PR #10, opened 2013-10-15
Adds base wercker config that installs drush

### LINKED ISSUE #6: Create a wercker manifest & project for unit testing
Once the base make file ins introduced, we should setup the basic wercker config and project for testing.  Minimally, this should run the core test suites, so the only requirement should be the drupal-core-org make file

desmondmorris:
Fixes #6


aaronschachter:
Sure

PR #21, opened 2013-10-15
Add drupal-org.make file for contrib modules and themes

### LINKED ISSUE #18: Setup drupal-org.make - first pass at contrib module usual suspects
Create the drupal-org.make file.

Lets add some of the modules we know that we will be using for sure.  It may make since to start figuring out how we document some of this stuff as well.  Remember to version lock

Here is a quick list I think we can be pretty sure of:
- ctools
- views
- features
- token
- pathauto
- libraries
- date
- admin_menu
- google_alnalytics
- optimizely
- jquery_update
- entity_api

aaronschachter:
Closes #18

### DIFF
--- a/drupal-org.make
+++ b/drupal-org.make
@@ -1,0 +1,79 @@
+; Contrib modules, themes and libraries
+
+api = "2"
+core = "7.x"
+
+; MODULES
+
+; Admin menu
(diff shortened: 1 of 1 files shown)


desmondmorris:
looks good

PR #22, opened 2013-10-15
Adds build-dosomething.make file

aaronschachter:
Adds a make file which installs the DoSomething profile and automatically downloads modules found in drupal-org.make


desmondmorris:
looks good

PR #26, opened 2013-10-15
Points ci scripts to build make file and dosomething profile

desmondmorris:


desmondmorris:
These changes are unnecessary as the CI tasks will be moving to our cerebro jenkins instance

See issue #35

PR #33, opened 2013-10-16
Use include notation to include core make file

### LINKED ISSUE #28: Use include notation in build make for core make
Instead of declaring the core project in the build make file, we can simply use this include notation
includes[] = drupal-core-org.make

aaronschachter:
Resolves #28

### DIFF
--- a/build-dosomething.make
+++ b/build-dosomething.make
@@ -1,8 +1,8 @@
 api = 2
-core = 7.23
+core = 7.x
 
-; Drupal
-projects[] = drupal
+; Include the definition for how to build Drupal core directly:
+includes[] = drupal-org-core.make
 
 ; Dosomething Profile
 projects[dosomething][type] = profile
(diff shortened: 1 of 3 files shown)


desmondmorris:
@aaronschachter I think you still need to set what the core version is

aaronschachter:
@desmondmorris it's been running for me locally... the core version gets set in the drupal-org-core.make file, so i removed it from this file.  can add it in though, if you prefer.

in the commerce kickstart make file, they use
api = 2
core = 7.x
in both the build-commerce-kickstart.make file

and then again in their drupal-org-code.make file:
api = 2
core = 7.x
projects[drupal][version] = 7.22

what do you think?

desmondmorris:
@aaronschachter yah I think per our convo the best bet is to denote '7.x' in all files and specify the core version only in core make file

aaronschachter:
@desmondmorris this PR has been updated accordingly

PR #40, opened 2013-10-16
Adds dsapi php library

### LINKED ISSUE #17: DSAPI PHP Library
Update the contrib make file to install the DSAPI PHP Library

desmondmorris:
Pulls in dsapi php client library from the make file

Resolves #17


aaronschachter:
ship it good

Issue #46, opened 2013-10-16
Build Script


desmondmorris:
We need a script to help us quickly build and rebuild our dev environment.  The immediate goal is to provide the following step[s]:
- build

This script should also create symlinks to the profile files, modules and themes so we can work on them in an actual instance

Some things to consider in the future, when we have environments and such:
- db sync
- db dump

PR #50, opened 2013-10-16
A simple build helper script.  This will evolve as necessary

desmondmorris:
This script will:
- Delete the existing html directory
- run drush make

We will want to make this play nice with settings and files symlinking in the future

Related to #46


aaronschachter:
Cool pic man.  ship it

Issue #41, opened 2013-10-16
Create "dosomething" subdir in the themes directory


desmondmorris:
Lets create a "dosomething" sub directory in the themes directory.  This will allow us to symlink that directory while developing locally.

PR #51, opened 2013-10-16
Adds symlinks to modules profile files

### LINKED ISSUE #46: Build Script

desmondmorris:
There is a dependency here #41 that should get fulfilled before we merge this 

Closes #46


aaronschachter:
looks good.

PR #54, opened 2013-10-17
Adds more core/contrib modules to install profile

aaronschachter:
Removes "node" from the list, as it's required by Drupal core.  Adds more stuff to the stuff.


desmondmorris:
Looks good.  Merging.

PR #56, opened 2013-10-17
Remove unused drupal-setup.sls

aaronschachter:
Legacy file from old vagrant setup.


mchitten:
Looks good.  There's actually more that can go but I'll hit them in a separate PR.

PR #58, opened 2013-10-17
Remove extra files; add note about install-not-working

mchitten:
- Installing is weird now.  Until it's properly fixed there's a new note on how to build your environment.
- Removes some unused files.


aaronschachter:
ship it

PR #59, opened 2013-10-17
Dosomething User prelim

mchitten:
Start of ~~DoSomething General~~ DoSomething User module.  Only has the `dosomething_user_valid_cell` function for now.

duvc has been rewritten to better understand phone numbers.  Of a sample of 10,000 phone numbers in the DB, it only missed 16, and only because they weren't valid phone numbers.

Tests that!


desmondmorris:
Looks good.  Merging

Issue #60, opened 2013-10-17
Installation error with build.sh


aaronschachter:
![screen shot 2013-10-17 at 3 19 42 pm](https://f.cloud.github.com/assets/1236811/1355314/f13e2396-3763-11e3-975c-84fd504a9573.png)

mchitten:
This appears to be fixed ... somehow.  It is, somehow, copying the files in the build script.
![profiles](https://f.cloud.github.com/assets/1836115/1553280/17ee1f1c-4e31-11e3-8344-18da1527f3a1.png)
So there you go.

PR #61, opened 2013-10-17
Removes profile symlinking for now. 

desmondmorris:
This is a hotfix that presents a temporary fix to #60 to get builds working properly again.

The symlinks for the profile specific files were not being referenced properly by the install script.


aaronschachter:
sounds good

PR #64, opened 2013-10-18
Better install-complete message.

mchitten:
Building with salt apparently stopped working because ssh agent forwarding doesn't work while vagrant is setting itself up.  That's not something we can fix.
- Removes "drush-make" step from salt file; replaces it with "db-prepare" step, that creates the database for your env.
- Provides a helpful message after `vagrant up` is complete, re: what to do next to finish building your site.
- Updates build.sh file to include some more set-up:
  - Includes `--install` parameter that runs:
    - `drush site-install`
    - `drush role-add-permission 'anonymous|authenticated user' 'access content'`
  - Symlink settings.php file from `/vagrant` to `sites/default`
  - `drush cc all`

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -50,8 +50,13 @@ rm -rf "$ABS_CALLPATH/$TARGET/profiles/dosomething/themes/dosomething"
 ln -s "$ABS_CALLPATH/themes/dosomething" "$ABS_CALLPATH/$TARGET/profiles/dosomething/themes/dosomething"
 
 # Clear caches and Run updates
-cd "$DRUPAL"
+cd "$TARGET"
+
+echo 'Installing site...'
+drush site-install dosomething -y --db-url=mysql://root@localhost/dosomething --site-name=DoSomething
+
 echo 'Clearing caches...'
+
 #drush cc all; drush cc all;
 #echo 'Running updates...'
 #drush updb -y;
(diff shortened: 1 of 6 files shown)


build.sh
```diff
@@ -50,9 +50,18 @@ rm -rf "$ABS_CALLPATH/$TARGET/profiles/dosomething/themes/dosomething"
 ln -s "$ABS_CALLPATH/themes/dosomething" "$ABS_CALLPATH/$TARGET/profiles/dosomething/themes/dosomething"
 
 # Clear caches and Run updates
-cd "$DRUPAL"
+cd "$TARGET"
+
```
desmondmorris, line 58: We should probably trigger this with an argument and some conditional logic.  We do not want to wipe the database every time we run the build script.

something like:
    sh build.sh --site-install

build.sh
```diff
@@ -50,9 +50,18 @@ rm -rf "$ABS_CALLPATH/$TARGET/profiles/dosomething/themes/dosomething"
...
+echo 'Installing site...'
+drush site-install dosomething -y --db-url=mysql://root@localhost/dosomething --site-name=DoSomething
+
+echo 'Granting basic permissions...'
+drush rap 'anonymous user' 'access content'
```
desmondmorris: drush rap.  I like that

aaronschachter:
@mchitten walked thru these changes with me. ship it

PR #107, opened 2013-11-06
Set variable for features default export path

### LINKED ISSUE #106: Installation: Set features default export path
The default path is set to sites/all/modules, which is no good for us when executing `drush fe` commands.

`variable_set('features_default_export_path', 'profiles/dosomething/modules/dosomething')`

aaronschachter:
Resolves #106


desmondmorris:
+1

Issue #115, opened 2013-11-08
Create new private github repo for internal dev scripts


aaronschachter:
This is where a import.sh would live, for example, to dl latest copy of database.

Issue #114, opened 2013-11-08
ds script cleanup: Remove "dbup" code


aaronschachter:
We want to make the dosomething repo public, so we definitely don't want to give the public details on how to download our DB. 

The "dbup" command isn't actually implemented and contains "coming soon" - but this issue is simply for removing all instances and documentation of it.

Issue #112, opened 2013-11-08
ds script cleanup: profile symlinks


aaronschachter:
There's commented symlinking of dosomething.info, dosomething.profile and dosomething.install file.

Let's either:
- fix it if it's broken
- remove if we can't do it, and update the wiki accordingly with how to make profile changes. https://github.com/DoSomething/dosomething/wiki/DoSomething-Profile

aaronschachter:
Related #60

mchitten:
Seems to be good now.

PR #123, opened 2013-11-08
Remove scripts from ds file

### LINKED ISSUE #113: ds script cleanup: commented drush commands
Lines 106-112 are commented out and contain TODO.  

Let's either fix and uncomment, or remove and update wiki with what needs to be done.

mchitten:
- Removes `dbup` and `dbdump` scripts from `ds` file.  There's a new `import.sh` file in the `Devel-Scripts` repository.
- Cleans up the file a bit.

Resolves #113, #114, #115.

Will need to investigate #112 re: symlinking profile.

### DIFF
--- a/ds
+++ b/ds
@@ -14,12 +14,6 @@ ds build [--install]
   --install
     Runs "drush site-install" to actually install the Drupal instance.
 
-ds dbup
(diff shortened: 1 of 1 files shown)


aaronschachter:
:+1:

Issue #68, opened 2013-10-19
Update varnish config to use file storage


mshmsh5000:
Right now, varnishd runs with 

`-s malloc,256m`

Not optimal in a 1024m environment.

Update to 

`-s file,/var/lib/varnish/$INSTANCE/varnish_storage.bin,512m`

(512m seems reasonable.) See

`/etc/default/varnish`

for details.

PR #124, opened 2013-11-08
Issue #68: Customizes Varnish startup defaults

mshmsh5000:
Issue #68: Configures Salt to manage etc/default/varnish; supplies new version of that file


mchitten:
Looks good.

PR #126, opened 2013-11-08
Doxygen

mchitten:
Ignores Doxygen files on Black Angus and comments the user function to test Doxgen stuff.


aaronschachter:
bodacious.

PR #127, opened 2013-11-11
Better phone # validation.

mchitten:
Slightly better phone number validation, including invalidation of consecutive-single-digit phone numbers (e.g. 999-999-9999) and "+1 (area code)" validation.


jonuy:
:+1: for the +1

... and everything else

PR #130, opened 2013-11-11
DS User Test for Administrator permissions

### LINKED ISSUE #109: dosomething_user: Web test for administrator permissions
- Create a dummy user and assign "administrator" role.
- Test that dummy user has permissions like "administer modules", "bypass node access", etc.

aaronschachter:
Closes #109
- Adds DosomethingUserWebTestCase to test that the "administrator" role grants the expected permissions to expected roles.  There's some good stuff in here that we may want to extract into our own DS Test class, mainly:
  - Each of our DrupalWebTestCase classes need the `protected $profile = 'dosomething'` set to work
  - Helper function for creating an administrator user
  - Helper function for testing user permissions with user_access

For now, keeping this in DoSomething User until we introduce more modules and need to extract out.
- Renames dosomething_user.unit.test into just dosomething_user.test and adds all classes into it, to keep our modules consistent with standard Drupal test file naming conventions.
- Comments out the `administer mailsystem` permission, because for some bizarre reason, the user_access check is failing.  Enabling the module shows the administrator does have the permission, so opening a separate issue up to get this code merged in sooner than later.

### DIFF
--- a/modules/dosomething/dosomething_user/dosomething_user.info
+++ b/modules/dosomething/dosomething_user/dosomething_user.info
@@ -52,4 +52,4 @@ features[user_permission][] = generate features
 features[user_permission][] = manage features
 features[user_permission][] = view the administration theme
 features[user_role][] = administrator
-files[] = dosomething_user.unit.test
+files[] = dosomething_user.test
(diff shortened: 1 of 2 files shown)


desmondmorris:
this looks good

Issue #43, opened 2013-10-16
Set paraneue base theme as default


desmondmorris:
Set the paraneue base theme as the default theme for now while we are in development.  Once it is in a good place, we will then need to create a proper subtheme.

PR #133, opened 2013-11-12
Moves paraneue to profile specific make file.  Downloads neue as library

desmondmorris:
Paraneue must be downloading form the profile specific make file.  This also downloads neue as an independent library

Related to #43


aaronschachter:
I think this PR may resolve https://github.com/DoSomething/dosomething/issues/145#issuecomment-28408351

PR #137, opened 2013-11-12
Resolves #98 by using Salt

### LINKED ISSUE #98: QA: Self-assigned SSL cert
For now. If we need a legit SSL cert for third-party interaction, we can go there.

mshmsh5000:
http is still on 8888; https is on 8889.

Auto-generated certs live in /etc/pki/salt.

Also refactors a few utility packages into utils.sls for clarity.


mchitten:
:+1:

desmondmorris:
It works!

screen shot 2013-11-13 at 11 09 22 am

ignore the error copy

PR #139, opened 2013-11-12
Blatantly bad

mchitten:
...


works.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
```
dosomething-machines, line 2: Missing file doc comment

dosomething-machines, line 2: Missing file doc comment

PR #148, opened 2013-11-13
Coder

mchitten:
- Includes `coder` module in make file.
- Sets up PHP CodeSniffer command as `codercs` in .bash_profile -- will be available after running `ds build [--install]`.
- Installs PHP_CodeSniffer pear package through salt.

Documentation for all of this coming soon.

### DIFF
--- a/drupal-org.make
+++ b/drupal-org.make
@@ -80,6 +80,10 @@ projects[devel][subdir] = "contrib"
 projects[environment_indicator] = "7.x-2.0"
 projects[environment_indicator][subdir] = "contrib"
 
+; Coder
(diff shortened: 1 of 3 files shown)


aaronschachter:
This looks good.. although the OCD in me questions whether or not to git rebase all the commits into a single one, to avoid the commits like "Blatantly bad" from showing up in the history.  It's one of those things that we either start to do it now, or don't do it at all.  Thoughts?  cc @desmondmorris

desmondmorris:
Yeah, I agree.  Would be good to get the test commits out of there.

aaronschachter:
:+1:

Issue #99, opened 2013-11-01
Switch to https for user login/edit/register


mshmsh5000:
Requires #98 for testing.

Techniques and discussion of [D7 Secure Pages](https://drupal.org/project/securepages) module status [here](http://drupal.stackexchange.com/questions/23149/redirect-user-login-to-https-all-other-requests-to-http).

We may not need Secure Pages, since we're constructing a custom auth flow; but that module's techniques for forcing requests to https are still useful.

DFurnes:
:heart:!

It might be worth defaulting to HTTPS for all pages on logged-in sessions to prevent session hijacking.

mshmsh5000:
Desmond suggested this in #98 as well. The session security concerns are largely mitigated by enforcing a secure cookie, but I agree that, performance and scalability considerations aside, it'd be nice to go secure for the whole site for authenticated users.

PR #158, opened 2013-11-19
99: Secure Pages install

mshmsh5000:
Resolves issue #99 by, duh, installing module, setting variables in dosomething.install, and updating Vagrant's port mapping.

### DIFF
--- a/dosomething.install
+++ b/dosomething.install
@@ -20,9 +20,29 @@ function dosomething_install_tasks($install_state) {
 function dosomething_configure() {
 	// Set admin theme to Seven:
   variable_set('admin_theme', 'seven');
+  
   // Use admin theme for editing content:
   variable_set('node_admin_theme', 1);
+  
   // Set features default export path to our custom module directory:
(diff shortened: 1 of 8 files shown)


mchitten:
Looks good

dosomething.install
```diff
@@ -20,9 +20,29 @@ function dosomething_install_tasks($install_state) {
...
+  
   // Set features default export path to our custom module directory:
   variable_set('features_default_export_path', 'profiles/dosomething/modules/dosomething');
+  
+  // Secure Pages settings:
+  variable_set('securepages_basepath', 'http://localhost:8888');
+  variable_set('securepages_basepath_ssl', 'https://localhost:8889');
+  variable_set('securepages_debug', 0);
+  variable_set('securepages_enable', '1');
+  variable_set('securepages_forms', 'user_login
```
desmondmorris, line 35: We may want to move these settings into settings.php in the future.  This way we have more control across different environments.

Issue #162, opened 2013-11-20
Remove major version prefixes from the drupal-org.make file


desmondmorris:
Example

```
; Admin menu
projects[admin_menu] = "7.x-3.0-rc4"
projects[admin_menu][subdir] = "contrib"
```

Should be

```
; Admin menu
projects[admin_menu] = "3.0-rc4"
projects[admin_menu][subdir] = "contrib"
```

Reference: https://drupal.org/comment/6667404#comment-6667404

aaronschachter:
Closed by #166

PR #166, opened 2013-11-20
Remove major version numbers from make file

angaither:
Resolves issue #162


desmondmorris:
@angaither Looks good.  I am going to merge.  We should change the author on those two commits

PR #169, opened 2013-11-21
Removes themes symlinking for now

desmondmorris:
This is borked currently.  Removing this symlink + pulling in paraneue via the make file, places the theme directories properly.


aaronschachter:
Sure

Issue #92, opened 2013-11-01
SMS: Install/enable suite of modules needed for SMS


jonuy:
- conductor
- conductor_sms (our custom implementation of it)
- dosomething_sms / sms_flow combo module
- add to make file

PR #170, opened 2013-11-21
Make file updated to include SMS modules and libraries

jonuy:
For issue #92 

Modules:
- conductor
- conductor_sms
- mobilecommons

Libraries:
- mobilecommons-php

### DIFF
--- a/drupal-org.make
+++ b/drupal-org.make
@@ -41,6 +41,10 @@ projects[mandrill][subdir] = "contrib"
 projects[metatag] = "1.0-beta7"
 projects[metatag][subdir] = "contrib"
 
+; Mobile Commons
+projects[mobilecommons] = "1.0"
+projects[mobilecommons][subdir] = "contrib"
+
 ; Module Filter
 projects[module_filter] = "1.8"
 projects[module_filter][subdir] = "contrib"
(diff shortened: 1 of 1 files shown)


drupal-org.make
```diff
@@ -74,6 +78,22 @@ projects[wysiwyg] = "2.2"
 projects[wysiwyg][subdir] = "contrib"
 projects[wysiwyg][patch][] = "https://drupal.org/files/wysiwyg-support_v4_ckeditor-1853550-46.patch"
 
+; GIT PROJECTS
+
+; Conductor
+projects[conductor][type] = "module"
+projects[conductor][download][type] = "git"
+projects[conductor][download][url] = "https://github.com/DoSomething/Conductor.git"
+projects[conductor][download][revision] = "ac9956d"
```
aaronschachter: Would it make sense to specify a branch instead of a revision?  More read-able?

jonuy: Ah, perhaps. I didn't know that was an option and was just sorta following what I saw in the drush make example file. 

If it's just the master branch, is it safe to not even bother specifying a revision or a branch?

aaronschachter: It might be safe if you won't be creating other branches in the repo... but might it make sense to start branching?  What's the game plan for these different repo's for the conductor and conductor_sms ?

desmondmorris: It will default to master.  So that line is uncessary

jonuy: Not much of a gameplan for those two modules right now. Hopefully they won't really require any changing. They were both modules created by a zivtech guy and managed on github only. But we made a few changes along the way that didn't go back upstream, hence the forked repo in the DS account.

PR #171, opened 2013-11-21
Enables paraneue theme by default. 

### LINKED ISSUE #43: Set paraneue base theme as default

desmondmorris:
Sets the default theme to paraneue. 

Closes #43

### DIFF
--- a/dosomething.install
+++ b/dosomething.install
@@ -18,11 +18,22 @@ function dosomething_install_tasks($install_state) {
  * Set up base configuration.
  */
 function dosomething_configure() {
-	// Set admin theme to Seven:
+  // Disable bartik
+  db_update('system')
+    ->fields(array('status' => 0))
+    ->condition('type', 'theme')
+    ->condition('name', 'bartik')
+    ->execute();
+
+  // Enable paraneue and the admin theme.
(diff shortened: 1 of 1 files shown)


dosomething.install
```diff
@@ -18,11 +18,29 @@ function dosomething_install_tasks($install_state) {
...
+
+  // Disable bartik
+  db_update('system')
+    ->fields(array('status' => 0))
+    ->condition('type', 'theme')
+    ->condition('name', 'bartik')
+    ->execute();
+
+  // Enable the admin theme.
+  db_update('system')
```
mchitten: Small thing, but why bother having two different `db_update`s to just run the same command on different themes? Wouldn't this work just as well?

``` php
db_update('system')
  ->fields(array('status' => 1))
  ->condition('type', 'theme')
  ->condition('name', array('paraneue', 'seven'), 'IN')
  ->execute();
```

desmondmorris: Si senor it will.  Just need an 'IN' in there too.  gracias.

PR #178, opened 2013-11-22
Add Entity contrib module

### LINKED ISSUE #159: Install and enable Entity API

aaronschachter:
We'll make use of the Entity contrib module for building out Campaigns.  This adds and enables it upon install.

Resolves #159


jonuy:
:+1:

PR #184, opened 2013-11-22
Add dosomething_campaign module

### LINKED ISSUE #31: Campaign: Stub module
Create a basic content type for campaigns.  This should be managed by a feature.  This should only be a stub as we do not have finalized comps/spec.

aaronschachter:
Provides a Campaign entity, which for now, just simply has an id and a title field.

Makes use the contrib Entity module to provide an admin UI for CRUD operations at /admin/campaigns.

Resolves #31

### DIFF
--- a/modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
+++ b/modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
@@ -1,0 +1,40 @@
+<?php
+/**
+ * @file
+ * Admin pages for dosomething_campaign.module.
+ */
+
+/**
+ * Form constructor for the Campaign entity form.
+ *
+ * @ingroup forms
+ */
+function campaign_form($form, &$form_state, $campaign, $op = 'edit') {
+  $form['title'] = array(
+    '#title' => t('Title'),
+    '#type' => 'textfield',
+    '#default_value' => isset($campaign->title) ? $campaign->title : '',
+    '#description' => t('Campaign title.'),
+    '#required' => TRUE,
+    '#weight' => -50,
+  );
+  $form['actions'] = array(
+    '#type' => 'actions',
+    'submit' => array(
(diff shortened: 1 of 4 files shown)


modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -0,0 +1,102 @@
...
+    'access callback' => TRUE,
+  );
+  return $items;
+}
+
+/**
+ * Access callback for CampaignEntity CRUD operations.
+ */
+function dosomething_campaign_access($op, $dosomething_campaign = NULL, $account = NULL) {
+  return TRUE;
```
deezone, line 53: I assume you're just fleshing things out as this access check seems pointless?

aaronschachter, line 53: Yup. There's a bunch more stuff to go into this module, but I'm adding it as is to get it set up so we can all jump in and build different aspects of it out.

modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
```diff
@@ -0,0 +1,40 @@
...
+}
+
+/**
+ * Form submission handler for campaign_form().
+ *
+ * @see campaign_form()
+ */
+function campaign_form_submit($form, &$form_state) {
+  $campaign = entity_ui_form_submit_build_entity($form, $form_state);
+  $campaign->save();
```
deezone, line 38: Does entity_ui_form_submit_build_entity() take care of all of CRUD? How are entries deleted?

aaronschachter, line 38: This form is only for adding and editing.  For deleting (for now), there's a "delete" link in the overview page, on /admin/campaign.  It looks like the way the node overview or taxonomy term list does.

PR #193, opened 2013-11-25
Campaign properties

### LINKED ISSUE #192: Campaign: Add properties for created, updated, uid, status
Autopopulate created, updated, and uid values

aaronschachter:
Populates the created, updated, and uid properties on a Campaign.
- Overrides Entity API `save()` method to save created, updated, and values on submitting campaign edit form
- Overrides overviewTable() to display the "changed" value as a column and sort by it

Resolves #192

### DIFF
--- a/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -112,6 +112,56 @@ class CampaignEntityController extends EntityAPIController {
  * Our custom controller for the admin ui.
  */
 class CampaignEntityUIController extends EntityDefaultUIController {
+
+  /**
+   * Generates the render array for a overview table for arbitrary entities
+   * matching the given conditions.
+   *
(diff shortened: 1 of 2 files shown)


modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -92,6 +92,20 @@ class CampaignEntity extends Entity {
  */
 class CampaignEntityController extends EntityAPIController {
 
+  /**
+   * Overrides save() method.
+   *
+   * Populates created, updated, and uid automatically.
+   */
+  public function save($entity, DatabaseTransaction $transaction = NULL) {
+    if (isset($entity->is_new)) {
```
desmondmorris, line 102: set the REQUEST_TIME in a variable so the created and changed values are not different for new entities.

$request_time = REQUEST_TIME;

desmondmorris:
everything looks good.  the above comment should be considered.

PR #196, opened 2013-11-25
Enables dosomething campaigns & taxonomy modules from the installation profile

desmondmorris:
@aaronschachter Any reason why we cannot enable these now?  It will make it easier for others to jump and contribute to both


aaronschachter:
go for it :+1:

Issue #189, opened 2013-11-25
Campaign: Install schema + Edit form


aaronschachter:
Add simple value fields into the install hook, and update the `campaign_form` function accordingly to store.

aaronschachter:
Progress being tracked here: https://docs.google.com/a/dosomething.org/spreadsheet/ccc?key=0AnDHGfy1v5bJdDk3NFIycGpNYmUxYmt0ZDVMdFNTUGc&usp=drive_web#gid=0

PR #197, opened 2013-11-25
Campaign column start + Admin settings form

aaronschachter:
Adds a "CTA Text" column and columns / fieldset for Timing in both the table schema, and admin `campaign_form`.

Also adds an admin settings form to store form descriptions in variables via admin settings form (instead of hardcoding descriptions).

Begins work for #189


desmondmorris:
This looks good.

Issue #91, opened 2013-11-01
SMS: consolidate dosomething_sms amd sms_flow modules into a single module


jonuy:
More likely bring `sms_flow` functionality into `dosomething_sms` to clarify the code base and remove any ambiguity about where SMS code lives.

Also, remove any workflows and endpoints no longer used.

jonuy:
This ended up getting handled by #981 and #1038

PR #200, opened 2013-11-25
Beginning of the great SMS code migration

### LINKED ISSUE #90: SMS: Purge private values from dosomething_sms and sms_flow code
Before migrating code over, transfer private values over to whatever solution comes out of #75

jonuy:
- Previously both dosomething_sms and sms_flow modules were used as homes to SMS code. This is all now consolidated again in the dosomething_sms module alone.
- Previous methods also relied on the sms_mobile_commons and smsframework modules. Those dependencies have been removed and we're just using the mobilecommons module in conjunction with the mobilecommons-php library to make our calls to - you guessed it - Mobile Commons.
  - Using the mobilecommons module also solves how we'll set private authentication credentials.
- I also started the process of bringing over some of our ConductorActivity plugins - everything under dosomething_sms/plugins/activity/misc. I only brought over the ones so far that rely on Mobile Commons API calls.

Resolves #90, #91

### DIFF
--- a/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
+++ b/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
@@ -1,0 +1,83 @@
+<?php
+/**
+ * @file
+ * Admin page callback for the DoSomething SMS module.
+ */
+
+/**
+ * Builds and returns the Optimizely Add/Update form.
+ * If there are only 3 arguments in the path it builds the add form and it adds a record.
+ * Otherwise it builds the update form where the fourth argument is the record ID in
+ * in the optimizely table.
+ */
+function dosomething_sms_settings_form() {
+
+  $form = array();
+
+  $form['configuration'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Configuration of DoSomething.org SMS cron()'),
+  );
+
+  $form['configuration']['dosomething_sms_cron_enable'] = array(
+    '#type' => 'radios',
+    '#title' => t('Cron Status'),
+    '#options' => array(
+      TRUE => t('Enabled'),
+      FALSE => t('Disabled'),
+    ),
+    '#default_value' => variable_get('dosomething_sms_cron_enable', FALSE),
+    '#weight' => 10,
+  );
+
+  $form['configuration']['dosomething_sms_cron_interval'] = array(
+    '#type' => 'select',
+    '#title' => t('Cron interval'),
+    '#description' => t('Time after which cron_example_cron will respond to a processing request.'),
+    '#default_value' => variable_get('dosomething_sms_cron_interval', 60 * 60 * 24 * 7),
+    '#options' => array(
+      60 * 60 => t('1 hour'),
+      60 * 60 * 24 * 1 => t('1 day'),
+      60 * 60 * 24 * 2 => t('2 days'),
+      60 * 60 * 24 * 5 => t('5 days'),
+      60 * 60 * 24 * 7 => t('1 week'),
+      60 * 60 * 24 * 7 * 2 => t('2 weeks'),
+      60 * 60 * 24 * 7 * 4 => t('4 weeks'),
+    ),
+    '#weight' => 20,
+  );
+
+  // Where to send the report
+  $form['configuration']['dosomething_sms_cron_recipients'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Email Recipients'),
+    '#description' => t('Seperate email addresses with commas.'),
+    '#default_value' => variable_get('dosomething_sms_cron_recipients', 'dlee@dosomething.org'),
+    '#size' => 60,
+    '#maxlength' => 256,
+    '#required' => TRUE,
+    '#weight' => 30,
+  );
+
+  $form['status'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('DoSomething.org SMS Cron Status'),
+  );
+
+  // Run job every interval (week by default)
+  $interval = variable_get('dosomething_sms_cron_interval', 60 * 60 * 24 * 7);
+  $next_run = variable_get('dosomething_sms_cron_next_execution', REQUEST_TIME);
+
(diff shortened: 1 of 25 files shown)


modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
```diff
@@ -0,0 +1,83 @@
+<?php
+/**
+ * @file
+ * Admin page callback for the DoSomething SMS module.
+ */
+
+/**
+ * Builds and returns the Optimizely Add/Update form.
```
aaronschachter, line 8: No it doesn't.

modules/dosomething/dosomething_sms/dosomething_sms.install
```diff
@@ -0,0 +1,139 @@
...
+}
+
+/**
+ * Implements hook_uninstall().
+ */
+function dosomething_sms_uninstall() {
+  db_drop_field('conductor_instance_pointer', 'timestamp');
+  drupal_set_message('dosomething_sms: The "timestamp" field has been removed from the conductor_instance_pointer table.', 'status');
+
+  db_drop_table('sms_flow_game');
```
aaronschachter, line 114: We don't need these `db_drop_table`'s in Drupal 7 -- uninstalling a module automatically drops its tables defined its `hook_schema`.

modules/dosomething/dosomething_sms/dosomething_sms.module
```diff
@@ -0,0 +1,810 @@
...
+ * Implements hook_ctools_plugin_directory().
+ */
+function dosomething_sms_ctools_plugin_directory($owner, $plugin_type) {
+  if ($owner == 'conductor') {
+    return "plugins/$plugin_type";
+  }
+}
+
+/**
+ * Impliments hook_cron().
```
aaronschachter, line 66: Misspelled.

modules/dosomething/dosomething_sms/dosomething_sms.module
```diff
@@ -0,0 +1,810 @@
...
+ */
+function dosomething_sms_cron() {
+  dosomething_sms_cron_test();
+}
+
+/**
+ * hook_cron() functionality that can be called as a cron job or as a test.
+ */
+function dosomething_sms_cron_test($manual_test=FALSE) {
```
aaronschachter, line 75: This is picky, but let's get into the habit of this.  Variable declarations should have spaces between the ='s `$manual_test = FALSE` https://drupal.org/coding-standards#operators

modules/dosomething/dosomething_sms/dosomething_sms.module
```diff
@@ -0,0 +1,810 @@
...
+    'administer dosomething_sms' => array(
+      'title' => t('Administer dosomething_sms module'),
+      'description' => t('Administer access to everything in module'),
+      'restrict access' => TRUE,
+    ),
+  );
+}
+
+/**
+ * Impliments hook_theme().
```
aaronschachter, line 258: Misspelled.

modules/dosomething/dosomething_sms/dosomething_sms.module
```diff
@@ -0,0 +1,810 @@
...
+ *   '555-555-5555', 128441, 128441, array('field_webform_mobile'),
+ *   $form_state['values'],
+ *   array(
+ *     'first_name' => 'Mike',
+ *     'club_name' => "Mike's cool DoSomething.org Club."
+ *   )
+ * );
+ * @endcode
+ */
+function dosomething_sms_flow_start($alpha_number, $alpha_optin, $beta_optin, $beta_numbers, $form_values, $beta_variables, $beta_to_alpha_campaign = FALSE, $type_override = NULL) {
```
aaronschachter, line 308: Wows, such parameters.  

Not mandatory but it might make sense to pass through an array of all these params instead of as individual parameters - where you have to worry about what order you're passing them through.

modules/dosomething/dosomething_sms/dosomething_sms.module
```diff
@@ -0,0 +1,810 @@
...
+ *   have not been any messages sent from beta -> alpha.
+ */
+function dosomething_sms_alpha_received_message($alpha, $nid, $type) {
+  // Query the sms_flow_records table.
+  $select = db_query("
+    SELECT
+      COUNT(*) AS sent
+    FROM sms_flow_records
+    WHERE `accepted_invite` = 1
+      AND REPLACE(
```
aaronschachter, line 440: What is going on here?

modules/dosomething/dosomething_sms/dosomething_sms.module
mshmsh5000, line 308: Also a better pattern in D7, which started explicitly capturing params as param arrays in a lot of places (e.g., [theme()](https://api.drupal.org/api/drupal/includes%21theme.inc/function/theme/7)).

modules/dosomething/dosomething_sms/dosomething_sms.module
mshmsh5000, line 440: Looks like post-facto cleaning up of the phone numbers. But looking at this table, there are only a few hundred with bad numbers, and they're the oldest records. I assume that this is because we've got better logic in place to normalize these data before it gets into the table.

Can we kill this, then?

modules/dosomething/dosomething_sms/dosomething_sms.module
```diff
@@ -0,0 +1,810 @@
...
+    else {
+      unset($active_workflows[$row->workflow_name]);
+    }
+  }
+
+  return $active_workflows;
+}
+
+/**
+ * Helper function to find all files within a directory that end with the given extension.
```
mshmsh5000, line 787: Do we have a utility module to stick this in?

desmondmorris, line 787: +1 - If possible, I think we should start maintaining a PHP library for our reusable helper and utility functions.  This way we can reuse it across multiple projects.  This way we avoid a dosomething_general || utility module, which is very difficult to maintain.

Issue #173, opened 2013-11-21
Campaign: Pitch and Action templates


aaronschachter:
Create stub files for the campaign entity's Pitch and Action templates:
- campaign-pitch.tpl.php
- campaign-action.tpl.php

For now, use a query string variable in order to switch between templates, and render basic Campaign fields such as Title.

PR #202, opened 2013-11-26
Campaign view entity and inc file

aaronschachter:
Overrides the `buildContent()` method in the `CampaignEntityController` to dynamically add all properties of the entity object on the `campaign/%campaign` view page.  

Still doesn't look pretty, but it's a way for editors to view content without opening the form (and is more readable than the existing `print_r`content).

Lays out foundation for #173 

Also moves the campaign entity classes into a separate inc file, to separate the code out and make it easier for multiple dev's to contribute.


modules/dosomething/dosomething_campaign/dosomething_campaign.info
```diff
@@ -3,3 +3,4 @@ description = Provides a custom Campaign Entity and its functionality.
 core = 7.x
 package = DoSomething
 dependencies[] = entity
+files[] = includes/dosomething_campaign.inc
```
desmondmorris, line 6: +1 for entity class stuff

desmondmorris:
Def good for now.

Issue #194, opened 2013-11-25
CRUD tests against Campaign entity


desmondmorris:
Implement baseline tests against Campaign entity CRUD functionality

aaronschachter:
Closing this, as we're rebuilding the campaign as a node.

PR #221, opened 2013-12-03
Add new columns into testCampaignForm

aaronschachter:
Tests that the newly added columns are saving correctly.

Adds tests for #194


angaither:
:+1:

PR #223, opened 2013-12-04
Split out Field API form elements in Campaign Form

### LINKED ISSUE #214: Campaign: Split out Field API fields on Campaign Form
Currently using Lullabot's method of using the Entity API Admin UI form, and calling `field_attach_form('campaign', $campaign, $form, $form_state)` to render all of the Field APi fields.  

However, this doesn't allow us to place the individual fields into the fieldsets, it just renders all the Field API fields at the bottom of the form.

This issue is to figure out how to render the fields inside the fieldsets they belong to, as per https://docs.google.com/a/dosomething.org/spreadsheet/ccc?key=0AnDHGfy1v5bJdDk3NFIycGpNYmUxYmt0ZDVMdFNTUGc&usp=drive_web#gid=0

aaronschachter:
Resolves #214

Found [this trick](http://stackoverflow.com/questions/18794648/custom-submit-handler-for-two-combined-forms-workflow-and-node-edit-forms) in order to render the separate Field API Field elements where we want in the Campaign Edit form.

### DIFF
--- a/modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
+++ b/modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
@@ -10,8 +10,15 @@
  * @ingroup forms
  */
 function campaign_form($form, &$form_state, $campaign, $op = 'edit') {
-  // Prefix for form description variables:
+
+  // Prefix for Entity column description variables:
   $desc_prefix = DOSOMETHING_CAMPAIGN_FORM_DESC_PREFIX;
+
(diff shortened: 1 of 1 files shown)


jonuy:
Looks fine.

Question though, is the ':' at the end of some of those single line comments something we should all be doing when we write similar comments?

aaronschachter:
I believe if you run the Coder Love Tough, it will complain about line comments that don't end in a punctuation mark, and so that created my habit. Per https://drupal.org/coding-standards/docs#drupal, comments should be complete sentences.

PR #225, opened 2013-12-04
Tags taxonomy vocabulary

### LINKED ISSUE #210: Taxonomy: Tags vocabulary
Needed for Campaigns, but possibly for other content as well.

I'm thinking that the DoSomething taxonomy feature should store the vocabulary, however it should not actually export the individual terms.  Any thoughts?

aaronschachter:
Let there be Tags.

Resolves #210


mchitten:
:+1:

PR #226, opened 2013-12-04
Campaign Tags auto-complete

### LINKED ISSUE #211: Campaign: Tags field
Depends on #210 

A free form text field, where editors can add new terms from the textbox.  Like tumblr or wordpress, or the internet.

Product Team will be responsible for checking terms for spelling errors, and all that.

aaronschachter:
Resolves #211

Creates a field_tags field on the campaign entity, and adds an auto-complete widget to the campaign form.

This also cleans up code committed for #214 - the easier way seems to be to use field_attach_form for the current `$form` variable instead of creating the duplicate, and just using the $options parameter to specify the field to attach.


mchitten:
Looks fine.

PR #231, opened 2013-12-05
Start of compound text module.

angaither:
@aaronschachter ehhh

### DIFF
--- a/ds
+++ b/ds
@@ -61,9 +61,8 @@ if [[ $1 == "build" ]]
     rm -rf "$ABS_CALLPATH/$TARGET/profiles/dosomething/modules/dosomething"
     ln -s "$ABS_CALLPATH/modules/dosomething" "$ABS_CALLPATH/$TARGET/profiles/dosomething/modules/dosomething"
 
-    # echo 'Linking themes'
-    # rm -rf "$ABS_CALLPATH/$TARGET/profiles/dosomething/themes/dosomething"
-    # ln -s "$ABS_CALLPATH/themes/dosomething" "$ABS_CALLPATH/$TARGET/profiles/dosomething/themes/dosomething"
+    echo 'Linking themes'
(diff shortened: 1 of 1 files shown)


modules/dosomething/dosomething_compoundtext/dosomething_compoundtext.install
```diff
@@ -0,0 +1,26 @@
...
+   return array(
+    'columns' => array(
+      'ds_title' => array(
+        'type' => 'varchar',
+        'length' => 255,
+        'not null' => FALSE,
+      ),
+      'ds_copy' => array(
+        'type' => 'varchar',
+        'length' => 99999,
```
mchitten: I think when you reach a varchar of length 99,999 you'd be safer to go with some sort of text field.

mchitten: Addendum to this: I'm not actually sure that it's possible to have a varchar of length 99,999.  It seems that the [max length is 65,536](http://www.pythian.com/blog/text-vs-varchar/#more-7129)...which is still kind of ridiculous.

desmondmorris: +1 for text type

modules/dosomething/dosomething_compoundtext/dosomething_compoundtext.install
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+
+/**
+ * Implements hook_field_schema().
+ */
+function dosomething_compoundtext_field_schema($field) {
+   return array(
+    'columns' => array(
+      'ds_title' => array(
```
aaronschachter: I think the code will end up being more legible if calling these columns text1 and text2, or value1 and value2, vs having the ds_ prefix.

PR #233, opened 2013-12-05
Restores wercker build steps and disables ds-api library pull for now

desmondmorris:
Disabling the ds-api library for now.  Since the make file clones multiple private repos, it is necessary to add ssh keys to the build pipeline for the both.  Currently the add-ssh-key build step does not accommodate this properly.


aaronschachter:
sure

PR #236, opened 2013-12-06
Campaign View Modes for Pitch and Action

aaronschachter:
Creates two custom view modes for the Campaign Entity: "Pitch" and "Action", and adds a Signup "Do This") Form into the Pitch view mode (form submit is not functional yet).

The action page can be viewed at campaign/[id], and the pitch page can be viewed by navigating to campaign/[id]/pitch.  This page callback will be updated to display the view_mode based on user signup status once signups are functional.

Also implements hook_preprocess_entity in order to allow for naming convention: campaign--pitch.tpl.php and campaign--action.tpl.php.

Works towards #173 -- all we need to do is actually create the tpl files in order to close it.

### DIFF
--- a/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -123,7 +123,7 @@ function dosomething_campaign_view_entity($campaign, $view_mode = 'action') {
 function dosomething_campaign_signup_form($form, &$form_state) {
   $form['submit'] = array(
     '#type' => 'submit',
-    '#value' => 'Do This',
+    '#value' => t('Do This'),
   );
   return $form;
 }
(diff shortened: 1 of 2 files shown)


modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -98,8 +111,30 @@ function campaign_load($id) {
...
+  return entity_view('campaign', array($campaign->id => $campaign), $view_mode);
+}
+
+/**
+ * Form constructor for the Campaign Signup form.
+ */
+function dosomething_campaign_signup_form($form, &$form_state) {
+  $form['submit'] = array(
+    '#type' => 'submit',
+    '#value' => 'Do This',
```
desmondmorris: t() please

desmondmorris:
do it!

Issue #251, opened 2013-12-11
Set base url & hostname


desmondmorris:
In order for securepages to work properly we will need to set the base_url and hostname in vagrant.

PR #252, opened 2013-12-11
Sets base_url and hostname to dev.dosomething.org

desmondmorris:
In order for this to work, you must add the following to the host machines hosts file:

```
127.0.0.1 dev.dosomething.org
```

References #251


desmondmorris:
We will need to make the securepages and base_url settings environment aware in the furture

PR #253, opened 2013-12-11
Adds environment aware configuration settings in settings.php. Fixes #75

### LINKED ISSUE #75: Dev / QA configuration variables
Programatic way to set variables like Google Analytics key.  Possibilities include in settings.php, or in the profile install hook.  

Could be useful for different variables for different environments?

desmondmorris:
Depends on #252


aaronschachter:
Looks good

PR #256, opened 2013-12-12
Adds varnish module and basic configuration settings. Fixes #104

### LINKED ISSUE #104: Actual Varnish-Drupal integration
https://drupal.org/project/varnish

7.x is still in beta -- amazing. But the benefit here is we don't have redundant Drupal page cache.

desmondmorris:


mshmsh5000:
How about [expire](https://drupal.org/project/expire) as an adjunct to Varnish? I've been playing with it, and it seems stable in 7.x.

desmondmorris:
@mshmsh5000 I agree expire would be a good addition.  I say lets create a new issue for it as it is not a dependency of or dependent on this patch

PR #270, opened 2013-12-13
Campaign tpl vars

### LINKED ISSUE #264: Campaign: Preprocess vars to include Compound text values
Add compound text vars into `dosomething_campaign_preprocess_entity`to allow for easy output in theme layer.

aaronschachter:
Loops through Compound Text field values, Taxonomy terms and outputs them onto the campaign.tpl.php.

Resolves #264

### DIFF
--- a/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -142,29 +142,21 @@ function dosomething_campaign_preprocess_entity(&$variables) {
     // Store the campaign entity to quickly reference its properties within the tpl:
     $campaign = $variables['elements']['#entity'];
     $variables['campaign'] = $campaign;
+    $compound_text_fields = array('faq', 'step_pre', 'step_post');
     // Loop through $campaign Entity to get to Compound Text values.
-    // Currently, DS Compound Text field types aren't visible to the entity_meta_wrapper().
-    $variables['campaign']->faq = array();
-    foreach ($campaign->field_faq[LANGUAGE_NONE] as $faq) {
-      $variables['campaign']->faq[] = array(
-        'title' => $faq['title'],
-        'body' => $faq['body'],
-      );
+    foreach ($compound_text_fields as $field) {
+      $variables['campaign']->{$field} = array();
+      // Currently, DS Compound Text field types aren't visible to the entity_meta_wrapper(),
+      // So we need to use LANGUAGE_NONE constant.
+      $field_name = 'field_' . $field;
+      foreach ($campaign->{$field_name}[LANGUAGE_NONE] as $value) {
+        $variables['campaign']->{$field}[] = array(
+          'title' => $value['title'],
+          'body' => $value['body'],
+        );
+      }
     }
-    $variables['campaign']->steps_pre = array();
-    foreach ($campaign->field_step_pre[LANGUAGE_NONE] as $step) {
-      $variables['campaign']->steps_pre[] = array(
-        'title' => $step['title'],
-        'body' => $step['body'],
-      );
-    }
-    $variables['campaign']->steps_post = array();
-    foreach ($campaign->field_step_post[LANGUAGE_NONE] as $step) {
-      $variables['campaign']->steps_post[] = array(
-        'title' => $step['title'],
-        'body' => $step['body'],
-      );
-    }
+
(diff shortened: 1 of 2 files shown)


modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -140,19 +140,61 @@ function dosomething_campaign_preprocess_entity(&$variables) {
   // If this is a campaign entity, create variables to simplify theming.
   if ($entity_type == 'campaign') {
     // Store the campaign entity to quickly reference its properties within the tpl:
-    $variables['campaign'] = $variables['elements']['#entity'];
-    $campaign = entity_metadata_wrapper('campaign', $variables['elements']['#entity']);
+    $campaign = $variables['elements']['#entity'];
+    $variables['campaign'] = $campaign;
+    // Loop through $campaign Entity to get to Compound Text values.
+    // Currently, DS Compound Text field types aren't visible to the entity_meta_wrapper().
+    $variables['campaign']->faq = array();
```
desmondmorris: The faq, steps_pre and steps_post can all be handled by  single loop.  Give your self a whitelist and iterate over that first;

```
$fields = array('faq', 'steps_pre', 'steps_post');

foreach($fields as $field) {
  if (!isset($variables['campaign']->$$field) && !is_array($variables['campaign']->$$field)) {
    $variables['campaign']->$$field = array();
  }

  $variables['campaign']->$$field[] = array(
    'title' => $field['title'],
    'body' => $field['body'],
  );  

}
```

This was not tested, but should get you close

aaronschachter:
Ready for re-review: Updated dosomething_campaign_preprocess_entity per @desmondmorris 's suggestion to loop through compound text and taxonomy fields values via a single loop for each field type.

themes/dosomething/paraneue_dosomething/templates/campaign/campaign.tpl.php
```diff
@@ -1,14 +1,70 @@
...
+	  <?php foreach ($campaign->faq as $faq): ?>
+		  <h4><?php print $faq['title']; ?></h4>
+		  <?php print $faq['body']; ?></p>
+	  <?php endforeach; ?>
+	  <?php endif; ?>
 	</div>
-	<?php if (!empty($gallery)): ?>
+
+	<div id="prep">
+		<h2>Prep</h2>
```
angaither, line 18: can we grab these titles from the header sections instead?

themes/dosomething/paraneue_dosomething/templates/campaign/campaign.tpl.php
```diff
@@ -1,14 +1,70 @@
...
+		  <?php print $step['body']; ?></p>
+	  <?php endforeach; ?>
+	  <?php endif; ?>
+	</div>
+
+	<?php if (!empty($campaign->action_type)): ?>
+  <h4>Action Type</h4>
+  <ul>
+  <?php foreach ($campaign->action_type as $term): ?>
+    <li><?php print l($term['name'], 'taxonomy/term/' . $term['tid']); ?></li>
```
angaither: send these in as a part of the variables array already formatted as links?

themes/dosomething/paraneue_dosomething/templates/campaign/campaign.tpl.php
aaronschachter, line 18: Which header section do you mean?

themes/dosomething/paraneue_dosomething/templates/campaign/campaign.tpl.php
aaronschachter: Good call, making this change

themes/dosomething/paraneue_dosomething/templates/campaign/campaign.tpl.php
angaither, line 18: from the campaign admin form headers.
Maybe overthinking it, but it would be good to have those stored in once place instead of two 
  '#title' => t('Know It'),

aaronschachter, line 18: @angaither I think we should be ok for now... this isn't the finalized tpl file but just a place to output all of the field values.  To do this ideally we'd add new variables. We'll need a cleaner way to handle all the variables we're using on the form, so for now I'm going to open an issue for it and document this there.

PR #282, opened 2013-12-18
Fix subdirectory for double_field

### LINKED ISSUE #281: Fix subdir for Double Field in drupal-org.make
Breaks the things.

angaither:
Fixes #281


aaronschachter:
:+1:

PR #288, opened 2013-12-19
dosomething_fact module

### LINKED ISSUE #265: Campaign: Facts
We need a similar field type for Facts, with Value / Source columns.  This will be used in as a "field_campaign_facts" field in the Campaign entity.

I've discussed building this out as an Entity with @desmondmorris and Fantini, where the Facts field would be an entity reference.  Although this is a clean solution on the back end, it would be tricky to implement as a select / add widget on the form itself.  For now, let's just implement as a Fact field type and create a mutli-value field on the Campaign entity.

aaronschachter:
Creates a Fact entity, and places entity reference fields on the campaign.

Uses ctools modal to place an "add fact" form on the campaign form.

Resolves #265


modules/dosomething/dosomething_fact/dosomething_fact.info
```diff
@@ -0,0 +1,6 @@
+name = DoSomething Fact
+description = Provides a custom Fact Entity and its functionality.
+core = 7.x
+package = DoSomething
+dependencies[] = entity
```
angaither, line 5: entity reference?

modules/dosomething/dosomething_fact/dosomething_fact.module
```diff
@@ -0,0 +1,132 @@
...
+    'title callback' => t('Fact'),
+    'page callback' => 'dosomething_fact_view_entity',
+    'page arguments' => array(1, 2),
+    'access callback' => TRUE,
+  );
+  // This menu callback needs to be visible by SecurePages in order for ctools modal to work.
+  $items['admin/dosomething-fact/modal-add/%ctools_js'] = array(
+    'page callback' => 'dosomething_fact_modal_add_callback',
+    'page arguments' => array(3),
+    'access callback' => TRUE,
```
angaither, line 50: permissions?

modules/dosomething/dosomething_fact/dosomething_fact.info
aaronschachter, line 5: It's been added as a dependency to dosomething_campaign, but isn't necessary for the fact.  The fact just has a few simple columns on the table, and depends on Entity to define the Fact entity.

angaither:
Other than those two comments, looks ok. :thumbsup:

modules/dosomething/dosomething_fact/dosomething_fact.module
aaronschachter, line 50: Good catch.  This will be handled by #289

PR #292, opened 2013-12-19
Varnish envvar #Fixes 291

desmondmorris:
This is ready to go.

This requires a "vagrant provision"


desmondmorris:
This needs moar love.  The env var wont stick.

@mshmsh5000 thoughts?

mshmsh5000:
@desmondmorris 

### Idea 1

How about overwriting the varnish secret after install? Then you can hard-code the value in both places.

`/etc/varnish/secret` can be a managed file in salt: https://gist.github.com/mshmsh5000/8049305

### Idea 2

If you're already busting the secret-ness of the secret file, just make it readable in Salt. Then settings.php can dance like

``` php
$varnish_secret_file = '/etc/varnish/secret';
if (is_readable($varnish_secret_file)) {
  $conf['varnish_control_key'] = file_get_contents($varnish_secret_file);
} else {
  // Panic
}
```

desmondmorris:
I like you bc you are smart.

desmondmorris:
That was not supposed to be sarcastic :) 

My vote is for #1 as I think that the file_get_contents is more expensive than reading the environment variable.

desmondmorris:
@mshmsh5000 a quick sanity check.  I ended generating a has value and adding it to a "secret: file managed by salt, as you suggested.

This will be one of those values that will need to be changed in production environments

mshmsh5000:
Looks good. The varnish conf is moving to the [settings.local.php file](https://github.com/DoSomething/dosomething/pull/574). We can fix after we get these two PRs in.

+1 ship it value add

PR #310, opened 2014-01-03
README for generating Facts/Campaigns

aaronschachter:
Adds README files for the DS Fact and DS Campaign Devel submodules.

Cleans up README's in the /modules and /themes directories by removing the 2nd READMEs in the modules/dosomething and themes/dosomething directories.


angaither:
:+1: looks good.

PR #314, opened 2014-01-03
Fact permissions / access callbacks

aaronschachter:
Creates basic permissions and checks them on relevant facts admin / view pages.


angaither:
:shipit: ship it!

PR #320, opened 2014-01-03
Fact feature: user permissions

aaronschachter:
Stores user permissions for fact CRUD in dosomething_fact module.


modules/dosomething/dosomething_fact/dosomething_fact.features.user_permission.inc
```diff
@@ -0,0 +1,50 @@
...
+      'administrator' => 'administrator',
+    ),
+    'module' => 'dosomething_fact',
+  );
+
+  // Exported permission: 'view fact'.
+  $permissions['view fact'] = array(
+    'name' => 'view fact',
+    'roles' => array(
+      'administrator' => 'administrator',
```
angaither, line 44: shouldn't anyone be able to see a fact?

aaronschachter, line 44: Nope, this is intentional.  The only users who should be able to see fact/50 are staff -- these pages will list out what campaigns reference the given fact.

The campaign preprocess loads the fact values into the template file, so an anon/regular auth user never  needs to see the fact view.

angaither:
:+1:

PR #328, opened 2014-01-06
Node campaign fields

angaither:
Refs #322


modules/dosomething/dosomething_campaign_node/dosomething_campaign_node.info
```diff
@@ -0,0 +1,89 @@
+name = dosomething_campaign_node
+description = Campaign node type
+core = 7.x
+package = Features
+version = 7.x-1.0
+project = dosomething_campaign_node
+dependencies[] = ctools
+dependencies[] = date
+dependencies[] = dosomething_taxonomy
```
aaronschachter, line 10: Let's add dosomething_fact as a dependency too, otherwise the fact entityreference fields will break.  Guess features aint that smart.

aaronschachter:
Looks great!  :+1:

Issue #317, opened 2014-01-03
Fact: Render source url as link in buildContent


aaronschachter:
Alter `buildContent` in `FactEntityController` to output  the url_source as a hyperlink.

Pseudo code: `<a href="url_source" target="_blank">url_source</a>`

aaronschachter:
This was closed by #339 -- @DeeZone reminder to use proper closing syntax https://help.github.com/articles/closing-issues-via-commit-messages

PR #339, opened 2014-01-07
DoSomething Facts - url_source as link

deezone:
Fixes issue #317

-> Adds l() to url_source #markup output when buildContent is called in FactEntityController class
-> minor changes to pass "coder" (codercs) review.


aaronschachter:
:+1:

PR #348, opened 2014-01-09
Removes capistrano deploy configuration files and other related files

desmondmorris:
Removing the capistrano stuff as we are not using it at this point. There is no plan in the near future to deploy via cap, with the Pantheon deployment work in the horizon


aaronschachter:
:+1:

PR #350, opened 2014-01-09
Moves post provision install message into salt directory for tidyness

desmondmorris:
To clean up the main directory a little bit, I move the post install provision instructions into the salt directory, where i think it may make more sense.  This is all in an effort to clean up our workspace a bit.


desmondmorris:
@aaronschachter thoughts on this one?

aaronschachter:
I like it.  Ship it

Issue #362, opened 2014-01-10
DS script: Rebuild permissions


aaronschachter:
Upon doing a fresh build --install, it prompts me to rebuild permissions.  I believe this because of enabling the view_unpublished module.

We can update the install script to execute `drush php-eval 'node_access_rebuild();'` as a last step -- as per https://drupal.org/comment/2978658#comment-2978658

@desmondmorris - thoughts?

desmondmorris:
@aaronschachter this seems okay to me.

PR #366, opened 2014-01-10
DS CLI clean up

desmondmorris:
This merge does a little housekeeping on the ds cli script.  A bit of block level commenting and abstraction of command functions.

This merge introduces the stub for the deploy task as well as address the node permissions rebuild task from #362


aaronschachter:
Superb!  (as per the mortal kombat voice)

desmondmorris:
@aaronschachter  added the node permission addition as a bonus

Issue #247, opened 2013-12-10
Campaign node: Textareas as Markdown


aaronschachter:
- Configure textareas to use Markdown
- Add dosomething_settings as dependency in dosomething.info file

aaronschachter:
Can we close this now that we are using field collections with markdown per #389 ?

angaither:
Yes, I think we are good here.

PR #368, opened 2014-01-10
Enable text filter on all text area campaign fields.

angaither:
Refs #247


aaronschachter:
These changes look fine, but I don't think this actually fixes #247 until #369 is resolved.  Can we change the PR description and keep #247 open until all fields where we need to use Markdown are using it.

Issue #351, opened 2014-01-09
Fact Devel: Notice messages when creating Facts


deezone:
Notice: Undefined index: fact in dosomething_fact_devel_create_facts() (line 80 of /vagrant/modules/dosomething/dosomething_fact/dosomething_fact_devel/dosomething_fact_devel.module).
Warning: array_keys() expects parameter 1 to be array, null given in dosomething_fact_devel_create_facts() (line 80 of /vagrant/modules/dosomething/dosomething_fact/dosomething_fact_devel/dosomething_fact_devel.module).
Notice: Undefined index: fact in dosomething_fact_devel_create_facts() (line 81 of /vagrant/modules/dosomething/dosomething_fact/dosomething_fact_devel/dosomething_fact_devel.module).

aaronschachter:
@DeeZone you need to use "Fixes #[ISSUE NUMBER]"  in order to automatically close issues.  Using 
Fixes Issue #[ISSUE NUMBER]" will not auto close becuase of the word "Issue" in between Fixes and the #

PR #371, opened 2014-01-12
Issue #351 - Fact devel PHP notice messages

deezone:
See issue #351

Fact devel PHP notice messages when selecting to delete existing devel fact entries when none exist.


aaronschachter:
:+1: Nice catch

PR #377, opened 2014-01-13
Double Field module patch

aaronschachter:
Applies d.o. patch for better devel_generate when creating campaigns with double_field.


angaither:
:+1:

Issue #349, opened 2014-01-09
Pantheon deployment from ds CLI


desmondmorris:
Add a "deploy" task to the ds cli bash script.  This command should run the build command and push the output up to the Pantheon instance.

In the future we should be able to deploy to feature branches on Pantheon as well.   Initially, this command should just deploy to the master branch on Pantheon, which represents the "development" environment.

Please update the deployment instructions in the WIKI.

Related to #277

PR #386, opened 2014-01-14
Initial pass at pantheon deploy script

### LINKED ISSUE #151: Method to swap out github username in make files
Currently, the only way you can test the make files is by committing them into the repo, since the build-dosomething.make is hardcoded to use the DoSomething repo: projects[dosomething][download][url] = "git@github.com:DoSomething/dosomething.git"

Maybe this is issue is just a nice to have, and we change the file to our forks locally to test ... and then just make sure to not commit that change to the make file back into the repo.

If this is the case, update documentation to do so: https://github.com/DoSomething/dosomething/wiki/Contributing

desmondmorris:
This PR introduces the first pass at deploying directly to Pantheon.  It requires that the Pantheon site be switched to SFTP mode and you must be a user on the Pantheon site.  You will also need to use the terminus `drush pauth` method to authenticate to Pantheon before using.

Usage:
`ds deploy`

**Some of this is subject to change and will be documented properly once a little more mature**

Other things in this merge:
- A bit more tidying of the DS script and adds the drops pantheon core as the drupal core used n the make file.
- Switches the profile download method to "file" instead of pulling from GIT.  This is will allow for easier local make file development.  Fixes #151 
#349

### DIFF
--- a/build-dosomething.make
+++ b/build-dosomething.make
@@ -6,5 +6,5 @@ includes[] = drupal-org-core.make
 
 ; Dosomething Profile
 projects[dosomething][type] = profile
-projects[dosomething][download][type] = git
-projects[dosomething][download][url] = "git@github.com:DoSomething/dosomething.git"
+projects[dosomething][download][type] = file
+projects[dosomething][download][url] = "drupal-org.make"
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .vagrant
 html/
+build/
 config/deploy_vars.yml
 docs
 doxy.sh
--- a/bin/ds
+++ b/bin/ds
@@ -1,6 +1,25 @@
 #!/bin/sh
 
 # ==============================================================================
+# GLOBALS
+# ==============================================================================
+
+# The base project directory
+BASE_PATH=`pwd -P`
+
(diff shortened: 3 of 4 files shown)


build-dosomething.make
```diff
@@ -6,5 +6,5 @@ includes[] = drupal-org-core.make
 
 ; Dosomething Profile
 projects[dosomething][type] = profile
-projects[dosomething][download][type] = git
-projects[dosomething][download][url] = "git@github.com:DoSomething/dosomething.git"
\ No newline at end of file
+projects[dosomething][download][type] = file
+projects[dosomething][download][url] = "drupal-org.make"
```
aaronschachter, line 10: Nice, does this mean we'll be able to test our drupal-org.make files locally?

desmondmorris, line 10: Yup.  It came to me in a dream.

I had a dreamt

aaronschachter, line 10: Sweet.  updating description, this will also resolve #151

aaronschachter:
ship it

build-dosomething.make
desmondmorris, line 10: Now we're cooking with heat

PR #387, opened 2014-01-14
Reverts back to git download type for profile

desmondmorris:
Changing this to local file, borks the install process.  In order to achieve this properly, i think we would need to stash the profile in its own directory


aaronschachter:
ok, bork away

PR #389, opened 2014-01-14
Field collection FTW

### LINKED ISSUE #369: Campaign: Double field -- text formats
The double field module we are using to provide campaign steps & FAQs doesn't support text formats. 
We either need to write a patch (and update DB schema) or switch to field collection module.

https://drupal.org/node/1891664

aaronschachter:
Fixes #369 by completely scrapping double_field and using field collection instead.


modules/dosomething/dosomething_campaign/dosomething_campaign.info
```diff
@@ -56,6 +59,12 @@ features[field_group][] = group_report_back|node|campaign|form
 features[field_group][] = group_sponsors|node|campaign|form
 features[field_group][] = group_taxonomy|node|campaign|form
 features[field_group][] = group_timing|node|campaign|form
+features[field_instance][] = field_collection_item-field_faq-field_compound_text_copy
```
angaither, line 62: looks like the old double fields are still referenced in this file, probably need to manually remove those

aaronschachter, line 62: I named all the fields the same names.  field_faq, field_step_pre, field_step_post

modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -78,6 +78,16 @@ function dosomething_campaign_preprocess_node(&$vars) {
     // Gather vars for pitch page.
     dosomething_campaign_preprocess_pitch_page($vars);
   }
+  elseif ($vars['view_mode'] == 'full'){
+    // Output entity_metadata_wrapper for easier theming..
+    $campaign = entity_metadata_wrapper('node', menu_get_object());
```
angaither, line 83: we can probably define this above the else and send it into the pitch page preprocess as well

themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -1,93 +1,21 @@
 <div class="content clearfix <?php print $classes; ?>">
+  <h3><?php print $campaign->field_call_to_action->value(); ?></h3>
 
-	<?php if (!empty($cover_image)): ?>
```
angaither, line 3: :bomb:

modules/dosomething/dosomething_campaign/dosomething_campaign.field_group.inc
```diff
@@ -24,8 +24,8 @@ function dosomething_campaign_field_group_info() {
     'weight' => '6',
     'children' => array(
       0 => 'field_step_photo',
-      1 => 'field_step_post',
-      2 => 'field_step_pre',
+      1 => 'field_step_pre',
+      2 => 'field_step_post',
```
angaither, line 28: these should be removed as well

modules/dosomething/dosomething_campaign/dosomething_campaign.module
aaronschachter, line 83: agree, going to leave here for now, and we can move accordingly.

modules/dosomething/dosomething_campaign/dosomething_campaign.field_group.inc
aaronschachter, line 28: same field names - these are ok

angaither:
:+1:

PR #396, opened 2014-01-15
Only display 'pitch' link on campaign nodes.

### LINKED ISSUE #388: Campaign: Pitch tab
- node/*/pitch access callback: new function -- should only appear for Campaign nodes and if you are an Editor or Admin
- node/*/pitch title =  "Pitch", remove the title callback function

angaither:
Remove title from pitch link.
Fixes #388


aaronschachter:
This works for changing the tab title, but for the visibility we'll need to revisit the access callback for node/*/pitch.  Ideally the check for node type should be done in a new function, like `dosomething_campaign_pitch_access`, so that anon or auth users can't hit the /pitch by just going to the URL.

aaronschachter:
:+1: :+1: :+1:

PR #401, opened 2014-01-16
Campaign update all the things.

angaither:
- match fields to pantheon form
- update group orders
- add helper text


aaronschachter:
Let's walk through this PR tomorrow.  Some thoughts:
- A field name I'd suggest changing is `confirmation_sentence`.  At first I figured this would be the msg appears when you sign up, but instead its reportback.  Possibly change that to something like `reportback_confirm_msg` so we could also have a `signup_confirm_msg` if we ever need one.
- Also unsure of what the baby_steps fields are for.

aaronschachter:
:+1:

PR #402, opened 2014-01-16
Issue #315 - Campaign: View campaigns_by_fact

### LINKED ISSUE #315: Campaign: View campaigns_by_fact
Build a view `campaigns_by_fact` which uses a Fact field as a contextual argument.  This view should have three different Block displays with the various fields as contextual arguments:
- Problem Fact
- Solution Fact
- More Facts

The view output should be a table, and display the Campaign Title (with a link to the node).

The view should be defined in the `dosomething_campaign` module, as its a view of campaign nodes.

To generate facts, use the dosomething_fact_devel module, under "Development > Generate Facts".  Next, generate campaigns using devel generate's "Generate Content" admin page.

deezone:
Fixes #315

View "campaigns_by_fact" has three blocks that when passed a fact ID as a contextual argument will display campaign titles with related link. The three block list based on campaigns fact assignments by:
- Problem Fact
- Solution Fact
- More Facts

The view is added to the dosomething_campaign module as a feature.


aaronschachter:
This looks good, go for it. I've updated the PR description so that will automatically close #315.  The text as it was written, "See issue #315", would not automatically close the issue.

deezone:
Replacing this request with #412 to avoid merge conflict.

PR #416, opened 2014-01-21
Signup views

aaronschachter:
- Exposes dosomething_signup table to views 
- Creates `user_signups` view of signups filtered by User as contextual argument 
- Places creates a block of User Signups on the user profile.


angaither:
:+1:

PR #421, opened 2014-01-21
Issue: #316 - Adds views to list related campaigns for fact

### LINKED ISSUE #316: Fact: Display campaigns on fact view page
Depends on #315 

Once campaigns_by_view is built, add the views for the different campaign fields into the `buildContent()` output in the `FactEntityController` class, to provide campaigns team ability to view what campaigns use a given fact.  
- Be sure to wrap this code within a `if (module_exists('dosomething_campaign'))` conditional, as dosomething_campaign is not a dependency for dosomething_facts.

deezone:
Fixes #316

Adds views_embed_view('campaigns_by_fact'... calls to Fact display to list related campaigns by:
- Problem Fact
- Solution Fact
- Other Fact

assignment.

Depends on campaigns_by_fact view, see Issue #315


aaronschachter:
@DeeZone Did you test to make sure the blocks appear on the fact/[fid] pages?  If so, good to go.

deezone:
Tested  fact/[fid], it's ugly but it works.

PR #423, opened 2014-01-21
Fixes php non-object notice.

### LINKED ISSUE #422: User: Is Staff PHP notice
View a campaign and:

Error message

Notice: Trying to get property of non-object in dosomething_user_is_staff() (line 50 of /vagrant/modules/dosomething/dosomething_user/dosomething_user.module).

Need to check if ($user == NULL) instead of checking for $user->uid

angaither:
Fixes #422


aaronschachter:
:+1:

Issue #354, opened 2014-01-09
Campaign: Template Preprocess text vars


angaither:
update module and template files to return the correct formatted data

PR #426, opened 2014-01-21
Preprocess vars for the pitch page.

angaither:
Refs #354


modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -80,18 +80,23 @@ function dosomething_campaign_entity_info_alter(&$entity_info) {
  */
 function dosomething_campaign_preprocess_node(&$vars) {
   if ($vars['type'] != 'campaign') { return; }
+
+  // Output entity_metadata_wrapper for easier theming.
+  $node = menu_get_object();
```
aaronschachter: Calling `menu_get_object()` here won't always work based on the view mode.  We could be viewing teasers of nodes or search results.

aaronschachter:
:+1:

PR #429, opened 2014-01-21
Added GLOBAL lookup for dosomething_fact_access() if not set

### LINKED ISSUE #427: Fact: Test user access to Facts
Based on settings in dosomething_fact_permission(), dosomething_fact_access() should control access to Fact pages.

deezone:
Fixes #427

Adds check for global $user value in dosomething_fact_access() if not passed to function.

### DIFF
--- a/modules/dosomething/dosomething_fact/dosomething_fact.module
+++ b/modules/dosomething/dosomething_fact/dosomething_fact.module
@@ -117,6 +117,9 @@ function dosomething_fact_menu() {
  *
  */
 function dosomething_fact_access($op, $account = NULL) {
+  if(!isset($account)) {
+    $account = $GLOBALS['user'];
+  }
   if ($op == 'update') {
     $perm = 'edit fact';
   }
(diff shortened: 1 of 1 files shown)


modules/dosomething/dosomething_fact/dosomething_fact.module
```diff
@@ -117,6 +117,9 @@ function dosomething_fact_menu() {
  *
  */
 function dosomething_fact_access($op, $account = NULL) {
+  if(!isset($account)) {
```
aaronschachter: spacing!

modules/dosomething/dosomething_fact/dosomething_fact.module
```diff
@@ -117,6 +117,9 @@ function dosomething_fact_menu() {
...
+    $account = $GLOBALS['user'];
```
aaronschachter: Is there any advantage to using this instead of `global $user;` ?

deezone: Nope, just trying to use $account as that's the parameter name.

deezone:
For what it's worth: http://drupal.stackexchange.com/questions/23000/what-is-the-best-way-to-get-current-user-object

Looks like is a long standing question: global $user; vs $GLOBALS['user'];

aaronschachter:
Let's keep our code consistent.  I'm using `global $user` all over the dosomething_signup module -- let's stick with always declaring it that way.  so maybe just

```
global $user;
$account = $user;
```

aaronschachter:
:+1:

PR #438, opened 2014-01-22
Automatically generate content on site install.

angaither:


aaronschachter:
Looks good.  Pls update wiki with documentation then ship it

PR #449, opened 2014-01-23
Campaign tpl - check if field collections vars are set

aaronschachter:
- Alters `dosomething_campaign_get_field_collection_values` to return NULL if empty values
- Adds `isset` check for field collection vars inside action page template


angaither:
:+1:

PR #455, opened 2014-01-24
Reportback: Insert / update functions 

aaronschachter:
Split out inserts and updates in `dosomething_reportback_form_submit` into individual functions, to make for better unit testing, and allowing other apps (like mobile) to call these functions as well.
- Passes through uid as a value instead of auto-inserting global $uid.


angaither:
:+1:

PR #464, opened 2014-01-24
DS Build tweaks

aaronschachter:
- Reverts all features to avoid funky permissions mystery upon fresh build
- Adds sponsors and tags terms into --dev function
- Creates more facts in --dev


angaither:
:+1:

PR #473, opened 2014-01-24
Sponsor images

### LINKED ISSUE #467: Campaign: Sponsors Image in tpl
preprocess & send image to tpl

angaither:
Fixes #467

### DIFF
--- a/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -86,10 +86,9 @@ function dosomething_campaign_preprocess_node(&$vars) {
 
     $node = menu_get_object();
     $wrapper = entity_metadata_wrapper('node', $node);
-
+print "<pre>"; print_r($vars['field_sponsor']); print "</pre>";
     // Vars needed for both the pitch & action page.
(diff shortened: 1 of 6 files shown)


modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -86,10 +86,13 @@ function dosomething_campaign_preprocess_node(&$vars) {
 
     $node = menu_get_object();
     $wrapper = entity_metadata_wrapper('node', $node);
-
     // Vars needed for both the pitch & action page.
     $vars['cta'] = $wrapper->field_call_to_action->value();
-
+    if (isset($vars['field_sponsor'][0])) {
```
aaronschachter: It's possible we may have multiple sponsors on a campaign.  Could leave code as is for now and create a separate issue, or could handle via this PR .

Also, it might be nice to use the $wrapper for easier readability.  It's up to you, but it would keep it consistent with a lot of the other variable declarations.  Not tested but e.g.

```
  foreach ($wrapper->field_sponsor->getIterator() as $delta => $term_wrapper) {
    // $term_wrapper may now be accessed as a taxonomy term wrapper.
    $label = $term_wrapper->label->value();
  }
```

aaronschachter:
:+1: Sweet

PR #474, opened 2014-01-24
Delete Reportback Form

aaronschachter:
Creates menu callback reportback/[id]/delete and checks for relevant "delete" access.

Upon deletion, redirects user back to the campaign that the deleted reportback was for.


angaither:
:+1:

PR #481, opened 2014-01-25
More reportback stuff

### LINKED ISSUE #469: Reportback: Add additional properties
new properties:
- Comments, long text.  caption: "Why is this campaign important to you?"
- num_participants, integer.
- updated, timestamp.

Also:.
-  Add into the reportback_form / form_submit functions.  
- Autoupdate the updated property via ReportbackEntity class save()
- reportback_form: On update, display the updated date as read only.

### LINKED ISSUE #477: Campaign: Rename reportback fields
Rename `field_report_back` fields as `field_reportback` instead, to keep the naming convention same sitewide.

### LINKED ISSUE #479: Reportback: Security
- [x] Sanitize quantity and why_participated fields
- [x] Prevent manipulation of rbid in browser to save to a different reportback

Having hidden form fields on the reportback form for rbid and nid means a web dev user could open up firebug, change the rbid or nid values, and write to the wrong records.

aaronschachter:
## Note - re-build ur local after this PR merges

Easiest to just do a new `ds --build` upon merging this PR to avoid database errors.  Otherwise if you don't want to do that you need to

```
drush -y fr dosomething_campaign
drush -y dis dosomething_reportback
drush -y dis dosomething_reportback
drush -y en dosomething_reportback
```

Fixes #469 by adding 3 new properties

Closes #477 by standarizing "reportback" naming convention and also fixes #479


aaronschachter:
Closing this because we're going to handle #479 via a md5 field instead of checking extracting rbid's and nid's from the current path.

PR #483, opened 2014-01-27
Reportback new properties

### LINKED ISSUE #469: Reportback: Add additional properties

aaronschachter:
Fixes #469 by adding some new reportback properties into the table and form, and sneaks in #477 too.


aaronschachter:
Easiest to do a fresh `ds build --install --dev` in order to apply these changes.  Otherwise run these drush commands to apply changes:

```
drush -y fr dosomething_campaign
drush -y dis dosomething_reportback
drush -y dis dosomething_reportback
drush -y en dosomething_reportback
```

modules/dosomething/dosomething_reportback/dosomething_reportback.install
```diff
@@ -36,12 +36,30 @@ function dosomething_reportback_schema() {
...
       'quantity' => array(
         'description' => 'The quantity of reportback_nouns reportback_verbed.',
         'type' => 'int',
         'not null' => TRUE,
         'default' => 0,
       ),
+      'why_participated' => array(
+        'description' => 'Why the user participated.',
+        'type' => 'varchar',
+        'length' => '255',
```
jonuy, line 54: Is the user answer to the why question getting dropped in here? Do you think 255 might be too small?

aaronschachter, line 54: Yes, this stores "why is this campaign important to you?".  I debated between doing the varchar or a long text, but went this route because we could always make it bigger.. and for now we'd get slightly better performance with the small field.  I'm happy to change it to a larger text field though - do you think thats a safer bet?

jonuy, line 54: Not sure. Sounds life you've thought about it though. I think it's fine for now, but it might be worthwhile talking to data about it and figuring out if there's a rough median length to the answers for that question.

jonuy:
A moment of silence for the underscores lost...

I think it looks fine.

modules/dosomething/dosomething_reportback/dosomething_reportback.install
aaronschachter, line 54: Agree.  Gonna keep it in for now, @mikefantini says he'll get back to me with a definitive answer.

Issue #417, opened 2014-01-21
Campaign: Anon user display


aaronschachter:
If anonymous user, pitch page should be returned where signup button should calls login form via modal.

Depends on neue to work again, and for #532 to actually signup user upon submit.

aaronschachter:
Assigning to myself to work on login / register form_alters to handle signups upon login.

PR #486, opened 2014-01-28
Display the pitch page to all users if not signed up for the campaign

### LINKED ISSUE #418: Campaign: Auth user display
`dosomething_campaign_preprocess_node` needs to check if `dosomething_signup_exists($nid)`.
- If TRUE: output the action page.
- if FALSE: output the pitch page, with a signup form.

angaither:
Fixes #418
Refs #417

### DIFF
--- a/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -80,7 +80,7 @@ function dosomething_campaign_entity_info_alter(&$entity_info) {
  */
 function dosomething_campaign_entity_view_mode_alter(&$view_mode, $context) {
   // Is this a campaign node?
-  if ($context['entity_type'] == 'node' && $context['entity']->type == 'campaign') {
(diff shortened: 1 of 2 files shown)


modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -74,6 +74,22 @@ function dosomething_campaign_entity_info_alter(&$entity_info) {
     'custom settings' => TRUE,
   );
 }
+/*
+ * Implements hook_entity_view_mode_alter().
+ * Used to change the view_mode of the campaign nodes.
+ */
+function dosomething_campaign_entity_view_mode_alter(&$view_mode, $context) {
+  // Is this a campaign node?
+  if ($context['entity_type'] == 'node' && $context['entity']->type == 'campaign') {
```
aaronschachter: I think we may need to also check to see if view_mode == full, to make sure the user is viewing an actual node page.  One possible view mode we could possibly use in the future "search result" -- which we would not want to return the pitch for.

angaither: fixed

aaronschachter:
:+1:

PR #488, opened 2014-01-29
Reportback Form fix

aaronschachter:
Gets rid of PHP notice by populating $entity->num_participants with NULL if form is in create mode.


angaither:
:+1:

Issue #444, opened 2014-01-22
Pattern Library integration


DFurnes:
We need a method of including the pattern library (our base styles and re-usable interface patterns, used across products and as a deliverable for partners), and also building product-specific styles in here where they belong.

Currently learning towards using Bower, as this lets us version-lock what we're pulling in rather than always pulling the latest off master; and fits into a natural front-end build process for development.

DFurnes:
:dog:

PR #491, opened 2014-01-29
Frontenders

DFurnes:
:construction: 

Things are still quite messy, but this gets local SCSS compilation working (and some very basic local JS compilation). Built files are currently gitignored – they will be compiled using the `ds build` task in the future so that we don't have to be storing compiled source in the repo and dealing with merge conflicts from that.


DFurnes:
This fixes #444. :bird:

aaronschachter:
@desmondmorris -- any thoughts?  this seems ok to me but i haven't been involved in the initial architecture hooking into neue.  (e.g. removing the download in the make file)

DFurnes:
@aaronschachter Re: removing the download in the make file, we're including Neue using Bower rather than Drupal's Library feature now. This lets us version our dependencies on the front-end the same way you would in a Ruby `Gemfile` or Node `package.json`.

Talked the plan for this over with Desmond earlier this week, so I'm going to merge this in so we can keep building off this. I'll also give a little overview to the rest of the team when we do our retrospective this Friday.

aaronschachter:
ok, rock.

desmondmorris:
@DFurnes this seems okay.  Safe to merge.

Some things that need to happen:
- NPM, bower and grunt need to be installed in the Vagrant instance
- The DS build step needs to be updated to run bower install 
- We may want a convenience DS cli command to update bower as well.  not necessary, but would be good to have.

I till take a look this evening https://github.com/DoSomething/dosomething/issues/498

PR #499, opened 2014-01-29
Cache themed image tags.

angaither:


aaronschachter:
This looks good. For future PR's please open an issue up first as a place for dev's to discuss / fight it out / say nothing at all, and associate the PR to it.

PR #503, opened 2014-01-30
Adds bower install command to ds cli tool.  

### LINKED ISSUE #504: Update ds binary with asset helpers
`ds` CLI should have access to Bower, NPM and Grunt so that everyone's development environment is able to pull the latest changes upon building or re-building.

desmondmorris:
```
# This will install bower components in paraneue subtheme
ds bower

OR

ds bower --install
```

```
# This will update bower components in paraneue subtheme
ds bower --update
```

@DFurnes @mmwtsn - After doing this, it may make better sense to include the bower goodness in the paraneue  base theme.  The subtheme will inherit from there.  Flags/thoughts/issues with that idea?

Fixes #504

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -58,8 +58,11 @@ ds build [--install]
   --install
     Runs "drush site-install" to actually install the Drupal instance.
 
-ds bower [--update]
+ds bower [--install|--update]
   Installs bower components in the paraneue subtheme
+
+  --install (default)
+  Runs bower install command.  This is the default behavior
 
   --update
   Runs bower update command instead of install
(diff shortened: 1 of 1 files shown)


mmwtsn:
:flags: Not having a separate `--install` flag breaks with convention set by `build` elsewhere in the CLI. Should install either need to take a flag or at least have that option?

Let's revisit the base theme discussion in person. Dave and I were talking through it yesterday evening and there are a few concerns we have with that approach.

desmondmorris:
Not sure if it necessarily breaks the convention as the default behavior for bower is install and the default behavior for build is build.  Neither expect/need an explicit flag.  But I do see where it may be confusing for folks.  There is no reason not to add it.

mmwtsn:
Good point. Either way, it doesn't hurt to have it in there. :thumbsup:

PR #521, opened 2014-01-30
Reportback Activity log

aaronschachter:
Need to either rebuild or uninstall / reinstall dosomething_reportback module if testing any reportback stuff.

This PR creates a dosomething_reportback_log table, to track all reportback insert, update, and delete transactions.  This has been requested by data team since we're no longer allowing users to submit multiple reportbacks, but instead giving them the ability to update one reportback per campaign they've signed up for.

### DIFF
--- a/modules/dosomething/dosomething_reportback/dosomething_reportback.install
+++ b/modules/dosomething/dosomething_reportback/dosomething_reportback.install
@@ -87,6 +87,12 @@ function dosomething_reportback_schema() {
         'not null' => TRUE,
         'default' => 0,
       ),
+      'op' => array(
+        'description' => 'Operation performed on the reportback.',
+        'type' => 'varchar',
+        'length' => '255',
+        'not null' => FALSE,
+      ),
       'timestamp' => array(
         'description' => 'The Unix timestamp of the update.',
         'type' => 'int',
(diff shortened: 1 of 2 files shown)


desmondmorris:
Only concern is maintaining this.  This will grow quite large after a while.  Cant we get the result from a single entity query?

aaronschachter:
Hey @desmondmorris  -- not sure if you're aware but one major change we're making to reportbacks is that there is only one reportback per user per campaign.  If you want to reportback again on a campaign you've reported back for, you're presented with the update form for your reportback instead of a new create reportback form.

Data team still wants the ability to see reportback activity over time though, instead of just that one last updated record, hence all the log activity.

Ideally, we'd stash all the reportback transaction activity on the User document in mongo instead of in this table to avoid it being huge.  When we get there, we could move this table to the User document.

Not sure if that clarifies, but if it doesn't change things: how would a single entity query work?  I think the only way it would work is if each update was a revision to the entity, which it's not set up to do right now. I suppose we could look into using that method instead of a log, it would make sense because it's built in,...  however using the separate log table and functions lets us easily change the `dosomething_reportback_log_insert` to write records to mongo later.

desmondmorris:
Ahh gotcha.  This makes sense

PR #527, opened 2014-01-30
Dosomething_image module updates.

### LINKED ISSUE #523: Image: create "image_get_image_url" function
The `dosomething_image` module is only able to return a themed image. In certain cases we'll need to return only the image path (e.g. setting the hero image on the Pitch and Action pages as inline `background-image` styles).

angaither:
- broke out image path logic
- added function to return themed image urls
- added hero image urls to campaign template

Fixes #523

### DIFF
--- a/modules/dosomething/dosomething_image/dosomething_image.module
+++ b/modules/dosomething/dosomething_image/dosomething_image.module
@@ -7,7 +7,7 @@
 include_once 'dosomething_image.features.inc';
 
 /**
- * Returns an image src for any image field
+ * Returns a themed image for any image field
  * @param int $nid
  *  The image node id.
  * @param string $ratio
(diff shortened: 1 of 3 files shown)


modules/dosomething/dosomething_image/dosomething_image.module
```diff
@@ -16,17 +16,38 @@ include_once 'dosomething_image.features.inc';
...
+ * @param string $style
+ *  The image style preset/size to return the image.
+ *
+ * @return string
+ *  An image url.
+ */
+function dosomething_image_get_image_url($nid, $ratio) {
+   $node = node_load($nid);
     switch ($ratio) {
       case 'vert':
```
aaronschachter, line 52: Can we rename these parmas as "portrait" and "landscape" to keep consistent site wide?

angaither, line 52: only if we can actually call it p, l, and s
landscape takes way too long to type (I HAVE TENDENTIOUS MUST BE CAUTIOUS)

aaronschachter:
Looks ok except for global find/replace comment.

PR #529, opened 2014-01-31
Browserify

### LINKED ISSUE #462: DS Client-Side: App-level SCSS and JS
Setting up a local SCSS and JS build system for app-specific styles and client-side behavior. This lets us keep Neue and Paraneue as abstract deliverables for affiliates without mixing a ton of our own app-level stuff in there.

We'll be using Bower to pull in DS Neue assets. A helper SCSS file will import mixins and variables for local SCSS development. This will be where we work on any app-specific stuff (rather than general re-usable patterns) which compiles into `app.css`. The `neue.css` and local `app.css` will be concatenated using the Drupal asset pipeline.

Scripts for app-specific features (such as pitch page logic, location finder, etc.) will live in the `dosomething` repo. We'll use Browserify to manage dependencies and compile code, and then concatenate with Neue assets using the Drupal asset pipeline.

DFurnes:
Fixes #462 after dosomething/paraneue#17 gets merged in.

Current order of things in the front-end world:
- Neue has CSS and JS for basic layout. Exports a SCSS helper file with mixins and variables for app dev.
- Paraneue pulls in Neue's CSS and JS, and jQuery. It themes Drupal base features and nothing more. Paraneue does **not** include its own app-level CSS or JS.
- Paraneue_DoSomething has app-level SCSS and JS (i.e. styles and scripts for pitch/action page) and a build system to keep that up to date.

Awkward bits:
- There are two separate Bower installations to install dependencies to Paraneue and Paraneue_DoSomething. Since jQuery and Neue are pulled in during Paraneue's build process, changing dependencies on Paraneue_Dosomething won't affect anything.
- We can't build JS with dependencies in Paraneue, since the Browserify build process in Paraneue_DoSomething won't know that the dependencies were already included in Paraneue, and so if they're referenced in both places then it will be included in both files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,4 @@ doxy.sh
 Doxyfile
 node_modules
 themes/dosomething/paraneue_dosomething/dist/
-themes/dosomething/paraneue_dosomething/vendor/
+themes/dosomething/paraneue_dosomething/bower_components/
--- a/themes/README.md
+++ b/themes/README.md
@@ -1,3 +1,9 @@
 DoSomething specific custom themes live here.
 
 We need this directory in order to make symlinking during development easier.
+
+## Theming Stack Overview
+
+ - `Neue` has CSS and JS for basic layout. Exports a SCSS helper file with mixins and variables for app dev. It is imported into `Paraneue` and `Paraneue_DoSomething` using a `bower.json` file in each respective directory.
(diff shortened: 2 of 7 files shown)


aaronschachter:
@DFurnes This PR contains some great info that would be worth adding as a README.md file in dosomething_paraneue (and possibly create a README.md for paraneue which highlights your 2nd bullet point)

DFurnes:
Good call.

Issue #513, opened 2014-01-30
Add drush make_local plugin


desmondmorris:
The drush [make_local](https://drupal.org/project/make_local) plugin allow us to grab projects from local paths.  This will be used specifically for pulling in the install profile and will make @aaronschachter very happy.

Related to #387

desmondmorris:
Testing this locally now

PR #536, opened 2014-01-31
Installs make_local drush plugin to werkcer env

desmondmorris:
Wercker env is breaking because the make_local plugin introduced in #513 was not ported over (my bad).


aaronschachter:
:+1:

PR #543, opened 2014-01-31
Pitch page for anon user

aaronschachter:
- Displays pitch page for anonymous user, and renders signup_button as a link instead of the signup form.
- Passes new param $label into dosomething_signup_form to declare the submit label value once within `dosomething_campaign_preprocess_pitch_page`.

### DIFF
--- a/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -82,12 +82,17 @@ function dosomething_campaign_entity_info_alter(&$entity_info) {
 function dosomething_campaign_entity_view_mode_alter(&$view_mode, $context) {
   // Is this a campaign node?
   if ($context['entity_type'] == 'node' && $context['entity']->type == 'campaign' && $view_mode == 'full') {
+    // If anonymous user:
+    if (!user_is_logged_in()) {
+      // Display pitch view mode.
+      $view_mode = 'pitch';
+      return;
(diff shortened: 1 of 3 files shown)


modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -82,12 +82,17 @@ function dosomething_campaign_entity_info_alter(&$entity_info) {
...
+      return;
+    }
     $node = $context['entity'];
-    // If the current user is not on staff, and not signed up for the campaign, show the pitch page.
-    if (!dosomething_user_is_staff()) {
-      if (!dosomething_signup_exists($node->nid)) {
-        $view_mode = 'pitch';
-      }
+    // Otherwise, if auth user not signed up for the campaign:
+    if (!dosomething_signup_exists($node->nid)) {
```
angaither: won't this force the pitch page on ds staff?

aaronschachter: Yes, if they aren't signed up.  I do think it will make it easier for internal testing to force the actual signup to see the full node.

aaronschachter:
Updated PR to keep check for `dosomething_user_is_staff` in `dosomething_campaign_entity_view_mode_alter`, so that staff will always see full action page when viewing a campaign node.  (They have access to the "Pitch" node tab to view the pitch page)

angaither:
:+1:

Issue #506, opened 2014-01-30
Video field collection


aaronschachter:
field_video as a collection with fields:
- field_video_provider
- field_video_id

aaronschachter:
This was not closed by https://github.com/DoSomething/dosomething/commit/c0e7eaf06dddaf5516c5e25939bf37a1a94f763b

aaronschachter:
This issue was closed by https://github.com/DoSomething/dosomething/commit/c5fae33e3a4b377bcf02a8057d85c10a097a8523

PR #550, opened 2014-02-03
Static content video

### LINKED ISSUE #528: Static Content: Add intro video field and corresponding logic

blisteringherb:
Adding intro video field to static content and supporting theme function to ds_helper module.  

Fixes #528


modules/dosomething/dosomething_static_content/dosomething_static_content.info
```diff
@@ -3,6 +3,8 @@ description = Content type and theme function definitions for the Static Content
 core = 7.x
 package = Features
 dependencies[] = ctools
+dependencies[] = dosomething_campaign
+dependencies[] = double_field
```
aaronschachter: I thought in code review, we had decided to use a Field Collection to handle the video composite fields? cc: // @angaither ? If we go double_field we need to add back into the make file

angaither: field collection, FTW!

modules/dosomething/dosomething_static_content/dosomething_static_content.features.field_base.inc
```diff
@@ -256,6 +257,37 @@ function dosomething_static_content_field_default_field_bases() {
     'type' => 'text',
   );
 
+  // Exported field_base: 'field_intro_video'
+  $field_bases['field_intro_video'] = array(
```
aaronschachter: My idea here was sharing a video field between the Static Content and Campaign content types... would it make sense to just name it field_video and share between both?

modules/dosomething/dosomething_static_content/dosomething_static_content.info
aaronschachter: Already an issue open for this #506

modules/dosomething/dosomething_static_content/dosomething_static_content.features.field_base.inc
blisteringherb: We can definitely have one shared field that we use most of the time, but are there any cases where we might have multiple video fields on a content type?

aaronschachter: Potentially it could happen, but currently they are not any cases for this.

aaronschachter:
This looks good, but two things:

1 - do you need to also export the field collection fields into the feature?  The only added field is `field_video`.  If you do a `drush fc --not-exported` and select field_base or field_instance, I'd imagine the fields that belong to the field_video collection would show up?

2 - minor, but are you using git rebase to merge dev?  this would be preferred to avoid the "Merge branch dev" commits.  See: https://github.com/DoSomething/dosomething/wiki/Contributing#wiki-using-git-rebase-to-keep-up-to-date

aaronschachter:
:+1:

PR #559, opened 2014-02-03
Reportback: No more num_participants

aaronschachter:
Remove num_participants field as per convo with Fantini.

Also provides documentation around log table and Mongo wish list.


aaronschachter:
Note: If testing reportbacks, uninstall and re-install this module

angaither:
:+1:

PR #560, opened 2014-02-03
Bower stuff

DFurnes:
Fixes missing Grunt plugin dependency. Also we're no longer pulling Neue into both Paraneue and Paraneue_DoSomething so that the code isn't duplicated (and therefore potentially having the version numbers differ). Fixes dosomething/paraneue#19.

:chicken:


mmwtsn:
That file path. Cool.

PR #567, opened 2014-02-03
Fact Search View tweaks

aaronschachter:
- Checks for "view fact" permission instead of role
- Allows sortable by Date Created, Date Updated


angaither:
:+1:

PR #572, opened 2014-02-03
Reportback Schema Test

aaronschachter:
Updates test with correct properties ("created" instead of "timestamp") and tests for dosomething_reportback_log table / fields.


angaither:
:+1:

PR #574, opened 2014-02-03
Local Drupal settings FTW

mshmsh5000:
This ships with a template for local settings: `default.settings.local.php`.

On the initial build, this will get copied to the non-versioned `settings.local.php`.  `bin/ds` will make sure this is symlinked to `sites/default/settings.local.php`. The main `settings.php` will find and include this.

This is now where you can store all instance-specific settings.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .vagrant
 html/
+settings.local.php
 build/
 config/deploy_vars.yml
 docs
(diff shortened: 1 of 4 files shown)


desmondmorris:
The code looks good.  Its borking the wercker build for a reason that is not apparent.

mshmsh5000:
I am sad bc 
screen shot 2014-02-03 at 5 54 04 pm

desmondmorris:
@mshmsh5000 sign in!

PR #585, opened 2014-02-04
Remove unused text format module/libs

### LINKED ISSUE #573: Remove CKEditor library
I assume this is not necessary, given that we're going to use Markdown across the board:

https://github.com/DoSomething/dosomething/commit/7e4b3b2b

Can we just revert? @desmondmorris since this was your commit, please comment then reassign.

### LINKED ISSUE #584: Remove WYSIWYG Module

angaither:
Fixes #584
Fixes #573

### DIFF
--- a/dosomething.info
+++ b/dosomething.info
@@ -41,7 +41,6 @@ dependencies[] = varnish
 dependencies[] = view_unpublished
 dependencies[] = views
 dependencies[] = views_ui
-dependencies[] = wysiwyg
 
 ; DOSOMETHING
 dependencies[] = dosomething_campaign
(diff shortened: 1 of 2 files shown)


aaronschachter:
Need to also remove modules from the dosomething.info file.

angaither:
@aaronschachter updated

aaronschachter:
:+1: Looks good.  How will this work with staging though?  Would we manually just disable the modules and delete the directories in html/profiles/dosomething/modules/contrib ? cc: @mshmsh5000 @desmondmorris

angaither:
drush-make would fix this, right? 
we need something a little like this when we deploy 
https://github.com/letharion/Drupal-build-scripts/blob/master/build.sh

aaronschachter:
yes, and yes

PR #590, opened 2014-02-04
Static content issets

blisteringherb:
Add additional checks in static content tpl to avoid PHP notices.

Depends on #550


aaronschachter:
:+1:

Issue #456, opened 2014-01-24
Reportback: Tests for CRUD functions


aaronschachter:
Depends on #455 

Web tests for:
- [x] dosomething_reportback_schema
- [ ] dosomething_reportback_insert_reportback
- [ ] dosomething_reportback_update_reportback
- [x] dosomething_reportback_save
- [x] dosomething_reportback_set_properties
- [x] dosomething_reportback_set_files
- [ ] dosomething_reportback_log_insert
- [ ] dosomething_reportback_entity_insert
- [ ] dosomething_reportback_entity_update
- [ ] dosomething_reportback_entity_delete
- [ ] dosomething_reportback_exists
- [ ] dosomething_reportback_access
- [ ] dosomething_reportback_form

barryclark:
@aaronschachter Still for Tues?

aaronschachter:
This would be a nice to have, but I'd imagine this is not critical for launch.  Was hoping I'd have bandwidth but not sure about that yet.

barryclark:
Cool, will leave it in 0.2.0 for now.

barryclark:
Backlogging this as it's not critical for Tuesday launch.

PR #591, opened 2014-02-04
Reportback Tests

aaronschachter:
Works towards #456.  Opening this is as a PR for now so it doesn't end up being a gigantic one to review.


desmondmorris:
This is looks okay.  Give the merge a shot.  Follow the directions below VVV

PR #593, opened 2014-02-04
Solr - Basic setup for indexing nodes

blisteringherb:
Adding search_api, search_api_solr modules and their respective settings to support indexing of nodes.


aaronschachter:
This looks ok. It's minor but I'd vote to rename the module `dosomething_search`.  It also needs to be added as a dependency in the dosomething.info

aaronschachter:
This all looks good - except missing one last step in dosomething.info -- adding the dosomething_serach module into the dependencies for DoSomething modules.  Add that then good to merge :+1:

PR #599, opened 2014-02-05
Proposed: Codebase re-org

desmondmorris:
In an effort to further cleanup our work area, I took a pass this suggested file structure.  I believe it is more intuitive and provides better separation of concerns.  The changes are as follows:
- All configuration/settings files now live in the **/config** directory.  This directory houses the settings.php file as well as the default local settings file. 
- All provision related files and directories now live in the **/provision** directory.  This is where the salty stuff is.  If we decide to move to chef or puppet, this is where they will live.
- The **/lib** directory is now home to our application libs (modules, themes, profiles).  The modules directory and themes direct have been moved here.  A new dosomething profile directory has been created and the profile related files have been moved here.  I believe this will make it much easier for people to distinguish which make related files belong to what.

Preview: https://github.com/desmondmorris/dosomething/tree/codebase-reorg

The necessary path changes have been made to the ds script, makes files and Vagrantfile.  Thoughts, concerns, opposition?

This is a breaking change that will require a rebuild of your environment.


aaronschachter:
:+1: looks rad.

desmondmorris:
Thanks @aaronschachter.  I want to make sure more folks get eyes on this too.

mshmsh5000:
I like the look of it. I assume this still creates and builds to html.

I'm thinking this morning about how we package our build for international distro (and thence to Pantheon). Let's make sure, now that you're doing this work, we have an eye on that goal too.

I imagine the workflow will look like
- Tag for release
- Run ds build
- Package and push html dir to Pantheon

What think ye?

barryclark:
+1 Really clear and simple @desmondmorris

desmondmorris:
@mshmsh5000 Yup it still works the same as before.

I think that work flow makes sense.  We should really incorporate it into our internal deployments as well.

PR #604, opened 2014-02-05
Moving ds_search module to the new lib directory

blisteringherb:
The ds_search module was left behind in the modules directory.  This moves it to the lib directory.


aaronschachter:
:+1:

PR #605, opened 2014-02-05
Removing ds_search from make file to prevent build breakage

blisteringherb:
- The ds_search module contains features that should create a
  new server, but this isn't working properly


desmondmorris:
ship it!

PR #610, opened 2014-02-06
Adding Cause taxonomy to Fact Page

blisteringherb:
Fixes #594

PR #611, opened 2014-02-06
11 Facts Permissions

### LINKED ISSUE #609: 11 Facts: Permissions
Export permissions for Fact Page content type

blisteringherb:
Adding permissions definitons for fact page

Fixes #609


aaronschachter:
Hmm, this commit has your changes from #610 in as a separate commit, which is funky.  Possibly merge #610 first?

Not a huge deal but FYI - you don't need to include the administrator role, the admin role has "bypass node access" conditions.  Would prefer to keep all of these identical, if you don't mind removing.

aaronschachter:
not a huge deal but can you explicitly add the perms for delete own / delete any content as well? see https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_image/dosomething_image.features.user_permission.inc

aaronschachter:
:+1:

Issue #606, opened 2014-02-05
Fact: Rebuild as node


aaronschachter:
Rebuild Fact module to create new content type, Fact, instead of custom entity.

This will allow use of EntityConnect for adding/editing facts "inline" from a campaign or fact_page.

Details:
- [x] Drop dosomething_fact table, remove install schema, and entity hooks / classes
- [x] Create "fact" content type with body for Fact copy, and Source multi-value longtext
- [x] Updating all Fact EntityReference fields to point to Fact nodes (not Entities)
- [x] Displaying campaigns_by_fact views on the Fact node content
- [x] Node access check to only make Facts pages visible to staff
- [x] Change Fact Search to search Fact nodes
- [x] Update Fact permissions

PR #612, opened 2014-02-06
Fact Content Type

aaronschachter:
Begins work on #606

This PR defines a Fact content type in the dosomething_fact module, and updates the existing Fact entityreference fields to use the Fact node type instead of custom entity.

I know I'm breaking the rules here a little bit, but in case you have campaigns and facts generated on your local and don't want to run a fresh --install, you'll need to truncate the facts tables.  I'll be doing this on the staging site, because we have a small enough data set that it's not worth writing proper update hooks for.

```
TRUNCATE TABLE field_data_field_fact_problem
TRUNCATE TABLE field_data_field_fact_solution
TRUNCATE TABLE field_data_field_facts
TRUNCATE TABLE field_revision_field_fact_problem
TRUNCATE TABLE field_revision_field_fact_solution
TRUNCATE TABLE field_revision_field_facts
```


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.info
```diff
@@ -5,7 +5,6 @@ package = DoSomething
 version = 7.x-1.0
 project = dosomething_campaign
 dependencies[] = date
-dependencies[] = dosomething_fact
```
blisteringherb, line 8: Shouldn't this still be a dependency since the campaign content type references facts defined in ds_fact?

aaronschachter, line 8: I removed it in attempt to prevent all the custom modules from being dependent on each other.  I believe that if you disable a content type's feature module which defines it, the content type will still exist in the database.  Would just need to wrap any functions like `dosomething_fact_get_values` within conditionals that check for `if (module_exists('dosomething_fact'))`

I don't feel super strongly about this, but I do think the less dependencies the safer. I've burned in our old world site before on tons of modules depending on a `dosomething_general` module, it accidentally being disabled, and then crashing the site :)

aaronschachter:
Also updated PR with Fact content type user permissions.

mshmsh5000:
If a module A requires module B to operate, then it's a dependency. If it can muddle on without module B, then it's not a dependency. I'm in favor of reducing explicit dependencies to what's absolutely necessary.

blisteringherb:
:+1:

aaronschachter:
Confirmed that disabling the dosomething_fact module won't kill the actual content type.

PR #620, opened 2014-02-07
Added DMOZ meta tag.

### LINKED ISSUE #618: DMOZ meta tag in header
Need to add `<meta name="robots" content="NOODP">` to head block to override incorrect Open Directory Project site information.

Relevent Trello Card: https://trello.com/c/o208GKs3/21-prevent-search-engines-from-using-dmoz-site-descriptions

DFurnes:
Closes #618. :radio:


lib/themes/dosomething/paraneue_dosomething/template.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+/**
+ * Overrides to modify Paraneue's <head> section.
```
aaronschachter: This should read as:

```
Implements hook_html_head_alter().
```

aaronschachter:
Should we move this into `dosomething_helpers.module` instead, so it's not dependent on the theme?

DFurnes:
Updated the comment.

I feel like anything on the presentation layer belongs in the theme. I suppose we can always just `grep` for it, but doesn't spreading the presentation layer out into a bunch of different places seem like a maintainability headache down the line?

aaronschachter:
I don't feel super strongly on this one, but my argument would be that this is not presentation layer, because we would want this in the <HEAD> regardless of any theme we use on the site.  Anybody else want to weigh in?

angaither:
If we want this functionality no matter the theme, then I agree with @aaronschachter it belongs in `dosomething_helpers.module`

DFurnes:
But if it's not manipulating application or database state (the Model or Controller parts of MVC), then it's got to be presentational, right? This is only concerned with how the page looks (albeit to a machine, rather than a person) and never touches any sort of "business logic".

DFurnes:
(That being said, we're arguing over a single `<meta>` tag that's probably going to be pulled out of the source in a few months anyways, so it's not a big deal. I'll defer to your best call, since I'm not as well-versed in the proper Drupal way of things as you guys are.)

DFurnes:
:tangerine:

DFurnes:
Merged manually; closing.

desmondmorris:
My two cents and sorry to chime in late on this one.  I have to agree, why would this not live in the theme?  This is specific only to head markup with no relation to the any configuration.  This would be a similar argument for keeping the meta viewport tag in the theme layer.

PR #621, opened 2014-02-07
Fact Cleanup

aaronschachter:
Removes all traces of the Fact entity.  Another PR towards #606 -- i have a checklist on that issue outlining what's been done and is left to do.

I will be manually deleting the dosomething_fact table from the staging database.  Discussed this with @mshmsh5000 that not worth writing update hooks since we're so early in adding Fact content into the site.


lib/modules/dosomething/dosomething_fact/dosomething_fact.views_default.inc
```diff
@@ -14,7 +14,7 @@ function dosomething_fact_views_default_views() {
   $view->name = 'fact_search';
```
angaither, line 14: we can probably remove this view all together & add facts to the 'search content' view

aaronschachter, line 14: I re-created it as a separate view, so you can search on the Fact itself (which is the node body) but also the Source field (which is a multi-value field).

angaither:
:+1:

PR #627, opened 2014-02-07
Static Content: Fixing PHP notices

blisteringherb:


aaronschachter:
Very nice :1 ship it value added

PR #628, opened 2014-02-07
Facts of life

aaronschachter:
Almost done fighting the good fight on #606.

This makes fact nodes only visible to staff, and removes the Body field and uses Title as Fact copy instead.  Fixes action page template.


angaither:
:+1:

PR #629, opened 2014-02-07
Action Guide content type

aaronschachter:
Defines Action Guide content type, shares fields defined by `dosomething_static_content`.


angaither:
:+1:

PR #630, opened 2014-02-07
Adds helper functions to display entityref parents

aaronschachter:
Fixes last item left in #606
- Adds helper functions which query entityreference fields to see where a given node is referenced.
- Moves `hook_node_access` check into `dosomething_user` to have one place to deny access to view content types that only staff should have view access to (facts, images).
- Displays three sections on a Fact node (which is an internal view only staff can see):
  - parent nodes which reference the Fact via field_facts (this will include campaigns and fact_page)
  - parent nodes which reference the Fact via field_fact_problem
  - parent nodes which reference the Fact via field_fact_solution
- Similarly displays parent nodes which reference a given image node on the image node's node_view (which again, only staff can see

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -23,3 +23,76 @@ function dosomething_helpers_form_alter(&$form, &$form_state, $form_id) {
     drupal_add_js(drupal_get_path('module', 'dosomething_helpers') . '/dosomething_helpers_nodeform.js');
   }
 }
+
+/**
+ * Gets node nid's that reference the given $nid in a given $field_name.
+ *
+ * @param int $nid
+ *  The node nid value stored as the given field's target_id.
+ * @param string $field_name
+ *  The machine name of the entityreference field to look for parents.
+ *
+ * @return mixed
+ *  Array of entity objects (not loaded).  FALSE if no parents found.
+ */
+function dosomething_helpers_get_entityref_parents($nid, $field_name) {
+  $query = new EntityFieldQuery;
+  $result = $query->entityCondition('entity_type', 'node')
+    ->fieldCondition($field_name, 'target_id', $nid)
+    ->execute();
+  if (isset($result['node'])) {
+    return $result['node'];
+  }
+  return NULL;
+}
+
+/**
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_image/dosomething_image.module
```diff
@@ -76,3 +76,26 @@ function dosomething_image_get_themed_image_url($nid, $ratio, $style) {
...
+      'field_gallery_image',
+      'field_hero_image',
+      'field_image_campaign_cover',
+      'field_image_psa_replacement',
+      'field_image_reportback_gallery',
+      'field_image_sponsor_logo',
+      'field_image_step_gallery',
+      'field_intro_image',
+    );
+    dosomething_helpers_add_entityref_links($node, $entityref_fields);
```
angaither, line 99: this would be a good place to add caching in a later lifetime.

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -22,4 +22,77 @@ function dosomething_helpers_form_alter(&$form, &$form_state, $form_id) {
...
+ *  Array of machine name of fields to check for.
+ */
+function dosomething_helpers_add_entityref_links(&$node, $fields) {
+  $i = 30;
+  foreach ($fields as $fld) {
+    $node->content[$fld] = array(
+      '#prefix' => '<div><h3>' . $fld .'</h3>',
+      '#suffix' => '</div>',
+      '#markup' => dosomething_helpers_get_entityref_parent_list($node->nid, $fld),
+      '#weight' => $i,
```
angaither, line 94: why 30 + 10?

aaronschachter, line 94: That's a good question.... it was just a random number I chose to weight these lists to show up at the end of the node content.  I suppose it could be bumped up to a greater number... or do some dynamic stuff to figure out the highest existing weight, but didn't go into it too much since this is only for internal usage.  Open to any suggestions if it seems too random...

angaither:
:+1:

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
angaither, line 94: We may want to move this into a tpl like static content eventually.

PR #631, opened 2014-02-07
Campaign Action Page

mmwtsn:
# WIP
## Known Issues
- [x] NEUE.pinToTop isn't triggering
- [x] grid overlay not available in development
- [x] grid widths not adding up (e.g. shift(2) + span(5) + span(5) = broken layout)
- [ ] content limits are not functioning (dummy content is too long)
- [ ] real content needed to visually compare against mock (ideally in an environment that will not get destroyed; e.g. staging)
- [ ] background placeholder image needs to be replaced
- [ ] fact solutions need to be decoupled from their facts so that they can be printed lower in the template
- [ ] _campaign.scss is very long and becoming unorganized
- [x] broken tags are being generated form the editor in the "Do It" section when Markup is selected
## Questions
- is this how Neat was supposed to be used? I'm `@include outer-container;` often

Fixes dosomething/dosomething#553

**Under** :construction: **not ready for merge.**

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/feature/_campaign.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/feature/_campaign.scss
@@ -18,9 +18,9 @@ section.campaign-wrapper {
   }
 }
 
-// Inner campaign template container that
-// holds all content from the node fields
-div.campaign-container {
+// Campaign header containing that campaign's
+// metadata sourced from the node fields
+header.campaign-meta {
   @include media( $tablet ) {
     @include span-columns(3);
     @include shift(1);
--- a/lib/themes/dosomething/paraneue_dosomething/scss/feature/_campaign.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/feature/_campaign.scss
@@ -3,41 +3,49 @@
 //  CAMPAIGN
 // ----------
 
-// Outermost campaign template wrapper that
-// has the hero image applied to it
-section.campaign-wrapper {
-  min-height: 600px;
-  padding: 250px 0 0;
+.c--wrapper {
   overflow: hidden;
-  background-repeat: no-repeat;
-  background-position: 43% 50%;
-  background-size: cover;
 
   @include media( $tablet ) {
     @include outer-container;
   }
-}
 
-// Campaign header containing that campaign's
-// metadata sourced from the node fields
-header.campaign-meta {
-  @include media( $tablet ) {
-    @include span-columns(3);
-    @include shift(1);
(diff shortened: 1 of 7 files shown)


DFurnes:
Regarding `NEUE.pinToTop` not triggering, we should be switching to use a `position: sticky` polyfill anyway. I'll look into that (DoSomething/ds-neue#121).

aaronschachter:
Something to watch out for with Facts.  Just implemented changes to allow multiple sources per a given fact.  Note change in the template file:  https://github.com/DoSomething/dosomething/blob/dev/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php#L19

To get these changes in, the easiest way would be do a fresh installation. If there are db changes to keep locally, you'll need to truncate the tables listed here,: https://github.com/DoSomething/dosomething/pull/612, and then rebase from dev to get the latest content type definitions.

DFurnes:
:dancers:

DFurnes:
:boom:

lib/themes/dosomething/paraneue_dosomething/scss/feature/_campaign.scss
```diff
@@ -1,43 +1,173 @@
...
 // ----------
 //  CAMPAIGN
 // ----------
 
-// Outermost campaign template wrapper that
-// has the hero image applied to it
-section.campaign-wrapper {
-  min-height: 600px;
-  padding: 250px 0 0;
+.c--wrapper {
```
DFurnes, line 5: Perhaps `cpm--wrapper` for a bit better readability?

lib/themes/dosomething/paraneue_dosomething/scss/feature/_campaign.scss
```diff
@@ -1,43 +1,173 @@
...
+      @include span-columns(4);
+      @include shift(2);
+    }
+  }
+
+  div.col.second {
+    @include media( $tablet ) {
+      @include span-columns(4);
+      @include shift(1);
+      @include omega;
```
DFurnes, line 166: We can just use the `nth-child` argument to `omega()` to use a single `.col` class for this. :turtle:

lib/themes/dosomething/paraneue_dosomething/scss/feature/_port-to-neue.scss
```diff
@@ -1,4 +1,8 @@
-div.campaign-container {
+@mixin section-padding {
+  // padding: 0 1.5rem;
```
DFurnes, line 2: :bangbang:

PR #632, opened 2014-02-10
User login updates

### LINKED ISSUE #614: User login/registration form fields
Currently, we're using the default Drupal login/registration fields in the new world. We're running into a couple issues where default behavior conflicts with our existing user flow: 
- We can't authenticate users if they provide either an email+password cell phone+password. This is inconsistent with our current login behavior.
- We prompt users to create a username when registering. I think the current behavior is to generate a unique username for them behind-the-scenes.

### LINKED ISSUE #619: Members need to be able to log in with email, cell (username?)
Dave and Maxwell originally flagged this. On our current site lots of members log in using the cell number, username or email address. 

Can users currently do this in the new login flow?

angaither:
- moved campaign signup calls to dosomething_user
- allow users to log in via mobile number or email
- validation/clean cell numbers

Fixes #619
Fixes #614

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -139,62 +139,9 @@ function dosomething_signup_form_submit($form, &$form_state) {
     return;
   }
   $nid = $form_state['values']['nid'];
-  // Signup shouldn't exist already, but check. 
+  // Signup shouldn't exist already, but check.
   if (!dosomething_signup_exists($nid) && dosomething_signup_insert($nid)) {
     drupal_set_message(t("You're signed up!"));
-  }
-}
-
-/**
- * Implements hook_form_alter().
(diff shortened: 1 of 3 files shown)


aaronschachter:
The code all looks okay - I can see that some of these function comments and names are copied from https://github.com/DoSomething/DoSomething-7.x/blob/dev/sites/all/modules/dosomething/dosomething_general/dosomething_general.module, and with along with it is errors in the commenting standards.  Can we run coder and fix?

Also let me know what you think, but I think it'd be cleaner to rename the functions to use _get_ in the function name instead of _find_ to keep consistent sitewide, and to use mobile instead of cell e.g. 

```
dosomething_user_get_user_by_mobile
dosomething_user_get_user_by_email
dosomething_user_get_clean_mobile
```

angaither:
@aaronschachter updated

aaronschachter:
:+1:

PR #634, opened 2014-02-10
Form tweaks

aaronschachter:
- Fixes issues with Season Dates in Campaign form
- Makes Fact Title a textarea, removes body from Fact node type instance


angaither:
:+1:

PR #635, opened 2014-02-10
11 Facts - back end theming

blisteringherb:


lib/modules/dosomething/dosomething_fact_page/dosomething_fact_page.module
```diff
@@ -5,3 +5,55 @@
...
+    }
+
+    $node = entity_metadata_wrapper('node', $vars['node']);
+    if ($node->field_video->value()) {
+      $vars['intro_video'] = theme('dosomething_video_embed', array('field' => $node->field_video->value()));
+    }
+
+    $vars['facts'] = dosomething_fact_get_fact_field_wrapper_values($node->field_facts);
+    foreach ($vars['facts'] as $key => $fact) {
+      $vars['facts'][$key]['sources'] = theme('item_list', array('items' => $fact['sources']));
```
aaronschachter: This works for now, although you may want to check with might want to check in with @mmwtsn or @DFurnes  -- they may prefer to have the raw array of sources data available as a variable vs a themed list, which is how it's set up on the campaign's tpl.php.

mshmsh5000:
@blisteringherb Any idea why the wercker build failed? Looks like a Drush-level failure, locating/installing the required Dosomething_\* modules.

aaronschachter:
This PR looks very scary now :( ... any idea why all the extra commits?

blisteringherb:
I don't know what happened here.  I opened a new PR.  I'm gonna close this one.

PR #643, opened 2014-02-10
Theming for 11 facts page

blisteringherb:
- Adding new tpl file for 11 facts
- Adding fact page module to modify theme registry paths
- Adding vars for 11 facts


aaronschachter:
:+1:

PR #653, opened 2014-02-11
User has unique mobile number.

### LINKED ISSUE #644: User: Mobile must be unique
Only one mobile number per user.  Per oldworld -- if a user tries to sign up with an existing phone number, there is messaging around "DId you forget your password?" with a link to reset password.

angaither:
Fixes #644


aaronschachter:
Looks good, this works for user/*/edit form too?

angaither:
Yup, the form_alter has a case for both `user_register_form` and `user_profile_form`

aaronschachter:
ohhh, i see.  very nice, ship it value added :+1:

PR #655, opened 2014-02-11
Adding entity connect perms for editors

blisteringherb:


aaronschachter:
Seems ok. Could arguably belong in `dosomething_campaign` too, or perhaps `dosomething_helpers`.  But i think fine to keep here for now.

Issue #450, opened 2014-01-23
Campaign: Mobile Commons opt in with campaign name on sign up


jonuy:
Two things should happen when a user signs up for a campaign and provides a mobile number:
- Get subscribed into a specified Mobile Commons opt-in path ID
- Pass along the campaign name to a Mobile Commons field

This'll be something like a POST to `https://secure.mcommons.com/profiles/join` with:
- `opt_in_path[]` = the opt in path id
- `person[phone]` = the mobile number to subscribe
- `person[campaign_name]` = the name of the campaign the user signed up for

Probably should use the `mobilecommons` module to make the API call.

aaronschachter:
Agree about using mobilecommons module to handle the API call.  Could multiple campaigns subscribe to the same opt-in path ID, or will each opt-in always be unique?

mikefantini:
@aaronschachter multiple campaigns will subscribe users to the same opt-in path ID (all non-staff pick campaigns will point to the same opt-in path). 

cc: @jonuy

mikefantini:
@aaronschachter the opt-in path for all non-staff pick campaigns is 164453. Are you good on adding this to the campaign node or is there more info you need from me?

cc: @jonuy

aaronschachter:
If this is same across the board, then we don't need to add this into a campaign node for now.  however we don't have anything communicating to Mobilecommons in this app right now. As per @jonuy 's comment, we should use the mobilecommons contrib module but haven't written any code using it yet.  Is mobile signups a requirement for 0.2?

mikefantini:
Yeah we do need these going into mobile commons for 0.2 and pull in
campaign_name. First time we would have a campaign that would use a
different mobile commons id is mid-march.

On Mon, Feb 10, 2014 at 3:24 PM, Aaron Schachter
notifications@github.comwrote:

> If this is same across the board, then we don't need to add this into a
> campaign node for now. however we don't have anything communicating to
> Mobilecommons in this app right now. As per @jonuyhttps://github.com/jonuy's comment, we should use the mobilecommons contrib module but haven't
> written any code using it yet. Is mobile signups a requirement for 0.2?
> 
> ## 
> 
> Reply to this email directly or view it on GitHubhttps://github.com/DoSomething/dosomething/issues/450#issuecomment-34677880
> .

## 

Mike Fantini
Head of Product, DoSomething.org
212.254.2390 x232

follow me: @mfantini12
follow us: @dosomething

[...]

PR #659, opened 2014-02-11
Mobilecommons Opt-In

aaronschachter:
Creates a function in `dosomething_signup` to opt-in a user into a mobilecommons path as specified by #450.

This looks like the request is successful (you need to enable mobilecommons module and add credentials to test) but it doesn't seem like the opt-in path is actually set up, because I'm not receiving any texts.  Not closing out #450 yet as we'll still need to alter this to pass through a campaign nid/name into the opt-in request as well.


blisteringherb:
:+1:

PR #662, opened 2014-02-12
Include custom modules & themes via make

desmondmorris:
**Why would I want to do something like this?**
Because it makes the project easier to package and deploy.  Currently, we would have to symlink the modules and themes directories every time we deploy or package up a dist.  This way, all of our custom modules and themes are included by default.

**Thats confusing, and annoying.**
No, its not.

**Yes it is**
Nope, I gave you a nice dosomething.make file in the profile directory.  This is where need to include the custom modules and themes.  Yes, it adds another step when creating new modules, but this should not happen very often

**But, I like how things work now.**
Well, thats too bad, plus everything should be business as usual.

Thoughts/Issues folks?

### DIFF
--- a/lib/profiles/dosomething/dosomething.make
+++ b/lib/profiles/dosomething/dosomething.make
@@ -1,0 +1,92 @@
+; Dosomething custom modules and themes
+api = 2
+core = 7.x
+
+includes[] = drupal-org.make
+
+; MODULES
+
+; Dosomething Action Guide
+projects[dosomething_action_guide][type] = "module"
+projects[dosomething_action_guide][download][type] = local
+projects[dosomething_action_guide][download][source] = './lib/modules/dosomething/dosomething_action_guide'
+projects[dosomething_action_guide][subdir] = "dosomething"
+
+; Dosomething Campaign
+projects[dosomething_campaign][type] = "module"
+projects[dosomething_campaign][download][type] = local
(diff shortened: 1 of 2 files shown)


lib/profiles/dosomething/dosomething.make
```diff
@@ -0,0 +1,92 @@
...
+projects[dosomething_taxonomy][subdir] = "dosomething"
+
+; Dosomething User
+projects[dosomething_user][type] = "module"
+projects[dosomething_user][download][type] = local
+projects[dosomething_user][download][source] = './lib/modules/dosomething/dosomething_user'
+projects[dosomething_user][subdir] = "dosomething"
+
+; THEMES
+; Dosomething User
```
angaither: Paranue?

angaither:
if this removes the need for the module & theme symlinks, can we remove those lines in the ds build script?

lib/profiles/dosomething/dosomething.make
desmondmorris: Ahh good catch.  I will fix that

desmondmorris:
@angaither We will still need the symlinks in the build because they allow us to work directly form the modules and themes folder from our host/vm machines

angaither:
@desmondmorris ahhhh, gotcha.

angaither:
:+1:

DFurnes:
screen shot 2014-02-12 at 11 57 49 am

Marry me.

PR #663, opened 2014-02-12
Capistranos return

desmondmorris:
### Capistrano is back

san juan capistrano

This merge re-introduces capistrano configs.  Currently it includes a config to deploy to "production".  This is the production web head I have been working on.

 `cap production deploy`, will ssh into the webhead, clone down our repo, run `drush make` against the build make file (which now includes custom profiles and themes #662) and symlinks the settings directory.

We will need to do some things with environment variables to rig up the db and caching etc.  We will also need to flesh out the staging config as well.

**Requirements**
- You must have Capistrano 3 installed
- You must be connected to the OpenStack VPN to run deploys at this point
- Your key must be present on the webhead

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -1,0 +1,25 @@
+# Load DSL and Setup Up Stages
+require 'capistrano/setup'
+
+# Includes default deployment tasks
+require 'capistrano/deploy'
+
+# Includes tasks from other gems included in your Gemfile
+#
+# For documentation on these, see for example:
+#
+#   https://github.com/capistrano/rvm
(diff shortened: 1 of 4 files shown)


mshmsh5000:
Looks good, my cap dry run seemed to work.

DFurnes:
@desmondmorris How will this work with building the compiled front-end CSS/JS, since those aren't stored in Git?

desmondmorris:
@DFurnes we will need to add a build assets task to the capistrano config.

PR #664, opened 2014-02-12
Update all image fields to use a view to return entity results.

### LINKED ISSUE #650: Image Node Reference view alter
When referencing an image node, the nid displays, the crops available displays & maybe the thumbnail

angaither:
Allows override of results, to display nid & thumbnail in results.
Fixes #650


desmondmorris:
Seems okay to me

PR #668, opened 2014-02-12
Edit ui image node fancify.

### LINKED ISSUE #667: Image: node updates
- automagically title image nodes
- hide title field
- automagically submit uploaded photos (users don't need to click 'upload')

angaither:
Fixes #667


angaither:
@aaronschachter added image title back in.

aaronschachter:
This looks good.  I think it's fine to keep in `dosomething_image` as is for now, but wonder if it's worth moving the JS into `dosomething_helpers` as a general file-upload script?

angaither:
mayhaps if we find the need for this with other file uploads.

PR #673, opened 2014-02-13
Static content

barryclark:


lib/themes/dosomething/paraneue_dosomething/scss/app.scss
```diff
@@ -13,7 +13,7 @@
 // import paraneue's bower installation of Neue
 @import "../../../../html/profiles/dosomething/themes/dosomething/paraneue/paraneue/bower_components/pattern-library/scss/helpers";
 
-// features
+// Features
```
mmwtsn, line 16: :+1:

lib/themes/dosomething/paraneue_dosomething/scss/feature/_content.scss
```diff
@@ -0,0 +1,8 @@
+
+// ---------
+//  CONTENT
```
mmwtsn, line 3: "Content" is rather generic. It would be good to change this to something more descriptive in the future.

PR #681, opened 2014-02-18
Adding Markdown support to static content

### LINKED ISSUE #641: Markdown editing for static content fields
Probably awaiting more info from #640.

blisteringherb:
Fixes #641


aaronschachter:
:+1: Looks good. we'll also need to update the Action Guide content type similarly as it uses these fields.  I can do this or you can take a stab at it @blisteringherb, as i'm in broken dev box land.

Issue #679, opened 2014-02-18
Campaign: Confirmation template


aaronschachter:
Create a page callback for confirmation of reportback submission.

Update Campaign's reportback submission to redirect user to this page upon success.

mikefantini:
This confirmation page should have 3 campaigns it suggests.

We'll show:
- hero image
- title
- call to action

This will look like the gallery on static content: https://docs.google.com/a/dosomething.org/viewer?a=v&pid=sites&srcid=ZG9zb21ldGhpbmcub3JnfGRvc29tZXRoaW5ndXh3aWtpfGd4OjNkNGY1ODQyODE2MDMwMmM

We need to have 4 campaigns the confirmation page can select from and add logic so that it doesn't suggest the campaign the user just reported back on.

PR #718, opened 2014-02-19
Campaign confirmation template

aaronschachter:
Works toward #679.  Creating this PR now so it won't be huge once we add in the suggested campaigns. 
- Creates menu, access, and page callback functions for a node/*/confirmation page, which is a campaign confirmation page.
- Redirects user to the campaign confirmation page upon submitting a reportback.

Still left todo for a future PR:
- Display 3 campaigns in the confirmation template, either via variables or dynamically finding campaigns user has not signed up for.
- Move module tpl into a templates folder, and create an override tpl in the dosomething_paranaue directory.


angaither:
:+1:

PR #723, opened 2014-02-19
Reportback fix

aaronschachter:
Forgot to remove this code when I removed the `num_participants` property from the reportback table.


angaither:
:+1:

PR #725, opened 2014-02-19
Blank Mobile Number Fix

### LINKED ISSUE #680: Login can fail with blank mobile number
No matter what username I use, aschachter or admin, it tells me the password is incorrect.  I've reset the password so I know it's correct.

The error message, "Have you forgotten your password?" is linking me to http://stage.dosomething.org/user/password?name=rmohammed, suggesting that this is a caching issue.  When I try to add a query string into the URL however, I still get directed to the rmohammed reset link.

screen shot 2014-02-18 at 12 13 35 pm

angaither:
Fixes #680


aaronschachter:
value added :+1:

PR #732, opened 2014-02-19
Report back modal classes

### LINKED ISSUE #717: Hook up report back modal
The modal doesn't need to be styled yet, but getting it triggering and functional so that we can test the full campaign flow is a priority. 

[Modal Documentation](https://github.com/DoSomething/ds-neue/blob/dev/js/modal.js)

mmwtsn:
# "Tell Us" Modal

Links the report back modal to the campaign template.

Fixes #717


aaronschachter:
Looks ok to me

PR #733, opened 2014-02-19
Update Profile/Add Users

### LINKED ISSUE #719: User edit form redirect
As an administrator, if you go to edit another user, you are redirected back to your own user edit form, and the changes are not saved.

### LINKED ISSUE #720: User Edit: Can't update profile and save mobile number
A user cannot update their own phone number on the user/[your-id]/edit

### LINKED ISSUE #730: User Add Fails without error
Can't add new user
the 'admin/user/create' form_id is the same as the user/register form_id so the action is forcing it to post to user/register

### LINKED ISSUE #731: User Edit: Fails if mobile is set
If a user updates profile and leaves mobile number, it fails. Must check if user number was previously set.

angaither:
Open to suggestions on how to better handle the case statements for having different events happen on login/registration. :confused: 

Fixes #719
Fixes #730
Fixes #731
Fixes #720

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -122,8 +122,12 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
       // After build form changes.
       $form['#after_build'][] = 'dosomething_user_register_after_build';
 
+      // Force action to post to the secure user registration, but not on the add people screen.
+      if ($form_id == 'user_register_form' && $_SERVER['REQUEST_URI'] != '/admin/people/create') {
+        $form['#action'] = $secure_url . '/user/register';
+      }
+
       // Force registration to submit to secure registration.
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -122,8 +122,12 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
       // After build form changes.
       $form['#after_build'][] = 'dosomething_user_register_after_build';
 
+      // Force action to post to the secure user registration, but not on the add people screen.
+      if ($form_id == 'user_register_form' && $_SERVER['REQUEST_URI'] != '/admin/people/create') {
+        $form['#action'] = $secure_url . '/user/register';
+      }
+
       // Force registration to submit to secure registration.
```
aaronschachter: Should remove this comment too?

aaronschachter:
Looks ok except for leftover comment. The if conditional inside the case statement seems fine to me.

PR #741, opened 2014-02-20
Confirmation Page content

### LINKED ISSUE #679: Campaign: Confirmation template

aaronschachter:
Closes #679
- Moves the dosomething_campaign module's reportback-confirmation.tpl.php file into a /theme subdirectory
- Creates an override reportback-confirmation.tpl.php in the dosomething_paraneue/templates/campaign directory
- Adds a bunch of `dosomething_campaign_get_recommended` functions to return campaigns which the user has not signed up for.
  - First finds any staff picks that the user has not signed up for
  - If less than 3, finds any campaigns which share the same Primary Cause that user has not signed up for
- Renders them as recommended campaigns on the Campaign Reportback Confirmation page.
- Adds documentation


lib/modules/dosomething/dosomething_campaign/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+The confirmation page displays 3 recommended campaigns for the user.  This is
+determined by first finding published campaigns which are staff picks, that the
+user has not signed up for.
+
+If there are less than 3 staff picks that the user hasn't signed up for, the
+confirmation page will also display published campaigns which share the same
+Primary Cause, for which the user has not signed up for.
+
+
```
angaither, line 61: :+1:

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -96,32 +96,73 @@ function dosomething_campaign_reportback_confirmation_page_access($node) {
...
 /**
+ * Returns an array of variables for a given recommended campaign $nid.
+ *
+ * @param int $nid
+ *   The node $nid to return variables for.
+ *
+ * @return array
+ *   An array of variables.
+ */
+function dosomething_campaign_get_recommended_campaign_vars($nid) {
```
angaither, line 136: Small thing, but perhaps the function should be something more generic like 'campaign_teaser_vars'
Actually, it's pretty similar to the pitch page variables.

aaronschachter, line 136: The pitch page would be a different crop I'd think... which would make sense to keep in a separate "recommended" function.  I'm happy to rename later if we want to reuse it, what do you think?

aaronschachter, line 136: Discussed with @angaither -- we can visit building a generic teaser vars function in a later PR.

PR #742, opened 2014-02-20
Switching to using apachesolr module instead of Search API

blisteringherb:
- Apachesolr module properly indexes all fields for all types


lib/modules/dosomething/dosomething_search/dosomething_search.apachesolr_environments.inc
```diff
@@ -0,0 +1,36 @@
...
+ * Implements hook_apachesolr_environments().
+ */
+function dosomething_search_apachesolr_environments() {
+  $export = array();
+
+  $environment = new stdClass();
+  $environment->api_version = 1;
+  $environment->env_id = 'solr';
+  $environment->name = 'localhost server';
+  $environment->url = 'http://192.168.1.21:8080/solr/collection1';
```
aaronschachter: I'd imagine we'd want to store the name and URL outside of code, no?  When I go that URL I get "The requested resource is not available."

lib/modules/dosomething/dosomething_search/dosomething_search.apachesolr_environments.inc
```diff
@@ -0,0 +1,41 @@
...
+  $environment->conf = array(
+    'apachesolr_direct_commit' => 0,
+    'apachesolr_index_updated' => 1392910246,
+    'apachesolr_last_optimize' => 1392910246,
+    'apachesolr_read_only' => '0',
+    'apachesolr_soft_commit' => 0,
+  );
+  $environment->index_bundles = array(
+    'node' => array(
+      0 => 'fact',
```
aaronschachter: I don't believe we want fact nodes to show up in search.  The fact node just exists for internal usage (so a user would get an access denied upon viewing a fact/[nid] page.

aaronschachter:
This looks good, although I wonder if it's going to break things without a new build, as the search module needs a new module to be downloaded via the build script.  @mshmsh5000 @desmondmorris -- thoughts?  I'd imagine that the search feature would just be disabled...

mshmsh5000:
I think we need to update the Jenkins build to run drush make into a temp directory, then switch that with the live html dir. Like

`drush make /vagrant/build-dosomething.make html-new`

and so on. Need to redo the symlinks to settings and files, etc.

mshmsh5000:
To be clear -- Jenkins-side changes shouldn't block this PR.

aaronschachter:
ok, sounds good :+1:

PR #750, opened 2014-02-20
Mobilecommons signup

### LINKED ISSUE #450: Campaign: Mobile Commons opt in with campaign name on sign up

aaronschachter:
Closes #450

Creates a `dosomething_signup_user_signup` function which handles inserting a signup record and the API request to opt-in to Mobilecommons.

Uses this new function to replace  manual `dosomething_signup_insert` calls in `dosomething_user_login_submit` and `dosomething_signup_form_submit`.

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -244,8 +244,12 @@ function dosomething_signup_user_signup($nid, $uid = NULL) {
 
   // Insert signup.
   if ($sid = dosomething_signup_insert($nid, $uid)) {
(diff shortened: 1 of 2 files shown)


angaither:
:+1:

Issue #752, opened 2014-02-20
Message Broker: Example Code


angaither:
Per @DeeZone 
- New User Registration (not login)
- Password Reset requests
- Campaign Signups (it's ok if this results in a new user registration as well producing two transactional email messages / message_broker_producer_request()s
- Campaign Report backs

barryclark:
Are you the assignee of this one, @angaither?

PR #753, opened 2014-02-20
Stub out external mail requests.

angaither:
Refs #752 

@DeeZone

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -257,7 +257,9 @@ function dosomething_signup_user_signup($nid, $account = NULL) {
     $params = array(
       'nid' => $nid
     );
-    dosomething_user_send_message_request('campaign-signup', $params);
+    if (module_exists('dosomething_user')) {
+      dosomething_user_send_message_request('campaign_signup', $params);
+    }
     // Set success message.
     $message = t("You're signed up!") . '<br />';
     $message .= t("Get started with") . ' ' . $node->title .'  ' . t("below!");
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -253,6 +253,11 @@ function dosomething_signup_user_signup($nid, $account = NULL) {
     $node = node_load($nid);
     // Opt-in to mobilecommons.
     dosomething_signup_mobilecommons_opt_in($account, $node->title);
+    // Send external message request
+    $params = array(
+      'nid' => $nid
+    );
+    dosomething_user_send_message_request('campaign-signup', $params);
     // Set success message.
     $message = t("You're signed up!") . '<br />';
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -371,6 +376,31 @@ function dosomething_user_clean_cell_number($number) {
...
+ *
+ * @param string $origin
+ *  Where the message request came from.
+ * @param array $params
+ *  Optional keyed array of data to send externally.
+ *  - nid - campaign nid.
+ * @TODO: add more as needed.
+ *
+ */
+function dosomething_user_send_message_request($origin, $params = NULL) {
```
deezone, line 377: This looks great for my needs.

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -141,6 +141,10 @@ function dosomething_signup_form_submit($form, &$form_state) {
   $nid = $form_state['values']['nid'];
   // Signup shouldn't exist already, but check.
   if (!dosomething_signup_exists($nid) && dosomething_signup_insert($nid)) {
+    $params = array(
+      'nid' => $nid,
+    );
+    dosomething_user_send_message_request('campaign-signup', $params);
```
aaronschachter: This code is old as of half hr ago or so -- the `send_message_request` should be called from within `dosomething_signup_user_signup`

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -371,6 +376,31 @@ function dosomething_user_clean_cell_number($number) {
...
+  //@TODO: gather array to send externally & make request.
+  switch ($origin) {
+    case 'user_register_form':
+      break;
+    case 'user_pass':
+      break;
+    case 'campgin-signup':
```
aaronschachter: I know its still skeleton code, so this is minor but FYI - spelling.  Same with "reqest" in the block comments.

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
angaither: :bomb:

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -358,6 +363,31 @@ function dosomething_user_clean_cell_number($number) {
...
+ *
+ */
+function dosomething_user_send_message_request($origin, $params = NULL) {
+  //@TODO: gather array to send externally & make request.
+  switch ($origin) {
+    case 'user_register_form':
+      break;
+    case 'user_pass':
+      break;
+    case 'campaign-signup':
```
aaronschachter: Missed this earlier, but these campaign values should use underscores instead of dashes, to match the format of the other cases: `user_register_form` , `user_pass`.  So change to `campaign_signup`, `campaign_reportback`.  Otherwise good to go.

angaither:
Closing, code is now in #775

PR #758, opened 2014-02-20
Campaign automatic revisions

aaronschachter:
Creates new revision every time a campaign is updated.  This change has already been made into the staging database.


blisteringherb:
:+1:

PR #767, opened 2014-02-21
Campagin end date not set.

### LINKED ISSUE #754: Campaign End Date error
If you create a campaign without a High Season Start/End:

EntityMetadataWrapperException: Unable to get the data property value2 as the parent data structure is not set. in EntityStructureWrapper->getPropertyValue() (line 438 of /vagrant/html/profiles/dosomething/modules/contrib/entity/includes/entity.wrapper.inc).

Need to modify `dosomething_campaign_preprocess_node` so that it only displays End Date if there's an actual value set in `$wrapper->field_high_season->value2->value())`

angaither:
Fixes #754


aaronschachter:
:+1:

PR #771, opened 2014-02-21
Help text changes

aaronschachter:
yeah, computer science!

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.features.field_instance.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.features.field_instance.inc
@@ -364,9 +364,9 @@ function dosomething_campaign_field_default_field_instances() {
     'bundle' => 'campaign',
     'default_value' => NULL,
     'deleted' => 0,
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.features.field_instance.inc
```diff
@@ -956,8 +952,6 @@ function dosomething_campaign_field_default_field_instances() {
       ),
     ),
     'entity_type' => 'node',
-    'entityconnect_unload_add' => 0,
-    'entityconnect_unload_edit' => 0,
```
angaither, line 960: Are we killing these guys intentionally?

aaronschachter, line 960: I looked at this too - and not exactly. When I did a features-update using the staging database, these changes came through.  I don't think it should affect the content type, as the form looks fine and the entityconnect buttons appear on all of the image fields.

angaither:
:+1:

PR #773, opened 2014-02-21
Allow editors publish/unpublish

aaronschachter:
The "administer nodes" permission will allow editors to publish and unpublish.

It also allows them to do things like change authoring information, "promoted" and "sticky" .. but we could always just hide this stuff if it's too much.


angaither:
:+1:

PR #775, opened 2014-02-21
Message broker call update1

deezone:
Refs #752 

First draft of: 
- dosomething_user_send_message_request() - suggested $payload values and resulting call to message_broker_producer_request()
- Updated sub calls to dosomething_user_send_message_request() in dosomething_reportback, dosomething_signup and dosomething_user. Need to complete list of $params values.

I'll build out in the next pull request the $params values in the calling functions to better define what is needed.


aaronschachter:
What's the difference between this and #753 ? Same comment applies with the mismatching of underscores and hyphens for expected values in the $origin param.

deezone:
@angaither I'm not sure if I did this right -> new branch off of your branch after cloning of your fork of the dosomething repo but...

Here is my first round of code changes, thoughts?

angaither:
This is looking good, except the underscores. 
You can grab user data from `global $user`
The campaign data is all in the campaign node, I can help out with that if you need it.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -484,6 +484,23 @@ function dosomething_reportback_save($entity, $values) {
...
+    // Send external message request
+    if (module_exists('dosomething_user')) {
+      if ($user == NULL) {
+        global $user;
+        $account = $user;
+      }
+      $params = array(
+        'email' => $account->mail,
+        'uid' => $account->uid,
+        'first_name' => '??',
```
angaither: Create a `dosomething_user_get_first_name` function
see https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_user/dosomething_user.module#L395

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -133,6 +144,17 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
 
     break;
 
+    case 'user_pass':
+      // Send external message request
+      $params = array(
+        'email' => $account->mail,
+        'uid' => $account->uid,
+        'first_name' => '??',
+        'reset_link' => '??',
```
angaither: see `user.module` 
Should be able to call `user_pass_reset_url($account)` to generate this link.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -484,6 +484,23 @@ function dosomething_reportback_save($entity, $values) {
...
+        'event_id' => '??',
```
angaither, line 497: is this supposed to be the reportback id? 
or the node id that the user is reporting back for?

deezone:
@angaither Can we merge this in as is? There's only a few more field values to collect for the dosomething_user_send_message_request() calls. The code will work with the current '??' values. I'd like to do some testing with the Message Broker system using this code with an updated version of the code base before going further with this pull request.

angaither:
:+1:

PR #782, opened 2014-02-21
Action guide preprocess / template file

aaronschachter:
- Creates a base tpl file for the Action Guide content type and places it into the module directory.
- Updates `dosomething_helpers_modules` variable to include action guide.
- Documentation!

FYI @blisteringherb  - There seems like there may be variables missing from the original preprocess gallery functions, like the Gallery Title.  We'll need to tweak both content types' preprocess and template files as we fix any bugs.

### DIFF
--- a/lib/modules/dosomething/dosomething_action_guide/dosomething_action_guide.module
+++ b/lib/modules/dosomething/dosomething_action_guide/dosomething_action_guide.module
@@ -5,3 +5,32 @@
  */
 
 include_once 'dosomething_action_guide.features.inc';
+
+/**
+ * Implements hook_preprocess_node().
+ */
+function dosomething_action_guide_preprocess_node(&$vars) {
+  if ($vars['type'] != 'action_guide') { return; }
+
+  if ($vars['view_mode'] == 'full') {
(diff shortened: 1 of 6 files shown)


lib/modules/dosomething/dosomething_action_guide/dosomething_action_guide.module
```diff
@@ -5,3 +5,32 @@
...
+    $text_vars = array(
+      'subtitle',
+      'intro_title',
+      'intro',
+      'additional_text_title',
+      'additional_text',
+    );
+    foreach ($text_vars as $label) {
+      $field = "field_" . $label;
+      $vars[$label] = $wrapper->{$field}->value();
```
angaither, line 27: Is this guy going to get mad about blank values?

aaronschachter, line 27: Nope, thats whats neat about the entity_metadata_wrapper value() jam, is that it will set the $vars[$label] to NULL if its not set.

angaither:
:+1:

Issue #675, opened 2014-02-18
ds script – update sandbox environment with content from staging site


mmwtsn:
Update `dev` flag of `ds build` to pull the latest from the staging database rather than generating placeholder content.

aaronschachter:
This is related to / possibly duplicating #607

mmwtsn:
Good catch. Should this exist as a new flag or a replacement for `dev`?

desmondmorris:
The steps documented here should do it for you.  This needs more eyes to test

https://gist.github.com/desmondmorris/3c5dd8392a26efc5a840

mshmsh5000:
@mmwtsn @desmondmorris Make sure you see [my addenda to those instructions](https://gist.github.com/desmondmorris/3c5dd8392a26efc5a840#comment-1178746).

desmondmorris:
@mshmsh5000 yah thanks for that.  I mentioned some of this here: https://github.com/DoSomething/dosomething/pull/789

PR #789, opened 2014-02-24
Adds hosts and black angus alias files via salt state.  Related to #675

desmondmorris:
This merge allows us to safely pull down the staging db via the drush `sql-sync` command.  This is all possible by using drush aliases.  The following command will pull down the ds_stage db and import it into the dosomething db within your vm:

```
drush sql-sync @blackangus.stage @self
```

**Important**
1.  You must be on the DS network for this to work.  So if you are remote, you must be connecting to the VPN.
2.  Currently, you must add the proper mysql credentials to the alias file for this to work.  You can find the full db string (here)[https://gist.github.com/desmondmorris/66ed52d15d6b3486235a].  You will need to replace the db string in /home/vagrant/.drush/blackangus.aliases.drushrc.php with the one in that link I provided.  I think we may be able to get around this by configuring the mysql db on stage to accept read only pulls from any request within our network (VPN). @mshmsh5000 thoughts on that?
3.  You must be with a working drupal webroot for the self alias to work.  So make sure you cd into /vagrant/html first.

**Coming soon**
With aliases we will also be able to easily rsync the files directory down and run drush commands against the staging server.

### DIFF
--- a/provision/salt/roots/salt/dosomething.sls
+++ b/provision/salt/roots/salt/dosomething.sls
@@ -1,0 +1,13 @@
+{% if grains['os'] == 'Ubuntu' %}
+
+ds-hosts:
+  file.managed:
+    - name: /etc/hosts
+    - source: salt://dosomething/hosts
+
+ds-aliases:
+  file.managed:
+    - name: /home/vagrant/.drush/blackangus.aliases.drushrc.php
+    - source: salt://dosomething/blackangus.aliases.drushrc.php
+
+{% endif %}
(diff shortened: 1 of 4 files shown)


desmondmorris:
Actually @mshmsh5000, any opposition to making the ds_read user passwordless since it is only accessible behind VPN?

mshmsh5000:
Re: passwordless: ehhhh okay. We have the jenkins user configured that way. And given that this code base will eventually be public, and we're linking to passwords here, it falls within "security theater" anyway. I'll work on it.

mshmsh5000:
K, ds_read is now passwordless for both legacy QA and beta stage catalogs.

```
$ mysql -u ds_read -h blackangus

[...]

mysql> show databases;
+--------------------+
| Database           |
+--------------------+
| information_schema |
| ds_stage           |
| qa_dosomething     |
+--------------------+
3 rows in set (0.00 sec)
```

PR #790, opened 2014-02-24
Moving causes down

### LINKED ISSUE #791: 11 Facts: Move causes selection down.

blisteringherb:
Fixes #791


aaronschachter:
Is this to move the "Causes" field lower on the edit form?  Does this actually fix? 

Haven't tested this but I'd imagine you'd need to drag the field with UI and features-update the fact_page feature, so the weights are fixed in the field_instance.inc?

blisteringherb:
The last commit should fix this.

aaronschachter:
:+1:

PR #792, opened 2014-02-24
Unset user.js globally.

### LINKED ISSUE #764: Preprocess JS - User.js
Per description #724 - the login modal on a pitch page does not work unless Drupal's core User module's User.js is removed or renamed manually.

This issue is to handle this programatically, ideally through a `hook_preprocess_js` to avoid messing with core.

angaither:
Fixes #764

cc: @desmondmorris

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -4,6 +4,18 @@ include_once('dosomething_helpers.theme.inc');
 include_once('dosomething_helpers.strongarm.inc');
 
 /**
+ * Implements hook_js_alter().
+ */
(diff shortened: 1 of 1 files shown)


aaronschachter:
Question: Would this make sense to define in paraneue_dosomething's template.php?  This issue is only due to neue JS? cc @desmondmorris

desmondmorris:
@aaronschachter  Yes, we are already doing things like this from the theme layer in the paraneue base theme.  The theme layer would be the best place for this.  cc: @DFurnes 

I would not let this block for now though.  It can always be moved.

angaither:
@desmondmorris @aaronschachter @DFurnes 
We could go this route instead https://github.com/DoSomething/paraneue/pull/24

angaither:
closing, unset in https://github.com/DoSomething/paraneue/pull/24

Issue #797, opened 2014-02-24
DS build: Zendesk library


aaronschachter:
Blocked by #796 

This issue blocks #770 from completion.
1. Update build script to install the Zendesk PHP library (https://github.com/zendesk/zendesk_api_client_php) into Drupal's libraries directory.  Update `dosomething_zendesk_libraries_info` with relevant path/filename if necessary.
2. Enable `dosomething_zendesk` in the dosomething.info

Issue #770, opened 2014-02-21
ZenDesk contact form


aaronschachter:
Use ZenDesk API to create a contact form that appears on campaigns.

Should create a ZenDesk ticket with the logged in user's UID, email, name and campaign they are submitting from. 

Notes:
Official ZenDesk PHP Library: http://developer.zendesk.com/blog/2014/02/20/introducing-the-zendesk-php-api-client-library/
Examples: https://github.com/zendesk/zendesk_api_client_php

barryclark:
@aaronschachter What are you thinking for the UI for this? Can we use our modals for the form?

aaronschachter:
Yup thats what i was a-thinkin

barryclark:
Sweet, loop in @DFurnes when you're working on this too pls.

PR #798, opened 2014-02-24
dosomething_zendesk

aaronschachter:
Refs #770

New module that will handle submitting tickets to Zendesk using the new official Zendesk PHP library: https://github.com/zendesk/zendesk_api_client_php

Not enabling this module inside ds build yet, because we need to handle #797 first to load the library.

### DIFF
--- a/lib/modules/dosomething/dosomething_zendesk/dosomething_zendesk.module
+++ b/lib/modules/dosomething/dosomething_zendesk/dosomething_zendesk.module
@@ -3,8 +3,9 @@
  * @file
  * Code for the DoSomething Zendesk module.
  */
-use Zendesk\API\Client as ZendeskAPI;
 
+// Namespace to allow use of ZendeskAPI class.
+use Zendesk\API\Client as ZendeskAPI;
 
 /**
  * Implements hook_menu().
@@ -44,7 +45,6 @@ function dosomething_zendesk_libraries_info() {
     ),
     'version' => 1
   );
-
   return $libraries;
 }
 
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_zendesk/dosomething_zendesk.module
```diff
@@ -0,0 +1,148 @@
...
+function dosomething_zendesk_get_client() {
+  $library = libraries_load('zendesk');
+  if (empty($library['loaded'])) {
+    return FALSE;
+  }
+  $subdomain = variable_get('dosomething_zendesk_subdomain');
+  $username = variable_get('dosomething_zendesk_username');
+  $token = variable_get('dosomething_zendesk_token');
+  $client = new ZendeskAPI($subdomain, $username);
+  $client->setAuth('token', $token);
```
angaither, line 55: I would imagine you'll get errors if these vars aren't set. 
Pass in a default value or check if isset?

aaronschachter, line 55: `variable_get` does pass through NULL as default, so there actually won't be errors returned by this function if the variables don't exist.

Errors would be returned when trying to make a request against a $client object with bad variables, which will be caught here: https://github.com/aaronschachter/dosomething/blob/zendesk/lib/modules/dosomething/dosomething_zendesk/dosomething_zendesk.module#L116

angaither:
:+1:

PR #800, opened 2014-02-24
Set image & cause tags in image search to 'and' not 'or'

### LINKED ISSUE #729: Enhancements to Filtering for Image Search
- make sure filter criteria is "and" not "or" when multiple tags are selected

cc: @barryclark

angaither:
Fixes #729


aaronschachter:
Is this missing code to make the actual change?  Seems like the only addition has been translatables.

Am I also allowed to be super OCD and request we change the menu title to "Search images" instead of "Search Images" -- to keep it consistent with all of the Drupal menu capitalizations? :\

angaither:
@aaronschachter updated!

aaronschachter:
:+1:

PR #802, opened 2014-02-24
Print do it copy.

### LINKED ISSUE #707: action page (do it) – last content section missing copy
The last section of the "do it" section of the action page is not displaying any copy. It is simply printing a single letter.

angaither:
Fixes #707


aaronschachter:
Good catch :+1:

PR #809, opened 2014-02-24
Administrator role

aaronschachter:
Makes the "administrator" role the "administrator" role forreals.

Per help text on "admin/config/people/accounts":
- This role will be automatically assigned new permissions whenever a module is enabled. Changing this setting will not affect existing permissions.


angaither:
:+1: go the extra mile. put this one to bed. total quality.

PR #817, opened 2014-02-25
Adds html5shiv for IE8.

### LINKED ISSUE #812: HTML5Shiv in header for IE8
Ye 'ol Internet Explorer Eight requires [html5shiv](https://github.com/aFarkas/html5shiv) to understand such modern niceties as `<main>` and `<header>` and `<nav>`. We shalt provide it in the `<head>` such that users of aforementioned venerable browser may be enlightened by the semantic goodness of these tags.

DFurnes:
Fixes #812.


mmwtsn:
:sailboat:

PR #821, opened 2014-02-25
Smart settings and cap deploy to pc staging server

desmondmorris:
This merge introduces a deployment task that ships code to a staging VM present in our PC environment.  The DNS is still resolving, but this will eventually point to staging.beta.dosomething.org.  You will need ssh keys on that machine to deploy and we will most likely set this deployment as a job that can be trigger in jenkins.

This also introduces a "smart" settings file that needs to be tested a bit more before replacing the existing settings files in our config folder.  Currently this file looks for environment variables to set the db and hostname values in settings.php.  It also looks for a environment specific settings file based on the current environment set.  This defaults to local.  This is only being used in the deployment currently.  Once this gets more local testing, I propose replacing settings.php and settings.local with this file.

Here is the list of currently support env vars:

```
DS_DB_MASTER_NAME
DS_DB_MASTER_USER
DS_DB_MASTER_PASS
DS_DB_MASTER_HOST
DS_DB_MASTER_PORT
DS_DB_MASTER_DRIVER
DS_DB_MASTER_PREFIX
DS_HOSTNAME
DS_INSECURE_PORT
DS_SECURE_PORT
```

Varnish salt & drupal salt should be included here eventually as well.  We can also use these to set the redis and storage paths.  By default, this should work without envvars in the local environment

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -10,7 +10,7 @@ set :repo_url, 'git@github.com:DoSomething/dosomething.git'
 set :branch, "dev"
 
 # Default deploy_to directory is /var/www/my_app
-set :deploy_to, '/home/ubuntu/app'
+# set :deploy_to, '/var/www/my_app'
 
 # Default value for :scm is :git
 set :scm, :git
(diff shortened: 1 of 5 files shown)


aaronschachter:
This seems ok to me

desmondmorris:
gracias. your golden @aaronschachter

PR #825, opened 2014-02-25
Condintionally Prints Scholarship Variable

### LINKED ISSUE #824: pitch (header) — scholarship always prints to the page
Regardless of whether or not a scholarship is specified on a campaign the scholarship highlight and arrow print.

mmwtsn:
On the pitch page, only print the scholarship variable and hand-drawn arrow if a scholarship was set on the campaign.

Fixes #824


aaronschachter:
there it is :+1:

PR #834, opened 2014-02-25
action (prove it) – gallery updates

### LINKED ISSUE #828: action (prove it) – gallery missing fallback elements
If no image is uploaded to the report back gallery image section of the campaign a default fallback image should load.

### LINKED ISSUE #829: action (prove it) – update pagination buttons
An updated icon font glyph set has been pushed to Neue. It should be used in place of the current greater than and less than signs used to paginate gallery images.

mmwtsn:
1. If the campaign does not have any report back images associated with it, a single fallback image without pagination is displayed.
   default
2. Pagination icons are now pulled in using our icon font.
   pagination
3. Markup cleaned up.

Closes #828 
Closes #829


DFurnes:
We should probably leave some `TODO` comments next to any placeholder images we're using so that they don't get lost. Otherwise all good. +1

DFurnes:
Also, looked like there was a bunch of re-indenting in the template? Just worth a double-check to make sure that was the case.

mmwtsn:
`@TODO` flags added! The indentation issue was simply Git having trouble with the diff. That commit removed a wrapping `div` and indented all previously nested code.

PR #843, opened 2014-02-25
Only staff members can see term pages.

### LINKED ISSUE #841: View term pages permissions
Currently all taxonomy term pages are visible to anon users: e.g. http://stage.dosomething.org/cause/poverty

I'd imagine we don't want anonymous users or authenticated users to view these pages, as it hasn't been in scope to theme or handle these term pages. This issue is to deny access to any non-staff users viewing a taxonomy term page.

angaither:
Fixes #841

### DIFF
--- a/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
+++ b/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
@@ -34,7 +34,7 @@ function dosomething_taxonomy_get_sponsors($sponsor_field_wrapper) {
  */
 function dosomething_taxonomy_taxonomy_term_view($term, $view_mode, $langcode) {
   // Don't grant access to term pages for non-staffers.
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
```diff
@@ -27,4 +27,14 @@ function dosomething_taxonomy_get_sponsors($sponsor_field_wrapper) {
...
-}
\ No newline at end of file
+}
+
+/**
+ * Implements hook_taxonomy_term_view().
+ */
+function dosomething_taxonomy_taxonomy_term_view($term, $view_mode, $langcode) {
+  // Don't grant access to term pages for non-staffers.
+  if (!dosomething_user_is_staff()) {
```
aaronschachter: Let's also add a check for `module_exists('dosomething_user')` to be safe here.  Then good to go.

PR #845, opened 2014-02-25
NPM Warnings!

mmwtsn:
Resolves two NPM warnings encountered when `ds grunt` is executed.

Closes DoSomething/paraneue#28


desmondmorris:
:+1:

Issue #768, opened 2014-02-21
Campaigns Template - Sponsor Fields + Modal


mikefantini:
Need new fields on campaigns node. These should live within the Sponsors and Partners Group.

Fields:
- partner name (not sponsor name because we may give this to a non-sponsor)
- body text
- partner photo
- partner video

Note:
- need the ability to add multiple per campaign
- need markdown (ability to add links and create separate paragraphs)

This will open up in a modal on the action page.

See annotated wires here: https://trello-attachments.s3.amazonaws.com/511573a78e553913770000ea/52d9ad6d1b4858440e574065/8c828a9e1d6211fd5acd421b4ce4ae34/Action_Page_-_Sponsor_Modal_v0.1.png

See mocks here: https://s3.amazonaws.com/uploads.hipchat.com/34218/237177/Xk7eqRtsusVCuKB/keri_actionpg%20mob_aero_modal.jpg

mikefantini:
cc: @barryclark

mikefantini:
@aaronschachter @barryclark put this into 0.2 since we need to get it done for March 17th.

aaronschachter:
Proposed build:
- The sponsor vocabulary becomes "partner" vocabulary, which has a Partner logo file field.
- Delete field_sponsors and field_partners multi-value fields from campaign.
- Create a new field collection, field_campaign_partners with fields:
  - Partner Term ID (entity reference - single value to Partners vocab)
  - Is Sponsor (boolean)
  - Partner Copy (Long text with Markdown)
  - Partner Image (Node ref to Image type)
  - Partner Video (Video field collection)

In the preprocess function, loop through all field_campaign_partners.
Foreach partner:
- If partner is sponsor: output the logo in the header for the "Powered by [logo]"
- If partner has Partner Copy set, output the Partner modal link and copy/image/video
- List the partner name in the campaign footer.

angaither:
Seems good, just a few questions: 

Can we store the following with the partner term:
- Partner logo
- Partner video
- Perhaps partner copy (though I can see why this may be different per campaign the partner sponsors... perhaps we can do a default living with the term and override per campaign as needed?) 

Also, since image nodes now have minimum requirements, I think we should leave the partner image as a field attached to the term.

aaronschachter:
The partner logo does live on the term, but yeah - the copy/image/video are supposed to live on the campaign node as they can change per campaign.  I believe the image/photo is something that is separate from the logo.

[...]

PR #849, opened 2014-02-25
Partners vocabulary

### LINKED ISSUE #833: Campaign: Sponsors field / taxonomy changes
Refs #768. Creating task list to help maintain sanity.  Keeping #768 open to handle preprocess logic.
- [x] Delete field_sponsors
- [x] Delete field_partners
- [x] Update dosomething_campaign feature with deletions
- [x] Delete sponsors vocabulary
- [x] Update dosomething_taxonomy feature with deleted sponsors
- [x] Create partners vocabulary with Partner logo image file field
- [x] Update dosomething_taxonomy feature added partners
- [x] Create field_partners field collection with partner tid, is_sponsor, image node, copy, field_video
- [x] Update dosomething_campaign feature with field_partners

aaronschachter:
Fixes #833

Replaces Sponsors vocabulary with Partners vocabulary, and creates a Partners field collection on the Campaign to handle the Partners/Sponsors modal.  

Within the field collection, the Partner is set as a sponsor or not.  This way, a partner term can be a sponsor on Campaign X but only a partner on Campaign Y.

The field_collection is defined within dosomething_taxonomy so it can be applied to any other content types that may be sponsored in the future, such as SMS Game or Fact Pages.

All removed fields and vocabularies have already been deleted from staging, so we should not run into any "changing field type" errors.

This PR gets us in a place to begin work on adding all of these variables into the template in #768


angaither:
:+1:  at the end of the day bottom line is we keep our core values

PR #850, opened 2014-02-25
Use a session to make nid persist after incorrect password.

### LINKED ISSUE #801: Pitch login bug
If you enter the wrong password, you are directed to user/login.

This means that once you enter the right password, you won't be signed up for the campaign, since you are no longer on the pitch page.

angaither:
Fixes #801

@desmondmorris @mshmsh5000 
IS THIS OK?
tumblr_m2e9dsu9tu1qdedm3o1_500

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -152,14 +152,16 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
  * @see dosomething_user_strongarm()
  */
 function dosomething_user_login_submit($form, &$form_state) {
-  // If nid is not present, nothing to sign up for.  Exit.
-  if (!isset($form_state['input']['nid'])) { return; }
-  $nid = $form_state['input']['nid'];
-  // Redirect to node (this is needed for user_register form).
-  $form_state['redirect'] = 'node/' . $nid;
-  // Signup global user for node $nid.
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -214,6 +216,7 @@ function _dosomething_user_add_campaign_data(&$form) {
     $obj = menu_get_object();
     // If we are viewing a campaign node:
     if (isset($obj->type) && $obj->type == 'campaign') {
+      $_SESSION['nid'] = $obj->nid;
```
aaronschachter, line 219: Reading up on using session data in D7.  Seems we may need to manually start the session for anon users: 

"Drupal 7 sometimes does not automatically start sessions for anonymous user" http://drupal.stackexchange.com/questions/4851/how-do-i-save-session-data-for-the-anonymous-users

DFurnes:
:heartbeat:

aaronschachter:
This solution has a bug: As an anonymous user, I view a campaign pitch page with nid X.  If I then navigate to user/login on my own, to simply sign in, won't I then be automatically signed up for campaign X?

mshmsh5000:
1. Session caching: worth looking at https://drupal.org/project/session_cache which has a smarter implementation.
2. Redirects: Would be much better, I think, to append `destination` param to user/login, so we can make this explicit.

PR #852, opened 2014-02-25
Fixing intro and hero image for static content

### LINKED ISSUE #836: Static content - intro_image not displaying
I have uploaded an image and included it in the intro_image field, however it never gets displayed. 

http://stage.dosomething.org/content/fantini-static-content-test
http://stage.dosomething.org/node/30/edit

blisteringherb:
- Imagecache presets are only by size so generating the images
  requires specifying the dimensions.

Fixes #836


angaither:
:+1:

PR #868, opened 2014-02-26
Added 'form action' to add campaign data funciton.

### LINKED ISSUE #801: Pitch login bug

angaither:
On password fail, will remain on current campaign page.
Fixes #801


aaronschachter:
:+1:

PR #886, opened 2014-02-27
Signup mobile commons

### LINKED ISSUE #857: New User Reg: Mobile Commons opt-in
Four things should happen when a user registers for the site NOT through signing up for a campaign (i.e. clicking Sign Up in the header) and provides a mobile number:
- [ ] Get subscribed into opt-in path 164905 on Mobile Commons
- [ ] Pass along user's first name
- [ ] Pass along user's DOB
- [ ] Pass along user's email

See #450 and #759 for related issues.

cc: @jonuy

### LINKED ISSUE #881: Campaign: Mobile Commons opt-in non-staff pick
The correct Mobile Commons opt-in path ID for non-staff pick campaigns is actually 164951.

This is a change from #450 because we had to change the opt-in path.

angaither:
Fixes #857
Fixes #881

First pass at mobile commons signups.
Added signups for new users. 
Moved code to include file dosomething_signups

Still left:
-  Move opt-in id's to admin form.
- Make an admin form

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -253,7 +253,7 @@ function dosomething_signup_user_signup($nid, $account = NULL) {
   if ($sid = dosomething_signup_insert($nid, $account->uid)) {
     $node = node_load($nid);
     // Opt-in to mobilecommons.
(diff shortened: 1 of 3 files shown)


aaronschachter:
Looks good, but technically doesn't close out #857 yet -- still need to pass first name and email as params.  Fine to merge now if you update the description to not close yet, or just add those extra params.

PR #903, opened 2014-02-28
Issue716 message broker install

### LINKED ISSUE #716: Message Broker and drush.make - how to include libraries with depenancies
The messagebroker-phplib library is needed for the message_broker_producer.module.  messagebroker-phplib is defined as a drupal library but messagebroker-phplib has additional dependancies of libraries defined in composer.json.

Are the following changes to drush.make sufficient for this addition?

```
; Message Broker PHP Library
libraries[messagebroker-phplib][download][type] = "git"
libraries[messagebroker-phplib][download][url] = "https://github.com/DoSomething/messagebroker-phplib.git"
```

@desmondmorris 
How do I get the additional composer defined dependancies defined in composer.json to be installed in /vendor as well? 

```
; GIT PROJECTS

; Message Broker Drupal 7.x Producer
projects[message_broker_producer][type] = "module"
projects[message_broker_producer][download][type] = "git"
projects[message_broker_producer][download][url] = "https://github.com/DoSomething/message_broker_producer.git"
projects[message_broker_producer][subdir] = "contrib"
```

deezone:
Fixes #716

Adds the:
- Message Broker Producer Drupal module
- related messagebroker-phplib library
  - adds messagebroker-phplib library dependency: php-amqplib

See #775 for `message_broker_producer_request('produceTransactional', $payload)` details on how the following user activities result in entries produced in the RabbitMQ "transactionals" queue:
- user_register
- user_password
- campaign_signup
- campaign_reportback

A consumer app that watches the transactional queue generates API calls to Mandrill resulting a transactional email message being sent to the user.


aaronschachter:
Will installing the package with Composer through the ds file will give you a vendor/autoload.php to include?  If so, doesn't this `hook_libraries_info` need to change: https://github.com/DoSomething/message_broker_producer/blob/master/message_broker_producer.module#L18

lib/profiles/dosomething/drupal-org.make
```diff
@@ -153,6 +153,12 @@ projects[conductor_sms][download][type] = "git"
 projects[conductor_sms][download][url] = "https://github.com/DoSomething/conductor_sms.git"
 projects[conductor_sms][subdir] = "contrib"
 
+; Message Broker Producer
+projects[message_broker_producer][type] = "module"
+projects[message_broker_producer][download][type] = "git"
+projects[message_broker_producer][url] = "https://github.com/DoSomething/message_broker_producer"
```
jonuy: Should this be `projects[message_broker_producer][download][url] =`? Just asking because it looks like that's what the conductor and conductor_sms config looks like.

aaronschachter: Good catch @jonuy

deezone:
@aaronschachter Good point but there's a work around within the library. hook_libraries_info points to MessageBroker.php (https://github.com/DoSomething/messagebroker-phplib/blob/master/MessageBroker.php) which takes care of the autoload stuff. In non-Drupal apps the autoload via composer.json points to MessageBroker/MessageBrokerObjectLibrary.php

```
"autoload": {
  "psr-0": {
    "MessageBrokerObjectLibrary": "MessageBroker/"
  }
}
```

Perhaps hacky but it works. Thoughts, suggestions?

aaronschachter:
walked thru changes with @DeeZone who explained purpose of MessageBroker.php which expects to find a vendor/autoload.php

:+1:

PR #909, opened 2014-02-28
Campaign Partials

### LINKED ISSUE #848: campaign.scss is growing unruly
Keeping all campaign-related styles in one file is creating a few headaches. This file should be broken out into partials by page section.

mmwtsn:
This bundle of commits creates separate partials for all sections of the campaign template including:
- pitch page
- action page
- confirmation page
- general layout classes

To simplify the maintenance process, all campaign partials are imported under the `campaign--wrapper` class to enforce namespacing. This import syntax breaks if any un-nested styles are passed across files. For this reason the general use `_campaign.scss` partial receives its own wrapping class inside that file.

This also corrects the ASCII art as it incorrectly documented Paraneue DS as Paraneue.

**NOTE:** Due to the nature of these commits, the visual diffs provided by Git below are useless.

Resolves #848


DFurnes:
SURE LETS DO IT. :godmode:

PR #910, opened 2014-02-28
Added fieldgroup permissions

### LINKED ISSUE #842: Admin perms: Field collection and field group
Because field_collection and field_group have been enabled since we've made the change to the admin role variable in #809, the permission is not granted to the admin role.  This unfortunately will be the case for all modules that were enabled previously.

This issue is to grant the "administrator" role the ability to administer field collections and field groups in dosomething_user.

angaither:
Fixes #842


lib/modules/dosomething/dosomething_user/dosomething_user.views_default.inc
```diff
@@ -217,6 +217,32 @@ function dosomething_user_views_default_views() {
   $handler->display->display_options['menu']['weight'] = '0';
   $handler->display->display_options['menu']['context'] = 0;
   $handler->display->display_options['menu']['context_only_inline'] = 0;
+  $translatables['content_search'] = array(
```
aaronschachter: ?? what is this?

angaither:
@aaronschachter removed that file.

aaronschachter:
:+1:

PR #911, opened 2014-02-28
Authy-auth-auth.

### LINKED ISSUE #420: Implement neue login/registration workflow

DFurnes:
Closes #420. :leaves: 

Extracted the view layer from the `dosomething_user` module into the theme. Yay, separation of concerns! :100:

Honestly, there isn't anything all too complicated in here... it's a rather underwhelming culmination of three days of deep `dpm()` diving in various different theme hooks. Luckily even the [guy who coded the Form API](http://john.albin.net/drupal/arrays-of-doom) thinks it was the worst idea ever, so _that_ makes me feel great.


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -100,9 +100,8 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
   switch ($form_id) {
     case 'user_login_block':
     case 'user_login':
-      // Form specific helper text.
-      $form['name']['#title'] = 'Email or Cell Number';
-      $form['pass']['#attributes']['placeholder'] = t('[password]');
+      // After build form changes.
+      $form['#after_build'][] = 'paraneue_dosomething_login_after_build';
```
aaronschachter: If we're going to move these functions into the theme's template.php, we need a wrapper around this guy to make sure that `paraneue_dosomething` is the current theme.

DFurnes: +1 agreed. I'd argue the best way to handle this is to have the function be a more generic `after_build` that any supporting theme can hook into if it wants to; not sure if this is possible.

aaronschachter: Right.. yeah then we should maybe have it live in `dosomething_user`.  I'll stop by to chat

DFurnes:
Also connected to DoSomething/paraneue#36.

DFurnes:
Oh man, I just tried to rebase to bring in line with master... :skull: Anyways, all these things should be merged into master so I don't think its an issue. Makes diffing a bitch though. Fuck.

aaronschachter:
FTW :+1:

DFurnes:
YOLO

PR #913, opened 2014-02-28
Added imagemagick rule to process images.

### LINKED ISSUE #912: Images: Filter Needed on Campaign Hero Images
We need to attach the image filter we want to the actual image itself because certain devices may not render the CSS in the way we want and therefore won't display the filter, making the white text on top of the photos unreadable.

@angaither is going to investigate ImageMagick to see if this will resolve the issue.

Best case scenario: ImageMagick is able to apply the filter we need  and is easy enough to implement for the 0.2 release.

Worst case scenario: We adjust the Image Node to have fields for:
- landscape crop, no filter
- landscape crop, filter
- square crop, no filter
- square crop, filter

angaither:
Fixes #912

This only creates a processed crop for the hero image as a proof-of-concept.


DFurnes:
This is super-fantastic. +1000

aaronschachter:
Code seems fine ... but upon pushing to stage, won't we get "function not found" errors without a full ds build to dl the module?

angaither:
@aaronschachter I think it should be fine if we run drush remake?
cc// @desmondmorris @mshmsh5000

aaronschachter:
Just chatted with @mshmsh5000 -- we deploy to staging by actually running `ds build`.  See https://github.com/DoSomething/dosomething/blob/dev/config/deploy.rb#L44.  Should be to good to go, if things break locally we just need to re-run the build.

Did you test the build with the new addition, just to double check that all is good?

angaither:
Nope, I'll test with a build before merging.

PR #922, opened 2014-02-28
Update Admin Perms

### LINKED ISSUE #915: Admin Perms
Export those permissions that were not applied to the admin role yet.

angaither:
Fixes #915


lib/modules/dosomething/dosomething_user/dosomething_user.features.user_permission.inc
```diff
@@ -59,6 +59,15 @@ function dosomething_user_user_default_permissions() {
     'module' => 'node',
   );
 
+  // Exported permission: 'access devel information'.
```
aaronschachter: Let's remove Devel, as this now becomes a dependency for dosomething_user to run, and we don't want Devel enabled on production.

Because it is disabled by default, when you enable the module, the role is automatically granted to an administrator.

PR #935, opened 2014-03-03
Remove corrupt field_partners field

### LINKED ISSUE #927: Campaign field_partners
`field_partners` was changed from a taxonomy term reference to a field collection, and it's causing this error:

```
EntityMetadataWrapperException: Invalid data value given. Be sure it matches the required data type and format. in EntityDrupalWrapper->set() (line 736 of /var/www/beta.dosomething.org/releases/20140228215948/html/profiles/dosomething/modules/contrib/entity/includes/entity.wrapper.inc).
```

We had one record in the old `field_data_field_partners` table before the field type change, which I had recreated as a new Partners Field Collection value after deleting from the old table.  Currently, the data in the `field_data_field_partners` looks weird.  Inspecting a devel/php on a node witih no partners values filled out in the form shows that there's an integer value set inside the `field_partners` values array.  This is why the previous empty and set checks didn't work.

Investigating if dropping the field and renaming the `field_partners` definition + features-revert will fix.

aaronschachter:
Fixes #927
- DS Modules: drops evil `field_partners` field collection field from features and database
- DS Build: Grabs the latest revision from Field Collection's dev 7.x-1.x branch to fix issue with cleanly deleting field_collection fields.

What's going on is that the field_partners field collection field is adding default  field_collection_item id's regardless of whether or not you add any in the node edit form -- which is causing the confusion.  You haven't added any partners for nid x, but a field_collection_item record with item_id y exists for field_partners and nid x.

Seems it's because of this issue: https://drupal.org/node/1239946 -- when a default value is set on any of the child fields.  field_partners has a field called is_sponored which is a boolean, so it automatically has a default value of 0, and also the video_id has a default provider value.

This PR doesn't fix that. This PR is just to delete the field and its corrupt data, which there wasn't any real data to begin with.

Going to open up a separate PR for getting the Partners field collection to work again re: the the default value.


aaronschachter:
Requires `drush -y updb && drush -y fra` upon deployment to fix.

aaronschachter:
Closing this, not ready for merge just yet.  The solution may be to simply implement `hook_field_collection_is_empty_alter` per https://drupal.org/node/1614578

PR #952, opened 2014-03-03
Added function to check if user is under 13.

### LINKED ISSUE #932: Site Registration - Under 13 should not be added to MailChimp nor MobileCommons
Users who are under 13 years old _should be_ allowed to register for the site and we should create user accounts for them and store their information in our database.

However, we should not pass their information along to either MailChimp or MobileCommons.

I.e. We should not automatically create accounts for under 13 users in either MailChimp or MobileCommons. Even if it is a special staff pick campaign, we should still not create accounts for under 13 users in either MailChimp or MobileCommons.

See #931  #857 and #892 for related issues.

angaither:
This is run on new user registrations to determine.
if we are allowed to sign him/her up for external ds communications.

Fixes #932

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -173,8 +173,11 @@ function dosomething_user_new_user($form, &$form_state) {
  * @return bool
  *  Returns true if user is under 13, else false.
  */
-function dosomething_user_is_under_thirteen() {
-  $birthday = dosomething_user_get_birthdate();
+function dosomething_user_is_under_thirteen($user = NULL) {
+  if ($user == NULL) {
+    global $user;
+  }
+  $birthday = dosomething_user_get_birthdate(NULL, $user);
   // Get the date 13 years ago today.
   $date_cutoff = strtotime('-13 years');
   if ($date_cutoff < $birthday) {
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -150,17 +148,40 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
...
+/**
+ * Checks that user is over 13.
+ *
+ * This is needed to make sure a new user is old enough
+ * to get any external messages from us.
+ *
+ * @return bool
+ *  Returns true if user is under 13, else false.
+ */
+function dosomething_user_is_under_thirteen() {
```
aaronschachter: Let's pass an ($account = NULL) param, which defaults to global $user if it doesn't exist, and would keep it consistent with the other dosomething_user functions

That way we can write tests by passing in different users.

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -150,17 +148,43 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
...
+    $params = array(
+      'email' => $account->mail,
+      'uid' => $account->uid,
+      'first_name' => dosomething_user_get_field('field_first_name', $account),
+    );
+    dosomething_user_send_message_request('user_register', $params);
+    // Sign user up for mobile commons
+    if (module_exists('dosomething_signup')) {
+      $opt_in = variable_get('dosomething_mobilecommons_general');
+      dosomething_signup_mobilecommons_opt_in($account, $opt_in);
```
blisteringherb, line 163: Does this function require a third param, $title?

https://github.com/DoSomething/dosomething/blob/5fbb4ff5b5a23e61e5c4ec2970e9dfba057ee06c/lib/modules/dosomething/dosomething_signup/includes/dosomething_signup.mobilecommons.inc#L11

angaither, line 163: good call, title should default to null, created #962 for that

blisteringherb:
:+1:

PR #953, opened 2014-03-03
Partners field collection fix

### LINKED ISSUE #936: Partners field_collection default value bug
The `field_partners` field collection was writing default values regardless of whether or not any data was entered, hence removing it (#935)

Known issue for adding default values when a field collection's child field has a default value set. https://drupal.org/comment/5375608

Need to investigate if this was case, because it could have been either/or:
- field_is_sponsor (would default to 0)
- video_id (this is a grandchild field, a child of video_id, which is a child of field_partners)

aaronschachter:
- Implements `dosomething_taxonomy_field_collection_is_empty_alter` to fix #936
- Places Partner/Sponsor info fields into a field group to improve node edit form


blisteringherb:
:+1:

PR #964, opened 2014-03-03
Add under 13 bool

### LINKED ISSUE #959: Site Registration - Field for users under 13
We need to create a field for users under 13 who register for the site.

This will a. allow us to easily see which users are under 13 and b. set us up in the future to automatically subscribe those users to MailChimp/Mobile Commons on their 13th birthday.

Relates to #932 

cc: @marahml @Bladt 
- Users should not be able to edit or view this field

angaither:
Added hidden field to user profile.
Set to 1 if user is under 13

Fixes #959


aaronschachter:
:+1:

PR #965, opened 2014-03-03
Fixin small bugs

aaronschachter:
baby bugs


angaither:
:+1: value added. best practice.

PR #966, opened 2014-03-03
Set mobile-commons title opt in to null

### LINKED ISSUE #962: Mobile Commons opt-in function
Title should default to NULL

angaither:
Fixes #962


aaronschachter:
Should also only send through campaign[title] if isset($title)

jonuy:
fwiw I also do a check on the Mobile Commons side to handle cases where `campaign_name` is null.

PR #975, opened 2014-03-04
Updated get_themed_image function.

### LINKED ISSUE #807: Remove width and height attributes in image tag helper
We should remove hard-coded width and height attributes from `dosomething_image_get_themed_image` so that we can size those motherfuckers using some CSS.

angaither:
Now it returns magic.
Added optional title, attributes.
Returns image tag with no height/width set.
Fixes #807

### DIFF
--- a/lib/modules/dosomething/dosomething_image/dosomething_image.module
+++ b/lib/modules/dosomething/dosomething_image/dosomething_image.module
@@ -23,23 +23,30 @@ function dosomething_image_form_image_node_form_alter(&$form, &$form_state, $for
  *  The image field to pull from the desired image ratio (vert, horiz, sq).
  * @param string $style
  *  The image style preset/size to return the image.
+ * @param string $title
+ *  Optional - image title
+ * @param array $attributes
+ *  Optional - key/value paired array of attributes to add to the image.
  *
  * @return array
- *  A themed image.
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_image/dosomething_image.module
```diff
@@ -23,23 +23,30 @@ function dosomething_image_form_image_node_form_alter(&$form, &$form_state, $for
...
  * @param string $style
  *  The image style preset/size to return the image.
+ * @param string $title
+ *  Optional - image title
+ * @param array $attributes
+ *  Optional - key/value paired array of attributes to add to the image.
  *
  * @return array
- *  A themed image.
+ *  A themed image url with no set height/width.
```
aaronschachter: Still returns an image right, not a url?

aaronschachter:
This looks ok (except for comment) - but will this change break things in other modules that reference this function?

DFurnes:
Can you switch the `title` parameter to specify `alt` text instead?

angaither:
@DFurnes should we do both?

mmwtsn:
The title attribute just provides a tooltip hover state. I don't think it's necessary for images.

DFurnes:
Yep, +1. Thanks Andrea. :100:

PR #977, opened 2014-03-04
Changes ds cli shell to bash instead of sh.  This will make it play nice...

desmondmorris:
Changes ds cli shell to bash instead of sh.  This will make it play nicer on diff flavors of linux (specifically  unbuntu)


mshmsh5000:
+1 Thank god. This tripped me up with the function declarations before.

Issue #894, opened 2014-02-27
SMS in-campaign transition flows


jonuy:
This is to implement the New World SMS UX plan where SMS users are transitioned from one Mobile Commons campaign to another as they progress through the activities of a DS campaign.
- [x] Create `dosomething_sms` module out of old world SMS code
- [x] Create endpoint to handle transition from starting MC campaign to in-progress MC campaign.
- [x] Create endpoint to handle SMS report backs and transition from in-progress MC campaign to completed MC campaign.

PR #981, opened 2014-03-04
Initial pass at dosomething_sms module

jonuy:
For Issue #894

The `dosomething_sms` module here only takes pieces from the old world `dosomething_sms` and old world `sms_flow` modules that are currently needed for the Comeback Clothes campaign.

Highlights:
- All things related to `dosomething_sms_start_campaign_transition` are for the new world SMS UX plan where users are transitioned from one Mobile Commons campaign to another.
- `dosomething_sms_game` will be used for tracking a user's progress through SMS tips. Not yet included because I have no content. This was previously handled by the `sms_flow_game` table.
- The module has `dosomething_sms_mobile_commons_opt_in()` and `dosomething_sms_mobile_commons_opt_out()` methods. Not sure if you want to use these or continue to do like what was done in the `dosomething_signup` module. And/or if this implementation should include checks on whether necessary modules and libraries exist.
- The implementation here depends on the 7.x-1.1 version of the `mobilecommons` module.

cc: @aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_sms/dosomething_sms.info
+++ b/lib/modules/dosomething/dosomething_sms/dosomething_sms.info
@@ -1,0 +1,6 @@
+name = DoSomething SMS
+description = Provides SMS integration for DoSomething.org
+core = 7.x
+dependencies[] = conductor_sms
+package = DoSomething
+configure = admin/config/dosomething/dosomething_sms
--- a/lib/modules/dosomething/dosomething_sms/dosomething_sms.install
+++ b/lib/modules/dosomething/dosomething_sms/dosomething_sms.install
@@ -40,55 +40,5 @@ function dosomething_sms_schema() {
     ),
   );
 
-  $schema['dosomething_sms_invite_records'] = array(
-    'description' => 'Stores alphas and betas who go through the SMS flow.',
-    'primary key' => array('sid'),
-    'fields' => array(
-      'sid' => array(
-        'description' => 'The id of the record.',
-        'type' => 'serial',
-        'unsigned' => true,
-        'not null' => true,
-      ),
-      'alpha' => array(
-        'description' => 'The alpha\'s phone number.',
-        'type' => 'varchar',
-        'length' => 255,
-        'not null' => true,
-      ),
-      'beta' => array(
-        'description' => 'The beta\'s phone number.',
-        'type' => 'varchar',
-        'length' => 255,
-        'not null' => true,
-      ),
-      'type' => array(
-        'description' => 'The type of object that the beta has been invited to.',
-        'type' => 'varchar',
-        'length' => 255,
-        'not null' => true
-      ),
-      'nid' => array(
-        'description' => 'The ID of the page / whatever that the person was invited to.',
-        'type' => 'int',
-        'unsigned' => TRUE,
-        'not null' => TRUE
-      ),
-      'accepted_invite' => array(
-        'description' => 'Whether or not the user accepted the invitation.',
-        'type' => 'int',
-        'unsigned' => TRUE,
-        'default' => 0,
-        'not null' => TRUE,
-      ),
-      'timestamp' => array(
-        'description' => 'The time that it all began...',
-        'type' => 'int',
-        'unsigned' => TRUE,
-        'not null' => TRUE
-      )
-    ),
-  );
-
   return $schema;
 }
--- a/lib/modules/dosomething/dosomething_sms/dosomething_sms.install
+++ b/lib/modules/dosomething/dosomething_sms/dosomething_sms.install
@@ -1,7 +1,7 @@
 <?php
 /**
  * @file
- * Install/Update/Uninstall functions for dosomething_sms module
+ * Installation and schema hooks for dosomething_sms.module.
  */
 
 /**
--- a/lib/modules/dosomething/dosomething_sms/dosomething_sms.install
+++ b/lib/modules/dosomething/dosomething_sms/dosomething_sms.install
@@ -1,0 +1,94 @@
+<?php
+/**
+ * @file
+ * Install/Update/Uninstall functions for dosomething_sms module
+ */
+
+/**
+ * Define the table schemas used for this module.
+ */
+function dosomething_sms_schema() {
+  $schema = array();
+
+  $schema['dosomething_sms_game'] = array(
+    'description' => 'Stores details to SMS games relative to a user / unique phone number.',
+    'fields' => array(
+      'game_id' => array(
+        'description' => 'Unique identifier for the game/quiz.',
+        'type' => 'int',
+        'unsigned' => TRUE,
+        'not null' => TRUE
+      ),
+      'phone' => array(
+        'description' => 'Phone number',
+        'type' => 'varchar',
+        'length' => 32,
+        'not null' => true
+      ),
+      'started_paths' => array(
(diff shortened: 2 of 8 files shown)


lib/modules/dosomething/dosomething_sms/dosomething_sms.module
```diff
@@ -0,0 +1,384 @@
...
+      $valid = TRUE;
+    }
+  }
+  return $valid;
+}
+
+/**
+ * Implements hook_menu().
+ */
+function dosomething_sms_menu() {
```
deezone: Drupal convention puts hook_menu towards the top of the .module file. At least put it before the callback functions. It makes it easier to understand the flow.

lib/modules/dosomething/dosomething_sms/dosomething_sms.module
```diff
@@ -0,0 +1,384 @@
...
+    'carrier' => $carrier,
+    'sender' => $sender,
+  );
+
+  // Pass message on to Conductor to handle and generate a response
+  $output = '';
+  conductor_sms_response($output, $context);
+
+  // If an output is given, reply in the xml format expected from Mobile Commons
+  if (!empty($output)) {
```
deezone: Any reason you're working in XML rather than JSON?

deezone: Looks like a good candidate for Message Broker... perhaps, someday?

lib/modules/dosomething/dosomething_sms/dosomething_sms.install
```diff
@@ -0,0 +1,44 @@
+<?php
+/**
+ * @file
+ * Install/Update/Uninstall functions for dosomething_sms module
```
deezone: Where are the install / uninstall functions? I guess this is for DS consumption only perhaps not needed once everything is setup?

lib/modules/dosomething/dosomething_sms/dosomething_sms.module
```diff
@@ -0,0 +1,384 @@
...
+    WHERE
+      phone = '" . $phone . "'
+      AND game_id = '" . $game_id . "'
+    LIMIT 1
+  ");
+
+  $record = current($select ->fetchAll());
+
+  if ($record && !empty($record->answers)) {
+    $return = json_decode($record->answers, TRUE);
```
deezone, line 292: JSON and XML?

lib/modules/dosomething/dosomething_sms/dosomething_sms.info
```diff
@@ -0,0 +1,6 @@
+name = DoSomething SMS
+description = Provides SMS integration for DoSomething.org
+core = 7.x
+dependencies[] = conductor_sms
```
aaronschachter, line 4: This is no longer needed, right?

lib/modules/dosomething/dosomething_sms/dosomething_sms.module
```diff
@@ -0,0 +1,384 @@
...
+  $context = array(
+    'message' => $message,
+    'keyword' => $keyword,
+    'carrier' => $carrier,
+    'sender' => $sender,
+  );
+
+  // Pass message on to Conductor to handle and generate a response
+  $output = '';
+  conductor_sms_response($output, $context);
```
aaronschachter, line 152: Oh.. hmm. Do you need all the conductor stuff in here?  Maybe let's discuss

lib/modules/dosomething/dosomething_sms/dosomething_sms.module
```diff
@@ -0,0 +1,384 @@
...
+ *
+ * @return array
+ *   The paths a user has been opted-in to.
+ */
+function dosomething_sms_game_get_paths($phone, $game_id) {
+  $phone = preg_replace('#[^0-9]#i', '', $phone);
+  $phone = substr($phone, -10);
+
+  $select = db_query("
+    SELECT started_paths
```
angaither: [dynamic query me!](https://drupal.org/node/310075)

lib/modules/dosomething/dosomething_sms/dosomething_sms.module
```diff
@@ -0,0 +1,384 @@
...
+ *
+ * @return
+ *   Associative array of the user's answers to the given game. Otherwise, NULL.
+ */
+function dosomething_sms_game_get_answers($phone, $game_id) {
+  $phone = preg_replace('#[^0-9]#i', '', $phone);
+  $phone = substr($phone, -10);
+
+  $select = db_query("
+    SELECT answers
```
angaither: dynamic me too!

lib/modules/dosomething/dosomething_sms/dosomething_sms.module
jonuy: Mobile Commons expects an XML response. At least that's all they supported before. This may have changed since. And yea, there are certain things that I think would make sense for the Message Broker to handle - like report backs because those are sometimes time consuming and are at risk of timeouts.

lib/modules/dosomething/dosomething_sms/dosomething_sms.install
jonuy: No install or uninstall functions I guess. I'ma change the file comment to match the other dosomething modules.

lib/modules/dosomething/dosomething_sms/dosomething_sms.module
jonuy, line 292: XML was just for sending responses back to Mobile Commons. JSON is used here as a serialized way of storing data in the table. @DeeZone suggestions for a different/preferred way of doing this?

lib/modules/dosomething/dosomething_sms/dosomething_sms.info
jonuy, line 4: @aaronschachter The `conductor_sms` module is still needed. Particularly the `conductor_sms_response()` method does the heavy lifting for taking a request from Mobile Commons, figuring out what ConductorWorkflow should handle it, and handling any responses we'd wanna send back.

We _could_ make some changes so that this dependency is no longer needed, but... we might just wanna put that effort behind eventually removing all of this altogether from the Drupal app?

lib/modules/dosomething/dosomething_sms/dosomething_sms.module
jonuy, line 152: Continuing off the other comment on conductor_sms, `conductor_sms_response` takes care of figuring out what workflow a request should go into, if someone's already in the middle of one, the response we should send back, etc.

lib/modules/dosomething/dosomething_sms/dosomething_sms.module
deezone: Looking at the Mobile Commons API docs it appears they support both XML and JSON. There are many reasons to switch if you're facing a restructure / rewrite.

lib/modules/dosomething/dosomething_sms/dosomething_sms.info
aaronschachter, line 4: Sadness, re: `conductor_sms`.  If it was only a few functions we needed, it'd be worth moving into this module but if it's a ton of tables and code, we can keep it is.  However you'd need to make the alterations to the make file and dosomething.info in order to download the conductor_sms into the `html/profiles/dosomething/modules/contrib` directory.

Issue #934, opened 2014-03-03
Image Styles features-revert error


aaronschachter:
`drush -y fra` or `drush -y fe [stuff]` results in this error: 

```
array_merge(): Argument #1 is not an array image.module:586 
```

which seems related to image styles -- https://drupal.org/node/1973278

angaither:
Finally saw this error on http://dev.dosomething.org:8888/admin/config/media/image-styles
It's because two modules were not enabled locally, try `drush en imagecache_actions imagecache_canvasactions -y`

aaronschachter:
Ah, good catch. Are they currently enabled on staging?  

Question to the group: Perhaps best practice to catch new modules should be to `dosomething.info` accordingly, and use that profile has the guideline for everything that should be enabled?

angaither:
Yes, just enabled them on staging. 
Seems like yes, all modules in ds.info should be enabled, maybe should this happen on ds build?

aaronschachter:
That sounds like a good way to make sure of it. 

I'm not sure how we'd actually go about checking that profile's info file (without parsing the file and running drush php-eval for (!if module_enabled($name) ) then drush enable module....

@mshmsh5000 @desmondmorris -- any thoughts / concerns?  I suppose one risk might be enabling a module automatically which needs auth variables in order to fully function.  and making sure that those modules handle the lack of auth variables gracefully.

angaither:
Fixed after modules were enabled on stage.

PR #984, opened 2014-03-04
Enable modules in dosomething.info on ds build.

angaither:
On `ds build` this will check the `dosomething.info` file and enable any modules that are not already enabled.

If we no longer need `securepages` module we should remove that from the info file before this is merged in.

Refs #934

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -156,6 +156,10 @@ function build {
   echo 'Updating database...'
   drush updb -y
 
+  # Enable modules in info file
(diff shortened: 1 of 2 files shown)


aaronschachter:
This looks great -- smart to have an auto-enabler, and will make us better about knowing which modules are turned on.  Looking at you `locale` module. 

Discussed with @angaither -- we can always remove from the `ds build` script if we run into issues.

PR #986, opened 2014-03-04
Mobilecommons update

### LINKED ISSUE #882: Mobilecommons module update
- [x] Create a new 7.x-1.1 release from dev branch to make use of `mobilecommons_request` method
- [x] Update drupal-org.make file
- [x] Update mobilecommons opt-in methods accordingly

aaronschachter:
Fixes #882

If you're testing any campaign signup stuff, run a `ds build` to grab the latest mobilecommons contrib module.  Or download and replace in your relevant local html/profiles/dosomething/modules/contrib/mobilecommons.

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/includes/dosomething_signup.mobilecommons.inc
+++ b/lib/modules/dosomething/dosomething_signup/includes/dosomething_signup.mobilecommons.inc
@@ -54,11 +54,15 @@ function dosomething_signup_get_mobilecommons_vars($account, $lid, $title = NULL
   $args = array(
     'opt_in_path[]' => $lid,
     'person[phone]' => $mobile,
-    // Expected format is YYYY-MM-DD.
-    'person[date_of_birth]' => dosomething_user_get_birthdate('Y-m-d', $account),
-    'person[first_name]' => dosomething_user_get_field('field_first_name', $account),
     'person[email]' => $account->mail,
   );
+  if ($first_name = dosomething_user_get_field('field_first_name', $account)) {
+    $args['person[first_name]'] = $first_name;
+  }
+  // Expected DOB format is YYYY-MM-DD.
+  if ($dob = dosomething_user_get_birthdate('Y-m-d', $account)) {
+    $args['person[date_of_birth]'] = $dob;
+  }
   if (isset($title)) {
     $args['person[campaign_name]'] = $title;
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_signup/includes/dosomething_signup.mobilecommons.inc
```diff
@@ -5,45 +5,62 @@
...
+
+  $args = array(
+    'opt_in_path[]' => $lid,
+    'person[phone]' => $mobile,
+    // Expected format is YYYY-MM-DD.
+    'person[date_of_birth]' => dosomething_user_get_birthdate('Y-m-d', $account),
+    'person[first_name]' => dosomething_user_get_field('field_first_name', $account),
+    'person[email]' => $account->mail,
+  );
+  if (isset($title)) {
```
angaither, line 66: if we are testing if this title isset before sending it, we should also test the first name & birthday as those aren't required fields on registration.

aaronschachter:
@angaither updated!

lib/modules/dosomething/dosomething_signup/includes/dosomething_signup.mobilecommons.inc
```diff
@@ -5,45 +5,66 @@
...
+ * @param string $title
+ *   Optional -- The campaign title the user has signed up for.
+ *
+ * @return mixed
+ *   Array of arguments to pass to a Mobilecommons request, or FALSE if no mobile.
+ */
+function dosomething_signup_get_mobilecommons_vars($account, $lid, $title = NULL) {
+   // If user doesn't have mobile number, exit function.
+  $mobile = dosomething_user_get_mobile($account);
+  if (!$mobile) {
```
angaither: hmm, one more thing. if this func is `get_mobilecommon_vars` shouldn't this check happen outside of the function, it should be only getting the vars

aaronschachter: Hmmm, I can see an argument for either... but because you always have to pass in a mobile # to mobilecommons, I think it could have its place here as well.  I don't mind making the change.

aaronschachter:
@angaither updated, updated!

angaither:
:+1:

PR #989, opened 2014-03-04
User Partners/Sponsors field

### LINKED ISSUE #772: User: Partner term field
- [x] Add the existing field_partner onto the User, to associate a user with a partner term(s).
- [x] Only staff can view and edit the User's field_partner values.

This allows us to track activity for sponsors, e.g. 
- Did they even login and look at the action page? 
- Did they reportback?

As far as giving Sponsor/Partners access to campaigns before they are published, we'll use staging and publish a campaign there.

_Original Description_

Create "sponsor" and "press" roles, with same permissions as generic member role. 
When a sponsor like H&M wants to view their site that's behind the gate, we can create them a login.

aaronschachter:
Fixes #772 by adding the existing `field_partner` Term Reference multi-value field to the User entity.  

Only staff can view/edit.


angaither:
:+1:

Issue #980, opened 2014-03-04
Dynamic login message - Welcome Back / You've Signed Up


mmwtsn:
Need to update the notification for when a user signs up for a campaign and when a user returns to a campaign they are already signed up for.

Notification messages should look like this:
https://s3.amazonaws.com/uploads.hipchat.com/34218/407723/jn8yiKwQ2EQROAt/Action%20Page%20Notification.pdf

aaronschachter:
Discussion with @mikefantini today:

Desired behavior for "Welcome back" message is any time a user logs in, NOT any time they return to a campaign node.  Ideally this should check whether or not the user is new, or for us to count number of logins (or at least set a field/flag for "returning user"), so you don't get "welcome back" upon your first login.

Need to investigate the one scenario where a user logs in via pitch page, and not displaying both messages.

mikefantini:
Yeah it's for any time a user logs in, _unless_ they are logging in to sign up for a campaign for the first time of that campaign's current run. 

Then they should get the "You're signed up" notification.

aaronschachter:
The campaign signup message has been completed in #992, so updating the issue title per what's left to build for this.

Per discussion with @mikefantini, the dynamic sign-in message that says "Welcome back" per comments above is not needed for 0.2  Moving this to backlog.

barryclark:
Didn't manage to get this in for v1.0. Closing this and can re-open if we do it in v2.

PR #992, opened 2014-03-05
Signup message

aaronschachter:
Refs #980 

Adds a `dosomething_signup_set_signup_message` to configure the campaign signup message.

Also error checks for any exceptions thrown from `message_broker_producer_request`.

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -307,10 +307,15 @@ function dosomething_signup_user_signup($nid, $account = NULL) {
  * Sets Drupal message for a signup for given $title.
  *
  * @param string $title
- *   Title to display within the Drupal message.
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -299,13 +299,23 @@ function dosomething_signup_user_signup($nid, $account = NULL) {
...
   }
 }
 
 /**
+ * Sets Drupal message for a signup for given $title.
+ *
+ * @param string $title
+ *   Title to display within the Drupal message.
+ */
+function dosomething_signup_set_signup_message($title) {
```
angaither: maybe title should default to NULL?

aaronschachter:
@angaither Updated `dosomething_signup_set_signup_message` with `$title` as an optional parameter.

angaither:
:+1:

PR #995, opened 2014-03-05
Updating Solr biases

### LINKED ISSUE #951: SOLR: Field Biasing
- [x] Author name should be omitted (currently set at 3.0)
- [x] The rendered comments associated with a node should be omitted (currently set at 0.5)
- [ ] Taxonomy term names only from the Cause vocabulary should be set at 3.0 (currently omitted)
- [ ] Taxonomy term names only from the Issue vocabulary should be set at 3.0 (currently omitted)

blisteringherb:
Fixes #951


lib/modules/dosomething/dosomething_search/dosomething_search.apachesolr_environments.inc
```diff
@@ -21,11 +21,117 @@ function dosomething_search_apachesolr_environments() {
...
   $environment->conf = array(
-    'apachesolr_direct_commit' => 0,
-    'apachesolr_index_updated' => 1392910246,
-    'apachesolr_last_optimize' => 1392910246,
-    'apachesolr_read_only' => '0',
-    'apachesolr_soft_commit' => 0,
+    'apachesolr_index_last' => array(
+      'node' => array(
+        'last_changed' => '1393971408',
+        'last_entity_id' => '715',
```
aaronschachter, line 27: I'm not too familiar with apachesolr config so forgive me if this is a dumb question but is this code generated from features?  Won't the last_changed and last_entity_id constantly be changing?

blisteringherb, line 27: These are exported by Features.  The consequence of having these vars in features is that it causes feature overrides and in this case reverting the feature doesn't actually revert.  Unfortunately there's no good way to fix this.  https://drupal.org/node/1984502  The "fix" involves not using features and scripting the setup of the variables.

aaronschachter, line 27: Aw, shizniz.   :+1: This is fine for now, but any thoughts about whether or not it makes to script later down the line so we don't run into that?

blisteringherb, line 27: I think if we notice that indexing is behaving strangely, we can move forward with scripting.  Since there were only a few people in that issue queue, I'm thinking it works pretty solidly for most applications.

PR #996, opened 2014-03-05
Removes secure pages module from info file & code.

### LINKED ISSUE #994: Remove Secure Pages Module
take out of info file
remove any code that references the module

angaither:
Fixes #994 

Refs #984


aaronschachter:
There are a bunch of variables being set in the dosomething.install that should be removed as well.

angaither:
@aaronschachter good call, updated.

aaronschachter:
:+1:

Issue #974, opened 2014-03-04
Reportback Files


aaronschachter:
- [x] Drop `field_image_user_reportback` field and remove all references to it in reportback feature.
- [x] Create new table `dosomething_reportback_files` 
- [x] Reportback form submit should insert record into it
- [x] Update /reportbacks view to display corresponding reportback files
- [x] Display uploaded files in reportback buildContent

aaronschachter:
Removing "blocker" label, form is in good shape for styling.  Back end code is in decent shape but still needs a little more love per checklist in issue description.

aaronschachter:
This issue was created to generally move Reportback Images out of the Drupal field and into their own db table, using a Drupal "file" Form API element to handle the upload.  

I've opened up issues for the remaining bugs and enhancements in order to close out this general issue, as the move has been completed.

PR #1008, opened 2014-03-05
Kill reportback field

aaronschachter:
Refs #974

Drops the Drupal Field on reportbacks and removes it from features.

Creates new db table to store them.


angaither:
:+1:

PR #1009, opened 2014-03-05
Replaces robots.txt during staging deployment

desmondmorris:
To prevent unwanted robots and crawlers on the public facing staging instance we must update the robots file in the webroot with: 

```
User-agent: *
Disallow: /
```

This is executed after deployments


aaronschachter:
Yes totes :+1: best practice

Issue #1014, opened 2014-03-05
action page (prove it) – report back modal markup


DFurnes:
We're abandoning the original field implementation for file uploads and using a custom-built form implementation. We need to be able to:
1. Upload a single image using a standard `<input type="file">` tag.
2. Display the previously uploaded image if editing an existing report back.

barryclark:
@DFurnes is this guy closed?

barryclark:
@DFurnes bump :bug:

DFurnes:
Still needs some work on the interface for uploading & viewing images.

PR #1016, opened 2014-03-05
File upload for Reportback form

aaronschachter:
Refs #974, #1014

Adds a "file" Form API element, and saves file into public://reportbacks/[nid].

This still a work in progress, this PR is to get the file element into the form to allow for fending.


aaronschachter:
actually we don't need this for stage just yet.  this could use more cleanup.

PR #1027, opened 2014-03-06
Reportback File upload

aaronschachter:
Refs #974, #1014
- Removes admin menu links for Reportback fields/display (no more fields)
- Adds a `file` element to the `dosomething_reportback_form` and adds relevant file validating / save handling
- Uses just one function: `dosomething_reportback_save($values)` to replace `dosomething_reportback_insert_reportback` and `dosomething_reportback_update_reportback` functions
- Moves a lot of the general Reportback API stuff into the Reportback Entity classes.
  - `save`: adds a file into dosomething_reportback_files on save if $entity->fid is set
  - `save` and `delete`: Handles the reportback log writes (instead of using hook_entity_[operation] functions)

TODO:
- Update documentation
- Display uploaded file on the form if updating -- last outstanding task in #1014
- Rename files upon upload to avoid overwriting
- Form validation
  - file upload is mandatory if new reportback, otherwise not mandatory
  - Make sure quantity value is numeric
- ~~Form tweaks (change labels, add placeholder values where applicable)~~


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -458,37 +459,25 @@ function dosomething_reportback_exists($nid, $uid = NULL) {
...
-        'email' => $account->mail,
-        'uid' => $account->uid,
-        'first_name' => dosomething_user_get_field('field_first_name', $account),
-        'event_id' => '??',
-        'impact_verb' => '??',
-        'impact_number' => '??',
-        'impact_noun' => '??',
-      );
-      dosomething_user_send_message_request('campaign_reportback', $params);
-    }
```
angaither, line 491: probably don't want to be removing this stuff.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -585,24 +529,33 @@ function dosomething_reportback_log_insert($op, $entity) {
...
 /**
- * Implements hook_entity_update().
+ * Returns the path to write a reportback file for given node $nid.
  */
-function dosomething_reportback_entity_update($entity, $type) {
-  if ($type == 'reportback') {
-    dosomething_reportback_log_insert('update', $entity);
-  }
+function dosomething_reportback_get_file_dir($nid) {
+  return 'public://reportbacks/' . $nid;
```
angaither, line 540: :+1:

angaither:
Looks ok, just keep the message request in!

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
aaronschachter, line 491: its not removed!  its been abstracted into `dosomething_reportback_message_broker`

angaither, line 491: :dancers: Merge away! :+1:

PR #1028, opened 2014-03-06
Hides asterisks on required form fields.

### LINKED ISSUE #869: Auth - Create an Account Required Fields Needed
First Name and Birthday should both be required fields and they should have \* indicating that they are required.

image

DFurnes:
Fixes #869.

Drupal.org gave some friendly advice along the way. :trollface: 
drooooopal

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/feature/_drupal.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/feature/_drupal.scss
@@ -8,6 +8,12 @@
   display: none;
 }
 
+// Hide asterisks on required form fields.
+.form-required {
+  display: none;
+}
+
+
 
 // @TODO: Remove date.css to override this more cleanly
 .container-inline-date .date-padding {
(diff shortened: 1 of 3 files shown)


aaronschachter:
Why not just hide the asterik with css instead of unset it?

DFurnes:
We can, but if that's the case then we should be adding a '#required' to the first name and birthday fields as well since neither of them have it.

DFurnes:
That being said, using a PHP hook to add inline CSS to the page with a `display: none !important` is almost certainly not the right way to be doing this.

aaronschachter:
If first name and birthday are required, it definitely makes sense to add to the form.

This doesn't work? https://drupal.org/comment/229295#comment-229295

angaither:
@DFurnes since these are required fields, rather not unset them as required... if adding it to first name and birthday we can do that

angaither:
@aaronschachter jinx

DFurnes:
@aaronschachter That'd work. I'm happy to do it that way instead.

Just read the documentation and I see that `#required` is setting validation on the server-side. Definitely agreed that we shouldn't be unsetting it, good call. Frustrating that all this stuff is intertwined though.

angaither:
Alright, I'll make and issue and do it :ballot_box_with_check:

DFurnes:
Pair coded with @desmondmorris, merging in.

PR #1029, opened 2014-03-06
Fact Styles and Markup Updates

mmwtsn:
Desktop: styles the "Know It" fact sources.
Mobile: adds and styles the fact sources below the "Prove It" section.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
@@ -108,6 +108,7 @@
     <script id="modal-partner-<?php print $delta; ?>" type="text/cached-modal">
       <a href="#" class="js-close-modal modal-close-button">×</a>
       <?php print $partner['copy']; ?>
+      <?php if (isset($partner['image'])): print $partner['image']; endif; ?>
     </script>
     <?php endforeach; ?>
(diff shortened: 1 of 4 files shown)


lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -94,17 +68,51 @@
...
+        <div class="legal"><sup><?php print ($key + 1); ?></sup><?php print $source; ?></div>
+      <?php endforeach; ?>
+    </script>
+    <?php endif; ?>
+
+    <?php if (isset($partner_info)): ?>
+    <?php foreach ($partner_info as $delta => $partner): ?>
+    <script id="modal-partner-<?php print $delta; ?>" type="text/cached-modal">
+      <a href="#" class="js-close-modal modal-close-button">×</a>
+      <?php print $partner['copy']; ?>
```
aaronschachter, line 110: I'm removing code below out of #1049 to avoid merge conflict (since the partner info modal has changed locations).  Add this in to print the Image field.

```
<?php if (isset($partner['image'])): print $partner['image']; endif; ?>
```

mmwtsn, line 110: @aaronschachter Thanks for the heads up. Added below. Could you check out 9194995 ?

PR #1030, opened 2014-03-06
Redirect to referer after login.

### LINKED ISSUE #1023: auth - after log in user is taken to /user; should reload page they started on
When I log in to the site I am redirected to /user. 

Instead I should remain on the page that I was on to trigger the auth flow. i.e. I'm on the homepage, I click on Log In, after I go through auth, I am taken back to the homepage.

https://trello.com/c/xC2uouKl

angaither:
Fixes #1023

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -214,14 +214,15 @@ function dosomething_user_is_under_thirteen($user = NULL) {
  */
 function dosomething_user_login_submit($form, &$form_state) {
   // If nid is not present, nothing to sign up for.  Exit.
-  if (!isset($form_state['input']['nid'])) { return; }
-  $nid = $form_state['input']['nid'];
-  // Redirect to node (this is needed for user_register form).
-  $form_state['redirect'] = 'node/' . $nid;
-  // Signup global user for node $nid.
-  if (module_exists('dosomething_signup')) {
-    dosomething_signup_user_signup($nid);
+  if (isset($form_state['input']['nid'])) {
+    $nid = $form_state['input']['nid'];
+    // Signup global user for node $nid.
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -214,14 +214,15 @@ function dosomething_user_is_under_thirteen($user = NULL) {
...
-    dosomething_signup_user_signup($nid);
+  if (isset($form_state['input']['nid'])) {
+    $nid = $form_state['input']['nid'];
+    // Signup global user for node $nid.
+    if (module_exists('dosomething_signup')) {
+      dosomething_signup_user_signup($nid);
+    }
   }
+  // After all logins redirect to page user was just on.
+  $form_state['redirect'] = $_SERVER['HTTP_REFERER'];
```
aaronschachter, line 225: Looks like it will work, just reading up on using HTTP_REFERER ... 

http://drupal.stackexchange.com/questions/44264/how-to-redirect-user-to-previous-page-after-form-submit suggests having a fallback in place, and "Remember that some browser does not set $_SERVER['HTTP_REFERER'] and if a user browses directly to the page, "

What are your thoughts about that?  Would an alternative be to set the form's action, like we do in the campaign_data ?

angaither:
I remember needing to set the action to submit the hidden forms to the correct urls, then this 'redirect' happens after that action.
If someone goes directly to the '/user' page they will be taken to their profile.

aaronschachter:
Right, it makes sense that you'd get taken their profile on `/user`.

I'm happy to merge and see how goes.  Google shows some talk of issues with varnish and HTTP_REFERER, not sure if we're set up with varnish on our local installs.

angaither:
:trollface: varnish....
we will give it a try.....  wishing on :stars:

PR #1038, opened 2014-03-06
Added dosomething_sms to the info and make files

jonuy:
Ran `ds build --install` and verified dosomething_sms and its dependencies got installed and enabled.

For #894


aaronschachter:
:+1:

Issue #692, opened 2014-02-18
campaign (header) – style sponsor logo


mmwtsn:
Once Aaron passes over the sponsor logo, one version needs to be styled.

mmwtsn:
H&M, Keri and Mike all prefer the lower right-hand corner treatment. As the design is not finalized, other versions will be tested in Optimizely.

mikefantini:
Here is the H&M logo. Need it to be added here: http://staging.beta.dosomething.org/taxonomy/term/369/edit?destination=admin/structure/taxonomy/partners

![h m logo red](https://f.cloud.github.com/assets/6330971/2350762/895e7a6c-a574-11e3-93a1-b205f5b0eb46.png)

mmwtsn:
Fixed by #1096

PR #1044, opened 2014-03-06
Sponsor logos

aaronschachter:
Let there be corporate.

Unblocks #692


desmondmorris:
This seems okay to me.

PR #1045, opened 2014-03-06
Redis ground work

desmondmorris:
This merge installs redis server into our VM and adds and enables the Drupal redis module.  It also pulls down the predis library.

This does not include any of the default settings that are necessary to wire this up.  These will come in a future merge.

### DIFF
--- a/config/settings.php
+++ b/config/settings.php
@@ -1,4 +1,13 @@
 <?php
+
+/**
+ * Lets do something about globals
+ */
+define('DS_PROFILE_PATH', 'profiles/dosomething');
+define('DS_MODULES_PATH', DS_PROFILE_PATH . '/modules');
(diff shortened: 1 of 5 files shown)


aaronschachter:
Just checking - any issues with enabling the module but not configuring it?  Errors, etc

desmondmorris:
Should be okay.  This will only work if the cache backends are set in settings.php

@aaronschachter but let me triple check :)

aaronschachter:
If all looks good with the triple check then :+1: 

email password

mshmsh5000:
Great -- I had this set up the same locally. I did have to wrangle the settings, because autoload stuff wasn't happening correctly:

https://gist.github.com/mshmsh5000/9573307

desmondmorris:
@mshmsh5000 that was golden.  thank you

mshmsh5000:
+1 Tested and value added.

PR #1047, opened 2014-03-06
Import user data

### LINKED ISSUE #1036: Users: Import user table from old world
- only keep active users
- store mail, password, created and old user id
- store mobile #

angaither:
This imports all the basic data needed for validation for a user to log in on the new site.
- email
- phone
- name

[Other user data](https://gist.github.com/angaither/4131041918493fc813fc) still needs to be migrated. 

Fixes #1036
@mshmsh5000

### DIFF
--- a/scripts/import-users.php
+++ b/scripts/import-users.php
@@ -28,6 +28,7 @@ foreach ($result as $user_row) {
     $user['access'] = $user_row->access;
     // Set the init value as the old world uid.
     $user['init']= $user_row->uid;
+    $user['status'] = $user_row->status;
     $account = user_save('', $user);
   }
 
--- a/scripts/import-users.php
+++ b/scripts/import-users.php
@@ -29,7 +29,7 @@ foreach ($result as $user_row) {
      $user['created'] = $user_row->created;
      $user['access'] = $user_row->access;
      // Set the init value as the old world uid.
-     $user['init']= $user_row->uid;
+     $user['init']= $user_row->init;
      $user['status'] = $user_row->status;
      $account = user_save('', $user);
   }
--- a/scripts/import-users.php
+++ b/scripts/import-users.php
@@ -3,32 +3,53 @@
  * Script to import users from old world into new world.
  *
  * to run:
- * import user data from old world site into temporary table
+ * import user data from old world site into temporary tables
+ * will need 'users', 'profile', and 'field_data_field_user_mobile'.
  * drush --script-path=../scripts/ php-script import-users.php
  *
  */
 
 $user_table = '_users_old_world';
+$profile_id = '_profile_old_world';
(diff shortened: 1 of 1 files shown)


scripts/import-users.php
```diff
@@ -0,0 +1,55 @@
...
+    $user = array();
+    // Will be overridden by uid.
+    $user['name'] = user_password(8);
+    $user['pass'] = $user_row->pass;
+    $user['mail'] = $user_row->mail;
+    $user['created'] = $user_row->created;
+    $user['access'] = $user_row->access;
+    // Set the init value as the old world uid.
+    $user['init']= $user_row->uid;
+    $account = user_save('', $user);
```
aaronschachter: Do we need to also bring in status and assign "authenticated user" role?  As the script is now, is an imported user able to login?

desmondmorris:
I would checkin with @joshcusano first.  I had a meeting with him last week and he has a table that exports this data daily

joshcusano:
@aaronschachter @angaither I have this running daily, just adding the pass column from the users table. Let me know if it could be helpful.

mshmsh5000:
@angaither If we import users from the old app, with passwords encoded with that salt string, we'll need to update the salt string in new the app's settings.php for this to work. Once we do this, it will require resetting passwords for all users we want to keep (i.e., users who have authored content in the beta app).

We'll want to run this test in QA ASAP to make sure the plan makes sense.

angaither:
@mshmsh5000 oh snap. yes. We can test whenever. 
I'll just need to import the temporary tables.

angaither:
Updated to keep old world uids in place.

mshmsh5000:
Oh yeah! Let's see if we can log in, now. I've copied the legacy QA salt string to beta QA.

aaronschachter:
This looks good.  One element to consider is roles.  We have both "staff" and "administrator" roles in the legacy app.  Also, wouldn't we want to add "authorized user" role as well?

angaither:
@aaronschachter Yeah, there's quite a few roles in the old app. I wasn't sure what to keep...

aaronschachter:
For sure there are. I did a bunch of work on roles last year, the only important ones are
- Authenticated user (although... i guess we could just grant this to everyone regardless of importing?)
- Staff
- Administrator

mshmsh5000:
+1 on granting Authenticated User to everyone.

Staff and Administrator sets are small enough that we can assign manually if'n we need to -- might be easier than doing this in code.

aaronschachter:
True indeed

angaither:
Well, if that's the case anyone who has a status of 1 is automatically an `authenticated user`
http://drupal.stackexchange.com/questions/57444/authenticated-users-missing-from-users-roles-table

aaronschachter:
:+1:

PR #1048, opened 2014-03-06
Added missing quote in image tag

angaither:
..yeah, fixes things.


aaronschachter:
:+1:  -- Would you mind also updating the comments in the function too so the accepted values are accurate?  they currently read as "The image field to pull from the desired image ratio (landscape, sq)."

Issue #958, opened 2014-03-03
Sponsor modals: Preprocess image


aaronschachter:
Refs #768 -  Preprocess `field_partners.field_image_partner`

aaronschachter:
Renaming issue to just account for preprocessing the Sponsor Modal's image. Moved preprocessing the video into #1086

PR #1049, opened 2014-03-06
Sponsor modal images

aaronschachter:
Moves preprocessing partners vars into its own function.

Additionally adds selected Sponsor Modal Image node into the campaign template if value is present.

Refs #958 and #768


blisteringherb:
:+1:

PR #1056, opened 2014-03-07
Fixes #939

### LINKED ISSUE #939: DS update-files removes symlinks
Run `ds update-files`, make changes, then wonder if you still know how to PHP.  

And then realize oh, just no more symlinks

Not sure if this is by design, but seems like a bug?

aaronschachter:
Let there be symlinks.


desmondmorris:
This looks good. Ship it

PR #1062, opened 2014-03-07
Adds ds pull command

### LINKED ISSUE #929: Update ds --help documentation for stage downloads
Refs #872 

Add documentation for the `ds update-db` and `ds update-files` into the output on `ds --help`

aaronschachter:
Replaces `ds update-db` and `ds update-files` functions with a `ds pull` command.

Adds usage documentation into `ds --help` documentation, fixes #929

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -59,7 +59,7 @@ function art {
 #===============================================================================
 function help {
   read -d '' help <<"USAGE"
-usage: ds [build [--install]|dbup|dbdump] [--help]
+usage: ds [build [--install]] [--help]
 
 Runs helper scripts for DoSomething environment.
 
--- a/bin/ds
+++ b/bin/ds
@@ -79,6 +79,15 @@ ds grunt [--watch]
   --watch
   Continues watching for changes and recompiles assets as necessary.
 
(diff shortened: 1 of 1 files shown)


bin/ds
```diff
@@ -334,6 +343,44 @@ function grunt {
 
 }
 
+#=== FUNCTION ==================================================================
+# NAME: pull_db
+# DESCRIPTION: Pulls down database from given environment onto local.
+#===============================================================================
+function pull_db {
+  if [[ $1 == "stage" ]]
```
desmondmorris, line 351: Do you think we should prepare this for production as well right now?

So instead of checking for "stage", we should be checking for the drush aliases keys, so "staging" or "production".  The you can just pass the alias to the sql-sync command a few lines below

aaronschachter, line 351: Could do... the alias for prod is set up already?  Sounds like just modifying the `check_alias_param` function and then yea modifying these `pull_db` and `pull_files` to run on whatever param its been given.

I figured we'd want to modify this to not hardcore "stage" but also think we could get away with it for now - I was under the impression that we only have the staging alias.

PR #1065, opened 2014-03-07
Campaign custom variables

### LINKED ISSUE #895: Campaign: sponsor color variable
We'll need a way to set custom sponsor colors and access them in the campaign template. Passing them over within a `$sponsor_color` variable should do the trick.
- Alternate header color (hex value)
- Alternate Prove It pattern (file upload)

Via #781

aaronschachter:
Fixes #895

Adds a fieldset for on an existing campaign node's edit form, where we can set a custom alt color and custom alt background image.

The variables in the template are named `alt_color` and `alt_bg_src`.

The values are stored as Drupal variables instead of fields, because only a small number of campaigns will actually need these values.  

The list of available variables are defined within a function called `dosomething_campaign_get_custom_varnames`, so it's easier to add, update, and maintain custom variables in the future.  Looking into possibly using this storage method for scholarships as well, since few campaigns have them.

Also renames incorrectly named `dosomething_image_get_image_url_by_fid` function as `dosomething_image_get_themed_image_by_fid`.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -120,8 +120,11 @@ function dosomething_campaign_get_custom_var_prefix($nid) {
 /**
  * Implements hook_node_delete().
  */
-function dosomething_campaign_node_delete() {
-  // @todo: Delete any custom variables associated with node.
+function dosomething_campaign_node_delete($node) {
+  // Delete any custom variables associated with node.
+  foreach (dosomething_campaign_get_custom_varnames($node->nid) as $name) {
+    variable_del($name);
+  }
 }
 
 /*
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -259,6 +365,29 @@ function dosomething_campaign_preprocess_node(&$vars) {
...
+ *
+ * @param array $vars
+ *   Node variables, passed from preprocess_node.
+ */
+function dosomething_campaign_preprocess_custom_vars(&$vars) {
+  // Get prefix to check for custom campaign variables.
+  $prefix = dosomething_campaign_get_custom_var_prefix($vars['nid']);
+  // Check for custom alt color.
+  $alt_color = variable_get($prefix . 'alt_color');
+  if (isset($alt_color) && !empty($alt_color)) {
```
jonuy: Just curious, isn't `!empty()` alone sufficient?

jonuy:
:+1: to more colors and images

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
aaronschachter: I think you're right.  Gonna update this then merge.

Issue #1054, opened 2014-03-06
Reportback: public files subdirectory error


aaronschachter:
The `dosomething_reportback_get_file_dir` function (https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module#L556) has been working on my local machine when testing the repoertback form, but it's not happy on staging and won't allow writing files.

The specified file temporary://me (alt).jpeg could not be copied, because the destination directory is not properly configured. This may be caused by a problem with file or directory permissions. More information is available in the system log.
There was an error. Please try again.

I actually got the same error from my home and its not working for @DFurnes either. I guess my local is magic.

If the function is changed to just write to "public://" -- it works for @DFurnes

related reading http://drupal.stackexchange.com/questions/30113/configuring-the-temporary-directory

aaronschachter:
`dosomething_reportback_get_file_dir` is attempting to write files to "public://reportbacks/[nid]" where nid is the campaign nid.  Also see `dosomething_reportback_form_validate_file` for where we're attempting to write the file. https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module#L349

PR #1066, opened 2014-03-07
Hotfix for reportback upload directory

aaronschachter:
This is a hotfix to get reportback uploads working on staging for today's deployment.

Tough to troubleshoot this on my work machine because when the upload directory is set to `reportbacks/[nid]`, it works.  However I can replicate #1054 from my personal machine, so I'll continue to investigate.


blisteringherb:
:+1:

PR #1067, opened 2014-03-07
Fixes #1064 - missing email value in payload

### LINKED ISSUE #1064: Message Broker - email value missing in $payload
When `message_broker_producer_request('produceTransactional', $payload);` is called in dosomething_user_send_message_request() the email address is missing in $params['email'].

deezone:
Fixes #1064

The payload value for email was missing in both the new user registration and the password reset. The email value is required to build the related message to be sent my Mandrilll.


jonuy:
Looks good.

PR #1070, opened 2014-03-07
Fixes #1069 - Move to mbp naming convention

### LINKED ISSUE #1069: Message Broker - rename message_broker to mbp
The word "message" is ambiguous and can be confused with the display of a "message" to the user via things like drupal_set_message. Rename Message Broker references to "mbp" is consistent with the Message Broker application naming convention.

deezone:
Fixes #1069

Removes the word "message" with mbp. "Message" is used to reference messages being displayed to the user via drupal_set_message()


aaronschachter:
I like it :+1:

PR #1074, opened 2014-03-10
View reportback images

aaronschachter:
Refs #974 

Uses `hook_views_data` to expose the `dosomething_reportback_file` db table to views.

Updates `node/*/reportbacks` view to display uploaded images.

Displays uploaded images in the reportback entity view, `reportback/[rbid]`.  This entity view is staff only.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.views_default.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.views_default.inc
@@ -85,6 +85,11 @@ function dosomething_reportback_views_default_views() {
   $handler->display->display_options['empty']['area']['empty'] = TRUE;
   $handler->display->display_options['empty']['area']['content'] = 'No reportbacks found.';
   $handler->display->display_options['empty']['area']['format'] = 'markdown';
+  /* Relationship: Reportback: User uid */
+  $handler->display->display_options['relationships']['uid']['id'] = 'uid';
+  $handler->display->display_options['relationships']['uid']['table'] = 'dosomething_reportback';
+  $handler->display->display_options['relationships']['uid']['field'] = 'uid';
+  $handler->display->display_options['relationships']['uid']['required'] = TRUE;
   /* Field: Reportback: Reportback rbid */
   $handler->display->display_options['fields']['rbid']['id'] = 'rbid';
   $handler->display->display_options['fields']['rbid']['table'] = 'views_entity_reportback';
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.views_default.inc
```diff
@@ -85,40 +98,74 @@ function dosomething_reportback_views_default_views() {
...
   $handler->display->display_options['fields']['quantity']['table'] = 'dosomething_reportback';
   $handler->display->display_options['fields']['quantity']['field'] = 'quantity';
+  /* Field: File: Path */
+  $handler->display->display_options['fields']['uri']['id'] = 'uri';
+  $handler->display->display_options['fields']['uri']['table'] = 'file_managed';
+  $handler->display->display_options['fields']['uri']['field'] = 'uri';
+  $handler->display->display_options['fields']['uri']['relationship'] = 'fid';
+  $handler->display->display_options['fields']['uri']['label'] = 'Pic';
+  $handler->display->display_options['fields']['uri']['alter']['alter_text'] = TRUE;
+  $handler->display->display_options['fields']['uri']['alter']['text'] = '<img src="[uri]" width="200" height="200"/>';
```
angaither, line 158: @aaronschachter is there any way we can theme this image with the functions you created instead of hardcoding an overwrite into the view?

angaither:
"This entity view is staff only." -- don't see this in the view...but also, monday.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.views_default.inc
aaronschachter, line 158: I was having a lot of trouble trying to get the dosomething_reportback module to override this views field with all sorts of various theme hooks, and ended up just doing this for now to keep moving.  This view uses the Seven admin theme, so I wasn't too interested in trying to add any code into that (or creating our own version etc)

aaronschachter:
@angaither https://github.com/aaronschachter/dosomething/blob/view_reportbacks/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.views_default.inc#L28

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.views_default.inc
angaither, line 158: Totally fine, it might be good to add it as a @TODO just so we can flag it if we care later

aaronschachter, line 158: Created #1076 to keep track of it.

angaither:
:+1:

PR #1097, opened 2014-03-10
Reportback form validation

### LINKED ISSUE #1087: Reportback form validation
- [x] Quantity must be a positive integer
- [x] File upload is required if creating a reportback, else not required for update.

aaronschachter:
Fixes #1087
- Checks that quantity is a non zero integer with handy Drupal core shiz `'#element_validate' => array('element_validate_integer_positive'),`
- Image upload is only mandatory upon reportback creation (not update)

Moves rendering of `reportback_form` into `hook_node_view` instead of `hook_preprocess_node` to get drupal messages to appear correctly after submitting (before this change, they wouldn't appear until refreshing the page after submit -- see http://pixeljets.com/blog/be-careful-drupalgetform-theme-layer)

Additionally, makes same change for `zendesk_form`.


angaither:
:+1:

PR #1098, opened 2014-03-10
Updated campaign hero images.

### LINKED ISSUE #912: Images: Filter Needed on Campaign Hero Images

angaither:
Fixes #912


aaronschachter:
Looks ok, beware of merge conflict. https://github.com/DoSomething/dosomething/pull/1096/files#diff-ae398bb8864fc3c26317f7ccaef04c0aL341

PR #1108, opened 2014-03-10
Fixes #1094

### LINKED ISSUE #1094: Reportback log bug
Upon creating a new reportback, the fid and num_files are not populated in the corresponding "insert" record in the `dosomething_reportback_log` table.

aaronschachter:
Fixes #1094

Changes the `Reportback.insertLog` method to query the `dosomething_reportback_file` table at the time of `insertLog`, instead of checking the `fids` property (which wont be up to date unless we loaded the entity again, after the insert into files table).


blisteringherb:
:+1:

PR #1141, opened 2014-03-11
Pitch [sign up] Fix

### LINKED ISSUE #1131: Campaign pitch page: Disappearing sign up button
If signed in and you are presented with the pitch page, the signup button does not appear.  

This is because of the recent change to print the header twice on the page, Drupal does not allow you to print the same form on the page twice: https://drupal.org/node/766146

mmwtsn:
These commits add a `dosomething_campaign_add_inline_css` function to the Campaign Module which allows us to print styles in the `<head>` rather than spitting them out inline in the template.

Only prints one `<header>` and fixes #1131


DFurnes:
Looks smart to me. +1

aaronschachter:
Pulled down and tested. Looks good.

PR #1150, opened 2014-03-11
Prove It image fix

aaronschachter:
Not sure why this was happening but could definitely replicate that the image was not appearing for an authenticated user, and would appear for an admin.

Using preprocess values from the `entity_metadata_wrapper` instead of the `$vars` array seems to do the trick.  Will update the issue with findings of why that is once we get some breathing room.


blisteringherb:
:+1:

aaronschachter:
Inspecting the `$vars` array as an authenticated user shows that the `reportback_image` variable is an array with an empty element.
screen shot 2014-03-11 at 3 25 29 pm

This is what it looks like for an administrator:
screen shot 2014-03-11 at 3 26 25 pm

Here's why this happened: `dosomething_user_node_access`.  This is in place to deny access to a user from viewing an Image node, e.g. navigating to node/840 and seeing just an image node.

Issue #1079, opened 2014-03-10
Reportback File renaming


aaronschachter:
Rename reportback files to a unique name upon upload, and avoid these:

PDOException: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '' for key 'uri': INSERT INTO {file_managed} (filesize, status, timestamp, uuid) VALUES (:db_insert_placeholder_0, :db_insert_placeholder_1, :db_insert_placeholder_2, :db_insert_placeholder_3); Array ( [:db_insert_placeholder_0] => 0 [:db_insert_placeholder_1] => 1 [:db_insert_placeholder_2] => 1394466530 [:db_insert_placeholder_3] => 6cb5af31-c702-4765-9b2c-88f59cec404e ) in drupal_write_record() (line 7201 of /vagrant/html/includes/common.inc).

Look into using https://api.drupal.org/api/drupal/includes%21file.inc/function/file_destination/7

PR #1152, opened 2014-03-11
Comeback Clothes SMS report backs

jonuy:
Tested and works locally so far...

@aaronschachter and whoever else ends up looking at this, a couple things in particular that I'd appreciate eyes on:
- `dosomething_user_clean_cell_number()`:
  - Mobile Commons includes the leading `1` for mobile numbers in their payload. So I updated this to allow 11 digit numbers to be valid. But let me know if you'd want to handle this differently.
- `dosomething_user_create_user_by_cell()`:
  - Since some numbers may be trying to submit a report back without having come to our site to create an account first, I'm silently creating an account for them. It's like what we did on the old site. Did I create all that I needed to for the user account?
  - Also, the `name`/`username` field doesn't seem to be sticking with the way I'm doing it. Any ideas?
- Everything in `ConductorActivitySmsReportBack`. Am I doing this right? It seems like files and getting saved and the `dosomething_reportback` table is getting updated.:
  - Summary of the logic in there:
    - Find an existing account. If none, then create one.
    - Find an existing report back.
    - Download and save the MMS image.
    - Create/update the report back submission.
    - Update things on the Mobile Commons end: updating profile, opting out of one campaign, and opting into another.

### DIFF
--- a/lib/modules/dosomething/dosomething_sms/plugins/activity/reportback/dosomething_sms_submit_reportback/ConductorActivitySmsReportBack.class.php
+++ b/lib/modules/dosomething/dosomething_sms/plugins/activity/reportback/dosomething_sms_submit_reportback/ConductorActivitySmsReportBack.class.php
@@ -58,6 +58,9 @@ class ConductorActivitySmsReportBack extends ConductorActivity {
     $state = $this->getState();
     $mobile = $state->getContext('sms_number');
 
+    // Mobile Commons sends the international code in its payload. Remove it.
+    $mobile = substr($mobile, -10);
+
     // Get user by cell number if it exists. Otherwise create it.
     $user = dosomething_user_get_user_by_cell($mobile);
     if (!$user) {
--- a/lib/modules/dosomething/dosomething_sms/plugins/activity/reportback/dosomething_sms_submit_reportback/ConductorActivitySmsReportBack.class.php
+++ b/lib/modules/dosomething/dosomething_sms/plugins/activity/reportback/dosomething_sms_submit_reportback/ConductorActivitySmsReportBack.class.php
@@ -64,8 +64,12 @@ class ConductorActivitySmsReportBack extends ConductorActivity {
       $user = dosomething_user_create_user_by_mobile($mobile);
     }
 
-    // Check for an rbid indicating that the user's submitted a report back previously.
-    $values = array();
+    // Initialize reportback values, defaulting to create a new reportback.
+    $values = array(
+      'rbid' => 0,
+    );
+
+    // Check for a previously submitted reportback to update instead.
     if ($rbid = dosomething_reportback_exists($this->nid, $user->uid)) {
       $values['rbid'] = $rbid;
     }
--- a/lib/modules/dosomething/dosomething_sms/plugins/activity/reportback/dosomething_sms_submit_reportback/ConductorActivitySmsReportBack.class.php
+++ b/lib/modules/dosomething/dosomething_sms/plugins/activity/reportback/dosomething_sms_submit_reportback/ConductorActivitySmsReportBack.class.php
@@ -61,7 +61,7 @@ class ConductorActivitySmsReportBack extends ConductorActivity {
     // Get user by cell number if it exists. Otherwise create it.
     $user = dosomething_user_get_user_by_cell($mobile);
     if (!$user) {
-      $user = dosomething_user_create_user_by_cell($mobile);
+      $user = dosomething_user_create_user_by_mobile($mobile);
     }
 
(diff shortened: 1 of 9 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -384,10 +385,41 @@ function dosomething_user_clean_cell_number($number) {
...
+/**
+ * Create a user account when provided with only a cell phone number.
+ *
+ * @param string $number
+ *   User's phone number.
+ */
+function dosomething_user_create_user_by_cell($number) {
+  if ($clean_number = dosomething_user_clean_cell_number($number)) {
+    $user_data = array(
+      'name' => $clean_number, // @todo This doesn't seem to work. UID gets set as the name instead. Can it not be only numbers? meh, seems like $clean_number . '@mobile' does nothing either.
```
aaronschachter: `dosomething_user_user_insert` is what's auto-setting the name as uid. Shouldn't really need todo anything here.

jonuy: Ah ok, that's what I was seeing happen. If we're cool with that just being the uid, then I'll take the 'name' line out altogether.

lib/modules/dosomething/dosomething_sms/plugins/activity/reportback/dosomething_sms_submit_reportback/ConductorActivitySmsReportBack.class.php
```diff
@@ -0,0 +1,118 @@
...
+      $values['rbid'] = $rbid;
+    }
+
+    // Get the MMS URL from the provided context.
+    $pictureUrl = $state->getContext($this->mmsContext);
+
+    // Download, save, and move the file to proper directory.
+    $pictureContents = file_get_contents($pictureUrl);
+    $file = file_save_data($pictureContents, $pictureFilename);
+    $file = file_move($file, dosomething_reportback_get_file_dir($this->nid));
```
aaronschachter, line 86: This looks fine for now.  Heads up that I'm working on a few bugs to better organize the directory and filenames (#1079, #1054).  More of a note to myself that I'll need to make these changes here too when I change them in the reportback form -- although we'll want to test on mobile as well just to be sure.

lib/modules/dosomething/dosomething_sms/plugins/activity/reportback/dosomething_sms_submit_reportback/ConductorActivitySmsReportBack.class.php
```diff
@@ -0,0 +1,118 @@
...
+    $mobile = $state->getContext('sms_number');
+
+    // Get user by cell number if it exists. Otherwise create it.
+    $user = dosomething_user_get_user_by_cell($mobile);
+    if (!$user) {
+      $user = dosomething_user_create_user_by_cell($mobile);
+    }
+
+    // Check for an rbid indicating that the user's submitted a report back previously.
+    $values = array();
```
aaronschachter: I'm a little surprised this worked for reportback creation, I'd imagine you'd need to initialize rbid = 0.  May be good to initailze it anyway:

```
// Initialize reportback values, defaulting to create a new reportback.
$values = array(
  'rbid' => 0,
);
```

lib/modules/dosomething/dosomething_user/dosomething_user.module
aaronschachter: You may need to still set a name, but can just set it to a random string like `user_password()`.  If its not mandatory, fine to leave out.

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -384,10 +385,41 @@ function dosomething_user_clean_cell_number($number) {
...
   return FALSE;
 }
+
+/**
+ * Create a user account when provided with only a cell phone number.
+ *
+ * @param string $number
+ *   User's phone number.
+ */
+function dosomething_user_create_user_by_cell($number) {
```
aaronschachter: I know you're kind of jumping into the middle of this, but it'd be great if we just stuck with either calling this `cell` or `mobile`.  We've built the field as `mobile` in the database, and also have functions like `dosomething_user_get_mobile()`.  Any interest in taking one for the team and global replacing `cell` as `mobile` ?  Otherwise let's create a new cleanup issue for it.

jonuy: Sounds good. I'll at least change this one to `mobile`. And throw in an issue to update anywhere else that uses `cell`.

PR #1161, opened 2014-03-11
Campaign form - Alt BG field fix

### LINKED ISSUE #1134: campaign (form) - Alt background pattern field bug
When I upload an image in the Alt background pattern field and hit save, the background shows up properly. However, if I go to edit the campaign later I am unable to save and get this error message:

image

aaronschachter:
Fixes #1134

Error was occurring because we're using a `file_managed` form element to store a File fid, and wasn't updating the corresponding records in the `file_usage` table.  This PR fixes that, implementing the `dosomething_campaign_save_alt_bg_fid` function to track the usage of files we're setting for each `alt_bg_fid` variable.

Also gets rid of the `dosomething_campaign_get_custom_varnames` function, because it was just not readable at all by looping through it on the save (and not worth it for just 2 variables)


angaither:
:+1:

Issue #1158, opened 2014-03-11
Change function names that use "cell" to use "mobile"


jonuy:
Examples: 
`dosomething_user_clean_cell_number()`
`dosomething_user_valid_cell()`
`dosomething_user_get_user_by_cell()`
etc.

PR #1166, opened 2014-03-12
Renaming methods that use 'cell' to use 'mobile'

jonuy:
For issue #1158


aaronschachter:
:+1: awesome!  thanks for cleaning this up, @jonuy

PR #1167, opened 2014-03-12
Reportback file directories

aaronschachter:
Writes reportback files into the "reportbacks/[nid]" directory within the public files directory, for whatever [nid] the reportback is associated with.

Renames reportback file with pattern `uid_[uid]-nid_[nid]-[index]`, retaining the file extension of the uploaded file.  e.g. "uid_5-nid_43-0.jpg", "uid_5-nid-43-1.png"


angaither:
:+1:

Issue #1061, opened 2014-03-07
File field directories


aaronschachter:
- [ ] Update the Image fields to write into a specific folder in files directory, maybe just "images", since they'll all be done thru the Image node type.  Or possibly "images/landscape", "images/square" etc.
- [ ] Update campaign alt bg to upload into /campaigns directory

aaronschachter:
Seems we've all been ok with the image files living in the directories they've been living in (the default). Can always re-open if anyone feels otherwise.

PR #1168, opened 2014-03-12
Image field directory

aaronschachter:
Refs  #1061

Upload Image field files into an "images" directory within the public dir.


angaither:
:+1: should we move all previously uploaded images?

aaronschachter:
That'd be nice... but what's the best way to go about that?  I've also been contemplating whether or not we should move the public directory to point to just `files` in the webroot instead of `sites/default/files`

PR #1169, opened 2014-03-12
DS Build: Revert features

aaronschachter:
Reverts features upon `ds build`.

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -161,8 +161,11 @@ function build {
   drush updb -y
 
   # Enable modules in info file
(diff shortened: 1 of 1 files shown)


angaither:
we should really be checking and reverting features that are overridden, this will slow things wayyy down.

aaronschachter:
what is the full command to use along with the `drush fl | grep "Overridden"`?

angaither:
`drush fl | grep 'Overridden' | awk '{print ($(NF-2))}' | xargs -i drush fr {} -y`

:neckbeard:

Issue #1015, opened 2014-03-05
action page (prove it) – report back modal styles


DFurnes:
We need to make the report-back form look pretty.

[Mobile Annotated Mock](https://a0465d35-a-5d563b7e-s-sites.googlegroups.com/a/dosomething.org/dosomethinguxwiki/home/product-archive/campaign-template/Report%20Back%20Modals%20-%20Annotated%20Mocks.png?attachauth=ANoY7cp0dpLc2GnVA0S2cUKv4Pmm7yGEXhQLuRZVk552Jcjuz95EiKydpNQImwrnnmH-lVIAaSKYf_ClqzutxpTesj1VXEHKTtaw_6nC03Skzr98u_iKZq67ToITUd0R7DzBPVUopUE6iPVeagWJrBGbR4WK2yGc7Bi0_wWtjjEibMZZeE-2lYGvXsPsGCspk655tqe-7v_g8EijFNYxk_-2F3koPNNXtHx22rszirpkD4BivJULoCwfSBrU5TaSiAdTaALZij8PNOegP0lFid17h8pKO-AXp-3mFAfVpo8gt14s02x4il-MQ4pmu6tMsdnCjlMEzNdI&attredirects=0)

mikefantini:
@DFurnes for sponsored campaigns we should be pulling in the Alt background pattern for the report back modal.

DFurnes:
Yup, just put in a PR for that. Modal styles won't be fully matching the mocks until this issue is closed.

mikefantini:
got it. thanks!

mikefantini:
@DFurnes if you finish up the client side validation this guy would be good to pick up next

mikefantini:
related to this issue, we got a request that the font of the photo submission is too dark to read

happens when both before you've uploaded a photo and once you've uploaded it.

Is it possible to make this font the same color as the font of the other text on the form?

cc: @crystalruthbell 

https://trello-attachments.s3.amazonaws.com/52ea92c7ccdaf8415b8cab3a/53220dbee43c4570337c751e/72152e318c1032ec329907b71c6e4ebd/Screen_Shot_2014-03-13_at_3.58.09_PM.png

related to https://trello.com/c/y0DWDWrw

PR #1173, opened 2014-03-12
Modal sponsor bg

### LINKED ISSUE #1159: campaign (all) – clean up inline styles
Inline styles can now be processed in the Campaign Module. Move all remaining inline styles over, make sure they are name spaced under the `inline--` class and document those changes in the [wiki](https://github.com/DoSomething/dosomething/wiki).

DFurnes:
Hooks up scroll indicator to nav; adds in sponsor background pattern override; removes inline styles in favor of override classes in the header.

Working on #1015. Closes #1159.


aaronschachter:
Nice PHP :+1:

PR #1185, opened 2014-03-12
Zero downtime deploy & cleanup

desmondmorris:
This merge provides:
- **zero-downtime deployment** by moving the build task to an earlier point in the in pipeline
- **cleanup** default settings that come stock with capistrano install
- **production** tier deploy task for web1 in beta RPC env

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -3,27 +3,8 @@ lock '3.1.0'
 
 set :application, 'beta.dosomething.org'
 set :repo_url, 'git@github.com:DoSomething/dosomething.git'
-
-# Default branch is :master
-# ask :branch, proc { `git rev-parse --abbrev-ref HEAD`.chomp }
-
 set :branch, "dev"
-
-# Default deploy_to directory is /var/www/my_app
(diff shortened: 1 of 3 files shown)


DFurnes:
"zero-downtime deployment"

+9999. Was just wondering about this. :tada:

config/deploy.rb
```diff
@@ -3,27 +3,8 @@
...
-# set :deploy_to, '/var/www/my_app'
-
-# Default value for :scm is :git
-set :scm, :git
-
-# Default value for :format is :pretty
-set :format, :pretty
-
-# Default value for :log_level is :debug
-set :log_level, :debug
```
aaronschachter, line 22: Just checking, we don't need these guys anymore ^^ ? Did we ever?

desmondmorris, line 22: Nope see the comments above each.  Those values are defaults

cc: @aaronschachter

aaronschachter:
:+1:

mshmsh5000:
OK +1 value shipped

PR #1189, opened 2014-03-12
Updating SMS transition gate with correct IDs

jonuy:
Yea, what the title says. The IDs used before were just for testing.

This should be the last step to close out #894


aaronschachter:
Looks fine. Depending on how many more campaigns we run in the beta app with this code, we may want to make these variables so we don't need to deploy code for these changes.

jonuy:
Yea, that's a good call. If I can't move away from this by the next campaign, then we'll do that.

PR #1192, opened 2014-03-12
Staff-only content types

### LINKED ISSUE #1153: Rework Image, Fact node non-staff access
- [ ] Replace `hook_node_access` with `hook_init` to prevent non-staff from viewing Images or Facts
- [x] Update Image node type to default to published status
- [x] Update Fact node type to default to published status (in case we revisit using $vars instead of the $wrapper in Fact functions)

Currently `dosomething_user_node_access` is in place to deny access to non-staff users from viewing an image or fact node.

This causes issues when using the node's `$vars` to preprocess image nodereferences, because auth users aren't able to view the image.  So we get [weird behavior](https://github.com/DoSomething/dosomething/issues/1119#issuecomment-37252482) where staff sees images in a node but regular users cannot: https://github.com/DoSomething/dosomething/pull/1150#issuecomment-37339291

Either we always use the node->field / wrapper->field values directly instead of `vars` in our preprocess, or implement a different way to deny access to non-staff for Image nodes.  @blisteringherb mentioned that at MSL used a module to show a 404 if you visited the image node directly, although we'd want this to only happen for non-staff.  The fact and image nodes are useful for staff to see where they are referenced.

aaronschachter:
Resolves #1153

Uses `hook_init` to determine if a non-staff user is viewing a staff-only content type.  If so, redirects to front page.  Returning `drupal_not_found` i was resulting in "headers already sent" errors, so going the `drupal_goto` route for now as it's an easy fix.

Updates the Fact and Image node types to publish by default, to allow using `$vars` variables when preprocessing values from these staff-only content types.


angaither:
:+1: value added. ship it.

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -7,23 +7,31 @@
...
+    // If node type is staff only:
+    if (in_array($node->type, dosomething_user_get_staff_only_types())) {
+      // If logged in user is not staff:
+      if (!dosomething_user_is_staff()) {
+        // No node for you.  Redirect to front page for now.
+        drupal_goto();
+      }
     }
-    // Otherwise, no facts for you.
-    return NODE_ACCESS_DENY;
```
blisteringherb, line 23: Would calling drupal_access_denied() instead of returning the constant fix the half rendering issue that you were running into?

https://api.drupal.org/api/drupal/includes%21common.inc/function/drupal_access_denied/7

aaronschachter, line 23: Still same behavior with `drupal_access_denied` too.  The goto works fine for now.

screen shot 2014-03-13 at 11 56 39 am

PR #1220, opened 2014-03-13
Search Page Styles

### LINKED ISSUE #493: search results – style
Related to #492, need to make sure this is given presentable theming. Obviously, let's not get too jiggy for 0.2 -- something that works and is consistent with global styles should be sufficient.

mmwtsn:
Button styles should be handled on the server side.
## Desktop Styles

desktop
## Mobile Styles

mobile

Resolves #493


mmwtsn:
@blisteringherb All the input element needs is a `btn` class and this should be good to go.

mmwtsn:
search

Buttons styles are live. All set :v:

DFurnes:
COOL SHIT.

Issue #1071, opened 2014-03-07
Message Broker - Gather remaining merge_var values


deezone:
Need help finding the missing details in the related entity.

dosomething_signup_mbp_request()
      'call_to_action' => '??',
      'step_one' => '??',
      'step_two' => '??',
      'step_three' => '??',

dosomething_reportback_mbp_request()
      'event_id' => '??',
      'impact_verb' => '??',
      'impact_number' => '??',
      'impact_noun' => '??',

aaronschachter:
I don't mind taking this on as i'm pretty familiar with the fields we need, but what is `event_id` for `dosomething_reportback_mbp_request` ?

barryclark:
@aaronschachter @DeeZone Looks like this one is fixed?

aaronschachter:
Nope still open.  @DeeZone let's connect tomorrow, still don't know what `event_id` would be.

deezone:
@aaronschachter 

`event_id` is the NID for the campaign.

mikefantini:
'step_one' => 'field_pre_step_header',
'step_two' => is not a variable field. this will always be "Snap a Pic"
'step_three' => 'field_post_step_header',

PR #1227, opened 2014-03-13
Reportback helper functions

aaronschachter:
- Sanitizes user input
- Adds calculated properties to make for easy variable gathering in #1071


blisteringherb:
:+1:

PR #1235, opened 2014-03-13
Script to import signup data to the new world.

### LINKED ISSUE #1217: Import Signup Data
All signups from the 'old world' should be imported to the new world.
Once we are running both, signups are only stored in the new world.

Import data where `dosomething_campaign_signups.nid = 731234 && 731098`

angaither:
Fixes #1217

Add optional 'timestamp' arg to `dosomething_signup_insert` to preserve old world timestamps.
Saves all signups as new world nids


aaronschachter:
:+1:

PR #1237, opened 2014-03-13
Installing redirect module

### LINKED ISSUE #960: Install Global Redirect
Add Global redirect module into the build via make file.  

Does not need to be enabled in dosomething.info

Needed for indexing old nodes, and for #955

blisteringherb:
- Redirect module will be used to redirect dummy campaigns that
  are created in the new world.

Fixes #960

### DIFF
--- a/lib/profiles/dosomething/dosomething.info
+++ b/lib/profiles/dosomething/dosomething.info
@@ -44,6 +44,7 @@ dependencies[] = metatag
 dependencies[] = mobilecommons
 dependencies[] = module_filter
 dependencies[] = pathauto
+dependencies[] = redirect
 dependencies[] = search
 dependencies[] = strongarm
 dependencies[] = uuid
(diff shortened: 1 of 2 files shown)


aaronschachter:
Not Global Redirect? I thought we had decided on that this morning. Any reason in particular to use redirect?

blisteringherb:
I always get the two mixed up.  I installed global redirect and realized that it doesn't have the functionality to allow you to add a redirect from one node to another.  The path_redirect module had that, but now it's part of the Redirect module.

@aaronschachter

aaronschachter:
The main need for this module is redirecting from beta nid to legacy _path_ (but not actually to a different nid).  I take it this should do the trick?

blisteringherb:
Yes, the Redirect module will handle that.

aaronschachter:
:+1: value added ship it

PR #1238, opened 2014-03-13
Polaroid Style Fix

### LINKED ISSUE #1226: action page (do it) - image floating in bottom left hand corner
I re-uploaded the landscape image that goes in the Do It section on the H&M campaign. The ratio of that image is 16:9. 

Now, on staging.dosomething.org/campaigns/comeback-clothes as an authenticated user I'm seeing this in the Do It section. The image is not in it's proper place but is instead hiding in the bottom left hand corner.

image

mmwtsn:
## Polaroid Styles

polaroid-styles

Tested across Safari, Firefox and Chrome from `1440px` down to `320px`.
Fixes #1226


barryclark:
:+1:

PR #1236, opened 2014-03-13
MBP Request variables

aaronschachter:
Fixes #1071

PR #1241, opened 2014-03-13
Rb cleanup

### LINKED ISSUE #1231: Reportback form cleanup
- [x] Fix Notice: Undefined property: ReportbackEntity::$nid in campaign
- [x] Cleanup reportback form to use helper functions to display node values instead of using `$wrapper` as a param

aaronschachter:
- Fixes bug in #1236
  - `$entity->node_title`, `noun`, `verb` variables were not available at time of `mbp_request`, so reloading the reportback to get them.
- Fixes #1231 
  - Simplifies `dosomething_reportback_form` to use the calculated properties instead of sending node `$wrapper` as a param
- Fixes PHP Notice on the reportback confirmation page if recommended campaign does not have an image


angaither:
:+1:

PR #1242, opened 2014-03-13
Sets base url protocol dynamically

desmondmorris:
This merge makes sure the base_url is using the proper protocol.  The logic was shamefully stolen from this poor [soul](http://stackoverflow.com/questions/4503135/php-get-site-url-protocol-http-vs-https)

This will make sure our assets get served behind the correct protocols


aaronschachter:
:+1:

Issue #1068, opened 2014-03-07
Reportback confirmation page error


aaronschachter:
Error on the node confirmation page, but not consistently.

I can replicate this by truncating the signups table and then submitting reportback form.  Directed to the confirmation page (http://dev.dosomething.org:8888/node/12/confirmation). Refreshing and same thing.

This was occurring for @desmondmorris on staging.

PDOException: SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ')) AND (node.status = '1') AND (field_data_field_staff_pick0.entity_type = 'node' at line 2: SELECT field_data_field_staff_pick0.entity_type AS entity_type, field_data_field_staff_pick0.entity_id AS entity_id, field_data_field_staff_pick0.revision_id AS revision_id, field_data_field_staff_pick0.bundle AS bundle FROM {field_data_field_staff_pick} field_data_field_staff_pick0 INNER JOIN {node} node ON node.nid = field_data_field_staff_pick0.entity_id WHERE (field_data_field_staff_pick0.field_staff_pick_value = :db_condition_placeholder_0) AND (field_data_field_staff_pick0.deleted = :db_condition_placeholder_1) AND (node.nid NOT IN ()) AND (node.status = :db_condition_placeholder_2) AND (field_data_field_staff_pick0.entity_type = :db_condition_placeholder_3) AND (field_data_field_staff_pick0.bundle = :db_condition_placeholder_4) LIMIT 3 OFFSET 0; Array ( [:db_condition_placeholder_0] => 1 [:db_condition_placeholder_1] => 0 [:db_condition_placeholder_2] => 1 [:db_condition_placeholder_3] => node [:db_condition_placeholder_4] => campaign ) in field_sql_storage_field_storage_query() (line 585 of /vagrant/html/modules/field/modules/field_sql_storage/field_sql_storage.module).

aaronschachter:
I can confirm that commenting out the recommended query stuff makes the error go away.  Will investigate and add error handling to catch whatever is going on.

PR #1243, opened 2014-03-13
More cleanup

aaronschachter:
Missed in a form in #1241

Quick fix for #1068 to avoid WSOD (this only happens in the bizarre case that a user has no signups, which actually should never happen, but could happen if you're a dev directly visiting a campaign or reportback confirmation)


angaither:
:+1:

PR #1250, opened 2014-03-14
Signup admin views / unsignup form

### LINKED ISSUE #1129: Signup: Remove signup functionality
This would be super useful for us at least internally for testing.

I'd imagine some users would want this functionality too.

This issue is to create a way to remove a campaign signup for staff only.

aaronschachter:
Creates a `node/*/signups` tab on campaigns, which is a new view `node_signups` which displays all signups for a given node.  Built to faciliate testing signups, provide overview to campaigns team, and visible to roles with permission to edit a campaign.

Also resolves #1129 by creating a `node/*/unsignup` tab to allow a staff user to delete their own signup. Built for better internal signup testing. Only visible to staff, and if you're signed up for the node.


angaither:
:+1:

PR #1255, opened 2014-03-14
Official Rules Fix

### LINKED ISSUE #1147: action page (prove it) - only show official rules if field has a file uploaded
We should only display a link for the official rules if there is a file uploaded to the "Official Rules" field.

image

### LINKED ISSUE #1205: action page (prove it) - Tweak official rules link
- [x] move the Official Rules link down lower
- [x] make it the secondary link type (underlined, not bold)

mmwtsn:
## Issue

The campaign "official rules" link was printing regardless of whether or not a PDF has been updated because the `src` was checked on the template which always has a base URL value set.

Visually, it was the wrong color and not spaced correctly.
## Fix

This change creates a `$official_rules` variable to check against in place of the site's base URL, makes the link white and gives it some breathing room.

Fixes #1147
Fixes #1205


DFurnes:
:electric_plug: :elephant:

Issue #1253, opened 2014-03-14
action page (do it) - after tips are not appearing


mikefantini:
The after tips are not appearing on the action page:

After tips:
![image](https://f.cloud.github.com/assets/6330971/2421596/f2553168-ab83-11e3-9b52-b0839e122079.png)

Example: 
http://staging.beta.dosomething.org/campaigns/music-march-out

And this should be treated as this issue outlines on mobile: https://github.com/DoSomething/dosomething/issues/1200

mikefantini:
upload trello card when closed: https://trello.com/c/Hmii36am

aaronschachter:
Assigning to @blisteringherb

PR #1269, opened 2014-03-14
Adding a check to make sure faqs on campaigns are an array

### LINKED ISSUE #1260: Campaign PHP Warning
Warning: Invalid argument supplied for foreach() in include() (line 209 of /vagrant/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php)

blisteringherb:
- Fixing PHP warning

Fixes #1260


aaronschachter:
It needs the isset to get rid of the warning, but is this actually the right variable? Its printing FAQs but the heading says Tips. cc @mmwtsn @barryclark.  Seems linked to #1253 ... could turn this PR into a 2 for 1?

blisteringherb:
@mikefantini Confirmed that it should be $step_pre.

aaronschachter:
:+1:

Issue #1213, opened 2014-03-13
Legacy campaign redirects


aaronschachter:
For campaigns with signups, we'll use the Beta NID's pitch page to serve as the gate.  After signup (stored in the beta app), we redirect to the legacy nid's new path.

| Campaign | Legacy NID | Beta NID | Signups |
| --- | --- | --- | --- |
| Mind on my money | 731098 | 850 | yes |
| ~~PB Jam Slam~~ | ~~731234~~ | ~~955~~ | ~~yes~~ |
| Bully Text | 730939 | 832 | no |
| Pregnancy Text | 729027 | 843 | no |
| Would you rather | 731226 | 51 | no |

mikefantini:
@angaither @aaronschachter for Would you rather the legacy NID is 731226 and the beta NID is 51

aaronschachter:
Updated issue description, per @mikefantini PBJ content will be ported into the beta node, and we won't need to redirect back to legacy node.  We will still need to import the PBJ signups.

aaronschachter:
The redirects are working, however for some redirects, its directing to specifically to "node/730939".  If we care about this, we need to just these redirects to point to paths.  Otherwise ready to be closed.

blisteringherb:
The 3 non-login redirects are working properly.

mikefantini:
@aaronschachter want to just create an issue in the backlog for that and close this one out.

aaronschachter:
WE DID IT

PR #1270, opened 2014-03-14
Added support for legacy campaign redirects.

### LINKED ISSUE #1219: Stub node redirects
Campaign stub nodes should know the status of the current user and redirect as needed to the correct page. 

| Authenticated User | Action |
| --- | --- |
| signed up | taken to old world action page |
| not signed up | taken to old world 'join' |

| Anonymous User | Action |
| --- | --- |
| not signed up | stay on new world pitch |

Refs #1213

angaither:
Only handles mind on my money redirects.
Does not add the module to the .info file, we don't want this turned on by default.
Hardcoding mind of my money urls to point to production, this won't work in stage/qa.

Fixes #1219
Refs #1213

### DIFF
--- a/lib/modules/dosomething/dosomething_legacy/dosomething_legacy.module
+++ b/lib/modules/dosomething/dosomething_legacy/dosomething_legacy.module
@@ -1,0 +1,38 @@
+<?php
+/**
+ * @file
+ * Code for the DoSomething Legacy module.
+ */
+define('MOMM_NID', 850);
+define('MOMM_NID_LEGACY', 731098);
+/**
+ * Determines if a campaign is a legacy campaign.
+ *
+ * @param object $node
+ *  A node object.
+ * @return bool
+ *  Returns true if node is a legacy campaign, else false.
+ */
+function dosomething_legacy_check($node) {
+  if ($node->nid == MOMM_NID) {
+    return TRUE;
+  }
+  return FALSE;
+}
+
+/**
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_legacy/dosomething_legacy.module
```diff
@@ -0,0 +1,38 @@
...
+ * Handles redirects to the legacy app.
+ *
+ * @param int $nid
+ *  Old world nid.
+ * @param bool $signedup
+ *  True if user is signed up for campaign, else false.
+ */
+function dosomething_legacy_goto($nid, $signedup) {
+  if ($signedup) {
+    drupal_goto('https://www.dosomething.org/node/' . MOMM_NID_LEGACY);
```
aaronschachter, line 33: This works for now, but just checking: won't we need to change this to a specific path once this is live?  Since there wont be a node/731098 on dosomething.org once we launch.

Maybe it makes to store these 2 values as variables in an admin form, so we can make this change without deploying code.

aaronschachter, line 33: Disregard, didn't realize that by default paths default to legacy app.  This looks good to go :+1:

angaither, line 33: Should work because by default paths resolve to the legacy app.

PR #1271, opened 2014-03-14
Add signup data

### LINKED ISSUE #1252: Import additional signup data
Refs #1217 

Legacy: Mind on my Money collects a user's school upon signup, storing it in the `data` column of `dosomething_campaign_signups`. The campaign queries this field to see whether or not to display the School modal.

We'll need to store this in our beta signups db, in order to perform the same check.

The easiest way to do this for now is likely just adding a `data` column to our `dosomething_signups` table, and having the legacy module query for it.  

Long term we may want to implement a `dosomething_signups_data` to collect multiple data points in a signup process.

angaither:
Adds signup data column in `dosomething_signup`
Updated import script to get `data` from old world.

Fixes #1252
Fixes #/1261

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.install
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.install
@@ -47,3 +47,17 @@ function dosomething_signup_schema() {
   );
   return $schema;
 }
+
+/**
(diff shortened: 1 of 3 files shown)


aaronschachter:
:+1:

PR #1257, opened 2014-03-14
Ie8 snafu

DFurnes:
Miscellaneous fixes and cleanup made while trying to fix the IE8 issue in #1207.

PR #1274, opened 2014-03-14
Replaces local html5shiv reference with the CDN version

### LINKED ISSUE #1207: Global - Site-specific CSS not loading correctly
When attempting to access the campaign website in IE8 on Windows XP, the site does not render properly. ~~The three lines icon that accesses the sign in link on a mobile device is displayed instead of a sign in link and instead of accessing a sign in link it does nothing when clicked.~~ (See #1058.)

image

desmondmorris:
The original path to html5shiv seems to be borked.  This replaces it with the cdn version and order is restored.

This is a temp fix until we sort of the local path reference.  Though, I do not see an issue with using the cdn version either

Fixes #1207

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/template_hooks/preprocess_page/html5shiv.php
+++ b/lib/themes/dosomething/paraneue_dosomething/template_hooks/preprocess_page/html5shiv.php
@@ -3,12 +3,12 @@
 /**
  * Adds html5shiv script for IE8 users.
  *
- * Implements hook_preprocess_page(). 
+ * Implements hook_preprocess_page().
  */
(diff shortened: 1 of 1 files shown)


mmwtsn:
The original path is missing the `html5shiv/` directory. Does the local version when linked correctly?

`/bower_components/dist/html5shiv.js` should be `/bower_components/html5shiv/dist/html5shiv.js`

desmondmorris:
@mmwtsn, I just spoke with @DFurnes, and I think that is the case.  It will work with the local version when linked properly.

Close this once the better fix is in place

DFurnes:
Added a fix in #1257 referencing the correct Bower path. Thanks @desmondmorris! :+1:

PR #1277, opened 2014-03-14
This Could Be You!

### LINKED ISSUE #875: Campaign Template - Default Image in Prove It Gallery
If there is no image uploaded to the Prove It Gallery (which there won't be when we launch) we need to use a default image so that it doesn't break or just look shitty.

Keri is working on this image. I will upload it to this issue once I have it.

mmwtsn:
## Changes
- updates "prove it" placeholder image

prove-it

Resolves #875


lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -282,8 +282,7 @@
         </div>
       <?php else: ?>
       <div class="carousel-wrapper">
-        <?php //@TODO: Remove Trello-hosted placeholder ?>
-        <figure class="slide visible"><img src="https://trello-attachments.s3.amazonaws.com/53037337ba957ad54dc80486/53063914418de170762c28c0/62291511fcc8d763184a04a99db007b7/placeholder.jpg" alt="This could be you!" /></figure>
+        <figure class="slide visible"><img src="/profiles/dosomething/themes/dosomething/paraneue_dosomething/images/this-could-be-you.jpg" alt="This could be you!" /></figure>
```
angaither: @mmwtsn can we set this in the module and send it out as a themed image?

mmwtsn: Sure. The inline link is gross.

angaither:
:dancers:
:+1: 
:shipit: 
:confetti_ball: 
:congratulations:

PR #1283, opened 2014-03-14
Set the mind on my money nids in the settings file.

### LINKED ISSUE #1272: Old world/new world nid lookup.
Let's not be so gross.
Add a lookup table or a variable in settings.php so that you aren't hardcoding nids all over the damned place.

angaither:
Allows it to be used in other modules.
yyeeeee. 

Fixes #1272


aaronschachter:
:+1:

Issue #1254, opened 2014-03-14
Webfonts serving with wrong mime-type; Firefox disapproves


DFurnes:
Continuing discussion from #1133. We're having issues with webfonts periodically not loading on Firefox. I think the likely cause for this is that it seems Apache is serving our webfonts with the mime type `application/octet-stream` and Firefox is getting upset that this isn't strictly what it's looking for.

Adding the following types to `.htaccess` on local _seems_ to do the trick:

``` apache
AddType application/vnd.ms-fontobject .eot
AddType application/font-sfnt .otf .ttf
AddType application/font-woff .woff
```

We need to look into a fix for this both for local vagrant boxes & our production server.

desmondmorris:
Content types updated on the staging server.

This was completed by replicated the steps in #1284

desmondmorris:
Content types added to the production webhead.

This was achieved by adding the types in /etc/nginx/mime.types

desmondmorris:
@DFurnes  this should be good to go on all tiers.  Once you reprovision for vm should be good out of the box.  You can test staging and production like so:

**Staging**
`curl -I http://staging.beta.dosomething.org/profiles/dosomething/themes/dosomething/paraneue/paraneue/bower_components/neue/assets/fonts/proxima-nova/ProximaNova-SBold.woff`

**Prod**
`curl -I https://beta.dosomething.org/profiles/dosomething/themes/dosomething/paraneue/paraneue/bower_components/neue/assets/fonts/proxima-nova/ProximaNova-SBold.woff`

DFurnes:
Looks great! :+1:

PR #1284, opened 2014-03-14
Adds font mimetypes to vhost config in VM

desmondmorris:
Adds proper font mimetypes to local VM instances

This will require a VM provision to take effect

References #1254


DFurnes:
Looks great :thumbsup:

PR #1295, opened 2014-03-16
remove margin-top on the solution

### LINKED ISSUE #1203: action page (plan it) - tighten up margin under The Solution
There's an inconsistently big margin under The Solution. It should be the same as The Problem. 

image

barryclark:
Fixes #1203

Ty @aaronschachter


aaronschachter:
:+1:

PR #1299, opened 2014-03-16
Fixes padding for static content and 11 Facts pages.

DFurnes:
Before:
screen shot 2014-03-16 at 2 57 58 pm

After:
screen shot 2014-03-16 at 3 07 38 pm

Ahh, much better. :leaves:


barryclark:
Looks good, ty Dave :+1:

PR #1300, opened 2014-03-16
Action guide modals

aaronschachter:
Renders action guide content into hidden modal div's in the campaign template.  

Leaves the modal's header as blank for now.


angaither:
:+1:

PR #1317, opened 2014-03-17
Theme cleanup

DFurnes:
- Turns off preprocessing for CSS and JS, since `@import` rules break our IE8 polyfills. Our files are already minified and concatenated by our front-end build so this isn't too bad.
- Removes HTML5Shiv in favor of including it in Paraneue (see DoSomething/paraneue#50).
- Re-organizes the theme hooks by the feature that they're modifying, rather than by hook.


desmondmorris:
seems okay to me

PR #1318, opened 2014-03-17
Adding install file to update vars to replace reg. success msg

### LINKED ISSUE #1109: site registration - notification message copy
We need to update the copy of the notification message a user receives when they register for the site.

Language should be: 
"You've created an account with DoSomething.org."

Language currently:
site reg copy

blisteringherb:
- This should possibly be moved to settings.php in the future?

Fixes #1109


aaronschachter:
Is there any reason to not add to `settings.php` now?  It seems like a basic enough change.. unless something I'm misisng?

aaronschachter:
Sweet :+1:

PR #1320, opened 2014-03-17
Added a variable to set the last user saved.

angaither:
This will help if things error out in the middle of the script.
We should set this variable on qa-beta once we add the script to the `SELECT MAX(uid) FROM users`

@mshmsh5000


aaronschachter:
Smart, value added for sure. Does it make a difference that its being called before the `dosomething_user_is_thirteen` ?

angaither:
@aaronschachter nah, we just need the uid from the last user_save call

aaronschachter:
ok ship it :+1:

PR #1331, opened 2014-03-17
Adding 'q' param for apachesolr views to enable elevate queries

### LINKED ISSUE #1306: /campaigns prioritization is off
@blisteringherb Would You Rather? (nid 51) is designated a staff pick but it is getting buried at the bottom of the list

Also The Bully Text (nid 832) has a photo and is a staff pick and is still not being pulled into the view.

blisteringherb:
- The Solr QueryElevationComponent can be leveraged to bubble up
  a static list of nodes for a given query.  For /campaigns, 
  the desired results can't be achieved with boosts alone, so
  the list and order of campaigns that should show up first has
  been added to elevate.xml, and the elevate component has been
  enabled in solrconfig.xml.
- Elevate queries specified in elevate.xml need a 'q' param
  to operate properly.  apachsolr_views doesn't automatically include
  one.
- Since this is a static list that's not editorially controllable,
  this may need to be rethought in the future.

Fixes #1306

@mikefantini

### DIFF
--- a/lib/modules/dosomething/dosomething_search/dosomething_search.module
+++ b/lib/modules/dosomething/dosomething_search/dosomething_search.module
@@ -62,10 +62,17 @@ function dosomething_search_apachesolr_query_alter(DrupalSolrQueryInterface $que
     ),
   );
 
+  // Add a 'q' param to apachesolr_views queries to enable the
+  // elevate.xml to work properly.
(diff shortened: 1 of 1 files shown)


aaronschachter:
Seems ok :+1:  Agree about reworking later to make this editorially controllable.

PR #1335, opened 2014-03-17
Sponsor logo fallback

aaronschachter:
If there's no sponsor logo, show the sponsor name instead.

more money is more problems


angaither:
:+1:

PR #1338, opened 2014-03-17
Updates client-side validation attributes with new IDs

### LINKED ISSUE #1332: auth - confirm password not recognizing identical passwords
When creating an account I enter a password and then enter a perfect match in the Confirm Password field and it says that I haven't entered a match (even though I have).

DFurnes:
`edit-pass-pass1-pass2-pass-pass-pass--3`

Fixes #1332.


DFurnes:
Merging in since it seems to fix the issue. Kinda fishy though. :fishing_pole_and_fish:

DFurnes:
For future reference, most likely cause of the changed ID seems to be moving the auth modal code from Paraneue to Paraneue_DoSomething. So far doesn't seem to be any corner cases where this breaks.

PR #1350, opened 2014-03-17
Makes admin interface look a bit nicer.

DFurnes:
Before:
screen shot 2014-03-17 at 6 05 13 pm

After:
screen shot 2014-03-17 at 6 22 04 pm


angaither:
:mag: :ring:

PR #1351, opened 2014-03-17
Tells user password must be 6+ characters long.

### LINKED ISSUE #1288: auth - we should tell the user the password needs to be 6+ characters
In testing I saw lots of people enter passwords <6 characters. If we're going to throw the error message we should warn the user that it needs to be 6 characters or more.

DFurnes:
Fixes #1288.

screen shot 2014-03-17 at 6 26 40 pm


angaither:
:+1:

PR #1366, opened 2014-03-18
Remove uuid dependencies

aaronschachter:
Removes uuid from module dependencies.

I got an import to run by doing the following:

```
drush -y pm-uninstall uuid-features
drush -y pm-uninstall uuid
```

Oh great that should work!

Nope still an error upon login:

```
PDOException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'base.uuid' in 'field list': SELECT base.uid AS uid, base.name AS name, base.pass AS pass, base.mail AS mail, base.theme AS theme, base.signature AS signature, base.signature_format AS signature_format, base.created AS created, base.access AS access, base.login AS login, base.status AS status, base.timezone AS timezone, base.language AS language, base.picture AS picture, base.init AS init, base.data AS data, base.uuid AS uuid FROM {users} base WHERE (base.mail = :db_condition_placeholder_0) ; Array ( [:db_condition_placeholder_0] => aschachter ) in DrupalDefaultEntityController->load() (line 191 of /vagrant/html/includes/entity.inc).
```

Ok, so then here's the kicker.

I re-enabled uuid.

It took 45 minutes and eventually timed out at user 14k something out 15k something.

BUT. now I can login.


desmondmorris:
ship it

aaronschachter:
actually this PR may be completely unnecessary.  we should be able to just disable the relevant modules and uninstall uuid, and then follow steps outlined in PR.  used it mostly for sake of uninstalling uuid without killing everything else.

aaronschachter:
Not needed, although it'll be a good idea to clean up some of the dependency hell we're heading towards.

PR #1367, opened 2014-03-18
Removing _blank target

### LINKED ISSUE #1315: /campaigns - Comeback clothes opens up in new tab
on /campaigns Comeback Clothes defaults to open up in a new tab (it's the only campaign I could find that was doing this)

blisteringherb:
Fixes #1315


barryclark:
Sweeeeet! :+1:

PR #1375, opened 2014-03-18
Added field to home page

angaither:
once I added this field and features reverted 20 times it worked.


aaronschachter:
Glad to see that fixes -- but we have a field_subtitle field that exists already, can we add that instead of adding a new field / db table?

angaither:
@aaronschachter updated.

aaronschachter:
:+1:

PR #1407, opened 2014-03-18
Update the register validate hook.

angaither:
When an admin is editing another user account this fails.
Send in the user object of the form being edited when getting mobile
number.
yeee.

Also, I have no idea if there's a github issue for this.


aaronschachter:
No issue for this one. If all looks good when testing with new user registration too, let's ship it value added

Issue #1395, opened 2014-03-18
/user/login styles broken


mmwtsn:
![user-login](https://f.cloud.github.com/assets/1479130/2448539/7c87e65c-aea7-11e3-9dc3-71757f7052ad.png)

DFurnes:
Looking into this. I think we can just style the default node template since we're overriding everything else.

DFurnes:
There's plenty that's not _great_ about this, but it's a better "default state" until we can figure out some better template inheritance stuff.

![screen shot 2014-03-18 at 3 50 33 pm](https://f.cloud.github.com/assets/583202/2452838/9d0646ec-aed6-11e3-9b0a-57dd9d903966.png)

mmwtsn:
Looks grand. I'll send up a PR with a default wrapper width once that gets merged in.

PR #1409, opened 2014-03-18
Default Container Widths

mmwtsn:
Sets a default width to all Drupal containers then overrides that width for "designed" pages (e.g. campaigns, static content, etc.).

Described in #1395


mmwtsn:
This makes me cringe but it's a decent improvement over the current state of affairs until a proper solution can be put in place.

mmwtsn:
Controversial and out of date! Closing.

PR #1413, opened 2014-03-18
Properly floats images on campaigns again.

DFurnes:
Wasn't floating images on campaigns properly after we re-jiggered the nav to deal with unstyled pages.


barryclark:
:+1:

PR #1414, opened 2014-03-18
Added script to import mobile info from temp tables.

angaither:
Import mobile number from the temp tables in production

if a user has no number, check the temp table
if one is found, update the user.


aaronschachter:
Looks good!! :+1:

PR #1415, opened 2014-03-18
Wrap the http referer in url alias call.

### LINKED ISSUE #1408: auth - failed log in on pitch page takes the user to a 404
To recreate:
- go to https://www.dosomething.org/campaigns/project-shutdown  as an anonymous user
- click sign up
- give an invalid email/password combo

I've only see it happen on pitch pages not when trying to log in on the homepage.

angaither:
This will fix the redirects after login from node/whatever to real guys

Fixes #1408


aaronschachter:
:+1:

Issue #1419, opened 2014-03-19
Data Migration


angaither:
- [x] mobile number
- [x] birthdate
- [x] first name

angaither:
The mobile number script is already [created](https://github.com/DoSomething/dosomething/blob/dev/scripts/import-mobile-data.php)
Just need to know how/where to run it.

angaither:
Mobile numbers finished importing on Friday.

First name & birthdate are running now.

barryclark:
@mshmsh5000 This is running reeeeaaally slow! About 1 field import per minute, so won't be completed for about a million years. Any ideas why?

angaither:
Finished.

PR #1422, opened 2014-03-19
Added script to grab first name and birthday data.

angaither:
Yeah, this does stuff.

We need to import the `field_data_field_user_first_name` as a temp table. 

Refs #1419

### DIFF
--- a/scripts/import-all-the-things.php
+++ b/scripts/import-all-the-things.php
@@ -1,0 +1,69 @@
+<?php
+
+/*
+ * Script to import user data from old world into new world.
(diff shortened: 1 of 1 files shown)


aaronschachter:
Looks good :+1: 

One thought, which might be overthinking, but if the script interrupts it'll just grab everyone without a mobile or first name again. Could potentially be worth creating a field on the User to track "yes we've imported this person's data" to avoid having to run it again for the full set... but... might be overthinking it. I suppose it depends on how long it takes to run.

Issue #1420, opened 2014-03-19
Campaign Hero Image meta tags


aaronschachter:
Implement `drupal_add_html_head` to add the URL of a campaign's Cover Image node as the relevant `<link rel="image_src" href="[url]">`. Also confirm that FB / Twitter reads thsi tag, and that we don't have to add relevant `og` metatags.

_Context:_

Attempting to use Metatags module to set the Image metatag field as the "Campaign Cover Image" entityreference token doesn't work, as technically we need to specifiy a URL based on a field on the entityreference's target node nid.

![screen shot 2014-03-19 at 12 44 39 pm](https://f.cloud.github.com/assets/1236811/2462873/f28a00f4-af85-11e3-9116-095f08f36679.png)

aaronschachter:
Appears within the `<head>` on my local but not on staging or production.

aaronschachter:
Guess this just didn't make it for deploy on Friday.  The metatag is there.

PR #1423, opened 2014-03-19
dosomething_metatag module

### LINKED ISSUE #1308: campaigns - default meta tags
- Page Title: [node:title] | [site:name]
- Description: [node:call_to_action]. Find all the tips you need here.
- Abstract: [node:call_to_action]. Find all the tips you need here.
- Keywords: [node:tags]
- Google News Keywords: [node:tags]
- ~~Image: [node:hero_image]~~ Must be handled via #1420

aaronschachter:
Fixes #1308

Provides a home to export metatag data, and handle things like #1420.  

This approach allows us to prevent metatag from becoming a dependency in every content type, allowing us to easily drop it one day if we ever want to.

Seems like it's time to kill `dosomething_settings` too.


angaither:
:+1:

PR #1430, opened 2014-03-19
Don't select all the things from the user table.

angaither:
Because of 'reasons' this was being killed in production.
This fixes that.


aaronschachter:
"Totes" :+1:

PR #1431, opened 2014-03-19
Member Support, User Search

### LINKED ISSUE #1428: Member support role
Needs ability to edit user accounts.

### LINKED ISSUE #1429: User search view
Create a view with exposed filters to allow searching by
- UID
- Email
- Mobile

Which links to the user profile.

Grant access to staff.

aaronschachter:
- Fixes #1428 -- new role `member support` can access admin theme, menu and most importantly: administer users.
- Fixes #1429 -- creates `user_search` view at path `/admin/members`.  Query users by uid, email, or mobile and visible to users with "access administration menu" permission granted.

I have no idea why `features-update` is removing so many dependencies in `dosomething.info`, but I like it.  The only change I can think of is the change in #809... but this shouldn't hurt anything by not requiring these modules as dependencies, as we're only defining permissions..


aaronschachter:
Still in progress: Making updates so that member support can view any reportback, and also just change the view name to "Users" instead of "Members".

aaronschachter:
PR is updated and ready for review. Member Support role can administer users, and view any signup or reportback.

angaither:
As long as this doesn't break ds build since modules are turned on from the info file.

aaronschachter:
Oh, I typed the wrong thing. Dependences are being removed from `dosomething_user.info` not `dosomething.info`.

angaither:
:+1:

PR #1459, opened 2014-03-20
Bower non-interactive flag

DFurnes:
Prevents potential hang if newer version of Bower makes any prompts during CI task.


bin/ds
```diff
@@ -126,9 +126,6 @@ function build {
   echo 'Install composer dependencies'
   composer
 
-  echo 'Install bower dependencies'
-  bower
```
aaronschachter, line 130: Is this intentionally removed? Won't the `bower` function will never get called on `ds  build` with removing this?

aaronschachter, line 130: Nevermind - saw the comment about it being called from the  `grunt` function.  Looks ok :+1:

PR #1463, opened 2014-03-21
Make mobile data import go faster

angaither:
Run one query, only load user if you need to.
Added time to the script.

This ran overnight. yes. please.


aaronschachter:
:+1:

PR #1471, opened 2014-03-21
Updated the other firstname and birthdate script to run faster.

angaither:
Only loads user when needed.
Run one query instead of 3.


aaronschachter:
:+1:

PR #1474, opened 2014-03-21
Communications Team: admin menu access

aaronschachter:
Grants communications team access to the admin menu, theme, and pages to easily find the third-party optin form.

Moves definition of the role into `dosomething_user` to keep the permissions defined in one place.


angaither:
:+1:

Issue #1449, opened 2014-03-20
action page (know it) - design QA todos


barryclark:
- [x] Make the source font color darker so its more readable
- [x] Add in padding above sources
- [x] Hide sources

mmwtsn:
Hiding the actual sources for each source is going to be tricky as they are rendered with the full source citation in a single paragraph tag, e.g.:

```
<p>"Textiles." United States Environmental Protection Agency. Accessed February 24, 2014. http://www.epa.gov/osw/conserve/materials/textiles.htm</p>
```

We decided on a Markdown input field for sources as we wanted to give the campaigns team the ability to correctly format their sources with italics, links, etc. It does not appear that has been the case.

Perhaps it's worth revisiting the idea of building a source template into the field that would allow us to control how the markup is generated and prevent content authors from needing to format the sources manually.

**EDIT** This was a misunderstanding on my part. Leaving it here as it's relevant documentation of our fact sources formatting issues.

PR #1477, opened 2014-03-21
Know It QA

mmwtsn:
## Changes
- Adds top and bottom margin to the sources part of the "Know It" section

References #1449


DFurnes:
Looks fine. :dolls:

PR #1478, opened 2014-03-21
Campaign Navigation QA

mmwtsn:
## Changes
- Removes padding on campaign nav to address recent change in Neue `ul` styles

:palm_tree:

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/feature/campaign/_nav.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/feature/campaign/_nav.scss
@@ -32,7 +32,7 @@
     background: $light-gray;
     text-align: center;
     overflow: hidden;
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/scss/feature/campaign/_nav.scss
```diff
@@ -32,7 +32,7 @@
     background: $light-gray;
     text-align: center;
     overflow: hidden;
-    padding: 0.5rem;
+    padding: 0rem;
```
DFurnes: Should be using unitless `0`.

mmwtsn: :baby: Still getting the hang of these California Style Sheets.

Issue #1454, opened 2014-03-20
Mailchimp Grouping ID vars


aaronschachter:
- [x] Add `mailchimp_grouping_id` variable to signup admin form, to collect numeric grouping_id
- [x] Add `mailchimp_group_name` variable to signup admin form, to collect alpha group name.
- [x] Pass relevant `mailchimp_grouping_id` and `mailchimp_group_name` to `dosomething_user_mbp_request`
- [x] Update relevant values in production upon deployment

Mailchimp doesn't currently expose an Interest Group ID, only a Group Name.

To get `listBatchSubscribe` to execute, we need to pass through the relevant Grouping ID along with the GroupName: https://longreply.com/r/876a2dcc

aaronschachter:
Updated values on production.

Issue #1476, opened 2014-03-21
Mailchimp vars cleanup: group_id


aaronschachter:
Depends on #1454 

Once message broker is properly handling `mailchimp_grouping_id` and `mailchimp_group_name` variables, remove all references to the `mailchimp_group_id` variables / form elements and delete the variables from the db.

deezone:
See https://github.com/DoSomething/mbc-registration-email/pull/7 to expand to use `mailchimp_grouping_id` and `mailchimp_group_name`

deezone:
Latest release (0.1.4) of mbc-registrations-email (https://github.com/DoSomething/mbc-registration-email/compare/0.1.3...master) is processing all variations of mailchimp_group_id / group_name / groupings / etc. It should be through the backlog of entries in the queue in the next 24 hours. I'll start the cleanup at that time.

aaronschachter:
This was fixed in #1576

PR #1481, opened 2014-03-21
Mailchimp grouping_id and group_name vars

aaronschachter:
This PR handles all code necessary for #1454.  Will close out once we update the config form with the relevant `grouping_id` and `group_name` values.

Adds `grouping_id` and `group_name` variables into signup config so we can avoid hardcoding them in Message Broker.

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
@@ -30,6 +30,11 @@ function dosomething_signup_optin_config($form, &$form_state) {
     foreach ($staff_picks as $staff_pick) {
       $nid = $staff_pick['nid'];
       $title = $staff_pick['title'];
+      $prefix = 'dosomething_signup_nid_' . $nid;
+      $mailchimp_grouping_id = $prefix . '_mailchimp_grouping_id';
+      $mailchimp_group_name = $prefix . '_mailchimp_group_name';
+      $mobilecommons_id = $prefix . '_mobilecommons_id';
+
       // Create a fieldset for each campaign.
       $form[$nid] = array(
         '#type' => 'fieldset',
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
```diff
@@ -38,18 +43,24 @@ function dosomething_signup_optin_config($form, &$form_state) {
         '#collapsed' => TRUE,
       );
 
-      // Each campagin gets a mailchimp & mobile commons id.
-      $form[$nid]['dosomething_signup_nid_' . $nid . '_mailchimp_id'] = array(
+      // Each campaign gets a mailchimp grouping_id/group_name & mobilecommons id.
+      $form[$nid][$mailchimp_grouping_id] = array(
+        '#type' => 'textfield',
+        '#title' => t('MailChimp Grouping ID'),
+        '#default_value' => variable_get($mailchimp_grouping_id),
```
deezone, line 50: Should there be support for additional "grouping_id"s? or do we want to insist on there only being one? `variable_get($mailchimp_grouping_id)` suggests there will only be one grouping. I assume so, just flagging for consideration.

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -485,9 +485,13 @@ function dosomething_user_mbp_request($origin, $params = NULL) {
         'STEP_TWO' => $params['step_two'],
         'STEP_THREE' => $params['step_three'],
       );
-      // Check for mailchimp group:
-      if (isset($params['mailchimp_group_id'])) {
+      // Check for mailchimp grouping_id+group_name:
+      $mailchimp = isset($params['mailchimp_group_name']) && isset($params['mailchimp_grouping_id']);
+      // If set, or if soon to be depecrated group_id isset:
+      if ( $mailchimp || isset($params['mailchimp_group_id']) ) {
         $payload['merge_vars']['MAILCHIMP_GROUP_ID'] = $params['mailchimp_group_id'];
```
deezone: Please add a @todo to move these out of 'merge_vars' on the future cleanup pass that also removes the mailchimp_group_id.

lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
aaronschachter, line 50: There is support for additional `grouping_id`'s.  This function is being called foreach staff pick.  The `mailchimp_grouping_id` variable is defined in line 35, it will be unique for each staff pick campaign.

lib/modules/dosomething/dosomething_user/dosomething_user.module
aaronschachter: It does say "soon to be deprecated" ... ideally we should just set these params one time, ideally in the `dosomething_signup` function and not here in `dosomething_user`.

aaronschachter: Also there is an open issue #1476 to address the "todos", so we don't need to leave multiple "todos" throughout the function.

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -485,9 +485,13 @@ function dosomething_user_mbp_request($origin, $params = NULL) {
...
+        $payload['merge_vars']['MAILCHIMP_GROUPING_ID'] = $params['mailchimp_grouping_id'];
```
deezone:
```
$payload['merge_vars']['MAILCHIMP_GROUPING_ID'] = $params['mailchimp_grouping_id'];
```

needs to be:

```
$payload['mailchimp_grouping_id'] = $params['mailchimp_grouping_id'];
```
- remove merge_var as the group_name / grouping_id is not a merge_var value but rather just a required payload value
- make `MAILCHIMP_GROUP_ID` or any variable in the payload that's not a merge_var lower case. Upper case is used for merge_var values in Mandrill and MailChimp.

aaronschachter:
@DeeZone do the changes look ok?

deezone:
:+1:

PR #1486, opened 2014-03-22
Adds mobile for MailChimp submissions

### LINKED ISSUE #892: User Registration: MailChimp Subscription
When a user registers for the site (whether through general site reg or through signing up for a campaign) she should get subscribed to MailChimp group NewMembers2012.

We should pass through to the user's MailChimp profile:
- email (key)
- first name
- birthday
- mobile number

cc: @barryclark

deezone:
Fixes #892

Adds mobile value submitted as part of "user_signup" to `dosomething_user_mbp_request() -> message_broker_producer_request()`

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -359,6 +359,7 @@ function dosomething_signup_get_mbp_params($account, $node) {
     'email' => $account->mail,
     'uid' => $account->uid,
     'first_name' => dosomething_user_get_field('field_first_name', $account),
+    'mobile' => dosomething_user_get_field('field_mobile', $account),
     'event_id' => $node->nid,
     'campaign_title' => $node->title,
     'campaign_link' => url(drupal_get_path_alias('node/' . $node->nid), array('absolute' => TRUE)),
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -359,6 +359,7 @@ function dosomething_signup_get_mbp_params($account, $node) {
     'email' => $account->mail,
     'uid' => $account->uid,
     'first_name' => dosomething_user_get_field('field_first_name', $account),
+    'mobile' => dosomething_user_get_field('field_mobile', $account),
```
aaronschachter: This should use the specific function `dosomething_user_get_mobile` in case we ever need additional logic around handling mobile #'s, that this function would take care of.

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -463,6 +463,7 @@ function dosomething_user_mbp_request($origin, $params = NULL) {
   switch ($origin) {
     case 'user_register':
       $payload['birthdate'] = $params['birthdate'];
+      $payload['mobile'] = $params['mobile'];
```
aaronschachter, line 466: Looking forward to refactoring so we aren't setting these vars in two places.

deezone, line 466: Agreed! I assume the refactoring will be in a few days once all the code changes settle down?

aaronschachter:
ship it :+1:

PR #1487, opened 2014-03-24
Adds dosomething_campaign.theme.inc

aaronschachter:
This campaign module is turnin cray.  Moves all preprocess functions into a separate file to help organize.


DFurnes:
THIS IS SOME NICE PHP

Issue #1484, opened 2014-03-21
Signup Data Form


aaronschachter:
Build functionality to include a signup data form on a campaign.

Required for Comeback Clothes:
- Collect address in form and save to user - depends on #1026
- Add link to campaign, which opens form in modal
- Track whether or not the user has submitted the form

Tech specs: https://docs.google.com/a/dosomething.org/document/d/15NfJ8UoUfkjImPpjKFo-NIYxoQq8NNU2mT4Whmaj_Vc/edit

mikefantini:
@aaronschachter the link to trigger this modal for Comeback Clothes will live as a bullet point in Stuff You Need in the Plan It section... this should be the last bullet point (i.e. below all content a campaigns lead enters in that field and below any action guide they associate with the campaign)

aaronschachter:
@mikefantini Do we want to add some type of confirmation message upon submit?  Is there a generic message we can use, or do we need to add another field into the Signup Data Form config?

mikefantini:
Yeah just put this in an FFA I'm starting. We're going to need this to be specific to each form.

PR #1496, opened 2014-03-24
New dosomething_signup_data_form table

aaronschachter:
Groundwork for #1484

This is still a work in progress, but submitting as a PR now so it's not millions of lines of code to review.
- Adds a new db table `dosomething_signup_data_form` which stores configuration for a node's signup data form.
- Adds additional form elements into the campaign node edit form to allow editing per node.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -17,6 +17,7 @@ function dosomething_campaign_form_campaign_node_form_alter(&$form, &$form_state
   // If we're updating an existing node, add the custom settings fieldset.
   if (isset($form['nid']['#value'])) {
     _dosomething_campaign_form_extras($form, $form_state);
+    dosomething_signup_node_signup_data_form($form, $form_state);
   }
 }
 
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -17,6 +17,7 @@ function dosomething_campaign_form_campaign_node_form_alter(&$form, &$form_state
   // If we're updating an existing node, add the custom settings fieldset.
   if (isset($form['nid']['#value'])) {
     _dosomething_campaign_form_extras($form, $form_state);
+    dosomething_signup_node_signup_data_form($form, $form_state);
```
angaither, line 20: didn't catch this before, but why are we only doing this on existing nodes?

style: underscore for helper functions

aaronschachter, line 20: Yeah i can't tell if this is a helper function or not. It should be the same for both, but couldn't decide. I suppose its a helper function so I can make underscores.

It's only on existing nodes because its way easier to save all of this custom variables/records with a nid already set, vs trying to grab it somewhere in the form submit functions.  We needed to get the custom settings vars in for H&M so i didn't spend much time trying to find the newly inserted uid to make it work on a node/add (and it didn't seem worth it).

I'd like to change this function so that if the #nid is not set, it prints help text like "Custom settings will be available upon saving this node".

angaither:
:+1:

PR #1500, opened 2014-03-24
Set default first name for mb requests.

### LINKED ISSUE #1378: email - create default replacement for first name if none exists
If we don't have a first name for a user in MailChimp we should have some fall back so that we don't use "Hey _|TEXT:FNAME|_!

image

angaither:
Fixes #1378


aaronschachter:
:+1:

PR #1502, opened 2014-03-24
Adding Optimizely to ds.info and adding default OPT project ID

### LINKED ISSUE #1034: Add and configure Optimizely
- [ ] Add relevant version of Optimizely into `dosomething.info`
- [ ] Update drupal-org.make to download it
- [ ] Upon deploy, configure Optimizely with relevant settings for H&M

blisteringherb:
- There's a new project called "New Site Tracking" set up in the
  Optimizely admin.  We'll need experiments to be added to that.
- Adding the "New Site Tracking" project as the default project for
  Optimizely.

@Bladt
@mikefantini

Fixes #1034

### DIFF
--- a/config/settings.php
+++ b/config/settings.php
@@ -119,3 +119,5 @@ if (is_readable('sites/default/settings.'. $environment .'.php')) {
 $conf['locale_custom_strings_en'][''] = array(
    'Registration successful. You are now logged in.' => "You've created an account with DoSomething.org."
  );
+
+$conf['optimizely_id'] = '747623297';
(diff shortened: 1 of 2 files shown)


deezone:
For what it's worth, "New Site Tracking" sounds like a title that will get out dated quickly but your setup looks fine.

:+1:

config/settings.php
```diff
@@ -119,3 +119,5 @@
 $conf['locale_custom_strings_en'][''] = array(
    'Registration successful. You are now logged in.' => "You've created an account with DoSomething.org."
  );
+
+$conf['optimizely_id'] = '747623297';
```
aaronschachter, line 123: Is this sensitive? :pencil:

blisteringherb, line 123: I don't believe so.  It's the name of the js file that Optimizely uses.

PR #1503, opened 2014-03-24
Kill that drupal padding

DFurnes:
Some misc. design QA fixes.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/auth/login.php
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/auth/login.php
@@ -62,12 +62,12 @@ function paraneue_dosomething_form_alter_login(&$form, &$form_state, $form_id) {
 function paraneue_dosomething_login_after_build($form, &$form_state) {
   // Customize field elements.
   $form['name']['#title'] = 'Email address or cell number';
-  $form['name']['#attributes']['placeholder'] = 'Email address or cell number';
+  $form['name']['#attributes']['placeholder'] = '';
   unset($form['name']['#description']);
- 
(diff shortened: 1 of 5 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/auth/login.php
```diff
@@ -62,12 +62,12 @@ function paraneue_dosomething_form_alter_login(&$form, &$form_state, $form_id) {
 function paraneue_dosomething_login_after_build($form, &$form_state) {
   // Customize field elements.
   $form['name']['#title'] = 'Email address or cell number';
-  $form['name']['#attributes']['placeholder'] = 'Email address or cell number';
```
angaither, line 65: Really? no helps?

DFurnes, line 65: NOPE. YOU'RE ON YOUR OWN.

angaither, line 65: Yikes :scream: :unamused: :grimacing:

angaither, line 65: Oh, my point before I got so scared was if you don't need this stuff instead of setting it to an empty string, `unset($form['name']['#attributes']['placeholder']`

aaronschachter, line 65: or... just remove the line. I don't believe placeholder is included by default.

DFurnes, line 65: Good point. :dancer:

PR #1504, opened 2014-03-24
User signup data form

aaronschachter:
Refs #1484

Creates a user facing Signup Data form and places it onto the campaign template.  Only displays if the "Collect additional signup data" checkbox is checked in the Signup Data fieldset.

Still a work in progress but wanted to get this in to avoid any potential conflicts with campaign tpl changes.

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
@@ -100,3 +100,34 @@ function dosomething_signup_node_signup_data_form_submit(&$form, &$form_state) {
     ))
     ->execute();
 }
+
+/**
+ * Form constructor for a user signup data form.
+ *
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -100,3 +100,34 @@ function dosomething_signup_node_signup_data_form_submit(&$form, &$form_state) {
...
+    '#prefix' => '<p>',
+    '#markup' => $config['form_copy'],
+    '#suffix' => '</p>',
+  );
+  $form['submit'] = array(
+    '#type' => 'submit',
+    '#value' => $label,
+    '#attributes' => array(
+      'class' => array(
+        'btn',
```
angaither, line 127: the fenders will lova ya!

angaither:
:+1:

PR #1516, opened 2014-03-25
Track user signup_data_form submissions

aaronschachter:
Refs #1484

Tracks when a user submits a `signup_data_form` via new fields added to the `dosomething_signup` table.

Next up will be actually presenting the user with fields to collect account data within the `dosomething_signup_user_signup_data_form`, and saving to the user account upon submit.


angaither:
:+1:

PR #1520, opened 2014-03-25
Fixes #1519: Sets Redis_Cache as default cache class

### LINKED ISSUE #1519: Cache bins should default to Redis
https://api.drupal.org/api/drupal/includes!cache.inc/7

We need `cache_default_class` to be `Redis_Cache`, and `cache_class_cache_form` to be `DrupalDatabaseCache`.

mshmsh5000:


desmondmorris:
+1 value added

PR #1527, opened 2014-03-25
Sets up homepage as node.

DFurnes:
Sets up homepage node (rather than using `page--front.tpl.php` ratfuckery). Node already exists on production.


angaither:
WHY WAS I THE ONE WHO MADE THAT COMMIT

angaither:
WIZARD

DFurnes:
:cat2: :star2:  :wind_chime: :japanese_ogre:

PR #1533, opened 2014-03-26
Collect user address in signup data form

### LINKED ISSUE #1484: Signup Data Form

aaronschachter:
Closes #1484
- Implements only what we need for Comeback Clothes, which is just enter/confirm user address in a modal.
- Adds a patch for addressfield module to enable a sweet `addressfield` form element to do that.
- Adds documentation, listing what hasn't been implemented yet.

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/README.md
+++ b/lib/modules/dosomething/dosomething_signup/README.md
@@ -1,0 +1,117 @@
+# DoSomething Signup
+
+This module provides functionality for users to signup for a node type which
+implements it (currently only Campaigns).
+
+Technically, a signup is simply a record in the `dosomething_signup` table, 
+which stores the user uid, the node nid, and the timestamp of the signup.
+
+## View Signups
+
+Each campaign has a corresponding `node/[nid]/signups` path, which calls
+a `node_signups` view, displaying the signups for the given node.
+
+Users with the custom `view any signup` permission are able to access this view. 
+
+## Unsignup
+
+Members (authenticated users) are currently unable to remove their signup from 
+a node.
+
+Staff users (see `dosomething_user`) who signed up for a campaign are able to
+remove their signup record at `node/[nid]/unsignup`.  This is available for
+testing purposes.
+
+
+## Third Party Opt-In
+
+Currently we opt-in users to Mailchimp and Mobilecommons upon signup, depending
+on the node they are signing up for.
+
+Users with `administer third party communication` are able to set custom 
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -114,6 +102,8 @@ function dosomething_signup_node_signup_data_form_submit(&$form, &$form_state) {
  *   The node which the signup data form is rendered on.
  */
 function dosomething_signup_user_signup_data_form($form, &$form_state, $nid) {
+  global $user;
+  $account = user_load($user->uid);
```
angaither: you shouldn't have to load the user again here...

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -155,7 +163,25 @@ function dosomething_signup_user_signup_data_form_submit($form, &$form_state) {
...
 /**
- * Saves signup_form_data to the relevant signup.
+ * Saves signup_form_data to fields on the user.
+ *
+ * @param array $values
+ *   The values passed from a user signup_data_form submission.
+ */
+function dosomething_signup_update_user_data($values) {
+  global $user;
+  $account = user_load($user->uid);
```
angaither: same here

lib/modules/dosomething/dosomething_signup/README.md
```diff
@@ -0,0 +1,117 @@
+# DoSomething Signup
+
```
angaither, line 2: :+1:  awesssome

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
aaronschachter: You do need to load the user here, because the address field is not available on just the `global $user`.  The address field is used as the default value for the address element.

angaither:
oookkk :shipit:

aaronschachter:
@angaither Updated

Issue #480, opened 2014-01-25
Signup: Security / hidden nid


aaronschachter:
Similar to #479  -- Because a signup form is rendered on a node, we can use arg(1) to determine nid for a signup, instead of passing through nid as a hidden field, which a user could change via firebug.

This would be needed for both the dosomething_signup_form for logged in users, and the form_alter for user login/registration

barryclark:
@aaronschachter thoughts on whether this needs done for launch on tues?

aaronschachter:
Was hoping that setting the hidden nid in the login form to `#access = FALSE` would handle this, as it is in #1538 .... but it's adding that property into the `_dosomething_user_add_campaign_data` function results in no signup. Still investigating.

angaither:
Looks like we can set the access to false, but instead of checking for `$form_state['nid']` in the submit handler we check for `$form['nid']`

Here's a dpm of the $form after login.
![](http://f.cl.ly/items/341x282Z1D1n3k2r2D15/Screen%20Shot%202014-04-07%20at%2012.56.19%20PM.png)

PR #1538, opened 2014-03-26
Reportback security

aaronschachter:
Well this was a lot easier than I thought.


angaither:
KABLAM :+1: 
do we need to do this in other places?

aaronschachter:
yes.. would make sense to do for the signup nid, but unfortunately when i tried it, its not working for the hidden nid in the login form (mentioned this in #480).

PR #1547, opened 2014-03-27
Reportback IP address

### LINKED ISSUE #1541: Reportback: Add IP column
Track IP to see who's submitting reportbacks from where.  Will allow us to track what is submitted via mobile.

aaronschachter:
Fixes #1541

Store IP address in the `dosomething_reportback_file` table to track where images are submitted from, and also store in the `dosomething_reportback_log` table to track every update.

Also fixes the reportback tabs on the view reportback page (admin only)

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc
+++ b/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc
@@ -88,6 +88,7 @@ class ReportbackEntity extends Entity {
       ->fields(array(
         'rbid' => $this->rbid,
         'fid' => $fid,
+        'remote_addr' => ip_address(),
         ))
       ->execute();
   }
@@ -118,6 +119,7 @@ class ReportbackEntity extends Entity {
           'why_participated' => $this->why_participated,
           'files' => implode(',', $fids),
           'num_files' => count($fids),
+          'remote_addr' => ip_address(),
         ))
         ->execute();
     }
--- a/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc
+++ b/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc
@@ -88,7 +88,7 @@ class ReportbackEntity extends Entity {
       ->fields(array(
         'rbid' => $this->rbid,
         'fid' => $fid,
-        'remote_addr' => ip_address(),
+        'remote_addr' => dosomething_helpers_ip_address(),
         ))
       ->execute();
   }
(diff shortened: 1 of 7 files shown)


lib/themes/dosomething/paraneue_dosomething/templates/reportback/reportback.tpl.php
```diff
@@ -2,19 +2,6 @@
...
-    <?php if (isset($edit_link)): ?>
-    <li><?php print $edit_link; ?></li>
-    <?php endif; ?>
-    <?php if (isset($delete_link)): ?>
-    <li><?php print $delete_link; ?></li>
-    <?php endif; ?>
-  </ul>
-  <?php endif; ?>
-</nav>
-
```
angaither, line 17: just making sure... is supposed to be removed?

lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc
```diff
@@ -88,6 +88,7 @@ class ReportbackEntity extends Entity {
       ->fields(array(
         'rbid' => $this->rbid,
         'fid' => $fid,
+        'remote_addr' => ip_address(),
```
angaither: Might want to check with @mshmsh5000 as this can be tricky because varnish only likes to expose it's own IP

lib/themes/dosomething/paraneue_dosomething/templates/reportback/reportback.tpl.php
aaronschachter, line 17: Yep. There are fixes in the `dosomething_reportback_menu` hook which makes these unnecessary now.

lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc
desmondmorris: Also, apache and nginx will give different results as well.  I have a solution for this in the sms workflow somewhere.  I will post when I find it

desmondmorris: The basis: https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_sms/dosomething_sms.module#L167-L170

A possible replacement solution:

`'remote_addr' => isset($_SERVER['HTTP_X_FORWARDED_FOR']) ? $_SERVER['HTTP_X_FORWARDED_FOR' : ip_address()`

aaronschachter:
@desmondmorris @angaither updated to use a new `dosomething_helpers_ip_address` function.

angaither:
:+1: kaplow. value added. ship it. :shipit:

PR #1548, opened 2014-03-27
Refactored file saving code for SMS report backs to maintain file extension

### LINKED ISSUE #1526: Reportback file extensions
The `dosomething_reportback_get_file_dest` function is not consistently writing image files with extensions.

extension sadness

jonuy:
Previous logic was to:
- download the image
- save the image
- figure out where it should really live
- then move the file there

Logic now simplified to:
- figure out where the file should live
- download and save it there

`basename($pictureUrl)` in particular is what allows us to get the file extension in the first place. Then passing `$fileDest` into `file_save_data()` lets us keep it. 

Fixes https://github.com/DoSomething/dosomething/issues/1526

cc: @angaither


angaither:
:+1: awesome.

PR #1549, opened 2014-03-27
Static content  misc

### LINKED ISSUE #1515: static content - link / button and misc styling
- [x] buttons should not have underline on rollover
- [x] text wrapping on top text
- [x] update section subheader to match campaigns template subheaders - and lose the purple

*removed the underline links item to separate out into issue for Dave on Neue.

weerd:
Need to double-check the confirmation and explore-campaigns pages since I'm not sure how to best view them on local. The explore-campaigns page seems to be generated by Views, but I don't see the template being used in Paraneue_DS.

Resolves #1515 (most of it anyhoo).

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign--pitch.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign--pitch.tpl.php
@@ -2,7 +2,7 @@
   <header class="header <?php print $classes; ?>">
     <div class="meta">
       <h1 class="title"><?php print $title; ?></h1>
-      <p class="cta"><?php print $cta; ?></p>
+      <h2 class="subtitle"><?php print $cta; ?></h2>
 
       <?php if (isset($end_date)): ?><p class="date"><?php print $end_date; ?></p><?php endif; ?>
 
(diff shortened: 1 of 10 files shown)


lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign--pitch.tpl.php
```diff
@@ -2,7 +2,7 @@
   <header class="header <?php print $classes; ?>">
     <div class="meta">
       <h1 class="title"><?php print $title; ?></h1>
-      <p class="cta"><?php print $cta; ?></p>
+      <h2 class="subtitle"><?php print $cta; ?></h2>
```
mmwtsn, line 5: Semantically, this makes sense but it brings the front-end terminology out of sync with the field/variable names. 

Should we update this in the module as well?

lib/themes/dosomething/paraneue_dosomething/templates/node--static_content.tpl.php
```diff
@@ -5,7 +5,7 @@
       <header class="header">
         <h1 class="title"><?php print $title; ?></h1>
         <?php if (isset($subtitle)): ?>
-    <h3 class="subtitle"><?php print $subtitle; ?></h3>
```
mmwtsn, line 8: :pineapple: Approved.

weerd:
We could update the module as well (at least the variable), but updating the fields in the DB might get hairy...

aaronschachter:
agree, easy enough to just change the preprocess vars to match.

weerd:
Cool. I can add another issue for me to update the preprocess vars in the .module :)

DFurnes:
:dart:

PR #1554, opened 2014-03-27
Added first & last name to address collection.

### LINKED ISSUE #1544: Signup Data Form: First Name and Last Name
If a signup data form is set to "Collect user address", values for the user first_name and last_name fields should be pre-populated / collected as well as the mailing address, and saved to the user upon submit.

angaither:
Fixes #1544


aaronschachter:
:+1:

PR #1556, opened 2014-03-27
Zendesk error catching

aaronschachter:
Abstracts Zendesk user search/creation into functions with try / catch statements.

If you create a new account, submit the zendesk form, and then immediately submit the form again, you'll get a nice error message instead of the WSOD style:

ahh nice error message

I've replicated the error still, but after the 4th submisison (within a minute), the user is then found and a new ticket created.

So per a suggestion I made in https://github.com/DoSomething/dosomething/issues/1388#issuecomment-38729401 ... we could create a zendesk user upon user registration, to avoid this potential lag issue.  My thoughts here are that it's going to be rare that a user submits the form immediately after submitting once, and not worth the overhead of creating a zendesk user for each user registration.  Thoughts?


blisteringherb:
:+1:

PR #1557, opened 2014-03-27
Video Collection - hook_field_collection_is_empty_alter

aaronschachter:
Implements `hook_field_collection_is_empty_alter` to prevent a dummy value being stored for `field_video` when no Video ID has been entered.


angaither:
:+1: looks good.

PR #1567, opened 2014-03-28
Action Page Alignment

### LINKED ISSUE #1446: Action page (structure) - Grid layout updates
- [x] Line up Steps and body text onto the grid, same as static content. 2 column left margin, rather than 3. 
- [x] Make gutter larger between the 2 column content layout (e.g. The Problem & The Solution)

mmwtsn:
## Changes
- Sets all shift values for left-most columns from `3` to `2`
- Reduces width of `<nav>` element to account for new layout
- Removes `<li>` style overrides now that default styles are set in Neue
- Removes left padding from `.banner` and `<footer>` to prevent misalignment with the grid columns
- Adds one-off breakpoint to remedy `<nav>` width issue to allow for a full-width navigation earlier
## Rationale

Static content pages across the site span `12` of `16` columns with a two column gutter on either side to center the content. Campaign Action Pages, however, were designed with a three column gutter. Having only one template laid out with a three column gutter was inconsistent with the site.
## Screenshot

new-layout

Resolves #1446

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
@@ -27,6 +27,7 @@
   // immediately before each section.step
   .banner {
     margin: 0;
+    padding: 0.1em 0 0; // Override default banner padding to prevent left indent
 
     span {
       @include section-padding;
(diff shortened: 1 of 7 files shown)


weerd:
:+1: looks good to me!

PR #1570, opened 2014-03-28
Removing PHP short tags from 2 templates.

### LINKED ISSUE #1565: Eliminate PHP short tags
The PHP config on QA is sensitive to this. Instead of changing the config, I'd rather fix the few outliers. I see [just one in this template](https://github.com/DoSomething/dosomething/blob/dev/lib/themes/dosomething/paraneue_dosomething/templates/node--static_content.tpl.php#L61):

screen shot 2014-03-28 at 12 30 34 am

weerd:
Resolves #1565


mshmsh5000:
Yay! +5

Value: RAD

weerd:
I'm gonna wait till @aaronschachter gives the ok on merging, after he's out of Features hell :)

PR #1576, opened 2014-03-28
Fixes #893, mailchimp payload cleanup

### LINKED ISSUE #893: Staff Pick Campaigns: MailChimp Group Subscription
Need the ability to subscriber users to a specific MailChimp Group for specific staff pick campaigns.

I.e. I sign up for H&M and I am subscribed into the H&M MailChimp Group.

This should happen whether I am a new or returning user.

See #892 for related MailChimp subscription issue.

deezone:
Fixes #893

Removes mailchimp_group_id from `$payload` values sent to Message Broker via `dosomething_user_mbp_request()` -> `message_broker_producer_request()`

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -494,6 +494,12 @@ function dosomething_user_mbp_request($origin, $params = NULL) {
       break;
     case 'campaign_signup':
       $payload['event_id'] = $params['event_id'];
+      // Check for mailchimp grouping_id+group_name:
+      $mailchimp = isset($params['mailchimp_group_name']) && isset($params['mailchimp_grouping_id']);
+      if ( $mailchimp ) {
+        $payload['mailchimp_grouping_id'] = $params['mailchimp_grouping_id'];
+        $payload['mailchimp_group_name'] = $params['mailchimp_group_name'];
+      }
       $payload['merge_vars'] = array(
         'FNAME' => $params['first_name'],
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -503,16 +509,6 @@ function dosomething_user_mbp_request($origin, $params = NULL) {
...
         'STEP_THREE' => $params['step_three'],
       );
-      // Check for mailchimp grouping_id+group_name:
-      $mailchimp = isset($params['mailchimp_group_name']) && isset($params['mailchimp_grouping_id']);
-      // If set, or if soon to be depecrated group_id isset:
-      // @todo: Remove check for group_id when ready for cleanup.
-      if ( $mailchimp || isset($params['mailchimp_group_id']) ) {
-        //@todo: Remove group_id when ready for cleanup.
-        $payload['mailchimp_group_id'] = $params['mailchimp_group_id'];
-        $payload['mailchimp_grouping_id'] = $params['mailchimp_grouping_id'];
```
aaronschachter, line 513: Dont we need these?

deezone, line 513: I move the code we want to keep to ln 497. I'd like keep 'merge_vars' at the end of the array as it's easier to look at the payload values when debugging when looking at the payload values in the queues.

aaronschachter, line 513: ohhhh i see.

aaronschachter:
If all tested and good to go, ship it :+1:

Issue #1511, opened 2014-03-25
static content - call to action


barryclark:
- [x] Large button on the CTA
- [x] .cta padding-bottom: 25px; .cta h3 { margin: 25px 0 10px 0; }
- [x] CTA on bottom of the page should have a bottom horizontal rule and spacing afterward

weerd:
Ended up doing 35px padding top and bottom of CTA and removing top margin on h3. No need to address underline in new style for CTA. Looks good!

PR #1577, opened 2014-03-28
Static content  gallery

### LINKED ISSUE #1513: static content - gallery
- [x] Gallery wrapper background #eee
- [x] Gallery h3 headers #444 18px bold

weerd:
Resolves #1513 and #1511

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_static.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_static.scss
@@ -39,6 +39,17 @@
     }
   }
 
+  .gallery-wrapper {
+    background-color: $light-gray;
+    // background-color: lighten($light-gray, 10%);
+  }
+
+  .cta-wrapper + .gallery-wrapper {
+    // Since cta's have a bottom border, if they preceed a gallery with a background,
+    // the border blends into the gallery background. Adds a margin to give it some room.
+    margin-top: 1.25rem;
+  }
+
   .intro {
     // @TODO: ಠ_ಠ
     margin-top: 20px;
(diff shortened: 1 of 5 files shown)


DFurnes:
Looks good! Two thoughts –

lib/themes/dosomething/paraneue_dosomething/scss/content/_static.scss
```diff
@@ -64,12 +69,11 @@
 
   .cta {
     text-align: center;
-    border-top: 2px solid $light-gray;
-    border-bottom: 2px solid $light-gray;
-    padding-bottom: 35px;
+    padding: 2.1875rem 0;
```
DFurnes, line 77: Should we round this to an even `2rem`?

lib/themes/dosomething/paraneue_dosomething/scss/content/_static.scss
```diff
@@ -64,12 +69,11 @@
...
 
     h3 {
       color: $copy-color;
+      margin: 0 0 0.625rem;
```
DFurnes: Same here. Thoughts on `0.5rem`?

lib/themes/dosomething/paraneue_dosomething/scss/content/_static.scss
weerd, line 77: Yeah, I was using values to match specific PX values requested in the issue. Would you guys prefer I round in the future if it make sense and feels like enough spacing?

DFurnes, line 77: Yeah, I'd like to avoid styling with specific pixel values wherever we can. Unless it breaks the design, in which case it may be worth coding with those pixel values just to indicate that its a precise value.

weerd:
Closing this PR due to mix up with rebase and inheriting other commits. :-1:

PR #1580, opened 2014-03-28
Zendesk Tickets: groups and priority

### LINKED ISSUE #928: Zendesk: Ticket Groups
- [x] Create admin form to configure the Zendesk group_id for each Cause term tid
- [x] Create Zendesk ticket with the group_id of a Campaign's Primary Cause
- [x] For staff picks, use priority "high" instead of "normal".

aaronschachter:
Closes #928

Creates admin form to configure each Cause term's with a corresponding Zendesk group_id.

Because each group is created with the same name as the Cause term, we can get away with populating the variables in a `hook_update_N` function.  

The admin form exists for if/when the names or groups should ever change.  Also validates entered values against the Zendesk API to make sure the entered group_id value is valid.

Creates ticket with the group_id if it exists.  Also if the form is rendered on a staff pick, the ticket is given high priority.


angaither:
:+1:

PR #1588, opened 2014-03-31
Update default local settings to use environment vars

### LINKED ISSUE #1585: /campaigns (staging) - HTTP 0; Request failed: Connection timed out
Can't access /campaigns on staging site (everything is fine in prod). Preventing me from showing an example of a bug in prod that I don't want visible to the public.

Getting this error message:

image

blisteringherb:
- Using environment variables to avoid putting the production solr
  IP address into the public repo.  If a rebuild happens in production
  this may require a restart of the php fpm pool to reload the settings.

Fixes #1585


aaronschachter:
Re: #1585, do the relevant variables need to be removed from the `dosomething_search` feature ?

PR #1590, opened 2014-03-31
Updated the import script again.

angaither:
- Add a var for the last saved
- Duh, only get users above that uid.

Refs #1419


aaronschachter:
:+1:

PR #1591, opened 2014-03-31
Paraneue paranoia

### LINKED ISSUE #1529: Merge Paraneue into main repo
Since Neue is already handling most of our style abstraction, we're thinking of merging Paraneue and Paraneue_DS to simplify workflow.

Looking at where our views are actually being generated, Paraneue is basically responsible for two partials (`html.tpl.php` and `page.tpl.php`) and the Drupal defaults, and everything else comes from the app-specific theme. Since we'll likely need to override the site chrome for international sites as well, it seems to basically boil down to just one template that's actually getting inherited.

With this in mind, we came to the conclusion that Neue itself is the shared "starter" that all of our child themes should be inheriting from. This also means we can keep all of our dependencies in one place, which makes the system as a whole a bit easier to grasp and keep tabs on (for example, with more tightly locking Bower versioning to releases).

DFurnes:
Merges Paraneue into Paraneue_DoSomething. Closes #1529.


DFurnes:
:cactus:

PR #1595, opened 2014-03-31
Require that js

DFurnes:
Sets up RequireJS for JavaScript asset pipeline. This sidesteps some weird issues we were having with Browserify compilation failing on different Vagrant boxes, and gives us some more flexibility in the future with how we pull in JS and assets.

modules oh my


weerd:
@_@ Oh man, it's happening.

PR #1602, opened 2014-03-31
Restores Uglify task for production build.

DFurnes:
Restores uglify task for production build. Accidentally got turned off when hooking up RequireJS.


weerd:
:+1:

PR #1606, opened 2014-03-31
Facebook OG Image

aaronschachter:
Adds og:image tag into the `<HEAD>` to make the relevant cover / hero image appear when sharing a campaign or fact page on the good Book.


angaither:
:+1:

PR #1608, opened 2014-04-01
RequireJS dev environment

DFurnes:
Sets up dev build to pull in RequireJS modules on-demand through the loader, rather than requiring a re-build of the `app.js` file on every change. :rocket:


mmwtsn:
:package:

PR #1609, opened 2014-04-01
Updates to 11 Facts to styles the CTAs and extract all static content...

### LINKED ISSUE #1587: Fact Page CTA alignment
Looks weaksauce.

example: http://staging.beta.dosomething.org/node/604

screen shot 2014-03-31 at 12 43 40 pm

weerd:
... CTA styling into a mixin for the time being.

Resolves #1587

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_fact-page.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_fact-page.scss
@@ -41,6 +41,10 @@
     }
   }
 
+  .cta-wrapper {
+    @include cta();
+  }
+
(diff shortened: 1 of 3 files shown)


DFurnes:
I'd argue this should be a class rather than a mixin, but I'm happy to hear arguments otherwise.

Looks fine to merge for now.

weerd:
Yeah, def. I made it a mixin for the time being and added a TODO. I'm in the same boat, but since we already need to address the header and abstract it, maybe we can discuss the CTA as an item as well or what makes the most sense.

DFurnes:
+1, makes sense.

PR #1615, opened 2014-04-01
Updates markup with new chrome classes in Neue 3.2.x.

DFurnes:
:warning: Do not merge yet. Requires Neue 3.2.x.

Removes wrapping `footer--wrapper` div and uses properly namespaced `.chrome--footer` and `.chrome--wrapper` class names. Also moves `main` tag to only contain the `$content` variable since that makes more semantic sense.


DFurnes:
Closing this since we reorganized `page.tpl.php` today and merge conflicts would be a pain.

PR #1623, opened 2014-04-01
Action Guide Styles

### LINKED ISSUE #1453: action guides - styling needed
- should have same modal header treatment as other modals (i.e. black band across the top with x to close modal)
- header content for this modal should come from the title field on the action guide content type
- for the rest of the styling, originally we talked about using same patterns as static content for action guides, which I think makes sense because they share the same fields that we display to the user
- we should keep the "Back to main page" as a second way to exit the modal
- need to make sure markdown works
- campaigns that have action guides include https://www.dosomething.org/campaigns/food-forward, https://www.dosomething.org/campaigns/school-standstill, https://www.dosomething.org/campaigns/free-pee-petition, and https://www.dosomething.org/campaigns/oppression-monopoly

mmwtsn:
## Changes
- Adds `title` with `.banner` class
- Misc. template and style updates

Resolves #1453
Requires DoSomething/neue#234


weerd:
Righty-o! :+1:

PR #1625, opened 2014-04-02
User profile

weerd:
Would appreciate it if a bender could take a look at this since I'm using hook_theme() to enable use of partials for outputting markup. @aaronschachter @angaither @desmondmorris  :smiley_cat:


lib/themes/dosomething/paraneue_dosomething/templates/user/page--user--edit.tpl.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+/**
+ * Generates site-wide User Edit page chrome.
```
aaronschachter: Why do we need a whole new page.tpl.php here?

weerd: To customize the page markup around the user form that Drupal's outputting. Plus, this starts to set the foundation to help w/ future needs for unique page types/markup instead of having a bunch of `if` statements in a single page.tpl.php.

aaronschachter: but why do we need a separate page markup for the user profile?

weerd: Oh, it's for the User Edit form page (not the user profile view). Drupal outputs it in the $content of the page tpl, but I need to be able to add markup around the form to get it to work w/ the grid and structure we've established.

DFurnes:
Looks good. :thumbsup:

lib/themes/dosomething/paraneue_dosomething/templates/user/page--user--edit.tpl.php
DFurnes: I'm not sure if it'll be necessary here, but this will help us with the issue we've been facing where we have separate chrome styling on designed pages vs. "untouched" default Drupal pages.

weerd: After some some useful discussion we've decided to hold off on the styling for the edit profile view and address it after, so we can address all the default page styling globally across the site.

thanks @aaronschachter

lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
```diff
@@ -6,5 +6,15 @@
...
 
+  $vars['navigation'] = theme('navigation');
+  $vars['footer'] = theme('footer');
+}
+
 
+/**
+ * Implements theme_preprocess_user_profile().
+ */
+function paraneue_dosomething_preprocess_user_profile(&$variables) {
```
angaither, line 17: would rather remove this stuff if we aren't using it yet.

aaronschachter, line 17: agree

weerd:
@angaither @aaronschachter I missed the above message earlier. Yeah, I agree. Imma remove it since it's not being used for the time being and likely won't need to ;)

PR #1626, opened 2014-04-02
Campaign PSA

### LINKED ISSUE #844: action (know it) – switch facts and video/placeholder
For the initial launch we are going with facts/sources in the left column of the campaign page in lieu of a YouTube video or PSA placeholder image.

**EDIT:** except on sponsored campaigns for which Fantini has requested that we keep all links in the rightmost column of the "Know It" section to allow for a video/placeholder image.

mmwtsn:
## Changes
- Calls a `~16:9` PSA/PSA replacement image
- Removes the default YouTube embedded video chrome to display cleaner videos
- Adds conditional logic to display modal links depending on the existence of a PSA/PSA replacement image
- Adds styles for responsive embedded videos
## Notes
- Modal links should included as a partial to DRY up the template
- `!important` should not be necessary here—we haven't gotten around to refactoring the indenting of content
- `first` and `second` layout modifiers should be replaced with Bourbon's `omega`
- `.col` should not be used as a selector—`.solution` / `.problem` should be moved higher up!
### PSA

video
### PSA Fallback Image

image
### Neither

neither

Closes #844

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -215,14 +215,14 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
   if (!empty($vars['field_video'])) {
     $vars['psa'] = theme('dosomething_video_embed', array(
       'field' => $wrapper->field_video->value(),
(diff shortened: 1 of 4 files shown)


DFurnes:
Cool :100: :pencil: :apple: :cookie:

PR #1631, opened 2014-04-02
Removes newline from end of file.

DFurnes:
tsk tsk


mmwtsn:
Someone call the lint man!

PR #1632, opened 2014-04-03
Reportback log fix

aaronschachter:
- Fixes bug in reportback log records for mobile submissions, where a `global $user` is not present.
- Updates records with uid == 0, using the uid value stored on the reportback.
  - Using db_query instead of db_update because db_update doesn't cleanly support joining tables, and this should be a one time jam. http://drupal.stackexchange.com/questions/53774/drupal-db-update-with-joins


jonuy:
jam on. :rocket:

PR #1639, opened 2014-04-03
 Allow users to remove campaign signups.

### LINKED ISSUE #1638: User Profile: Remove signups
Allow users to remove own signups from their profile.
- [x] remove row from signup table
- [ ] remove from mailchimp 
- [ ] remove from mobile commons

angaither:
:warning:
- add link in 'signups' view 
- update access callback, allow non-staffers to have access.

Fixes #1638

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -429,8 +429,8 @@ function dosomething_signup_get_signup_nids_by_uid($uid) {
  *   Whether or not logged in user can remove signup (and if it exists).
  */
 function dosomething_signup_node_unsignup_access($node) {
-  // Only allow access if user staff and is signed up.
-  return dosomething_user_is_staff() && dosomething_signup_exists($node->nid);
+  // Only allow access if user is signed up.
+  return dosomething_signup_exists($node->nid);
 }
 
 /**
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.views_default.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.views_default.inc
@@ -228,6 +228,24 @@ function dosomething_signup_views_default_views() {
     'image_style' => '310x310',
     'image_link' => '',
   );
+  /* Field: Content: Nid */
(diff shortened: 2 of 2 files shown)


aaronschachter:
this would also need to remove your subscription from mail chimp and mobile commons...

angaither:
OHHH :ok_hand: 
cc: @jonuy @DeeZone 
Is removing signups possible? Easy? Hard?

jonuy:
Possible from SMS. Worth noting that there are up to 3 Mobile Commons campaigns that a user would need to be unsubscribed from for a DS staff-pick campaign.

Using the `dosomething_sms` module, we could do this for each of the Mobile Commons campaign ids with something like:
`dosomething_sms_mobilecommons_opt_out($phone, $mobileCommonsCampaignId);`

aaronschachter:
I thought we had decided in the Kickoff meeting that version 1 wouldn't include the "remove signup" links, because of the required third party opt-outs.  Also there could be potential data "gotchas" where people reportback but then can cancel their signup? cc @julielorch @mikefantini

joolaay:
Yup we said V1 wouldn't include "remove signup", but since Andrea had some extra time we thought she could start working on it. We can wait to release it along with other updates in V2...no time pressure here.

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -429,8 +429,8 @@ function dosomething_signup_get_signup_nids_by_uid($uid) {
  *   Whether or not logged in user can remove signup (and if it exists).
  */
 function dosomething_signup_node_unsignup_access($node) {
-  // Only allow access if user staff and is signed up.
-  return dosomething_user_is_staff() && dosomething_signup_exists($node->nid);
+  // Only allow access if user is signed up.
+  return dosomething_signup_exists($node->nid);
```
aaronschachter, line 433: Adding this would cause the "View" and "Remove signup" tabs to appear on a campaign for an authenticated user, wouldn't it?

aaronschachter:
@angaither We're no longer using the `user_signups` view to display a user's signups on the profile.   Should we close this?

PR #1640, opened 2014-04-03
Updates to pass missing vars array to theme function.

### LINKED ISSUE #1641: Fix WatchDog issue reporting undefined variables
@_@

https://s3.amazonaws.com/uploads.hipchat.com/34218/305836/IfVWxqTGuGy7TTy/errors_sample.txt

weerd:
Should fix 'undefined variable' issue coming from watchdog. @_@ Didn't catch this when I made the update yesterday.

Basically navigation partial was calling variables that were no longer defined, and I needed to pass in the $vars array so variables would be set.

Resolves #1641


angaither:
Seems ok for a quick fix but it would be much nicer if we only sent the vars that the nav needed.

`$header_vars = array('search_box', 'logged_in', etc);
theme('navigation', $header_vars);`

DFurnes:
+1 @angaither

weerd:
@angaither Yeah, I was thinking of doing that but wasn't sure what would be best. I can update quick :)

Issue #1643, opened 2014-04-03
campaign (action guides) – "banner" font sizes on mobile


mmwtsn:
## Description

Pending the design approval of @crystalruthbell, Mike suggested we reduce the font size of the "banner" header type in the action guide modals to allow for longer titles as the current limit is `~12` characters.

![modal](https://cloud.githubusercontent.com/assets/1479130/2608536/0b9c1d74-bb6c-11e3-84b2-1841e74f0520.png)
### Notes
- This change should affect all modal styles
- This is waiting on a fix to [re-instate](https://github.com/DoSomething/dosomething/issues/1644) the default styles from Neue

DFurnes:
This is fixed in Neue 6.0! Closing.

PR #1645, opened 2014-04-03
Restores Neue Banner Styles

### LINKED ISSUE #1644: modal banner – default styles from Neue overridden
We had [pushed some changes](DoSomething/neue#234) to the default banner styles in Neue. Now that those changes are deployed the [app-specific overrides](https://github.com/DoSomething/dosomething/blob/dev/lib/themes/dosomething/paraneue_dosomething/scss/content/_modals.scss#L4-L15) can safely be removed 'cause they fuckin' it up.

mmwtsn:
# Changes
- Documented in #1644 

Resolves #1644
Related to #1643


weerd:
Sounds good. :+1: 

And by sounds, I mean looks.

PR #1651, opened 2014-04-03
Updating missing vars fix to ONLY pass in the necessary variables for th...

weerd:
...e Navigation. This will be the standard moving forward.

@DFurnes this should also make it easier to eventually pass along specific data to the navigation via the $navigation_vars array to apply additional classes, etc... (e.g. 'alternate' or 'reverse' for the text from black to white ;)


DFurnes:
Looks fantastic. :+1:

angaither:
:+1:

weerd:
Thanks for checking it out :dancer:

PR #1653, opened 2014-04-03
Dynamic gradients

### LINKED ISSUE #1586: Switch image gradient from Drupal image style to CSS.
Chatted with Andrea and Keri about switching the black to transparent gradient used on images throughout the site to be be generated dynamically via CSS (with png fallback for IE8). This will allow the gradient to be customized and tweaked rapidly and allow for making it adjustable depending on device, etc. Would likely solve the gradient dithering seen on some of the images as well.

weerd:
Sweet sweet gradient action :dancers: 

Resolves #1586


mmwtsn:
Looks good.

Should this be accompanied by changes on the module side?

weerd:
Figured since update was related to the module and the styles. The module update is pretty tiny. But not sure if the proper approach is to separate module updates from theme updates @_@.

weerd:
@angaither says those functions need to be used to still help w/ retrieving the image and for caching.

DFurnes:
This looks great! :ok_hand:

PR #1663, opened 2014-04-04
Padding Hack Refactor

### LINKED ISSUE #1629: campaign – refactor mobile padding
## History

When building out the campaign template we opted to use padding to control "gutters" on mobile. With the exception of images and other full-width elements, a `section-padding` mixin was applied. In order to account for different possible markup combinations generated by our Markdown parser we added some [gross overrides](https://github.com/DoSomething/dosomething/blob/dev/lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss#L70-L86).
## Options

A simple solution would be to remove full-width images on mobile and set the default padding on our outer container. If we're not able to change the image treatment, it would be preferable to set a negative margin on the full-width elements as we are with our `h2.banner` [styles](https://github.com/DoSomething/dosomething/blob/9c6aa45e4941cf188dfc8d5aa903bf59012d8a12/lib/themes/dosomething/paraneue_dosomething/scss/content/_modals.scss#L14).

mmwtsn:
## Changes
- Removes padding hack that indented all elements on the campaign page
- Removes misc. duplicate style

Resolves #1629

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
@@ -65,25 +65,7 @@ body.node-type-campaign {
   // Campaign "step" wrapper that contains all campaign sections
   .step {
     overflow: hidden;
-    margin: 1rem 0;
-
-    h3,
-    h4,
-    p,
-    div {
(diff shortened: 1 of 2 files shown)


DFurnes:
:ghost:

PR #1671, opened 2014-04-07
Campaign signup security.

### LINKED ISSUE #480: Signup: Security / hidden nid

angaither:
Fixes #480

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -248,8 +248,8 @@ function dosomething_user_is_under_thirteen($user = NULL) {
  */
 function dosomething_user_login_submit($form, &$form_state) {
   // If nid is not present, nothing to sign up for.  Exit.
-  if (isset($form_state['input']['nid'])) {
-    $nid = $form_state['input']['nid'];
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -290,8 +290,8 @@ function _dosomething_user_add_campaign_data(&$form) {
       // Add a hidden nid field.
       $form['nid'] = array(
         '#type' => 'hidden',
-        // @TODO: Encrypt me.
         '#value' => $obj->nid,
+        '#access' => FALSE,
```
aaronschachter, line 294: I had tried this, but it seemed like signups didn't work anymore upon login.  This is working for you?

aaronschachter, line 294: Oh I missed your comment in the issue.  If this is tested and good to go, awesome.

PR #1672, opened 2014-04-07
Removing textarea.js file that Drupal adds since it was breaking the JS ...

weerd:
...on campaign pages.


weerd:
@DFurnes :cactus:

DFurnes:
Non-value removed => value added! :100:

Issue #1665, opened 2014-04-04
action page (action guides) -copy running over modal


mikefantini:
copy is running over the modal of action guides. 

example: https://beta.dosomething.org/campaigns/i-want-you-quit-because  

https://trello-attachments.s3.amazonaws.com/52ea92c7ccdaf8415b8cab3a/533eb9e7334d9e9f53da6a2d/692x681/0ac9cad1aba7bd7b54130849f8125f67/Screen_Shot_2014-04-04_at_9.55.33_AM.png

weerd:
Should the link be explicitly output showing the full link, or would it potentially be better if the maybe the initial text of the list item was created as a link, with the `href` defined as the link that is currently printed. Might not be feasible based on how the info is input; just a thought but I could look into it.

mikefantini:
so like we talked about let's have the text wrap. also happening here with URLs (https://www.dosomething.org/campaigns/comeback-clothes when go to view the sources): 
![image](https://cloud.githubusercontent.com/assets/6330971/2634759/6975ab7c-be7d-11e3-921a-1c92fd4b8157.png)

weerd:
@mikefantini the above screenshot is for content outside of the modal, so I'll do a different fix for the sources. Do you want to add a separate ticket?

PR #1676, opened 2014-04-07
Updates to the action pages for the modal and the Prove It footer.

### LINKED ISSUE #1202: action page (contact) - need contact us on mobile
To do this add a period to "In partnership with H&M." Then have "Have a question? Contact us." next to that.

Also update text to read "Contact us." Instead of "Give us a shout"

weerd:
Resolves #1202 & #1665

With updated layout for the Prove It footer, I was able to combine most of the rules for the two items in the footer into unified rules and clean things up a bit :)


DFurnes:
:thumbsup:

PR #1677, opened 2014-04-07
Campaign finder feature flag

DFurnes:
#### Changes:
- Adds a feature flag to theme settings to toggle Campaign Finder (allowing us to test it on QA without deploying to production).
- Points Campaign Finder to use Black Angus rather than Morgan's laptop.
- Initializes Campaign Finder module only if its present on page.


lib/themes/dosomething/paraneue_dosomething/js/finder/CampaignFinder.js
```diff
@@ -20,7 +20,7 @@ define(["jquery", "lodash", "finder/CampaignResults"], function($, _, CampaignRe
        Configure this too. What's the base URL of the Solr server that you want
        the browser pinging? And what collection?
        */
-    baseURL: "http://192.168.1.139:8080/solr/",
+    baseURL: "http://192.168.1.169:8008/solr/",
```
desmondmorris, line 23: You may want to add this to the Drupal.settings javascript variable.  Just to cover your ass when/if that thing changes.  That value is already stashed in a variable, so you would just need to inject via drupal_add_js. Something like:

```
drupal_add_js(
  array(
    'myModule' => array(
      'solr_path' => variable_get('solr_path', 'http://localhost:8983');  // I dont know the right variable off of teh top of my head, but you get the idea
     )
  ),
  'setting'
);
```

Not a blocker right now, just a suggestion.

lib/themes/dosomething/paraneue_dosomething/templates/home/node--home.tpl.php
```diff
@@ -1,3 +1,105 @@
+<?php
+
+/**
+ * If theme setting is checked, show the new Campaign
+ * Finder/new homepage design.
+ * @see dosomething_user.module
+ */
+if( theme_get_setting('show_campaign_finder') ) {
```
desmondmorris, line 8: You may want to inject a variable into the template via preprocess or theme function rather than execute the theme_get_setting function.  Just to keep db calls out of the template.

Also, not a blocker for merge

desmondmorris:
Looks good.  I added a few non blocking considerations for a rainy day.

DFurnes:
Cool cool. Both good points. :+1:

PR #1687, opened 2014-04-08
Address finder style

### LINKED ISSUE #1539: action page (address collection) - style modal
We need the modal styled to collect addresses from users in order to send them banners.

This should be styled like the other modals on the page.

See #1484 for related issue.

DFurnes:
screen shot 2014-04-08 at 11 46 32 am

Fixes #1539.


weerd:
@DFurnes I take it the form gets output in this call? `<?php print $action_guide['content']; ?>`

weerd:
:+1:

aaronschachter:
@weerd The signup data form is output here: https://github.com/DFurnes/dosomething/blob/address-finder-style/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php#L227

weerd:
Cool. ok ;)

PR #1697, opened 2014-04-08
Campaign finder lazy load

DFurnes:
More CSS cleanup; lazy-loading thumbnails. No longer automatically triggers finder on initial load, in preparation for a static initial state on homepage.


barryclark:
Very nice very nice :poodle:

PR #1698, opened 2014-04-08
Reportback form: Custom field

### LINKED ISSUE #1647: campaigns - additional report back question
For now just need a radio button hard coded into Comeback Clothes.

Question: 
- Did you recycle the clothes through H&M?

Options:
- Yes
- No

Eventually we'll want this in the campaign creation interface so that campaigns can add their own additional report back questions.

---
- [x] New table: `dosomething_reportback_field` -- stores a node's additional reportback fields
- [x] New table:  `dosomething_reportback_field_data` - stores the user submitted field data per reportback
- [x] Render reportback_field as form element on reportback form
- [x] Write reportback_field form value to dosomething_reportback_field_data
- [x] Log reportback_field form value to dosomething_reportback_log
- [x] Update hook to populate record in `dosomething_reportback_field` for Comeback Clothes (nid 362)

aaronschachter:
Fixes #1647

Renders an additional field in the reportback form based on configurations in the "Reportback Fields" section.
-- Hardcodes the "options" element for radios for now, in the interest of getting this out for production tomorrow.

Adds a dosomething_reportback_field record for Comeback Clothes with status = 0, which means the additional question will not appear by default.  Must manually edit node 362 and check the "Add additional reportback field" checkbox to get it to appear.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -48,6 +48,7 @@ function dosomething_campaign_menu() {
  * Adds form elements for non-field campaign elements.
  */
 function _dosomething_campaign_form_extras(&$form, &$form_state) {
+  global $user;
   // If this is a node create form, display help text and exit out of extras.
   if (!isset($form['nid']['#value'])) {
     $form['custom'] = array(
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -56,6 +56,11 @@ function _dosomething_campaign_form_extras(&$form, &$form_state) {
     );
     return;
   }
+  // If admin:
+  if (user_access('administer modules') && module_exists('dosomething_reportback')) {
```
angaither: might be better to check user role here.

angaither:
Seems ok for comeback clothes though

aaronschachter:
@angaither updated to use role instead of a permission.  ok to merge?

angaither:
:+1:

Issue #1417, opened 2014-03-19
Import: Update node authors 


aaronschachter:
Not critical, but I think this is worth doing at some point.

Update the current all node uid and revisions to the corresponding staff user accounts, instead of using the stage uid's we imported from:

![who are these people](https://f.cloud.github.com/assets/1236811/2460747/1714996e-af6f-11e3-8f4b-e2180b0763c9.png)

angaither:
This code is all pushed to production, just waiting to test run this in QA.

angaither:
Ran in QA & production.

PR #1699, opened 2014-04-08
Update node authors script.

angaither:
Refs #1417
will fix, once run in production.

### DIFF
--- a/scripts/update-node-authors.php
+++ b/scripts/update-node-authors.php
@@ -1,0 +1,44 @@
+<?php
+/*
+ * Script to update node authors from the random uids from stage to acutal uids.
+ *
+ * To get the authors array, ran this query on stage and then query for matching uids on prod.
+ * SELECT DISTINCT u.uid, u.mail FROM users u INNER JOIN node n on n.uid = u.uid;
+ *
+ * drush --script-path=../scripts/ php-script update-node-authors.php
+ *
+ */
+
+// stage uid, beta uid, email for reference.
+$authors = array(
+  1  => 1,        /*'admin@example.com',             */
+  2  => 615996,   /*'ddeluca@dosomething.org',       */
+  3  => 632265,   /*'rmohammed@dosomething.org',     */
+  4  => 868854,   /*'spipergoldberg@dosomething.org',*/
+  5  => 1176550,  /*'aschachter@dosomething.org',    */
+  6  => 623863,   /*'fsheikh@dosomething.org',       */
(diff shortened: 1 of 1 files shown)


aaronschachter:
Looks ok. Kinda OCD, but could be a relatively small lift to update the node_revisions table as well... but in this case we'd need to check more than just the original author, as I know Ben has made revisions.

angaither:
@aaronschachter updated!

scripts/update-node-authors.php
```diff
@@ -0,0 +1,63 @@
...
+  9  => 1193105,  /*'nmody@dosomething.org',         */
+  11 => 1192054,  /*'aruderman@dosomething.org',     */
+  14 => 1157702,  /*'jlorch@dosomething.org',        */
+  18 => 1258186,  /*'dfurnes@dosomething.org',       */
+  19 => 547446,   /*'bclark@dosomething.org',        */
+  22 => 329195,   /*'mfantini@dosomething.org',      */
+);
+
+// Need to update the revisions first, otherwise it zeros out uids
+$revision_results = db_query("SELECT nid, vid, uid from node_revision");
```
aaronschachter, line 33: Aren't there some authors that would be returned in this list that don't appear in the $authors array?  I'm mostly thinking of Ben, hasn't he made revisions?  But would good to be sure if there are any other missing uid's as well...

angaither, line 33: Did he make revisions on stage or production? When I run a query getting all the users his email isn't returned.

`SELECT DISTINCT u.uid, u.mail FROM users u INNER JOIN node  n on n.uid = u.uid INNER JOIN node_revision nr on nr.uid = u.uid;`

angaither, line 33: Nevermind, my db was jenky after I ran the update. duhhh

angaither:
@aaronschachter for real this time.

scripts/update-node-authors.php
```diff
@@ -0,0 +1,65 @@
...
+foreach($revision_results as $result) {
+  $old_uid = $result->uid;
+  $new_uid = $authors[$old_uid];
+  if ($new_uid)  {
+    echo "Updating node revision: " . $result->nid . ' : ' . $result->vid . " from " . $old_uid . " to " . $new_uid . "\n";
+    db_update('node_revision') // Table name no longer needs {}
```
aaronschachter: Doesn't this guy need a WHERE condition?

aaronschachter:
Looks good :+1:

Issue #1592, opened 2014-03-31
action page (prove it) – add handwritten scholarship callout


mmwtsn:
## Description

Create a scholarship call out mimicking treatment on the pitch page pointing to the button.
## Mocks

These weren't officially mocked and signed off but here's a quick sketch of the general idea across mobile and desktop. Please ignore any styles unrelated to the scholarship call-out are out of date in these mocks.
### Mobile

![desktop](https://cloud.githubusercontent.com/assets/1479130/2618976/e392973c-bc28-11e3-947c-e33a0816e2a2.png)
### Desktop

![mobile](https://cloud.githubusercontent.com/assets/1479130/2618977/e3a3225a-bc28-11e3-8c64-6e32c144733b.png)

Broken out from #1455

Issue #1184, opened 2014-03-12
action (header) – add scholarship call out


mmwtsn:
## Description

Add handwritten scholarship call-out to the mobile action page.
### Mobile

![action-mobile](https://cloud.githubusercontent.com/assets/1479130/2618990/13698e20-bc29-11e3-8c8b-646e1933c4a6.png)
**NOTE:** Some of the elements of this design may have changed since the mock was created.

Related to #1592

mikefantini:
this is only needed for mobile

PR #1703, opened 2014-04-08
Action pages  font updates

### LINKED ISSUE #1658: neue – update handwritten font to COVERED BY YOUR GRACE
cc: @crystalruthbell

weerd:
Resolves #1658, #1592 & #1184

@DFurnes & @barryclark I reworked a bit how the scholarship stuff is getting handled, so we may want to chat through this one if anything seems unclear. Just gimme a holla!


DFurnes:
Since this relies on changes (updated fonts) in Neue, let's bump the required version number in [bower.json](https://github.com/DoSomething/dosomething/blob/dev/lib/themes/dosomething/paraneue_dosomething/bower.json#L23) and then tag those changes appropriately in Neue. Good habit to avoid things getting out of sync by mistake.

weerd:
Cool. @DFurnes Should I do that or have you already done that?

PR #1704, opened 2014-04-08
Add commas and ands to in partnership with list.

### LINKED ISSUE #1457: action page (partners) - comma and "and" logic with multiple partners
If there are multiple partners attached to a campaign we should have commas and the word "and" separating them.

So if there are two partners, we just need "and" separating them. If there are three or more partners we'd need commas separating each and "and" right before the last comma.

example: https://beta.dosomething.org/campaigns/cell-phones-survivors

image

angaither:
Moved to module
Fixes #1457

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
@@ -394,6 +394,9 @@
             In partnership with
             <?php foreach ($partners as $key => $partner) :?>
               <?php print $partner['name']; ?>
+              <?php if (count($partners) > 2 && $key == (count($partners) -2)): print 'and '; ?>
+              <?php elseif (count($partners) > 2 && $key < (count($partners) -1)): print ','; ?>
+              <?php elseif ($key < (count($partners) -1)): print 'and'; endif;?>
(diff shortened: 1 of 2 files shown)


lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -394,6 +394,9 @@
             In partnership with
             <?php foreach ($partners as $key => $partner) :?>
               <?php print $partner['name']; ?>
+              <?php if (count($partners) > 2 && $key == (count($partners) -2)): print 'and '; ?>
```
aaronschachter: Should it be `' and '` ?

angaither: It actually works fine without that, there's spaces before the commas, even though I didn't add them. @DFurnes @weerd do you know if extra spaces are being added?

weerd:
Nothing's going on CSS-wise that would be affecting it. I suggested to @angaither to run the output through trim() before printing to try and remove extra whitespace so that it can me more consistent :)

angaither:
@aaronschachter updated!

aaronschachter:
Looks great :+1:

PR #1714, opened 2014-04-09
Refactors Solr logic into separate adapter class.

DFurnes:
:snake:


DFurnes:
Tested side-by-side and haven't noticed any regressions in either rendering or the generated URL.

PR #1725, opened 2014-04-09
Single-select on Campaign Finder

### LINKED ISSUE #1722: Campaign Finder - Single select filters
The user should only be able to to select one thing from each category. 

When a 2nd item is selected it should remove the previous item.

DFurnes:
Makes options on the campaign finder single-select by field group. Fixes #1722.


weerd:
I think it looks swell. :+1:

Issue #1718, opened 2014-04-09
Reportback Field: PHP Notice Undefined property


aaronschachter:
Notice: Undefined property: ReportbackEntity::$rbid in ReportbackEntity->getFieldValue() (line 96 of /vagrant/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc).

PR #1726, opened 2014-04-09
Moar fixie

aaronschachter:
Getting rid of all these dang PHP errors!! 
#1718 was happening because there's an `entity_create` that was happening before loading the existing reportback, which isn't necessary if a reportback already exists.


angaither:
:+1: :shipit: :elephant: value added.

PR #1735, opened 2014-04-09
Fixing incorrect deactivation of choices when Time or Type selected

### LINKED ISSUE #1723: Campaign Finder - Filter items in Cause not being deactivated correctly
When you select a filter in Time or Type it doesn't deactivate links with no results in the filters in Cause. 

Selecting "Donate Something" as shown below should deactivate some of the filters in Cause and Time. 

image

blisteringherb:
Fixes #1723


DFurnes:
:boom:

PR #1745, opened 2014-04-09
Fix to radio button inputs on report back modal on campaigns.

### LINKED ISSUE #1693: Reportback Form: Radio styles
Depends on #1647, which is in review.

CSS is makin these radios look all cray.

To enable this locally (after #1647 is merged):
- Run `drush -y updb` (or pull from stage)
- Navigate to node/362/edit
- Under "Reportback Fields", check the checkbox.
- Save the node

---

screen shot 2014-04-08 at 3 25 54 pm

weerd:
Resolves #1693

Isolated the fix to only the reportback modal. I was concerned that doing it more globally could bork other forms!

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_modal-reportback.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_modal-reportback.scss
@@ -21,5 +21,22 @@
         }
       }
     }
+
+    .form-type-radio {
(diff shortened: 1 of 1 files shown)


aaronschachter:
Sure :+1: this would be good to get in for today, any fenders wanna weigh in?

barryclark:
:+1: !!

PR #1754, opened 2014-04-10
Fixes reportback field submit error

aaronschachter:
A new reportback entity must be created with a nid within the `entity_create` call, in order for it to be aware any `dosomething_reportback_field` field values it should be collecting.

The changes to the `db_merge` call don't actually fix anything, just makes it a little more readable.


angaither:
:+1:

PR #1763, opened 2014-04-11
Fixing inconsistency when filters are selected in different orders

### LINKED ISSUE #1762: campaign finder - inconsistency of ordering filters
See in the example below, when I select Homelessness + Poverty, Face to Face is disabled because there is no combination of campaigns that is Homelessness + Poverty _and_ Face to Face. However, if I _first_ select Face to Face, Homelessness + Poverty is _not disabled_ and so then when I go to select Homelessness + Poverty I get no results.

Homeless + Poverty selected, Face to Face disabled:
image

Face to Face selected, Homelessness+Poverty _not disabled_:
image

but has _no results_:
image

blisteringherb:
- There's still an issue with re-enabling all the filters which 
  may make this seem still borked, but it's separate.

Fixes #1762

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/finder/SolrAdapter.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/finder/SolrAdapter.js
@@ -18,7 +18,7 @@ define(["jquery", "lodash"], function($, _) {
      * Configure this too. What's the base URL of the Solr server that you want
      * the browser pinging? And what collection?
      **/
-    baseURL: "http://98.129.111.171:8080/solr/",
+    baseURL: "http://98.129.111.172/solr/",
     collection: "collection1",
 
     // The current throttle timeout identifier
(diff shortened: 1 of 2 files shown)


lib/themes/dosomething/paraneue_dosomething/js/finder/SolrAdapter.js
```diff
@@ -18,7 +18,7 @@ define(["jquery", "lodash"], function($, _) {
      * Configure this too. What's the base URL of the Solr server that you want
      * the browser pinging? And what collection?
      **/
-    baseURL: "http://98.129.111.171:8080/solr/",
+    baseURL: "http://98.129.111.172/solr/",
```
aaronschachter, line 21: Is this intentional?

aaronschachter:
Addition of parenthesis seems to make sense.. but just need confirmation from @blisteringherb  that the URL change was intentional.

blisteringherb:
Yes, this URL points to the new staging server, and the staging server has the Varnish rules on it.

weerd:
:+1: Looks fine. I don't think the changes in the values for the input (adding parenthesis) would affect much on the client side.

PR #1774, opened 2014-04-14
Improve performance by reducing solr obj size and enabling caching

### LINKED ISSUE #1773: Finder: Improve Performance
More fields than necessary are returned, only return the necessary fields and use a named callback function for the jsonp call to allow caching.

blisteringherb:
Fixes #1773

@DFurnes
@aaronschachter

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/finder/SolrAdapter.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/finder/SolrAdapter.js
@@ -1,6 +1,9 @@
 define(["jquery", "lodash"], function($, _) {
   "use strict";
 
+  // Create a callback function for caching the jsonp response.
+  window.solrResponse = function(response) {};
+
   /**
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/js/finder/SolrAdapter.js
```diff
@@ -1,6 +1,9 @@
 define(["jquery", "lodash"], function($, _) {
   "use strict";
 
+  // Create a callback function for caching the jsonp response.
+  window.solrResponse = function(response) {};
```
DFurnes, line 5: Can we make this part of the `SolrAdapter` rather than attaching it to the global scope?

Edit: Never mind, I think I see what's going on.

DFurnes:
Restricting the returned fields is great. :+1:

Let's talk over the caching, curious about the approach there.

PR #1778, opened 2014-04-14
Removes mailsystem

### LINKED ISSUE #1759: Remove mailsystem from profile and make
According to @DeeZone we don't need this module. so let's kill it.

deezone:
Fixes #1759

The Mail System module is no longer needed as all mail functionality is now running through the Message Broker. The Message Broker handles all communication with the Mailchimp and Mandrill services.


aaronschachter:
Looks good :+1:

angaither:
What's the process for removing modules in production, do we need to do anything after this deployment?

aaronschachter:
Ohh yeah.  Actually we should disable and uninstall now before this is removed from codebase and everything breaks. @DeeZone can you disable and uninstall on stage and prod using UI (or drush) ?

deezone:
@aaronschachter 
mailsystem is a dependency for DoSomething User. Is it **not** possible to disable / uninstall the mail system module without disabling the DoSomething User module? Perhaps this needs to be a two step deployment process?

aaronschachter:
I think you're right @DeeZone.. perhaps the first PR should just remove the dependency, so we can disable and uninstall.

deezone:
@aaronschachter Step one, removal of DoSomething User dependencyy on Mail System in place.

aaronschachter:
:+1:

Issue #1780, opened 2014-04-14
Remove image cache actions


angaither:
- [x] Remove all image cached processed presets
- [x] Remove the module from `dosomething.info`
- [x] Remove the module from `drupal-org.make`

aaronschachter:
We'll want to make sure the module is disabled and uninstalled before moving from the .info and .make

PR #1782, opened 2014-04-14
Remove image cache presets that used image cache actions

angaither:
Refs #1780


aaronschachter:
Looks ok - just checking: these presets are no longer called anywhere else?

aaronschachter:
Will need to also remove from dosomething_image.info.  Please make sure to find/destroy all instances.

Issue #1764, opened 2014-04-13
Campaign Finder - when I change a selection, grayed out options don't reset


mikefantini:
If I do the following combination:
- \+ Environment
- \+ Improve a Space
- \- Environment
- \- Improve a Space

I am now unable select Face to Face, even though there is _nothing else selected_. Clearly there is something in the database that is tagged with Face to Face, so I should be able to select it.

Video here: http://screencast.com/t/bM6bUKaRiZ0

Not sure if this is related to other solr issues or not so wanted to make sure it was documented separately.

mikefantini:
Easier combination: Select Education. 5+ hours then becomes inactive. Then de-select Education, and 5+ hours remains inactive. Even though, again I should be able to select 5+ hours since there are campaigns that have 5+ hours in the database.

DFurnes:
This should be fixed now.

Issue #1777, opened 2014-04-14
Campaign Finder - when I make a selection within a category, all other options should be unselectable


mikefantini:
For instance, if in cause I select Animals, all other causes should be unselectable.

We should also add some sort of x in a circle so it's clear to the user that they can remove their selection (need to make sure this treatment doesn't get too chunky on mobile).

cc: @julielorch @crystalruthbell @barryclark

DFurnes:
Fixed. Let's revisit the styling at standup tomorrow morning.

Issue #1760, opened 2014-04-10
campaign finder - filter disabling behaving incorrectly


mikefantini:
Goal: prevent the user from choosing a combination of filters that would provide 0 results

The ideal situation for the finder interactions:
- cause is limited by selections that are made to time and type (but disregards selections that are made to cause)
- time is limited by selections that are made to type and cause (but disregards selections made to time)
- type is limited by selections that are made by cause and time (but disregards selections made to type)
- the results are limited by selections that are made to cause, time, and type

For example, if I select 5+ hours, then Animals, I should be able to change my mind on Time and select 2-5 hours (because the combination of Animals + no preference of Type should allow me to select 2-5 hours and get results). As you can see, currently I can't do that: http://screencast.com/t/lKdUH7Lj3P

The reason (I suspect) that I can't change my mind on Time is because Solr is saying that there is no combination that has Animals + 5+ hours + 2-5 hours, however, it _should not_ be factoring the 5+ hours into it's calculation.

This appears to be happening across the board. In this example I should be able to change my mind to Share Something from Donate Something but I can't. The only Type I can change is to Improve a Space because Party Animal happens to be tagged with both Donate Something _and_ Improve a Space, so it seems that Solr is factoring in choices made within a particular category as limiting factors for which choices should be available: http://screencast.com/t/WqMd4301GCHc

Again, since the goal of the graying out is to prevent me from making combinations that will return 0 results, I _should_ be able to select Animals _and_ Share Something.

Hopefully one of these drawings help clarify as well:

![image](https://cloud.githubusercontent.com/assets/6330971/2674243/e3540394-c0fd-11e3-93b0-357ae3ae52fe.png)

![photo 42](https://cloud.githubusercontent.com/assets/6330971/2674299/1a973bae-c0ff-11e3-9a79-ed74ac38831b.JPG)

cc: @mshmsh5000 @julielorch @weerd

weerd:
I need to connect with @blisteringherb and confirm this is a bender issue and not a fender issue since the query that is sent to Solr is built in JS... we can chat further during standup :)

blisteringherb:
Chatted w/ @weerd on the best way to fix this.  We'll need to keep track of both the first clicked and the last selected and then clear those once everything has been deselected.  Working on doing this now.

DFurnes:
Per discussion this morning, side-stepped this issue by disabling all other fields in a field group when one is selected (requiring you to deselect then re-select, which updates facets from Solr).

Issue #1727, opened 2014-04-09
Campaign Finder - Selecting a Type filter clears inactive filters in Type


barryclark:
To replicate: Select Homelessness + Poverty from Cause, then select Host An Event from Type. It now clears all of the inactive filter items from the Type... the desired result is that they should stay inactive.

![image](https://cloud.githubusercontent.com/assets/692762/2657637/114bf6ee-bfff-11e3-867b-ea488af77dfd.png)

![image](https://cloud.githubusercontent.com/assets/692762/2657607/b6ae921e-bffe-11e3-8e59-1095bb835620.png)

mikefantini:
also happening with time category as well: http://screencast.com/t/aY3QXMs84Y3y

mikefantini:
@blisteringherb has this been fixed and pushed to staging?

mikefantini:
nope I'm wrong, this is still an open issue

mikefantini:
trello card here: https://trello.com/c/xSjUYzq1

mikefantini:
this then results in previous selections being de-selected: http://screencast.com/t/IF3UGSjRvn

but again, we just need to solve the root issue and then hopefully this bug will never re-appear

DFurnes:
Fixed.

PR #1785, opened 2014-04-14
Disable those fields

### LINKED ISSUE #1765: Campaign Finder - can't switch between choices within a category after second category selected
If I do the following combination:
- \+ 1 hour or less
- \+ Donate Something

I am able to do the campaign Fantini & "The Test". So we know that 1 hour or less + Donate Something yields results and should be an accepted combination.

However, if I do:
- \+ Make Something
- \+ 1 hour or less
- \- Make Something

I am then unable to select Donate Something even though we know the combination of  1 hour or less + Donate Something yields results. Really I should be able to just switch from Make Something to Donate Something without even having to de-select Make Something.

Probably related to #1764 and possibly a duplicate since fixing that should fix this, but there seems to be a lot of funny stuff going on since this is so complicated that wanted to be safe and create a new issue.

http://screencast.com/t/2u2TuJrABt

DFurnes:
:cactus: 

Fixes #1765 and #1777.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/finder/FormView.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/finder/FormView.js
@@ -69,17 +69,8 @@ define(["jquery", "lodash", "unveil", "finder/Finder", "finder/Campaign", "finde
       _.each(FormView.$fields, function(fieldGroup) {
         // Attach ourselves to the "change" event on the input fields
         fieldGroup.change(function () {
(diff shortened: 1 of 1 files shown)


DFurnes:
This desperately needs to get refactored, but this should make it _functionally_ work the way we want.

DFurnes:
Should also fix bugs #1764, #1760, #1727.

weerd:
:+1:

PR #1795, opened 2014-04-15
Last step in removing the Mail System module

### LINKED ISSUE #1759: Remove mailsystem from profile and make

### LINKED ISSUE #1794: Mail System Module removal - part II
For part one see:
https://github.com/DoSomething/dosomething/pull/1778

After part I the Mail System module is no longer a dependency for the DoSomething User module. The Mail System module can now be safely disabled and removed.

deezone:
Fixes #1794
Fixes #1759

Last step in removing the Mail System module.

The Mail System module has been disabled and uninstalled on both the production and staging sites.


aaronschachter:
:+1:

PR #1804, opened 2014-04-15
Admin campaign status page

### LINKED ISSUE #1802: campaign status page - show character limit violators
need an admin status page that lists all campaigns that have titles over 20 characters and all campaigns that have call to action over 65 characters.

aaronschachter:
Creates a page `admin/content/campaign-status` to resolve #1802


angaither:
:+1:

angaither:
Would be cool if this status was shown on the node itself as well

Issue #74, opened 2013-10-29
Valid Address API


aaronschachter:
Use service like USPS to validate that an address is legit.  Should be an external DS Library to be used by any webapp / app.

barryclark:
This isn't necessary for the mailing work this sprint, @angaither, but is a "nice to have" if you get time.

mikefantini:
@barryclark @angaither after chatting with @Bladt sounds like we should definitely try to get this in this sprint as it's a huge time/money suck if we don't have it. Going to put it into this sprint as medium priority.

angaither:
[USPS](https://www.usps.com/business/web-tools-apis/address-information.htm)
- free
- only takes xml
- no geocode

[Geocode.io](http://geocod.io/)
-  2,500 queries/day 
- takes json
- geocodes & has 'accuracy indicator' on addresses 
- @desmondmorris is all about it

![](http://f.cl.ly/items/0R1N1Q1F1H0L3i1L3Z1V/Screen%20Shot%202014-03-24%20at%204.37.02%20PM.png)

[Google](https://developers.google.com/maps/documentation/geocoding/)
- 2,500 queries/day
- json/xml
- geocodes

DFurnes:
Potential bullet point to add to the Google list: "The Geocoding API may only be used in conjunction with a Google map; geocoding results without displaying them on a map is prohibited."

mikefantini:
removed the comment from #1539 so that we keep that to styling and keep discussions on this issue about whether to validate client side or server side

angaither:
I think we should handle the actual address parsing on the client side, this way we know invalid addresses aren't saved with the user. We can handle asking 'is this what you meant' before the data is saved. 
@DFurnes thoughts?

DFurnes:
This should be a check on the sever-side before the data is saved to the user's profile then, right?

Client-side validation is really just a nice way to help the user catch errors without wasting their time on a full page reload. We shouldn't be relying on it to prevent bad data from hitting the database, since it won't run if a user has JavaScript disabled, or scripts don't load due to an error, etc.

DFurnes:
If we're caching these validations on the server, we can make an AJAX call to that check on client-side so the page doesn't have to refresh if the user needs to make any changes.

Let's chat on Monday about how we should handle this so we're all on the same page. :)

angaither:
AJAX call FTW. I'm on it. 

[...]

PR #1812, opened 2014-04-15
Validate address on form

angaither:
- parses XML from the UPS api
- form errors on address validation
- js errors on lack of first, last name.

TODO: make sure the errors appear in the collection modal.

Refs #74

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user_valid_address.inc
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user_valid_address.inc
@@ -40,7 +40,7 @@ function dosomething_user_ups_api_settings($form, &$form_state) {
  *  User ID
  *
  */
-function dosomething_user_validate_address($uid) {
+function dosomething_user_validate_user_address($uid) {
   $user = user_load($uid);
   $address = $user->field_address[LANGUAGE_NONE][0];
   $first_name = $user->field_first_name[LANGUAGE_NONE][0]['value'];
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user_valid_address.inc
```diff
@@ -118,10 +206,10 @@ function dosomething_user_compile_ups_xml($first_name, $address) {
  * @return string
  *   XML string
  */
-function dosomething_user_compile_xml_address($first_name, $address) {
+function dosomething_user_compile_xml_address($first_name, $last_name, $address) {
 
   $addr_xml = array(
-    'ConsigneeName' => $first_name, // Name
+    'ConsigneeName' => $first_name . $last_name, // Name
```
aaronschachter: Does this need a space? `$first_name . ' ' . $last_name` ?

aaronschachter:
Looks ok, except for question above.  

Pretty sure we'll want to revisit the error and success messages, we should create an issue to determine what the copy should be.

lib/modules/dosomething/dosomething_user/dosomething_user_valid_address.inc
```diff
@@ -48,13 +48,101 @@ function dosomething_user_validate_address($uid) {
...
+
+/**
+ * Parse XML results from UPS api.
+ *
+ * @param array $xml
+ *  XML results from the UPS api.
+ * @return array $address
+ *  A address array in the format the addressfield expects.
+ */
+function dosomething_user_parse_xml_results($xml) {
```
aaronschachter: Not critical, but I think it'd make sense to add `ups` into the function name when the functions relate specifically to UPS things. `dosomething_user_parse_ups_xml_results`

PR #1823, opened 2014-04-16
Update address validation errors

### LINKED ISSUE #1817: Address field validation errors needed
### Client Side
- [ ] if no first name/after first name 
- [ ] if no last name/after last name
### Server Side
- [ ] ambiguous address (ie we think you meant this)
- [ ] no results (no choices came back form UPS api)

@mikefantini

angaither:
Fixes #1817

I'm already mad this is in two functions.


aaronschachter:
Good ones. :+1:

PR #1826, opened 2014-04-16
Signup Data Form: Display submitted copy

### LINKED ISSUE #1806: action page (address collection) - post-form-submission state
If a user has submitted a form requesting we send them stuff (i.e. banners for Comeback Clothes) we should prevent them from submitting that form a second time.

If I have already requested banners, I should still see the link to trigger the address collection modal. However, when I click on the link the modal I see should hide the address fields and the copy from the "form copy" field and instead give me a message saying "We received your request blah blah blah".

@aaronschachter I assume you'd rather have this be a standard message across all rather than being another field. Definitely open to trying to start but I feel like for Thumb Wars we'll run into a bunch of questions, but happy to tackle that if/when it comes if you'd prefer. Let me know what you think.

If we do go with just one standard message, for now let's do "You requested your package on [time stamp in the form of Month, xxst/nd/rd/th] and it's on it's way! Questions? Fill out our contact us at the bottom of the page and we'll get right back to you."

aaronschachter:
Fixes #1806

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
@@ -86,6 +86,12 @@ function _dosomething_signup_node_signup_data_form(&$form, &$form_state) {
     '#title' => t('Confirmation message'),
     '#default_value' => $values['confirm_msg'],
   );
+  $form[$fieldset]['config'][$prefix . 'form_submitted_copy'] = array(
+    '#type' => 'textarea',
+    '#title' => t('Form submitted copy'),
+    '#default_value' => $values['form_submitted_copy'],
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -138,15 +146,32 @@ function dosomething_signup_node_signup_data_form_submit(&$form, &$form_state) {
...
     '#suffix' => '</h2>',
   );
+  // Load signup entity.
+  $signup = signup_load(dosomething_signup_exists($nid));
+  // If the user has submitted form already:
+  if ($timestamp = $signup->signup_data_form_timestamp) {
+    // Get filtered form_submitted_copy.
+    $copy = dosomething_signup_filter_form_submitted_copy($config['form_submitted_copy'], $timestamp);
+    // Display the form submitted copy:
+    $form['submitted_copy'] = array(
```
angaither, line 161: default might be nice?

angaither:
Looks ok though :+1:

Issue #1801, opened 2014-04-15
[Campaign Finder] Subpixel rendering issue with campaign boxes


weerd:
At certain sizes of the browser, the campaign boxes sometimes clear and cause gaps due to subpixel rendering issues. Need to find a solution; potentially force a few clear-fixes on specific boxes. Likely use nth-child() to address clearing box, and ignore the support for IE8...

joolaay:
@weerd I'm actually seeing a lot of these gaps when other people open the finder - maybe you can talk me through this so I understand properly?

weerd:
@julielorch sure. It's a browser specific issue. Sometimes happens on certain browsers more than others. We can discuss further :)

barryclark:
I think this got fixed, right @weerd?

DFurnes:
Yup, this was fixed by #1835.

barryclark:
Thanks @DFurnes

PR #1835, opened 2014-04-16
Fixing the block that doesn't clear stuff.

weerd:
Addresses stuff in #1801 but should take another look.


barryclark:
:+1:

DFurnes:
:racehorse:

PR #1846, opened 2014-04-17
Don't validate null string addresses.

### LINKED ISSUE #1845: address collection - required when editing a user's profile
You are required to enter an address when editing a user's profile.

User gets "hmmm we couldn't find that address.. Try again". Happens whether you are an admin editing a user account or an auth user editing your own account.

angaither:
Fixes #1845


aaronschachter:
:+1:

Issue #1840, opened 2014-04-17
EntityMetadataWrapper errors


aaronschachter:
Happened after merging #1826.  All features are default.

On campaigns:

EntityMetadataWrapperException: Unknown data property field_scholarship_amount. in EntityStructureWrapper->getPropertyInfo() (line 335 of /var/www/beta.dosomething.org/releases/20140417134155/html/profiles/dosomething/modules/contrib/entity/includes/entity.wrapper.inc).
The website encountered an unexpected error. Please try again later.

Static content:

EntityMetadataWrapperException: Unknown data property field_video. in EntityStructureWrapper->getPropertyInfo() (line 335 of /var/www/beta.dosomething.org/releases/20140417134155/html/profiles/dosomething/modules/contrib/entity/includes/entity.wrapper.inc).
The website encountered an unexpected error. Please try again later.

aaronschachter:
Was hoping that https://drupal.org/comment/6507278#comment-6507278 (hitting save on the "Fields" admin of the campaign content type) would fix.  No luck. Still investigating.  Other comments in that thread suggest upgrading to the latest dev version of UUID (sadness)

aaronschachter:
Per https://drupal.org/comment/8399767#comment-8399767: "Now here's the bit i think helped. The issue seems to occur when the enabling script runs while over the max_execution_time and fails to install some information in the database properly.this fails to install the modules correctly."

The database may be borked because the install step takes too long to complete.This seems kind of crazy because the `dosomething_signup_data_form` table has no more than like 100 rows or so.  One way to test may be to restore an old copy of staging DB, set the max_execution_time, and see if that fixes.  Not sure how to actually repair the DB at this point (if this is the issue)

aaronschachter:
Here's where things seem to get pretty crazy.  @weerd had pulled down codebase yesterday morning, with the Signup Entity code stuff in place. things were working fine all day. We noticed that when @DFurnes cleared cache, his local died with this error.

So, @weerd made a copy of his local DB before clearing cache. cleared cache. site dies.  restored the old DB. cleared cache and site is still dead.  Right now we're testing his local to see if commenting out redis stuff from settings.php fixes, as I've had luck with that locally.  Aware that can't be the fix, but hoping it could help point us in the right direction.

aaronschachter:
I've been able to replicate this error by simply adding this one offending line into `dosomething_signup.signup_data_form.inc`:

```
[...]

Issue #1850, opened 2014-04-17
international - error when report back


mikefantini:
On http://nigeria.international.dosomething.org/ when I go to report back on a campaign I am currently taken to this screen:

![image](https://cloud.githubusercontent.com/assets/6330971/2736786/edd7379a-c671-11e3-8c7d-04de286c1d82.png)

Previously this error occurred when reporting back:
![upload 2014-04-04 at 9 41 23 am](https://cloud.githubusercontent.com/assets/6330971/2736804/454b6820-c672-11e3-9544-0162c8d72efb.png)

In both instances reported back with a photo I have used previously to test report backs (successfully) and properly filled out # of items collected and why is this important to you.

cc: @barryclark

mshmsh5000:
@desmondmorris Problem with /tmp?

desmondmorris:
@mshmsh5000 @mikefantini This is an issue with message broker.  Its a similar issue to #1848.  The call to `dosomething_user_mbp_request` is causing the WSOD.

desmondmorris:
Looks like `dosomething_user_mbp_request` has be deprecated and the logic has been abstracted to dosomething_mbp_request.  The platform codebase on international is out of date.

desmondmorris:
@mshmsh5000 @mikefantini I just ran deploy to international and it looks like we are experiencing the same entitywrapper issues that @aaronschachter has been fighting through.  I was forced to rollback the deploy.  We can not deploy here until #1840 is resolved.

mikefantini:
@desmondmorris @mshmsh5000 #1840 is closed, your call on when we should deploy the fix. I can just talk through it right now since it's just with registration.

desmondmorris:
@mikefantini I have updated the international platform and each of the individual sites.

PR #1851, opened 2014-04-17
Hotfix for #1840 - remove signup as entity

aaronschachter:
:warning: Not ready to merge, opened for visibility.

This is making things die.

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.info
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.info
@@ -12,4 +12,4 @@ features[user_permission][] = view any signup
 features[views_view][] = node_signups
 features[views_view][] = user_signups
 files[] = dosomething_signup.test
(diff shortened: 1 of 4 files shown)


aaronschachter:
We replaced staging DB with a version from 6am, before I ran the update hook this morning. However the error is still occurring.

Locally: I restored the DB using `ds pull stage`, cleared cache with these changes, and things were working.  Then running after dosomething_signup_update_7004, borked again.  

Retracing my steps to try and replicate (taking the same database DB from stage) and now its not working.  :trollface:

aaronschachter:
Also FYI @DFurnes never ran the update hook, but cleared cache and his local borked too.

desmondmorris:
@aaronschachter is this PR still relevant?  Wondering if it is safe to update the international tier.  Releated to #1850

aaronschachter:
Good call, dev branch should be in good shape now.

PR #1853, opened 2014-04-17
Stubs out support for produceCampaignCache

### LINKED ISSUE #1844: Message Broker add "campaign_create" and "campaign_update"
Message Broker needs the following field values from any new or updated campaign.

The values will be cached by the Message Broker system and used to assemble the daily digest messages that are tailored for each user based on their campaign activity.

Using hook_node_update() a call to `dosomething_mbp_request('produceCampaignCache', $params)` where:

```
$params = array(
  'application_event' => 'create_campaign' OR 'update_campaign',
  'application_id' => 2, (Drupal site)
  'title' => '??',
  'url' => '??',
  'call_to_action_text' => '??',
  'do_it_text' => '??',
  'cover_image_url' => '??',
  'sponsor_image_url' => '??',
  'update_text' => '??',
  'update_date' => '??',
)
```

**Notes**:
- The cover image needs its own image style to meet email image dimension requirements. Investigate if images can be accessed from user email (http) requests off site.

deezone:
Adding in support for campaign creation and update hooks to send requests / details to Message Broker.

Fixes #1844

See:
https://github.com/DoSomething/message_broker_producer/pull/23
https://github.com/DoSomething/message-broker/issues/36


deezone:
@aaronschachter Please review this pull request. I think it's a complete solution that will not require any additional logic in the dosomething code base due to the use of hook_node_insert() and hook_node_update().

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
```diff
@@ -14,6 +14,44 @@
  *  Optional keyed array of data to send externally.
  */
 function dosomething_mbp_request($origin, $params = NULL) {
+
+  $productionType = NULL;
```
aaronschachter: minor, but let's stick to Drupal coding standards here, no camelCase. `$production_type`

deezone: Ah, good point. My brain keep flipping between the two standards as I jump between the Drupal side and the OO rabbit code.

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
```diff
@@ -14,6 +14,44 @@
...
+/**
+ * Build payload for transactional producer requeusts. Any adjustments to the
+ * data before becoming a payload should be done in this function.
+ *
+ * @param string $origin
+ *  Where the message request came from.
+ * @param array $params
+ *  Optional keyed array of data to send externally.
+ */
+function dosomething_mbp_transactional($origin, $params = NULL) {
```
aaronschachter: This function name is very general, upon first look it suggests it fires a transactional message.  Let's change the name to reflect what you mention in the comments, maybe something like `dosomething_mbp_get_transactional_array()` ?

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
```diff
@@ -67,11 +105,95 @@ function dosomething_mbp_request($origin, $params = NULL) {
...
+
+/**
+ * Build payload for campaign cache message requests.
+ *
+ * @param string $origin
+ *  Where the message request came from.
+ * @param array $params
+ *  Optional keyed array of data to send externally.
+ */
+function dosomething_mbp_campaign_cache($origin, $params = NULL) {
```
aaronschachter: Same here.  This function name could provide a bit more clarity.  Perhaps something like  `get_campaign_cache_array()` ?

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
```diff
@@ -67,11 +105,95 @@ function dosomething_mbp_request($origin, $params = NULL) {
...
+  return $payload;
+}
+
+/**
+ * Build payload for campaign cache message requests.
+ *
+ * @param string $origin
+ *  Where the message request came from.
+ * @param array $params
+ *  Optional keyed array of data to send externally.
```
aaronschachter: Should also include a `@return array`

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
```diff
@@ -14,6 +14,44 @@
...
+
+/**
+ * Build payload for transactional producer requeusts. Any adjustments to the
+ * data before becoming a payload should be done in this function.
+ *
+ * @param string $origin
+ *  Where the message request came from.
+ * @param array $params
+ *  Optional keyed array of data to send externally.
+ */
```
aaronschachter, line 64: Should also include a `@return array`

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
```diff
@@ -67,11 +105,95 @@ function dosomething_mbp_request($origin, $params = NULL) {
...
   }
-  catch (Exception $e){
-    watchdog('dosomething_mbp', $e, array(), WATCHDOG_ERROR);
-    return FALSE;
+
+  return $payload;
+}
+
+/*
+ * Impliments hook_node_insert()
```
aaronschachter: Spelling.  Coding standards also specify that a period should be at the end of the comment, `e.g. Implements hook_node_insert().`

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
```diff
@@ -67,11 +105,95 @@ function dosomething_mbp_request($origin, $params = NULL) {
...
+
+/*
+ * Collects and format values from $node submission and populates $params.
+ *
+ * @param string $action
+ *  Type of mbp requet
+ * @param object $node
+ *  Submitted values for campaign
+ */
+function dosomething_mbp_collect_params($action, $node) {
```
aaronschachter: Same here, this function is doing more than just collecting params, its doing that but also sending a MBP request (which I wouldn't have guessed by the function name).  Perhaps this is something like `dosomething_mbp_campaign_request($action, $node)` ?  Or `set_mbp_campaign_cache` ?  Maybe we should chat in person.. i could see a similar function being used for other node types as well (mainly SMS Game, for example.. if we ever added transactional emails)

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
```diff
@@ -67,11 +105,95 @@ function dosomething_mbp_request($origin, $params = NULL) {
...
+ */
+function dosomething_mbp_node_update($node) {
+  dosomething_mbp_collect_params('update_campaign', $node);
+}
+
+/*
+ * Collects and format values from $node submission and populates $params.
+ *
+ * @param string $action
+ *  Type of mbp requet
```
aaronschachter: Minor, but spelling.

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
```diff
@@ -67,11 +105,95 @@ function dosomething_mbp_request($origin, $params = NULL) {
...
+ *
+ * @param object $node
+ *  Submitted values for campaign
+ */
+function dosomething_mbp_node_insert($node) {
+  dosomething_mbp_collect_params('create_campaign', $node);
+}
+
+/*
+ * Impliments hook_node_update()
```
aaronschachter: Same.

deezone:
@aaronschachter One last review please.

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
```diff
@@ -5,15 +5,62 @@
...
+  }
+
+}
+
+/**
+ * Build payload for transactional producer requests.
+ *
+ * NOTE:
+ * Any adjustments to the data before becoming a payload should be done in
+ * this function.
```
aaronschachter, line 54: very nice

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
```diff
@@ -65,13 +115,103 @@ function dosomething_mbp_request($origin, $params = NULL) {
...
+
+/*
+ * Collects and format values from $node submission and populates $params.
+ *
+ * @param string $action
+ *   Type of mbp request
+ * @param object $node
+ *   Submitted values for campaign
+ */
+function dosomething_mbp_collect_params($action, $node) {
```
aaronschachter: We should call this function something like `dosomething_mbp_campaign_request` and do the node type check from within inside the `hook_node_[op]` functions.  The function name `collect_params` doesn't suggest its actually sending a request.

Keeping the campaign specific code into one function will prevent this function from becoming a gigantic switch / conditional for the different node types.

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
```diff
@@ -65,13 +115,103 @@ function dosomething_mbp_request($origin, $params = NULL) {
...
+ * Implements hook_node_update().
+ *
+ * Respond to the update of campaign nodes to produce Message Broker entries
+ * in the directCacheExchange.
+ *
+ * @param object $node
+ *  Submitted values for campaign
+ */
+function dosomething_mbp_node_update($node) {
+  dosomething_mbp_collect_params('update_campaign', $node);
```
aaronschachter: Per  comment below:

```
if ($node->type == 'campaign') {
  dosomething_mbp_campaign_request('update_campaign', $node);
}
```

deezone:
@aaronschachter Looks like this pull request has become a little old and nasty with other changes in the code base getting pulled in. I assume it would be better to toss this request and open a fresh one?

aaronschachter:
Hmm yikes.  As long as the commits that are not yours have the same hashes, it should be ok.  I suppose it depends on how much work it would be to create a fresh one.  If not, then that's definitely cleaner...

PR #1856, opened 2014-04-18
Revert epic fail code

aaronschachter:
Thanks to @angaither -- This is what we're testing, seems to hotfix #1840 for reals.


angaither:
SAME TIME STUFF.

PR #1858, opened 2014-04-18
Redirect with modal open on form error.

DFurnes:
Redirects to page with modal hash in URL if there's a form error (will trigger modal on load with updates in Neue).

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
@@ -223,14 +223,14 @@ function dosomething_signup_data_validate_address($form, &$form_state) {
 
   // Did we not get any results?
   if (in_array('sorry', $formatted_address)) {
-    drupal_goto($_SERVER['REQUEST_URI'], array('fragment' => 'modal-signup-data-form'));
     form_set_error('dosomething_user_validate_address', t('Hmmm, we couldn’t find that address. Please try again.'));
+    drupal_goto($_SERVER['REQUEST_URI'], array('fragment' => 'modal-signup-data-form'));
   }
   // Did it come back from the api as ambiguous? -- Check with the user.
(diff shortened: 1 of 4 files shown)


barryclark:
:+1: !

weerd:
:sun_with_face: :+1:

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -223,11 +223,13 @@ function dosomething_signup_data_validate_address($form, &$form_state) {
 
   // Did we not get any results?
   if (in_array('sorry', $formatted_address)) {
+    drupal_goto($_SERVER['REQUEST_URI'], array('fragment' => 'modal-signup-data-form'));
```
aaronschachter: Won't this reload the page and the user has to fill out the form again?

DFurnes: Yup, but only on form errors (which would currently reload the page but not include the modal hash at the end).

aaronschachter: The difference for current behavior though is that the form reloads with the user input still filled out (but they do need to figure out to click on the link again).  Not going to keep me up at night or anything, its just something we lose by doing the `drupal_goto`.

DFurnes:
So this isn't great, but it should be fine enough for now. If Drupal outputs a `.messages` class at the top of the page, we copy it into a container in the modal. We lose the user's filled form state

screen shot 2014-04-18 at 1 55 32 pm

DFurnes:
Gonna merge this guy in so that @mikefantini can check it out on staging. This is only a temporary fix until we can run the address validation as a client-side check before submitting the form, so it's a bit messy but not permanent.

barryclark:
Nice job @DFurnes @angaither :+1:

PR #1863, opened 2014-04-18
Makes sure to redirect with modal open (but doesn't lose form state).

DFurnes:
This should let us use the URL fragment to re-trigger the modal after form is submitted, but stop form state from surviving.

/cc @aaronschachter


barryclark:
:+1:

PR #1865, opened 2014-04-18
Uses same arrow as footer in campaign finder.

DFurnes:
Uses arrow from footer in Campaign Finder drop downs for consistency, per conversation in Design HipChat room. Tested in Safari, Chrome, Firefox, IE8, IE9, IE11. :sailboat: 

screen shot 2014-04-18 at 5 45 08 pm

screen shot 2014-04-18 at 5 45 13 pm

Not pictured: same swanky animation we use in the footer drop down. :nail_care:


barryclark:
:+1:

PR #1871, opened 2014-04-21
Signup Data Form: Already submitted

aaronschachter:
Displays the `form_submitted_copy` message on the signup data form if the user has already submitted it, preventing re-submit.

See https://github.com/DoSomething/dosomething/issues/1840#issuecomment-40942430 for details on why previous commits around this functionality/code caused the epic EntityMetadataWrapper error (spoiler: I can't write PHP `define` statements correctly).


blisteringherb:
:+1:

PR #1874, opened 2014-04-21
Campaign block template

aaronschachter:
Creates a themeable `campaign_block` function and template file, and updates the user profile to call them for each recommended  / signup campaign.

TODO:
- Add signup date into the `$vars` if exists
- Remove deprecated `user_signups` view from codebase


blisteringherb:
:+1:

PR #1878, opened 2014-04-28
Drupal.js

DFurnes:
ok this is actually the opposite of what we're doing but come on just look at it this is great

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/config.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/config.js
@@ -12,7 +12,7 @@
     baseUrl: "/profiles/dosomething/themes/dosomething/paraneue_dosomething/js/",
     include: "main",
     paths: {
-      "jquery": "../bower_components/jquery/dist/jquery",
+      "jquery": "../bower_components/jquery/jquery",
       "neue": "../bower_components/neue/neue",
       "mailcheck": "../bower_components/mailcheck/src/mailcheck",
       "lodash": "../bower_components/lodash/dist/lodash",
--- a/lib/themes/dosomething/paraneue_dosomething/js/config.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/config.js
@@ -8,21 +8,18 @@
 (function() {
   "use strict";
 
+
   require.config({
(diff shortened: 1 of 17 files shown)


lib/themes/dosomething/paraneue_dosomething/js/config.js
```diff
@@ -12,7 +12,7 @@
     baseUrl: "/profiles/dosomething/themes/dosomething/paraneue_dosomething/js/",
     include: "main",
     paths: {
-      "jquery": "../bower_components/jquery/dist/jquery",
+      "jquery": "../bower_components/jquery/jquery",
```
aaronschachter: Why is this path changing?

aaronschachter:
This all looks good technically, but are can you think of any gotchas we need to look out for, to make sure this doesn't break something unintentionally?

weerd:
Boo. Looks fine, but yeah, agree with @aaronschachter Keep in mind certain syntax things... like .on() and stuff...

weerd:
Just checked and `.on()` was added as of v1.7 so that should be fine, but we should still double check stuff @_@

weerd:
also @DFurnes :+1: on the image ;)

lib/themes/dosomething/paraneue_dosomething/js/config.js
DFurnes: The jQuery folks changed the path things were bundled up in 1.11.x. So we're movin' back to the old ways.

DFurnes:
Double-checked the 1.9-1.11 release notes and the major changes are just performance improvements and removing deprecated APIs (which is the gotcha that's causing Drupal to hurt, although admittedly the `$.browser` one that it's been crying about has been deprecated since v1.3 c.a. January 2009 so...)

aaronschachter:
not sure what the `$.browser` thing means but hey lets rock it :+1:

DFurnes:
This should be good to merge. Also ended up cleaning up some of the RequireJS configuration so net win. 

:chart_with_upwards_trend:

weerd:
Cleanup looks cool :+1: :sun_with_face:

PR #1880, opened 2014-04-28
User profile  markup

### LINKED ISSUE #1655: [User Profile] Generate markup for user profile page.
@julielorch @mikefantini

weerd:
Includes modifications to the new approach for using a template for the campaign blocks, etc.

Resolves #1655


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -414,7 +414,7 @@ function dosomething_campaign_get_campaign_block_vars($nid, $image_size = '740x4
   }
   return array(
     'nid' => $wrapper->getIdentifier(),
-    'title' => l($wrapper->label(), $path),
+    'title' => $wrapper->label(),
```
weerd, line 417: @aaronschachter I updated the `'title'` in the vars for the campaign block to only be the title text for the campaign, and in the template create a link if needed using `l(...)`. I updated where it's used in the report back tpl, and I saw that the "scholarships" page uses that as well, but I didn't find a tpl for it. Might just be an old function that's currently not in use, but wanted to flag it in case you have more info on it!

aaronschachter, line 417: Yeah the scholarships page is a prototype that @angaither built, that would need theming before it went live anyway.

lib/themes/dosomething/paraneue_dosomething/templates/campaign/campaign-block.tpl.php
```diff
@@ -1,5 +1,19 @@
...
+ * - $vars: An array of campaign variables.
+ *   - [nid]: Node ID for campaign (integer).
+ *   - [title]: Title of campaign (string).
+ *   - [call_to_action]: Call to action text for campaign. (string).
+ *   - [image]: URL path for campaign image (string).
+ *   - [src]: URL path for campaign node (string).
+ */
+?>
+
+<article class="campaign -teaser" id="campaign-<?php print $vars['nid']; ?>">
```
aaronschachter, line 15: Is this space intentional? `campaign -teaser` ?

lib/themes/dosomething/paraneue_dosomething/templates/campaign/campaign-block.tpl.php
```diff
@@ -1,5 +1,19 @@
...
+/**
+ * Returns the HTML for a single Campaign Block.
+ *
+ * Available Variables
+ * - $vars: An array of campaign variables.
+ *   - [nid]: Node ID for campaign (integer).
+ *   - [title]: Title of campaign (string).
+ *   - [call_to_action]: Call to action text for campaign. (string).
+ *   - [image]: URL path for campaign image (string).
+ *   - [src]: URL path for campaign node (string).
```
aaronschachter: Sorta picky but I'd probably call this guy `url` or `path` instead of `src`, which feels like it would be the name of the image source.

aaronschachter:
All comments are pretty minor and easy to refactor later (e.g. the variable name coment).  If that class name is correct, and the reportback confirmation page is working correclty, then good by me :+1:

lib/themes/dosomething/paraneue_dosomething/templates/campaign/campaign-block.tpl.php
weerd, line 15: Yep, intentional. It's a class that acts as a modifier. New system  ;)

lib/themes/dosomething/paraneue_dosomething/templates/campaign/campaign-block.tpl.php
weerd: Cool. Yeah, I felt the same way, but it was called [src] already and didn't want to change it for fear of disasters. I can change it and resubmit.

DFurnes:
Looks good. :+1:

lib/themes/dosomething/paraneue_dosomething/templates/campaign/campaign-block.tpl.php
DFurnes: +1. May be worth using this for the campaign blocks in the "blank slate" on the homepage as well so we don't have the same thing implemented both places.

lib/themes/dosomething/paraneue_dosomething/templates/campaign/campaign-block.tpl.php
DFurnes, line 15: :dancers:

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
weerd, line 417: Cool. Ok, good to know :)

aaronschachter:
variable names look good :+1:

weerd:
Awesome. Imma mergin'!

PR #1888, opened 2014-04-29
Uses AMD Neue interface from Neue 3.5.

DFurnes:
References DoSomething/neue#252.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/bower.json
+++ b/lib/themes/dosomething/paraneue_dosomething/bower.json
@@ -20,7 +20,7 @@
     "tests"
   ],
   "dependencies": {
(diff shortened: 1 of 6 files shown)


weerd:
:+1: Will also test once on stage :)

DFurnes:
So just for future reference, the cool benefit here is that RequireJS will be able to pull in Neue code as individual modules (rather than as one big concatenated file like it was before). This means we can debug an issue issue on, say, line 32 of `modal.js` rather than just knowing there's an error _somewhere_ on line 41352 in the concatenated source. :)

screen shot 2014-04-29 at 11 02 58 am

weerd:
nice :)  good to know!

Issue #1889, opened 2014-04-29
Admin menu environment indicator


aaronschachter:
- [x] Remove environment_indicator module from repo. Adding lots of extra JS we don't want for such a simple function.
- [ ] Borrow code from https://drupal.org/project/simplei

Per Desmond:
i would use a assoc array for the color settings though
https://gist.github.com/desmondmorris/fd79977a13a33bda540a
it will make lines 13:16 much smarter here http://drupalcode.org/project/simplei.git/blob/HEAD:/simplei.module#l5
something like that

PR #1892, opened 2014-04-29
Remove environment_indicator module from profile

aaronschachter:
We tried finally turning it on and it adds all that dang javascript!  Get it out of here.

Refs #1889 -- goal is to add some custom code later to handle indicating environments.


desmondmorris:
ship it

angaither:
looks good, is it in the .info file?

aaronschachter:
nope!  should be good to go.

PR #1905, opened 2014-04-29
Restores IE8 polyfills.

DFurnes:
Brings back IE 8 polyfills, since they're not being included in base Neue build now.

:truck:


weerd:
:+1:

PR #1911, opened 2014-04-30
Fixed registration link fallback URL.

### LINKED ISSUE #1900: User registration link goes to 'access denied'
The 'Create a new account' link when opened in a non-modal goes to /user/registration

it should go to user/register

DFurnes:
Fixes #1900.


aaronschachter:
:+1:

PR #1919, opened 2014-04-30
Collect and save why a user signs up

### LINKED ISSUE #1912: Signup Data Form: Collect/display why_signed up from user
Blocked by #1903 

If the signup data form is configured to collect why the user signed up, add it as a form element into the `dosomething_signup_user_signup_data_form` and save to the signup entity's `why_signedup` property.

Also, display the `why_signedup` value when viewing a signup entity.

aaronschachter:
Fixes #1912


angaither:
:+1:

PR #1922, opened 2014-04-30
Some initial very basic styles for the User Profile edit form.

### LINKED ISSUE #1918: [User Profile] Style the profile edit form with some basic styles.
This should really be addressed globally for all default pages, but it's going to be a significant refactor push. Instead will attempt to address just these form pages for now and then move the code to apply globally during a refactoring sprint.

weerd:
This is ugly... but only temporary. Will be refactoring the global default pages soon and this page will be addressed along with that push. :grimacing:

Resolves #1918

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_user.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_user.scss
@@ -10,6 +10,55 @@
 }
 
 
+// @TODO: Refactor and incorporate into default page styles.
+// I know this is fugly, but it's only temporary! Who are you to judge!?!
+.page-user-edit {
+  .chrome--nav {
+    a {
+      color: #222 !important;  // @TDOD: blarg!
+    }
+
+    .hamburger {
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/scss/content/_user.scss
```diff
@@ -10,6 +10,55 @@
 }
 
 
+// @TODO: Refactor and incorporate into default page styles.
+// I know this is fugly, but it's only temporary! Who are you to judge!?!
```
DFurnes, line 14: judgment

DFurnes:
:+1:

lib/themes/dosomething/paraneue_dosomething/scss/content/_user.scss
weerd, line 14: Ok. If you have a ginormous machine gun and look like you can kill a guy with your pinky... then fine. Judge all you want. :sun_with_face:

PR #1923, opened 2014-04-30
Fix for the black background issue on search results page.

### LINKED ISSUE #1775: Search results background swap to white
Let's change the background of search results to white to be consistent with static content styling. I would love to spend sometime thinking about search/results UX and design but its already solid aside from the dark background. 

Please also drop the extra search field that appears on the results page and just keep the search term in the nav search field. 

If you need help talking through how this affects mobile just grab me. I think a page with results and making the user go back into their nav to search again makes the most sense for now. 

screen shot 2014-04-14 at 1 19 32 pm

weerd:
Quickie :sunglasses: 

Resolves #1775

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_search.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_search.scss
@@ -10,6 +10,7 @@
 
       @include span-columns(8);
       @include shift(4);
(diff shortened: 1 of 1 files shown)


barryclark:
:+1: gogogogogo :racehorse: :racehorse: :racehorse: :racehorse:

PR #1934, opened 2014-05-01
Signup: Response data

### LINKED ISSUE #1920: Signup Data Form: Need ability to skip form
As a user I want the ability to skip the required form I get presented with.

### LINKED ISSUE #1924: Signups: Data Form Response column
Add a new column `signup_data_form_response` to store how users respond to the signup data form.  When data team wants to pull which users have requested thumbsocks, banners, etc. - they need to pull all signups where the response = 1.
- [x]  Add an int column `signup_data_form_response` to the `dosomething_signup` table, with default NULL.
- [x]  For any existing signup record with a `signup_data_form_timestamp`, update the `signup_data_form_response` column to 1.
- [x]  Upon submitting a signup_data_form, store 1 if user submits the form but 0 on "Skip" response.
- [x]  Upon viewing a signup_data_form again, if the response=1, show the `form_submitted_copy`. If response=0, user should be allowed to submit form to "opt in" and ask for a banner.

Illustration:
photo

aaronschachter:
Fixes #1924
Fixes #1920

Allows configuration of a signup data form to allow a user to hit a "Skip" button.  This response is stored in the `signup_data_form_response` column, so that we don't prompt a user to "Skip" again when they view the form in the modal.  It's a one time jam.

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
@@ -64,9 +64,21 @@ function _dosomething_signup_node_signup_data_form(&$form, &$form_state) {
   );
   $form[$fieldset]['config'][$prefix . 'required'] = array(
     '#type' => 'checkbox',
-    '#title' => t('Display after signup'),
+    '#title' => t('Prompt after signup'),
     '#default_value' => $values['required'],
     '#description' => t('If checked, the form modal will be displayed after the user first signs up.'),
+  );
+  $form[$fieldset]['config'][$prefix . 'required_allow_skip'] = array(
+    '#type' => 'checkbox',
+    '#title' => t('Allow skip'),
+    '#default_value' => $values['required_allow_skip'],
+    '#description' => t('If checked, the form modal will provide a Skip button when the user is first prompted.'),
+    // Should only be visible if form is required.
+    '#states' => array(
+      'visible' => array(
+        ':input[name="' . $prefix . 'required"]' => array('checked' => TRUE),
+      ),
+    ),
   );
   $form[$fieldset]['config'][$prefix . 'link_text'] = array(
     '#type' => 'textfield',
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -64,10 +64,22 @@ function _dosomething_signup_node_signup_data_form(&$form, &$form_state) {
...
+    '#title' => t('Prompt after signup'),
     '#default_value' => $values['required'],
     '#description' => t('If checked, the form modal will be displayed after the user first signs up.'),
   );
+  $form[$fieldset]['config'][$prefix . 'required_allow_skip'] = array(
+    '#type' => 'checkbox',
+    '#title' => t('Allow skip'),
+    '#default_value' => $values['required_allow_skip'],
+    '#description' => t('If checked, the form modal will provide a Skip button when the user is first prompted.'),
+    // Should only be visible if form is required.
```
angaither, line 76: A little confused here, if signup data is required, filling out the form can be skipped?

angaither:
Looks ok technically though

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
aaronschachter, line 76: When the form is set as required, it means the user must be prompted with it upon signup.  

If required_allow_skip, they are able to click "Skip" -- which allows us to store the fact that they did submit the form, they just didn't opt-in to get things sent to their address.

angaither, line 76: Phew. So we would be collecting like "why signed up" but not their address, because they don't want socks?

aaronschachter, line 76: If you click Skip, we don't even store "why signed up", just store the timestamp: https://github.com/aaronschachter/dosomething/blob/signup_data_form_response/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc#L451

PR #1942, opened 2014-05-01
Signup data form fix.

aaronschachter:
Adding the extra conditional means needing some parenthesis for `$timestamp` to be set correctly.


angaither:
:+1:

Issue #1913, opened 2014-04-30
Signup Data Form: Prompt user after signup


aaronschachter:
Blocked by #1904

If a signup data form is required, present the signup data form modal when the user first visits the action page.

For now, this will be done in somewhere in the theme template.php checking to see if the user has submitted the modal (by checking the timestamp).  If not, `drupal_add_js`: https://gist.github.com/aaronschachter/304bb24b0d5d81ac17b8

aaronschachter:
Assigning to @DFurnes and changing to "fender" -- all that's left is getting the modal to render once you visit a campaign and haven't submitted the signup form.  I've set up http://staging.beta.dosomething.org/campaigns/thumb-wars with the proper signup data form configuration, you can see the inline JS in the footer.

DFurnes:
Found the problem, `$` is not defined in the global scope because Drupal forces jQuery into `noConflict` mode. Switching it to this should fix it:

``` js
require(["neue/modal"], function(Modal) { Modal.open( jQuery("#modal-signup-data-form"), false); });
```

I'll throw in a PR in a minute.

PR #1945, opened 2014-05-02
Prompt signup data form

aaronschachter:
Refs #1913

Abstracts all variable logic for the signup_data_form into a new function, `dosomething_campaign_add_signup_data_form_vars($node)`.

Stores a flag `$node-> required_signup_data_form` which indicates the signup data form should be displayed when a user is viewing an action page.  Adds inline JS into the theme if this flag is set.

@DFurnes Tried the second code snippet but still no luck.  This PR sets up all the variables in order to know when to prompt the user, will pass #1913 over to you to actually get the JS working to automatically open the modal.


blisteringherb:
:+1:

PR #1955, opened 2014-05-02
Removing unnecessary fields and grouping for better usability

### LINKED ISSUE #1954: [Campaign Group] Remove unnecessary fields and rearrange
The image and video fields are unnecessary because the sponsors field collection already has those fields.  Also the Pre signup and post signup fields should be grouped.

blisteringherb:
Fixes #1954

@aaronschachter


aaronschachter:
Looks good :+1:

PR #1958, opened 2014-05-02
User profile  design updates

### LINKED ISSUE #1929: [User Profile] Style cta for when no signups showing.

weerd:
Lots of updates to design.

Resolves #1929

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_user.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_user.scss
@@ -168,6 +168,11 @@
     .subtitle { margin-bottom: 0; } // @TODO: temp for now; apply universally.
   }
 
+  // @TODO: Abstract out as a pattern to Neue
+  // @CTA (refernced to later remove it when CTA is patternized)
+  .cta {
+    @include cta();
+  }
 
   section {
     padding-bottom: 1rem;
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_user.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_user.scss
@@ -60,24 +60,38 @@
 }
 
 
-// @TODO: Move to Neue.
+// @TODO: Campaign Tile. Move to NEUE.
 .tile {
+  height: 0;
+  padding-bottom: 100%;
   position: relative;
 
   > .wrapper {
     background-color: #fff;
     display: block;
     height: 100%;
-    min-height: 200px;
+    left: 0;
+    overflow: hidden;
+    position: absolute;
+    top: 0;
     width: 100%;
 
     &:before {
+      bottom: 0;
       content: '';
       display: block;
       height: 100%;
       position: absolute;
       width: 100%;
       @include linear-gradient(rgba(#000, 0) 40%, rgba(#000, .2) 70%, rgba(#000, .5) 87%, rgba(#000, .85) 100%);
+    }
+  }
+
+  .no-cssgradients & {
+    > .wrapper {
+      &:before {
+        background: transparent image-url('black-gradient.png') 0 bottom repeat-x;
+      }
     }
   }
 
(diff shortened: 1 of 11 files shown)


lib/themes/dosomething/paraneue_dosomething/scss/content/_user.scss
```diff
@@ -81,6 +87,14 @@
     }
   }
 
+  .no-cssgradients & {
```
DFurnes, line 90: Is this something we could do with a normal CSS fallback in a mixin instead of Modernizr?

``` scss
$gradient: rgba(#000, 0) 40%, rgba(#000, .2) 70%, rgba(#000, .5) 87%, rgba(#000, .85) 100%;
background: transparent image-url('black-gradient.png') 0 bottom repeat-x;
@include prefixer(linear-gradient, $gradient, webkit moz spec);
```

Just seems like a lot of selector action.

lib/themes/dosomething/paraneue_dosomething/scss/content/_user.scss
```diff
@@ -121,21 +140,102 @@
...
-      width: 33.333333333%;
+    @include media($tablet) {  // @TODO: custom mq; address when more mq's added.
+      width: 25%;
+    }
+  }
+
+
+  // Style for empty gallery item to help balance the space.
+  .empty {
+    background-color: lighten($light-gray, 10%);
```
DFurnes, line 165: Why are we using a custom gray here? If it's necessary, is this something that we should be standardizing as an `$off-white` or something like that?

lib/themes/dosomething/paraneue_dosomething/scss/content/_user.scss
weerd, line 90: Wouldn't you be doing an extra HTTP request on modern browsers for a PNG they don't need since they support gradients though?

lib/themes/dosomething/paraneue_dosomething/templates/user/user-profile.tpl.php
```diff
@@ -11,21 +11,32 @@
...
 
-  $signedup = $user_account['campaigns_signedup'];
-  $recommended = $user_account['campaigns_recommended'];
+  $signedup = $user_account['signedup'];
+  $signedup_count = $user_account['signedup_count'];
+  $recommended = $user_account['recommended'];
   $address = $user_account['address'];
+
+  // $signedup = null;
+  // krumo($user_account);
```
DFurnes: :goat:

lib/themes/dosomething/paraneue_dosomething/scss/content/_user.scss
weerd, line 165: It was used on the CTA a while back. I think it would help with not going too crazy on lots of differently named grays if we just use the variables for the grays specified, but if an area calls for a slightly lighter or darker version of said gray (for example, hovers could be one), it's just using a Sass function with a specified percentage.

lib/themes/dosomething/paraneue_dosomething/templates/user/user-profile.tpl.php
weerd: oops.
I thought I gots rid of those. hrm.

lib/themes/dosomething/paraneue_dosomething/scss/content/_user.scss
weerd, line 90: it's a bit of selector action, but only really ends up applying to IE8 ;)

DFurnes, line 90: It shouldn't make the HTTP request if the rule is overridden, I think.

weerd, line 90: Hmmm, not sure. Modernizr is for this kind of stuff though. Allowing you to target via browser capabilities/support. Plus, once we no longer need to support IE8, all you need to do is remove any declarations using `.no-`.

Issue #1811, opened 2014-04-15
Signup refactor: Entity functions


aaronschachter:
Refs #1810.  Would be great to get in for #1887 
- [x] Implement `entity_create` in `dosomething_signup_insert`
- [ ] Implement `entity_delete` in `dosomething_signup_delete`
- [x] Implement `hook_entity_insert` instead of check in `dosomething_signup_user_signup`
- [x] Implement `entity_save` in `dosomething_signup_user_signup_data_form_submit` / `dosomething_signup_update_signup_data`
- [x] Pass a signup `$entity` as a param to `dosomething_signup_user_signup_data_form` instead of `$nid`

aaronschachter:
Having issues trying to use `entity_create`, it's inserting a record for nid = 1 when i am using `entity_create('signup', $values)`.  Investigating.

aaronschachter:
The issues were around naming the function `dosomething_signup_insert`.  It was looking for modules which invoke `signup_insert`.  Changing the function name to `dosomethign_signup_create` fixed.  Same for `dosomething_signup_delete` - was causing errors until changing the name to `dosomething_signup_delete_signup`

Issue #1896, opened 2014-04-29
Campaign Group: Child signup should also signup for Group


aaronschachter:
When signing up for a child campaign of a Campaign Group, a signup entity should also be created for the parent Campaign Group, in order to opt-in the user for the 3rd party communications of the Campaign Group.

aaronschachter:
@mikefantini @barryclark  Does the campaign group appear on a user profile? Or should the profile filter out Campaign Group signups?

mikefantini:
@aaronschachter campaign group will not appear on the user profile

PR #1964, opened 2014-05-05
dosomething_signup_create

aaronschachter:
Refs #1811

Uses `entity_save` instead of a `db_insert` to tap into Entity goodness later.

There was a namespace collision with the function name `dosomething_signup_insert`, where it was calling a `hook_insert` and inserting a signup with nid =1, so the function name has been globally changed to `dosomething_signup_create`.


aaronschachter:
Updated PR title to actually make sense.  This doesn't add any new functionality but sets up to be able to properly use `hook_entity_insert` for #1896

aaronschachter:
@blisteringherb Would you mind reviewing this one?  Moving to the entity API functions to be able to make use of `hook_entity_insert` for #1896

blisteringherb:
:+1:

PR #1965, opened 2014-05-05
Dependency man

DFurnes:
## Changes:
- Includes Neue's polyfills using Bower rather than using the versions bundled with Neue.
- Conditionally load polyfills using feature detection.

dependencyman


weerd:
:+1: Coooool

weerd:
:sun_with_face:

PR #1970, opened 2014-05-05
Updates after moving some styles over to Neue

weerd:
@DFurnes, deleting all the things!


DFurnes:
Cool cool cool. :sailboat:

PR #1971, opened 2014-05-05
Reportback view: Expose email as a filter

aaronschachter:
Adds email as exposed filter for the `node/*/reportbacks` view.  

Also adds it to `node/*/signups` just because.

@angaither Would you mind reviewing this one?


angaither:
:+1:

PR #1974, opened 2014-05-05
Adding a field to determine whether the signup form is displayed

### LINKED ISSUE #1961: Campaign Group: Display Signup Form
Create a new field `display_signup_form` to determine whether or not the Campaign Group Signup Form should be displayed on the Campaign Group template.

This will be used instead of the "are any child campaigns published?" check.

blisteringherb:
Fixes #1961


lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.features.field_instance.inc
```diff
@@ -302,6 +302,55 @@ function dosomething_campaign_group_field_default_field_instances() {
...
+      'teaser' => array(
+        'label' => 'above',
+        'settings' => array(),
+        'type' => 'hidden',
+        'weight' => 0,
+      ),
+    ),
+    'entity_type' => 'node',
+    'field_name' => 'field_display_signup_form',
+    'label' => 'display_signup_form',
```
aaronschachter: Picky, but this will come up later. Field label should be human readable.

PR #1995, opened 2014-05-06
Refactoring  using new neue patterns

weerd:
@DFurnes @barryclark Check it!

### DIFF
--- a/lib/modules/dosomething/dosomething_static_content/dosomething_static_content.module
+++ b/lib/modules/dosomething/dosomething_static_content/dosomething_static_content.module
@@ -48,7 +48,7 @@ function dosomething_static_content_preprocess_node(&$vars) {
               $vars[$label] = dosomething_image_get_themed_image($vars[$field][0]['entity']->nid, 'landscape', $size);
             break;
             case 'link':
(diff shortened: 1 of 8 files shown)


lib/modules/dosomething/dosomething_static_content/dosomething_static_content.module
```diff
@@ -48,7 +48,7 @@ function dosomething_static_content_preprocess_node(&$vars) {
               $vars[$label] = dosomething_image_get_themed_image($vars[$field][0]['entity']->nid, 'landscape', $size);
             break;
             case 'link':
-              $vars[$label] = l($content[$field][0]['#element']['title'], $content[$field][0]['#element']['url'], array('attributes' => array('class' => 'btn')));
+              $vars[$label] = l($content[$field][0]['#element']['title'], $content[$field][0]['#element']['url'], array('attributes' => array('class' => 'btn medium')));
```
DFurnes: `.medium` is the default size for buttons, so this shouldn't be necessary unless something else is overriding the styles somewhere.

DFurnes:
Looks solid aside from the one comment. :muscle:

lib/modules/dosomething/dosomething_static_content/dosomething_static_content.module
weerd: ah. good to know.

PR #1997, opened 2014-05-06
Neue 3.8

DFurnes:
Updates for Neue 3.8, address form style updates, and includes Modernizr as a separate script in head to fix some intermittent issues with classes not being applied.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/system/html.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/system/html.tpl.php
@@ -12,6 +12,7 @@
 
 <head>
   <title><?php print $head_title; ?></title>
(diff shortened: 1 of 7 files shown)


lib/themes/dosomething/paraneue_dosomething/templates/system/html.tpl.php
```diff
@@ -12,6 +12,7 @@
 
 <head>
   <title><?php print $head_title; ?></title>
+  <script type="text/javascript" src="/<?php print PARANEUE_DS_PATH; ?>/bower_components/neue/js/vendor/modernizr.js"></script>
```
weerd, line 15: Nice. Hopefully this solves those issues! :sun_with_face:

weerd:
:+1:

PR #1998, opened 2014-05-06
Campaign Groups: Signup upon login/register

aaronschachter:
- Creates a `dosomething_signup_get_login_signup_nid` function which is called from within the user login/register forms, to determine whether or not user is logging in from a node which should create a signup.
- Only creates signup on login for a Campaign Group if the `field_display_signup_form` is checked.
- Displays the `field_signup_confirm_msg` after signup, if the field exists on the node.

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -155,7 +155,7 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
       array_unshift($form['#validate'], 'dosomething_user_login_validate');
       $form['#submit'][] = 'dosomething_user_login_submit';
       // Helper text & additional data.
-      _dosomething_user_add_campaign_data($form);
+      _dosomething_user_add_signup_data($form);
 
     break;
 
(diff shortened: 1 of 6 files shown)


barryclark:
@blisteringherb if you get the chance to review this before eod that'd be sweet. @aaronschachter and @angaither are both in a meeting just now.

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -419,19 +419,17 @@ function dosomething_user_user_insert(&$edit, &$account, $category = NULL) {
...
-        '#value' => $obj->nid,
-        '#access' => FALSE,
-      );
-      $form['#action'] = '/' . drupal_get_path_alias('node/' . $obj->nid);
-    }
+function _dosomething_user_add_signup_data(&$form) {
+  if (!module_exists('dosomething_signup')) { return; }
+
+  // If we're viewing a node to signup for:
+  if ($nid = dosomething_signup_get_login_signup_nid()) {
```
angaither, line 426: can't you pass the form into this function to see the node type instead of doing a call to `menu_get_object` ?

aaronschachter, line 426: The form is the user_login / user_registration form though, so there won't be a node type. This function has been renamed from `dosomething_user_add_campaign_data`, since it needs to check for more than just a campaign, but also a campaign group.  (and in the future, a SMS game when that gets built)

aaronschachter:
@angaither @blisteringherb is this good to go?

blisteringherb:
:+1:

PR #1669, opened 2014-04-07
Adds the beginnings of a release shell script

desmondmorris:
# DEPRECATED

See #1999
#### What's this PR do?

Provides a bash script for creating new tagged releases.
#### Where should the reviewer start?

**[./scripts/release.sh](https://github.com/desmondmorris/dosomething/blob/2e94d27d2e17cb2adbf32e16f9641f41c441612d/scripts/release.sh)** - This is all of the logic.  Pretty straight forward
**[./VERSION](https://github.com/desmondmorris/dosomething/blob/2e94d27d2e17cb2adbf32e16f9641f41c441612d/VERSION)** - The version marker.  Right now we will handle bumps here manually
#### How should this be manually tested?
1.  Manually bump the version number in VERSION
2.  Run `./scripts/release.sh`
3.  Profilt

Do not worry, we can always remove these test tags later.
#### What are the relevant tickets?
#1574

@angaither @aaronschachter

PR #1574, opened 2014-03-28
Adding a check for drush when calling drupal_goto in hook_init

blisteringherb:
- Drush doesn't like drupal_goto in hook_init().
  https://drupal.org/node/1230132

Fixes #1568

PR #1999, opened 2014-05-07
Adds branch argument to enable passing of tags to capistrano

desmondmorris:
#### What's this PR do?

This pull request adds logic to set the default branch that capistrano uses via a command line argument.  The branch variable can also be set to a tag.  This will allow us to deploy from specific tags.

https://github.com/desmondmorris/dosomething/blob/d800ff0223fb3b29ae7c3be1f2b8e8e4698360c7/config/deploy.rb#L7-L10

We should run a deployment to the staging environment via the command line.
`capistrano staging deploy branch=v0.2.2`

Once this is in place, our Hubot deployment scripts must be updated to accept a tag parameter
#1574 #1669

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -3,7 +3,12 @@ lock '3.1.0'
 
 set :application, 'beta.dosomething.org'
 set :repo_url, 'git@github.com:DoSomething/dosomething.git'
+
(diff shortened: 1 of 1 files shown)


aaronschachter:
Looks good :+1:

Issue #1987, opened 2014-05-06
[Campaign Group] Node form: Signup Form fields, groups


aaronschachter:
- [x] Rename the "Call to Action" field group as "Signup Form".  Place the `field_display_signup_form` field into this group as well.
- [x] Hide the "CTA Link URL" field: this will never be used.

PR #2003, opened 2014-05-07
Renaming CTA group to Signup Form and moving display checkbox

blisteringherb:
Refs #1987


aaronschachter:
This looks good.  #1987 still has openitem about the CTA link though, so i'm going to alter the description to not "fix"

PR #2005, opened 2014-05-07
Falls back to normal image upload field on IE8.

DFurnes:
IE8 doesn't allow changing the value of `input[type="file"]` fields using JavaScript, which was causing errors on report backs for IE8 users. Also makes file field a bit easier to read for the way IE8 renders the filename portion of the field.

Fixes https://trello.com/c/b5JlKZEp/1125-having-a-hard-time-reporting-back-for-pb-jam-slam

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/campaign/ImageUploader.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/campaign/ImageUploader.js
@@ -69,7 +69,10 @@ define(function(require) {
 
   // When we open a modal, we prepare any `.js-image-upload`s that we find there
   Events.subscribe("Modal:opened", function(topic, args) {
-    prepareImageUploadUI(args);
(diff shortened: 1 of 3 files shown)


weerd:
While this might be frowned upon, this is a perfect case for IE conditional comments.

`<!--[if IE 8]>       <html class="no-js lt-ie9 ie8" lang="en"> <![endif]-->`

Then in the js, instead of browser detecting, can just use `if ($('html').hasClass('.ie8')...`

DFurnes:
Hmm, yeah. I feel more comfortable doing that though than user-agent sniffing. Especially since `$.browser` is deprecated and removed in jQuery 1.9+.

weerd:
Huzzah! Cool. Give it the ol' switcheroo and I'll take another look!

weerd:
:+1: Looks good señor!

Issue #1960, opened 2014-05-02
[Static content] need the ability to add a partner to static content


mikefantini:
Should work the same way as it does on campaigns (although for now we don't need fields for "Why is x sponsoring" so will only need partner name and whether they are a sponsor or not).

If it's a sponsor their logo should appear in the header AND we should have their name with "In partnership with" at the bottom.

If it's not a sponsor their name should just appear along with "In partnership with" at the bottom.

Need this ready and tested for May 27th launch of the Hunt

aaronschachter:
This requires adding the `field_partners` field collection to the Static Content node type, and using similar / same functions that campaign use to display the header and footer.

barryclark:
Wanna take care of this one, @aaronschachter? ... and shoot it over to @weerd for the front end after.

aaronschachter:
Assigning to @weerd - ready for fending.

mikefantini:
@weerd yeah tried adding sponsor to http://staging.beta.dosomething.org/about/hunt-member-created-campaign-challenge and logo isn't showing up

can you take a look once other grouped campaigns stuff is in a better place?

weerd:
Partner shows up in footer now. Working on the login in the header :)

Also, didn't mean to close the issue. Accidentally pressed the wrong button :(

mikefantini:
@weerd still not seeing sponsor logo in header. See http://staging.beta.dosomething.org/about/hunt-create-challenge

mikefantini:
@weerd also doesn't seem to be showing up along the bottom either

weerd:
@mikefantini  doh! sorry. I was looking at grouped campaigns. My bad. I'll do an update on this today and add it to static content as well! Got confused cause it was labeled with "grouped campaigns" :sun_with_face:

weerd:
http://staging.beta.dosomething.org/about/hunt-create-challenge link to confirm!

mikefantini:
yup! confirmed... thanks @weerd

PR #2006, opened 2014-05-07
Partners in static content

aaronschachter:
@blisteringherb Can you please review?

Completes bender work for #1960 -- Adds the `field_partners` field collection to the Static Content type. Ready for fending.

Moves the `dosomething_helpers_format_partners_list` into `dosomething_helpers_preprocess_partners_vars` so we only need to call the one function, `dosomething_helpers_preprocess_partners_vars`.


blisteringherb:
:+1:

PR #2009, opened 2014-05-07
Replacing CTA link with text field and adding removing CTA link field

### LINKED ISSUE #2007: [Campaign Group] Use a new field for the signup form
The URL field on a link field isn't easily hidden and requires a lot more code than makes sense.

blisteringherb:
Fixes #2007, #1987


angaither:
:+1:

lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.install
```diff
@@ -0,0 +1,13 @@
...
+ * @file
+ * Installation and schema hooks for dosomething_campaign_group.module.
+ */
+
+
+/**
+ * Deletes field_cta_link.
+ */
+function dosomething_campaign_group_update_7001(&$sandbox) {
+  field_delete_field('field_cta_link');
```
aaronschachter: This should be `field_delete_instance` ? `field_delete_field` will kill from static content: https://api.drupal.org/api/drupal/modules!field!field.crud.inc/function/field_delete_field/7

Please ping when fixed for re-review

PR #2010, opened 2014-05-07
Add version numbers to features.

### LINKED ISSUE #1986: Deployments not always reverting features
Ran ds build stage, and features are overridden.

screen shot 2014-05-06 at 11 17 37 am

angaither:
This will fix the features not reverting automatically.
When the feature was listed with a versoin number it changed the arg of the title of the module.
Adding a version number to all will make reverts consistent.
Also updated the custom features to all be in the `DoSomething` features group

Fixes #1986

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -162,7 +162,7 @@ function build {
   drush --script-path=../bin/ php-script enable_modules.php
 
   echo 'Reverting Features'
-  drush fl | grep 'Overridden' | awk '{print ($(NF-2))}' | xargs -i drush fr {} --force -y
+  drush fl | grep 'Overridden' | awk '{print ($(NF-3))}' | xargs -i drush fr {} --force -y
 
   echo 'Clearing caches...'
   drush cc all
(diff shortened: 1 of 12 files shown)


bin/ds
```diff
@@ -162,7 +162,7 @@ function build {
   drush --script-path=../bin/ php-script enable_modules.php
 
   echo 'Reverting Features'
-  drush fl | grep 'Overridden' | awk '{print ($(NF-2))}' | xargs -i drush fr {} --force -y
+  drush fl | grep 'Overridden' | awk '{print ($(NF-3))}' | xargs -i drush fr {} --force -y
```
aaronschachter, line 165: What is this change doing?

angaither, line 165: It changes the argument number for the module name

aaronschachter, line 165: ok. so if we make changes to features, do we always need to update the version number?

aaronschachter:
lets shippy ship :+1:

PR #2013, opened 2014-05-07
Partner footer links

aaronschachter:
Adds a new field, `field_partner_url` into the `field_partners` field collection.  

Modifies `dosomething_helpers_format_partners_list` so that if the `field_partner_url` is set, render the Partner name as a link.


aaronschachter:
@angaither Would you mind reviewing?

angaither:
Looks ok, but might be nice to add either placeholder, or description text to add the expected format.

PR #2015, opened 2014-05-07
Refactoring faq, partner and more facts modals to use a common tpl

### LINKED ISSUE #2000: [Campaign Group] FAQ content does not appear in modal
Ideally we should create a themeable function for the FAQ field, to reuse code between the Campaign and Campaign Group content types.

blisteringherb:
- Creating a tpl for faq, partner and more facts

Fixes #2000


aaronschachter:
This is really awesome, @blisteringherb . As long as all looks good on campaigns (can test on Comeback Clothes for all the modals), lets ship it :+1:

PR #2030, opened 2014-05-08
Add permission to edit home page content.

angaither:
The homepage stuff was pushed yesterday, without this permission.


aaronschachter:
:+1:

PR #2031, opened 2014-05-08
Grouped campaigns

aaronschachter:
Changes readable name of `campaign_group` to Grouped Campaign, adds a README to document Signups.


angaither:
:+1:

PR #2033, opened 2014-05-08
Zendesk form for Grouped Campaigns

### LINKED ISSUE #1898: [Grouped campaign] Zendesk forms
The Campaign Group will need to link to a Zendesk modal.  If a Zendesk group exists for the Campaign Group NID, the ticket should be created in that group.
- [x] Create variable storage for `dosomething_zendesk_nid_[NID]_group_id`
- [x] Create zendesk ticket for the variable if it exists, else use cause group

aaronschachter:
Fixes #1898
- Adds form input on the Zendesk admin form to store zendesk group_ids for `campaign_group` nodes 
- Alters the `dosomething_zendesk_form` to first check to see if a zendesk group_id variable exists for the current node nid.  if so, passes this value as the group_id for create the Zendesk ticket.  if not, checks for the node's primary cause, if it exists (which is the default behavior for campaigns)

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.module
+++ b/lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.module
@@ -85,6 +85,18 @@ function dosomething_campaign_group_preprocess_node(&$vars) {
 }
 
 /**
+ * Implements hook_node_view().
+ */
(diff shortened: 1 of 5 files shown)


aaronschachter:
@blisteringherb Can you please review, and merge if good to go?

blisteringherb:
:+1:

PR #2035, opened 2014-05-08
Renaming pre signup fields and adding logic to display when campaigns are published

### LINKED ISSUE #2021: [Campaign Group] Rename pre-signup fields
- [x] Rename field group as "Pre Live"
- [x] These fields should only be displayed in no Campaigns are published.  See #1993
- [x] Rename "Challenges Title" and "Challenges" to Re-Live header and Pre-Live copy.

blisteringherb:
Fixes #2021

@angaither would you mind taking a look at this?


angaither:
:+1:

PR #2037, opened 2014-05-08
Updating user profile link to use standard Drupal link function.

### LINKED ISSUE #2019: [user profile] Use l function for link to user profile
Link to user profile should print `l($name, 'user/'. $user->uid)` instead of using `<a href="/users/<?php print $user->uid ?>`

Linking to "users/[uid]" doesn't work for any user with a readable username (like staff members), and will break things if we ever change that path alias.

weerd:
@aaronschachter 

Resolves #2019


barryclark:
@aaronschachter can you review this guy plz. ty

aaronschachter:
looks good. all these links should be using the l function tho like logout.

barryclark:
@weerd looks good to merge

weerd:
@aaronschachter cool. I'll update those as well when I go back to working on the nav stuff. :sun_with_face:

PR #2044, opened 2014-05-08
Turning off Markdown for all textfields

### LINKED ISSUE #2025: [Grouped Campaigns] Change intro title so it defaults to plain text.
@aaronschachter adding you to this but not sure if it should be for Morgan?

Anyway to change it so there is NO option to change it to markdown? We can't have <p> tags outputting inside of the heading tags...

blisteringherb:
Fixes #2025


angaither:
:+1:

PR #2046, opened 2014-05-08
Refactoring and adding color settings field to campaign

### LINKED ISSUE #2041: [Grouped Campaigns] alt color custom style field needed
This would allow campaign lead to change the color of H2's on the page.

Same as we have on campaign node:

image

blisteringherb:
Fixes #2041


angaither:
:+1:

PR #2050, opened 2014-05-08
Adding logic for displaying campaigns and pre-live fields

### LINKED ISSUE #1993: [Campaign Group] Logic for displaying Pre-Live fields vs field_campaigns
Display the campaigns as grid if at least one child campaign is live.

If not, display the pre-live fields.

blisteringherb:
Fixes #1993

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.module
+++ b/lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.module
@@ -15,14 +15,21 @@ function dosomething_campaign_group_preprocess_node(&$vars) {
         'subtitle',
         'intro_title',
         'intro',
-        'additional_text_title',
-        'additional_text',
         'signup_form_button_text',
         'post_signup_title',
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.module
```diff
@@ -80,8 +87,6 @@ function dosomething_campaign_group_preprocess_node(&$vars) {
     }
   }
 
-  $vars['campaigns'] = dosomething_campaign_group_get_campaigns($vars['node']);
```
angaither, line 83: don't need this anymore?

angaither:
If that is meant to go, looks good.  :+1:

lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.module
blisteringherb, line 83: I moved it up to ln 25.

PR #2056, opened 2014-05-09
Checking for signup status to display button and fields

### LINKED ISSUE #1992: [Campaign Group] Display post-signup fields based on signup status
Currently displays regardless of signup status.

blisteringherb:
Fixes #1992

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.module
+++ b/lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.module
@@ -30,7 +30,7 @@ function dosomething_campaign_group_preprocess_node(&$vars) {
     }
 
     $sid = dosomething_signup_exists($vars['node']->nid);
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.module
```diff
@@ -31,6 +29,12 @@ function dosomething_campaign_group_preprocess_node(&$vars) {
       unset($vars['campaigns']);
     }
 
+    $sid = dosomething_signup_exists($vars['node']->nid);
+    if (!empty($sid)) {
```
aaronschachter: I believe `dosomething_signup_exists` returns NULL if doesnt exist, just checking: `!empty` the same?

aaronschachter:
As long as tested with `empty` for not having a signup, looks good :+1:

Issue #2017, opened 2014-05-07
Set up AJAX validation for fields/field sets


DFurnes:
Set up a re-usable function that sends field names/values to the server, and accepts a response of field names & success/suggestions.

Example:

POST to `https://dosomething.org/api/validate/address` with data `street="30 Kimblery St"`...
JSON\* Response: `{ "street": { success: false, suggestion: "30 Kimberly Drive", ... }}`

*or XML. :weary: 

/cc @angaither

PR #2057, opened 2014-05-09
Return ups results in menu hook differntly.

angaither:
Update the menu hook for start of address validation improvements
- parse the xml from UPS
- print it as json

Refs #2017


aaronschachter:
looks good

PR #2060, opened 2014-05-09
Fieldset address

DFurnes:
:rabbit:


DFurnes:
Merging this in so we have things ready for demo tomorrow. Shouldn't be anything too dangerous. :cactus:

Issue #1915, opened 2014-04-30
Campaign: Pitch page label


aaronschachter:
Display a custom value for a Signup form label.

For The Hunt, we may (or definitely?) need the ability to change the pitch page ClickyClickyBlue button label to read something like "Accept Challenge."

Initially we were thinking this would be set on the Campaign Group, however @blisteringherb brought up a good point that perhaps its just a setting on an actual Campaign node.. so any Campaign could potentially have a different Pitch page label.

aaronschachter:
This may make sense to store as a variable instead of creating a new field, similar to how to we store the `alt_color` and `alt_bg_file` variables in the "Custom Settings" section on the campaign node form.

aaronschachter:
Updated issue description: we now have the ability to store the value now in #2061, this issue will be closed once the Signup form is configured to read from it (or use a default if not set)

PR #2061, opened 2014-05-09
Adding pitch page label to campaign and campaign group custom settings

blisteringherb:
Refs #1915


weerd:
I think it looks good :+1: :sun_with_face:

aaronschachter:
My comment seems to have been lost in this conversation.. it read: 

Code looks good to ship. Updating issue description because it doesn't close issue #1915 yet: The pitch page button label still needs to use this value (if it exists). If doesn't exist, should default to current value ("Sign Up")

blisteringherb:
@aaronschachter Good point. Thx for updating.

PR #2062, opened 2014-05-09
Making images clickable for published campaigns

### LINKED ISSUE #1897: Campaign Group: "Teaser" display of unpublished child campaigns
Related to #1993 

If a child campaign is unpublished, its campaign-block should not actually link to the campaign.  Once it's published, it should be linked.

This is how we'll handle the daily challenges.  Each day, we'll need to publish the relevant campaign node for that day.

blisteringherb:
Fixes #1897


aaronschachter:
:+1:

PR #2063, opened 2014-05-09
Adding add'l larger images to search and updating relevant code

### LINKED ISSUE #1709: [Finder] Higher resolution images
We should be using higher resolution images for the default tiles (400x400) to avoid ugly scaling. Likewise, we need a bigger image for the "big" tile (720x720).

blisteringherb:
Fixes #1709

@DFurnes would you mind checking this to make sure the changes to
the finder make sense?


DFurnes:
Looks good. :+1:

PR #2074, opened 2014-05-12
Restores Zendesk tickets

aaronschachter:
@angaither Please review.

Reverts the Zendesk repo to a commit that works.  Ran a `ds build` with this change locally and all looks good.


angaither:
I'm still getting the error locally after this build, but we can merge and test in stage.

PR #2086, opened 2014-05-13
Go home, grandpa

### LINKED ISSUE #1921: Signup Data Form: Old people messaging
- [x] Create a `dosomething_user_is_old_person` function which returns TRUE if user is over 25
- [x] Create a new property for a signup data form, "Prevent old people from submitting", which will not be checked by default.
- [x] Modify the auto-prompt user (refs #1913) functionality to check for this setting, and if set, auto-prompt based on whether not `dosomething_user_is_old_person`
- [x] Create a new property for "Old people copy", which is only visible in the config form when "Prevent old people from submitting" is checked
- [x] Modify the user-facing signup data form to display the "Old people copy" if an old person is viewing the signup data form (which would be done by clicking the modal link in the action page)

aaronschachter:

Fixes #1921.

Adds two new properties to the Signup Data Form configuration:
- Prevent old people from submitting
- Old people copy

Displays old people copy if an old person clicks on the modal form.  Also makes sure not to auto-prompt an old person if the form is required.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -322,6 +322,7 @@ function dosomething_campaign_add_signup_data_form_vars(&$node) {
   $node->content['signup_data_form_link'] = $config['link_text'];
   // Load signup sid.
   $sid = dosomething_signup_exists($node->nid);
+
   // If no signup exists:
   if (!$sid) {
     // Staff is viewing a campaign which they haven't signed up for.
@@ -330,6 +331,19 @@ function dosomething_campaign_add_signup_data_form_vars(&$node) {
     );
     return;
   }
+
+  // If form is set to prevent Old People from submitting, and you're Old:
+  if ($config['prevent_old_people_submit'] && dosomething_user_is_old_person()) {
+    // Set the form to the Old People copy config property.
+    $node->content['signup_data_form'] = array(
+      '#markup' => $config['old_people_copy'],
+    );
+    // By exiting out of function now, we don't check to set the 
+    // required_signup_data_form variable, meaning Old People will never be
+    // auto prompted.
+    return;
+  }
+
   // Load the signup entity.
   $signup = signup_load($sid);
   // Pass to the user signup data form.
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
@@ -80,6 +80,24 @@ function _dosomething_signup_node_signup_data_form(&$form, &$form_state) {
       ),
     ),
   );
+  $form[$fieldset]['config'][$prefix . 'prevent_old_people_submit'] = array(
+    '#type' => 'checkbox',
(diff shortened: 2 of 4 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -330,6 +331,19 @@ function dosomething_campaign_add_signup_data_form_vars(&$node) {
     );
     return;
   }
+
+  // If form is set to prevent Old People from submitting, and you're Old:
+  if ($config['prevent_old_people_submit'] && dosomething_user_is_old_person()) {
```
angaither, line 336: do we want to ignore if `dosomething_user_is_staff` ?

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -80,6 +80,24 @@ function _dosomething_signup_node_signup_data_form(&$form, &$form_state) {
       ),
     ),
   );
+  $form[$fieldset]['config'][$prefix . 'prevent_old_people_submit'] = array(
```
angaither, line 83: `$prefix . '_prevent_old_people_submit'`

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -80,6 +80,24 @@ function _dosomething_signup_node_signup_data_form(&$form, &$form_state) {
...
+    '#type' => 'checkbox',
+    '#title' => t('Prevent Old People from submitting'),
+    '#default_value' => $values['prevent_old_people_submit'],
+    '#description' => t('If checked, Old People will be unable to submit the form.'),
+  );
+  $form[$fieldset]['config'][$prefix . 'old_people_copy'] = array(
```
angaither, line 89: same

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
aaronschachter, line 336: I think that would make things a little more confusing for staff.  if they get an old person message, they'll know that things are working properly, and also have the system knowledge to know to change their birthday to test.

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
aaronschachter, line 83: `$prefix` does include a `_`: See https://github.com/aaronschachter/dosomething/blob/old_person/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc#L35

angaither, line 83: That's pretty confusing... do we normally do it like that?

aaronschachter, line 83: It's always been like that in this function at least... and similar to `dosomething_helpers_get_custom_var_prefix` as well, where the '_` is included in the prefix.

angaither:
:+1:

PR #2087, opened 2014-05-13
Signup Button label override

### LINKED ISSUE #1915: Campaign: Pitch page label

aaronschachter:

Closes #1915 by outputting a campaign's `pitch_page_label` variable on the Signup Button if it's set.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -242,6 +242,12 @@ function dosomething_campaign_preprocess_signup_data_form(&$vars) {
  *   The corresponding entity wrapper for the node in $vars.
  */
 function dosomething_campaign_preprocess_pitch_page(&$vars, &$wrapper) {
+  // Check for a signup button label override.
+  $label_varname = 'dosomething_campaign_nid_' . $vars['nid']. '_pitch_page_label';
+  $label = variable_get($label_varname, '');
+  if (empty($label)) {
+    $label = NULL;
+  }
   // Adds a signup_button variable.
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -242,6 +242,12 @@ function dosomething_campaign_preprocess_signup_data_form(&$vars) {
  *   The corresponding entity wrapper for the node in $vars.
  */
 function dosomething_campaign_preprocess_pitch_page(&$vars, &$wrapper) {
+  // Check for a signup button label override.
+  $label_varname = 'dosomething_campaign_nid_' . $vars['nid']. '_pitch_page_label';
+  $label = variable_get($label_varname, '');
+  if (empty($label)) {
+    $label = NULL;
```
angaither, line 249: Shouldn't the label be set to the default 'Sign up' here?

aaronschachter, line 249: The default is set in the `dosomething_signup_preprocess_signup_button` function, found in `dosomething_signup.theme.inc`:

```
function dosomething_signup_preprocess_signup_button(&$vars, $label = "Sign Up") {
```

angaither:
:+1:

PR #2088, opened 2014-05-13
Fixes lost merge_var values bug

### LINKED ISSUE #2081: Sign Up transactional emails don't receive the needed merge_vars
`FNAME`, `CAMPAIGN_TITLE`, and `CAMPAIGN_LINK` are getting lost before the payload is sent through to the message broker.

The offending code looks like it's here (dosomething_mbp.module):

```
case 'campaign_signup':
      dosomething_mbp_get_common_campaign_payload($payload, $params);
      $payload['merge_vars'] = array(
        'CALL_TO_ACTION' => $params['call_to_action'],
        'STEP_ONE' => $params['step_one'],
        'STEP_TWO' => $params['step_two'],
      );
      break;

    case 'group_campaign_signup':
      dosomething_mbp_get_common_campaign_payload($payload, $params);
      $payload['merge_vars'] = array(
        'CAMPAIGN_COPY' => $params['transactional_email_copy'],
      );
      break;
```

When you're doing `$payload['merge_vars'] = array(...`, you're losing the values set when you passed the reference to `$payload` through to `dosomething_mbp_get_common_campaign_payload()`. Either return `$payload` from the function instead of passing by reference or do...

```
$payload['merge_vars']['CALL_TO_ACTION'] = $params['call_to_action'];
... etc...
```

cc: @mikefantini, @DeeZone

deezone:
Fixes #2081

Initial `$payload[merge_var']` values were been over written for `campaign_signup` and `group_campaign_signup` `$origin` payload values.

### DIFF
--- a/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
+++ b/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
@@ -91,19 +91,15 @@ function dosomething_mbp_get_transactional_payload($origin, $params = NULL) {
 
     case 'campaign_signup':
       dosomething_mbp_get_common_campaign_payload($payload, $params);
-      $payload['merge_vars'] = array(
-        'CALL_TO_ACTION' => $params['call_to_action'],
-        'STEP_ONE' => $params['step_one'],
-        'STEP_TWO' => $params['step_two'],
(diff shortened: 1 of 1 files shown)


aaronschachter:
:+1:

Issue #2094, opened 2014-05-13
[Grouped Campaigns] Add Additional Text fields in the backend


weerd:
The section for Additional Text is missing from the backend content input!

aaronschachter:
On the content type, the `field_additional_text_title` and `field_additional_text` fields are labelled as "Pre-live Copy" and "Pre-live Header": https://github.com/DoSomething/dosomething/commit/927a0ca1ad6593cc94cf04e72f524e10c301f906

@mikefantini Just want to confirm: The additional header/copy should always be visible regardless of the campaign group state (pre-live, live, post-signup).  Correct?

mikefantini:
So we need fields:
- pre_live_title
- pre_live_copy
- post_signup_title
- post_signup_copy
- additional_text_title
- additional_text_copy
- photo for additional text region

The additional text title and copy will always be visible regardless of grouped campaign state.

The pre-live title and copy will only be visible when child campaigns aren't live.

The post-signup title and copy will only be visible _after_ a user has signed up for the grouped campaign.

PR #2103, opened 2014-05-13
Pre Launch fields for Grouped campaigns

aaronschachter:
@angaither Please review.

Refs #2094

Adds pre-launch fields and sets the additional text fields back as "Additional text" fields.  Still needs an additional text image to close the issue.


angaither:
:+1: (drupal)

PR #2106, opened 2014-05-13
Fixes campaign signup button

aaronschachter:
Whoops.  @angaither Please review, this is needed for deploy today.


desmondmorris:
This looks okay but, why not add an optional label parameter to `dosomething_signup_preprocess_signup_button`.  Then you can just call `dosomething_signup_preprocess_signup_button($vars, $label);` without the conditional logic

aaronschachter:
it is an optional parameter... you can see the history here: https://github.com/DoSomething/dosomething/pull/2087/files

The trouble is that the way the campaign form works now, once you save the node edit form, it saves these "extra*" variables with an empty value... so you can't just check if the variable exists or not.

*extra = Values we're saving in the form as variables, as to not add more fields onto the campaign node type.

desmondmorris:
gotcha.  thanks for the context.

PR #2118, opened 2014-05-14
Updates the revert script.

### LINKED ISSUE #1986: Deployments not always reverting features

angaither:
This handles reverting features that are in a needs review state.
In the piped command it ignores the word 'review' so that modules are listed in the same place.

Fixes #1986

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -162,7 +162,7 @@ function build {
   drush --script-path=../bin/ php-script enable_modules.php
 
   echo 'Reverting Features'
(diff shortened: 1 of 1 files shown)


aaronschachter:
Has this been tested by changing features locally and running?  if yes, then go for it. 

I still personally would prefer to just use `drush -y fra` to keep this code a little more readable, but can't really speak to the resources that this method saves. Curious to hear input from others heavily involved in beta bending (cc: @desmondmorris @mshmsh5000 @blisteringherb )

PR #2119, opened 2014-05-14
Basic required field client-side validation.

### LINKED ISSUE #1917: Signup Data Form: why_signedup JS validation
Blocked by #1912 

If the signup data form is configured to collect the `why_signedup` property, JS validation should check to make sure a value exists.

### LINKED ISSUE #1957: [Signup Data Form] error messages for leaving required fields blank
Ideally we would treat required fields for all signup data forms (like state or zip code when collecting addresses) the same as we treat last name.

This way the page wouldn't re-load with error messages but instead prevent the user from submitting the form with those blank fields.

http://screencast.com/t/A5dByVsMIx2q

### LINKED ISSUE #2091: Signup Data - User feedback for each field when filling out the form
Decide whether to prompt the user to change it based on how severe the changes are. Some are only capitalization changes.

DFurnes:
Fixes #1917. Fixes #1957. Fixes #2091.


weerd:
:+1:

PR #2127, opened 2014-05-14
Fixes Old People Copy bug in Signup Data Form

aaronschachter:
@angaither Please review. 

The ol' copy paste routine.  Gets em every time.


angaither:
:+1:

PR #2129, opened 2014-05-15
Sms Game

aaronschachter:
@angaither @barryclark @desmondmorris @blisteringherb @mshmsh5000  Please give a look if you have time.  Nothing urgent, Development on SMS Game isn't scheduled until next week -- this is a prototype and up for discussion.  

I have to say that it was really quick and easy to build out, being that all the fields and preprocess functions all exist already. Creating a new content type / module / preprocess functions / template seems like it would have been way more painful.
- Adds a new `field_campaign_type` to the Campaign content type, with options `campaign` and `sms_game`.
- Adds a new view mode for the Campaign entity, `sms_game`, which is returned if the `field_campaign_type` is set to `sms_game.`
- Adds JS into the campaign node edit form to conditionally hide field_groups / fieldsets that only apply to campaigns.  This is not complete for SMS Game, as some shared fields like Scholarship and Rules will need to be moved out of the "Prove It" section, which isn't visible for a SMS game. 
- Adds a `node--campaign--sms-game.tpl.php` to the theme to render a SMS Game
- Beginnings of the SMS Game Signup form
- Beginnings of abstracting the campaign preprocess functions into smaller functions that can be re-used between different campaign types.

Progress so far: 

sms game


desmondmorris:
+1 for using view modes here

angaither:
This is awesome, totally validates our feels that this is probably the way to go since dev time was so quick.
Should have done this for campaign groups! :anguished: 
Also is that javascript you wrote?! :smiley_cat:

aaronschachter:
The Javascript was the painful part.  It's embarassing how long it took to write that.

I agree, if we go this route, it does feel right that `grouped_campaigns` coudl be either a campaign type, or possibly even a static content type.  The kicker with it is that it  shares about the same # of fields from both content types.  

The good news with grouped campaigns is that we just need to launch the Hunt, but then none else on the horizon... so we have some room to refactor.  If it was built as a campaign, it would require adding even MORE fields to the campagin type (mainly the Gallery field collection).  

That's the danger in going this route, is adding more and more fields ot the campaign type to encompass for all the variations.  However, I think for SMS Game, I'd liek to add a custom table that stores the properties, instead of adding more and more fields (like the Mobilecommons Alpha/Beta opt-in paths, which is different from the regular Mobilecommons Opt in path variable)

desmondmorris:
@aaronschachter I still think a single content type with variants and a lot of fields is easier to manage than a bunch of types with mostly the same fields.  (that sounded alot better before I wrote it).  But you are right there is always room to refactor and its good to see an example of this route.

angaither:
@desmondmorris couldn't agree more!
Thanks for taking the time to stub this out @aaronschachter

aaronschachter:
Absolutely, it was one of those things where as I was building it, once i got the view modes working, it seemed like yessss this is much better than the dual content type.  Thanks for reviewing, @angaither and @desmondmorris 

Sooo, this should be safe to merge, I've tested with campaigns to make sure I didn't break anything.  It'll be easy to keep adding SMS Game stuff piece by piece, but should merge sooner than later to get in the preprocess changes to avoid conflicts later. What say yous? Mergey?

Issue #1841, opened 2014-04-17
Version paths for static assets + long expire times


mshmsh5000:
We want the best of both worlds:
1. Have browsers cache static assets for a long time.
2. Retain the ability to force browsers to load updated assets.

We have the first quality right now, but not the second: `file.css` is usually served with a 30-day [TTL](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.9), but browsers may then ignore when we update `file.css` within the TTL period.

The classic solution for providing the second quality is versioning within the path. So `file.css` becomes, e.g., `/v1.0.1/file.css`. `/v1.0.1` gets stripped out by the web server, making it just a logical designation. The next deployment will update this to, e.g., `/v1.0.2`, or `/v1.1`.

This approach would fit nicely with our desire to version every release and supply the web heads with an env var set to the deployed version. This could get injected at the application level on page construction.

mshmsh5000:
@desmondmorris @DFurnes @weerd @aaronschachter @barryclark Take a think on this, plz. /cc @mikefantini

mikefantini:
@mshmsh5000 @desmondmorris @aaronschachter @DFurnes @weerd @barryclark since we just launched a new homepage and it looks super whack without the proper css (see image below), is there any way we could change the TTL period to be 1 day just momentarily and then update it to 30-day TTL?

![rge-screenshot 1](https://cloud.githubusercontent.com/assets/6330971/2736522/903a7eb0-c66e-11e3-8e55-1779a86b7588.png)

barryclark:
Assigning this one to @mshmsh5000

DFurnes:
:+1:

Makes sense to me to make a set of theme helper methods that generate the stylesheet, script, or image tag with the proper versioned path on the Drupal application side. Shouldn't have any problems with URLs in CSS then since those are all relative & will be tied to the cache on the old versioned CSS.

We'll also need to make sure we tie Bower assets to the same versioned path, since our Neue CSS in there changes fairly frequently.

mikefantini:
Went to our site using intern computer on Chrome, Windows 7, and saw this: 
![image](https://cloud.githubusercontent.com/assets/6330971/2818934/7f2a5c34-cee9-11e3-90f2-5ae33c0e8506.png)

@barryclark let's chat when you get in and prioritize this accordingly.

barryclark:
@mshmsh5000 is release tagging blocking this moving forward? 

[...]

PR #2132, opened 2014-05-15
Adds VERSION marker and drupal variable

desmondmorris:
### What's this PR do?

This pull request updates the deployment logic to create markers to indicate the current release version.  On deploy, a file named `VERSION` is created with the tag version and added to the webroot - an [example](http://staging.beta.dosomething.org/VERSION).  A Drupal variable `ds_version` is also updated with the current tag version as well.  It can be accessed like so: `variable_get('ds_version', NULL);`
### Where should the reviewer start?

https://github.com/desmondmorris/dosomething/blob/version-marker/config/deploy.rb#L28-L29
### How should this be manually tested?

Pull down this branch and run a test deployment on QA or the staging server.  I have already done so on staging.

Something like: `capistrano staging deploy branch=vX.X.X`

Then, visit: http://staging.beta.dosomething.org/VERSION AND from vagrant check the value of the ds_version variable like so: `drush @ds.staging vget ds_version`
### Any background context you want to provide?

This functionality gives us the ability to pull in akamai hosted paraneue assets by version number.  We also can check the version status across tiers.
### What are the relevant tickets?
#1841

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -11,13 +11,6 @@ end
 
 set :ssh_options, { :forward_agent => true }
 
-# Default value for :linked_files is []
-# set :linked_files, %w{config/database.yml}
-
-# Default value for linked_dirs is []
-# set :linked_dirs, %w{bin log tmp/pids tmp/cache tmp/sockets vendor/bundle public/system}
-
-
 namespace :deploy do
 
   desc "Run ds build tasks"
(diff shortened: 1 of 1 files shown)


config/deploy.rb
```diff
@@ -11,13 +11,6 @@
 
 set :ssh_options, { :forward_agent => true }
 
-# Default value for :linked_files is []
-# set :linked_files, %w{config/database.yml}
-
-# Default value for linked_dirs is []
-# set :linked_dirs, %w{bin log tmp/pids tmp/cache tmp/sockets vendor/bundle public/system}
```
aaronschachter, line 18: So all this is gone since we're using dyanmic directories for the version #'s ?

desmondmorris, line 18: Yah, that is the default stuff that is in that file when you scaffold a new cap project.  All that stuff is rails related.  Figured I would clean it up while i was in there

aaronschachter:
:+1:

PR #1838, opened 2014-04-16
Issue1748-Create_dosomething_mbp-v2

deezone:
See original pull request that could not be merged:
https://github.com/DoSomething/dosomething/pull/1752

Fixes #1748
- moves dosomething_user_mbp_request() to dosomething_mbp_request()
- replaces calls to dosomething_user_mbp_request() with dosomething_mbp_request()
  - dosomething_reportback_mbp_request()
  - dosomething_signup_mbp_request()
  - dosomething_user_user_pass_submit()
  - dosomething_user_new_user()
- **More work to be done** on the permissions related to this issue.

PR #2133, opened 2014-05-15
Removes unneeded feature file

deezone:
Feature reverts should stop complaining with the removal of this unneeded feature file. The `dosomething_mbp` module do not require features exported to files as it does not have UI components.


desmondmorris:
+1 ship it

aaronschachter:
WAIT there is an `administer message broker` permission defined.

aaronschachter:
It was formerly defined by `dosomething_user` and there was an issue to move it into DS MBP so `message_broker` was not a dependecy of `dosomething_user`.  I will find the related issues.

aaronschachter:
See https://github.com/DoSomething/dosomething/issues/1748#issuecomment-40020391. 

The `dosomething_mbp` module is not properly defining this because it's not including the relevant features.inc.

The proper way to fix would be to `drush -y fe dosomething_mbp "user_permission: administer message broker"` -- not sure if thats correct syntax for the permission, you can find it by using `drush fc`

aaronschachter:
This module SHOULD define the permissions, because they are needed for administrators, and message broker is already a dependency.

desmondmorris:
/me retracts previous comment due to ignorance

deezone:
Should the permissions be defined in `message_broker_producer` or `dosomething_mbp`? All of the admin UI lives in `message_broker_producer`. `dosomething_mbp` has all of the DS specific code based logic.

aaronschachter:
`dosomething_mbp`, because the permission for the "administrator role" is DS specific. `message_broker_producer` is considered a contrib module.

deezone:
OK, sounds good.

aaronschachter:
@DeeZone You'll have to do some investigation with `dosomething_user`.  Right now `drush fc --exported user_permission` shows the `user_permission:administer message_broker_producer` defined in `dosomething_user.`  However there was a PR (#1838) that tried to remove this and add it in, which seems to be causing `dosomething_user` to be unable to fully revert. #1838 never added a features.inc include. Once you get it out of `dosomething_user`, use the `drush fe` command to export it into `dosomething_mbp` properly.

PR #2139, opened 2014-05-15
Updating all features with a node type to have the sms display mode.

angaither:
All features with node types are showing as overriden because they don't have the field instance display for the new sms view mode.

this will fix the overriden features.


aaronschachter:
:+1: Good find. As @angaither and I discussed, this seems to be the only "Gotcha" in using view modes as a sub-content-type... the view mode is available to any content type.  Doesn't really matter much since we never use the "Manage display", but.. thats Dries for you.

PR #2140, opened 2014-05-15
Remove the needs review reverts.

angaither:
I changed my mind...the 'needs review' shouldn't be automaticaly reverted because bad things could happen, I guess Dries makes these different for a reason and we shouldn't force magic upon them when he wants a human to fix the issue.


aaronschachter:
ok. we'll just need to keep an eye on features upon every deploy...but probably should be doing this anyway. :+1:

PR #2158, opened 2014-05-15
Campaign Starter Statement field group

aaronschachter:
@angaither Please review.

Moves the "Starter Statement" field into its own field_group, to easily make it accessible to both campaign types, and also in a field_group that makes sense (moving it out of "Know it").

Makes the field required and updates help text for its old field_group accordingly.


angaither:
:+1:

PR #2168, opened 2014-05-16
SMS Game Signup form

aaronschachter:
@angaither Please review.

Continues development on SMS Game Campaign type.
- Adds form elements for alpha phone, beta phones
- Creates signup for auth user upon submit
- Redirects user to reportback confirmation page after submit
- Alters confirmation page logic to display for anon users if node is a sms game. 
- Outputs starter statement field into tpl

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
@@ -6,22 +6,47 @@
  */
 
 /**
- * Form constructor for the SMS Game form.
+ * Form constructor for an Alpha/Beta signup form.
+ *
+ * @param int $num_betas
+ *   Number of inputs to provide for beta numbers.
+ * @param int $alpha_opt_in
+ *   Opt-in code for the alpha (user submitting form).
+ * @param object $wrapper
+ *   Opt-in code for the beta (other mobile numbers provided).
  */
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
```diff
@@ -31,3 +57,20 @@ function dosomething_signup_sms_game_form($form, &$form_state) {
...
   return $form;
 }
+
+/**
+ * Form submission handler for dosomething_signup_alpha_beta_form().
+ *
+ * Redirects user to the reportback confirmation page for given node.
+ */
+function dosomething_signup_alpha_beta_form_submit($form, &$form_state) {
+  $values = $form_state['values'];
```
angaither, line 122: need to make sure to check these phone numbers against the regex https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_user/dosomething_user.module#L110

Also we have something in the ds_user module that checks if that phone number is already on an account, do we want/need that for this?

aaronschachter, line 122: Yeah that's a good point re: rege, I will add that in.

Asked @mikefantini about this and I think for right now, we don't need to worry about if the # belongs to a different account.

aaronschachter:
@angaither Updated with a server side validation function.

angaither:
:+1:

PR #2177, opened 2014-05-16
Confirmation page tweaks

aaronschachter:
@angaither Please review
- Moves the `field_reportback_confirm_msg` out of "Prove It" group and into a new group "Confirmation", so it's editable to both campaign types
- Creates a title callback to display different reportback confirmation page titles based on campaign type


angaither:
:+1:

PR #2192, opened 2014-05-19
UPS client-side address validation/suggestions

### LINKED ISSUE #1956: [Signup Data Form] wrong but close addresses
When someone enters an invalid address but it's close to the real address, can we have them confirm the address we think it is before submitting it?

In this example, 21 Pineapple Lane isn't real, but 21 Pineapple Street is. Here even though the user submits Pineapple Lane, we enter it as Pineapple Street and save that to their profile.

http://screencast.com/t/rP8DJqxfwI

Ideally we would have the user confirm this suggestion before just going with it.

Some error message like "Hmmmm we couldn't quite find that address, is this what you meant?" and then display for them the address we think it is.

@barryclark not sure if this is a fender or bender task.

DFurnes:
Client-side validation of addresses through UPS API. Closes #1956.
Requires DoSomething/neue#281.

For review: @barryclark @weerd

screen shot 2014-05-19 at 12 09 38 pm

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -243,14 +243,10 @@ function dosomething_user_add_validation_attributes_to_address_fields($form, &$f
 
   // Add `data-validate` attributes:
   $address['address']['street_block']['thoroughfare']['#attributes']['data-validate'] = 'address1';
-  $address['address']['street_block']['thoroughfare']['#attributes']['data-validate-required'] = '';
   $address['address']['street_block']['premise']['#attributes']['data-validate'] = 'address2';
   $address['address']['locality_block']['locality']['#attributes']['data-validate'] = 'city';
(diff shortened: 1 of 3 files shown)


angaither:
Oh man, that's pretty stuff. :sunflower:

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -243,14 +243,10 @@ function dosomething_user_add_validation_attributes_to_address_fields($form, &$f
 
   // Add `data-validate` attributes:
   $address['address']['street_block']['thoroughfare']['#attributes']['data-validate'] = 'address1';
-  $address['address']['street_block']['thoroughfare']['#attributes']['data-validate-required'] = '';
```
weerd, line 246: Who calls is 'thoroughfare', lol. :gb:

weerd:
:+1:

PR #2194, opened 2014-05-19
Adds admin switch for sending campaign values to Message Broker

### LINKED ISSUE #2085: DS MBP setting for campaign cache
@DeeZone Not sure if this taken care of or not, feel free to close if so.

Ideally the `hook_node_insert` and `hook_node_update` functions which update the message broker's campaign cache should only execute if a variable is set to TRUE.  

The variable should only be set to TRUE on production, whereas making content changes on QA or Staging should not alter any MBP campaign cache (the variable would be set to FALSE)

---
- [x] Add Admin interface path
- [x] Add checkbox admin setting to manage campaign_caching setting. This will only be enabled on the production site to avoid caching campaign settings created / edited on non-production sites.
- [x] Add check for setting in `dosomething_mbp_request()` to control sending campaign caching request to `message_broker_producer_request()`.

deezone:
Fixes #2085
- Adds admin switch for sending campaign values to Message Broker. Disable on non production sites to ensure that only production changes are stored in mb-campaigns-api.
- Adds todos for related payload changes needed.

### DIFF
--- a/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.admin.inc
+++ b/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.admin.inc
@@ -1,0 +1,25 @@
+<?php
+
+/**
+ * @file
+ * Module settings UI.
+ */
+
+/**
+ * Configuration form
+ */
+function dosomething_mbp_config_form($form, &$form_state) {
+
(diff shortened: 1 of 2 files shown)


aaronschachter:
:+1:

PR #2196, opened 2014-05-19
Update ds make file

### LINKED ISSUE #1854: DS Build: drupal-org.make doesn't respect version numbers
- [x] Audit what versions of all contrib modules that are in the codebase: https://gist.github.com/aaronschachter/c279be7308f9132da096
- [x] Update `drupal-org.make` to specify `[version]` for each module, and updating any versions which don't match what we're currently using (e.g. Entity and Field Collection examples below)

---

The build is not using the specified version numbers, but instead grabbing the latest recommended release.  This makes for trickier debugging when shit starts failing.

A guess would be maybe we have to specify the whole version number, e.g. "7.x-1.2"

Examples:

makefile Entity:
`projects[entity] = "1.2"` 
In entity.info:
`
; Information added by Drupal.org packaging script on 2014-04-08
version = "7.x-1.5"`

makefile Field Collection:
`projects[field_collection] = "1.0-beta5"` 
but in `html/profiles/dosomething/modules/contrib/`, FC is `"7.x-1.0-beta7"`

angaither:
- adds `[version]` to the makefile
- updates all modules to the version that they were using already anyway
  Fixes #1854

I started doing this all fancy and automated, but then ran into

So then I did it manually...

### DIFF
--- a/lib/profiles/dosomething/drupal-org.make
+++ b/lib/profiles/dosomething/drupal-org.make
@@ -28,7 +28,7 @@ projects[ctools][version] = "1.4"
 projects[ctools][subdir] = "contrib"
 
 ; Date
-projects[date][version] = "2.6"
+projects[date][version] = "2.7"
 projects[date][subdir] = "contrib"
 
 ; Diff
@@ -36,7 +36,7 @@ projects[diff][version] = "3.2"
 projects[diff][subdir] = "contrib"
 
 ; Entity
-projects[entity][version] = "1.2"
+projects[entity][version] = "1.5"
 projects[entity][subdir] = "contrib"
 
 ; Entity Connect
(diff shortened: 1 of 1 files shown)


lib/profiles/dosomething/drupal-org.make
```diff
@@ -104,44 +104,44 @@ projects[redirect] = "1.0-rc1"
 projects[redirect][subdir] = "contrib"
```
aaronschachter, line 104: Looks like redirect is missing.... just double check that all listed modules have that `[version]` defined and looks good.

aaronschachter:
Please test a `ds build` to make sure all looks good, and double check any missing modules.  This closes #1854

angaither:
Yeah, tested with a `ds build` and it worked.

Issue #1710, opened 2014-04-08
Reportback Field: Dynamic radio options


aaronschachter:
Allow for dynamic options: Currently, a Reportback field with type == 'radios' has hardcoded options.  Allow key|label values to be added similar to how Drupal renders select lists for Fields (and hopefully find the same code that translates the key|value and line breaks into the relevant $options array).

aaronschachter:
No longer supporting Reportback Fields.

PR #2205, opened 2014-05-19
Campaign Custom Settings tab

aaronschachter:
@angaither Please review.  Needed for deploy today.

Creates a new page callback, `node/%node/custom-settings` and places the Reportback Field config form there. This will also be the home for the other non-content-type settings, such as the Signup Data Form config, custom settings etc.

This makes the form submit logic much easier for the Reportback Field form, and will stop the form from saving empty values into `dosomething_reportback_field`, since it won't be called on every node save.

Adds a `textarea` option into the Reportback Field type options, to allow adding an additional textarea into the Reportback Form.

Also removes the Reportback field "options" input to keep the config form cleaner, until we need it (if/when we determine to implement #1710)

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -46,6 +46,17 @@ function dosomething_campaign_menu() {
     'type' => MENU_LOCAL_TASK,
     'weight' => 60,
   );
+  // Campaign custom settings admin page.
+  $items['node/%node/custom-settings'] = array(
+    'title' => 'Custom Settings',
+    'page callback' => 'dosomething_campaign_admin_custom_settings_page',
+    'page arguments' => array(1),
+    'access callback' => 'dosomething_campaign_admin_custom_settings_page_access',
+    'access arguments' => array(1),
+    'file' => 'dosomething_campaign.admin.inc',
+    'type' => MENU_LOCAL_TASK,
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -124,6 +135,23 @@ function dosomething_campaign_node_delete($node) {
...
+    'node/*/custom-settings' => TRUE,
+  );
+  return $paths;
+}
+
+/**
+ * Page access callback for admin custom settings form.
+ */
+function dosomething_campaign_admin_custom_settings_page_access($node) {
+  return ($node->type == 'campaign' && dosomething_user_is_staff());
```
angaither, line 152: Not needed on campaign groups?

angaither:
Looks good :+1:
As long as it's not needed for campaign groups. 
Is there a different issue to add all the other custom stuff to this later?

angaither:
Once the other vars live on that form will they submit null values if only some are filled out?

aaronschachter:
Not needed for campaign groups (no reportbacks there).  I'll open up an issue to port the other forms there.

aaronschachter:
The quick fix would be to just embed multiple forms in the page, but you can only submit one at a time -- so the answer to that is no, it won't save null values in other records.

To make it one submit button for all, we'll need to revisit each form constructor to make it one big form.. .which may or may not be worth doing.

PR #2207, opened 2014-05-19
Revert to using <input> rather than HTML5 <button>. Sad Drupal.

### LINKED ISSUE #2206: [signup data] can't submit form
When I get presented with the Thumb Wars order modal, and fill it out correctly and submit the form, the page reloads, but the modal reappears automatically and did not submit my request.

If I skip and then try to submit it, it works properly.

cc: @barryclark @aaronschachter

DFurnes:
Drupal seems to be choking when we use a `<button type="submit">` instead of an `<input type="submit">`. Oh well, Google lied. Fixes #2206.


aaronschachter:
:+1: Get that out of there

Issue #2153, opened 2014-05-15
[Campaign Template] - new report back question for TW


mikefantini:
Campaigns team wants to ask "Who did you share your Thumb Socks with and how did it go?" on Thumb Wars.

your call @aaronschachter if we want to handle this like we did Comeback Clothes or if we want to build it so they can add an additional question.

aaronschachter:
The functionality for "Comeback Clothes" allows you to add 1 additional reportback form element on the reportback form, and right now it only allows for radios.  We can expand this functionality to allow an editor to select radios or textarea as the additional reportback form question.

mikefantini:
cool let's do that.. do editors need a dev to enter "field name"

aaronschachter:
Yeah that would be best. Do we have a node on production now?

mikefantini:
@aaronschachter https://beta.dosomething.org/campaigns/thumb-wars

aaronschachter:
This has been added to production. Editors can edit the form label via https://beta.dosomething.org/node/1508/custom-settings

PR #2208, opened 2014-05-19
Report back workflow for Thumb Wars SMS

jonuy:


aaronschachter:
Looks ok, although there is an extra question #2153 in this reportback form... @mikefantini ok to launch without it?

PR #2211, opened 2014-05-19
Updates to fix some wonky stuff.

### LINKED ISSUE #1960: [Static content] need the ability to add a partner to static content

weerd:
Resolves #1960

@DoSomething/front-end


aaronschachter:
Looks ok here...

PR #2212, opened 2014-05-19
Quick and dirty fix for caching issues pre-Thumb-Wars launch.

DFurnes:
For review: @mshmsh5000


mshmsh5000:
Don't tase/cache me +1

PR #2215, opened 2014-05-20
Fixes Grouped Campaign error

aaronschachter:
@angaither Please review.

Error would only occur when a Grouped Campaign node has no child campaigns.


angaither:
:+1:

PR #2220, opened 2014-05-20
Upgrade the drupal version for safety.

### LINKED ISSUE #1839: Upgrade Drupal core for security release
New security release: https://drupal.org/drupal-7.27-release-notes
- [x] Add to the make file,
- [x] TEST. Run `ds build` locally and smoke test.

angaither:
built and tested locally, didn't see any issues.

There is actually a newer version of drupal core, but for some reason upgrading to that version was failing.

Fixes #1839

### DIFF
--- a/drupal-org-core.make
+++ b/drupal-org-core.make
@@ -5,7 +5,7 @@ core = 7.x
 
 ; Drupal Core
 projects[drupal][type] = core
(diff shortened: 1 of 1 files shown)


aaronschachter:
Ok lets give it a shot :+1: 

We should create a tag before we merge this though, to easily rollback.

PR #2227, opened 2014-05-20
Fixes padding on campaign banners to inherit properly.

DFurnes:
Quick cleanup for "banner" styles. References: DoSomething/neue#286.


weerd:
:+1: :bread:

PR #2229, opened 2014-05-20
Prepare dist package for asset versioning.

DFurnes:
Let's wait to merge this until after today's deploy. :construction_worker: 

This just sets Grunt to copy Bower components and JS source into the `dist/` directory, so that it can be packaged up by the versioned assets Jenkins task. Followup PR will swap the different asset path variables based on whether `ds_version` exists or not.

References #1841.


desmondmorris:
this looks good.  
great job

PR #2240, opened 2014-05-20
Theme setting to toggle CDN assets on or off.

### LINKED ISSUE #1841: Version paths for static assets + long expire times

DFurnes:
Added a theme setting so we can toggle whether or not to use CDN assets. If `ds_version` isn't set (like on a local development box), it will default to the `latest` asset folder on the CDN.

Fixes #1841.
For review: @desmondmorris

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/paraneue_dosomething.info
+++ b/lib/themes/dosomething/paraneue_dosomething/paraneue_dosomething.info
@@ -13,6 +13,7 @@ features[] = main_menu
 features[] = secondary_menu
 
 ; Settings
+settings[use_cdn_assets] = 0
 settings[show_campaign_finder] = 0
 settings[show_sponsors] = 1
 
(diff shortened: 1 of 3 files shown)


desmondmorris:
Awesome.  This is some quality drupaling and form apiery.

Non-blocking for merge - Would it be worth adding a drop down to select an exact version?

DFurnes:
That'd be cool. It'd also be nice to eventually decouple Neue on the CDN from the main app release, so that we could reference the same cached copy from multiple places. I'll make an issue for future enhancements.

PR #2241, opened 2014-05-20
Adding sshfs and add'l cpus for speed

blisteringherb:
- Loading the sshfs commands if the plugin is installed
- Adding an additional cpu and more memory to the box

Does this look like it makes sense?

@mshmsh5000 
@angaither

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -4,7 +4,8 @@ Vagrant.configure("2") do |config|
   config.vm.box_url = "http://files.vagrantup.com/precise64.box"
 
   config.vm.provider "virtualbox" do |v|
-    v.customize ["modifyvm", :id, "--memory", 1024]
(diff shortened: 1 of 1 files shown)


angaither:
I mean, looks great to me. 
How do you set ENV variables?

mshmsh5000:
This looks good to me, but doesn't make sense to merge until we automate some business with Salt: viz., copying the /vagrant director to an internal dir.

mshmsh5000:
@blisteringherb I've upped the memory allocation in https://github.com/DoSomething/dosomething/pull/2390

But not the VCPU allocation. I also hard-coded the SSHFS mount point, because I think it might be too much to get devs to set env vars locally to make this work.

Thoughts?

blisteringherb:
@mshmsh5000  Definitely makes sense.  Anything to make setting it up easier.

PR #2243, opened 2014-05-20
Not using an nonexistant default setting.

DFurnes:
:dancer:


desmondmorris:
:dancer: :dancers: 
:fire: :fire: :fire: :fire:

PR #2246, opened 2014-05-21
Adding grouped campaigns to search

### LINKED ISSUE #2095: [Grouped Campaigns] needs to show up in finder results
Not sure if this is set up or not yet (can't test because home page is not up and /campaigns is only showing two campaigns on staging), but we need to make sure that Grouped Campaigns appear in the finder and on /campaigns.

Also need to make sure that content type (i.e. whether they are a grouped campaign or a campaign) doesn't affect where they are prioritized in the listing.

I think the best rule to handle whether we show the grouped campaign in the finder or not is a check box on the node itself. But open to other suggestions.

Assigning this over to you @aaronschachter.

cc: @blisteringherb

blisteringherb:
Fixes #2095


aaronschachter:
:+1:

PR #2249, opened 2014-05-21
Fixing hierarchy and some styles.

### LINKED ISSUE #2239: [Grouped Campaigns] flip campaigns view and additional text
When campaigns are live on the Grouped Campaigns view, we need to flip the campaign listings view and the additional text header and copy:

image

So it should be: 
- Intro
- Post-Signup (if that's the proper state for the user)
- Campaign Listing
- Additional Text

weerd:
Also hiding the Drupal Admin menu on mobile and tablet, cause why?!

Resolves #2239

@DoSomething/front-end

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_container.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_container.scss
@@ -24,6 +24,14 @@
     }
   }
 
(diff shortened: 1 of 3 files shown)


barryclark:
Looks good @weerd! :+1:

weerd:
Thanks!

PR #2250, opened 2014-05-21
Node signups fix

aaronschachter:
@blisteringherb Please review

The comma separator in the SID field was causing the outputted signup/[sid] link to break in the Node Signups view on production.


blisteringherb:
:+1:

PR #2252, opened 2014-05-21
Added script to null out default birthday

### LINKED ISSUE #2242: null out 1969-12-31 birthdays
null out all birthdays that were set to the default 1969-12-31 00:00:00

angaither:
Sets Dec 31, 1969 to NULL. This is about 1,400 people.
Will need to be tested on QA.

Fixes #2242

### DIFF
--- a/scripts/null-fake-birthdays.php
+++ b/scripts/null-fake-birthdays.php
@@ -1,0 +1,18 @@
+<?php
+/*
+ * Script to null out birthdays that are set to 1969-12-31 00:00:00.
+ *
+ * drush --script-path=../scripts php-script null-fake-birthdays.php
+ *
+ */
+
+echo "Updating filler birthdays \n";
+
+db_update('field_data_field_birthdate')
+  ->fields(array(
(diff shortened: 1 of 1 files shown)


desmondmorris:
Code looks good.  Should we capture those uids before they get wiped?  Just in case you want to ~~spam~~ contact them to update there stuff

aaronschachter:
Could also output # of affected rows as a nice to have, but nothing required.  https://api.drupal.org/api/drupal/includes%21database%21database.inc/function/db_update/7#comment-6969

deezone:
Adding myself to the issue as I'll need to make a SQL dump of the users table once this data cleaning is in place to update the Message Broker User API data. @angaither please poke me when you run this script.

angaither:
Not sure if it's needed to poke them to update thoughts? cc @mikefantini

mikefantini:
not concerned with asking them to update their stuff. let's just null these out and call it a day. they can always update via the profile (eventually)

PR #2253, opened 2014-05-21
Fixes issue with address field validation in IE.

DFurnes:
jQuery's [serializeArray](http://api.jquery.com/serializearray/) seems not to work when given a `fieldset` in Internet Explorer. Giving it the specific fields we want validated seems to fix things.

:construction_worker: Doing some more cross-browser testing before merging this.


mikefantini:
@DFurnes can you check old mozilla too?

DFurnes:
Tested IE8, IE9, IE11 and all working now. Is there a specific version of Firefox I should test?

mikefantini:
try Firefox 26 @DFurnes

DFurnes:
Tested Firefox 18. :japanese_ogre:

weerd:
:+1:

PR #2259, opened 2014-05-21
Uglify and sourcemap-ify with RequireJS.

DFurnes:
Since we had to change up how we bundle up front-end assets a bit for CDN distribution, just switching to a traditional compilation step rather than using the RequireJS loader to dynamically load in assets when on dev. **This means that you'll need to run `ds grunt` in order to compile changes in scripts**, which I know is kinda a pain, but it also means we'll have much closer dev/prod parity which is a good thing.

Also enables source maps, which is gonna be super helpful for debugging things on prod.

:cactus:


weerd:
:+1: Looks cool; not sure how much I understand of the Require stuff @_@

Issue #2258, opened 2014-05-21
Signups: Refactor existing Signup variables


aaronschachter:
Refactor existing Campaign Signup Mobilecommons / Mailchimp variables to use `dosomething_helpers_variable` per #2740 

Standard Campaign Signups:
- [x] Update `dosomething_helpers_variable` with relevant `variable` table values
- [x] Update all `variable_get` methods that find the values to use values from `dosomething_helpers_variable` or better yet the Campaign class
- [x] Alter the admin third party config form to use `dosomething_helpers_set_variable` instead of setting records in the `variable` table
- [x] Delete the existing signup variables

SMS Game Signups:
- [x] Migrate `dosomething_signup_variable` data into  `dosomething_helpers_variable`
- [x] Update Friends Form
- [x] Drop `dosomething_signup_variable` table

Issue #2175, opened 2014-05-16
Signups: table to store third-party subscription values


aaronschachter:
A SMS Game needs to store 2 values: the Mobilecommons alpha opt_in path and beta opt_in path.

We don't want to add fields to store these 2 values, as they don't relate to all general campaigns.
- [x] Create a new table `dosomething_signup_variable` which will store the SMS Game values, but also store relevant Mailchimp Group Name, Group ID and Mobilecommons Opt IN paths for regular campaigns.
  - [x] entity_type
  - [x] entity_id
  - [x] name
  - [x] value
- [x] Add a new section to the admin config form to provide inputs for all SMS games.

PR #2260, opened 2014-05-21
Signup variable table

aaronschachter:
@angaither Can you pls review?

Creates a custom variable table and get/set functions for Signup third party values (and other potential settings as well).  Details in #2175.  

Configures admin form to set variables for SMS Games.  No changes yet to existing Campaigns (will be handled by #2258)

This allows us to begin work on actually sending the API request to Mobilecommons in the SMS Game.


angaither:
:+1:

Issue #1321, opened 2014-03-17
Footer Menu


angaither:
Figure out a better, less hardcodey way to render the footer/nav menu.

barryclark:
Sweet! Thanks :octopus:

DFurnes:
@angaither Would love to get your thoughts on DoSomething/paraneue#49.

barryclark:
@angaither @DFurnes Pretty sure this can be closed as @desmondmorris reworked the footer?

PR #2261, opened 2014-05-21
Make footer links editable

### LINKED ISSUE #2092: Make footer links customizable
It is required that international affiliates be able to manage links in the footer.  Currently the footer content is hard coded in a template file and will require a developer to update.  
We should provide the ability to update via theme settings and push the configs into the footer template file.

This work will also allow the non international footer links to be managed without a deployment.
- [x] Create fieldset in dosomething_paraneue theme settings file.  Add fields to manage the 3 columns of links including the headers.  Include a toggle for disabling/enabling social icons
- [x] Add preprocess function that feeds the footer theme settings into the footer.tpl.php file
- [x] Update footer.tpl.php to use variables rather than hard coded links.

See: https://trello.com/c/3dwQI1TM/4-footer-customization

desmondmorris:
#### What's this PR do?
- Provides UI for editing footer links via the theme settings form
- Provides a UI for toggling social links
- Modifies footer theme function to accept link and social variables
- Updates the theme footer call in preprocess_page adding the social and links variables from theme settings
- Provides default settings via the theme info file to prevent disaster.

We will need to revisit and re-org this theme-settings form as it is quickly growing.  A proper gameplan/rules may be worth thinking about @DFurnes 
1. **Footer.tpl.php** - https://github.com/desmondmorris/dosomething/blob/footer-funtime-nbd/lib/themes/dosomething/paraneue_dosomething/templates/system/partials/footer.tpl.php
2. **Footer theme function** - https://github.com/desmondmorris/dosomething/blob/footer-funtime-nbd/lib/themes/dosomething/paraneue_dosomething/includes/theme.inc#L30-L45
3. **Preprocessor to stuff vars into the template** - https://github.com/desmondmorris/dosomething/blob/footer-funtime-nbd/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc#L35-L62
1.  Checkout my branch
2.  Clear cache
3.  Visit the admin theme settings for paraneue dosomething.  Toggle social links.  Visit home and confirm the enabled/disabled social icons in the footer are represented properly
4.  Return to the admin page and update the footer links by overriding the column headers, links and class values.  Take note to the link formatting (Link Text|http://url.com)

This is the starting point to making the footer completely manageable via theme settings.

Fixes #2092

**Super primitive link adding capabilities**:  Should be good enough for now
screen shot 2014-05-21 at 2 55 05 pm

cc: @weerd @DFurnes @aaronschachter - its not particularly artisanal, but its a start.  chime in please

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/paraneue_dosomething.info
+++ b/lib/themes/dosomething/paraneue_dosomething/paraneue_dosomething.info
@@ -17,6 +17,17 @@ settings[use_cdn_assets] = 0
 settings[show_campaign_finder] = 0
 settings[show_sponsors] = 1
 
+settings[footer_links_first_column_heading] = 'Help'
+settings[footer_links_first_column_links] = 'Contact Us|about/get-touch\nHotlines|about/hotline-list\nFAQs|about/faq'
+settings[footer_links_first_column_class] = 'help'
+settings[footer_links_second_column_heading] = 'Get to Know Us'
+settings[footer_links_second_column_links] = 'Partners|about/our-partners\nDonate|about/donate-pretty-please\nTMI Agency|http://www.tmiagency.org'
+settings[footer_links_second_column_class] = 'knowus'
+settings[footer_links_third_column_heading] = 'About'
+settings[footer_links_third_column_links] = 'What is DoSomething.org?|about/who-we-are\nOur Team|about/our-people\nJobs|about/were-hiring\nInternships|about/internships\nOld People|about/old-people\nSexy Financials|about/sexy-financials\nInternational|about/international
+'
+settings[footer_links_third_column_class] = 'about'
+
 ;Exclude
 ;javascript
 exclude[js][] = 'misc/jquery.js'
--- a/lib/themes/dosomething/paraneue_dosomething/paraneue_dosomething.info
+++ b/lib/themes/dosomething/paraneue_dosomething/paraneue_dosomething.info
@@ -18,14 +18,23 @@ settings[show_campaign_finder] = 0
 settings[show_sponsors] = 1
 
 settings[footer_links_first_column_heading] = 'Help'
-settings[footer_links_first_column_links] = 'Contact Us|about/get-touch\nHotlines|about/hotline-list\nFAQs|about/faq'
+settings[footer_links_first_column_links] = 'Contact Us|node/516
+Hotlines|node/518
+FAQs|node/1052'
 settings[footer_links_first_column_class] = 'help'
 settings[footer_links_second_column_heading] = 'Get to Know Us'
-settings[footer_links_second_column_links] = 'Partners|about/our-partners\nDonate|about/donate-pretty-please\nTMI Agency|http://www.tmiagency.org'
+settings[footer_links_second_column_links] = 'Partners|node/532
+Donate|node/539
+TMI Agency|http://www.tmiagency.org'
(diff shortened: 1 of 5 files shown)


lib/themes/dosomething/paraneue_dosomething/paraneue_dosomething.info
```diff
@@ -17,6 +17,17 @@ settings[use_cdn_assets] = 0
 settings[show_campaign_finder] = 0
 settings[show_sponsors] = 1
 
+settings[footer_links_first_column_heading] = 'Help'
+settings[footer_links_first_column_links] = 'Contact Us|about/get-touch\nHotlines|about/hotline-list\nFAQs|about/faq'
```
aaronschachter: I'm not crazy about using the paths as the input here, it'd be nicer to use `node/[nid]`. Even though you are using the `l` function in the preprocess, i believe this would break things if the path changes.

lib/themes/dosomething/paraneue_dosomething/templates/system/partials/footer.tpl.php
```diff
@@ -10,49 +10,54 @@
...
+  <?php endforeach; ?>
 
   <div class="social mobile">
     <ul>
+      <?php if ($social['facebook'] !== 0): ?>
       <li><a class="social-link" title="dosomething on Facebook" href="https://www.facebook.com/dosomething">&#xe600;</a></li>
-      <li><a class="social-link" title="@dosomething on Twitter" href="https://www.twitter.com/dosomething">&#xe603;</a></li>
+      <?php endif; ?>
+      <?php if ($social['twitter'] !== 0): ?>
+      <li><a class="social-link" title="@dosomething on Twitter" href="https://twitter.com/dosomething">&#xe603;</a></li>
```
aaronschachter, line 50: Minor here too, but I'd imagine the title would say something like `@dosomethingUK` vs `@dosomething` on all affiliates...

aaronschachter, line 50: Actually all this stuff should really be configurable... e.g. Twitter URL, FB URL and whatnot. Can create a separate issue for this as it shouldn't block merging....

angaither:
Also, fixes #1321

lib/themes/dosomething/paraneue_dosomething/paraneue_dosomething.info
desmondmorris: +1

lib/themes/dosomething/paraneue_dosomething/templates/system/partials/footer.tpl.php
desmondmorris, line 50: yah I think that will be included in the next pass

aaronschachter:
Looks good to me :+1:

DFurnes:
Only seeing the first link in each heading when I test on my local:
screen shot 2014-05-22 at 4 26 39 pm

Looks like the `\n` escape characters from the default values aren't getting parsed correctly?
screen shot 2014-05-22 at 4 27 44 pm

DFurnes:
Looks good once I manually added the line breaks in the theme settings.

desmondmorris:
@aaronschachter @DFurnes that should do the trick.

DFurnes:
Looks good! :thumbsup:

PR #2262, opened 2014-05-21
Poor man's cache buster. :(

DFurnes:
Adds a query string to bust cache for today's deploy since we're having issues with CDN assets.
For review: @barryclark 

cache money


DFurnes:
Fixed the issue! Closing.

mad cache

angaither:
But dave, tell us how!

DFurnes:
The Jenkins task to build assets was running `bower install` rather than `bower update` so it wasn't pulling in new Bower dependencies. So it was using an old copy of `neue.js` which wasn't compatible with the local `app.js` code.

PR #2270, opened 2014-05-22
Removed index lookup in field_transactional_email_copy value

### LINKED ISSUE #2269: $params[transactional_email_copy] bug
`$params['transactional_email_copy']` is returning the first character in the `field_transactional_email_copy` field value.

"We’ve gotcha down for The Hunt..." returns as "W".

In dosomething_mbp.module:

```
$payload['merge_vars']['CAMPAIGN_COPY'] = $params['transactional_email_copy'];
```

deezone:
Fixes #2269

No need to use `['value']` when using `entity_metadata_wrapper`.


aaronschachter:
:+1: ship it value added

PR #2272, opened 2014-05-22
Grouped campaigns  empty modal list

### LINKED ISSUE #1960: [Static content] need the ability to add a partner to static content

### LINKED ISSUE #2236: [Grouped Campaigns] Modal Links container showing when no modal links

weerd:
Resolves #1960 (along with an update to Neue)
Resolves #2236 

@aaronschachter can you take a look at the update I did to the Campaign Group module, to keep it from populating the $modals var with an empty `<ul>` if no modal content is added.

@DFurnes & @barryclark


aaronschachter:
Bender approved :+1:

DFurnes:
Looks solid :hammer:

weerd:
:dancers:

PR #2277, opened 2014-05-22
Adds nl2br

### LINKED ISSUE #2276: [Grouped Campaigns] transactional email no honoring line breaks
Email is not honoring line breaks entered into the Transactional Email Copy field on grouped campaigns

See "Transactional Email Copy" filed on http://staging.beta.dosomething.org/node/1194/edit

and the email I get...

image

cc: @aaronschachter 

also note don't worry about the [[Button: Share Your Idea]] thing, that's just in the copy and will be removed.

deezone:
Fixes #2276

Support for line breaks in `field_transactional_email_copy`.


deezone:
@aaronschachter A quickie to support line breaks in the `transactional_email_cop` field. Please review when you have sec.

aaronschachter:
Seems like you didn't start from an updated dev, as the changes show your old code before #2270. Probably easier to just pull latest dev and make the change.  Actual change looks fine.

PR #2281, opened 2014-05-22
Checking for field existence before getting value

blisteringherb:
- When indexing a grouped campaign make sure that the fields
  exist before calling value()


aaronschachter:
:+1: Looks good

PR #2285, opened 2014-05-22
Fixes saving custom variables

aaronschachter:


lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -420,7 +420,7 @@ function dosomething_helpers_save_custom_vars(&$form, &$form_state) {
   foreach ($variables as $variable) {
     $$variable = "{$prefix}{$variable}";
     // Save alt color and pitch page label.
-    if (in_array($variable, array('alt color', 'pitch_page_label'))) {
+    if (in_array($variable, array('alt_color', 'pitch_page_label'))) {
       dosomething_helpers_save_custom_var($$variable, $values[$$variable]);
```
weerd, line 424: is this correct `$$variable`?

weerd, line 424: I see it created at top, but seems odd.

weerd:
Curious about the weird variable name, but other than that :+1:

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
aaronschachter, line 424: Hmmm. This is @blisteringherb's changes, but let me try to change all of the `$$variable` to `$variable` to see. hard to tell if typo or not.

weerd, line 424: Yeah, debating if it was maybe a copy and paste typo while selecting multiple items?

aaronschachter:
Looks like the `$$variable` is intentional. tested with changing it to `$variable` and it will not save.  However, maybe look into refactoring? per http://stackoverflow.com/questions/4179489/double-dollar-sign-php

weerd:
Yeah, maybe add a `@TODO` to this part for future ref...?

PR #2286, opened 2014-05-22
Update campaign confirmation message per SPG's request.

DFurnes:
:100: :racehorse: :checkered_flag: :boom:

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
@@ -34,7 +34,7 @@ function paraneue_dosomething_preprocess_page(&$vars) {
   // Show Hunt MCC form confirmation if query string exists
   $node_type = menu_get_object()->type;
   if ($node_type == 'campaign_group' && array_key_exists('submitted_mcc', $_GET)) {
-    drupal_set_message('Thanks for submitting your idea!');
+    drupal_set_message('Thanks! We\'ll be in touch soon with next steps. If you\'d like to submit more than one campaign idea, go for it!');
   }
 
   // Footer Output Setup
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
```diff
@@ -34,7 +34,7 @@ function paraneue_dosomething_preprocess_page(&$vars) {
   // Show Hunt MCC form confirmation if query string exists
   $node_type = menu_get_object()->type;
   if ($node_type == 'campaign_group' && array_key_exists('submitted_mcc', $_GET)) {
-    drupal_set_message('Thanks for submitting your idea!');
+    drupal_set_message('Thanks! We\'ll be in touch soon with next steps. If you\'d like to submit more than one campaign idea, go for it!');
```
angaither: `t()` dat

angaither:
:+1:

lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
```diff
@@ -34,7 +34,7 @@ function paraneue_dosomething_preprocess_page(&$vars) {
   // Show Hunt MCC form confirmation if query string exists
   $node_type = menu_get_object()->type;
```
aaronschachter, line 35: This may cause errors if `type` is not set.  We should be checking for said errors on pages like the user login or password.

DFurnes, line 35: Did some quick clicking around on my local and not seeing any errors on campaigns, static content, user, or home pages.

@barryclark Something to watch out for while testing just in case.

aaronschachter, line 35: It's there, just behind the scenes.  Check out `admin/reports/dblog`

Notice: Trying to get property of non-object in paraneue_dosomething_preprocess_page() (line 35 of /var/www/beta.dosomething.org/releases/20140522191445/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc).

Can replicate by visiting `user/password` as an anon user from incognito.

PR #2287, opened 2014-05-22
Fixes issue with modal lists and adding some margins at bottom of lists ...

weerd:
...in content.

@aaronschachter & @barryclark


barryclark:
:+1: looks good!

PR #2289, opened 2014-05-22
Grouped Campaign Alt Color

### LINKED ISSUE #2041: [Grouped Campaigns] alt color custom style field needed

aaronschachter:
@angaither @weerd please review.
#### What's this PR do?

Adds a Grouped Campaign's alt_color variable into inline CSS in the `<HEAD>` markup.

1.`dosomething_helpers_add_inline_css`
1.  Check out my branch
2.  Set the alt color for a Grouped Campaign node
3. Verify the H2 color changes
4. Set the alt color for a Campaign node
5. Verify the H2 color changes
6. Verify that a different campaign has normal color.

Adding the `dosomething_campaign_preprocess_custom_vars` function didn't do the trick here: https://github.com/DoSomething/dosomething/pull/2283/files#diff-0e259d21c6440ea2681ced1bd67ae238R80
because it's hardcoded to find custom variables with `campaign` in the variable name.

The `preprocess_custom_vars` function is only used for the sake of the `dosomething_helpers_add_inline_css` function (formerly called `dosomething_campaign_add_inline_css` before it was moved into `dosomething_helpers`) , so i've just created a new function `dosomething_helpers_preprocess_custom_vars` which gets the custom vars by node type, and then is called from within `dosomething_helpers_add_inline_css`

TL;DR Just make one call to `dosomething_helpers_add_inline_css` to add the relevant inline CSS for styles variable overrides.

Fixes #2041
#### Screenshots

screen shot 2014-05-22 at 3 40 47 pm

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -43,9 +43,6 @@ function dosomething_campaign_preprocess_node(&$vars) {
       dosomething_helpers_preprocess_partners_vars($vars);
     }
 
-    // Preprocess any custom campaign variables.
-    dosomething_campaign_preprocess_custom_vars($vars);
-
     // Add inline css based on vars.
     dosomething_helpers_add_inline_css($vars);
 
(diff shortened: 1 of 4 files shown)


weerd:
:+1: Tested locally and works! Also really like the refactored approach :dancer: Annnnd it illustrated that I missed adding a class on one of the headlines @_@

angaither:
:+1:

PR #2297, opened 2014-05-22
Fixing broken indexing because of missing low_season field

### LINKED ISSUE #2267: [Grouped Campaign] Error when indexing on stage
Attempting to reindex per https://github.com/DoSomething/dosomething/wiki/Solr-Search-Troubleshooting to get Grouped Campaign visible in finder, and error:

An AJAX HTTP error occurred. HTTP Result Code: 500 Debugging information follows. Path: /batch?id=1700171&op=do StatusText: Service unavailable (with message) ResponseText: EntityMetadataWrapperException: Unknown data property field_low_season. in EntityStructureWrapper->getPropertyInfo() (line 335 of /var/www/beta.dosomething.org/releases/20140521205842/html/profiles/dosomething/modules/contrib/entity/includes/entity.wrapper.inc).

@blisteringherb Do we need to add Campaign Group fields into the index?  If you are able to take a look, please update with what the issue is and how to fix.

blisteringherb:
- Low season field doesn't exist on the grouped campaigns content
  type so it causes an entity field error if indexing tries to
  check that field's value

Fixes #2267

### DIFF
--- a/lib/modules/dosomething/dosomething_search/dosomething_search.module
+++ b/lib/modules/dosomething/dosomething_search/dosomething_search.module
@@ -122,9 +122,8 @@ function dosomething_search_apachesolr_index_document_build_node(ApacheSolrDocum
 
     $node = entity_metadata_wrapper('node', $entity);
     foreach (array('high_season', 'low_season') as $season) {
-      $field = $node->{"field_" . $season};
-      if (isset($field)) {
-        $value = $field->value();
+      if (isset($node->{"field_" . $season})) {
+        $value = $node->{"field_" . $season}->value();
         if (isset($value)) {
           $now = time();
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_search/dosomething_search.module
```diff
@@ -122,9 +122,8 @@ function dosomething_search_apachesolr_index_document_build_node(ApacheSolrDocum
 
     $node = entity_metadata_wrapper('node', $entity);
     foreach (array('high_season', 'low_season') as $season) {
-      $field = $node->{"field_" . $season};
-      if (isset($field)) {
-        $value = $field->value();
+      if (isset($node->{"field_" . $season})) {
```
angaither, line 125: OMG! @DFurnes and I were just talking about variable variables earlier today!

angaither:
:+1:

PR #2300, opened 2014-05-22
Proper IE8 IE9 feature check usage.

DFurnes:
:cactus:


DFurnes:
Verbal :+1: from @barryclark.

PR #2304, opened 2014-05-23
SMS Game Tweaks

aaronschachter:
- Fixes bug to set the form's `num_betas` hidden element correctly.  I changed the form constructor parameter to `$num_friends` but forgot to set this value in the hidden element.
- Prevents a signup from being created if a user logs in on a SMS Game Campaign node.


angaither:
:+1:

PR #2311, opened 2014-05-23
Zendesk for all

aaronschachter:
@angaither Please review

Allows anonymous users to submit Zendesk form, by prompting for email. Authenticated users do not get prompted.

### DIFF
--- a/lib/modules/dosomething/dosomething_zendesk/dosomething_zendesk.module
+++ b/lib/modules/dosomething/dosomething_zendesk/dosomething_zendesk.module
@@ -93,6 +93,16 @@ function dosomething_zendesk_form($form, &$form_state, $node = NULL)  {
     '#value' => 'normal',
   );
 
+  // If anonymous user:
+  if (!user_is_logged_in()) {
+    // Add email as a required field.
+    $form['email'] = array(
+      '#type' => 'textfield',
+      '#title' => 'Your Email',
+      '#required' => TRUE,
+    );
+  }
+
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_zendesk/dosomething_zendesk.module
```diff
@@ -138,12 +152,28 @@ function dosomething_zendesk_form($form, &$form_state, $node = NULL)  {
...
+/**
  * Submit callback for dosomething_zendesk_form().
  */
 function dosomething_zendesk_form_submit($form, &$form_state) {
-  global $user;
   $values = $form_state['values'];
-  $values['email'] = $user->mail;
+  if (user_is_logged_in()) {
+    global $user;
+    $values['email'] = $user->mail;
```
angaither, line 174: This is super confusing. Instead of setting it as a form value, can you set `$email  = user->mail or $vaules['mail']`

aaronschachter, line 174: The `$values` array is what's being sent to the API, and it expects an `email` parameter.  I'd argue it would be more confusing to change everything to use `$values['mail']` ...

angaither, line 174: Oh, bleh, sorry I thought this was from the `form_state` array that's why.

angaither:
:+1:

PR #2316, opened 2014-05-23
Grouped Campaign URL pattern

aaronschachter:
Exports Pathauto pattern variable value for the `campaign_group` node type.


weerd:
Strongarm @_@ :+1:

PR #2317, opened 2014-05-24
Control mbp requeusts by node type.

### LINKED ISSUE #2315: [SMS Game] No transactionals upon signup
If you're logged in and submit the Friends Signup form, you'll get a transactional email. Logic needs to change to only fire transactionals on signup when campaign_type == `campaign`

deezone:
Fixes #2315

Control mbp calls by node type to prevent "Friends Signup" form submissions from triggering mbp requests.

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -374,7 +374,9 @@ function dosomething_signup_third_party_subscribe($account, $node) {
   $opt_in = isset($opt_in_override) ? $opt_in_override : $opt_in_general;
   dosomething_signup_mobilecommons_opt_in($account, $opt_in, $node->title);
   // Send message broker request.
-  dosomething_signup_mbp_request($account, $node);
+  if ($node->type == 'campaign' || $node->type == 'campaign_group') {
+    dosomething_signup_mbp_request($account, $node);
+  }
 }
 
 /**


aaronschachter:
Has this been tested? Because it shouldn't work :) An SMS Game is still a `campaign` node type.  I think it makes sense for the campaign type check to happen earlier, from within `dosomething_signup_entity_insert` where we have a loaded node.  The function to use is `dosomething_campaign_get_campaign_type`.

However, a PR should not be opened without having testing locally.  There is a SMS Game node at node/1199 if you pull down the staging database.  I'm also happy to take a stab at this issue too if you have more pressing issues (SMS Games aren't going live for a few days at least)

deezone:
No, it has not been tested. Perhaps it's a misuse of pull requests but in this case I'm looking for feedback on proposed code changes. If there's a better approach please let me know.

Now I have my feedback, I'll try the approach you've suggested.

aaronschachter:
Hmm, I suppose that could be open for discussion. 

I think one way to avoid me barking about PR's like this is to specifically ask in the PR: "Hey, this hasn't been tested but, is this the right direction?"  

Or hit up a dev in hip chat to ask about it first if you don't know where to start, vs blindly writing code.  That way you can test your changes.

One thing I don't like about trial PR's is a commit log of irrelevant commits... the OCD in me would definitely suggest creating a clean branch to get rid of any "test" commits.

deezone:
Noted, complete respect to all "OCD"s of the world. I'll delete this request :)

PR #2319, opened 2014-05-27
Fixing CTA bug

### LINKED ISSUE #2318: [grouped campaigns] call to action
Call to action is no longer appearing in the header for grouped campaigns even though the field is filled out:

image

cc: @barryclark

weerd:
Resolves #2318

@barryclark @aaronschachter


aaronschachter:
There it is :+1:

PR #2325, opened 2014-05-27
Updates to fix some IE8 issues that cropped up.

### LINKED ISSUE #2324: [Grouped Campaigns] IE8 bug causing gallery and aside items to not float properly.
Bourbon Neat relies on `:last-child` to remove margins on right side for floated grid items. However, this pseudo property is not supported in IE8. Classes were added (PR #2321) to help out resolve this in triad gallery. 

The class names need to be slightly modified and classes need to be added for `-columned` items to support the layout in those sections as well for IE8, and styled appropriately.

weerd:
Resolves #2324

@barryclark 

@aaronschachter (I included you just cause I changed the class names in the module to use the "modifier" style I'm putting into practice, which prefixes those classes with a hyphen.)


aaronschachter:
Looks good :+1:

weerd:
Righty-o!

Issue #2331, opened 2014-05-27
[Static content] Blockreference field


aaronschachter:
Add a field to the Static Content type which allows us to embed a themeable function onto Static Content.

This would allow Static Content nodes to be the wrapper around pages that currently require custom menu hooks and page callbacks / templates:
- #2322 - Fact landing page
- The MCC form, `volunteer/hunt/campaign-form`
- The scholarships prototype page, `about/scholarships`

Having this content live within the Static Content node gives us a lot for free:
- fields for happy content editing
- URL paths for happy path changing
- metatags support
- status field for unpublished/published (currently the Scholarships visibility is set using a drupal variable, this can be changed to an unpublished Static Content node set to use "Scholarship Listing" as a custom content value).

Open to suggestions besides Custom Content, but the idea behind it is that it requires a custom build to implement. Field names that have been rejected:
- Dynamic Content (might not be dynamic, like the MCC form)
- Themed Content (everything is themed)
- Content Plugins (very similar, maybe too similar to Ctools plugins)

aaronschachter:
Updated issue title - we've decided to implement this via Blockreference module.

PR #2339, opened 2014-05-27
Displaying themable content within Static Content

aaronschachter:
Adds a new `field_custom_content` select field to the Static Content type.

cc: // @mshmsh5000 @angaither @blisteringherb @desmondmorris -- Anyone who is around to take a looksy.

To embed a themeable function `my_function` in a Static Content node, update the allowed values for this field with the new function name. `my_function|My Sweet Code Snippet`.  The Static Content function `dosomething_static_content_preprocess_custom_content_vars` checks if a function name exists with this stored value, and if so, adds its output into the variables to display in the template.

For the Fact Page Archive, I've created a `dosomething_fact_page_archive_custom_content`, which returns the list of all published Fact Pages.

I've created a Static Content node locally, by selecting "11 Facts Archive" for the Custom Content field.  This is currently the only option available in the Custom Content field.

Here is the result:

facts-archive

At the risk of repeating the convo with myself in #2331, here are things we get for free by embedding this functionality into the Static Content node type:
- Published/published
- Header, subtitle, intro copy
- Metatags
- Editable URL alias


aaronschachter:
It may be useful to view the original approach I had started on which led me to this route: https://github.com/aaronschachter/dosomething/compare/fact_page_archive

This original approach required: 
- Defining a menu item and its URL in code
- custom access callback to determine whether or not to display the page
- custom page callback. 
- All intro content to either be handled via code or an admin variable somewhere
- Custom code to add metatag info
- Creating an entire page template to add the header, footer (like we had to in `templates/campaign/submit-campaign-idea.tpl.php`)

aaronschachter:
Update: Pow-wow-ed with @mshmsh5000 @barryclark and @weerd . Going to clone this branch and implement the custom content as a Drupal block, and add a Blockreference field (https://drupal.org/project/blockreference) into the mix instead of the text based `field_custom_content`.  Will keep this PR open in case of any gotchas with using Blocks.

aaronschachter:
#2361 and Blocks for the win. Closing this shiz.

PR #2345, opened 2014-05-28
Sms Game tweaks

aaronschachter:
- Submitting SMS Game form creates a signup for the account which belongs to alpha number, if an account exists. Regardless of logged in or not.
- Don't send transactionals for SMS Game form submitting.


jonuy:
:+1:

PR #2358, opened 2014-05-28
Readding requirejs settings changed when bug fixing.

weerd:
CC @DFurnes


weerd:
Back to the future!

(resetting code back to the way it was)

weerd:
:+1:

PR #2360, opened 2014-05-29
Temp fix to help make nav more legible.

weerd:
CC @DFurnes 

Temporary fix until Nav refactoring!


weerd:
This code is so beautiful I could cry! :+1: :+1:

PR #2361, opened 2014-05-29
Blocks in Static Content

### LINKED ISSUE #2322: [Fact Page] Landing page
A page which presents a list of all Fact Page nodes, grouped by Primary Cause.

Needs:
- [x] Page header / headline.
- [x] Intro copy.
- [x] List of Fact Page titles, each links to its node.
- [x] Metatag descriptions.  <-- This is actually kind of a pain, because need to handle all the `drupal_add_html_head` stuff ourselves. Same will occur for the Scholarships landing page. Wondering if we need some sort of a Dynamic Content vs. Static Content node type. Thinking out loud but could possibly have a few fields like Title, Subtitle, and a select list to add various plugins (like, select a Scholarships listing, or a Fact Page listing)

### LINKED ISSUE #2331: [Static content] Blockreference field

aaronschachter:
- Closes #2331 by adding Blockreference module and `field_blocks` to Static Content
- Closes #2322 by creating a Fact Page List block.
- Adds a `field--blockreference.tpl.php` to remove all markup from a Blockreference field content.


aaronschachter:
Discussed with @barryclark he says it's great code.

PR #2366, opened 2014-05-29
Group Fact Pages by Primary Cause

aaronschachter:
Creates theme function to output Fact Page node lists by Primary Cause (doesn't do any combination grouping of Sex + Relationships, Physical + Mental Health).

_NOTE:_ Stage does not have `field_primary_cause` filled out for many Fact Pages nodes, which is why there won't be many results.  @julielorch has filled this field out on Production but not for all nodes on Staging.

fact-page-cause

### DIFF
--- a/lib/modules/dosomething/dosomething_fact_page/dosomething_fact_page.module
+++ b/lib/modules/dosomething/dosomething_fact_page/dosomething_fact_page.module
@@ -27,18 +27,29 @@ function dosomething_fact_page_block_view($delta = '') {
 
   switch ($delta) {
     case 'fact_page_list':
-      $block['content'] = dosomething_fact_page_list_block_content();
+      // Gather list of Fact Page links.
+      $links = dosomething_fact_page_get_fact_page_list_links();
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_fact_page/dosomething_fact_page.module
```diff
@@ -95,14 +106,21 @@ function dosomething_fact_page_preprocess_node(&$vars) {
...
  * @return array
  *  Array of links.
  */
-function dosomething_fact_page_get_fact_page_list_links() {
- $query = db_select('node', 'n')
-    ->fields('n', array('nid', 'title'))
-    ->condition('type', 'fact_page')
-    ->condition('status', 1);
+function dosomething_fact_page_get_fact_page_list_by_tid($tid) {
+  //return $tid;
```
weerd: this supposed to be deleted? `//return $tid;`

weerd:
For the parts I understood, looks good! :+1:

aaronschachter:
I'll take it!

PR #2370, opened 2014-05-29
Introducing campaign runs

aaronschachter:
Adds new module for new content type, `campaign_run`.

Barebones fields for now.  Starts on a `field_winners` field collection, which will store Winner information for the Run.


aaronschachter:
@barryclark said this module looks pretty spectacular and i am free to merge it into dev.  here i go

PR #2373, opened 2014-05-30
Fixes SMS Game header image

aaronschachter:
@weerd Mind testing this fix locally?

Can test by pulling down this branch, and confirming that the Campaign Cover Image displays for:
- SMS Games
- Campaigns
- Grouped Campaigns

As mentioned in the issue, https://github.com/DoSomething/dosomething/issues/2349#issuecomment-44606427, this bug is kinda nuts.  The different `view_modes` have different `$vars['field_image_campaign_cover]` array keys, causing the conditional in `dosomething_helpers_preprocess_hero_images` to fail for the `sms_game` view mode.

This PR changes the function to use values returned from the `EntityMetaDataWrapper` stop the bleed, but this seems to open possibilities for other weird language issues in other fields where the `$vars` arrays could be different among view_modes.  

I'm beginning to think instead of defining custom view modes, all we might need is just different tpl and preprocess functions based on the Campaign type and "view" (pitch, closed).  Going to look into this a bit.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -36,7 +36,7 @@ function dosomething_campaign_preprocess_node(&$vars) {
     }
 
     // Hero Images.
-    dosomething_helpers_preprocess_hero_images($vars);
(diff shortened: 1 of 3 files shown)


aaronschachter:
Opening a new PR to avoid using a separate `view_mode` for rendering SMS Games, which is what is causing the strange difference in field array structure.

PR #2381, opened 2014-05-30
Removes SMS Game view mode

### LINKED ISSUE #2349: [SMS Game]  header image doesn't display
Bug in preprocess. Shows if Campaign Type == `campaign` but not if `sms_game`

aaronschachter:
Fixes #2349 in a hardcore way. It's because #2349 is some kind of crazy bug where there is a difference in the `$vars['field_image_campaign_cover']` array structure between the two different view modes, `full` and `sms_game`.

The cleanest solution here is to get rid of the view mode altogether.  We don't need it.  All we need to do is suggest a different theme template file based on the Campaign Type instead of looking for a `sms_game` view mode, which is seen here: https://github.com/aaronschachter/dosomething/compare/view_mode_no_more?expand=1#diff-f3920561a5ca7c40913d6f8452fb9c0fR55
## How to test:
- Pull down this branch and `drush -y fra`
- Verify Campaign Cover Images display for:
  - SMS Games
  - Campaigns
  - Grouped Campaigns
## History

The majority of these changes are removing the custom `sms_game` view mode entirely from the site, which means updating all the Features definitions to get rid of it.  

The fact that the `sms_game` view mode exists in Features on all of these content types we don't use it for is a good indicator that we're not using it correctly.

The history here is that the Campaign started as a custom Entity, and not a node.  We used a Pitch entity view for the Campaign entity, and it seemed to work fine to keep it for when we made the switch from defining Campaigns as custom entities instead of Nodes.

Little did we realize that the Pitch view mode was being exported for all field isntances for other content types too (since we started the Campaign node type as one of the first ones).  Will open a separate issue to do the same for Pitch view mode, it should just be handled via serving a different `theme_hook_suggestions` based on the Campaign Type and User Signup status.

### DIFF
--- a/lib/modules/dosomething/dosomething_action_guide/dosomething_action_guide.features.field_instance.inc
+++ b/lib/modules/dosomething/dosomething_action_guide/dosomething_action_guide.features.field_instance.inc
@@ -30,12 +30,6 @@ function dosomething_action_guide_field_default_field_instances() {
         'type' => 'hidden',
         'weight' => 0,
       ),
-      'sms_game' => array(
-        'label' => 'above',
-        'settings' => array(),
-        'type' => 'hidden',
-        'weight' => 0,
-      ),
       'teaser' => array(
         'label' => 'above',
         'settings' => array(),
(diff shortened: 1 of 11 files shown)


weerd:
Looks like it's working! I also flushed all cache if that would possibly play a role.

Campaign cover images show up on:

[x] SMS Games
[x] Campaigns
[x] Grouped Campaigns

:+1: :dancer: :dancers:

Issue #2383, opened 2014-05-30
Refactor / remove Pitch custom Entity view mode


aaronschachter:
Refs #2381 

We don't need a different view_mode, we just need to serve a different tpl file for a Pitch Page.  We are doing this already, but we're checking for the view_mode to be set to `pitch`.  This custom view mode then shows up in all of the other content type Features that don't make any use of it.
- [x] Move the pitch display logic out of `dosomething_campaign_entity_view_mode_alter` and into  `dosomething_campaign_node_view`, by setting a simple variable `$node->display_pitch_page =1`
- [x] Change `dosomething_campaign_preprocess_node` to check for `isset($node->display_pitch_page)` instead of the `view_mode == pitch`
- [x] Remove `dosomething_campaign_entity_info_alter` function to remove the Pitch view mode definition
- [x] Remove `pitch` view mode from all Features `field_instance.inc` files.  See https://github.com/aaronschachter/dosomething/commit/6a93c7a53d73dfcc997e8d8e62c2a7a6b4b10288. For some content types, this will be done by doing a drush -y fua, but for others, the `pitch` view_mode will be stored within fields `field_config_instance` table.  To completely purge the `sms_game` view mode from features, I had to manually remove code from the field_instance.inc in the features of content types that stored it.  Details in https://gist.github.com/aaronschachter/a2c287e8b142fa0737c1
- [x] In `dosomething_campaign_entity_view_mode_alter`, move the last remaining `dosomething_legacy` logic into `dosomething_legacy.module`.
- [x] Remove now empty `dosomething_campaign_entity_view_mode_alter` function.
- [x] Change `node/*/pitch` page logic to not return a rendered node in `pitch` view mode.
- [x] Update DS Campaign README with changes (no more custom view modes)

PR #2385, opened 2014-05-30
Adds dosomething_campaign_is_pitch_page

aaronschachter:
Refs #2383

Major change is in `dosomething_campaign.module`:

Changes the preprocess logic to use the return of new function, `dosomething_campaign_is_pitch_page` instead of checking for `$view_mode == 'pitch'`.  The logic in `dosomething_campaign_is_pitch_page` is code moved out of `dosomething_campaign_entity_view_mode_alter`, as we look to not rely on view modes for any of this various Campaign type / state templating.
- Also removes `pitch` view_mode from content type field_instance.inc feature files.

@weerd Would you mind pulling down and test?
- Confirm that anon user sees pitch page
- Confirm that auth user not signed up also seses pitch page
- Confirm that auth user signed up sees full node
- Confirm that staff sees the full node regardless of signup status

This keeps things functional as cleanup continues to remove the custom view mode out of the codebase / features.  Up next will be removing everything out of the features - but this is the big functional change.

### DIFF
--- a/lib/modules/dosomething/dosomething_action_guide/dosomething_action_guide.features.field_instance.inc
+++ b/lib/modules/dosomething/dosomething_action_guide/dosomething_action_guide.features.field_instance.inc
@@ -24,12 +24,6 @@ function dosomething_action_guide_field_default_field_instances() {
         'type' => 'text_default',
         'weight' => 7,
       ),
-      'pitch' => array(
-        'label' => 'above',
-        'settings' => array(),
-        'type' => 'hidden',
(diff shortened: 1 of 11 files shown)


weerd:
These are confirmed:

[x] Confirm that anon user sees pitch page
[x] Confirm that auth user not signed up also seses pitch page
[x] Confirm that auth user signed up sees full node
[x] Confirm that staff sees the full node regardless of signup status

:+1:

Issue #2347, opened 2014-05-28
rem polyfill error breaks DS.org in IE8


barryclark:
![image](https://cloud.githubusercontent.com/assets/692762/3107245/5d884a3a-e683-11e3-84c9-460f03fdbaa0.png)

via @weerd: There's an error w/ the polyfill for IE8 for rem support (rem.js). it's erroring out and blocking any proper rendering for respond.js in IE8. 

via @DFurnes: I think IE 8 issue _might_ be fixed if you take rem.js and respond.js out of excludeShallow config and remove Features.\* if statements by their respective require() calls in main.js

weerd:
@tjrosario suggested this might be a cross-origin resource sharing issue. We tested this by turning off CDN assets on stage and IE8 works great again.

barryclark:
Unfortunately doesn't look like this has been resolved by the #2344 fix. cc @mshmsh5000 @weerd @tjrosario @DFurnes

weerd:
So it looks like the Rem Polyfill is no longer an issue, but for some reason respond.js is not running, which is why the site is stuck on mobile view on IE8.

barryclark:
cc'ing @DFurnes @tjrosario on this

DFurnes:
Waiting on a HTTP endpoint for our CDN, but this should be fixed next time we deploy on production and will be a quick setting change on staging.

PR #2400, opened 2014-06-02
Bundle em up

DFurnes:
This should fix our issue lazy-loading in assets over CDN on IE8. Since we're bundling up most of the rest of our assets, it seems to make sense to just bundle everything outright to side-step the IE8 issues and save us the overhead of including the full RequireJS loader.

References #2347.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/bower.json
+++ b/lib/themes/dosomething/paraneue_dosomething/bower.json
@@ -30,6 +30,7 @@
     "unveil": "~1.3.0",
     "jquery-once": "~1.2.6",
     "respond": "~1.4.2",
-    "REM-unit-polyfill": "~1.2.3"
+    "REM-unit-polyfill": "~1.2.3",
+    "almond": "~0.2.9"
   }
 }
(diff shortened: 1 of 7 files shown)


weerd:
So this might not be ideal at all, but would conditionally loading in IE8 stuff in the head, before `app.js` gets called, using IE conditional comments, help to avoid having to load these into all browsers? Might not matter. Just thinking out loud...

weerd:
Aside from comment above, looks good! :+1:

DFurnes:
Yeah, my concern there is falling back to browser sniffing rather than feature detection. They're only 7KB combined, so I'm not too worried about bloat for other browsers.

DFurnes:
Let's keep an eye on it. At some point, I'd like to get back to dynamic script loading for some things, but maybe something we wait on until IE8 support is a non-issue.

weerd:
Cool. Sounds good.

weerd:
While IE8 conditional comments aren't technically browser sniffing and only apply to IE8... browser sniffing would apply to the REM polyfill, since it fill the support gap for any non-rem enabled browser.

PR #2405, opened 2014-06-03
Sms campaign  styles

weerd:
- Added labels to SMS Game phone number fields.
- Updated container module to work better and work according to updates to design. 
- Reworked the sources markup and functionality. Visible without JS, and hidden with JS with toggle to show, and cleanup styles.

@DFurnes


DFurnes:
Looks solid! :hammer:

weerd:
Huzzah! :package: :airplane:

PR #2407, opened 2014-06-03
Add org friend role

### LINKED ISSUE #2398: new user role needed to view all unpublished content
Need a new user role to create accounts for friends of the org (board members, strategic partners, sponsors) so we can show them unpublished content on prod.

This user role should have zero extra privileges other than the ability to view unpublished content on the production site. Would want this to only be content that would have the potential to be user facing (i.e. any of the content types we have) but not anything we wouldn't make user facing (like other user's contact info). Wouldn't want them to have edit access, just view access to these unpublished pages.

@aaronschachter recommended that this role would just be granted a single "view all unpublished" permission.

Open to names for the role, but I think "friend of org" makes sense since this role will be used for:
- board members
- strategic partners
- marketing partners
- sponsors

angaither:
- this role ONLY has the permission to view any unpublished content
- Fixes #2398


aaronschachter:
:+1: ship it

Issue #2379, opened 2014-05-30
[Campaign Close] Press Field Collection


aaronschachter:
New field collection `field_press`:
- Image (Gallery Image)
- Copy (Image Description)

PR #2409, opened 2014-06-03
Add press field group to campaign run.

angaither:
This is just the additional field group.
Refs #2379


aaronschachter:
looks good :+1:

Issue #2378, opened 2014-05-30
[Campaign Close] Winners field collection


aaronschachter:
Fields needed:
- [x] User (EntityReference)
- [ ] Reportback FID
- [x] Winner Name  _this can be loaded from the user object_
- [x] Winner Type (Select list. Scholarship, Grant)
- [x] Winner Description 
- [x] WInner Quote

For display:

If set, render the Reportback FID's `editorial_caption` (#2354) within the template output as well.

angaither:
@aaronschachter
if we have the campaign nid and the user reference two of these things can be filled in automatically.
- reportback fid = query for the report back and get the photo
- Winner name = `field_user_first_name`

aaronschachter:
For reportback FID, a user can upload multiple files to a reportback so we would need to store a specific FID.. 

Actually this issue may need to be updated to use an Image node. @mikefantini can speak to this, as far as usually we ask winners to send through a different image.

aaronschachter:
The idea behind adding the additional Winner Name is if we didn't happent o have a user UID for whatever reason. I think it's fine to implement for comeback clothes by using the User's field_first_name though for now, and add the name into the FC later if we ever end up needing it.

mikefantini:
@angaither @aaronschachter because of SMS games (where there is no report back) we would need an actual image node

@aaronschachter does this affect how other things for this were built?

angaither:
Do we need to worry about the reportback FID then? 
Or when the reportback is flagged, those images will be manually made into image nodes?

aaronschachter:
I think we don't need to worry about Reportback FID for now.  For now we'll just make image nodes for comeback clothes and can revisit how to promote Reportback Files to it.

mikefantini:
yeah I think that makes sense

mikefantini:
wait sorry we'll be ditching FID for winner for right now, but will we still be using it for the report back gallery up at the top of the screen @aaronschachter ?

aaronschachter:
yes that is https://github.com/DoSomething/dosomething/issues/2393

PR #2410, opened 2014-06-03
Adds the easy text fields to winner field group.

angaither:
Currently ignoring reportback files & the display. 
Refs #2378


aaronschachter:
:+1:

PR #2412, opened 2014-06-03
Friends form markup

aaronschachter:
@weerd @angaither Please review.

Creates fieldsets to wrap the alpha/beta form elements:

Gratuitious markup screenshot


weerd:
Sensational! :+1:

PR #2414, opened 2014-06-03
Removing press field

angaither:
- turning this into a gallery type field. 
- stay tuned for exciting things to come.


aaronschachter:
Was this pushed to stage?  If not, this looks good and ship it.  But if so we need to add in deleting the field instance.

angaither:
Aw, damn. looks like it was. Alright.

angaither:
@aaronschachter updated

aaronschachter:
looks good

PR #2415, opened 2014-06-03
Fix SMS Game for auth user with no mobile

aaronschachter:
When the user is logged in, we use the existing function `dosomething_signup_get_mobilecommons_vars`... the kicker is when the user doesn't have a mobile saved.  

This PR checks to see if the phone param has been set after executing `dosomething_signup_get_mobilecommons_vars`.  If not, sets the param needed to the value passed from the form.
## Testing:
- [x] Submit SMS Game as anon user
- [x] Submit SMS Game as auth user with no mobile value saved
- [x] Submit SMS Game as auth user with a mobile value saved

Make sure no errors


angaither:
Nice, good catch. :+1:

PR #2416, opened 2014-06-03
Sms campaign  optimizely script

### LINKED ISSUE #2404: [SMS game] - need to get optimizely up on staging for Bully Text
@Bladt mentioned that we need to add `<script src="//cdn.optimizely.com/js/1158700146.js"></script>` into the header of http://staging.beta.dosomething.org/campaigns/bully-text-0 in order for him to make the optimizely tests we want to run for user testing live; can we get that set up today so we can send off to usertesting.com by EOD?

weerd:
Adding script for optimizely :)

Resolves #2404 

cc @aaronschachter @DeeZone


aaronschachter:
:+1:

PR #2419, opened 2014-06-03
Exclude forms.js because it was exploding. And seemingly useless.

DFurnes:
:+1: :space_invader: :beer: :bear: :beers: :soon:


DFurnes:
@weerd likes it!

PR #2421, opened 2014-06-03
Make two links on Hunt MCC form open in new tab.

DFurnes:
screen shot 2014-06-03 at 5 31 31 pm

:cat: :pencil2: :page_with_curl: :bangbang:


barryclark:
:+1:

Issue #2382, opened 2014-05-30
[Campaign Close] Celebs Field Collection


aaronschachter:
Reuses gallery fields, but `field_celebs`, used to store content about celebs doing campaigns.  This is the section that is not present on the mockup currently, and is also not required for Comeback Clothes (as of the time of creating this issue)

Creating as its own field collection allows us to grab Celebrities doing campaigns easily, vs trying to figure out what Gallery field_collection it is.

PR #2428, opened 2014-06-04
Campaign close preprocess

angaither:
First pass at preprocess the campaign close vars.

There is a _klout gallery_ of galleries, each gallery has a `type` that determines the display of the gallery
- celeb mention
- celeb in action
-  press

Also added a winners array to the template.
For now it grabs the user id, first name, quote and description. 

Todo: images for both klout and winners. 

@aaronschachter and @weerd 

Refs #2382
Refs #2379


lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.features.field_base.inc
```diff
@@ -97,6 +120,33 @@ function dosomething_campaign_run_field_default_field_bases() {
     'type' => 'text',
   );
 
+  // Exported field_base: 'field_type'
+  $field_bases['field_type'] = array(
```
aaronschachter: Is there a reason specifically to call this `field_type` instead of `field_gallery_type` ? We have a `field_campaign_type`..

blisteringherb:
:+1:

PR #2430, opened 2014-06-04
Fixes borked reference to helpers module

desmondmorris:
#### What's this PR do?

The reference to dosomething home is incorrect in the profile make file.  We never noticed this because we do symlink magic in the ds build script.  When running the installation you see this:

screen shot 2014-06-04 at 5 06 42 pm

https://github.com/desmondmorris/dosomething/blob/dev/lib/profiles/dosomething/dosomething.make#L39-L49

Run install.php from a fresh build sans "--install" and confirm that the error in the above screenshot does not appear

Just trying to make the profile play nice.


aaronschachter:
:+1: I've made this change locally in attempts to get simpletests to work, shippy

PR #2432, opened 2014-06-04
Fact page styles

### LINKED ISSUE #2369: [Fact Page] Styles for Fact Page List
Node does not exist on Production yet, but stage node can be found here: http://staging.beta.dosomething.org/facts

tjrosario:
Resolves #2369 for allowing the new Facts listing page to use the new HTML/CSS framework.


weerd:
:+1: set it free!

PR #2434, opened 2014-06-05
Fixes redirect for momm by sending a node object in to check

### LINKED ISSUE #2433: [Mind On My Money] - redirect no longer working
When you go to /money the user should be taken to the pitch page on beta.

Then, when they sign up for the campaign, they should be taken to the legacy page.

This redirect is no longer working.

cc: @barryclark @aaronschachter @blisteringherb

angaither:
Fixes #2433


aaronschachter:
:+1:

PR #2441, opened 2014-06-05
Updated fact page markup to use new refactored markup / css system

### LINKED ISSUE #2435: Update fact view template to use new container markup system

tjrosario:
Fixes #2435

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_fact-page.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_fact-page.scss
@@ -17,11 +17,27 @@
   }
 }
 
+// Styles for Facts Listing page
+.container--facts-list {
+  h3 {
+    margin-bottom: 0.5rem;
+  }
+}
+
 .fact_page-wrapper {
   background: #fff;
 
   h2, h3, h4 {
     color: $purple;
+  }
+
+  ol {
+    list-style: none;
+    padding-left: 0;
+  }
+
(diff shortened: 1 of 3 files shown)


lib/themes/dosomething/paraneue_dosomething/scss/content/_fact-page.scss
```diff
@@ -17,6 +17,13 @@
...
 }
 
+// Styles for Facts Listing page
+.container--facts-list {
+  h3 {
+    margin-bottom: 0.5rem;
+  }
+}
+
 .fact_page-wrapper {
```
weerd: You should be able to get rid of the `fact_page-wrapper`, now that the `<article>` defines the container for the page. We're aiming to remove the need for `something-wrapper` classes.

lib/themes/dosomething/paraneue_dosomething/scss/content/_fact-page.scss
```diff
@@ -24,6 +31,15 @@
...
+  ol {
+    list-style: none;
+    padding-left: 0;
+  }
+
+  li {
+    margin-bottom: 1.5rem;
+  }
+
   .header-wrapper {
```
weerd: Same with this one. See above comment.

I didn't see an item in the markup w/ a `header-wrapper` class that this would apply to...

weerd:
Cool! Looks good :+1:

PR #2446, opened 2014-06-06
Added missing bracket.

DFurnes:
:racehorse:


weerd:
tableflip-1349463264

:+1:

PR #2448, opened 2014-06-09
Polyfill fixes

DFurnes:
- Uses cross-domain proxy for Respond.js.
- Move polyfills to end of body to prevent DOM not loaded issues.
- Rename `LIB_ASSET_PATH` to `VENDOR_ASSET_PATH` to be a bit clearer.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/system/html.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/system/html.tpl.php
@@ -29,6 +29,11 @@
       <link type="text/css" rel="stylesheet" href="<?php print DS_ASSET_PATH; ?>/ie.css" media="all" />
 
       <script type="text/javascript" src="<?php print VENDOR_ASSET_PATH; ?>/html5shiv/dist/html5shiv.min.js"></script>
+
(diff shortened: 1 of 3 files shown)


lib/themes/dosomething/paraneue_dosomething/templates/system/html.tpl.php
```diff
@@ -45,9 +42,17 @@
   <?php print $page_top; ?>
   <?php print $page; ?>
 
+  <!--[if lte IE 8]>
+      <script type="text/javascript" src="<?php print VENDOR_ASSET_PATH; ?>/respond/dest/respond.min.js"></script>
```
weerd: So now that `respond.js` is being added before the `<body>` tag, we'll likely have a bit longer of a FOUC effect.
From RespondJS readme: "Reference the respond.min.js script (1kb min/gzipped) after all of your CSS (the earlier it runs, the greater chance IE users will not see a flash of un-media'd content)"

DFurnes: Moved it back up into the head, just leaving REM-Unit-Polyfill in the footer because that's the guy that was having the intermittent problems. That being said, there's going to be a FOUC until the rem polyfill kicks in, so it's still not going to be "ideal" for IE users.

weerd:
:+1:

Issue #2356, opened 2014-05-28
Reportback: Promote a Reportback File


aaronschachter:
Promote a specific Reportback File, to a campaign or campaign_run Reportback gallery.  If we wanted to use Flag module where, we'd want need to add File Entity into the mix to do this.  File Entity may come in handy for the Editorial Caption issue #2354.  Create a corresponding view `node/[nid]/reportbacks/promoted`.

Notes on File Entity pros and cons investigation here: https://gist.github.com/aaronschachter/df421ea414894aed7ddc
- The promoted reportback files which are promoted should be sortable. Ideally this sortable form would live on the `node/[nid]/reportbacks/promoted` page. Possible implementations could be:
  - Our own custom property on the reportback/reportback files table
  - Something like Entityqueue, which may add more code/functionality than we want.

---

Once this is in place:
#2394: The Campaign Action Page Reportback Gallery should display Promoted files, in order of sequence.   Same for the Campaign Run gallery, perhaps it would display the top 3.

aaronschachter:
Renamed this issue and created #2395 to handle flagging a Reportback Entity, as these are two different Entities to flag (Entity for bad, specific File for good)

aaronschachter:
Closed by #2499

PR #2458, opened 2014-06-09
File Entity + Editorial Caption

aaronschachter:
**UPDATE:**

Adds File Entity and Flag module into the repo.  Sets the pathauto alias for Files to `file/[fid]`, the File entity's default URI, which prevents `url_alias` from created when we upload files.

We'll use a Promoted Flag in order to promote Reportback Files to the Campaign Run gallery, and add a description field to the flag entity itself for a better editorial experience.  This PR adds Flag into the mix as well to allow development to begin on Dynamic Galleries.

---

**Original PR:**

Adds File Entity module into the DS profile in order to:
- Add an Editorial Caption field to the File Entity and link to edit file from Reportbacks view.
- Allow use of Flag module to promote Reportback Files to Campaign / Run galleries, refs #2356

Will add a comment with a gist I've been working on detailing out pros/cons, but wanted to open the PR first. 
## Screenshots

Reportback File List
Reportback File Edit
admin-content-files

### DIFF
--- a/lib/modules/dosomething/dosomething_settings/dosomething_settings.info
+++ b/lib/modules/dosomething/dosomething_settings/dosomething_settings.info
@@ -13,3 +13,4 @@ features[ctools][] = strongarm:strongarm:1
 features[features_api][] = api:2
 features[filter][] = markdown
 features[user_permission][] = use text format markdown
+features[variable][] = pathauto_file_pattern
--- a/lib/modules/dosomething/dosomething_settings/dosomething_settings.strongarm.inc
+++ b/lib/modules/dosomething/dosomething_settings/dosomething_settings.strongarm.inc
@@ -1,0 +1,21 @@
+<?php
+/**
+ * @file
+ * dosomething_settings.strongarm.inc
+ */
+
+/**
+ * Implements hook_strongarm().
+ */
+function dosomething_settings_strongarm() {
+  $export = array();
+
+  $strongarm = new stdClass();
+  $strongarm->disabled = FALSE; /* Edit this to true to make a default strongarm disabled initially */
+  $strongarm->api_version = 1;
+  $strongarm->name = 'pathauto_file_pattern';
+  $strongarm->value = 'file/[file:fid]';
+  $export['pathauto_file_pattern'] = $strongarm;
+
+  return $export;
+}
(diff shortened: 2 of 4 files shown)


aaronschachter:
I've updated the pull request description with screenshots and have also listed out some Pros and Cons of adding File Entity into our build here: https://gist.github.com/aaronschachter/97c483fa9d4d9c315a40

Would be great if others to chime in with opinions, cc @angaither @desmondmorris @mshmsh5000 @barryclark @blisteringherb 

I'm leaning towards adding it in.  Gives us a ton of functionality out of the box that we'd otherwise have to write custom code for.  This is all really for admin purposes of setting properties and promoting Reportback files: I don't think we'd gain much by writing our own custom code to handle.  Note that the flagging/promoting of Reportback Files will come in a subsequent pull request if we go this route.

desmondmorris:
If one of the necessitates is to allow us to flag files, then I think file entity is a good option.  Though, shouldn't the editorial caption belong to the report back entity and not the file?

aaronschachter:
Flagging files is definitely a big reason to add the module in.  As far as the editorial caption:

The editorial caption is for a promoted Reportback Image, which you can see a mock of in the [Campaign Closed State mocks](https://sites.google.com/a/dosomething.org/dosomethinguxwiki/home/product-archive/campaign-template?pli=1)

Each reportback can have multiple files, which we store in a table `dosomething_reportback_file`, which is simply a table of 3 fields: Reportback ID, File ID, and the IP address .  

We could potentially store the editorial caption here, however it would require custom code to create the form to set the field, and to load the field when loading the Reportback file, for when we display in a Campaign Gallery.  

I went the route of using File Entity to get all that stuff out of the box, although one could argue that an Editorial Caption only applies to specific image files, and shouldn't be available to ALL image files. The field value itself is only saved into the DB when someone goes into a file and edits it.

desmondmorris:
I think you achieve this all through the Flag module.  Add the caption to the actually flag.

aaronschachter:
Chatting with @desmondmorris about that last comment. 

If editors will always be the **only** users who add a caption to a Reportback Image, it definitely makes sense to store on the Flag.

However if we ever wanted to explore allowing users to submit their own funny captions, it would make sense to keep this field on the actual File.

thoughts, @barryclark @mikefantini ?

aaronschachter:
Per our chat today. Going to remove the `field_editorial_description` from the File entity itself and build it out on a Promoted Flag, using Flag module to build out dynamic Reportback Galleries.

Will update this PR and ping for a plus 1 shortly.

aaronschachter:
Updated this PR to remove the `field_editorial_caption` from the File Entity, and add Flag module into the mix instead.  Ready for review, cc:// @angaither @desmondmorris

angaither:
:+1:

PR #2459, opened 2014-06-09
linked gallery item image

### LINKED ISSUE #2454: Static Content - Gallery (link photos)
Add ability for static content pages to link photos like on this What is DoSomething.org page.

screen shot 2014-06-09 at 12 14 30 pm

tjrosario:
Fixes #2454

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/static_content/node--static_content.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/static_content/node--static_content.tpl.php
@@ -68,10 +68,10 @@
             <?php foreach ($gallery['items'] as $gallery_item): ?>
               <div class="gallery-item">
                 <?php if (isset($gallery_item['image'])): ?>
-                  <?php print $gallery_item['image']; ?>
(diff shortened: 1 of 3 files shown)


lib/themes/dosomething/paraneue_dosomething/templates/static_content/node--static_content.tpl.php
```diff
@@ -68,10 +68,10 @@
             <?php foreach ($gallery['items'] as $gallery_item): ?>
               <div class="gallery-item">
                 <?php if (isset($gallery_item['image'])): ?>
-                  <?php print $gallery_item['image']; ?>
+                  <a href="<?php print $gallery_item['image_url']; ?>"><?php print $gallery_item['image']; ?></a>
```
weerd: Is the image link a required field? If not, should likely check to make sure it was filled in before outputting as a link around the image.

aaronschachter: Good catch, it's not required. yes we need to check if isset

weerd:
Cooooool! :+1:

Issue #2452, opened 2014-06-09
11-facts - sources


joolaay:
Holla,

Let's hook up sources with the same treatment we use in campaigns. 

_Sources_ should go to the bottom of each fact page. When clicked, it should open a drop down with  gray sourcing info. cc @crystalruthbell

tjrosario:
@crystalruthbell once staging is updated, sample Fact with source can be found here: http://staging.beta.dosomething.org/facts/11-facts-about-education-and-poverty-america

joolaay:
OOooh ballz. We actually have to get the web address sources that are
hooked up to the facts on the backend. So there should be a list of 11
sources in a BIG ass drop down. I can come explain - Crystal, do you know
what I'm talking about?

On Tue, Jun 10, 2014 at 1:07 PM, Tom Rosario notifications@github.com
wrote:

> Closed #2452 https://github.com/DoSomething/dosomething/issues/2452.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/DoSomething/dosomething/issues/2452#event-129886029.

## 

Julie Lorch

aaronschachter:
@tjrosario Check out the markup in the base template for Fact Page: nodes: https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_fact_page/node--fact_page.tpl.php#L34

aaronschachter:
Update -- the module's tpl file was never updated to output sources the way we are doing so in the theme's tpl file, per this commit https://github.com/DoSomething/dosomething/commit/eac289a0f5eed0d6ec30703758391cab5ba7590f

We need to set the `$sources` variable back to the array of $sources. Ideally we should also clean this up and remove the `additional_text` variable and field from Fact Pages.

PR #2464, opened 2014-06-10
Updated source toggle js to accomodate fact page source markup

tjrosario:
The markup for sources is unique for Facts because it's type is string versus an array in all other scenarios, therefore the JS had to be modified to accommodate.


weerd:
Is there a way to preprocess the output to make it a list? Maybe something to consider for later, for a quick refactor. Otherwise :+1:  :)

tjrosario:
Yes, agreed.  It should be streamlined across the board.  Will make a separate ticket for it.

tjrosario:
Fixes #2452

Issue #2468, opened 2014-06-10
Refactor Fact sources markup


tjrosario:
Refactor fact sources markup to match current source pattern

Relates to #2452

PR #2469, opened 2014-06-10
updated fact page module to returns sources as an array and updated markup on the front-end to match the current sources pattern

### LINKED ISSUE #2452: 11-facts - sources

tjrosario:
Fixes #2452 and #2468


aaronschachter:
Bender approved :+1:

weerd:
Nice! Awesome work :) :+1:

Issue #2377, opened 2014-05-30
[Campaign template] Hide campaign from Finder


aaronschachter:
Refs desired functionality specified here: https://github.com/DoSomething/dosomething/issues/2348
- [ ] Add setting to hide node from Finder (without unpublishing)
- [ ] Configure SOLR to exclude node from Finder based on this setting

This may be a simple checkbox, but could also possibly suggest that we need a Campaign State field, with different states:
- Pre-Launch
- Live
- Closed

It depends on if there are different views for Pre-Launch and Closed.  @mikefantini @barryclark Thoughts?  I don't think it makes much sense that you'd be able to find any campaigns within the Finder that don't have Campaign State == 'Live'

Ideally this could be more of a variable than a field, as we wouldn't necessarily need to be tracking revisions and the data is really simple.  Potentially similar to the `dosomething_signup_variable` concept.. maybe that should just be abstracted out to a simple `dosomething_variable` table where we can store things like campaign status (and both the Campaign and Grouped Campaign content types could make use of it).

mshmsh5000:
I can see an editorial case for having manual control over whether a particular campaign gets indexed. I see three possibilities:
1. An `is_search_indexed` field exposed to editors, and manually managed
2. A variable not exposed to editors, managed when the campaign transitions to a new state (pre-live, live, closed)
3. A field whose value is managed during campaign transitions, but which editors can also see and manually override

As usual, I lean towards the option that gives us the most flexibility, which is the third. But I'd love to hear opposing opinions, @barryclark / @mikefantini / @aaronschachter.

aaronschachter:
I like option 3, would this field be what actually transitions the campaign into the Closed state?  Vs. anything automated based on End Date values?  

We actually don't have a closed date anywhere on the campaign: just low season and high season -- which don't actually detail when the campaign ends, but controls its weight in the Finder.

mikefantini:
Late following up on this but I think the critical path for comeback clothes we should follow is (as @barryclark touched upon in his email on Monday):
- editor manually puts the campaign in a closed state (since new information is needed I'm not comfortable yet with campaigns automatically going into a closed state since the editor may not have put in the necessary info)
- when a campaign is in the closed state, it should be removed from the finder AND /campaigns; but is still discoverable via search

I'm going to confirm this in management meeting this morning and will update this issue based on that conversation.

mikefantini:
confirmed from management meeting that for now, when something goes into the closed state it should:
[...]

PR #2470, opened 2014-06-10
Campaign status field

### LINKED ISSUE #2450: [Campaign Close] Entity reference on campaign node
When we have multiple campaign closes we need a way to determine which campaign run is the active run. 

We also need a way to view the past campaign closed references.

My recommendation is to have the campaign run keep the date that the campaign closes and both the campaign & the campaign run to have a [reverse reference](https://drupal.org/project/cer). 

This way each year the campaign lead is only making a new campaign run and doesn't have to worry about last year's node.

angaither:
- adds `campaign_status` field to the campaign, currently only has two states.
- updated `dosomething_campaign_is_closed` function to use the new field
- display the current campaign run nid under the status field

Refs #2377 
Closes #2450


aaronschachter:
:+1:

PR #2471, opened 2014-06-10
Campaign Run node updates

### LINKED ISSUE #2465: [campaign close] - add User First Name field to Winner field collection
Field name "User First Name" (open to it being "Winner First Name" or whatever you think makes sense).

Help text "User First Name should only be needed if there isn't a user to lookup. If there is a user to lookup, then we'll pull in their first name from the UID."

cc: @aaronschachter 
- [ ] Add `field_winner_name` to `field_winners` field collection
- [ ] Use value in preprocess for when a User UID not set

angaither:
- Updated campaign run based on [mocks](https://docs.google.com/a/dosomething.org/file/d/0B5fENCbpVZv9VG9YbzFWUURJQTA/edit)
- Added first name field.
- Use that value if no uid.

Closes #2465


lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.features.field_base.inc
```diff
@@ -33,6 +33,33 @@ function dosomething_campaign_run_field_default_field_bases() {
     'type' => 'field_collection',
   );
 
+  // Exported field_base: 'field_gallery_type'
+  $field_bases['field_gallery_type'] = array(
```
aaronschachter, line 37: Did anything change here, or is this just weird features update stuff?

angaither, line 37: features update stuff

aaronschachter:
:+1: value has been added, value added.

Issue #2449, opened 2014-06-09
Campaign Close - Pre-signup button adds user to MailChimp


barryclark:
~~Blocked by #2463~~
- [x] New pre-signup form rendered in campaign header
- [x] Pre-signup form submit 
  - [x] Write a presignup record
  -  [ ] Subscribe user to Mailchimp
  - [x]  Upon submit, display the copy  "Sweet, we'll send you an email as soon as [campaign_title] re-opens. In the meantime, <a href="/campaigns">find a campaign</a> you can do right now."

aaronschachter:
@barryclark @mikefantini 
- Can you confirm this is Mailchimp only and no Mobilecommons?  
- should a message appear after signing up, and if so what is the copy
- Is a transactional email to be sent?

mikefantini:
@aaronschachter 
- yes Mailchimp only and not Mobile Commons
- yes, the user should get a notification, for copy let's go with "Sweet, we'll send you an email as soon as [campaign_title] re-opens. In the meantime <a href="/campaigns">find a campaign</a> you can do right now."
- I don't think we need to send a transactional email letting them know that we will be sending them an email in the future, but @marahml do you have any thoughts on this?

For context, when a campaign is in the closed state, users will be able to sign up to get notified via email when the campaign re-opens. The idea is that we will add the user to the campaign's MailChimp group and then send out a broadcast when the campaign re-opens.

Eventually we may want to do message broker-y things, but this is the idea for the first implementation.

aaronschachter:
We are currently handling Mailchimp subscriptions via Message Broker.

@deezone Is there a function to call which will only subscribe a user to a Mailchimp Group (and no transactional)?

mikefantini:
@aaronschachter right... i guess i meant rather than send a broadcast, handle the sending of the message once the campaign re-opens via the message broker.

Nothing we need to worry about yet though

aaronschachter:
@mikefantini  What I meant was that just to get a user subscribed to Mailchimp in the beta app, we use Message Broker to handle that.  There is no Mailchimp code within the DS repo.

mikefantini:
right makes sense

marahml:
@mikefantini @aaronschachter yes - we should send users a transactional email anytime they sign up for something on the site, even if its just for future updates. (ie. We'll tell you when Thumb Wars is ready again next year, but in the meantime do these campaigns).

[...]

Issue #2472, opened 2014-06-10
[Campaign Close] Presignup on user login/register


aaronschachter:
Alter login logic so that if a user is signing in on a closed campaign, call `dosomething_signup_user_presignup` instead of `dosomething_signup_user_signup`.  See #2473

PR #2473, opened 2014-06-10
Presignup form

aaronschachter:
@angaither Please review.

Refs #2449

Places a presignup form on the Campaign Closed template, and writes a presignup record upon submitting.

If a user clicks it over and over, they'll get the confirmation message but we'll only store the presignup once.

Sets up placeholder code for adding Mailchimp Presignup subscription: https://github.com/DoSomething/message_broker_producer/issues/35


angaither:
:+1: Looks good. Does anything need to change in dosomething_user?
Specifically `_dosomething_user_add_signup_data`

aaronschachter:
Yup, I assigned you #2472.

barryclark:
@DeeZone @marahml @mikefantini This means that we won't need to do any of the MailChimp message broker work for launch on the 19th, we'll be able to tackle it after that, as we're storing the pre-signups.

Issue #2475, opened 2014-06-10
[Campaign Close] Cleanup dosomething_campaign_is_closed function


aaronschachter:
Now that we have a `field_campaign_status` in place, the `dosomething_campaign_is_closed` function should simply return whether or not the field value == `closed` and we should create a separate closed page access callback.
- [x] Simplify `dosomething_campaign_is_closed` to return value == closed
- [x] Create a `_dosomething_campaign_closed_page_access` callback which checks for staff only, and whether or not a campaign run exists. Alter the `node/*/closed` page to use this callback.
- [x] Alter `dosomething_campaign_preprocess_node` check for closed to be an OR:

```
  if (dosomething_campaign_is_closed($node) || [staff is viewing closed page] ) {
    // Use the campaign closed template to theme.
    $vars['theme_hook_suggestions'][] = 'node__campaign__closed';
   ...
```

PR #2478, opened 2014-06-10
Cleanup dosomething_campaign_is_closed function

aaronschachter:
@angaither Please review

Issue details in #2475 but this removes staff / current path checks out of `dosomething_campaign_is_closed` and into an access callback and check within the preprocess.

Without a proper access callback, and simply checking if `dosomething_campaign_is_closed`, the user will be presented with both the "View" and "Closed" tab when the campaign is actually closed.


angaither:
:+1:

aaronschachter:
I've found a bug in this where it's not displaying the Closed page for anonymous users but the Pitch instead. Will ping when i've updated the PR.

angaither:
On all campaigns?

PR #2485, opened 2014-06-11
Changed the preprocess to match the actual field name.

angaither:
Noticed this bug when working with @tjrosario on the template.


aaronschachter:
:+1:

Issue #2486, opened 2014-06-11
Campaign closed Pitch Page displaying


angaither:
When an anon user views a closed campaign, the pitch page is displaying, not the 'closed' page.

PR #2487, opened 2014-06-11
On login submit, sign the user up for the correct campaign state.

### LINKED ISSUE #2472: [Campaign Close] Presignup on user login/register

angaither:
Check to see if campaign is closed, if so 'presignup' else signup.

This currently works for logged in users. Will probably also worked for anon users once #2486 is fixed.

Fixes #2472


aaronschachter:
:+1: Looks good.

PR #2488, opened 2014-06-11
Check for `is_closed` before `is_pitch`

### LINKED ISSUE #2486: Campaign closed Pitch Page displaying

angaither:
If a campaign is closed, we want to show that before we show the pitch page.
This is no code changes, just order changes.

Fixes #2486

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -49,6 +49,8 @@ function dosomething_campaign_preprocess_node(&$vars) {
     // Use the campaign closed template to theme.
     $vars['theme_hook_suggestions'][] = 'node__campaign__closed';
     dosomething_campaign_run_preprocess_closed($vars);
+    // Preprocess common vars between all campaign types.
+    dosomething_campaign_preprocess_common_vars($vars, $wrapper);
     return;
   }
 
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -41,17 +41,6 @@ function dosomething_campaign_preprocess_node(&$vars) {
   // Add inline css based on vars.
   dosomething_helpers_add_inline_css($vars);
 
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -63,6 +52,19 @@ function dosomething_campaign_preprocess_node(&$vars) {
...
+  if (dosomething_campaign_is_pitch_page($node)) {
+    // Use the pitch page template to theme.
+    $vars['theme_hook_suggestions'][] = 'node__' . $vars['type'] . '__pitch';
+    // Gather vars for pitch page.
+    dosomething_campaign_preprocess_pitch_page($vars, $wrapper);
+    return;
+  }
+
+  // Preprocess common vars between all campaign types.
+  dosomething_campaign_preprocess_common_vars($vars, $wrapper);
```
aaronschachter, line 66: I was going down this route and then realized my vagrant was all messed up. BUT. this line.  `dosomething_campaign_preprocess_common_vars` is preprocessing some things that we need for a campaign closed state... its basically used for preprocessing campaigns in a full view mode that aren't used for pitch page.  

I was looking into altering the pitch page logic to make sure a campaign isnt' closed... yet we still want the ability to staff to view the Pitch page at `node/*/pitch`

aaronschachter, line 66: one easy way to handle could just be to call preprocess_common_vars from within the closed check.  and then it just gets called also outside of the closed check outside for non-pitch campaign displays

angaither, line 66: Looking at whats in that function, it seems the only thing we need form it is the zenddesk form and the media vars. Everything else is specific to when the campaign is open. I was thinking of just adding those two into preprocess for the closed page.

https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc#L82-L105

angaither:
Scratch that, it calls a slew of other functions. I added it into the closed template.

aaronschachter:
:+1:

PR #2498, opened 2014-06-12
Fixes #2497: Uses private subnet IPs for stage/prod deploy

### LINKED ISSUE #2497: Cap deploys to public production and staging IPs
The [config settings](https://github.com/DoSomething/dosomething/blob/dev/config/deploy/production.rb) refer to a public IP, which is sub-optimal: (a) it publishes a public IP that we don't / shouldn't need for a web head, and (b) it requires both a public IP and an open SSH port in the firewall.

Instead, use private subnet equivalents for production deploy targets.

mshmsh5000:
Requires the deploying server to be either on the private subnet, or to have access (i.e., via VPN).

### DIFF
--- a/config/deploy/production.rb
+++ b/config/deploy/production.rb
@@ -1,8 +1,8 @@
 set :deploy_to, "/var/www/beta.dosomething.org"
 
-role :app, %w{dosomething@98.129.111.169}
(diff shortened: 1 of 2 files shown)


desmondmorris:
ship it

Issue #2394, opened 2014-05-31
[Campaign Close] Dynamic Reportback gallery


aaronschachter:
Depends on #2356.  If there is a flag in place to promote Reportback Files, we can display the top X promoted instead of implementing this via a Reportback File reference field #2393.

For short term, we'll implement #2393 first as it's a much smaller lift, so I'm labelling this issue as low priority for now.

This would also apply a Campaign Action Page, could use the smae functionality instead of the Image node Entityreference field for the "Prove It" gallery.

aaronschachter:
Ready for theming. Assigning over to @tjrosario

PR #2499, opened 2014-06-12
Campaign Reportback gallery

aaronschachter:
Handles back-end for #2394

Adds a "promoted" Flag used to promote Reportback Files  The promoted flag has two fields:
- `field_weight` -- The sequence in which the file should appear in the Campaign Reportback Gallery. New field.
- `field_image_description` -- The editorial caption. Existing field.

Editors can add a reportback file from within the `node/*/reportbacks` view. Upon adding it, they are prompted to enter an editorial caption.

There is now a `node/*/gallery` page for editors to sequence the promoted files using drag and drop, and maintain the captions for each.  This is implemented via Form API and Drupal's `drupal_add_tabledrag` functionality.
## Next steps:

Use `dosomething_reportback_get_gallery_vars` to output the Reportback Gallery into the Closed template.
## Screenshots:

reportback-gallery1
reportback-gallery2
reportback-gallery3
reportback-gallery4


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -730,3 +762,57 @@ function dosomething_reportback_get_reportback_field_info($nid, $name = NULL) {
...
 }
+
+/**
+ * Returns flagging_id, fids, rbid's for given $nid ordered by field_weight.
+ */
+function dosomething_reportback_get_gallery_item_ids($nid) {
+  // @todo: Future proof this.  Need to add a condition
+  // where flag fid == promoted in case we add another file flag.
+  $sql = "SELECT flagging_id, rbf.fid, rbf.rbid, field_weight_value
+    FROM flagging
```
angaither, line 773: not needed for merge, but would be nice to dynamic query this.

angaither:
:+1:

screenshots! yaaaaa

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
aaronschachter, line 773: Agree.  At first I thought this query was gonna be way harder than it is. Theres still some cleanup to do here (will show Gallery tab on all nodes) and I can give it a shot then.

PR #2500, opened 2014-06-12
Add page override for fact page.

DFurnes:
Adds a page override template for fact page.

Before:
before

After:
after


DFurnes:
The fact page can probably just use the default template, but this is a quick fix.

barryclark:
:+1:

PR #2501, opened 2014-06-12
Salt Provisioning: fix apache2/php.ini provisioning.

sergiitk:
Currently vagrant provisioning does not override system php.ini file for apache2 with the one provided by `lamp-drupal` salt state tree.
The reason for that is a logic error in `lamp-drupal.sls`. It was assumed that `/etc/php5/apache2/php.ini` is provided by `php5` package, however, it is provided by `libapache2-mod-php5`.
Thus salt-managed php.ini is copied too early so it gets overridden by the default version when `libapache2-mod-php5` package is being installed.
This commit corrects that by making `libapache2-mod-php5` responsible of managing `php5/apache2/php.ini`.

### DIFF
--- a/provision/salt/roots/salt/lamp-drupal.sls
+++ b/provision/salt/roots/salt/lamp-drupal.sls
@@ -13,11 +13,6 @@ php5-pkgs:
       - php5-gd
       - php5-memcache
       - php-apc
-  file.managed:
-    - name: /etc/php5/apache2/php.ini
-    - source: salt://php5/apache2/php.ini
-    - require:
-      - pkg: php5-pkgs
 
 php5-cli-config:
   file.managed:
@@ -105,6 +100,11 @@ libapache2-mod-php5:
     - installed
     - require:
       - pkg: apache2
+  file.managed:
+    - name: /etc/php5/apache2/php.ini
+    - source: salt://php5/apache2/php.ini
+    - require:
+      - pkg: php5-pkgs
 
 memcached:
   pkg:


sergiitk:
This also enables displaying php errors and increases memory limit.
Here's the difference between default php.ini and the one, provided by DS provisioning.

``` diff
@@ -289 +289 @@
-zlib.output_compression = Off
+zlib.output_compression = On
@@ -465 +465 @@
-memory_limit = 128M
+memory_limit = 384M
@@ -521 +521 @@
-error_reporting = E_ALL & ~E_DEPRECATED
+error_reporting = E_ALL | E_STRICT
@@ -538 +538 @@
-display_errors = Off
+display_errors = On
@@ -1870,0 +1871,4 @@
+
+; PHP Error
+; auto_prepend_file = "/srv/salt/php5/php_error.php"
+; php_error.autorun = On
```

sergiitk:
One thing though. There's no doubts that php.ini should be provisioned from `provision/salt/roots/salt/php5/apache2/php.ini`, but I'm a little worried about displaying errors on production.
Because of the bug I described above `php.ini` wasn't delivered to vagrant box so errors were muted. Now it's fixed so we better make sure production errors on production are explicitly disabled.

aaronschachter:
@sergii-tkachenko How can we test this?  If I pull down your branch, do I need to `vagrant destroy` and then `vagrant up`, or does it work to just pull down the branch and test around the site?

sergiitk:
@aaronschachter To perform the test, vagrant must be destroyed and then reprovisioned. So yes, `vagrant destroy` and `vagrant up`.

Simple pull request will change nothing since salt-managed files are not mounted to vbox realtime. All directives described in .sls scripts actually take place when `vagrant provision` is running. It is started automatically on the first up.

However, if you decided to change .sls and apply it to existing vbox, you can run `vagrant provision` manually. Salt will detect the change and apply it to the box.

But it's still not a clean test: second `vagrant provision`, in fact, would replace apache2/php.ini disregarding of my fix — libapache2-mod-php5 is already installed at this point and can't reoverride salt-managed php.ini.

sergiitk:
By the way, @mshmsh5000 confirmed that this change doesn't affect production.
Production severs are not provisioned by this scripts at all.

mshmsh5000:
+1 /cc @sergii-tkachenko Provisioning and php.ini worked for me.

sergiitk:
@mshmsh5000, That's great, thank you!

aaronschachter:
I've recently vagrant destroyed / up and not seeing PHP errors. @angaither also seems to be having the same issue...

sergiitk:
@aaronschachter I'll look into that.
@mshmsh5000 I'd suggest installing [`php-xdebug`](http://xdebug.org/) and turn on html_errors. Since salt provisioning is currently used only for development boxes errors displaying should be as loud as possible, am I right? It also prettifies var_dumps: http://xdebug.org/docs/display

sergiitk:
@aaronschachter From what I see, this is not the same issue:

```
$ diff -s /vagrant/provision/salt/roots/salt/php5/apache2/php.ini /etc/php5/apache2/php.ini
Files /vagrant/provision/salt/roots/salt/php5/apache2/php.ini and /etc/php5/apache2/php.ini
are identical
```

mshmsh5000:
@sergii-tkachenko Yes, Salt only provisions dev boxes. So loud errors should be fine.

PR #2507, opened 2014-06-12
Reportback Gallery preprocess

aaronschachter:
@angaither Please review.

Completes bender work for #2394, making ready for handoff to fender.

Also fixes bug where Gallery tab would appear for all nodes, by adding `dosomething_reportback_gallery_form_access`.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -788,10 +788,11 @@ function dosomething_reportback_get_reportback_field_info($nid, $name = NULL) {
 function dosomething_reportback_get_gallery_item_ids($nid, $num_items = NULL) {
   // @todo: Future proof this.  Need to add a condition
   // where flag fid == promoted in case we add another file flag.
-  $sql = "SELECT flagging_id, rbf.fid, rbf.rbid, field_weight_value
+  $sql = "SELECT flagging_id, rbf.fid, rbf.rbid, file.uid, field_weight_value
     FROM flagging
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -202,6 +201,18 @@ function dosomething_reportback_access($op, $entity, $account = NULL) {
...
 
+/**
+ * Access callback for a campaign reportback gallery form.
+ */
+function dosomething_reportback_gallery_form_access($node) {
+  // If viewing a campaign-campaign:
+  if (dosomething_campaign_get_campaign_type($node) == 'campaign') {
+    global $user;
+    $access_roles = array('administrator', 'editor');
+    $has_access = array_intersect($access_roles, $user->roles);
```
angaither: can't you just call `user_access` here?

aaronschachter: There isn't really an easy one... unless we add a new permission.
- An administrator doesn't have `edit any campaign content` because they have `bypass node access`.
- A member support is able to `view any reportback` but they shouldn't be able to set these.

aaronschachter: Open to suggestions though...

angaither: Maybe if the user has the ability to flag/unflag things? Wasn't that perm exported in the last PR?

aaronschachter: totes. updating.

angaither:
:+1:

PR #2508, opened 2014-06-12
Adds additional fields to send to campaign cache MBP transactions

### LINKED ISSUE #2237: Add campaign details to dosomething_mbp_campaign_request()
`dosomething_mbp_campaign_request()` needs access to the following node variables:
- path to image reference in edit-field-image-campaign-cover-und-0-target-id
- fact problem text referenced in edit-field-fact-problem-und-0-target-id
- fact solution text referenced in edit-field-fact-solution-und-0-target-id
- high season start date referenced in:
  - edit-field-high-season-und-0-value-month
  - edit-field-high-season-und-0-value-day
  - edit-field-high-season-und-0-value-year
- high season end date referenced in:
  - edit-field-high-season-und-0-value2-month
  - edit-field-high-season-und-0-value2-day
  - edit-field-high-season-und-0-value2-year
- low season start date referenced in:
  - edit-field-low-season-und-0-value-month
  - edit-field-low-season-und-0-value-day
  - edit-field-low-season-und-0-value-year
    -low season end date referenced in:
  - edit-field-low-season-und-0-value2-month
  - edit-field-low-season-und-0-value2-day
  - edit-field-low-season-und-0-value2-year
- report back details:
  - edit-field-reportback-noun-und-0-value
  - edit-field-reportback-verb-und-0-value
  - edit-field-reportback-copy-und-0-value
  - edit-field-reportback-confirm-msg-und-0-value

to ultimately send in $payload for a `message_broker_producer_request()`.

deezone:
Fixes #2237 
- Adds campaign related field values to payload for Message Broker campaign cache:
  - high season start
  - high season end
  - low season start
  - low season end
  - campaign image
  - fact problem
  - fact solution
  - call to action
- Adds admin link to module from module listing page.

### DIFF
--- a/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
+++ b/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
@@ -242,61 +242,60 @@ function dosomething_mbp_node_update($node) {
  */
 function dosomething_mbp_campaign_request($action, $node) {
 
-  $url = 'http://dosomething.org/' . drupal_lookup_path('alias', "node/" .
-    $node->nid);
-  $field = field_get_items('node', $node, 'field_call_to_action');
-  $call_to_action = $field[0]['value'];
-  // edit-field-image-campaign-cover-und-0-target-id
-  // @todo: reference field, need to look up referenced image path
-  $field = field_get_items('node', $node, 'field_image_campaign_cover');
-  $image_campaign_cover = $field[0]['value'];
-  // edit-field-fact-problem-und-0-target-id
-  // @todo: reference field, need to look up referenced fact node problem
-  $field = field_get_items('node', $node, 'field_fact_problem');
-  $fact_problem = $field[0]['value'];
-  // edit-field-fact-solution-und-0-target-id
-  // @todo: reference field, need to look up referenced fact node solution
-  $field = field_get_items('node', $node, 'field_fact_solution');
-  $fact_solution = $field[0]['value'];
-  /**
-   * @todo: Add high season dates:
-   * START
-   *   edit-field-high-season-und-0-value-month
-   *   edit-field-high-season-und-0-value-day
-   *   edit-field-high-season-und-0-value-year
-   * END
-   *   edit-field-high-season-und-0-value2-month
-   *   edit-field-high-season-und-0-value2-day
-   *   edit-field-high-season-und-0-value2-year
-   *
-   * Low season dates:
-   * START
-   *   edit-field-low-season-und-0-value-month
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
```diff
@@ -242,61 +242,60 @@ function dosomething_mbp_node_update($node) {
...
   $url = 'http://dosomething.org/' . drupal_lookup_path('alias', "node/" .
     $node->nid);
+  $params['url'] = $url;
+
   $field = field_get_items('node', $node, 'field_call_to_action');
-  $call_to_action = $field[0]['value'];
-  // edit-field-image-campaign-cover-und-0-target-id
-  // @todo: reference field, need to look up referenced image path
+  if (isset($field[LANGUAGE_NONE][0]['value'])) {
+    $params['call_to_action'] = t($field[LANGUAGE_NONE][0]['value']);
```
aaronschachter: Are you using `t` here to strip out the funky characters? If its for actually translating, thats the whole point of the LANGUAGE_NONE

deezone: Yes, as you suggested, t() is to get around the funky Word characters. My understanding is they will simply get stripped out. On the other hand, I like your suggestion of just leaving the text as is. It's more a user error than a tech problem. 

What's your final position on the Drupal side?

aaronschachter: Um I haven't really come to one. Would be curious what the rest of team Drupal has run into to deal with this. cc @desmondmorris @blisteringherb @angaither @mshmsh5000 

I had a colleague in a previous life once say that none of the filter stuff really truly does the job and it's more of a training issue to just copy/paste into Notepad first then into the text field.

Have you tested that using `t()` does indeed strip out the characters? I haven't, I just thought it could be a quick win.

deezone: I've backed out of the use of t() as it will require enabling filtering to clean text for Word weirdness. Current configuration does not clean the text. ie:

```
If youâ€™re lucky, youâ€™ll see ...
```

I'll wait for the decision on the addition of the filters for a later date... if ever.

deezone:
@aaronschachter I'm I good to go on this request?

aaronschachter:
Sure, good to merge :+1: Ideally we should use the functions we're already using in the `dosomething_campaign_preprocess_node` to avoid duplicating logic of if fields are set, but this is fine for now...

PR #2512, opened 2014-06-12
Fix extra header on user profile page.

DFurnes:
This took far too long to figure out...

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/user/page--user--%.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/user/page--user--%.tpl.php
@@ -1,0 +1,22 @@
+<?php
+
+/**
+ * Generates minimal site-wide page chrome.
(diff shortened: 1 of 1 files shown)


aaronschachter:
Not really needed for merge because it's more of a refactory question.. but is the fix here to create a page template without `$variables['header']` ?  Would it not make sense to take the header out of the user profile tpl and rework it to be displayed in the header variable?

aaronschachter:
Brings up questions of where is the right place to put the header.

barryclark:
:+1: 'ing to get this merged as we need to deploy soon. But let's talk more about @aaronschachter's thoughts above and also chat about the way we're implementing this header, etc.

DFurnes:
@aaronschachter Yep, 100%. Just quicker than trying to get that variable accessible outside.

aaronschachter:
Ok, sweet. let's create an issue to not forget about it though.

PR #2514, opened 2014-06-12
Revert previous deletion.

DFurnes:
Fixes Explore Campaigns page by adding back in SCSS removed in earlier PR.


barryclark:
:+1:

PR #2516, opened 2014-06-12
Campaign status solr

### LINKED ISSUE #2377: [Campaign template] Hide campaign from Finder

blisteringherb:
Hiding closed campaigns from the /campaigns page and the finder.  Closed campaigns will still be visible in search.

Fixes #2377


aaronschachter:
sure :+1:

barryclark:
Sweet! Thanks @blisteringherb. Let's merge + test on stage on Monday. :racehorse:

Issue #2518, opened 2014-06-13
Disable, uninstall, and remove DoSomething Legacy module


aaronschachter:
- [x] Disable and uninstall DoSomething legacy on production
- [x] Remove DS Legacy module from repo
- [x] Also remove MOMM variable from settings file.
- [x] Do a search in all DS modules/themes for references to DS Legacy and remove.

PR #2520, opened 2014-06-13
Move MOMM redirect into dosomething_legacy

### LINKED ISSUE #2383: Refactor / remove Pitch custom Entity view mode

aaronschachter:
Moves the redirect to legacy Mind on Money into the Dosomething Legacy module, in anticipation of disabling the module once Mind closes.

Refs #2518

Closes #2383

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -474,27 +474,6 @@ function dosomething_campaign_is_closed($node) {
   return FALSE;
 }
 
-/*
- * Implements hook_entity_view_mode_alter().
- *
- * Sets "pitch" view_mode for campaign nodes based on user signup status.
- *
- * This function also handles redirects to legacy nodes.
- */
-function dosomething_campaign_entity_view_mode_alter(&$view_mode, $context) {
-  // Is this a campaign node?
-  if ($context['entity_type'] == 'node' && $context['entity']->type == 'campaign') {
-    $node = $context['entity'];
-    // Is this a legacy campaign?
-    if (module_exists('dosomething_legacy') && dosomething_legacy_check($node)) {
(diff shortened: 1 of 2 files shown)


angaither:
Might not make sense to change this now, but it doesn't really need to be a `view_mode_alter` hook anymore, we can check in init or node load since it's just a redirect. 
Up to you.
:+1:

PR #2522, opened 2014-06-13
Added an admin form to administer run default copy.

### LINKED ISSUE #2402: [campaign close] "Results coming soon" state
We need to add a state to the campaign close template for those few days after a campaign closes that we don't yet have final signup & reportback numbers. This message will display in place of the results numbers at the top of the template. 

The content editor needs a way to be able to display it instead of the numbers (maybe a checkbox?). For now that message can read "Results coming soon", the editor doesn't need to be able to change the message. @mikefantini is going to finalize the wording. 

cc @weerd

### LINKED ISSUE #2481: [campaign close] - change button text
to read "NOTIFY ME"

angaither:
Values on the form: 
- no winners yet' state
- no total count yet state
- pre-signup button copy 

These values are then passed to the campaign run node template for display. 

Fixes #2481
Fixes #2402

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.admin.inc
+++ b/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.admin.inc
@@ -16,6 +16,7 @@ function dosomething_campaign_run_default_copy_config($form, &$form_state) {
   $form['dosomething_campaign_run_signup_button_copy'] = array(
     '#type' => 'textfield',
     '#title' => t('Signup Button Copy'),
+    '#required' => TRUE,
     '#default_value' => variable_get('dosomething_campaign_run_signup_button_copy'),
   );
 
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.admin.inc
```diff
@@ -0,0 +1,38 @@
...
+* Menu callback for campaign run default text config.
+* @see dosomething_campaign_run_menu().
+*/
+function dosomething_campaign_run_default_copy_config($form, &$form_state) {
+  $form = array();
+
+  // The pre-signup button text.
+  $form['dosomething_campaign_run_signup_button_copy'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Signup Button Copy'),
```
aaronschachter, line 18: These variables should be required.

aaronschachter:
Looks good -- I think we need to set these values in an update hook with the copy in the PR's so that the issues will be fixed upon deploy to stage/prod without us having to update them manually

PR #2529, opened 2014-06-13
Check for logged in user before submitting reportback form

aaronschachter:
We've had a few reportbacks with uid 0, which can be replicated by navigating to a campaign while logged in. From a different tab, log out.  Then go back to the original tab with the campaign and submit the reportback. The reportback is submitted to uid 0.

This fixes that and prevents the submission from happening.

screen shot 2014-06-13 at 1 34 19 pm


angaither:
YAAAAISSS
:+1:

Issue #1746, opened 2014-04-09
Dosomething Web Tests failing


aaronschachter:
Occurring for both the Signup and Reportback web tests.

Complains about not able to find the `field_config` table.

[...]

PR #2534, opened 2014-06-16
Profile/testing cleanup

aaronschachter:
Refs #1746 -- cleanup in attempt to get SimpleTest to run for a DoSomething web test, but does not resolve :(

Removes SimpleTest dependency from DS User to allow for disabling / enabling SimpleTest module. I needed to do this order to get rid of an error about it complaining about not finding a securepages test file.


angaither:
Nevermind, that was green. MONDAY.

looks good :+1:

PR #2547, opened 2014-06-16
Reportback view

### LINKED ISSUE #1560: Reportback view page enhancements
- [x] List each file's fid and link to the original file, along with remote_addr
- [x] Display the reportback's log history

aaronschachter:
Fixes #1560

Renders all reportback data in the reportback entity view for much better debugging.  e.g. File Entity is setting the File's "timestamp" field when it imports it, messing with the actual Reportback view.

reportback-admin


angaither:
:+1:

Issue #2549, opened 2014-06-16
Reportback View tweaks


aaronschachter:
- [x] Fix link to Reportback entity.  It's inserting a space character: http://staging.beta.dosomething.org/reportback/16%20, causing the reportback entity to think it's not in full view mode
- [x] Sort by File FID.  This is nasty but because we've added File Entity late into the game, as File Entity processes the file type of each image we have uploaded, it updates its timestamp value to the date its been processed. For an example, checkout the timestamps on https://cloud.githubusercontent.com/assets/1236811/3292293/9fa92786-f58d-11e3-9cd3-52d6d3bbd2ae.png

PR #2551, opened 2014-06-16
Reportback view fixes

aaronschachter:
- Sort by FID as a failsafe for most recent files - details in #2549
- Fix reportback path


angaither:
:+1:

PR #2553, opened 2014-06-16
Fixing the pitch page content.

weerd:
@DFurnes @barryclark @tjrosario 

Fixes for the pitch page tagline content! Used the `.container` structure approach :)


tjrosario:
+1

PR #2554, opened 2014-06-17
Don't add campaigns to a profile that are closed, with no reportback.

### LINKED ISSUE #2427: [campaign close] - signed up but didn't report back should drop off profile
If a user signed up for a campaign but did not report back before the campaign closed, it should just be removed from the user's profile completely.

cc: @julielorch

angaither:
Fixes #2427


aaronschachter:
:+1:

PR #2555, opened 2014-06-17
Sms gaming

### LINKED ISSUE #2538: [SMS game] - need to display copy indicating # of ppl who have participated in SMS Games
We want to show users how many ppl have participated in SMS games since they launched back in 2012.

Ideally this would be automated so we don't have to update copy on an ongoing basis, but that's not needed just yet if that's too difficult for launch.

@Bladt @joshcusano do you have thoughts on how to show the total number of people who have participated in SMS Games?

image

aaronschachter:
Fixes #2538 

Introduces variables to set copy in the Signup Friends Form on SMS Games, and outputs them within the form.

sms-game-admin
.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
@@ -5,6 +5,28 @@
  */
 
 /**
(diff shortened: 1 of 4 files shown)


angaither:
:+1:

PR #2559, opened 2014-06-17
Updating DO IT field feature to use markdown based on Sharon Actor's exc...

### LINKED ISSUE #2474: [Campaign] Allow markdown in "After Copy" field
To be more consistent with other fields in the "Do It" section of the campaign node, add the capacity for using markdown in the "After Copy" text area field to match the other text area items in that fieldset. Also will help with consistent tag output in the template.

weerd:
...ellent instructions.

Resolves #2474

@aaronschachter


aaronschachter:
:+1: who is the real sharon

weerd:
That's some deep, contemplative stuff. :goat:

Issue #2491, opened 2014-06-11
[campaign close] - update LOVE FROM CELEBS section header copy


mikefantini:
update section header "LOVE FROM CELEBS" to read "THE BUZZ"

![image](https://cloud.githubusercontent.com/assets/6330971/3248021/4a03e098-f18a-11e3-9508-26c2bc3df3d2.png)

@barryclark not sure if it's bender or fender so just assigning to you

barryclark:
One for tom, thanks

Issue #2490, opened 2014-06-11
[campaign close] - rules for displaying info in winner section


mikefantini:
- [x] if anything in the field_winners group is filled out display in the "CONGRATULATIONS TO..." section
- [x] if nothing in the field_winners group is filled out display the section header ("CONGRATULATIONS TO...") and the default winner text (#2402)

Mobile mocks with (outdated) copy:
![image](https://cloud.githubusercontent.com/assets/6330971/3247957/a9d8331c-f189-11e3-9bb9-a89cf06f76ef.png)

barryclark:
@crystalruthbell 

Are we moving to using italics for all placeholder type messaging? Should this become a pattern?

cc @weerd @DFurnes @tjrosario

tjrosario:
I've made this it's own pattern specific to campaign closed for now, but we should consider porting it over to neue.  I left a todo comment in the CSS just as a reminder.

barryclark:
Awesome, @tjrosario. Thanks.

PR #2565, opened 2014-06-17
Campaign closed mods

### LINKED ISSUE #2492: [campaign close] - don't display THE BUZZ section header if no content exists in that section
If nothing is filled out in any of the fields in group_the_buzz don't display "THE BUZZ" section header.

tjrosario:
closes #2492 #2491 #2490

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module
+++ b/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module
@@ -75,7 +75,8 @@ function dosomething_campaign_run_preprocess_text_vars(&$vars, $wrapper) {
     $placeholder_copy = variable_get('dosomething_campaign_run_no_total_copy');
     // Replace the ['label'] section with the actual label (eg clothes or tacos).
     $copy = str_replace('[label]', $vars['total_quantity_label'], $placeholder_copy);
-    $vars['total_quantity'] = $copy;
+    //$vars['total_quantity'] = $copy;
+    $vars['total_quantity_placeholder'] = $copy;
   }
 }
 
@@ -93,7 +94,7 @@ function dosomething_campaign_run_preprocess_winners(&$vars, $nid) {
   $winner_count =  count($closed_node->field_winners[LANGUAGE_NONE]);
   // If there are no winners, send default text.
   if ($winner_count == 0) {
-    $winners = variable_get('dosomething_campaign_run_no_winner_copy');
+    $vars['default_winners'] = variable_get('dosomething_campaign_run_no_winner_copy');
   }
   else {
     $winners = array();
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module
```diff
@@ -75,7 +75,8 @@ function dosomething_campaign_run_preprocess_text_vars(&$vars, $wrapper) {
     $placeholder_copy = variable_get('dosomething_campaign_run_no_total_copy');
     // Replace the ['label'] section with the actual label (eg clothes or tacos).
     $copy = str_replace('[label]', $vars['total_quantity_label'], $placeholder_copy);
-    $vars['total_quantity'] = $copy;
+    //$vars['total_quantity'] = $copy;
```
aaronschachter: Is this supposed to be commented?  Let's remove it if we don't need it anymore...

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign--closed.tpl.php
```diff
@@ -124,68 +129,91 @@
...
-    <h2 class="container__title banner"><span>Love From Celebs</span></h2>
-    <div class="wrapper">
 
-      <div class="container__body">
+  <?php
+    if (isset($additional_text_title) OR
+        isset($additional_text) OR
+        (isset($psa) AND $psa !== '') OR
+        (isset($klout_gallery) AND !empty($klout_gallery))
+        ):
```
aaronschachter: It would be a lot cleaner to create a new variable within the prepreocess function, something like `$display_celebs_section` which is set to this boolean expression.  Then we can just check for `if ($display_celebs_section)` here

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign--closed.tpl.php
```diff
@@ -124,68 +129,91 @@
...
+          <?php foreach ($klout_gallery as $key => $klout_gallery_item) :?>
+
+            <?php // Assign specific gallery css classes depending on 
+                  // the gallery type
+                  // If the gallery type is 'mention' then it will use the -duo pattern, 
+                  // else use -triad
+            ?>
+
+            <?php
+              switch ($klout_gallery_item['type']) {
```
aaronschachter: This should all be in preprocess too.

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign--closed.tpl.php
tjrosario: Agreed.  Will move the conditional to preprocessing

aaronschachter:
Looks ok here :+1:

PR #2567, opened 2014-06-17
Field Collection s. show

aaronschachter:
This is not ready for merge but opened for visibility.  I NEED HALP

Tested a build with latest dev version of Field Collection, as it claims to fix https://drupal.org/node/2141781

The build completes and the field collection updates run. From drush I can revert features and clear cache.

~~However the entire site dies with:
Fatal error: Call to undefined function unicode_check() in /var/www/vagrant/html/includes/common.inc on line 5130~~

### DIFF
--- a/lib/profiles/dosomething/drupal-org.make
+++ b/lib/profiles/dosomething/drupal-org.make
@@ -56,7 +56,10 @@ projects[features][version] = "2.0"
 projects[features][subdir] = "contrib"
 
 ; Field Collection
-projects[field_collection][version] = "1.0-beta7"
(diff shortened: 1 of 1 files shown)


aaronschachter:
The `Fatal error: unicode_check` seems to have been a weird Vagrant thing where something was dying. Upon a vagrant reload, pulling from stage and running a build with this branch, I am able to revert features.

desmondmorris:
Per a convo with @aaronschachter, I think this approach will be fine.  We just must keep an eye for a proper release with that commit in tact

PR #2573, opened 2014-06-17
added winner subheaders

### LINKED ISSUE #2568: [campaign close] - winner section subheader, not pulling in type of winner
Need to pull in type of winner and affix "Winner" to the end of it like we have in the mocks:
image

So we should pull in field_winner_type and then affix "Winner" to it, so if I select Scholarship as the field_winner_type, the page would read "Scholarship Winner"

Currently looks like:
image

cc: @barryclark @angaither

tjrosario:
Fixes #2568


aaronschachter:
:+1:

PR #2575, opened 2014-06-17
Added campaign_run to staff only content type array.

### LINKED ISSUE #2560: [campaign close] - prevent auth/anon users from visiting the campaign run node directly
since we have to publish campaign run nodes, let's add a check that prevents auth/anon users from visiting the campaign run node directly, like how we do for Facts or Image nodes

@barryclark assigning to you to assign out

angaither:
Fixes #2560


aaronschachter:
nailed it :+1:

PR #2576, opened 2014-06-17
User presignup tweaks

aaronschachter:
@angaither Please review.
- Sets the presignup confirm message upon user login/presignup.
- Hides presignup form if you're logged in and presigned up.
- Displays "We'll hit you up" text instead of the scholarship amount in the Presignup Form Callout, stores copy in a variable


angaither:
:+1:

PR #2588, opened 2014-06-18
Filter current campaign from confirmation results

aaronschachter:

Because an anonymous user can submit the Friends form for SMS Game nid X, the confirmation page could potentially display the campaign X as a recommended campaign since there is no signup to filter by.

This PR makes changes to the recommendation page callback to filter out the current campaign NID from results.


angaither:
:+1:

PR #2589, opened 2014-06-18
Refactoring Capistrano to work with multistage international

blisteringherb:
#### What's this PR do?
- Adding an include file to accommodate international staged deployments.
- Refactoring non-international stages to use the same shared_links code
- Adding aliases to support international prod pulls
- Refactoring and adding policies to drush to prevent issues with using international prod aliases

The biggest changes are in the deploy.rb file.  All of the code that was in the individual stage deploy files was abstracted here to one shared_links task

This branch in this repo will have to be added to a new Jenkins job for testing.

Because the international deployment is a multisite and multistage installation it requires more than just the one 'international' stage for Capistrano.  It seemed prudent while refactoring that to also refactor the other stage files and tasks.

@desmondmorris 
@mshmsh5000

### DIFF
--- a/config/deploy/includes/international.rb
+++ b/config/deploy/includes/international.rb
@@ -1,6 +1,6 @@
 set :deploy_to, "/var/www/international.dosomething.org"
 
-case :stage
+case ENV['stage']
 when production
   role :app, %w{dosomething@dosomething@deploy.international.dosomething.org}
   server 'deploy.international.dosomething.org', user: 'dosomething', roles: %w{app}
--- a/config/deploy/includes/international.rb
+++ b/config/deploy/includes/international.rb
@@ -6,6 +6,6 @@ when production
   server 'deploy.international.dosomething.org', user: 'dosomething', roles: %w{app}
 
 when qa
-  role :app, %w{dosomething@10.241.0.33}
-  server '10.241.0.33', user: 'dosomething', roles: %w{app}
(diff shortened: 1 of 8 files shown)


desmondmorris:
This looks great.  +1

This should not break the jenkins deploy jobs should it?

blisteringherb:
This shouldn't break the Jenkins deploy jobs.  

They'll need an additional parameter, stage, and to be modified slightly to make the international deployments work, but the current jobs as they are should continue to just work.  George might also need some modifications.

```
#!/bin/bash -e
export TERM=xterm
[[ -s "$HOME/.rvm/scripts/rvm" ]] && source "$HOME/.rvm/scripts/rvm"
rvm use 2.0
eval `ssh-agent -s` 
ssh-add

if [[ -z "$TAG" ]]
then
  cap $TIER deploy
elif [ -n "$STAGE" ] && [ -n "$TAG"]
then
  cap $TIER deploy branch=$TAG stage=$STAGE
else
  cap $TIER deploy branch=$TAG
fi
```

config/deploy/includes/international.rb
```diff
@@ -0,0 +1,11 @@
+set :deploy_to, "/var/www/international.dosomething.org"
+
+case ENV['stage']
+when production
+  role :app, %w{dosomething@dosomething@deploy.international.dosomething.org}
```
mshmsh5000: Redundant `dosomething@`?

PR #2590, opened 2014-06-18
Tablet mods

tjrosario:
Fixes totals column collision at tablet


lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign-closed.scss
```diff
@@ -34,11 +34,28 @@
       font-size: 1.2rem;
     }
 
+    @include media($desktop) {
+      strong {
+        font-size: 5rem;
```
weerd: Try to see if you can use one of the predefined $font-"insert-size-here" predefined variables from Neue so it'll match one of the sizes in the style guide if possible.

weerd:
:+1:

PR #2591, opened 2014-06-18
SMS Confirmation Page: Login/register modal

### LINKED ISSUE #2541: [SMS game] - prompt user to create account/log in on confirmation page
For SMS Games, we should prompt the user to create an account or log in after they have submitted the form.

Let's take the real estate we give to "Find More Campaigns" and replace it with this call out.

Clicking on the button should prompt auth modal.

Ideally we would have CTA copy above the button like we do on static content.

@aaronschachter let me know how you want to handle this copy... I can either get copy to you in this issue or if you're going to build that thing for admin's to control copy I'll just put it in there.

image

aaronschachter:
@angaither Please review

Fixes #2541

Sets a bunch of variables to prevent needing to make copy changes via code.


angaither:
:+1:

PR #2595, opened 2014-06-18
removed capitalization from winners text and fixed vertical centering on tablet

tjrosario:
Removes auto caps on winners text and vertical alignment of the content of each column


weerd:
:+1:

Issue #2596, opened 2014-06-18
[campaign close] - create new image crop


mikefantini:
now that we have the media style gallery with thumb nail images I think 640x640 is too restrictive for a minimum image size for photos.

Can we reduce it to 300x300 for now?

We can always change it back if we end up implementing #2380 and don't think this will disrupt other times we try to go with this two column gallery.

DFurnes:
Where are we getting these 300x300 photos from? That's very low resolution.

mikefantini:
@DFurnes some of them are coming from twitter.. we'll be chatting about this post-stand up today if you want to join

mikefantini:
updated the title

let's call this "thumbnail"

minimum size is 200x200

angaither:
Closed with #2619

PR #2605, opened 2014-06-19
Added a `field_image_thumb` type.

angaither:
- added new image field type called `thumbnail` to be used for very small icons/logos/avitars
- updated the klout gallery images to pull in thumb type on `press` and `mention` galleries. 
- using `100x100` size for now
  Refs #2596


aaronschachter:
:+1:

PR #2608, opened 2014-06-19
Added FC dev updates to features to fix overriden things.

angaither:
In order to get features out of a perpetually overridden state the code needs to match what's in the DB. 
This code wasn't added after the updates were made by features collection.


aaronschachter:
Nice catch :+1: Pulled down and tested

PR #2619, opened 2014-06-19
Update image styles again

angaither:
Created a new `100x100` image style
Updated `hmax-75` to match other patterns
Sent press images as 75px max, and celeb mentions as a little large & square. 

Refs #2596


aaronschachter:
:+1:

PR #2624, opened 2014-06-20
Attempt to disbale autocomplete.

### LINKED ISSUE #2460: [SMS game] - remove auto-filling in phone number info
When a user goes to fill in a phone number, the browser is suggesting data to auto-fill in the field (i.e. if I used 555-555-5555 before in my browser as a phone number, it suggests I use that for my phone number AND all three of the friends' phone number fields).

Let's just remove all phone number suggestions if we can for all of those field's (user's phone number included).

Making high priority because it creates a pretty terrible experience for the user because you have to delete data in the fields every time you try to fill out the form.

weerd:
Not 100% this will work to eliminate the autofilling of the multiple phone number fields we saw during user testing. This approach could work, but it's dependent on the browser actually paying attention to this attribute on the `<input>`. 

This seems to be a very browser specific and OS specific issue which is harder to address universally.

Resolves #2460

@DFurnes @aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
@@ -129,6 +129,7 @@ function dosomething_signup_friends_form($form, &$form_state, $node, $num_friend
       'data-validate' => 'phone',
       'data-validate-required' => '',
       'placeholder' => t('(555) 555-5555'),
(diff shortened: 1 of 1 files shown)


DFurnes:
:checkered_flag:

aaronschachter:
Sure :8ball: :+1:

weerd:
huzzah! let's give it a whirl.

PR #2629, opened 2014-06-20
11 facts  conversion point

### LINKED ISSUE #2451: 11-facts - add additional conversion point
Holla,

11-facts testing has revealed that a SECOND button between facts 5 and 6 increases conversions.

Button will: Look exactly like the existing button (this snippit doesn't include gray background b/c it was from optimlizely), and it will have THE SAME CTA as the lower button (the CTA is editable in Admin view).

screen shot 2014-06-09 at 11 25 19 am

weerd:
Resolves #2451

@DFurnes  Don't look! It's hideous!


DFurnes:
:see_no_evil:

PR #2633, opened 2014-06-20
Kills DoSomething Legacy module

aaronschachter:
Now that Mind on My Money is closed, we're done with redirecting to legacy.


desmondmorris:
best pr ever

PR #2638, opened 2014-06-23
Refactor get user field

### LINKED ISSUE #2606: User: Get First Name function
Add a `dosomething_user_get_first_name` function which uses `dosomething_user_get_field`, but also does string processing to make sure it's Capital letter, lowercase.
- no all lowercase
- no all capitals

We currently have users with lowercase / all caps first names, so just implementing this via `hook_user_save` will not fix for existing users.

angaither:
## Overview

Refactoring calls to `dosomething_user_get_field`
## What does this PR do?
- Creates a generic `dosomething_user_get_field` function 
  - This function calls handlers if more specific formatting is needed
  - The handlers should not be called directly
## How to test
- Check out this PR
- Test out the user profile page for any oddities
- Manually `drush php-eval` functions to make sure the results are as expected, a few to test...
  - `drush php-eval "var_dump(dosomething_signup_get_mbp_params(user_load(10), node_load(84)));"`
  -  `drush php-eval "echo dosomething_user_is_under_thirteen(user_load(10))"`
  - `drush php-eval "echo dosomething_user_get_field('field_first_name', user_load(10), 'ucwords');"`

Bonus! Fixes #2606

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module
+++ b/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module
@@ -111,7 +111,7 @@ function dosomething_campaign_run_preprocess_winners(&$vars, $nid) {
       $winners[$i]['uid'] = $winner[$entity_id]->field_user[LANGUAGE_NONE][0]['target_id'];
       // If there is no user, get the first name from the text field.
       if (!isset($winners[$i]['uid'])) {
-        $winners[$i]['fname'] = $winner[$entity_id]->field_winner_name[LANGUAGE_NONE][0]['value'];
+        $winners[$i]['fname'] = dosomething_user_get_field('field_first_name', user_load($winners[$i]['uid'], 'ucwords'));
       }
       // Else, grab the first name from the user account.
       else {
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -349,7 +349,7 @@ function dosomething_user_new_user($form, &$form_state) {
       'email' => $account->mail,
       'uid' => $account->uid,
       'first_name' => dosomething_user_get_field('field_first_name', $account),
-      'birthdate' => dosomething_user_get_birthdate(),
+      'birthdate' => dosomething_user_get_field('field_birthdate', $account),
     );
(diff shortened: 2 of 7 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -635,51 +635,103 @@ function dosomething_user_create_user_by_mobile($number) {
...
+/**
+ * Helper function to format strings.
+ *
+ * @param string $string
+ *  The string to format
+ * @param string $format
+ *  The format that should be returned.
+ *    strtolower, strtoupper, ucwords
+ */
+function dosomething_user_format_string($string, $format) {
```
aaronschachter, line 705: Minor but I could see us moving this into `dosomething_helpers` to potentially use in other modules.

lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module
```diff
@@ -111,7 +111,7 @@ function dosomething_campaign_run_preprocess_winners(&$vars, $nid) {
       $winners[$i]['uid'] = $winner[$entity_id]->field_user[LANGUAGE_NONE][0]['target_id'];
       // If there is no user, get the first name from the text field.
       if (!isset($winners[$i]['uid'])) {
-        $winners[$i]['fname'] = $winner[$entity_id]->field_winner_name[LANGUAGE_NONE][0]['value'];
+        $winners[$i]['fname'] = dosomething_user_get_field('field_first_name', user_load($winners[$i]['uid'], 'ucwords'));
```
aaronschachter, line 114: I've pulled down this branch to test... I am not seeing the Winner First names here. Is it working for you? 

no-names

aaronschachter:
Code looks ok - just seemed to find the one bug in the Campaign Close template. Another place we could use this is in the User Signup Form, where we collect addresses, but this is also minor and not needed for merge..

lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module
angaither, line 114: Alright, check again. It wasn't respecting the default values set in handler functions.

aaronschachter:
:+1:

PR #2641, opened 2014-06-23
Use the field_first_name token instead of $account->name

### LINKED ISSUE #2502: User Profile PageTitle
Implement `drupal_set_title` to override the default behavior, which uses the user's username as the page title. Which in our case, is set to the uid so it makes no sense.

Maybe just have it read as "My Profile"? "My campaigns?" 

Screenshot shows bug in user profile, but note the `<TITLE>` value.
screen shot 2014-06-12 at 2 34 44 pm

angaither:
Fixes #2502

### DIFF
--- a/lib/modules/dosomething/dosomething_metatag/dosomething_metatag.features.metatag.inc
+++ b/lib/modules/dosomething/dosomething_metatag/dosomething_metatag.features.metatag.inc
@@ -118,7 +118,6 @@ function dosomething_metatag_metatag_export_default() {
   // Exported Metatag config instance: user.
   $config['user'] = array(
     'instance' => 'user',
-    'disabled' => TRUE,
     'config' => array(
       'title' => array(
         'value' => '[user:field_first_name] | [site:name]',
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_metatag/dosomething_metatag.features.metatag.inc
```diff
@@ -121,7 +121,7 @@ function dosomething_metatag_metatag_export_default() {
     'disabled' => TRUE,
```
aaronschachter: That's odd that this fixes... the metatags is set to disabled so that it doesn't display that "metatags" fieldset on the user profile, or add the relevant metatags into the html...

aaronschachter:
Pulled down and tested, reverting all features. Did not fix.

screen shot 2014-06-23 at 4 07 55 pm

lib/modules/dosomething/dosomething_metatag/dosomething_metatag.features.metatag.inc
angaither: Hm weird features should have removed that when I updated, strangeeeeeness. Anyways, it's updated now.

aaronschachter:
:+1:

PR #2647, opened 2014-06-23
Cross browser mods

tjrosario:
Adds fallback for nth-child hooks in IE8

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -382,6 +382,7 @@ function dosomething_campaign_preprocess_reportback_gallery(&$vars, $style = '30
   // Loop through gallery_vars to only output what we need for theming:
   foreach ($items as $delta => $item) {
     $vars['reportback_gallery'][$delta]['image'] = $item['image'];
+    $vars['reportback_gallery'][$delta]['order_class'] = dosomething_helpers_get_gallery_item_order_class($delta);
     if (isset($item['caption'])) {
       $vars['reportback_gallery'][$delta]['caption'] = $item['caption']['safe_value'];
     }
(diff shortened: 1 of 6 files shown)


DFurnes:
Looks good. Does `dosomething_helpers_get_gallery_item_order_class` belong in the theme though, since it's only view related?

weerd:
Cool. Now I understand what you meant earlier :)

tjrosario:
@DFurnes it was moved out of theme into helpers as it is used by static_content and campaign_run now.

weerd:
:+1:

DFurnes:
Even if it's used by multiple modules, it's a theme-related concern. It only makes sense in the context of the Paraneue theme and the markup we're using there. Modules should be holding business logic - if you look at the majority of the other functions in the `dosomething_helpers.module` file, that's what they are.

It's not a blocker, just something worth considering and potentially re-organizing later.

tjrosario:
@DFurnes Noted. Is there a best practice you are using for shared theme-specific methods?

DFurnes:
It looks like we've been putting things like this in [dosomething_helpers.theme.inc](https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc) so far. I'd argue that [helpers.inc](https://github.com/DoSomething/dosomething/blob/dev/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc) in the theme would be the best place for template helper functions like this since they're being used to generate theme-specific markup.

Definitely willing to hear other arguments though... I'm just worried that `dosomething_helpers` seems to be becoming a dumping ground.

/cc @aaronschachter @angaither Any thoughts?

aaronschachter:
I agree on all points, @DFurnes . Moving this function in particular into a theme `helpers.inc` sounds good here.

DFurnes:
Cool, made a low priority issue so we don't forget about it. :thumbsup:

tjrosario:
Great!

PR #2650, opened 2014-06-23
Update to grouped campaign template to add in updated callout.

weerd:
I noticed that the callout on the Grouped Campaign page lost it's styling due to the updates earlier today. I had missed updating this template along with the others.

@DFurnes @tjrosario

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
@@ -27,6 +27,7 @@
 }
 
 .campaign--pitch,
+.campaign--grouped,
 .campaign--action .container--prove {
   .message-callout {
     .__copy {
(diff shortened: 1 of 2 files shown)


DFurnes:
LOOKS SOLID BRAH

lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
```diff
@@ -35,9 +36,8 @@
   }
 }
 
-
-// Campaign Pitch Page
-.campaign--pitch {
+.campaign--pitch,
+.campaign--grouped {
```
tjrosario, line 40: Do you foresee adding more classes to this definition?  Would be nice to streamline it with a single class if it starts to bloat up.

weerd, line 40: Agreed. We'll see what happens in the near future. If more need to be added, we should probs. streamline it. But good mentioning it so it's on our radar :)

PR #2659, opened 2014-06-24
Make all beta numbers required

### LINKED ISSUE #2654: [SMS game] - form submits with invalid number
image

I also get validation on the form itself (i.e. field shakes and error message appears above the form) before the form submits.

Form only submits upon the first time you try to submit. Otherwise you just get the error on the form itself.

Ideally the form just wouldn't submit at all if there was an error.

Happened on IE11 and chrome on windows 7

Also some inconsistent behavior. Sometimes page reloads, sometimes it takes you to same URL but /campaigns appears on the page (see: #2653)

cc: @aaronschachter

aaronschachter:
@weerd Please review.

Fixes #2654 also.


blisteringherb:
:+1:

weerd:
Looks good to me :)

PR #2662, opened 2014-06-24
Support iPhone copy/paste contact numbers

aaronschachter:
@angaither Can you review?

When you copy/paste from iPhones, it adds a leading 1. This PR adds a condition into `dosomething_user_clean_mobile_number` to account for this.

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -599,6 +599,12 @@ function dosomething_user_clean_mobile_number($number) {
     if (strlen($trimmed_number) == 10) {
       return $trimmed_number;
     }
+    // Else if it contains a leading 1.
+    // (iPhones return the 1 when you copy/paste a number from your contacts).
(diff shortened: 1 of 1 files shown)


angaither:
I think `dosomething_user_valid_mobile` should take care of the leading 1 issue.
https://github.com/dosomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_user/dosomething_user.module#L83

angaither:
If that's not working though, this code is ok :+1:

aaronschachter:
Yeah, `dosomething_user_valid_mobile` is ok with leading 1's, but it returns a boolean: https://github.com/dosomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_user/dosomething_user.module#L110

The validation is failing within `dosomething_user_clean_mobile_number` because it expects 10 digits: https://github.com/dosomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_user/dosomething_user.module#L599

May be worth refactoring to combine `dosomething_user_valid_mobile` and `dosomething_user_clean_mobile_number` functions.

PR #2667, opened 2014-06-24
Updaing campaign modal lists and sources.

### LINKED ISSUE #2643: SMS Template: Sources in modal formatting is off
screen shot 2014-06-23 at 3 24 48 pm

Can we make these numbers superscript and in line that way the line spacing is fixed and readability increases?

weerd:
Resolves #2643

@DFurnes @tjrosario

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/campaign/sources.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/campaign/sources.js
@@ -2,11 +2,12 @@ define(function(require) {
   "use strict";
 
   var $ = require("jquery");
+  var Events = require("neue/events");
+  var Modal= require("neue/modal");
 
-  var $sources = $(".sources") || null;
 
-  if ($sources) {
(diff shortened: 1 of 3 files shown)


lib/themes/dosomething/paraneue_dosomething/js/campaign/sources.js
```diff
@@ -2,11 +2,12 @@ define(function(require) {
   "use strict";
 
   var $ = require("jquery");
+  var Events = require("neue/events");
+  var Modal= require("neue/modal");
```
DFurnes: I don't think you need to include Modal here, right?

DFurnes:
Looks good other than the one unused variable.

lib/themes/dosomething/paraneue_dosomething/js/campaign/sources.js
weerd: Cool. I thought I did, but wasn't sure. I'll remove and confirm still works, and then merge!

PR #2668, opened 2014-06-24
Set default=NULL in user_birthdate function

### LINKED ISSUE #2666: [Thumb Wars] Sign In Error Message.
I found this weird bug using chrome where I would search for the Thumb Wars campaign and hit the sign up button, sign into my account and then end up with this error message: 
image
***Note: I can still sign into my account, but that error message still stays on the site.

angaither:
Fixes #2666


aaronschachter:
:+1:

Issue #2642, opened 2014-06-23
SMS Template: Kick down help text to a darker gray


crystalruthbell:
![screen shot 2014-06-23 at 3 21 36 pm](https://cloud.githubusercontent.com/assets/6550604/3362624/c23608cc-fb0b-11e3-8831-09b73a252641.png)

Should increase readability to level to a darker gray.

mikefantini:
@weerd assigning over to you. not sure how this will change if we ever decide to use a different background image.

I'll let you and @crystalruthbell talk through this.

weerd:
I'm going to bump it to use a sass function to darken it, but we should discuss with @DFurnes cause we're falling out of the set of gray variables we established in Neue. We could add more, or it might be a good time to use Sass Lists for all the gray variables... ;)

weerd:
In regards to the background image dependency, we don't have an alternative right now. Maybe some functionality to consider to toggle between dark and light text design depending on the perceived "darkness or lightness" of an image. For the time being we'll need to aim to use a lighter colored image on SMS.

PR #2669, opened 2014-06-24
Sms  disclaimer color

weerd:
...n template.

Resovles #2642 

@DFurnes don't be mad!


DFurnes:
Why not use `$dark-gray`?

weerd:
Waaay to dark. Basically same color as regular text. I tried it.

DFurnes:
Gotcha. Merge away. Let's definitely take some time to think about this in the future though. :sailboat:

weerd:
Indeed! :dancer:

PR #2672, opened 2014-06-24
Reportback Flag permissions

aaronschachter:
Missed these in building out all the flag stuff.  I've manually added them to prod to allow campaigns team to start flagging.


angaither:
:+1:

PR #2679, opened 2014-06-25
Update MBP admin path to admin/config/dosomething

aaronschachter:
Cleanup task to move all of our admin config pages into a single folder.


angaither:
Code is fine, but I think there is a use case for some of this stuff living under the web services menu

UPS Api keys
Zend desk keys
etc etc

Those things aren't really `dosomething` configs

aaronschachter:
Hmm, I'd argue that they are, because they are custom DS modules, hence putting it all in one folder.  Both of those instances you've mentioned have been moved to that `admin/config/dosomething` menu already.  Going to merge this for now but happy to talk through best places for each.

PR #2680, opened 2014-06-25
Fact page intro

aaronschachter:
- Changes `field_intro` to use markdown, displays in the template.
- Removes the `field_additional_text` field instance. This was used for including Sources as per the legacy 11 fact pages, but we are not using this field anymore and instead displaying the Sources linked to Facts.


angaither:
:+1:

Issue #2531, opened 2014-06-16
User URL pattern


aaronschachter:
We're currently storing 1571614 unnecessary aliases for user profiles.

![url-aliases](https://cloud.githubusercontent.com/assets/1236811/3287935/0535410a-f55e-11e3-918a-82c40f69a970.png)
- [x] Set user pathauto pattern to the default URI: `user/[user:uid]`. This will prevent alias records from being created
- [ ] Delete all user aliases in prod db.

aaronschachter:
Reopening - will close once user aliases are deleted on prod.

aaronschachter:
Aww yeah.

![users-aliases](https://cloud.githubusercontent.com/assets/1236811/3404536/1d86a2e6-fd72-11e3-9771-af011f3c452c.png)

PR #2682, opened 2014-06-25
Remove unnecessary pathauto defaults 

aaronschachter:
Sets pathauto patterns to empty strings for
- Users
- Taxonomy terms
- Files (this was already set to the default, but setting to empty string to keep consistent)

This prevents creating unneded records in the `url_alias` table.  Example: we have a million records in there for the user alias for `users/[uid]` , where the default path is just `user/[uid]`

Once this is merged in, we'll manually delete the aliases for users and taxonomy terms.  See #2531


angaither:
:+1:

PR #2683, opened 2014-06-25
Remove user_signups view (no longer used)

aaronschachter:
- `user_signups` was once used to display user signups on the profile. It is no longer used.
- Update `node_signup` view to display for Grouped Campaigns, as well as Campaigns.


angaither:
:+1:

PR #2687, opened 2014-06-26
Move Submit Campaign form into a block

### LINKED ISSUE #2685: [member created campaigns] - unpublish the form for now
Need to unpublish https://www.dosomething.org/volunteer/hunt/campaign-form

We'll probably adjust some of the questions and re-use soon, but need to take the form down for now.

aaronschachter:
Fixes #2685 by removing the page calback to the Submit Campaign form.  

Moves the form into a custom block, to be embedded into Static Content for when we need it again.  I've tested creating a static content page and embedding the block, seems to look and work fine.

static-content-view
static-content-edit


angaither:
:+1: 
Do we need to make this node on stage/production?

aaronschachter:
Yep, I was going to do that once we deploy to each.

aaronschachter:
Confirmed with Alyssa that the test submission I sent from the Static Content implementation was received.

PR #2688, opened 2014-06-26
Reportback / signup forms housekeeping

aaronschachter:
Closes an old issue of moving all the reportback / signup forms into relevant `forms.inc` files.  

This keeps the `.module` files a little bit cleaner, with basic entity CRUD functions and relevant Drupal hooks.


angaither:
:+1: nice cleanup. value added.

PR #2690, opened 2014-06-26
A campaign class

aaronschachter:
This PR is the start of an idea I've discussed with a few devs in our team: a Campaign class to provide a simplified version of a loaded campaign node.

The idea here is moving all the preprocess / entity reference logic all into one class constructor, so there are standard variables we can use site wide, vs always needing to drill down into field arrays and loading reference nodes.  This Campaign class would be the object would be returned in a Campaign Content API GET request (vs all of the crazy field array stuff a Services node GET request would return).

I've made changes in two places to demonstrate how it could possibly simplify things: the campaign preprocess for Fact Problem and Fact Solution, and the message broker campaign cache request for Fact Problem and Fact Solution.

If we're on board with this approach, we can keep working on the class, and standardize the Campaign variable getting stuff site-wide.  This PR mainly for visibility and to open to discussion on pros and cons of this approach.. we can slowly migrate different fields into the campaign object, vs. needing to do all at once.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -44,18 +44,22 @@ function dosomething_campaign_load($node) {
   foreach ($filtered_text as $property) {
     $field_name = 'field_' . $property;
     if ($value = $wrapper->{$field_name}->value()) {
-      // Store raw input (without markup).
-      $campaign->{$property} = $value['value'];
+      // Store filtered text.
+      $campaign->{$property} = $value['safe_value'];
     }
   }
 
   // Taxonomy fields.
+  $campaign->primary_cause = NULL;
   if ($primary_cause = $wrapper->field_primary_cause->value()) {
     $campaign->primary_cause['tid'] = $primary_cause->tid;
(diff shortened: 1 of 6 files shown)


aaronschachter:
Screenshot of the object so far:
campaign-load

angaither:
This is looking awesome & it so much cleaner!

DFurnes:
Oh my god, this is amazing! :tada:

deezone:
Looks great but missing the high / low season start / end dates **or** the creation date, whichever is easier. Can I open a new issue for these items?

aaronschachter:
Yup, its not complete yet. I'll be adding more properties in today, aiming to use the Campaign class for all of the variables being set in `dosomething_mbp_campaign_request`.

deezone:
Please let me know when the end-point to access this functionality is available. As discussed, I'd like to switch out the use of the use of mb-campaign-api with this solution for the digest functionality in the Message Broker system.

aaronschachter:
I've created issues #2711 for setting up the endpoint, and #2712 to remove the related mb-campaign-api code once the switch has been made.

deezone:
Sounds good. Once the end point is operational I'll remove all of the mb-campaign-api code. This is going to be great, thanks for making this happen!

desmondmorris:
TYBG

PR #2691, opened 2014-06-26
Added password reset customization.

### LINKED ISSUE #1004: pw reset - remove username in field title
Field name should just read "E-mail address" not "Username or e-mail address" since most users don't know their username or that they even have one and it will just create confusion.

image

### LINKED ISSUE #1005: pw reset - notification that email was sent is needed
When I request a new password on /user/password I should receive a notification "An email has been sent to [user_email] to reset your password."

angaither:
Fixes #1004
Fixes #1005


aaronschachter:
Code looks fine. Just curious -- why theme and not `dosomething_user` ? this would be something we'd need regardless of theme, no?

angaither:
All the user form customizations are already in the theme. idk I agree, but I just put it were it with everything else.

aaronschachter:
Ok.. fine for now to keep it all in the same place, but I think it's worth a refactor to move in the module at some point.

:+1: ship it

barryclark:
Thanks @angaither!

PR #2696, opened 2014-06-26
Rem no more

DFurnes:
Replace `rem` units with `pixels`, removing our need for the polyfill on IE8. References DoSomething/neue#309.


weerd:
Oh man

DFurnes:
Merging this guy in so it doesn't get out of sync.

PR #2700, opened 2014-06-27
Updates to preprocess to fix header in default pages.

weerd:
@DFurnes @angaither 

Everything worked great except on default pages where the header disappears since there is no node output.

Aiming to find a different solution to be able to output the header in the static node tpl. This works for now.


lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
```diff
@@ -47,6 +47,25 @@ function paraneue_dosomething_preprocess_page(&$vars) {
   $vars['navigation'] = theme('navigation', $navigation_vars);
 
 
+  krumo($vars);
+  krumo($vars['node']);
```
aaronschachter: no krumo

weerd:
@aaronschachter fixed :)

DFurnes:
:sailboat:!

PR #2704, opened 2014-06-27
refactored fact listing to return chunked arrays to play nicely with cta's

### LINKED ISSUE #2703: Refactor facts listing to return chunked arrays
The cta that is placed in the middle of fact's needs to span full width so we need to split the array of facts into chunks so we have more control over markup

tjrosario:
Fixes #2703

### DIFF
--- a/lib/modules/dosomething/dosomething_fact_page/dosomething_fact_page.module
+++ b/lib/modules/dosomething/dosomething_fact_page/dosomething_fact_page.module
@@ -98,6 +98,7 @@ function dosomething_fact_page_preprocess_node(&$vars) {
     $node = entity_metadata_wrapper('node', $vars['node']);
     $values = dosomething_fact_get_fact_field_vars($node->field_facts);
     $vars['facts'] = $values['facts'];
+    $vars['facts_chunked'] = paraneue_dosomething_chunk_array($vars['facts'], 5);
     $vars['sources'] = $values['sources'];
   }
 }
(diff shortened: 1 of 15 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -29,3 +29,19 @@ function paraneue_dosomething_get_gallery_item_order_class($index) {
...
+ * @param array $array
+ *   An array to be chunked
+ *
+ * @param int $size
+ *   Specifies size of each chunk
+ *
+ * @return array
+ *   Array of array chunks
+ */
+function paraneue_dosomething_chunk_array($array, $size) {
```
DFurnes: Is there a reason not to just use `array_chunk` directly here?

lib/modules/dosomething/dosomething_fact_page/dosomething_fact_page.module
```diff
@@ -98,6 +98,7 @@ function dosomething_fact_page_preprocess_node(&$vars) {
     $node = entity_metadata_wrapper('node', $vars['node']);
     $values = dosomething_fact_get_fact_field_vars($node->field_facts);
     $vars['facts'] = $values['facts'];
+    $vars['facts_chunked'] = paraneue_dosomething_chunk_array($vars['facts'], 5);
```
aaronschachter: This function should really live in `paraneue_dosomething_preprocess_node`, since it's specific to the theme.  The module preprocess is more for variables which would be needed regardless of the specific theme.

Within `paraneue_dosomething_preprocess_node`, you'll see checks for `$type`.  You'll want to add another condition:

```
  elseif ($type == "fact_page") {
    // Maybe a comment here about what its doing
    $vars['facts_chunked'] = paraneue_dosomething_chunk_array($vars['facts'], 5);
  }
```

tjrosario:
Closing due to corrupt PR

PR #2709, opened 2014-06-27
refactored facts list to be returned in chunks for support for cta's emb...

### LINKED ISSUE #2703: Refactor facts listing to return chunked arrays

tjrosario:
Fixes #2703


DFurnes:
Looks good. :sailboat:

Issue #2708, opened 2014-06-27
[campaign close] - incorrect behavior for scholarship winner names


mikefantini:
When viewing closed campaigns as an anonymous user there is no first name displayed for any of the scholarship winners:

![image](https://cloud.githubusercontent.com/assets/6330971/3415754/99d28c04-fe24-11e3-989a-6f3cde307103.png)

When viewing closed campaigns as an authenticated user, that user's first name is pulled in as the first name for all of the scholarship winners:

![image](https://cloud.githubusercontent.com/assets/6330971/3415764/b6207876-fe24-11e3-9c35-97ca5a71e244.png)

PR #2710, opened 2014-06-27
Fix funky first names on winners & broken edit pages.

### LINKED ISSUE #2707: [campaign runs] - error when trying to access run edit view
Error message when trying to edit campaign runs:

image

angaither:
Get the value from the winner first name field if no uid
Use get field function on winners with uid.
Refs #2708

Only remove flagged reportbacks query, if there are flagged reportbacks.

Fixes #2707

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module
+++ b/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module
@@ -244,8 +244,7 @@ function dosomething_campaign_run_form_campaign_run_node_form_alter(&$form, &$fo
   if (isset($campaign_nid)) {
     // Get totals & language from the campaign.
     $signup_total_members = dosomething_signup_get_signup_total_by_nid($campaign_nid);
-    // @TODO: fix this
-    // $reportback_quantity = dosomething_reportback_get_reportback_total_by_nid($campaign_nid);
+    $reportback_quantity = dosomething_reportback_get_reportback_total_by_nid($campaign_nid);
     $language = dosomething_reportback_get_noun_and_verb($campaign_nid);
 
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module
```diff
@@ -111,11 +111,11 @@ function dosomething_campaign_run_preprocess_winners(&$vars, $nid) {
       $winners[$i]['uid'] = $winner[$entity_id]->field_user[LANGUAGE_NONE][0]['target_id'];
       // If there is no user, get the first name from the text field.
       if (!isset($winners[$i]['uid'])) {
-        $winners[$i]['fname'] = dosomething_user_get_field('field_first_name', user_load($winners[$i]['uid'], 'ucwords'));
+        $winners[$i]['fname'] = $winner[$entity_id]->field_winner_name[LANGUAGE_NONE][0]['value'];
       }
       // Else, grab the first name from the user account.
       else {
-        $winners[$i]['fname'] = user_load($winners[$i]['uid'])->field_first_name[LANGUAGE_NONE][0]['value'];
+        $winners[$i]['fname'] = dosomething_user_get_field('field_first_name', user_load($winners[$i]['uid'], 'ucwords'));
```
aaronschachter, line 118: This has a typo, it's passing `ucwords` as a parameter to `user_load`. Should be 

`dosomething_user_get_field('field_first_name', user_load($winners[$i]['uid']), 'ucwords');`

Because `user_load` actually does accept a 2nd param, and `dosomething_user_get_field` calls `dosomething_user_get_field_first_name` which sets the default `$format` to `'ucwords'` -- this should all still work, though we should fix to correctly call the function.

Issue #2676, opened 2014-06-25
Campaign Alternate Header Image


aaronschachter:
Sponsored campaign needs ability to display a different Header image on Pitch/Action pages, but still display the regular Campaign Cover Image in the Finder, Profile, etc.

Needed for July 9, for campaign launch.
- [x] Add a variable into Node Edit > Custom Settings > Styles to enter a Image node nid.
- [x] If said variable exists, display as the Pitch / Action Page header.  May need to tweak the "dark image" preprocess logic to check for this variable.

aaronschachter:
Implementing a table in DS Helpers called `dosomething_helpers_variable` to store this information. See #2726

aaronschachter:
This can be set in the Custom Settings form by entering the Node nid of the desired Image node.

![heaer-image-form](https://cloud.githubusercontent.com/assets/1236811/3460257/58479b4a-0215-11e4-90d3-ad5376978f7e.png)

Working on porting over all variables into this form to keep it all in one place / table per #2726

PR #2723, opened 2014-07-01
A field_variables field

aaronschachter:
This is not ready for merge just yet but opened for visibility and feedback. cc: @blisteringherb @desmondmorris -- Any thoughts / opinions?

Considering building a field collection of variables, instead of individual fields for a BG File Override, or a Header Image override (#2676).

The intention is that dosomething_helpers_form_extras stuff is getting a bit unwieldy: https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module

Would be easier if the variables were a part of the $node object vs always needing to execute `variable_get` to get various properties.

vs. this route https://github.com/aaronschachter/dosomething/compare/alt_header?expand=1
where we get into variable_get land when it comes time to `dosomething_campaign_preprocess_page`

screen shot 2014-07-01 at 8 31 21 am


desmondmorris:
Talked with @aaronschachter in chat.  Its okay with me for now.  We may want to review all of the field collection stuff at a later time though.

Issue #2712, opened 2014-06-30
Remove DS MBP Campaign Cache functionality


aaronschachter:
Blocked by #2711 

Once an endpoint is available to retrieve campaign data and Message Broker is configured to use it, all of the campaign cache related code in `dosomething_mbp` can be removed.
- [ ] Delete relevant variables in an install hook
- [ ] Remove all related code

aaronschachter:
This will be up on prod upon our next prod deployment -- it's currently live on staging: 

```
curl http://staging.beta.dosomething.org/api/v1/content/362  -H "Content-type: application/json" --header "Accept: application/json"
```

aaronschachter:
Closed by https://github.com/DoSomething/dosomething/commit/0451b0e22d310f2f006cbf8dc5a5d4310eca870a

PR #2519, opened 2014-06-13
Campaign  full template refactor

weerd:
#### What's this PR do?

This PR completely refactors the main campaign template and establishes some new patterns that will be moved over into Neue. It updates the campaign template to use the new `.container` module pattern we have been establishing for content across the site.
#### Where should the reviewer start?

Start by pulling down this branch and testing it locally. You should probably run `drush cc all` once the code is pulled down and before starting to test it.
#### How should this be manually tested?

Test a bunch of different campaigns. Make sure that things are looking correct and close to what they used to look like. Some spacing should be different now that it's more consistent and using the spacing rules defined in the `.container` module. Just make sure no campaigns look borked.
#### Any background context you want to provide?

In the "Drop It Off" stage the spacing for the paragraph and stuff below it is off. Known issue that will be fixed in a different PR to update the use of a markdown field instead of plain text in the CMS. Just FYI.
#### What are the relevant tickets?

Resolves #2307
Resolves #2028
Resolves #2293
Resolves #2425
Resolves #2028
Resolves #1206
Resolves #1177

PR #2735, opened 2014-07-01
More campaign class work

### LINKED ISSUE #2734: Add campaign class properties
Blocks #2712 

Add properties for:
- [x] `$campaign->is_staff_pick`
- [x] `$campaign->image_cover['src']`

Should unset the `$campaign->variables` property when outputting content to endpoint.  Possibly pass a parameter to `dosomething_campaign_load($node, $public = FALSE)`.  If `$public` is TRUE, hide non-content variable data type stuff.

aaronschachter:
- Fixes #2734 to unblock #2712
- Swaps out more entity wrapper within theme.inc vars with relevant campaign properties
- Fixes PHP warnings introduced by #2519, which occur when Plan It fields are not populated
  - Notice: Undefined variable: plan in include() (line 130 of /var/www/vagrant/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php).
  - Notice: Undefined index: plan in dosomething_campaign_preprocess_action_page() (line 188 of /var/www/vagrant/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc).

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -7,7 +7,12 @@
 /**
  * Returns a scaled down version of a campaign loaded $node.
  *
- * @return object
+ * @param obj $node
+ *   A loaded Node object.
+ * @param bool $public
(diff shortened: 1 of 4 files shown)


deezone:
:+1:

PR #2736, opened 2014-07-01
Deny access to file/[fid] paths

aaronschachter:
Updates `dosomething_user_init` to deny access to any File Entity path.

Also tweaks logic to only check for first two args, and removing the check for `arg(2)`.  If a user currently goes to a path like `node/823/test` -- they are able to view the node, since the check for arg(2) fails.

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -13,19 +13,20 @@ include_once 'dosomething_user.theme.inc';
  * Implements hook_init().
  */
 function dosomething_user_init() {
+  $args = array(arg(0), arg(1));
+
+  // If viewing a file/[fid] page:
+  if ($args[0] == 'file' && is_numeric($args[1])) {
+    dosomething_user_non_staff_redirect();
+  }
+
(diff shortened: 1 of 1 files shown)


sergiitk:
##### Minor code styling

I'm not sure if it make sense to create an array to store args.

I understand, accessing array is much faster than executing a function, but it's still a microoptimization, which impairs code readability.

Here's [one of examples](https://github.com/drupal/drupal/blob/7.x/modules/statistics/statistics.module#L62) from core modules. Even though `arg(1)` called twice, it wasn't assigned to a local variable.

By the way, function `arg()` [uses static cache](https://github.com/drupal/drupal/blob/7.x/includes/bootstrap.inc#L2859) so it's not so heavy as expected.

sergiitk:
Another one thing: it's better to check not for Drush in particular, but to detect if it's CLI, so
`+    if (!function_exists('drush_main')) {`
becomes
`+    if (drupal_is_cli()) {`

The [comment mentioned](https://www.drupal.org/node/1230132#comment-4790690) was initially about Drupal 6. Function drupal_is_cli() function was introduced in Drupal 7.

sergiitk:
Everything else is great!

PR #2742, opened 2014-07-02
Setting and fetching tier.  Capistrano doesn't automatically provide it

blisteringherb:
The symlinks for production aren't working properly after deployment because Capistrano doesn't automatically provide :stage as a variable.  This code will explicitly set and fetch the values for the deploy_env for production, qa, and staging.  The symlinking process is different for international.


aaronschachter:
Looks good :+1:

Issue #1507, opened 2014-03-24
Disable optimizely in ds pull 


aaronschachter:
Promote developer sanity and disable the optimizely module upon `ds pull stage`.  Technically needs to happen in the `pull_db` function.

deezone:
If possible, I'd like to figure this out. I'll work through the details and perhaps poke you a few times if it's OK with you? I'm close to a new release of the module and I'd like to understand / work on the `ds pull stage` disable options.

barryclark:
Was there any progress on this one @DeeZone?

deezone:
@barryclark Nope but I hope to have some extra bandwidth next week. I'll add this to my list.

angaither:
@DeeZone did you get a chance to look at this? is it done? I don't see anything in the build script about this, but I'm not sure if it happened elsewhere

deezone:
Ha! A 1.5 year later and...

> I'm close to a new release of the module and I'd like to understand / work on the ds pull 
> stage disable options

Is true once again. I'm working on a `7.x.-3.x` release of the module. I'll take a look and close this issue out this week.

PR #2760, opened 2014-07-03
Stage file proxy

### LINKED ISSUE #1575: Add Stage File Proxy, enable in QA/Vagrant environments
Important for keeping QA files in sync, and obviates the need to pull files in Vagrant.

blisteringherb:
This PR will download and enable the Stage File Proxy module when the 'dev' flag is passed to ds build.

Fixes #1575

@aaronschachter

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -366,7 +366,7 @@ function pull_files {
   then
     echo "Pulling down files from staging"
     cd $WEB_PATH
-    drush rsync -y @ds.staging:%files @self:%files
+    drush rsync @ds.staging @self -y
     symlinks
   fi
 }
--- a/bin/ds
+++ b/bin/ds
@@ -164,10 +164,6 @@ function build {
   echo 'Enabling modules in dosomething.info'
   drush --script-path=../bin/ php-script enable_modules.php
 
-  # Enable modules in info file
-  echo 'Enabling development modules'
-  drush en -y stage_file_proxy
-
   echo 'Reverting Features'
   drush fra -y --force
 
@@ -237,6 +233,7 @@ function symlinks {
 function dev {
   echo 'Enabling modules'
   drush en devel devel_generate -y
+  drush en -y stage_file_proxy
 
   echo 'Auto-generating Content'
   drush generate-terms sponsors 20
--- a/bin/ds
+++ b/bin/ds
@@ -233,7 +233,6 @@ function symlinks {
 function dev {
   echo 'Enabling modules'
   drush en devel devel_generate -y
-  drush en -y stage_file_proxy
 
   echo 'Auto-generating Content'
   drush generate-terms sponsors 20
@@ -354,6 +353,7 @@ function pull_db {
     echo "Pulling down db from staging"
     cd $WEB_PATH
     drush sql-sync @ds.staging @self -y
+    drush en -y stage_file_proxy
   fi
 }
 
@@ -366,7 +366,7 @@ function pull_files {
   then
     echo "Pulling down files from staging"
     cd $WEB_PATH
-    drush rsync @ds.staging @self -y
+    drush rsync -y @ds.staging:%files @self:%files
     symlinks
   fi
 }
@@ -502,7 +502,7 @@ fi
 
 
 #----------------------------------------------------------------------
-# test 
+# test
 #----------------------------------------------------------------------
 if [[ $1 == "test" ]]
 then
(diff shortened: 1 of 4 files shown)


config/default.settings.local.php
```diff
@@ -56,3 +56,5 @@
 $conf['apachesolr_path'] = '/solr/collection1';
 $conf['apachesolr_read_only'] = 1;
 
+$conf['stage_file_proxy_origin'] = "http://www.dosomething.org";
```
aaronschachter: Because we pull down from stage, and use it as our dev database, maybe it makes sense for us to set this to staging.beta.dosomething.org?  Then potentially the `ds pull --db` function in `ds` could `drush -y en stage_file_proxy`.  If we go that route, we could also knock out #1507

bin/ds
```diff
@@ -229,6 +229,7 @@ function symlinks {
 function dev {
   echo 'Enabling modules'
   drush en devel devel_generate -y
+  drush en -y stage_file_proxy
```
aaronschachter: Whoa, we haven't used this `--dev` function in ages. See comment around possibly integrating this into the `ds pull`. If we go that route, its tempting to just kill this `dev` function completely...

blisteringherb:
I like the idea of the dev function even though it only enables devel and devel_generate at the moment, but I can move enabling the stage_file_proxy into ds pull.

aaronschachter:
I'm definitely down to keep the `--dev` function, esp for enabling `devel` and `stage_file_proxy`.  Perhaps we just need to remove all the Devel Generate stuff.

blisteringherb:
I'm thinking that maybe you wouldn't need stage_file_proxy unless you ran a ds pull (--db), so it might be better suited to go in the pull_db function.

config/default.settings.local.php
```diff
@@ -56,3 +56,5 @@
 $conf['apachesolr_path'] = '/solr/collection1';
 $conf['apachesolr_read_only'] = 1;
 
+$conf['stage_file_proxy_origin'] = "http://staging.beta.dosomething.org";
```
aaronschachter, line 68: Can you confirm this works for you locally? I can test tomorrow, I've set up the module manually locally but I think we may also need to set the `stage_file_proxy_origin_dir` variable to `sites/default/files` ?

aaronschachter:
This seems crazy. Everything looks alright in this PR, but I pulled this branch down, ran a `ds build` and got this WSOD message:

Fatal error: Call to undefined function drupal_session_initialize() in /var/www/vagrant/html/includes/bootstrap.inc on line 2247

Trying a `ds build` a second time and I still get it.  Switched to my dev branch, ran ds build, and now 

`Fatal error: Call to undefined function drupal_bootstrap() in /var/www/vagrant/html/index.php on line 20`

After I do a `vagrant provision` from the dev branch, things are back to normal.

I don't know if that's really helpful.  Would be useful if someone else can pull down and test to verify this PR works.

aaronschachter:
A third re-attempt to ds build and Fatal error: Call to undefined function module_load_all() in /var/www/vagrant/html/includes/bootstrap.inc on line 2481

This is really crazy. It

In the output it says:
stage_file_proxy-7.x-1.5 downloaded.   

But in html/profiles/dosomething/modules there is no stage_file_proxy module downloaded.

blisteringherb:
@aaronschachter I'm not sure what that could be from.  I tried reproducing the issue by pulling down the branch, vagrant destroying, and then vagrant up, and I was able to do a ds build successfully.  I can also confirm that the module works with the staging URL.

angaither:
This mostly works for me after a build, however in order to get the `$conf['stage_file_proxy_origin']` variable set correctly in `settings.local.php` I had to first remove that file.

Not sure how we want to approach adding things to local settings file as we probably don't want to overwrite any custom settings, but we need to add new variables.

Another question, in `ds pull` once this is merged would want to remove the `--files` pull portion?

:+1:

aaronschachter:
@blisteringherb and I were discussing this... the only reason to keep the `--files` would be to get the latest modules/theme files from the html directory, if you didn't want to run a `ds build` to get the latest.  I'm not really opposed to just forcing people to get in the habit of running the build.  

If we can hook up https://github.com/DoSomething/dosomething-dist into vagrant, that would remove the need for that though.  That repo runs `ds build` every time we create release.

aaronschachter:
Re: https://github.com/DoSomething/dosomething/pull/2760#issuecomment-51093919

This is not working for me either, and deleting `settings.local` doesn't sound like a good solution.  Perhaps we should just set the variable in the `ds pull --db` when we enable the module?

PR #2763, opened 2014-07-03
Fix missing subtitle on pitch page.

### LINKED ISSUE #2762: Subtitle is missing on pitch page
The subtitle field seems to be missing on the pitch page.

screen shot 2014-07-03 at 2 14 25 pm

DFurnes:
Fixes #2762.

/cc @aaronschachter


aaronschachter:
Whoa, good catch. Thanks @DFurnes

PR #2766, opened 2014-07-03
Updating the search index display mode for indexed type

### LINKED ISSUE #2388: Search results show file names within preview
Would be rad if we could hide this. Reads pretty pretty weird. 

screen shot 2014-05-30 at 2 55 23 pm

Hahahahhahahahah

blisteringherb:
Restricting which fields get indexed as 'content' to avoid displaying file names or other unnecessary information in highlighted search results.

Fixes #2388

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 .vagrant
-.DS_Store
+*.DS_Store
 html/
 settings.local.php
 build/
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -1,5 +1,5 @@
 .vagrant
-.DS_Store
+*.DS_Store
```
aaronschachter, line 2: just checking -- this is intentional ?

aaronschachter:
Picky but can you `git commit --amend -m "Fixes #2388"` to provide a better commit message?

.gitignore
blisteringherb, line 2: Yes, I had a bunch of _.DS_Store files in my dosomething mount.  Since we never need .DS_Store files I was adding that.

aaronschachter:
Ship it :+1: I spot checked a few of the fields, but didn't go thru every single one. Guess we'll find out if things look weird :)

PR #2767, opened 2014-07-03
Changing Campaign fields to markdown

aaronschachter:
Changes the `text_processing` for `field_solution_support` to use filtered text.

Updates all current `field_solution_support` and `field_post_step_copy` values to use Markdown.

How to test:
- Pull this branch down
- Run `drush -y fr dosomething_campaign` to change the text format for Solution Support
- Run `drush updb` to bulk update the field values.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.features.field_instance.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.features.field_instance.inc
@@ -1975,7 +1975,7 @@ function dosomething_campaign_field_default_field_instances() {
     'label' => 'Supporting Statement on the Solution',
     'required' => 0,
     'settings' => array(
-      'text_processing' => 0,
+      'text_processing' => 1,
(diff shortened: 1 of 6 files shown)


sergiitk:
Functionality — OK.
Style:
- [`dosomething_campaign.install:66`](https://github.com/DoSomething/dosomething/pull/2767/files#diff-7d934c7e3d7aaf001b4c5b0a0c20568bR66) When it's only one field in `db_select` you can use [`addField`](https://api.drupal.org/api/drupal/includes%21database%21select.inc/function/SelectQuery%3A%3Afields/7) method to avoid another one array.
- [`dosomething_campaign.install:79`](https://github.com/DoSomething/dosomething/pull/2767/files#diff-7d934c7e3d7aaf001b4c5b0a0c20568bR79) Instead of string concatenation it's better to use watchdog's thrid argument [`variables`](https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/watchdog/7).

PR #2768, opened 2014-07-03
SMS report back for #ImABoss

jonuy:
Same deal as the report backs from before.

mData endpoint setup to POST to `dosomething.org/dosomething_sms/imaboss-reportback`. The only real thing to note is `$activity->nid = 2401;` which should be the #ImABoss node id.

Tested stepping through the SMS flow on my local by posting requests at the endpoint to mimic Mobile Commons requests.

### DIFF
--- a/lib/modules/dosomething/dosomething_sms/workflows/dosomething_sms.workflows.inc
+++ b/lib/modules/dosomething/dosomething_sms/workflows/dosomething_sms.workflows.inc
@@ -193,3 +193,59 @@ $activity->name = 'end';
 $activity->inputs = array('submit_reportback');
 
 $workflows[$workflow->name] = $workflow;
+
+/**
+ * Report back for Fifth Harmony #ImABoss 2014.
+ */
+$workflow = new ConductorWorkflow();
+$workflow->wid = 'new';
+$workflow->name = 'dosomething_sms_imaboss_reportback';
+$workflow->title = 'Fifth Harmony #ImABoss 2014 Report Back';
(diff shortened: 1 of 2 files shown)


aaronschachter:
:+1:

lib/modules/dosomething/dosomething_sms/workflows/dosomething_sms.workflows.inc
```diff
@@ -193,3 +193,59 @@ $activity->name = 'end';
...
+$activity->name = 'submit_reportback';
+$activity->inputs = array('strip_signature');
+$activity->outputs = array('end');
+$activity->nid = 2401; // #ImABoss node id
+$activity->mmsContext = 'ask_picture:mms';
+$activity->propertyToContextMap = array(
+  'quantity' => 'ask_quantity:message',
+  'why_participated' => 'ask_why_participated:message',
+);
+$activity->optInPathId = 168589;
```
deezone, line 243: Perhaps move this below `$activity = $workflow->newActivity('end');` to complete the `$activity = $workflow` entries before assigning the Id values?

jonuy, line 243: These id values apply to the `$activity` assigned above (`$activity = $workflow->newActivity('dosomething_sms_submit_reportback');`). So it needs to happen before the `$activity = $workflow->newActivity('end');`

deezone:
:+1:

PR #2769, opened 2014-07-03
Tests are great

DFurnes:
## Changes:
- Moves [functional-testing](https://github.com/dosomething/functional-testing) into the main repo. This means we can build tests alongside features (a la TDD), and follows common practices for test organization in a project.
- Tests can be run using `ds test` or `ds test [directory or file]`. This will run CasperJS tests, and perform pre-test and post-test scripts. There's also a sneaky `CI_TESTING` argument to set the CasperJS `url` variable to the QA instance for running tests on QA.
- Some modifications to login, register, and campaign signup tests:
  - Using modal login/register form now to better reflect real user behavior.
  - Tests existence of proper headers on campaign signup test.
## To test this PR:

Re-provision your Vagrant box to ensure that PhantomJS and CasperJS are installed. Pull database from staging to ensure that you have the `QA_TEST_ACCOUNT@example.com` account.
## What to expect:

screen shot 2014-07-03 at 5 13 12 pm


blisteringherb:
:+1:  This looks good to me.  I don't know much about the testing, but there's no risk to introducing it.

aaronschachter:
Agree - no risk in introducing it. I think we may want to look into creating the test Campaign node programatically through a drush command, and then testing against that.

 https://qa.dosomething.org/campaigns/test-campaign is unpublished, so I'd image the tests fail.  Plus you then need to worry about our stage and QA databases not being in sync. I'm pretty sure I already unknowingly made changes to the Test Campaign node in staging just to test my own stuff.

Something to add in later though.  Keep on fighting the good fight, @DFurnes -- this looks great so far.

DFurnes:
Good call on creating test nodes programmatically. Definitely want to look into that. :+1:

Issue #2718, opened 2014-06-30
Add reportback image path to mbp payload


deezone:
Add the path to the image related to a campaign report back to the payload sent to mbp `dosomething_mbp_request('campaign_reportback', $params)` requests.

deezone:
@aaronschachter Where can get the image id / url to the image uploaded with a campaign report back?

Looking at `dosomething_reportback_mbp_request()` I can see the submitted `$entity`:

```
ReportbackEntity Object
(
    [files_table:protected] => dosomething_reportback_file
    [log_table:protected] => dosomething_reportback_log
    [field_data_table:protected] => dosomething_reportback_field_data
    [entityType:protected] => reportback
    [entityInfo:protected] => Array
        (
            [label] => Reportback
            [base table] => dosomething_reportback
            [entity keys] => Array
                (
                    [id] => rbid
                    [label] => rbid
                    [revision] => 
                    [bundle] => 
                )

            [entity class] => ReportbackEntity
            [uri callback] => entity_class_uri
            [controller class] => ReportbackEntityController
            [module] => dosomething_reportback
            [views controller class] => EntityDefaultViewsController
            [fieldable] => 
            [static cache] => 1
            [field cache] => 1
            [load hook] => reportback_load
            [bundles] => Array
                (
                    [reportback] => Array
                        (
                            [label] => Reportback
                        )

                )

            [view modes] => Array
                (
                )

            [translation] => Array
                (
                )

            [schema_fields_sql] => Array
                (
                    [base table] => Array
                        (
                            [0] => rbid
                            [1] => uid
                            [2] => nid
                            [3] => created
                            [4] => updated
                            [5] => quantity
                            [6] => why_participated
                        )

                )

            [token type] => reportback
            [apachesolr] => Array
                (
                    [indexable] => 
                    [status callback] => 
                    [document callback] => 
                    [reindex callback] => 
                    [bundles changed callback] => 
                )

            [configuration] => 
        )

    [idKey:protected] => rbid
    [nameKey:protected] => rbid
    [statusKey:protected] => status
    [defaultLabel:protected] => 
    [rbid] => 123
    [uid] => 1700273
    [nid] => 15
    [created] => 1404157597
    [updated] => 1404157597
    [quantity] => 1
    [why_participated] => cuz
    [fids] => Array
        (
        )

    [node_title] => Momm-o-grams
    [noun] => cards
    [verb] => given
    [field] => 
)
```

[...]

PR #2777, opened 2014-07-07
Expose reportback fid after insert

aaronschachter:
The `ReportbackEntity->save()` adds the corresponding `dosomething_reportback_file` record after the entity is saved, in order to insert a record with the newly created `rbid`.  This is why the `hook_insert_entity` method was not finding any `fids` array when loading the reportback.

Unblocks #2718
cc @DeeZone @deadlybutter


itsjoekent:
@aaronschachter thanks a ton man!

PR #2782, opened 2014-07-07
added tablet media query for polaroid positioning

### LINKED ISSUE #2765: [campaign template] - Do It image not showing up on mobile
On mobile, the field_image_step_gallery from the Do It gallery is not being pulled in properly.

It should be pulling in the landscape crop that is uploaded.

tjrosario:
Fixes #2765


DFurnes:
:palm_tree:

PR #2785, opened 2014-07-07
Removing styles no longer used for static pages now that they've been re...

weerd:
...factored to use the container module.

@DFurnes @tjrosario Cleanup time!


DFurnes:
this makes me so happy

weerd:
lol. best approval yet.

PR #2787, opened 2014-07-07
Add campaign signup tests

DFurnes:
Add tests checking for proper pitch page content; and signing up for a campaign from the pitch page as an unregistered user, registered & logged-out user, or as a logged-in user.

screen shot 2014-07-07 at 5 03 46 pm

Jira Issue #FT-6

### DIFF
--- a/scripts/tests/before_tests.sh
+++ b/scripts/tests/before_tests.sh
@@ -13,17 +13,28 @@ cd /var/www/vagrant/html
 drush sql-query "DELETE FROM flood WHERE identifier LIKE '%127.0.0.1';"
 
 
-# Delete test user account created in `user/register.js`
-echo "Creating fresh test accounts..."
-{
-  ## Delete users created during previous test runs...
-  drush user-information QA_TEST_ACCOUNT@example.com | grep "User ID" | sed -e 's/[ ]*User ID[ ]*\:[ ]*//g' | xargs -i drush user-cancel {} -y
-  drush user-information QA_TEST_USER_REGISTER@example.com | grep "User ID" | sed -e 's/[ ]*User ID[ ]*\:[ ]*//g' | xargs -i drush user-cancel {} -y
-  drush user-information QA_TEST_CAMPAIGN_SIGNUP_EXISTING@example.com | grep "User ID" | sed -e 's/[ ]*User ID[ ]*\:[ ]*//g' | xargs -i drush user-cancel {} -y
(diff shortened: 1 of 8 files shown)


scripts/tests/before_tests.sh
```diff
@@ -5,14 +5,25 @@
...
-# Get test user account UID...
-uid=$( drush user-information QA_TEST_ACCOUNT@example.com | grep "User ID" | sed -e 's/[ ]*User ID[ ]*\:[ ]*//g' )
 
-# Remove all campaign signups for test account
-drush sql-query "DELETE FROM dosomething_signup WHERE uid='$uid';"
+# Delete test user account created in `user/register.js`
+echo "Creating fresh test accounts..."
+{
+  ## Delete users created during previous test runs...
+  drush user-information QA_TEST_ACCOUNT@example.com | grep "User ID" | sed -e 's/[ ]*User ID[ ]*\:[ ]*//g' | xargs -i drush user-cancel {} -y
```
aaronschachter: Not critical, but would be nice to make this a function to avoid copy/pasting all this `grep` stuff.

tests/campaign/signup.js
```diff
@@ -3,29 +3,96 @@
...
+      "field_first_name[und][0][value]": "Panda",
+      "field_birthdate[und][0][value][date]": "01/04/1989",
+      "mail": "QA_TEST_CAMPAIGN_SIGNUP_NEW@example.com",
+      "pass[pass1]": "QA_TEST_CAMPAIGN_SIGNUP_NEW",
+      "pass[pass2]": "QA_TEST_CAMPAIGN_SIGNUP_NEW",
+    }, true);
+    
+    // Make sure the signup happened.
+    casper.waitForSelector('.messages.status', function() {
+      test.assertSelectorHasText('.messages.status', 'You\'re signed up for', 'User can register and sign up for a campaign');
```
aaronschachter: Can we make the `'You\'re signed up for'` a global variable somewhere to use for all of these tests? That way if the copy ever changes, we can change it once.

Potentially if we had a Signup API -- we could query for a signup sid here, no?  To verify that a signup record exists in the db.

scripts/tests/before_tests.sh
DFurnes: Mmm, good call.

tests/campaign/signup.js
DFurnes: Both good points. :+1:

Issue #2674, opened 2014-06-24
Fact Page: Facts field collection


aaronschachter:
Desired functionality is to append copy to the end of a fact within a fact page, e.g. 

> Safe, insecticide-treated bed nets repel and kill disease-carrying mosquitoes to reduce the spread of malaria. Check out <a href="#">Comeback Clothes</a> to learn how to fight disease.
- [x] Build a field collection `field_fact_collection` into the Fact Page content type.
  - [x] `field_fact` -- Fact. mutli-value entity reference to a fact node. This could potentially just reuse `fact_solution` or `fact_problem`
  - [x] `field_compound_text_copy` -- Fact Copy. Use existing field
  - [x] `field_campaigns` -- Links. Use existing field.  This potentially could even just be `field_node` so its not content type specific, but then could just be something like `field_nodes` or hell why not `field_entity`
- [x] Build a token system to use [link] to replace any instances with a `l` to the node specified in `field_campaigns` 
- [x] Import existing `field_facts` values to the new field collection.  This is nasty. OR an intern manually does this.
- [x] Update the preprocess / template
- [x] Drop `field_facts` Entity Reference field once all Fact Pages have been migrated.

aaronschachter:
Deploying to stage failed, because the `field_fact_collection` did not exist when running 7002, suggesting that the code added to manually revert the feature did not work.  This is kinda crazy because it worked for both me and @sergii-tkachenko.

```
Updating database...
DEBUG [6fc07136]    Added the administrator role to uid 1                                  [success]
DEBUG [6fc07136]     Dosomething_fact_page  7002  Imports existing field_facts data into the        
DEBUG [6fc07136]                                  field_fact_collection field.                      
DEBUG [6fc07136]     Dosomething_fact_page  7003  Removes field_facts from Fact Page content type.
DEBUG [6fc07136]    Do you wish to run all pending updates? (y/n): 
DEBUG [6fc07136]    Added the administrator role to uid 1                                  [success]
DEBUG [6fc07136]    y
DEBUG [6fc07136]    Added the administrator role to uid 1                                  [success]
DEBUG [6fc07136]    y
DEBUG [6fc07136]    Invalid field name given: field_fact_collection is not a Field           [error]
DEBUG [6fc07136]    Collection field.
DEBUG [6fc07136]    y
DEBUG [6fc07136]    Performed update: dosomething_fact_page_update_7002                         [ok]
DEBUG [6fc07136]    y
[...]

PR #2789, opened 2014-07-07
Fact collection field collection

aaronschachter:
Refs #2674

@sergii-tkachenko Would you mind reviewing?
## How to test
- Pull down staging DB
- Run the following commands (which is the same order that our `ds build` script executes in as well):
  - `drush updb` -- Should run `dosomething_fact_page_update_7002`
  - `drush -y fr dosomething_fact_page`
- Confirm your local Fact Page nodes display the same Facts as they do on staging
- Confirm the instance of `field_facts`, a multi-value Entityreference field, has been removed from the `fact_page` contentt ype.
## Overview
- Creates a field collection, `field_fact_collection`, with fields:
  - Fact (entityreference to the Fact node)
  - Copy
  - Campaign
- Migrates the Fact Page's `field_facts` values into the new `field_fact_collection`, into the `field_fact` field.
- Alters `dosomething_fact_page_preprocess_node` to use a new method, `dosomething_fact_page_get_field_fact_collection_vars`, which inspects the field collection data
- If copy field is set, appends the copy to the end of the fact. Links to the campaign (if set) whereever an editor uses a [link] token.
- Deletes the `field_facts` instance from the Fact Page node
## Screenshot

screen shot 2014-07-08 at 9 53 47 am
screen shot 2014-07-08 at 9 56 00 am


aaronschachter:
Update: Copy/Campaigns are now being displayed in the template file, closing #2674. updated screenshots and issue description accordingly.

sergiitk:
Syntax comments (in progress):
- [`dosomething_fact_page.install:21`](https://github.com/DoSomething/dosomething/pull/2789/files#diff-4c5be4919aeb275975780bd9b3456999R23): `dosomething_helpers_get_node_vars` can return `NULL` so it's better to do something like

```
  if (empty($fact_pages)) {
    return;
  }
```
- [`dosomething_fact_page.install:29`](https://github.com/DoSomething/dosomething/pull/2789/files#diff-4c5be4919aeb275975780bd9b3456999R29): it's either no curly braces or `continue;` on the new line
- [`dosomething_fact_page.install:32`](https://github.com/DoSomething/dosomething/pull/2789/files#diff-4c5be4919aeb275975780bd9b3456999R32): `$delta` not used

sergiitk:
:+1:

Issue #2790, opened 2014-07-07
[campaign template] - Do It image in wrong location on mobile/tablet


mikefantini:
Related to #2765 the image should be _above_ Snap a Pic, not below:

Live:
![image](https://cloud.githubusercontent.com/assets/6330971/3503246/8b809900-062d-11e4-912c-f28b20262549.png)

Mock
![image](https://cloud.githubusercontent.com/assets/6330971/3503248/938ca062-062d-11e4-9cc6-4f39cee55209.png)

mikefantini:
Also on mobile/tablet it should be pulling in the landscape crop not the square crop

weerd:
(Moving my comment from the PR into this issue)

@mikefantini or @crystalruthbell I'd argue that this isn't a content image and more of a design element and thus not required for mobile... might be something to address in future. It doesn't detract from the content message if missing on mobile. But probably a discussion for another day ;)

weerd:
Also if we're calling landscape image on mobile as well as storing the square crop for the polaroid for desktop, then we're downloading 2 images on mobile; one of which isn't used unless viewing afterwards on desktop. Something to consider since it's an extra image download that can affect performance for mobile.

If performance is a concern, easiest solution is to not show an image on mobile and only show the polaroid on desktop (makes only 1 image download, but still isn't shown... not ideal but meh...). Otherwise, alternative solution is to load the landscape first for mobile, and then if desktop use JavaScript to load in and switch to the square version, which is a bit more involved but also doable.

Just throwing these out there for reference ;)

@DFurnes and I discussed this when I was refactoring the polaroid and the template to use the container stuffz and decided scenario 1 above was the best solution for the time being (not show square image on mobile and only show on tablet -> desktop).

DFurnes:
+1 to using the same crop (square) on mobile for now.

Best solution would be adding `<picture>` support with something like [Picturefill](http://scottjehl.github.io/picturefill/) imo.

weerd:
Yeah def. Picturefill would be ideal, and we could start harnessing it for other image needs throughout the site :)

mikefantini:
@weerd @DFurnes including the image on mobile is mainly focused on breaking up all the text (which can get both a. boring and b. overwhelming especially on mobile).

Ideally it does also provide value as generally it is meant as another way to show the user others taking action (visually showing people how to do the campaign).

[...]

PR #2795, opened 2014-07-08
repositioned polaroid images to appear above 'snap a pic' headline

tjrosario:
References #2790


weerd:
Looks good :+1:  

@mikefantini or @crystalruthbell I'd argue that this isn't a content image and more of a design element and thus not **required** for mobile... might be something to address in future. It doesn't detract from the content message if missing on mobile. But probably a discussion for another day ;)

PR #2797, opened 2014-07-08
Hotfix to get Fact pages back

aaronschachter:
PR #2789 seems to have failed because `dosomething_fact_page_update_7002` did not properly revert the `dosomething_fact_page` feature, which adds the `field_fact_collection` that we're attempting to import the `field_facts` data into.

not sure why it worked locally for me and @sergii-tkachenko but i'd rather not chance it on the prod site.

Instead of trying to do it all the field-adding and importing in one shot, this PR sets a `dosomething_fact_page_fact_collection_imported` to `FALSE` to indicate that the fact collections have not been imported, and to use the existing field fact values.

Comments out `dosomething_fact_page_update_7003` because that's the code we'll need once features are reverted and the field collection is added, in order to finally close out #2674 for good.

### DIFF
--- a/lib/modules/dosomething/dosomething_fact_page/dosomething_fact_page.module
+++ b/lib/modules/dosomething/dosomething_fact_page/dosomething_fact_page.module
@@ -95,7 +95,19 @@ function dosomething_fact_page_preprocess_node(&$vars) {
       }
     }
 
-    $values = dosomething_fact_page_get_fact_collection_vars($vars['node']);
+    // @todo: Remove this conditional once fact_collection has been imported.
+    // If the fact collections have been imported:
+    if (variable_get('dosomething_fact_page_fact_collection_imported') == TRUE) {
+      // Get values from field_fact_collection.
+      $values = dosomething_fact_page_get_fact_collection_vars($vars['node']);
+    }
+    // Else if not imported yet:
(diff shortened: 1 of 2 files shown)


aaronschachter:
This will work on the staging DB because i've restored to an earlier backup from today, where the schema of `dosomething_fact_page` is still 7001.

lib/modules/dosomething/dosomething_fact_page/dosomething_fact_page.module
```diff
@@ -95,7 +95,19 @@ function dosomething_fact_page_preprocess_node(&$vars) {
       }
     }
 
-    $values = dosomething_fact_page_get_fact_collection_vars($vars['node']);
+    // @todo: Remove this conditional once fact_collection has been imported.
+    // If the fact collections have been imported:
+    if (variable_get('dosomething_fact_page_fact_collection_imported') == TRUE) {
```
sergiitk, line 100: You can just do `if (variable_get('dosomething_fact_page_fact_collection_imported')) {` :smile:

sergiitk:
Looks sain for me. Will test it and get back for you.

sergiitk:
:+1: All tests passed on my env.

PR #2798, opened 2014-07-08
Issue2718 add reportback image url to mbp payload

### LINKED ISSUE #2718: Add reportback image path to mbp payload

deezone:
Fixes #2718

Adds report back image markup value to payload values sent to mbp.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -370,6 +370,8 @@ function dosomething_reportback_save($values) {
     $entity = entity_load_single('reportback', $rbid);
   }
   try {
+    // Send reportback transactional message
+    dosomething_reportback_mbp_request($entity);
     // Set entity properties.
     dosomething_reportback_set_properties($entity, $values);
     // Set entity files.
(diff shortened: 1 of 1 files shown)


aaronschachter:
:+1:

deezone:
It's not clear to me how ReportbackEntityController->save gets called to trigger the `dosomething_reportback_mbp_request($entity);` call? Is there some Drupal magic connecting `function dosomething_reportback_save()` to the class?

aaronschachter:
Yes, this one actually is some Drupal magic.  We create a reportback entity in `dosomething_reportback_save` and use the save method on it: https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module#L380

And in the ReportbackentityController class, we override the `save` method: https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc#L376

deezone:
Gotta love OO. Thanks for the explanation, makes perfect sense now that you point it out.

PR #2802, opened 2014-07-09
Import field_facts into field_fact_collection

aaronschachter:
Now that we've deployed the latest feature changes to the Fact Page to add the `field_fact_collection` field collection, we can safely import the values from `field_facts` into it and drop `field_facts` from the content type.

@sergii-tkachenko -- mind giving this a quick review so we can deploy today?


sergiitk:
:+1:

PR #2803, opened 2014-07-09
Remove deprecated Fact Page import variables and functions

aaronschachter:

Refs #2674 and #2797

Now that the `field_facts` data has been imported into `field_fact_collection` and has been removed from the Fact Page content type, this PR gets rid of all the code we don't need anymore (and don't want lingering around as a function which someone could potentially call)


sergiitk:
:+1:

PR #2804, opened 2014-07-09
Editor permissions to add Partners terms

### LINKED ISSUE #877: Partners: Editors Need Ability to Add
Editors need access to http://stage.dosomething.org/admin/structure/taxonomy/partners/add

cc: @barryclark

aaronschachter:
Adds Taxonomy Access Fix contrib module to introduce "Add terms in [vocab]" permissions.

Grants Editor role permission to add terms to Partners to fix #877.

Also removes the permission to edit tags from Editors.  This was in place but they never accessed it because the Structure menu wasn't available (Taxonomy Access Fix now makes it available to them).  Per convo with @mikefantini -- too many cooks in the kitchen to have editors allow access to edit.


aaronschachter:
Editors will now see a "Structure" menu which will only allow them access to Partners terms.

screen shot 2014-07-09 at 6 02 24 pm

PR #2805, opened 2014-07-10
I18N: Make "What is DoSomething.org?" editable.

sergiitk:
[DOS-12](https://jira.dosomething.org/browse/DOS-12):
Change the "What is DoSomething.org?" header item to "Who Are We?" for
all affiliate sites.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/system/partials/navigation.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/system/partials/navigation.tpl.php
@@ -18,9 +18,9 @@
       </li>
 
       <li>
-        <a href="<?php print url('node/538');  ?>">
+        <a href="<?php print url('node/' . $who_we_are['link']); ?>">
           <?php //<strong>What is D<span>o</span>S<span>omething</span>.org?</strong> @TODO: Temp hidden. ?>
-          <strong><?php print $who_we_are; ?></strong>
+          <strong><?php print $who_we_are['text']; ?></strong>
           <span>Young people + social change.</span>
         </a>
       </li>
--- a/lib/themes/dosomething/paraneue_dosomething/templates/system/partials/navigation.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/system/partials/navigation.tpl.php
@@ -3,7 +3,6 @@
 /**
  * Generates site-wide global page nav.
  **/
-
 ?>
 
 <nav class="chrome--nav<?php if(isset($modifier_classes)): print ' ' . $modifier_classes; endif; ?>">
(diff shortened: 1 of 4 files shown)


sergiitk:
DOS-12 is ready, "What is DoSomething.org?" is changeable now. However, it may also be needed to change a link to actual "About us" page. @barryclark, @julielorch could you clarify that?
@angaither could you please review this PR?

barryclark:
@sergii-tkachenko I don't think the link to the About Us page is changing. Is that right @julielorch?

joolaay:
nope - not changing. We just want a different title in the header for "What is DoSomething.org"

sergiitk:
@barryclark What I'm worried about is node id.
I don't know how international setup were init. If the starting point for these is DS production, it will work. If these were built from scratch (`ds build --install`), node id will change.

lib/themes/dosomething/paraneue_dosomething/templates/system/partials/navigation.tpl.php
```diff
@@ -21,7 +20,7 @@
       <li>
         <a href="<?php print url('node/538');  ?>">
```
sergiitk: Currently nid is hardcoded: `node/538`.

barryclark:
@sergii-tkachenko Thoughts on using a link to "about/who-we-are" instead of the node id?

barryclark:
Meh, actually that's a crappy idea, as if we changed the node's alias the link would be dead.

sergiitk:
@barryclark exactly. Plus, non-english sites may want to have aliases in its language.

aaronschachter:
No menus should be hardcoded with node nid's, we should use variables similar to what we use for the footer....

sergiitk:
@barryclark But so is ugly hardcoding a nid.
Moreover, I'd suggest using [uuid](https://www.drupal.org/project/uuid) module.

angaither:
@sergii-tkachenko uuid is installed :)

sergiitk:
Thank you for pointing that out, @angaither!
It turned out I accidentally put ignored `html/profiles/dosomething/modules/` in my SublimeText.

sergiitk:
I'm thinking to use [uuid_entity_autocomplete](https://www.drupal.org/project/uuid_entity_autocomplete) module.
Since we have a lot of nids hardcoded in the templates and anyway will replace these for i18n project, it'd may make sense to provide nice UI for selecting a node.

angaither:
this looks ok! :+1: 
I don't feel great about that uuid module, it doesn't seem to be supported or widely used

sergiitk:
@angaither I agree with you, but I'd like to have a little experiment with the module before throwing it out.
I'll merge what we have now and get back to you after the experiment.

sergiitk:
@angaither, you were right, this uuid_entity_autocomplete module was rubbish. After selecting a node in the dropdown it just replaces input a node title with meaningless UUID.

sergiitk:
@angaither But take a look at the [Entity Autocomplete](https://www.drupal.org/project/entity_autocomplete). It's 2,362  sites using this module / 17,851 downloads.

Here's how selection works (It is already limited to Static Content pages only):
image
With page selected:
image

And compare to current look and UI:
image

I've saved a patch from my experiment and it'd take just a minute to commit it.
@angaither @julielorch @barryclark Should I go for it?

PR #2807, opened 2014-07-10
Page  footer templatization

### LINKED ISSUE #2806: [Global] Abstracting Info-Bar pattern markup into a template theme.
Turn the markup for the info-bar pattern footer that shows up on all campaign types and static pages into a template theme call, so the markup is defined in one place and another templates can call it via Drupal's theme function.

weerd:
Refactoring the `info-bar` pattern markup into a template that can be called from other templates using Drupal's `theme()` function.

Resolves #2806

@aaronschachter (or @angaither since Sharon Actor is on vacay :tropical_drink: ) and @DFurnes

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
@@ -26,7 +26,7 @@ function paraneue_dosomething_preprocess_page(&$vars) {
   }
 
   /**
-   * Navigation
+   * Global Main Navigation
    */
 
   $modifier_classes = 'white floating';
@@ -48,7 +48,7 @@ function paraneue_dosomething_preprocess_page(&$vars) {
 
 
   /**
-   * Header
+   * Global Page Header
    */
 
   $header_vars = array(
@@ -64,7 +64,7 @@ function paraneue_dosomething_preprocess_page(&$vars) {
 
 
   /**
-   * Footer
+   * Global Page Footer
    */
 
   $columns = array('first', 'second', 'third');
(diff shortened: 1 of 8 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
```diff
@@ -133,5 +134,18 @@ function paraneue_dosomething_preprocess_node(&$vars) {
...
     $vars['facts_chunked'] = array_chunk($vars['facts'], 5);
   }
+
+
+  /**
+   * Info Bar
+   */
+
+  $info_bar_vars = array(
+    'zendesk_form'       => $vars['zendesk_form'] || $vars['content']['zendesk_form'], // @TODO: not ideal... unify these!
```
weerd, line 144: This is up next. Would like to make it so that there's only 1 array item to check, but currently grouped campaigns dumps the zendesk value into the content array. It's the only one that does that. Will need to update the module, and may need some bendering help! :tropical_fish:

angaither:
Look at that drupal! :+1: :elephant:

weerd:
drupal-icon-3d-small

DFurnes:
Nice! :flags:

lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
```diff
@@ -133,5 +134,18 @@ function paraneue_dosomething_preprocess_node(&$vars) {
...
+    'sponsors'           => $vars['sponsors'],
```
aaronschachter, line 145: All this stuff needs to be checked before it's outputted as variable.  Go to a campaign without sponsors:

```
Notice: Undefined index: sponsors in paraneue_dosomething_preprocess_node() (line 149 of /var/www/vagrant/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc).
```

PR #2811, opened 2014-07-10
Updates to cleanup variables and fix a Zendesk bug introduced in previou...

### LINKED ISSUE #2810: [Global] Cleanup variables
Unify the way the variables are set so that it's easier to pass them along to the `theme()` function for templating and fix a bug with Zendesk introduced in prior pull request update.

weerd:
...s update.

Resolves #2810

@DFurnes @angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.module
+++ b/lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.module
@@ -133,21 +133,9 @@ function dosomething_campaign_group_preprocess_node(&$vars) {
    
     // Include Zendesk Form
     $vars['zendesk_form'] = drupal_get_form('dosomething_zendesk_form', $node);
-    
+
   }
 }
-
-/**
- * Implements hook_node_view().
- */
-// function dosomething_campaign_group_node_view($node, $view_mode, $langcode) {
-//   if ($node->type == 'campaign_group' && $view_mode == 'full') {
-//     // Add Zendesk form variable:
-//     if (module_exists('dosomething_zendesk')) {
-//       $node->content['zendesk_form'] = drupal_get_form('dosomething_zendesk_form', $node);
-//     }
-//   }
-// }
 
 /**
  * Returns a Campaign Group parent nid for a given node $nid.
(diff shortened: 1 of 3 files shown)


DFurnes:
:sun_with_face:

DFurnes:
:dancers:

weerd:
huzzah! :goat:

lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.module
```diff
@@ -130,22 +130,14 @@ function dosomething_campaign_group_preprocess_node(&$vars) {
     if (!empty($modal_items)) {
       $vars['modals'] = theme('modal_links', array('modals' => $modal_items));
     }
-    
-  }
-}
+   
+    // Include Zendesk Form
+    $vars['zendesk_form'] = drupal_get_form('dosomething_zendesk_form', $node);
```
aaronschachter, line 135: The zendesk form is in the `hook_node_view` because adding forms via preprocess can lead to irregular messaging (you submit a form and you don't get a message and then refresh the page and then you see a message).  

If this works ok but please test.

aaronschachter, line 135: Isn't this now inconsistent with `dosomething_campaign` where we are setting the `zendesk_form` variable in `preprocess_node` ? Perhaps we can look into this next week when i'm back in but moving this out of `hook_node_view` i think will lead to the irregular messaging...

weerd:
Hmm, interesting... ok. I can let @barryclark know that we should test this thoroughly on Grouped Campaigns. I talked this through with @angaither before I implemented so I could get some input. 

It was previously inconsistent with other nodes in terms of where the `zendesk_form` value was being stored (within `$vars['content']['zendesk_form']` as opposed to `$vars['zendesk_form']` like everywhere else). This made it difficult to unify the way values were passed across to the template partial when using the `theme()` function.

If there is another/better way to unify where the `zendesk_form` stuff is stored, I'm down :)

weerd:
I just tested on Staging site and it seemed to work fine :goat:

barryclark:
Looks like the test is purely the zendesk modal on the campaign group page, not any of the individual campaigns, right?

weerd:
Yep yep! Only Campaign Group page. All the other scenarios are same as before and code didn't change ;)

barryclark:
Worked fine for me on staging too... and confirmed that both @weerd and my messages made it into Zendesk.

PR #2812, opened 2014-07-10
Updating markup for new Neue modal syntax.

### LINKED ISSUE #1078: modals - refine modal experience on iOS
Ideally we can make some tweaks to modals on iOS so that we can limit some of the user's ability to scroll outside of the modal.

Awkwardness can be seen by activating nav and/or going through site reg and scrolling down.

### LINKED ISSUE #1673: [IE Bug] Add background fallback for modal lightbox.
screen shot 2014-04-07 at 2 40 07 pm

rgba() not supported in IE8 so need to add a fallback. Either a transparent repeatable png, or a solid background color.

DFurnes:
Updating modal markup for DoSomething/neue#320. Adds analytics code that used to exist in Neue into the DS repo.

Fixes #1673. Fixes #1078.

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/modal-links.tpl.php
+++ b/lib/modules/dosomething/dosomething_helpers/modal-links.tpl.php
@@ -1,21 +1,21 @@
 <ul class="modal-links">
   <?php if (isset($modals['faq'])): ?>
-    <li><a href="#modal-faq" class="js-modal-link">Check out our FAQs</a></li>
+    <li><a href="#" data-modal-href="#modal-faq">Check out our FAQs</a></li>
   <?php endif; ?>
 
   <?php if (isset($modals['more_facts'])): ?>
-    <li><a href="#modal-facts" class="js-modal-link">Learn more about <?php print $modals['issue']; ?></a></li>
+    <li><a href="#" data-modal-href="#modal-facts">Learn more about <?php print $modals['issue']; ?></a></li>
   <?php endif; ?>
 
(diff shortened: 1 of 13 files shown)


lib/modules/dosomething/dosomething_helpers/modal-links.tpl.php
```diff
@@ -1,21 +1,21 @@
 <ul class="modal-links">
   <?php if (isset($modals['faq'])): ?>
-    <li><a href="#modal-faq" class="js-modal-link">Check out our FAQs</a></li>
+    <li><a href="#" data-modal-href="#modal-faq">Check out our FAQs</a></li>
```
tjrosario, line 3: What is the reason for using a data attribute for the 'href' ?

DFurnes:
Tested the new modals with a bunch of random browser versions (including IE8-IE11) on BrowserStack and everything seems to work fine.

lib/modules/dosomething/dosomething_helpers/modal-links.tpl.php
DFurnes, line 3: The idea is that we generally want to have a canonical link (i.e. "Login" button goes to `/user/login` when JS is disabled), and that's what the `href` should point to.

In hindsight, I don't know if that's ever going to happen for a bunch of these links on the campaign page so maybe we should look into something more like the old syntax as an alternative way of linking to a modal.

weerd:
Looks good :+1: :tractor:

PR #2818, opened 2014-07-14
Zendesk Form variables

aaronschachter:
Stores the Form's' header and label text values as variables.

Also fixes lots of undefined errors for campaigns and/or nodes that do not have the `$vars` set.  cough, @weerd 
## Screenshots

zendesk_form

zendesk_admin

### DIFF
--- a/lib/modules/dosomething/dosomething_zendesk/dosomething_zendesk.admin.inc
+++ b/lib/modules/dosomething/dosomething_zendesk/dosomething_zendesk.admin.inc
@@ -45,6 +45,26 @@ function dosomething_zendesk_config_form($form, &$form_state)  {
     '#description' => t('Do not share this token publicly.'),
     '#default_value' => variable_get('dosomething_zendesk_token', ''),
   );
+  $form['copy'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Form copy'),
+    '#collapsible' => TRUE,
+    '#collapsed' => TRUE,
+  );
+  $form['copy']['dosomething_zendesk_form_header'] = array(
+    '#type' => 'textarea',
+    '#title' => t('Form Header'),
+    '#description' => t("Copy displayed in the form header."),
+    '#required' => TRUE,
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_zendesk/dosomething_zendesk.admin.inc
```diff
@@ -45,6 +45,26 @@ function dosomething_zendesk_config_form($form, &$form_state)  {
...
+    '#collapsed' => TRUE,
+  );
+  $form['copy']['dosomething_zendesk_form_header'] = array(
+    '#type' => 'textarea',
+    '#title' => t('Form Header'),
+    '#description' => t("Copy displayed in the form header."),
+    '#required' => TRUE,
+    '#default_value' => variable_get('dosomething_zendesk_form_header'),
+  );
+  $form['copy']['dosomething_zendesk_form_body_label'] = array(
```
sergiitk, line 61: It's not in coding standards, but, in my opinion, aligning arrows is great for readability.
image
vs
image

I use [Alignment](https://sublime.wbond.net/packages/Alignment) plugin (not, 287K installs) for that.
But it's up to you, just ignore that if you don't like it.

lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
```diff
@@ -144,12 +144,29 @@ function paraneue_dosomething_preprocess_node(&$vars) {
...
+  foreach ($info_var_names as $variable) {
+    // If the vars is set for this variable name:
+    if (isset($vars[$variable])) {
+      // Store its value.
+      $value = $vars[$variable];
+    }
+    else {
+      $value = NULL;
+    }
+    $info_bar_vars[$variable] = $value;
```
sergiitk: As an alternative:

``` php
$info_bar_vars[$variable] = NULL;
if (!empty($vars[$variable])) {
  $info_bar_vars[$variable] = $vars[$variable];
}
```

sergiitk:
Functionality: :+1: 
image

PR #2823, opened 2014-07-14
Metatag defaults for abstract/description

aaronschachter:
Setting the abstract/description in the homepage node is not setting the metatags when viewing the homepage. Didn't realize that this stuff was being reverted upon every deployment when I marked https://github.com/DoSomething/dosomething/issues/2343#issuecomment-47233342 as done.

I think we added Twitter Cards module after we initially created this module, hence all the new exportables.


sergiitk:
:+1:

PR #2824, opened 2014-07-14
Intl: Allow overriding header logo.

sergiitk:
#### What's this PR do?

Adds an option to override header logo.
image
- Open paraneue_dosomething [theme settings](http://localhost:8888/admin/appearance/settings/paraneue_dosomething)
- Upload a png to LOGO OVERRIDE → File, save form.
  Logo examples can be found [here](https://jira.dosomething.org/browse/DOS-11)
- Open frontpage and see the logo changed
- Open theme settings, click Remove on the left of LOGO OVERRIDE → File
- Open frontpage and make sure it's default logo

Fixes [#DOS-11](https://jira.dosomething.org/browse/DOS-11)

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
@@ -34,10 +34,17 @@ function paraneue_dosomething_preprocess_page(&$vars) {
     $modifier_classes = 'black floating';
   }
 
+  $logo = NEUE_ASSET_PATH . '/assets/images/ds-logo.png';
+  if (theme_get_setting('header_logo_override')) {
+    $file = file_load(theme_get_setting('header_logo_override'));
+    if (!empty($file->uri)) {
+      $logo = file_create_url($file->uri);
+    }
+  }
   $navigation_vars = array(
     'base_path'        => $vars['base_path'],
     'modifier_classes' => $modifier_classes,
-    'logo'             => NEUE_ASSET_PATH . '/assets/images/ds-logo.png',
+    'logo'             => $logo,
     'front_page'       => $vars['front_page'],
     'logged_in'        => $vars['logged_in'],
     'search_box'       => $vars['search_box'],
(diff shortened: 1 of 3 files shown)


aaronschachter:
Is there a reason in particular to not just re-work the theme to use Drupal's built in logo stuff?

screen shot 2014-07-14 at 5 31 51 pm

aaronschachter:
I get that this may get us there quicker, but it feels like duplication of functionality already built in... unless there's something else going on with the logo that i'm not aware of...

sergiitk:
The reason why I decided to use custom field is that is the asset builder. So the final link to the logo becomes https://dosomething-a.akamaihd.net/sites/default/files/assets/v0.2.134/bower_components/neue/assets/images/ds-logo.png
For fact, at the moment we're not using `logo` feature at all. I even checked a diff of the frontpage with and without `features[] = logo` line.
Maybe there's a way to make it work with `features[] = logo`, but I just changed what I found we're using at the moment.

aaronschachter:
Thanks for clarifying, @sergii-tkachenko. I still would prefer that we use the core logo uploader, only to just have less custom code in the site.  @DFurnes @weerd -- Is it critical that the logo uses the `ASSET_PATH`?  Is there any reason to not just rework this to use the logo file field provided by Drupal core?  I'd think it would be a relatively lift to do so...

sergiitk:
@aaronschachter I agree with you. I'll look into implementation using Drupal native logo feature. 
However, I'll clarify priorities on the international standup. It might be critical to have this functionality right now.

sergiitk:
@aaronschachter, @barryclark confirms it is a priority. We can refactor it afterwards. Do you have comments on current code change?

aaronschachter:
No, code looks okay... I just question whether or not it's a really quick change to just use the default functionality - but it's your call. I'd like us to definitely refactor sooner than later, this will be easy to forget about. Do you want to create an issue to refactor and re-assign to yourself so the convo doesn't get lost?

sergiitk:
@aaronschachter  Yes, I will create it.

PR #2828, opened 2014-07-15
Campaign  pitch seo update

weerd:
@DFurnes 

Update to add content for Pitch page and adjusting styles. The content is added as a separate partial that's not included for now, so the code exists in the repo and can be added at a later date :sun_with_face:

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/bower.json
+++ b/lib/themes/dosomething/paraneue_dosomething/bower.json
@@ -20,7 +20,7 @@
     "tests"
   ],
   "dependencies": {
-    "neue": "DoSomething/neue#~4.4.1",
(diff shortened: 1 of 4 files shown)


lib/themes/dosomething/paraneue_dosomething/bower.json
```diff
@@ -20,7 +20,7 @@
     "tests"
   ],
   "dependencies": {
-    "neue": "DoSomething/neue#~4.3.1",
+    "neue": "DoSomething/neue#~4.4.1",
```
DFurnes: Should this be `4.4.0`?

weerd: oop!

DFurnes:
:sailboat:!

Issue #2749, opened 2014-07-02
Static Content gallery: Theme based on Gallery Style field value


aaronschachter:
Refs #2646 - blocked by #2748

`dosomething_static_content_preprocess_gallery_vars` is now adding a `style` variable for each Gallery.

`default`: Should be rendered as is.

`2col`: The 2 column style as per the Closed Campaign

`3col_short`: The 3 column style per the Press gallery

These values are defined on the `field_gallery_style` content type.

barryclark:
For my own reference: Need to be able to use the 3 different gallery types used here https://www.dosomething.org/campaigns/comeback-clothes

barryclark:
Shooting you over this one, @tjrosario. @aaronschachter will connect with you on the details of it. cc @mikefantini

mikefantini:
@tjrosario when trying to add the 3 column short gallery to static content, the gallery isn't pulling in images that are uploaded as the thumbnail crop.

see http://staging.beta.dosomething.org/about/fantiniz-partnerz as an example

using this issue since it's still open to report the bug rather than creating a new issue

let me know if you need any other info

tjrosario:
@aaronschachter @angaither the `image_url` field is returning blank for the gallery photos on static pages.  Would one of you be able to take a look or point me in the direction for a fix?

angaither:
@tjrosario which field are you trying to use? what the nid?

tjrosario:
![image](https://cloud.githubusercontent.com/assets/1566749/3679516/185873f6-12a5-11e4-9e05-371a4319e235.png)

NID is 1292

angaither:
That image only has something in the `field_thumbnail` is that what you're trying to grab?

aaronschachter:
For the "3column short" gallery - it is supposed to display the Thumbnail field. So, we need to alter the prrocess, right now it's hardcoded to use Square:

https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_static_content/dosomething_static_content.module#L109

If the gallery style is default, keep this code as is, passing `'square'` as the `'ratio'` parameter to `dosomething_image_get_themed_image`.

```
    $vars['galleries'][$i]['items'][$a]['image'] = dosomething_image_get_themed_image($gallery_item['field_gallery_image']['#items'][0]['target_id'], 'square', '300x300');
```

If NOT default, 'thumb'  should be passed:

```
[...]

PR #2829, opened 2014-07-15
added gallery style helper to work with current gallery patterns

tjrosario:
References #2749


lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign--closed.tpl.php
```diff
@@ -287,8 +287,25 @@
     </div>
   </section>
 
-  <?php if ($info_bar): ?>
-    <?php print $info_bar; ?>
-  <?php endif; ?>
+  <footer class="info-bar">
```
aaronschachter, line 290: this shouldn't be in here, the `$info_bar` is new code from #2807. If you rebase from dev, this change should not be present. Check out https://github.com/DoSomething/dosomething/wiki/Contributing#using-git-rebase-to-keep-up-to-date

aaronschachter:
Looks good here except for comment about out of date template.

PR #2832, opened 2014-07-15
Editor user edit

aaronschachter:
## Overview

Closes https://jira.dosomething.org/browse/DOS-58 
- Enables User profile pictures
- Hides picture upload from all non-staff

Closes https://jira.dosomething.org/browse/DOS-59
- Allows editors ability to edit a User's First Name, Last Name, City, State, and Profile Picture by creating a new page callback and form at `user/*/info`
  - I chose to implement this way vs. unsetting a bunch of form elements within the `dosomething_user_form_alter`, because the only way an editor could even access the `user/*/edit` form would be to grant permission to `administer users` -- which also allows them to delete users if they went to the URL directly.  
  - This is also in the long run potentially less of a headache to maintain, since we'd always have to unset profile form elements in the `form_alter` as we add new ones.
## Screenshots

screen shot 2014-07-15 at 3 57 52 pm


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -757,3 +777,112 @@ function dosomething_user_non_staff_redirect() {
...
+  }
+  $form['info']['city'] = array(
+    '#title' => "City",
+    '#type' => 'textfield',
+    '#default_value' => $city,
+  );
+  $last_name = dosomething_user_get_field('field_last_name', $account);
+  $form['info']['state'] = array(
+    '#title' => "State",
+    '#type' => 'textfield',
```
angaither, line 824: Might be nice to make this a dropdown using the addressfield list of 'administrative area' choices if it's easy to do

aaronschachter, line 824: Yeah, I had trouble finding it in the `addressfield` code.  Figured this was an easy way to at least get functionality in, and could always add back in.

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -757,3 +777,112 @@ function dosomething_user_non_staff_redirect() {
...
+      ),
+    ),
+  );
+  // Store the current address.
+  $edit['field_address'] = $account->field_address;
+  // Overwrite City/State if values are present.
+  $edit['field_address'][LANGUAGE_NONE][0]['locality'] = $values['city'];
+  $edit['field_address'][LANGUAGE_NONE][0]['administrative_area'] = $values['state'];
+  // Hardcoded to 'merica for now.
+  $edit['field_address'][LANGUAGE_NONE][0]['country'] = 'US';
```
angaither, line 880: Before saving this, it should probably be run through the valid address checker

aaronschachter, line 880: This is only to set City/State though... not a full address.  is it possible to verify for just city/state combo?

lib/modules/dosomething/dosomething_user/dosomething_user.module
angaither, line 824: Might be worth a shot to use the custom address field that's being used on the signup forms?
That way the validation would happen already as well

lib/modules/dosomething/dosomething_user/dosomething_user.module
angaither, line 880: Hm. Hm. I think you need the zip at least. Just makes me a little nervous to have this without validation... though if we aren't using it to mail things I guess it doesn't matter....

angaither:
Anyways, the code is all fine :+1:

lib/modules/dosomething/dosomething_user/dosomething_user.module
aaronschachter, line 880: Yeah... the sole purpose of adding these fields is to display the City / State here: https://jira.dosomething.org/browse/DOS-25

I didn't think it made sense to store this stuff on the campaign as separate fields, since it's actually related to a real user (and we already have the fields for it). What do you think?

angaither, line 880: Totally makes sense to put it on the user, do we only get the city/state from them? 
Is this from the google form?

aaronschachter, line 880: Yeah I'm assuming editors will be adding this info from Google Form submissions if we don't have an address stored for them already. @mikefantini sound right?

mikefantini, line 880: these will be super hands on for the forseeable future so we can get full address and have it validated properly

PR #2833, opened 2014-07-15
Campaign tests

DFurnes:
Adds campaign action page tests to test suite.
JIRA issues: FT-4, FT-5

screen shot 2014-07-15 at 3 19 16 pm


scripts/tests/before_tests.sh
```diff
@@ -5,36 +5,51 @@
...
 
-## Make fresh test accounts...
+echo "Making fresh test accounts..."
 drush_create_test_user QA_TEST_ACCOUNT
 drush_create_test_user QA_TEST_CAMPAIGN_SIGNUP_EXISTING
+drush_create_test_user QA_TEST_CAMPAIGN_ACTION
 
+echo "Signing action page test account up for test campaign..."
+example_campaign_nid=1261
+drush_signup_user $example_campaign_nid QA_TEST_CAMPAIGN_ACTION@example.com
```
angaither, line 55: way to bash

angaither:
:+1: value added

angaither:
There should be a cleanup task as well to remove these users after, otherwise the test will fail the second time around

DFurnes:
Right now, I'm removing the users in the `before_tests.sh` script, since I don't think I can guarantee that `after_tests.sh` will run if the user is impatient and cancels the test before it finishes. Might make sense to try to run cleanup both before and after?

PR #2835, opened 2014-07-16
Drush command to create campaigns from JSON

### LINKED ISSUE #2793: Testing: Create a campaign node from json file
Drush command which will read a JSON file, consisting of Campaign properties (per #2732) and values, and creates the node.

The capser.js test will also read from this file and test values against it for various fields.

aaronschachter:
Fixes #2793

@sergii-tkachenko Would you mind reviewing?

cc @DFurnes 

Adds a drush command `drush campaign-create` to create a campaign node from a given JSON file, and adds said JSON file into the `tests/campaign` directory.

```
vagrant@dev:/var/www/vagrant/html$ drush campaign-create ../tests/campaign/campaign.json
Created node nid 1292.
```

The function and JSON files are minimal at the moment, just to get this up and running.  The idea here is that the Capser JS tests will read from the JSON file as well when checking for the presence of various fields, so we're not relying on content within the database as is, and hardcoding tests based on that content.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.drush.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.drush.inc
@@ -7,7 +7,10 @@ function dosomething_campaign_drush_command() {
   $items = array();
   // Name of the drush command.
   $items['create-campaign'] = array(
-    'description' => 'Creates a campaign node',
+    'description' => 'Creates a campaign node from given JSON file.',
+    'arguments' => array(
+      'filename' => 'Name of the file to read.'
+    ),
     'callback' => 'dosomething_campaign_drush_create_campaign',
   );
   return $items;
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.drush.inc
```diff
@@ -0,0 +1,31 @@
+<?php
+
+/**
+ * Implementation of hook_drush_command().
+ */
+function dosomething_campaign_drush_command() {
+  $items = array();
+  // Name of the drush command.
+  $items['create-campaign'] = array(
```
sergiitk: I'd better name it `campaign-create`, following pattern `user-create`, `pantheon-site-create`, etc

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.drush.inc
```diff
@@ -0,0 +1,31 @@
...
+    'description' => 'Creates a campaign node from given JSON file.',
+    'arguments' => array(
+      'filename' => 'Name of the file to read.'
+    ),
+    'callback' => 'dosomething_campaign_drush_create_campaign',
```
sergiitk: Add `required-arguments => TRUE,`

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.drush.inc
```diff
@@ -0,0 +1,31 @@
...
+  );
```
sergiitk, line 19: Consider adding

``` php
'examples' => array(
  'drush create-campaign ../tests/campaign/campaign.json' => 'Create test campaign.',
),
```

sergiitk:
:+1: Awesome!

PR #2836, opened 2014-07-16
Campaign created fields

aaronschachter:

Adds a `created_by` Entity Reference and a `created_copy` for subtasks in https://jira.dosomething.org/browse/DOS-25

Also updates uploaded file status to permanent and adds `file_usage` record when upload is present.


sergiitk:
I'm not sure if characters count is relevant for Created By field.
Is it something assigned automatically?

image

aaronschachter:
Yeah... it's definitely not relevant but happens in all of the EntityReference fields.  Would be a separate issue outside of the scope of this PR if we want to fix (but the node form is never facing... so not a biggie).

sergiitk:
:+1:

PR #2837, opened 2014-07-16
Adding some additional checks for tables and fields

blisteringherb:
- Checking for table existence before selecting from it in install
- Checking for field existence before calling value() on it

@aaronschachter 
@mshmsh5000


mshmsh5000:
Seems pretty conservative to me, but I haven't tested.

aaronschachter:
:+1: all kosher

PR #2809, opened 2014-07-10
Refactoring ds build script for multisite deployments

blisteringherb:
#### What's this PR do?
- Updates the ds build script to take an '--intl' flag to build on each of the countries' sites  
- Creates a build helper function to abstract the common tasks between single and multisite builds  
- Updates settings.php to use the hostname 'solr' for the solr url instead of an IP address  
- Moves all capistrano tasks to one file and removes extra international includes file
- Adding solr hostname setup to salt for dev environments
- Adding new countries to aliases file
#### Where should the reviewer start?

The most significant changes are in bin/ds in the `build`, `install` and `build_helper` functions.  There are also changes that affect how the arguments for the ds build script are parsed.
#### How should this be manually tested?
- Pull down the code, making sure both sites/dosomething and ../dosomething-mount have the updated intl-deploy code, vagrant destroy your current environment, vagrant up to make sure all Salt changes have happened.
- Run `ds build`, `ds build --intl --install`, `ds pull stage.$AFFILIATE --ds --files`, etc.
#### Any background context you want to provide?

The previous ds build script didn't take a multisite installation into account, this PR addresses that.
#### What are the relevant tickets?

Fixes #DOS-18

cc: @angaither @sergii-tkachenko - please take a look if you're available

PR #2842, opened 2014-07-17
Create a function to determine if current site is an international affiliate.

sergiitk:
#### What's this PR do?
- Creates a variable in settings.php to determine if current site is an international affiliate
- Create a helper function returning if current site is an international affiliate
#### Todo

Set the variable to TRUE for international affiliate, e.g. `in_array($hostname, $affiliates);`
Reference: #2809 

Fixes [#DOS-76](https://jira.dosomething.org/browse/DOS-76)

### DIFF
--- a/config/settings.php
+++ b/config/settings.php
@@ -119,3 +119,12 @@ $conf['locale_custom_strings_en'][''] = array(
  );
 
 $conf['optimizely_id'] = '747623297';
+
+/**
+ * DoSomething specific settings.
+ */
+// Determines whether current site is an international affiliate.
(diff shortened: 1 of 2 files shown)


sergiitk:
@tjrosario  You can use `dosomething_settings_is_affiliate` to determine if the site is an international affiliate. Since #2809 is still a blocker for this task (there are no `$affiliates` yet), for now this function will return only `FALSE`.
However, you can add this condition after you finished styling/theming. So your code will not be shown on actual dosomething.org.

aaronschachter:
:+1:

PR #2844, opened 2014-07-17
Reportback API endpoint

aaronschachter:
Right now, this will only work for when the logged in user (via `api/v1/user/login`) is an administrator. 

cc @jonuy 

 This PR opens the door to building out SMS Reportbacks within the DS mData rResponder repo, and to eventually move all SMS Reportback functionality out of the Drupal app.  Instead the mData repo would post reportback data to the `reportback/save` endpoint.  The mData app would just login as an administrator (we could create an account like `machines@dosomething.org`, or build out Token based authentication.

Will update the API wiki when this is merged:

**GET** `www.dosomething.org/api/v1/reportback/:rbid` -- Returns reportback content for a given reportback rbid.

**POST** `www.dosomething.org/api/v1/reportback/save` -- Creates or updates a reportback based on given uid and nid in the data:
- nid
- uid
- quantity
- why_participated
- file_url -- an Image URL to save as a file and associate with the reportback

**Note: This will not work yet with campaigns that require custom reportback fields.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -738,3 +738,34 @@ function dosomething_reportback_get_noun_and_verb($nid) {
 
   return $results;
 }
+
+/**
+ * Implementation of hook_services_resources().
+ */
+function dosomething_reportback_services_resources() {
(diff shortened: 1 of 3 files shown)


jonuy:
Cool! Code seems fine.

PR #2848, opened 2014-07-17
Pulls image lazy-loading for tiles out of the campaign finder.

DFurnes:
Tiles with an [Unveil](http://luis-almeida.github.io/unveil/) lazy-loaded `data-src` image should now work properly when included outside of the campaign finder. (Fixes issue with international sites.)


DFurnes:
Long-term, we should probably wrap this in with whatever we decide on for a responsive image solution.

blisteringherb:
:+1:

PR #2851, opened 2014-07-18
Campaign finder  link fix and cleanup

### LINKED ISSUE #2850: [Campaign Finder] Update to fix missing links on default homepage finder tiles
After home page refactor, the campaign tiles for the initial state on the finder are missing links to their respective campaigns.

weerd:
Resolves #2850

@DFurnes Cleanup for JS and some CSS.

@aaronschachter Update to add the link href correctly based on what we discussed.

### DIFF
--- a/lib/modules/dosomething/dosomething_home/dosomething_home.module
+++ b/lib/modules/dosomething/dosomething_home/dosomething_home.module
@@ -47,12 +47,9 @@ function dosomething_home_preprocess_node(&$vars) {
       $tiles['image'] = dosomething_image_get_themed_image_url($image_nid, 'square', '400x400');
     }
 
-    $tiles['link'] = l('<span>' . $title . '</span>', 'node/' . $nid, array(
-      'attributes' => array(
-        'class' => array('full-link')
-      ),
-      'html' => TRUE
-    ));
+
+
+    $tiles['link'] = drupal_get_path_alias('node/' . $nid);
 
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_home/dosomething_home.module
```diff
@@ -47,12 +47,9 @@ function dosomething_home_preprocess_node(&$vars) {
...
 
-    $tiles['link'] = l('<span>' . $title . '</span>', 'node/' . $nid, array(
-      'attributes' => array(
-        'class' => array('full-link')
-      ),
-      'html' => TRUE
-    ));
+
+
+    $tiles['link'] = drupal_get_path_alias('node/' . $nid);
```
aaronschachter: looks fine. i suppose could call it `'url'` if `link` suggests its markup... but up to you.

DFurnes: +1 renaming to `'url'`.

DFurnes:
Seems fine. :nail_care:

lib/modules/dosomething/dosomething_home/dosomething_home.module
weerd: good note. I'm updating it to this:

```
$tiles['path'] = 'node/' . $nid;
$tiles['pretty_path'] = drupal_get_path_alias($tiles['path']);
```

aaronschachter:
sure :+1:

PR #2853, opened 2014-07-18
Campaign seo  adding content

weerd:
@DFurnes quick check please? Just content updates on includes that will be used only via optimizely.

@Bladt The markup and content for the 3 different campaigns to try the SEO stuff on optimizely are included.


DFurnes:
Looks good. Might be good to just add a comment to each template mentioning that these aren't expected to be used by Drupal so we don't get confused later on.

PR #2856, opened 2014-07-21
Remove bower copy task that was causing trouble with old configs.

DFurnes:
Accidentally deleted branch for #2854 and GitHub wouldn't let me re-open the pull request. So here we are.


weerd:
:+1:

PR #2857, opened 2014-07-21
Member Faming modal

aaronschachter:
Adds a Creator property to the Campaign class with relevant Creator data, and creates a new tpl to display.

cc @weerd 
- Closes https://jira.dosomething.org/browse/DOS-60
- Unblocks https://jira.dosomething.org/browse/DOS-61

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -347,21 +347,6 @@ function dosomething_campaign_get_campaign_status($node) {
 }
 
 /**
- * For given node entity wrapper, return its field_variables as an array.
- */
-function dosomething_campaign_get_field_variables($wrapper) {
-  if (!isset($wrapper->field_variables)) { return array(); }
-
-  $vars = array();
-  foreach ($wrapper->field_variables->value() as $variable) {
-    $name = $variable->field_variable_name[LANGUAGE_NONE][0]['value'];
-    $value = $variable->field_variable_value[LANGUAGE_NONE][0]['value'];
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
```diff
@@ -21,6 +21,16 @@
  *   - type (string). 'campaign' | 'sms_game'
  *
  *   - call_to_action (string). Plain text.
+ *   - creator (array)
+ *     - uid (int)
+ *     - first_name (string). Plain text
+ *     - last_initial (string). Plain text.
```
weerd, line 27: Should this be `last_name` and then in theme, we can just use a quick PHP function (like substr()) to grab the first letter? Feel like this info should be available, and not just the initial cause that's what the design calls for? Not sure though. Thoughts?

aaronschachter, line 27: https://jira.dosomething.org/browse/DOS-25 calls for Last Initial. I've built this so this Creator data is available from the Content API as well, which is why I elected to only display Last Initial and not Last Name to the public.

weerd, line 27: Gotcha! Ok, makes sense. Good to know :dancer:

aaronschachter, line 27: This will be easy enough to swap out later with last name if we decide to display it, but yeah the design does not match the request in DOS-25

weerd:
Cooooool. Looks good :+1:

Issue #2375, opened 2014-05-30
[Campaign template] Campaign header and footer templates


aaronschachter:
Templates for:
- [x] Campaign header: Headings (Title, CTA, Scholarship)
- [x] Campaign header: Sponsors
- [x] Campaign header; Scholarship
- [x] Campaign footer - Required for #2367 

Strip out the Campaign node's headers and footers into themeable functions, so we can easily place them on the Campaign Run node template.  Should drastically clean up node tpl files for Campaign, Campaign Pitch, and SMS Game tpl files as well.

aaronschachter:
Discussed with @weerd today -- there's so many various permutations of the actual `<header>` markup that it's too complex to move into its own themeable function.  Instead we'll create snippets of code to make the template much more manageable, e.g.

```
<header role="banner" class="-hero <?php print $classes; ?>">
  <div class="wrapper">
    <?php print $campaign_headings; ?>
    <?php if (isset($signup_button)): ?>
      <div class="__signup">
      <?php print render($signup_button); ?>
      <?php print $scholarship; ?>
      </div>
    <?php endif; ?>
    <?php print $campaign_sponsors; ?>
  </div>
</header>
```

PR #2860, opened 2014-07-21
Campaign headings template

aaronschachter:
Refs #2375: Adds a `campaign-heads` tpl file into `paranaue_dosomething` to make campaign templating better.

@weerd Can you please review?


weerd:
Awesomeness! :+1: :tropical_drink:

PR #2861, opened 2014-07-21
Added a pre-commit hook to check for common mistakes.

angaither:
## What does this PR do?
- adds in pre-commit hooks that fire before committing
- adds the symlinked git hook into the `ds build` task
## What are we checking for?
- git conflicts to be committed
- drupal debug code (`dsm`, `dpm`, `krumo`)
## How do I test this?
- pull down this code
- add some `dpm`s
- `git commit`
- you will see this error

```
Error: git pre-commit hook forbids committing "dpm" to hooks/pre-commit
--------------
To commit anyway, use --no-verify
```

The code was generously provided by [this man](https://github.com/bobgilmore/githooks)

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -223,6 +223,9 @@ function symlinks {
   echo 'Linking themes'
   rm -rf "$WEB_PATH/profiles/dosomething/themes/dosomething/paraneue_dosomething"
   ln -s "$LIB_PATH/themes/dosomething/paraneue_dosomething" "$WEB_PATH/profiles/dosomething/themes/dosomething/paraneue_dosomething"
+
+  echo 'Add git pre-commit hook'
+  ln -s "../../hooks/pre-commit" ".git/hooks/pre-commit"
 }
 
 #=== FUNCTION ==================================================================
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,0 +1,37 @@
+#!/usr/bin/env ruby
+
+
+user = ENV["USER"]
+error_found = false
+
+# "Forbidden" regular expressions
+FORBIDDEN_STRINGS = [
+  />>>>>>/,    # Git conflict markers
+  /<<<<<</,    # ''
+  /dpm/,         # drupal debug statements
+  /krumo/,
+  /dsm/,
+]
+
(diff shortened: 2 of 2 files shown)


aaronschachter:
Do you need to rebuild?  I've pulled down and testing from my `dosomething-mount` but able to commit with `dpm`'s in my code...

This also looks interesting, as far as running Coder in pre-commit-hook https://www.drupal.org/node/2054259#comment-7970545

aaronschachter:
Ran `ds symlinks` . but still able to commit `dsm`s...

angaither:
From inside `dosomething-mount` 
if you run `ls .git/hooks/` does 'pre-commit' show up?
if not it means the symlinks didn't work.
Make sure to run bin/ds symlinks inside of `/var/www/vagrant` then you should see the pre-commit file and it will give you the warning before committing a `dsm`

aaronschachter:
Ah, the trick was I ran `ds symlinks` and not `bin/ds symlinks`.  Why is there a difference?  If you run just `ds symlinks` the `Add git pre-commit hook` does not appear.

What are the steps to get everyone to do this? Does everyone need to run `bin/ds symlinks` or will it happen upon a new `ds build` ?  Code looks good :+1: just need clear steps for how to get this working for everyone.

angaither:
It'll happen once people do a `ds build` or manually run `bin/ds symlinks`

aaronschachter:
Not a biggie but FYI when I run `ds pull stage`, I get this output:

```
Linking modules
Linking themes
Add git pre-commit hook
ln: failed to create symbolic link `.git/hooks/pre-commit': No such file or directory
```

Ds pull stage calls `ds symlinks` to recreate the sylinks once files are pulled down from stage.

angaither:
Ah, ew okay, I can add it in at a different point, or just have people manually add the symlink, since it only needs to happen one time. Thoughts?

aaronschachter:
Hmm, I wonder if this is could be a reason that pulling files from stage seems to be taking longer. Maybe we add it once in `ds build` but keep it out of `ds symlinks`?  Or check if the symlink exists before creating?

PR #2865, opened 2014-07-22
Campaign visual tests

DFurnes:
Adds visual regression tests for campaign page. Using [PhantomCSS](https://github.com/Huddle/PhantomCSS), the contents of selector will be saved to a file and compared to an image which is committed to the repository. 

If the contents of the selector no longer match the expected image, the test will fail. To approve the change, the expected image can be replaced so that the test passes (and also indicate the change for easy review in the pull request)! Easy!
## Highlights:

Most important changes are in `common.js` (PhantomCSS configuration) and `action.js` (splitting tests into visual and functional sections, and adding in PhantomCSS tests). 

Additional minor changes made while testing were @d3baae1 (removing unused "size" modifier classes) and @29f1398 (triggering ImageUploader on DOM ready, since we no longer need to wait for Modal event as of Neue 4.4).
## Test Output

screen shot 2014-07-22 at 3 50 26 pm
## Example Test Failure

Test output:
visual-error-example

Visual diff showing change, generated in `tests/visual/failures`:
step1 16 fail


barryclark:
:sparkling_heart: :sparkling_heart: :sparkling_heart: :sparkling_heart:

scripts/tests/vendor/phantomcss.js
```diff
@@ -0,0 +1,580 @@
+/*
+James Cryer / Huddle / 2014
```
angaither: oh cool, did we hire him?

scripts/tests/vendor/ResembleJs/README.md
```diff
@@ -0,0 +1,78 @@
+Resemble.js
+==========
+
+Analyse and compare images with Javascript and HTML5. [Resemble.js Demo](http://huddle.github.com/Resemble.js/)
```
angaither: is there a way not to track this stuff in our repo?

scripts/tests/vendor/phantomcss.js
DFurnes: deal with it

PR #2866, opened 2014-07-22
DS Build: Force revert all Features

aaronschachter:
After discussing with the team, it sounds like we'll always want to revert when our Features are in "Needs Review" state -- so let's simplify this by just force reverting all features, and see if this gets rid of our "Table already exists" issues.


angaither:
:+1:

PR #2875, opened 2014-07-24
Static Content Gallery -- Use Thumbnail field when Gallery Style != default

aaronschachter:
@tjrosario Can you pls review? cc @angaither 

per #2749 the gallery needs to display the Thumbnail field on the Image node instead of the Square field if the Gallery Style is anything than default.

This PR changes `dosomething_static_content_preprocess_gallery` to use the correct Image field based on the the Gallery Style, but there are CSS issues when the `-duo` class is used.

It also adds a `paraneue_dosomething_preprocess_node_field_gallery` function to keep Neue specific preprocessing in the theme layer, and breaks out Campaign node specific preprocessing into `paraneue_dosomething_preprocess_node_campaign`.


aaronschachter:
CSS issues 

(as is -- using -duo):
css_as_is

Removing width:

screen shot 2014-07-23 at 9 23 12 pm

angaither:
:+1:

PR #2880, opened 2014-07-24
Campaign finder  bug

weerd:
@DFurnes This resolves the duplicating results issue that started occurring after the Campaign Finder refactoring. While this solves the issue, I'm not sure why the $container wasn't being emptied properly before? Please take a gander and let me know what you think!

cc @blisteringherb


DFurnes:
man javascript am i right

PR #2882, opened 2014-07-25
Hotfix for DOS-103

aaronschachter:
Refs: https://jira.dosomething.org/browse/DOS-103

I think what's going on here is that when Apache SOLR runs cron, the `$node = menu_get_object()` is not reliable.

This PR alters the preprocessing to only occur when we are viewing a full view mode, when a `$vars['node']` will be available.


sergiitk:
:+1:

PR #2883, opened 2014-07-25
Removing git commit symlink from build.

angaither:
this is only needed on first build, doesn't need to happen every time.

I've added this step into the [wiki](https://github.com/DoSomething/dosomething/wiki/Building-your-local-environment) as it only needs to be done once on the first build of your local environment.


aaronschachter:
:+1:

PR #2878, opened 2014-07-24
DoSomething templates i18n: wrap PHP templates text into t().

sergiitk:
#### What's this PR do?
- Wraps all Paraneue DoSomething hardcoded strings found in tpl files into [`t()`](https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/t/7) function
#### Any background context you want to provide?

This is step 1. The next step will be checking all javascripts and theme functions.
#### What are the relevant tickets?

[#DOS-91](https://jira.dosomething.org/browse/DOS-91)

PR #2862, opened 2014-07-21
Campaign Scholarship theme function

aaronschachter:
Continues work on #2375 to improve campaign tpl maintainability.

@weerd Please review when you have a chance.

PR #2885, opened 2014-07-25
Campaign template fixes

aaronschachter:
@sergii-tkachenko Can you pls review?

Fixes https://jira.dosomething.org/browse/DOS-108 -- MIssed this callout in #2862

Fixes https://jira.dosomething.org/browse/DOS-110 -- Bug in #2878 was translating HTML.

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc
@@ -138,6 +138,7 @@ function dosomething_helpers_preprocess_partners_vars(&$vars) {
 function dosomething_helpers_format_partners_list(&$vars) {
   $partners = $vars['partners'];
   $formatted_partners = '';
+  $and = t('and');
   foreach ($partners as $key => $partner) {
     $name = $partner['name'];
     // If there is a partner url set:
@@ -147,13 +148,13 @@ function dosomething_helpers_format_partners_list(&$vars) {
     }
     $formatted_partners .= $name;
     if (count($partners) > 2 && $key == (count($partners) -2)) {
-      $formatted_partners .= ' and ';
+      $formatted_partners .= ' ' . $and .' ';
     }
     elseif (count($partners) > 2 && $key < (count($partners) -1)) {
       $formatted_partners .= ', ';
     }
     elseif ($key < (count($partners) -1)) {
-      $formatted_partners .= ' and ';
+      $formatted_partners .= ' ' . $and .' ';
     }
   }
   $vars['formatted_partners'] = $formatted_partners;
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc
```diff
@@ -147,13 +148,13 @@ function dosomething_helpers_format_partners_list(&$vars) {
     }
     $formatted_partners .= $name;
     if (count($partners) > 2 && $key == (count($partners) -2)) {
-      $formatted_partners .= ' and ';
```
sergiitk, line 150: I think using `$and` variable instead on explicit `t('and')` is micro-optimization vs readability again. [`t`](https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/t/7) uses static variable cache.
I never seen this in core modules. In my opinion, it's better to keep up with the style declared.

aaronschachter, line 150: ok, i don't mind changing.

aaronschachter, line 150: Updated!

lib/themes/dosomething/paraneue_dosomething/templates/system/partials/info-bar.tpl.php
```diff
@@ -20,9 +20,9 @@
     </div>
     <?php endif; ?>
 
-    <?php if (isset($sponsors)): ?>
+    <?php if (isset($formatted_partners)): ?>
       <div class="sponsor">
-      <?php print t("In partnership with @partners", array("@" => $formatted_partners)); ?>
```
sergiitk, line 25: I'm not sure if we should keep `@`-reference.
I mean, clearly, it was my fault I missed passing `@partners` to the second argument, but keeping the reference may provide additional context to translators.
It can be:
`t("In partnership with @partners", array("@partners" => $formatted_partners));`
so translators see "In partnership with @partners" in the translation interface.

But it's your call, I'm not sure myself what'd be better.

aaronschachter, line 25: The `$formatted_partners` only contains Partners taxonomy term names.  I can't imagine we'd actually want to translate individual Partner names like "Toyota" or "Sprint".

sergiitk, line 25: That's not exactly what I was thinking.
It's not about translating actual partners, but what string translators will see.
It's either `"In partnership with @partners"` or `"In partnership with"`
@partners is just a placeholder

aaronschachter, line 25: But the bug happens because `$formatted_partners` contains HTML... so it can't go into the `t` function.

sergiitk:
Ups, I merged it. The point is :+1:

PR #2891, opened 2014-07-28
User Registration Source field

aaronschachter:

Adds a new `field_user_registration_source` field to the User entity.  It's a plain text field without a default value. This will be used by the College Prowler User import, setting it to something like `import_college_prowler` or whatever value we want.
- Hides the field from non-staff
- Makes the field read-only to staff


angaither:
:+1:

PR #2892, opened 2014-07-28
Campaign fixtures taxonomy

DFurnes:
## Changes
- Creates an array of fact nodes for the More Facts modal.
- Creates a "Test Issue" issue in the Causes taxonomy if it doesn't already exist.

For review: @aaronschachter @angaither 
## Usage

To create test campaign: `drush campaign-create ../tests/fixtures/campaign.json`
## Screenshot

test-campaign-taxonomy

### DIFF
--- a/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.helpers.inc
+++ b/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.helpers.inc
@@ -1,0 +1,67 @@
+<?php
+/**
+ * @file
+ * Helper functions for dosomething_taxonomy functionality.
+ */
+
+
+/**
+ * Creates a taxonomy term with the given name if it doesn't already exist.
+ * @return term object
+ */
+function dosomething_taxonomy_create_term_once($name, $vocabulary_machine_name) {
+  if ($vocabulary = taxonomy_vocabulary_machine_name_load($vocabulary_machine_name)) {
+    $vid = $vocabulary->vid;
+    $existing_term = dosomething_taxonomy_get_term_from_name($name, $vid);
+
+    $term = new stdClass();
+
+    $term->name = $name;
+    $term->vid = $vid;
+
+    if($existing_term) {
+      $term->tid = $existing_term;
+    }
+
+    taxonomy_term_save($term);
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.helpers.inc
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Helper function to dynamically get the tid from the term_name
+ * Source: https://api.drupal.org/comment/18909#comment-18909
+ *
+ * @param $term_name Term name
+ * @param $vid ID of the vocabulary to search the term in
+ *
+ * @return Term ID of the found term or else FALSE
+ */
+function dosomething_taxonomy_get_term_from_name($term_name, $vid) {
```
aaronschachter: I would rename this as `dosomething_taxonomy_get_tid_by_name` as it returns a tid and not a term object.

lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.helpers.inc
```diff
@@ -0,0 +1,63 @@
...
+
+  return FALSE;
+}
+
+/**
+ * Creates a vocabulary with the given machine name if it doesn't already exist.
+ * Otherwise, update the existing vocabulary with the new name.
+ * @return vocabulary object
+ */
+function dosomething_taxonomy_create_vocabulary_once($name, $machine_name) {
```
aaronschachter: Hmm. I'm not into the naming `create_vocabulary_once` convention.. because by definition a Create operation should just create once.  Plus this doesn't create if one exists. 

The vocabularies should always exist because of the DoSomething Taxonomy features being enabled though.  I think you can just get away with `taxonomy_vocabulary_machine_name_load('issue')`in`dosomething_taxonomy_create_term_once` ...

DFurnes: Yeah, if "Issue" is guaranteed to exist even without an pulling down staging DB then that makes sense.

aaronschachter:
:+1:

PR #2893, opened 2014-07-28
User Profile variables

aaronschachter:
Without explicitly setting the `user_picture_dimensions` variable, profile pictures save to 85x85.  This fixes.


weerd:
:+1:

PR #2898, opened 2014-07-29
Indexed fields

blisteringherb:
Readding the Staff Pick field for the Campaign and Campaign Group content type and the Campaign Status field for campaigns to fields indexed by Solr.  The staff pick field is necessary for appropriately boosting the results, and the campaign status field is necessary for correctly excluding closed campaigns from the finder and /campaigns.

Addresses: https://jira.dosomething.org/browse/DOS-117
@aaronschachter


aaronschachter:
:+1:

PR #2899, opened 2014-07-29
Campaign Downloads field

aaronschachter:

Creates a new `field_downloads` as a multi-value File field and adds to the Campaign node type and Campaign class.
- Closes https://jira.dosomething.org/browse/DOS-96 -adding `field_downloads`
- Closes https://jira.dosomething.org/browse/DOS-97 - outputs in Campaign Action Page tpl


lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -124,19 +124,32 @@
...
+                  </ul>
+                <?php endif; ?>
+
+                <?php if (isset($signup_data_form_link)): ?>
+                  <ul>
+                    <li><a href="#" data-modal-href="#modal-signup-data-form"><?php print $signup_data_form_link; ?></a></li>
+                  </ul>
+                <?php endif; ?>
+
+              <?php endif; // if content['category'] == 'materials' ?>
```
sergiitk, line 151: Did you forget to remove the comment?

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
```diff
@@ -179,6 +182,16 @@ function dosomething_campaign_load($node, $public = FALSE) {
     $campaign->{$property} = dosomething_helpers_get_field_collection_values($wrapper, $field_name);
   }
 
+  $campaign->downloads = NULL;
```
sergiitk, line 185: `$campaign->downloads = array();`?

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
aaronschachter, line 151: No that's intentional, indicating it's closing the if statement for materials.

sergiitk, line 151: Yes, makes sense.

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
aaronschachter, line 185: All the other properties are being set to NULL in this class if they don't exist.  Seemed to make more sense for an API response vs setting to an empty array.  I'm open to discussing whether or not the defaults should be empty arrays, but would prefer to leave this as is for now to keep consistent, and then we can chat about whether or not these properties should be empty arrays or set to NULL

sergiitk, line 185: I think empty array makes sense because below you're doing dynamic assignment `$campaign->downloads[] = …`.

sergiitk:
:+1:

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
aaronschachter, line 185: Chatted with @sergii-tkachenko -- we'll discuss more and refactor the Campaign object to use empty arrays for array properties later.

PR #2901, opened 2014-07-29
Campaign fixture images

DFurnes:
## Changes

Adds :camera:s to campaign test node via JSON. Closes [FT-20](https://jira.dosomething.org/browse/FT-20).

For review: @angaither @aaronschachter 
## Screenshot

It's starting to look like a real campaign!
campaign-fixtures-images


tests/fixtures/campaign.json
```diff
@@ -5,6 +5,15 @@
     "issue": "Test Issue",
     "status": "active",
     "active_hours": "2",
+    "image_campaign_cover": {
+      "title": "Test Cover Image",
+      "src": {
+        "landscape": "/var/www/vagrant/tests/fixtures/cover-landscape.jpg",
```
angaither: will this path work on stage & QA?

DFurnes: Hmm, good question. @blisteringherb?

angaither: I think it should be ok on stage, but QA is not a vagrant box. 
Can the paths be relative?

DFurnes: Let me play around with it. It would definitely make more sense if they were relative to `html/..`.

aaronschachter: If there was a hardcoded path liek http://flickr.com/whatever this could be an easy solution too.... but then relies on that image always living there

lib/modules/dosomething/dosomething_image/dosomething_image.helpers.inc
```diff
@@ -0,0 +1,50 @@
...
+
+  node_save($node);
+
+  return $node;
+}
+
+/**
+ * Create an image field from a file path.
+ */
+function dosomething_image_create_image_field_from_path($path) {
```
aaronschachter: This is really creating a File from path.  can rename this accordingly?

angaither:
:shipit: :+1:

PR #2902, opened 2014-07-29
Campaign  faming updates

weerd:
@DFurnes quickie!


DFurnes:
:snowboarder:

aaronschachter:
@weerd Can you start rebasing these commits?  Three commits for such small changes.

weerd:
@aaronschachter sure. It was really 1 commit, but then I had file corruption issues from the SSHFS and the only way to correct it is to do another commit, so I made another small change and then reset it :goat: Not sure if there's a better way. But yeah, I can combine them next time.

PR #2903, opened 2014-07-30
bithday vs birthday

deezone:
Stupid typo on my part...


aaronschachter:
:+1: So Birthday Mail SMS has just never been working at all (did we ever get this working on staging?)

PR #2904, opened 2014-07-30
Adding some style defaults for the creator modal.

weerd:
@DFurnes quickie for modal. Adding in defaults, making sure it looks pretty even with very little data :goat:


DFurnes:
:goat: :trophy:

PR #2906, opened 2014-07-30
Sms Workflow entity

aaronschachter:
Creates a `dosomething_sms_workflow` table to store SMS Reportback Workflows.

Next PR will query this table to actually create keywords and endpoints.

Refs https://jira.dosomething.org/browse/DOS-125

screen shot 2014-07-30 at 12 59 13 pm
screen shot 2014-07-30 at 12 59 42 pm

### DIFF
--- a/lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
+++ b/lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
@@ -4,6 +4,39 @@
  * Code for dosomething_sms admin functionality.
  */
 
+function dosomething_sms_workflow_overview_page() {
+  $admin_path = 'admin/config/dosomething/dosomething_sms';
+  $add_path = $admin_path . '/add';
+  $add_link = l(t('Add SMS Workflow'), $add_path);
+  $output = '<ul class="action-links"><li>' . $add_link. '</li></ul>';
+  $rows = array();
+  $header = array(
+    t('ID'),
+    t('Endpoint'),
+    t('Node'),
+    t('Status'),
+    t('Op'),
+  );
+  $result = db_select('dosomething_sms_workflow', 'sw')
+      ->fields('sw')
+      ->execute()
+      ->fetchAll();
+  foreach ($result as $workflow) {
+    $nid = $workflow->nid;
+    $rows[] = array(
+      $workflow->id,
+      $workflow->endpoint,
+      l($nid, 'node/' . $nid . '/reportbacks'),
+      $workflow->status ? t('Enabled') : t('Disabled'),
+      l(t('Edit'), $admin_path . '/' . $workflow->id . '/edit'),
+    );
+  }
+  return $output . theme('table', array(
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
```diff
@@ -0,0 +1,115 @@
+<?php
+/**
+ * @file
+ * Code for dosomething_sms admin functionality.
+ */
+
+function dosomething_sms_workflow_overview_page() {
+  $admin_path = DOSOMETHING_SMS_WORKFLOW_ADMIN_PATH;
+  $add_path = $admin_path . '/add';
+  $add_link = l(t('Add SMS Workflow'), $add_path);
```
sergiitk: `$add_link = l(t('Add SMS Workflow'), $admin_path . '/add');` ?

lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
```diff
@@ -0,0 +1,115 @@
...
+    t('Op'),
+  );
+
+  foreach (dosomething_sms_get_sms_workflows() as $workflow) {
+    $nid = $workflow->nid;
+    $rows[] = array(
+      $workflow->id,
+      $workflow->endpoint,
+      l($nid, 'node/' . $nid . '/reportbacks'),
+      $workflow->status ? t('Enabled') : t('Disabled'),
```
sergiitk, line 26: Does it work? I'd put it in parentheses, but that's ok if it works.

aaronschachter, line 26: It does, viewable in screenshot. I can update this though.

lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
aaronschachter: Good point - if nothing else critical I'll make this change in my next PR, as there's another one coming up soon.

lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
```diff
@@ -0,0 +1,115 @@
...
+      l(t('Edit'), $admin_path . '/' . $workflow->id . '/edit'),
+    );
+  }
+  return $output . theme('table', array(
```
sergiitk: I wouldn't overload return so much, but split it to `$output .= theme(…` , `return $output`;

aaronschachter: Good point.

lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
sergiitk, line 26: No, leave it like this then.

sergiitk:
:+1:

PR #2908, opened 2014-07-30
Dynamic workflow

aaronschachter:
Creates keywords and workflow based on active records in the `dosomething_sms_workflow` table.

Closes DOS-125 
Closes https://jira.dosomething.org/browse/DOS-125 or however I can get Jira to see this stuff

screen shot 2014-07-30 at 2 29 20 pm


sergiitk:
:+1:

PR #2911, opened 2014-07-30
Only display zenddesk form on ds.org, not on affiliate sites.

angaither:
Fixes [DOS-124](https://jira.dosomething.org/browse/DOS-124?filter=-1)

cc: @sergii-tkachenko


aaronschachter:
Code is fine, but why not just disable the module, per comments in the issue?

sergiitk:
:+1:

sergiitk:
@aaronschachter We have no intl-specific deployment code yet (no special features, no modules, no themes).

aaronschachter:
Hmm yeah. Could be done by removing dosomething_zendesk from the dosomething.info dependencies, but ya, the quick fix is fine

PR #2913, opened 2014-07-30
Added 'explore campaigns' to editable menu text.

angaither:
Fixes [DOS-101](https://jira.dosomething.org/browse/DOS-101)

TODO: I've left the "Explore Campaigns" link code commented out for now.
I'm not sure what we want to do with that, if we allow an override (with an entity reference), you won't be able to set the link to `/campaigns`

Thoughts?

cc @sergii-tkachenko  @aaronschachter


aaronschachter:
I don't know why that code is there, has there ever been a request to override the Campaigns page to actually not direct to `/campaigns`? @sergii-tkachenko @barryclark

aaronschachter:
:+1:

angaither:
@aaronschachter the request was in the Jira ticket, I'm not sure if it's actually needed...

PR #2914, opened 2014-07-30
Updates to tweak some layout issues with the new promotions approach in ...

weerd:
...the header.

@DFurnes some small tweaks to help with layout issues as we tested the promotions on staging, etc. :tropical_fish:


DFurnes:
Looks fine to merge. :accept:

A wee bit wary of some of the hacks we're having to do to make this work, but nothing blocking.

weerd:
:mushroom:

PR #2915, opened 2014-07-30
SMS Worfklow debugging

aaronschachter:
Adds settings and functionality for:
- Bypass IP access checks for Mobile Commons IP's.  Useful for testing from a browser REST console.
- Log `$context` input and `$output` if logging variable is set.

### DIFF
--- a/lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
+++ b/lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
@@ -46,6 +46,10 @@ function dosomething_sms_workflow_overview_page() {
     'header' => $header,
     'rows' => $rows,
   ));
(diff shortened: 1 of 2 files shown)


aaronschachter:
Hmm so this isn't exactly good.  Turned debugging on for our staging SMS campaign and got this cool error.  Tough to debug any of this stuff since mobilecommons can't see our local machines.  Probably just going to remove this code tomorrow (for now we can just keep debugging off)
img_1644

deezone:
Should we (I) make an issue for this as I assume it's filling up our error logs somewhere?

aaronschachter:
Right now it's only on staging, and it's turned off -- so we just have the usual PHP notices in the error logs :) I'll be cleaning this up tomorrow am and also adding help text to the Workflows form.

PR #2916, opened 2014-07-31
Workflow Form

aaronschachter:
- Removes debugging call that seemed to break things when on
- Adds lots of help text and form validation
- Gives `communication team` role access to add and edit SMS Workflows

### DIFF
--- a/lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
+++ b/lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
@@ -51,8 +51,12 @@ function dosomething_sms_workflow_overview_page() {
     'header' => $header,
     'rows' => $rows,
   ));
-  $config_form = drupal_get_form('dosomething_sms_workflow_config_form');
-  $output .= render($config_form);
+
+  if (user_access('administer modules')) {
+    $config_form = drupal_get_form('dosomething_sms_workflow_config_form');
+    $output .= render($config_form);
+  }
+
   return $output;
 }
 
--- a/lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
+++ b/lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
@@ -9,8 +9,13 @@
  */
 function dosomething_sms_workflow_overview_page() {
   $admin_path = DOSOMETHING_SMS_WORKFLOW_ADMIN_PATH;
+
+  $help = "A SMS Workflow enables Reportbacks via MobileCommons for a Campaign.";
+  $output = '<p>' . t($help) . '</p>';
+
   $add_link = l(t('Add SMS Workflow'), $admin_path . '/add');
-  $output = '<ul class="action-links"><li>' . $add_link. '</li></ul>';
+  $output .= '<ul class="action-links"><li>' . $add_link. '</li></ul>';
+
   $rows = array();
   $header = array(
     t('ID'),
@@ -46,8 +51,6 @@ function dosomething_sms_workflow_overview_page() {
     'header' => $header,
     'rows' => $rows,
   ));
-  $output .= '<h2>Development</h2>';
-  $output .= '<p>These should be turned off in production.</p>';
   $config_form = drupal_get_form('dosomething_sms_workflow_config_form');
   $output .= render($config_form);
   return $output;
@@ -63,18 +66,13 @@ function dosomething_sms_workflow_form($form, &$form_state, $workflow = NULL) {
       '#value' => $workflow->id,
     );
   }
+  $desc = "The unique path for this workflow, appended to the https://www.dosomething.org/dosomething_sms URL that we use for mDatas.<p>e.g. If the endpoint is set to <em><strong>comebackclothes-reportback</strong></em>, the URL for its associated mData needs to be set to <em>https://dosomething.org/dosomething_sms/comebackclothes-reportback</em>.</p>Only use alpha-numeric characters and dashes (no spaces).";
   $form['endpoint'] =  array(
     '#type' => 'textfield',
     '#title' => t('Endpoint'),
-    '#description' => "The unique path that this workflow is be accessed from.",
+    '#description' => t($desc),
     '#required' => TRUE,
     '#default_value' => isset($workflow->endpoint) ? $workflow->endpoint : NULL,
-  );
-  $form['title'] =  array(
-    '#type' => 'textfield',
-    '#title' => t('Human readable title'),
-    '#required' => TRUE,
-    '#default_value' => isset($workflow->title) ? $workflow->title : NULL,
   );
   $form['nid'] =  array(
     '#type' => 'entity_autocomplete',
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
```diff
@@ -9,8 +9,13 @@
  */
 function dosomething_sms_workflow_overview_page() {
   $admin_path = DOSOMETHING_SMS_WORKFLOW_ADMIN_PATH;
+
+  $help = "A SMS Workflow enables Reportbacks via MobileCommons for a Campaign.";
```
deezone, line 13: Being picky but should all this markup be in a template?

aaronschachter, line 13: Hmm I'd say no to this one - this is just an admin form that never is user facing.

lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
```diff
@@ -46,10 +51,12 @@ function dosomething_sms_workflow_overview_page() {
...
   ));
-  $output .= '<h2>Development</h2>';
-  $output .= '<p>These should be turned off in production.</p>';
-  $config_form = drupal_get_form('dosomething_sms_workflow_config_form');
-  $output .= render($config_form);
+
+  if (user_access('administer modules')) {
+    $config_form = drupal_get_form('dosomething_sms_workflow_config_form');
+    $output .= render($config_form);
+  }
```
deezone, line 58: A message when the user doesn't have the rights to access the form? I'm guessing members of the communication team will access this page who might not have the right admin rights if access ever gets more granular?

lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
```diff
@@ -63,19 +70,14 @@ function dosomething_sms_workflow_form($form, &$form_state, $workflow = NULL) {
       '#value' => $workflow->id,
     );
   }
+  $desc = "The unique path for this workflow, appended to the https://www.dosomething.org/dosomething_sms URL that we use for mDatas.<p>e.g. If the endpoint is set to <em><strong>comebackclothes-reportback</strong></em>, the URL for its associated mData needs to be set to <em>https://dosomething.org/dosomething_sms/comebackclothes-reportback</em>.</p>Only use alpha-numeric characters and dashes (no spaces).";
```
deezone, line 73: Template and / or external l()?

lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
aaronschachter, line 58: This form is only for setting those Debug variables on though, which is kind of out of scope of what would be useful for the Communication Team.  The communication team is only able to add and edit Workflows as per this PR.

lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
aaronschachter, line 73: This shouldn't be a clickable link though, this is help text to specify what the URL to set the mData is when you're setting up everything in Mobile Commons.

lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
```diff
@@ -91,6 +93,14 @@ function dosomething_sms_workflow_form($form, &$form_state, $workflow = NULL) {
     // Default to 1 for new workflows.
     '#default_value' => isset($workflow->status) ? $workflow->status : 1,
   );
+  $desc = t("This title is used internally to describe this workflow. e.g. <em>Comeback Clothes 2014.</em>");
```
deezone, line 96: This value is also referenced in the mData web service url, ex:
http://www.dosomething.org/dosomething_sms/imaboss-reportback
where "imaboss-reportback" would be this value.

aaronschachter, line 96: You're referring to the `endpoint` form value.. which is this help text that is meant to explain that: https://github.com/DoSomething/dosomething/pull/2916#discussion_r15643135

aaronschachter, line 96: This description is for the `title` field, which seems to only be used when the Workflow is constructed in `workflows.inc`

deezone, line 96: Yep, you're right - endpoint vs "human readable".

lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
deezone, line 73: Reading through the whole value for $desc this looks good.

lib/modules/dosomething/dosomething_sms/dosomething_sms.admin.inc
```diff
@@ -138,13 +159,20 @@ function dosomething_sms_workflow_form_submit($form, &$form_state) {
...
+    '#title' => t('Development'),
+    '#collapsible' => TRUE,
+    '#collapsed' => TRUE,
+    '#description' => t('These settings should always be disabled in production.'),
+  );
+  $form['dev']['dosomething_sms_workflow_log'] =  array(
     '#type' => 'checkbox',
     '#title' => t('Enable logging'),
     '#description' => t("Logs requests to any SMS Workflow endpoint."),
     '#default_value' => variable_get('dosomething_sms_workflow_log'),
```
deezone, line 173: Sweet!

lib/modules/dosomething/dosomething_sms/dosomething_sms.features.user_permission.inc
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+/**
+ * Implements hook_user_default_permissions().
+ */
+function dosomething_sms_user_default_permissions() {
+  $permissions = array();
+
+  // Exported permission: 'administer dosomething_sms'.
+  $permissions['administer dosomething_sms'] = array(
```
deezone, line 14: An additional admin type permission to grant to the communication team?

aaronschachter, line 14: I'm pretty sure we don't use `administer dosomething_sms` anywhere, so i've just used this existing permission.  I've logged in as a communication team to test, and all they see is the SMS workflows page (in addition to the Opt-In config form which they already have access to)

lib/modules/dosomething/dosomething_sms/dosomething_sms.module
```diff
@@ -24,7 +24,7 @@ function dosomething_sms_menu() {
     'description' => 'Overview of all SMS Workflows',
     'page callback' => 'dosomething_sms_workflow_overview_page',
     'access callback' => 'user_access',
-    'access arguments' => array('administer modules'),
+    'access arguments' => array('administer dosomething_sms'),
```
deezone, line 27: Ah, here are the extra rights!

deezone:
:+1:

lib/modules/dosomething/dosomething_sms/dosomething_sms.info
```diff
@@ -1,9 +1,9 @@
-name = DoSomething SMS
+name = dosomething_sms
```
angaither, line 1: Does this module name need to change in `dosomething.info` as well?

aaronschachter, line 1: Ahh sneaky Features Export, changing shit. This is just that user facing name that appears in the Features list.  The dosomething.info uses the module machine name `dosomething_sms`.  I can change this later to make it look more pertty.

PR #2917, opened 2014-07-31
Adds num_participants field to Reportback table

aaronschachter:
Adds a `num_participants` column to the `dosomething_reportback` table, to be used by The Hunt and various other campaigns (not all).  Hence setting to NULL by default.

Closes subtask https://jira.dosomething.org/browse/DOS-67


angaither:
:+1:

PR #2919, opened 2014-07-31
Communications team permissions

aaronschachter:
Grants Communications Team some more permissions to make the SMS Workflows easier to use:
- Access to Entity Autocomplete -- allows selecting the Campaign node as dropdown instead of needing to know the nid
- Access to view any reportback -- will allow them to verify that a SMS Reportback has been received.


angaither:
:+1:

PR #2920, opened 2014-07-31
Temporary fix for JS issues on staging.

DFurnes:
Temporary fix for JS issues on staging. Will hopefully be resolved with better solution for `drupal.js` issues later today. :cactus:


blisteringherb:
:+1:

PR #2922, opened 2014-07-31
Set the title of unpublished campaigns in a group to 'Stay Tuned'

angaither:
Fixes [DOS-126](https://jira.dosomething.org/browse/DOS-126?)


aaronschachter:
:+1:

PR #2923, opened 2014-07-31
Reportback form: Number of participants

aaronschachter:
Creates a new admin form on the `node/*/custom-settings` which allows editors to check whether or not to collect Number of Participants on the reportback form.

If checked, the form adds the additional field and saves to the reportback.
screen shot 2014-07-31 at 1 50 16 pm

screen shot 2014-07-31 at 1 49 57 pm


angaither:
:+1:

PR #2924, opened 2014-07-31
Fixing bubble query to respect queried field in filter

blisteringherb:
The bubble function query was being added to the wrong place and was overriding the existing query.  This PR adds them together.

@DFurnes


DFurnes:
Looks good. :cactus:

PR #2925, opened 2014-07-31
Mcc  form update

weerd:
@DFurnes @aaronschachter @barryclark 

Here's the update for the MCC Form. I've include some logic for blocks, because I needed the output for the form as a block without any of the block wrapper markup. With the approach I took it would allow for indicating other block ids that we want to output with clean markup. 

@DFurnes suggested maybe nuking the wrapper markup for all block output at a later date anyhoo, but we'd ideally test the site thoroughly before that happens. The `$clean_ouput` provides an intermediary approach to eventually get to that point :tropical_drink: 

**EDIT**: It was decided that the markup around blocks wasn't really necessary, so we nuked it now instead of later! :bomb:

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/campaign/submit-campaign-idea.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/campaign/submit-campaign-idea.tpl.php
@@ -2,7 +2,7 @@
 
   <form class="wrapper" action="https://docs.google.com/a/dosomething.org/forms/d/1YY2BTXz65IcqyjaTf6GT8FZx0Z2E1251LcSf5LIhc0s/formResponse" method="post" id="ss-form" target="_self" onsubmit="">
     <p class="required">
-      <strong><?php print t('All fields are required.'); ?></strong>
+      <strong>All fields are required.</strong>
     </p>
 
     <?php // First Name Field ?>
@@ -10,13 +10,11 @@
       <div dir="ltr" class="ss-item ss-item-required ss-text">
         <div class="ss-form-entry">
           <div class="ss-q-title">
-            <label class="ss-q-item-label" for="entry_212037837"><?php print t('First Name'); ?> <span class="ss-required-asterisk">*</span></label>
+            <label class="ss-q-item-label" for="entry_212037837">First Name <span class="ss-required-asterisk">*</span></label>
           </div>
-          <div class="ss-q-help ss-secondary-text" dir="ltr"></div><input type="text" name="entry.212037837" value="" class="ss-q-short" id="entry_212037837" dir="auto" aria-label="<?php print t('First Name'); ?>" aria-required="true" required="" title="">
+          <div class="ss-q-help ss-secondary-text" dir="ltr"></div><input type="text" name="entry.212037837" value="" class="ss-q-short" id="entry_212037837" dir="auto" aria-label="First Name" aria-required="true" required="" title="">
           <div class="error-message" id="1205439792_errorMessage"></div>
-          <div class="required-message">
-            <?php print t('This is a required question'); ?>
-          </div>
+          <div class="required-message">This is a required question.</div>
         </div>
       </div>
     </div>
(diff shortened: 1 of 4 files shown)


aaronschachter:
I think we're actually fine to just get rid of all the block markup.: The only other place we're using Blocks at all is in the Fact Page archive (dosomething.org/facts).  Would you mind testing to see if killing the block markup messes up that page?  If not, lets just get rid of it so we don't need hte preprocess.

weerd:
Ok. Sounds good.

weerd:
@aaronschachter ok! I removed the surrounding markup. Fact Page Archive works fine without it. This is much cleaner! :penguin:

aaronschachter:
:+1: bender approved

lib/themes/dosomething/paraneue_dosomething/templates/campaign/submit-campaign-idea.tpl.php
```diff
@@ -1,646 +1,352 @@
...
+    <p class="required">
+      <strong>All fields are required.</strong>
+    </p>
+
+    <?php // First Name Field ?>
+    <div class="ss-form-question errorbox-good -compact -alpha">
+      <div dir="ltr" class="ss-item ss-item-required ss-text">
+        <div class="ss-form-entry">
+          <div class="ss-q-title">
+            <label class="ss-q-item-label" for="entry_212037837">First Name <span class="ss-required-asterisk">*</span></label>
```
mshmsh5000, line 13: @weerd please run this by @sergii-tkachenko on the translation front. I'm going to assume that all hard-coded text like `First Name` should be passed through `t()` from now on, but defer to him on that.

aaronschachter, line 13: @mshmsh5000 @mikefantini @barryclark  Would the MCC Form ever be available to international affiliates tho?

weerd, line 13: I had it at first, but @DFurnes and I discussed it. This form is only meant for US members... so running the extra functions on it for something that didn't need translatin' didn't seem necessary... However we could be wrong?

sergiitk, line 13: @mshmsh5000 @weerd Yes, it should be `<?php print t('First Name'); ?>`

mshmsh5000, line 13: Possibly unknowable at this point. It's true that it's _meant_ for US members.

I'm fine if we want to put off the `t()` calls for now; it's reasonable to imagine that this form will go through more edits than just string translation before we could deploy it internationally.

mshmsh5000, line 13: So, like, stet

weerd, line 13: @aaronschachter Potentially, but we'd need a different form anyways with new content. There's specifically a line asking about US members "Can your idea be done by anyone in the United States?" I'm assuming international affiliates would use a different form, and then those fields would need to be translated... these are hardcoded fields. Grey area? :confused:

mikefantini, line 13: yeah @aaronschachter @weerd @mshmsh5000 @sergii-tkachenko there is no chance we would open this form up to international affiliates. If we open up MCC to international instances it would be a more developed product, not a google form we theme.

weerd, line 13: @mshmsh5000 yeah, plus I chatted with @aaronschachter about future form stuff. Right now it's only 1 hard-coded form that we can add to a static node (always the same form). So in future if we want to add more Google forms, and a variety of them, we need to refactor and change how the forms are added anyhoo. :)

DFurnes, line 13: Yeah, I don't think it's worth the time it would take to go into the generated markup and add `t()` functions by hand unless we know the form is going to be used internationally. Since the MCC form has a bunch of US-specific questions, I assume we'd be creating a separate Google Form if affiliates were ever going to have their own.

weerd, line 13: Cool. Yeah, definitely agree that we should use the `t()` once we go international with the MCC forms stuff, but that'd be a more fleshed out form. Right now this form was more of a copy and paste from Google Forms and then cleaning up the crazy HTML that Google gave me and adding some fender fanciness :goat: 

Good discussion though, the chat activity on this PR makes me feel so popular! :bowtie:

PR #2927, opened 2014-07-31
Joining bubble query with AND

blisteringherb:
The bubble query has to be joined to the other query with AND so that filtering works properly and the correct number of documents are returned.

@DFurnes @weerd


DFurnes:
ok

weerd:
:ramen: :+1:

PR #2931, opened 2014-07-31
Adding active hours as an indexed field for campaigns

blisteringherb:
The field_active_hours is not being indexed by solr which is greying out the time filters.

@angaither


angaither:
:+1:

PR #2932, opened 2014-07-31
Updating number validation in reportbacks to not allow 0.

weerd:
@DFurnes Number validation update that you helped me with :tropical_drink:

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/validation/reportback.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/validation/reportback.js
@@ -5,7 +5,8 @@ define(function (require) {
 
   // validate number of items
   Validation.registerValidationFunction("reportbackNumber", function(string, done) {
-    if( string !== "" && /^[,0-9]+$/.test(string) ) {
+    var trimmedString = string.replace(" ", "");
+    if( trimmedString !== "" && /^[1-9]\d*$/.test(trimmedString) ) {  //^[1-9]\d*$
       return done({
         success: true,
         message: "That's great!"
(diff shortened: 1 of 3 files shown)


aaronschachter:
Is it worth renaming this function to something more generic like `PositiveInteger` vs `ReportbackCount` ?

weerd:
Potentially... Señor @DFurnes can you weigh in since this is your baby? :baby:

DFurnes:
Renaming to `positiveInteger` sounds like a good call to me. :+1:

weerd:
Okie doke. Will do. Updating incoming!

DFurnes:
:shipit:

lib/themes/dosomething/paraneue_dosomething/js/validation/reportback.js
```diff
@@ -4,8 +4,9 @@ define(function (require) {
   var Validation= require("neue/validation");
 
   // validate number of items
-  Validation.registerValidationFunction("reportbackNumber", function(string, done) {
-    if( string !== "" && /^[,0-9]+$/.test(string) ) {
+  Validation.registerValidationFunction("positiveInteger", function(string, done) {
+    var trimmedString = string.replace(" ", "");
+    if( trimmedString !== "" && /^[1-9]\d*$/.test(trimmedString) ) {
```
aaronschachter, line 9: Just making sure... (I suck at regex) but this isn't filtering out numbers with 0's in it, is it?  Like if you test with 10 or 101

weerd, line 9: Tested and it's fine :) Only filters out, just `0`, but not integers > 0, with 0's in them :)

PR #2933, opened 2014-08-01
Parent opt in

aaronschachter:
Resolves https://jira.dosomething.org/browse/DOS-129 by not sending third-party opt-ins for a Campaign which is a child of a Campaign Group.

Also adds logging variables, allowing us to log Message Broker payloads, MobileCommons requests, and Signup Entity insert/updates, which I used to test.


deezone:
Nice logging addition. Looks good to me :+1:

PR #2934, opened 2014-08-01
Store and print NIDs and Entity IDs of created test fixtures.

DFurnes:
:construction: 

Still working on this.

Open question: How should I be returning this list of NIDs and entity IDs? Looking for PHP advice. :apple: :pencil:

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -330,6 +330,8 @@ function dosomething_campaign_load_stats(&$campaign) {
  */
 function dosomething_campaign_create_node_from_json($string) {
   $data = json_decode($string);
+  $created_nodes = array();
+  $created_entities = array();
   $node = new stdClass();
 
   $node->type = $data->type;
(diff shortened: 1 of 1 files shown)


aaronschachter:
If I remember correctly, the purpose of returning the list of id's (vs just the newly created `$node`, which would be the expected return of `create_node_from_json`) to delete all the newly created test nodes / entities?

What are you planning to use to actually delete all of the nodes / entities?  Drush command once the Casper tests run?

Random (possibly bad) idea but I wonder if there's a crazy hack we can do, like setting a common property on all of them, perhaps a property like `created` to -1 or something like that, so we'd know which ones to delete, without needing to worry about what id's were created.  We could run something like  `drush delete-ds-test-data`

DFurnes:
That sounds clever. The nice thing about that is that we can clean up test data at any time too, rather than relying on having the stored NIDs from the previous run.

aaronschachter:
Yeah... the more i was thinking about this, node_delete will get you pretty close as far as deleting all the field_collection_item id's.  It's more of whether or not we want a bunch of Coffee Images and terms (which would not be deleted).  Not sure if `created` is a property you can overwrite on the `node_save` / `taxonomy_term_save` though... but maybe just using a UID like 1 could do the trick.

DFurnes:
Closing this guy in favor of #2957.

PR #2939, opened 2014-08-05
File proxy update

aaronschachter:
Updates Stage File Proxy module to 1.6, which provides an admin config form at `admin/config/system/stage_file_proxy`

Sets the origin variable within `ds pull stage --db` to fix bug described in https://github.com/DoSomething/dosomething/pull/2760#issuecomment-51093919

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -350,10 +350,13 @@ function grunt {
 function pull_db {
   if [[ $1 == "stage" ]]
   then
-    echo "Pulling down db from staging"
+    echo "Pulling down db from staging..."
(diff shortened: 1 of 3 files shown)


angaither:
Oh nice, does this mean we should also take the variable out of `settings.local`?

aaronschachter:
I guess so?  I'm not sure how that's supposed to work properly. @blisteringherb mentioned that it should have copied the `$conf` variable into `settings.local`, but that didn't work for me.  I suppose it makes sense to just remove it from the settings completely.  I'll do that.

PR #2940, opened 2014-08-05
Only load campaign stats for API requests

aaronschachter:
Only queries for signups when we load the campaign via API, which is the only place this is used right now.  Should significantly cut down on number of queries in general.

Closes https://jira.dosomething.org/browse/DS-166


angaither:
:+1:

PR #2942, opened 2014-08-05
RequireJS bundles

DFurnes:
# Changes
- Compile separate `lib.js` (jQuery, etc.) and `app.js` bundles. This lets us squeeze `drupal.js` in between, fixing some frustrating dependency issues on the Drupal side of things.
- Updates test runner to include `drupal.js` and a mocked `Drupal.settings` variable. A+! :100: 
- Don't do crazy acrobatics to fit everything into `dist` directory. We'll follow standard pattern of bundling whole `paraneue_dosomething` directory when preparing assets for CDN (just like Neue, and most NPM/Bower packages).

Note: I'm leaving `drupal.js` up to Drupal for now (rather than either requiring a full Drupal environment to bundle assets, or having a second copy added to version control), and hopefully the extra files can all be concatenated by Drupal once we get a chance to add the [CDN](https://www.drupal.org/project/cdn) module.
# Related Jenkins changes:

The [Deploy-Assets](http://admin.dosomething.org/jenkins/job/Deploy-Assets/) Jenkins job will need to be updated to store whole `paraneue_dosomething` directory for future releases, rather than just the `dist` directory. Since we'll prune unnecessary dependencies before storing it, this should only take a few MB more than before. Here's the plan:

``` diff
  #!/bin/bash -e
  export TERM=xterm
  source ~/.bash_profile
  nvm use 0.10
  eval `ssh-agent -s` 
  ssh-add

+ # Install dependencies & build assets
  cd lib/themes/dosomething/paraneue_dosomething
  npm install
  bower update --config.interactive=false
  grunt prod

+ # Clean up devDependencies from node_modules
+ npm prune --production
+ 
+ # Copy paraneue_dosomething into versioned directory on CDN
+ cd ..
  rm -rf ~/assets/$VERSION ~/assets/latest
- cp -rf dist ~/assets/latest
+ cp -rf paraneue_dosomething ~/assets/latest
- mv dist ~/assets/$VERSION
+ cp paraneue_dosomething ~/assets/$VERSION
```


sergiitk:
That's just amazing! :+1:

sergiitk:
Wow. How did you squeeze `drupal.js` to lib?

P.S. This PR totally goes to my favorites. Cheers, @DFurnes!

DFurnes:
It's just living in Drupal world for now. I'm looking into setting up the [CDN module](https://www.drupal.org/project/cdn) so that hopefully we can let Drupal deal with Drupal scripts and RequireJS deal with our own scripts, and still get them all bundled up into one nice HTTP friendly package.

But for now, this gets us part-way there!

sergiitk:
Great idea!

PR #2943, opened 2014-08-06
Reportback form: Number of participants variable

aaronschachter:
@sergii-tkachenko Can you pls review?

Adds a variable to make the label of the "Number of participants" Reportback Form field editable via Campaign Settings admin form.

Closes https://jira.dosomething.org/browse/DS-190


sergiitk:
:+1:

PR #2946, opened 2014-08-06
Dosomething Reward module

aaronschachter:
Closes https://jira.dosomething.org/browse/DS-132

This is a very basic PR -- Adds a new `dosomething_reward` module, which adds a Reward entity, stored in a `dosomething_reward` table. 

Doesn't actually do anything yet.  Opened now for readability.

This PR purposely does not add the module into the `dosomething.info` file.  It will not be enabled in affiliate sites.


angaither:
:+1: 
:gem: value added. :shipit: 
ship it.

PR #2947, opened 2014-08-06
Reportback Count Reward

aaronschachter:

Upon creating a new Reportback entity, check if User has 2 or more Reportbacks.  

If so, create a Reward with reward_type `reportback_count`.

Closes https://jira.dosomething.org/browse/DS-138

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -734,6 +734,28 @@ function dosomething_reportback_get_reportback_total_by_nid($nid, $type = 'quant
 }
 
 /**
+ * Returns count of Reportbacks for a given node $uid.
+ *
+ * @param int $uid
+ *   Optional - the user uid to get count of Reportbacks for.
+ *   If not given, uses global $user->uid.
+ *
+ * @return int
+ *   The number of Reportbacks for the User.
+ */
+function dosomething_reportback_get_count_by_uid($uid = NULL) {
+  if ($uid == NULL) {
+    global $user;
+    $uid = $user->uid;
+  }
+  // Select all reportbacks for given $uid.
+  $query = db_select('dosomething_reportback', 'rb')
+    ->fields('rb', array('rbid'))
+    ->condition('uid', $uid);
+  return $query->execute()->rowCount();
+}
+
+/**
  * Returns array of Reportback rbid's which have been flagged.
  *
  * @param int $nid
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -737,7 +737,7 @@ function dosomething_reportback_get_reportback_total_by_nid($nid, $type = 'quant
  * Returns count of Reportbacks for a given node $uid.
  *
  * @param int $uid
- *   Optional - the user uid to get count of Reportbacks for.
+ *   (optional) The User uid to get count of Reportbacks for.
  *   If not given, uses global $user->uid.
  *
  * @return int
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -734,6 +734,28 @@ function dosomething_reportback_get_reportback_total_by_nid($nid, $type = 'quant
 }
 
 /**
+ * Returns count of Reportbacks for a given node $uid.
+ *
+ * @param int $uid
+ *   Optional - the user uid to get count of Reportbacks for.
```
sergiitk: Optional parameters are indicated by [(optional)](https://www.drupal.org/node/1354#param).

lib/modules/dosomething/dosomething_reward/dosomething_reward.module
```diff
@@ -24,3 +24,76 @@ function dosomething_reward_entity_info() {
...
   return $info;
 }
+
+/**
+ * Checks if a Reward exists.
+ *
+ * @param string $reward_type
+ *   The type of Reward to check for.
+ * @param int $uid
+ *   Optional - the user uid of Reward record to check.
```
sergiitk: Same here.

lib/modules/dosomething/dosomething_reward/dosomething_reward.module
```diff
@@ -24,3 +24,76 @@ function dosomething_reward_entity_info() {
...
+ *   If not given, uses global $user->uid.
+ *
+ * @return mixed
```
sergiitk: It's possible to specify multiple return statements separated by pipe: `object|false`, see [API documentation and comment standards](https://www.drupal.org/node/1354#types).

Or you can simply avoid it, as in `entity_load_single()`:

``` php
/**
 * A wrapper around entity_load() to load a single entity by name or numeric id.
 *
 * @todo: Re-name entity_load() to entity_load_multiple() in d8 core and this
 * to entity_load().
 *
 * @param $entity_type
 *   The entity type to load, e.g. node or user.
 * @param $id
 *   The entity id, either the numeric id or the entity name. In case the entity
 *   type has specified a name key, both the numeric id and the name may be
 *   passed.
 *
 * @return
 *   The entity object, or FALSE.
 *
 * @see entity_load()
 */
function entity_load_single($entity_type, $id) {
  $entities = entity_load($entity_type, array($id));
  return reset($entities);
}
```

lib/modules/dosomething/dosomething_reward/dosomething_reward.module
```diff
@@ -24,3 +24,76 @@ function dosomething_reward_entity_info() {
...
+/**
+ * Creates a Reportback Count Reward for a Reportback, if applicable.
+ *
+ * This function is to be called when a Reportback has been inserted.
+ * @see ReportbackEntityController->save()
+ *
+ * @param obj $reportback
+ *   The newly inserted Reportback entity.
+ *
+ * @return mixed
```
sergiitk: See https://github.com/DoSomething/dosomething/pull/2947/files#diff-62967dede5d2b819281b05f8cff076d0R37

lib/modules/dosomething/dosomething_reward/dosomething_reward.module
```diff
@@ -24,3 +24,76 @@ function dosomething_reward_entity_info() {
...
+  return FALSE;
+}
+
+/**
+ * Creates a Reportback Count Reward for a Reportback, if applicable.
+ *
+ * This function is to be called when a Reportback has been inserted.
+ * @see ReportbackEntityController->save()
+ *
+ * @param obj $reportback
```
sergiitk: Should be `object`, not `obj`,

sergiitk:
@aaronschachter code itself look fine. Would you like me to test in on VM?

aaronschachter:
Sure, that would be super helpful if you don't mind.  You'll have to enable the `dosomething_reward` module (this is an example of a new module we won't be turning on for affiliate sites anytime in the near future)

aaronschachter:
And then can test by reporting back for a campaign, and verifying in the database a Reward is created for a user with 2 or more Reportbacks.  Also, that a Reward is not created again for a user once they've earned a Reward from reporting back.

PR #2948, opened 2014-08-07
Core upgrade

aaronschachter:
Fixes https://jira.dosomething.org/browse/DS-199


aaronschachter:
I've ran this successfully on my local twice now, the first and third times.  The 2nd time I got this error:

Fatal error: Call to undefined function drupal_session_initialize() in /var/www/vagrant/html/includes/bootstrap.inc on line 2254

But I was running into similar errors this week regardless of the version of core.

angaither:
:+1: worked for me! ship it! :shipit:

PR #2949, opened 2014-08-07
Fix PHP notice for Undefined num_participants property 

aaronschachter:
Occurs when a Campaign is set to collect `num_participants` in the Reportback Form.  The property needs to be initialized when we create the blank Reportback entity to save.

Fixes https://jira.dosomething.org/browse/DS-203


blisteringherb:
:+1:

Issue #2466, opened 2014-06-10
Core + Contrib updates


aaronschachter:
Now that we have a versioned makefile (good), we may be stuck with needing to update all modules anytime we want to write our own update hook (ehh) :\

```
vagrant@dev:/var/www/vagrant/html$ drush up
 Name                      Installed Version  Proposed version  Message                   
 Drupal                    7.27               7.28              Update available          
 Fieldgroup (field_group)  7.x-1.3            7.x-1.4           Update available          
 Redis (redis)             7.x-2.6            7.x-2.10          Update available          
 Views (views)             7.x-3.7            7.x-3.8           SECURITY UPDATE available 
```

aaronschachter:
Weirdly enough, answering "no" did exactly what I wanted, skip everything else and run the `dosomething_signup` update hook I'm working on.  Guess that makes sense since we aren't downloading the latest modules automatically.  Can test on pull stage once i'm ready to submit my PR for the signup stuff i'm working on.

```
Do you really want to continue with the update process? (y/n): n
Aborting.                                                                                       [cancel]
 Dosomething_signup  7009  Adds dosomething_signup_presignup table.
Do you wish to run all pending updates? (y/n): y
Performed update: dosomething_signup_update_7009   
```

aaronschachter:
Core updated by #2948. Preparing a PR for contrib updates.

PR #2950, opened 2014-08-07
Contrib security updates

aaronschachter:
Refs #2466 -- updates modules with security updates.

There are more contrib modules to update, but wanted to test each update vs blindly updating -- these are the most critical.  I've tested locally and all seems to be fine.


blisteringherb:
:+1:

PR #2951, opened 2014-08-07
Added reportback and signup global header view totals

angaither:
Yeah, apparently they were manually counting this. Yikes.


aaronschachter:
:+1:

PR #2952, opened 2014-08-07
Adds CDN contrib module to installation profile

aaronschachter:
But does not enable.

This module will only be used on production, not in staging or in local development (since we pull from staging).

Refs https://jira.dosomething.org/browse/DS-109


DFurnes:
:+1:

PR #2954, opened 2014-08-07
Rewards admin view

aaronschachter:
@angaither Can you pls review?

Creates a table view of Reward entities, for internal staff use.  Closes https://jira.dosomething.org/browse/DS-208

screen shot 2014-08-07 at 4 18 54 pm


lib/modules/dosomething/dosomething_reward/dosomething_reward.info
```diff
@@ -3,5 +3,8 @@ description = Provides Reward entity and functionality.
 core = 7.x
 package = DoSomething
 version = 7.x-1.0
-dependencies[] = entity
```
angaither, line 6: probably still need this guy, right?

angaither:
:+1: code looks ok as long as entity is supposed to be removed!

lib/modules/dosomething/dosomething_reward/dosomething_reward.info
aaronschachter, line 6: Ugh dumb `features-update`.  Good catch.  I think it's just going to keep on removing it if I add it back in, but I'll  give it a shot locally to see if it does the same thing.  Apparently we have it as a dependency in Signup but not Reportback.

PR #2955, opened 2014-08-08
Reward Redeem Form

aaronschachter:
Chipping away at https://jira.dosomething.org/browse/DS-127 subtasks.
- Creates the `dosomething_reward_redeem_form` constructor and submit function, which sets the Reward redeemed timestamp when submitted
- Renders the Redeem Form on the Reportback Confirmation page if `dosomething_reward` module is enabled (should still remain disabled while under development)
- Creates admin settings form to set copy for Redeem Form links and modal
- Cleans up Reportback Confirmation theme function -- send an array of variables instead of specifying specific parameters


blisteringherb:
:+1:

PR #2956, opened 2014-08-08
DoSomething Shipment

aaronschachter:
Creates a Shipment entity, and creates a Shipment when a Reward has been redeemed.
- Closes https://jira.dosomething.org/browse/DS-133
- Continues work on redeem form in DS-134
- Closes https://jira.dosomething.org/browse/DS-209 by creating admin table view of Shipments
- Closes https://jira.dosomething.org/browse/DS-144 by only creating Rewards for non-OldPeople


lib/modules/dosomething/dosomething_reward/dosomething_reward.forms.inc
```diff
@@ -14,20 +14,23 @@
...
     '#access' => FALSE,
   );
-  $form['size'] = array(
+  $form['shirt_size1'] = array(
+    '#type' => 'radios',
+    '#required' => TRUE,
+    '#title' => t("Your T-Shirt size"),
+    '#options' => $size_options,
+  );
+  $form['shirt_size2'] = array(
```
blisteringherb: Any thoughts about changing 'shirt_size1' to 'shirt_size', and changing 'shirt_size2' to 'shirt_size_friend'?  When debugging I often find it easier to use more explicit names for form fields.

aaronschachter: I'm open to it, but would need to make the same change to the `dosomething_shipment` table schema as well.

The reason why I specifically chose `shirt_size1` and `shirt_size2` as names over `size` and `size_additional` is what happens if/when we ever need to add a third?   Then it's kind of funky to have a `size_additional2` or something like that.

aaronschachter: .. or, if there was just some sort of use case where the 2nd shirt isn't actually for a friend, but a mentor, or mentoree (is that a word?), or teacher, or whatever.

sergiitk: I agree with @blisteringherb here.

PR #2957, opened 2014-08-08
Test cleanup

DFurnes:
Re-thinking how best to clean up test nodes based on discussion in #2934.
## Changes
- Flags all nodes created from JSON with a `created` value of `-1` so we can find them later.
- Creates a `drush test-node-delete` task to remove all nodes with `created == -1`.

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -5,6 +5,7 @@ include_once 'dosomething_helpers.features.inc';
  * Code for the Dosomething helpers feature.
  */
 
+include_once('dosomething_helpers.drush.inc');
 include_once('dosomething_helpers.theme.inc');
 include_once('dosomething_helpers.strongarm.inc');
 include_once('dosomething_helpers.variable.inc');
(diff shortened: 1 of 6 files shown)


lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -24,6 +25,26 @@ function dosomething_helpers_menu() {
...
+  $query->entityCondition('entity_type', 'node');
+  $query->propertyCondition('created', -1);
+
+  $result = $query->execute();
+  if (isset($result['node'])) {
+    $test_nids = array_keys($result['node']);
+    node_delete_multiple($test_nids);
+
+    return $test_nids;
+  } else {
```
aaronschachter, line 43: Picky line spacing time

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -24,6 +25,26 @@ function dosomething_helpers_menu() {
...
+    return null;
```
aaronschachter, line 44: Picky `NULL` capitalization constant time

aaronschachter:
:+1: Picky things not needed for merge, but worth fixing in a future PR

PR #2958, opened 2014-08-11
DS User Address Form element

aaronschachter:
@angaither Can you please review when you have a chance?

This PR moves the Address Form API elements out of the `dosomething_signup_user_signup_data_form` and into a new function called `dosomething_user_address_form_element` so it can be re-used in the Reward Redeem Form to close out https://jira.dosomething.org/browse/DS-134

I've tested on Thumb Wars to make sure the Address Collection works there, and it's working fine.  JS Validation seems to working on the Reward Redeem Form as well, but the "Fix it" error messages are not displaying.. any clues to why? cc @DFurnes 

Refs https://jira.dosomething.org/browse/DS-141


angaither:
:+1: looks good!

PR #2960, opened 2014-08-11
Remove user email vars from exported variables.

angaither:
This will allow affiliate sites to use default transactional emails, instead of MB emails.

[DS-214](https://jira.dosomething.org/browse/DS-214?filter=-1)


aaronschachter:
:+1:

PR #2961, opened 2014-08-11
Fix ds pull stage: force "yes" on files override dialog.

sergiitk:
Drush only accepts `-y` argument before the command.

With this fix `ds pull stage` will not ask
`You will destroy data from /var/www/vagrant/html/ and replace with data from
dosomething@staging.beta.dosomething.org:/var/www/beta.dosomething.org/current/html//`
`Do you really want to continue? (y/n):`
anymore.

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -369,7 +369,7 @@ function pull_files {
   then
     echo "Pulling down files from staging"
     cd $WEB_PATH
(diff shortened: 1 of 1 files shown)


aaronschachter:
Nice one :+1: ship it value added

PR #2962, opened 2014-08-11
Reward logging

aaronschachter:
If the `dosomething_reward_log` variable is set, log lots of messages around automatic creation of Rewards and Shipments.

Refs https://jira.dosomething.org/browse/DS-219

### DIFF
--- a/lib/modules/dosomething/dosomething_reward/dosomething_reward.module
+++ b/lib/modules/dosomething/dosomething_reward/dosomething_reward.module
@@ -6,6 +6,8 @@
 
 include_once 'dosomething_reward.features.inc';
 include_once 'dosomething_reward.forms.inc';
+
+DEFINE('DOSOMETHING_REWARD_LOG', variable_get('dosomething_reward_log') ? TRUE : FALSE);
 DEFINE('DOSOMETHING_REWARD_NUM_REPORTBACK_COUNT', 2);
 
 /**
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_reward/dosomething_reward.module
```diff
@@ -228,18 +228,45 @@ function dosomething_reward_reportback_count($reportback) {
   $uid = $reportback->uid;
   $account = user_load($uid);
 
+  $log = variable_get('dosomething_reward_log');
+  $log_type = 'dosomething_reward_reportback_count';
```
sergiitk: Would it be more explicit to name it `$log_category`? I understand, [`watchdog()`](https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/watchdog/7) calls it `$type`, but it's also

> `$type`: The **category** to which this message belongs.

lib/modules/dosomething/dosomething_reward/dosomething_reward.module
```diff
@@ -228,18 +228,45 @@ function dosomething_reward_reportback_count($reportback) {
   $uid = $reportback->uid;
   $account = user_load($uid);
 
+  $log = variable_get('dosomething_reward_log');
```
sergiitk: I'd make it more clear:

``` php
$log = variable_get('dosomething_reward_log', FALSE) ? TRUE : FALSE;
```

lib/modules/dosomething/dosomething_reward/includes/dosomething_reward.inc
```diff
@@ -21,7 +21,13 @@ class RewardEntity extends Entity {
    * Sets the Reward as redeemed and creates a relevant Shipment if applciable.
    */
   public function redeem($values = array()) {
+    $log = variable_get('dosomething_reward_log');
```
sergiitk: Same here:

``` php
$log = variable_get('dosomething_reward_log', FALSE) ? TRUE : FALSE;
```

lib/modules/dosomething/dosomething_reward/dosomething_reward.module
sergiitk: Consider defining it as a constant on the top of `dosomething_reward.module`.
I doubt it'll change its value during script runtime.

Thus you'll be able to reuse it in `dosomething_reward.inc`.

aaronschachter: Yeah that's a really good idea.

sergiitk:
Great! :+1:

PR #2963, opened 2014-08-11
New SMS Game variables

aaronschachter:
Work done for subtasks in https://jira.dosomething.org/browse/DS-206
- Adds a `dosomething_signup_sms_game_multiplayer_endpoint` variable to store the URL to post multi-player SMS Games to.
- Adds two new helper variables, `sms_game_mp_story_id` and `sms_game_mp_story_type` for Multi-Player SMS Game campaign nodes.  
- Refactors the SMS Game helper variable set/get opt-in form to use an array of SMS Game variables as (similar to the array of standard Campaign helper variables, to make it easier to add new variables

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
@@ -8,7 +8,7 @@
 /**
  * Form constructor for Signup opt-in configuration.
  */
-function dosomething_signup_optin_config_form($form, &$form_state) {
+function dosomething_signup_opt_in_config_form($form, &$form_state) {
 
   // General Drupal variables.
   $name = 'dosomething_mobilecommons_opt_in_path_user_register';
@@ -33,8 +33,8 @@ function dosomething_signup_optin_config_form($form, &$form_state) {
   );
 
   // Node helpers variables.
-  $optin_helpers = dosomething_signup_get_optin_config_form_helpers_vars();
-  foreach ($optin_helpers as $key => $vars) {
+  $opt_in_helpers = dosomething_signup_get_opt_in_config_form_helpers_vars();
+  foreach ($opt_in_helpers as $key => $vars) {
     // Output group header:
     $form[$key] = array(
       '#markup' => '<h3>' . $vars['title'] . '</h3>',
@@ -46,7 +46,7 @@ function dosomething_signup_optin_config_form($form, &$form_state) {
       // Loop through each node:
       foreach ($vars['node_vars'] as $node_vars) {
         // Add a form element for the node's helpers variables.
-        dosomething_signup_optin_config_form_add_elements($form, $node_vars, $helpers_vars);
+        dosomething_signup_opt_in_config_form_add_elements($form, $node_vars, $helpers_vars);
       }
     }
   }
@@ -59,9 +59,9 @@ function dosomething_signup_optin_config_form($form, &$form_state) {
 }
 
 /**
- * Submit handler for dosomething_signup_optin_config_form().
+ * Submit handler for dosomething_signup_opt_in_config_form().
  */
-function dosomething_signup_optin_config_form_submit($form, &$form_state) {
+function dosomething_signup_opt_in_config_form_submit($form, &$form_state) {
   $values = $form_state['values'];
 
   // Save general Drupal variables.
@@ -74,10 +74,12 @@ function dosomething_signup_optin_config_form_submit($form, &$form_state) {
   }
 
   // Save node helpers variables.
-  $optin_helpers = dosomething_signup_get_optin_config_form_helpers_vars();
-  foreach ($optin_helpers as $vars) {
+  $opt_in_helpers = dosomething_signup_get_opt_in_config_form_helpers_vars();
+  foreach ($opt_in_helpers as $vars) {
     // If no nodes exist, go to next group.
-    if (empty($vars['node_vars'])) { continue; }
+    if (empty($vars['node_vars'])) {
+      continue;
+    }
 
     // Loop through each node:
     foreach ($vars['node_vars'] as $node_vars) {
@@ -108,9 +110,9 @@ function dosomething_signup_optin_config_form_submit($form, &$form_state) {
  * @param array $helpers_vars
  *   The helpers variable names to get/set for this node.
  */
-function dosomething_signup_optin_config_form_add_elements(&$form, $node_vars, $helpers_vars) {
+function dosomething_signup_opt_in_config_form_add_elements(&$form, $node_vars, $helpers_vars) {
   $nid = $node_vars['nid'];
-  $title = isset($node_vars['title']) ? $node_vars['title'] : NULL;
+  $title = !empty($node_vars['title']) ? $node_vars['title'] : t("Undefined");
   // Create a fieldset for each campaign.
   $form[$nid] = array(
     '#type' => 'fieldset',
@@ -157,7 +159,7 @@ function dosomething_signup_get_signup_helpers_var_names() {
  *
  * @return array
  */
-function dosomething_signup_get_optin_config_form_helpers_vars() {
+function dosomething_signup_get_opt_in_config_form_helpers_vars() {
   $vars = array();
   $helpers_var_names = dosomething_signup_get_signup_helpers_var_names();
   $vars['staff_pick'] = array(
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
@@ -62,6 +62,7 @@ function dosomething_signup_optin_config_form($form, &$form_state) {
 function dosomething_signup_optin_config_form_submit($form, &$form_state) {
   $values = $form_state['values'];
 
+  // Save general Drupal variables.
   $config_variables = array(
     'dosomething_mobilecommons_opt_in_path_user_register',
     'dosomething_mobilecommons_opt_in_path_general_campaign_signup',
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
```diff
@@ -29,49 +31,26 @@ function dosomething_signup_optin_config_form($form, &$form_state) {
...
+    $form[$key] =array(
+      '#markup' => '<h3>' . $vars['title'] . '</h3>',
+    );
+    // If nodes exist for this group:
+    if (!empty($vars['node_vars'])) {
+      // Get the helpers_vars names for this group.
+      $helpers_vars = $vars['helpers_vars'];
+      // Loop through each node:
+      foreach ($vars['node_vars'] as $node_vars) {
+        $type = $vars['form_element_type'];
```
aaronschachter: Just realized this isn't used. About to push a cleanup commit.

lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
```diff
@@ -29,49 +31,26 @@ function dosomething_signup_optin_config_form($form, &$form_state) {
...
-      $form[$nid][$nid . '_beta'] = array(
-        '#type' => 'textfield',
-        '#title' => t('Mobile Commons Beta Opt-In Path'),
-        '#default_value' => dosomething_helpers_get_variable($nid, 'mobilecommons_friends_opt_in_path'),
-      );
+
+  // Node helpers variables.
+  foreach (dosomething_signup_get_optin_config_form_helpers_vars() as $key => $vars) {
+    // Output group header:
+    $form[$key] =array(
```
sergiitk: Missed whitespace after `=` sign.

lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
```diff
@@ -29,49 +31,26 @@ function dosomething_signup_optin_config_form($form, &$form_state) {
...
-        '#default_value' => dosomething_helpers_get_variable($nid, 'mobilecommons_opt_in_path'),
-      );
-      $form[$nid][$nid . '_beta'] = array(
-        '#type' => 'textfield',
-        '#title' => t('Mobile Commons Beta Opt-In Path'),
-        '#default_value' => dosomething_helpers_get_variable($nid, 'mobilecommons_friends_opt_in_path'),
-      );
+
+  // Node helpers variables.
+  foreach (dosomething_signup_get_optin_config_form_helpers_vars() as $key => $vars) {
```
sergiitk: It may not be the case, but generally I'd avoid passing function call to the foreach statement.
This prevents the ambiguity if somebody decided to change the value by its reference: `…  as $key => &$vars) {`.

lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
```diff
@@ -29,49 +31,26 @@ function dosomething_signup_optin_config_form($form, &$form_state) {
...
+    // Output group header:
+    $form[$key] =array(
+      '#markup' => '<h3>' . $vars['title'] . '</h3>',
+    );
+    // If nodes exist for this group:
+    if (!empty($vars['node_vars'])) {
+      // Get the helpers_vars names for this group.
+      $helpers_vars = $vars['helpers_vars'];
```
sergiitk: In shortcut variables it's better to use reference assignment so, for example, an array will not be copied by accident. It will populate memory stack and may cause memory limit overflow:

``` php
$helpers_vars = &$vars['helpers_vars'];
```

Especially with array of nodes.

lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
```diff
@@ -93,69 +73,52 @@ function dosomething_signup_optin_config_form_submit($form, &$form_state) {
...
-      $nid = $vars['nid'];
-      $node = node_load($nid);
-      foreach ($variables as $name => $description) {
-        $value = $values[$nid . '_' . $name];
-        dosomething_helpers_set_variable($node, $name, $value);
-      }
-    }
-  }
+  // Save node helpers variables.
+  foreach (dosomething_signup_get_optin_config_form_helpers_vars() as $vars) {
```
sergiitk: Same here.

lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
```diff
@@ -93,69 +73,52 @@ function dosomething_signup_optin_config_form_submit($form, &$form_state) {
...
+    // If no nodes exist, go to next group.
+    if (empty($vars['node_vars'])) continue;
```
sergiitk: > Always use curly braces even in situations where they are technically optional.
> https://www.drupal.org/coding-standards#controlstruct

lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
```diff
@@ -93,69 +73,52 @@ function dosomething_signup_optin_config_form_submit($form, &$form_state) {
...
-  $title = $vars['title'];
-  $variables = dosomething_signup_get_admin_config_campaign_variables();
-
+function dosomething_signup_optin_config_form_add_elements(&$form, $node_vars, $helpers_vars) {
+  $nid = $node_vars['nid']; 
   // Create a fieldset for each campaign.
   $form[$nid] = array(
     '#type' => 'fieldset',
-    '#title' => $title . ' | NID ' . $nid,
+    '#title' => $node_vars['title'] . ' | NID ' . $nid,
```
sergiitk: Check whether `$node_vars['title']` is empty?

lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
```diff
@@ -166,23 +129,73 @@ function dosomething_signup_optin_config_form_add_elements(&$form, $vars) {
...
+  $form['sms_game'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('SMS Game'),
+  );
+  $var_name = 'dosomething_signup_sms_game_multiplayer_endpoint';
+  $form['sms_game'][$var_name] = array(
+    '#type' => 'textfield',
+    '#required' => TRUE,
+    '#title' => t('SMS Multi-player Game Endpoint'),
+    '#description' => t("URL to post a multi-player SMS Game request, e.g. <em>http://ds-heroku.biz/sms-multiplayer-game/create</em>"),
```
sergiitk, line 197: Does it show `<em>` correctly?

lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
```diff
@@ -166,23 +129,73 @@ function dosomething_signup_optin_config_form_add_elements(&$form, $vars) {
 }
 
 /**
- * Returns list of helper variables that a standard Campaign can implement.
+ * Returns multi-dimensional array of relevant Signup helpers variable names.
  *
  * @return array
- *   Associative array keyed by variable name.
+ *   Returns multi-dimensional array, keyed by the Campaign type.
```
sergiitk: `keyed` → `indexed`?

lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
aaronschachter, line 197: Yep, it does.
screen shot 2014-08-12 at 10 56 20 am

sergiitk, line 197: Good! Just double-checking.

lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
aaronschachter: good call. will make a bunch of these changes in one commit.

lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
aaronschachter: This is actually just an array of Node nid's and titles, not nodes. but i will make this change too

lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
```diff
@@ -29,49 +31,26 @@ function dosomething_signup_optin_config_form($form, &$form_state) {
...
-        '#default_value' => dosomething_helpers_get_variable($nid, 'mobilecommons_opt_in_path'),
-      );
-      $form[$nid][$nid . '_beta'] = array(
-        '#type' => 'textfield',
-        '#title' => t('Mobile Commons Beta Opt-In Path'),
-        '#default_value' => dosomething_helpers_get_variable($nid, 'mobilecommons_friends_opt_in_path'),
-      );
+
+  // Node helpers variables.
+  $optin_helpers = dosomething_signup_get_optin_config_form_helpers_vars();
```
sergiitk: `optin` is it correct?

lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
```diff
@@ -93,59 +73,44 @@ function dosomething_signup_optin_config_form_submit($form, &$form_state) {
...
-        $value = $values[$nid . '_' . $name];
-        dosomething_helpers_set_variable($node, $name, $value);
-      }
-    }
-  }
+  // Save node helpers variables.
+  $optin_helpers = dosomething_signup_get_optin_config_form_helpers_vars();
+  foreach ($optin_helpers as $vars) {
+    // If no nodes exist, go to next group.
+    if (empty($vars['node_vars'])) { continue; }
```
sergiitk: Drupal says,

> https://www.drupal.org/coding-standards#controlstruct
> The opening curly should be on the same line as the opening statement, preceded by one space. The closing curly should be on a line by itself and indented to the same level as the opening statement.

But I'm OK with that if this is common in DS code.

lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
```diff
@@ -93,59 +73,44 @@ function dosomething_signup_optin_config_form_submit($form, &$form_state) {
...
+ *   The helpers variable names to get/set for this node.
  */
-function dosomething_signup_optin_config_form_add_elements(&$form, $vars) {
-  $nid = $vars['nid'];
-  $title = $vars['title'];
-  $variables = dosomething_signup_get_admin_config_campaign_variables();
-
+function dosomething_signup_optin_config_form_add_elements(&$form, $node_vars, $helpers_vars) {
+  $nid = $node_vars['nid'];
+  $title = isset($node_vars['title']) ? $node_vars['title'] : NULL;
```
sergiitk: It's better to pass empty string instead of `NULL`:
1. This way it's more clear that variable type is string
2. `NULL` will not be concatenated with string. <small>It may produce strict warning, but I'm not 100% sure.</small>

On the other hand, I would even go with

``` php
$title = !empty($node_vars['title']) ? $node_vars['title'] : t("Undefined");
```

This way it's easier to pinpoint the error. As for me, title `| NID 42` looks worse than `Undefined | NID 42`.
In addition `empty()` will also cover empty titles. I assume empty title is a sign that something went wrong as well.

lib/modules/dosomething/dosomething_signup/dosomething_signup.admin.inc
aaronschachter: It's always been that way, but yeah I think this could use a rename as `opt_in`.

sergiitk: That's up to you. I'm not a native English speaker. I thought it may mean something I don't understand.

sergiitk:
Great! :+1: Thank you for fixing all small styling issues!

PR #2964, opened 2014-08-12
Fixing translation error on SMS link that happened during global site tr...

weerd:
...anslation.

Quick fix for that bug with the translation of link in the SMS page.


sergiitk:
:+1:

weerd:
thanks! :tropical_drink:

PR #2965, opened 2014-08-12
Removing the  var from campaign thumbnail template since we don't use it...

weerd:
... and it's not being populated with anything anyways.

Fixy-fix-fix :goat: 

@aaronschachter @barryclark 

Fixes https://jira.dosomething.org/browse/DS-202


aaronschachter:
:+1: Please keep an eye on the log table!

weerd:
Will do :smile:

PR #2966, opened 2014-08-12
Fixed weird bug with Plan It section when action guides are added. It wa...

weerd:
...s dumb.

Quick fix! I was being dumb. Should resolve the layout issues in Plan It section.

@aaronschachter @angaither

Thanks Aaron for the help on finding the bug! :beer: 

EDIT: Jira issue: https://jira.dosomething.org/browse/DS-120


aaronschachter:
:+1: There's a GH issue or a Jira issue for this right?

PR #2968, opened 2014-08-12
Use the slave server for reportback view queries.

angaither:
Exported this feature that's already being used in production.

[DS-192](https://jira.dosomething.org/browse/DS-192)


aaronschachter:
:+1:

deezone:
For what it's worth, specific queries can specify the use of the slave database. An example from the Optimizely module:

```
  $query = db_select('optimizely', 'o', array('target' => 'slave'))
    ->fields('o')
    ->condition('o.enabled', 1, '=')
    ->orderBy('oid');
```

I suspect this is the same thing as what you're doing through Views.

PR #2969, opened 2014-08-12
Multi-player SMS Game signup

aaronschachter:
- Renames `dosomething_signup_friends_form` to `dosomething_signup_sms_game_form`
- Sets hidden form values for a Multi-player SMS Game if the relevant opt-in variables are set.
- Sets hidden form value for the type of SMS Game (`single_player` or `multi_player`)
- Creates `dosomething_signup_sms_game_multi_player_signup_request` function to post a request to the `dosomething_signup_sms_game_multiplayer_endpoint` endpoint.
- Adds constants for Signup logging.

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -682,3 +682,58 @@ function dosomething_signup_get_signup_total_by_nid($nid) {
     ->execute();
   return $result->rowCount();
 }
+
+/**
+ * Handles a SMS Game Signup Form signup.
+ *
+ * @param string $sms_game_type
+ *   Expected values: 'single_player' | 'multi_player'.
+ * @param array $values
+ *   Values to use for the signup request.
+ *   @see dosomething_signup_sms_game_form()
+ */
+function dosomething_signup_sms_game_signup_request($sms_game_type, $values) {
+  if ($sms_game_type == 'single_player') {
+    // Send opt_in request to Mobilecommons API.
+    return dosomething_signup_mobilecommons_opt_in_friends($values);
+  }
+  // Default to multi-player signup:
+  dosomething_signup_sms_game_multi_player_signup_request($values);
+}
+
+/**
+ * Posts to to the Multi-player SMS Game URL with relevant $values.
+ *
+ * @param array @values
+ */
+function dosomething_signup_sms_game_multi_player_signup_request($values) {
+  // Initialize post data.
+  $data = array();
+  $data_vars = array(
+    'alpha_first_name',
+    'alpha_mobile',
+    'beta_mobile_0',
+    'beta_mobile_1',
+    'beta_mobile_2',
+    'story_id',
+    'story_type',
+  );
+  foreach ($data_vars as $var) {
+    $data[$var] = $values[$var];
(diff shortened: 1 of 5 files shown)


jonuy:
I feel like I can't be as thorough as Sergii. But from what I can tell, I like it.

lib/modules/dosomething/dosomething_signup/includes/dosomething_signup.inc
```diff
@@ -80,7 +80,7 @@ class SignupEntityController extends EntityAPIController {
       }
     }
     parent::save($entity, $transaction);
-    if (variable_get('dosomething_signup_log_signups')) {
+    if (DOSOMETHING_SIGNUP_LOG_SIGNUPS) {
```
angaither, line 83: I think we had problems with using a constant in an include file before. I vaguely remember this....

angaither:
But if it's working :+1:

lib/modules/dosomething/dosomething_signup/includes/dosomething_signup.inc
aaronschachter, line 83: There was something that broke once because I defined a constant without the quotes... maybe that's what you are thinking of.

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -679,3 +682,58 @@ function dosomething_signup_get_signup_total_by_nid($nid) {
...
+    $data[$var] = $values[$var];
+  }
+  $options = array(
+    'method' => 'POST',
+    'data' => drupal_http_build_query($data),
+    'timeout' => 60,
+    'headers' => array('Content-Type' => 'application/x-www-form-urlencoded'),
+  );
+
+  $url = variable_get('dosomething_signup_sms_game_multiplayer_endpoint');
```
blisteringherb, line 731: Will the default value for this var live in settings.php or strongarm?

aaronschachter, line 731: That's an interesting question, I was just assuming this would only be stored in the DB. 

@jonuy are there any security reasons to avoid storing the URL in code ? (keeping in mind the repo should hopefully be open source soon) 

 Also FYI @blisteringherb the URL has not been finalized yet, stage is set up to point to @jonuy 's heroku instance.

jonuy, line 731: @aaronschachter I suppose it could be abused if someone wanted to. By knowing the URL, they could constantly POST to it that would then spam users each time. But... I suppose that's a potential problem we already face anyway.

aaronschachter, line 731: Yeah, i'm fine to just keep this in the DB unless you have objections @blisteringherb .  I am assuming there are no plans in the near future to implement this for affiliates.

@jonuy we would just need to manually change the variable in production and in staging when the URL is finalized

jonuy, line 731: Safe assumption. There are no near future plans to bring this to our affiliates.

And as far as finalizing the URL, we'll have that in place some time tomorrow after Desmond deploys to nodejitsu.

PR #2970, opened 2014-08-13
Added indicies to reportback tables for faster views.

angaither:
# closes DS-192


blisteringherb:
:+1:

aaronschachter:
Should we update the schema definition too? i guess it's kinda picky but feels a little more best practice / cleaner for any new instances.  Not really necessary for merge tho

PR #2971, opened 2014-08-13
Shipment item and additional_item properties

aaronschachter:
Closes https://jira.dosomething.org/browse/DS-232

Removes `shirt_size` and `shirt_size_additional` columns and adds the following columns:
- item
- item_option
- item_additional
- item_additional_option

Also removes the `shipment_type`, as the `item` column will be used to store what item is being shipped.

These changes are done in the `dosomething_shipment.install` file because it hasn't been enabled on stage yet.
- Updates the `user_shipments` view with new columns
- Updated Reward Redeem Form to prompt for all new columns and create the Shipment record with submitted values


angaither:
:+1:

PR #2972, opened 2014-08-13
moved duo gallery to be global

### LINKED ISSUE #2749: Static Content gallery: Theme based on Gallery Style field value

tjrosario:
Fixes #2749


weerd:
Cooool :+1:  And yeah, we should eventually move to Neue as well :beers:

PR #2973, opened 2014-08-14
DoSomething API

aaronschachter:
This PR moves our exported Services endpoint into a new module, `dosomething_api` and out of `dosomething_settings`.  

Affiliate sites don't need the API endpoint, so this should only be enabled on the US site.  I will enable manually on staging and production upon deployment.

Next steps:
- move all custom resources hooks into this module (instead of defining separately in `dosomething_reportback` and `dosomething_user`)
- define custom endpoint for a campaigns index: https://jira.dosomething.org/browse/DS-221


lib/modules/dosomething/dosomething_api/dosomething_api.services.inc
```diff
@@ -47,18 +47,6 @@ function dosomething_settings_default_services_endpoint() {
...
-        'retrieve' => array(
-          'enabled' => '1',
-        ),
-      ),
-      'actions' => array(
-        'save' => array(
-          'enabled' => '1',
-        ),
-      ),
-    ),
```
angaither, line 61: Just checking that this should be removed...

angaither:
:+1: looks ok. does anything on the lobby dashboard side need to change?

lib/modules/dosomething/dosomething_api/dosomething_api.services.inc
aaronschachter, line 61: Yup, it's on purpose.  The reportback endpoint was just a prototype that we're not using for anything yet, so there's no need to enable it.

aaronschachter:
Nope all of the endpoints are the same, the only gotcha is just enabling this module manually when we deploy to production today (since we don't have affiliate specific module lists yet)

angaither:
Yeah, we should make that.

aaronschachter:
Agree.  There's an issue for it here https://jira.dosomething.org/browse/INF-16

PR #2976, opened 2014-08-14
"US"  specific module deployment

angaither:
This is not ready for merge, but here's an idea for site specific deployments.
- gives the ability to turn on/off modules with a deployment 
- the ability to have affiliate specific modules (currently only the file for :us: exists)
- the `key` is the affiliate name, passed as an arg with the drush script to get the correct site-specific modules


angaither:
WHY DIDN'T MY EMOJI PR TITLE WORK :no_good: :warning:

bin/default.modules.json
```diff
@@ -0,0 +1,81 @@
...
+  "simpletest" : 1,
+  "taxonomy" : 1,
+  "addressfield" : 1,
+  "admin_menu" : 1,
+  "admin_menu_toolbar" : 1,
+  "apachesolr" : 1,
+  "apachesolr_search" : 1,
+  "apachesolr_views" : 1,
+  "blockreference" : 1,
+  "conductor" : 1,
```
aaronschachter: FYI - A few of these modules like `conductor`, `conductor_sms`, `mobilecommons`, `dosomething_sms`, won't be needed by default.  Working on a specific list in that Jira issue but it's not finalized yet (so maybe hold off on making these changes until it is)

bin/ds
```diff
@@ -162,7 +162,7 @@ function build {
 
   # Enable modules in info file
   echo 'Enabling modules in dosomething.info'
```
aaronschachter, line 164: Minor, but we'll want to change these comments and output to signify that we're not using `dosomething.info` anymore.

aaronschachter:
This looks good to me, just minor comments. Assuming you've tested and it's worked for local `ds build` ?

One thing to consider is that all the affiliate sites will already have modules they don't need enabled. So, for a module like `dosomething_sms`... currently we would need to manually disable this in every instance.  What's a bummer is that there a few modules related to it (`conductor`, `conductor_sms`).  Also `message_broker_producer` and `dosomething_mbp` don't need to be on currently although there are talks of affiliate MBP stuff going on...

deezone:
Should the optimizely module be enabled by default? I'm guessing they'll have no use for it initially.

bin/default.modules.json
angaither: No problem, we can remove that stuff.

angaither:
@aaronschachter re: disabling each module manually-- we can have a key for all affiliates (as they are all the same right now) and set the status in the json file to 0 and that will disable it for all of them

aaronschachter:
@dee The related issue is https://jira.dosomething.org/browse/INF-16, let's work on keeping the list there vs this PR.

@angaither That sounds good.  I guess the only gotcha could be when we want to enable a module but not do so in code.. but I'm not really sure if we'd ever want that actually.

PR #2977, opened 2014-08-15
Added an index to the nid table on dosomething-signup

angaither:
Closes [DS-167](https://jira.dosomething.org/browse/DS-167)


aaronschachter:
:+1:

PR #2979, opened 2014-08-15
Added cache for dosomething_reportback table.

angaither:
Caches the reportback galleries.
The `dosomething_reportback_get_gallery_item_ids` query was showing up in the slow query new relic log.

@aaronschachter can you pull this down and make sure it creates/stores things properly in the table?

still todo: clear the cache on flagging.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
@@ -237,6 +237,8 @@ function dosomething_reportback_schema() {
       'rbid' => array('rbid'),
     ),
   );
+  // Add a custom cache table.
+  $schema['cache_dosomething_reportback'] = drupal_get_schema_unprocessed('system', 'cache');
   return $schema;
 }
 
@@ -378,4 +380,12 @@ function dosomething_reportback_update_7009() {
    if (!db_index_exists($table, $column)) {
     db_add_index($table, $column, array($column));
   }
+ }
+
+/**
+ * Creates dosomething reportback cache table.
+ */
+ function dosomething_reportback_update_7011() {
+   $schema = drupal_get_schema_unprocessed('system', 'cache');
+   db_create_table('cache_dosomething_reportback', $schema);
  }
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
@@ -386,6 +386,8 @@ function dosomething_reportback_update_7009() {
  * Creates dosomething reportback cache table.
  */
  function dosomething_reportback_update_7011() {
-   $schema = drupal_get_schema_unprocessed('system', 'cache');
-   db_create_table('cache_dosomething_reportback', $schema);
+   if (!db_table_exists('cache_dosomething_reportback')) {
+     $schema = drupal_get_schema_unprocessed('system', 'cache');
+     db_create_table('cache_dosomething_reportback', $schema);
+   }
  }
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
```diff
@@ -379,3 +381,11 @@ function dosomething_reportback_update_7009() {
     db_add_index($table, $column, array($column));
   }
  }
+
+/**
+ * Creates dosomething reportback cache table.
+ */
+ function dosomething_reportback_update_7011() {
+   $schema = drupal_get_schema_unprocessed('system', 'cache');
+   db_create_table('cache_dosomething_reportback', $schema);
```
aaronschachter: Doesn't this need a `db_table_exists` check to avoid "Table exists" errors on new installations?

aaronschachter:
Pulling down to test.

aaronschachter:
I've pulled down and ran `drush updb`, creating the `cache_dosomething_reportback` table.  I've visited node/362 and node/850 locally but no cache records are being created in the table.  Is there an extra step I'm missing for testing?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
```diff
@@ -237,6 +237,8 @@ function dosomething_reportback_schema() {
       'rbid' => array('rbid'),
     ),
   );
+  // Add a custom cache table.
+  $schema['cache_dosomething_reportback'] = drupal_get_schema_unprocessed('system', 'cache');
```
aaronschachter, line 241: Does it make sense to rename this to `cache_dosomething_reportback_gallery`?  This table name suggests it would be caching the Reportback entities.

angaither, line 241: yeah that was intentional, I thought we might want to start caching those as well.

angaither:
the cache is being pulled from redis, not in the drupal database.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
aaronschachter, line 241: So.. then should we store the gallery cache in a separate `cache_dosomething_reportback_gallery` table?

aaronschachter:
Testing by adding a `dsm($cache->data)`, but the cache isn't being updated upon cache clear.  Needs stuff like this? https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_image/dosomething_image.module#L168

But without the cache clearing in place upon flagging, this will make gallery stuff somewhat un-usable for the campaigns team in the meantime.

angaither:
Yeah, that's what I'm working on now. :pencil2:

angaither:
@aaronschachter added drush command & clearing once a new reportback image is added to a gallery

aaronschachter:
Pulled down and tested by using a `dsm($cache->data)` to confirm cached data.  Looks good :+1:

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -872,3 +879,20 @@ function dosomething_reportback_services_resources() {
   );
   return $resources;
 }
+
+/**
+ * Implements hook_flag_flag().
+ */
+function dosomething_reportback_flag_flag($flag, $entity_id, $account, $flagging) {
+    if ($flag->name == 'promoted') {
+      dosomething_reportback_cache_clear_all();
```
aaronschachter, line 888: Not urgent, but we should potentially only clear the Reportback Gallery cache for the $nid (which we could calculate through the $entity_id).

PR #2980, opened 2014-08-18
Updated image style for 3col galleries.

angaither:
Pull smaller images when there are more images.

Refs DS-252
cc: @tjrosario the css is stretching these to 100% width and making them pixely


aaronschachter:
:+1:

DFurnes:
Seems cool. My suggestion is that we just switch `width: 100%` to `max-width: 100%`.

weerd:
Agreed! I think that sounds right @DFurnes, especially since going with a more fixed height on some of these images within the gallery.

DFurnes:
Neue patch merged and released, so should be good for next staging deploy (or `ds grunt` on local).

@angaither Should be safe to merge! Thanks for the css!

PR #2983, opened 2014-08-19
Reward page / form template

aaronschachter:

This PR will unblock fenders from theming the Reward Redeem page / form. 
- Adds tpl file for the Redeem Reward page / modal, moving it out of the Confirmation page tpl.
- Only displays the page/form if the Reward has not been redeemed
- Adds lots of variables for the Reward page/modal, including Shirt image nodes.
- Only creates Reportback Count Rewards if the `dosomething_reward_enable_reportback_count` variable is set, to allow enabling/disabling without any code changes.


sergiitk:
:+1: Awesome!

PR #2984, opened 2014-08-19
Removing capistrano deployment files

blisteringherb:
Removing capistrano deployment files from DS repo.  

Capistrano scripts now live in their own repo here: https://github.com/DoSomething/dosomething-deploy

Fixes: https://jira.dosomething.org/browse/DS-72

@mshmsh5000


aaronschachter:
seems ok here :+1:

mshmsh5000:
Love it! +1

We need to reconfigure the Jenkins job now, eh?

blisteringherb:
I think the Jenkins job just needs the github URLs updated to point to the https://github.com/DoSomething/dosomething-deploy repo, correct?  The project URL for the code that needs to be deployed is in deploy.rb.

Issue #1537, opened 2014-03-26
Reportback delete form - bugs


aaronschachter:
Minor issue as the form is for admin only.  The delete reportback form does indeed delete it from the database, but this happens too

![screen shot 2014-03-26 at 12 18 09 pm](https://f.cloud.github.com/assets/1236811/2527199/666b90f6-b502-11e3-8775-ebaf60d0cd91.png)

PR #2985, opened 2014-08-19
Fix errors on Reportback Delete form.

aaronschachter:
Fixes an oldie but a goodie, #1537.  

Only staff can delete reportbacks, so this was never user facing.  Keeps the existing logic though in case one day users ever can delete their own reportbacks.


angaither:
:+1:

PR #2986, opened 2014-08-19
Fixes Reward Form address validation

aaronschachter:
Fixes https://jira.dosomething.org/browse/DS-141

Refs #2958


DFurnes:
Looks solid. :elephant:

PR #2987, opened 2014-08-19
Re-organize test directory.

DFurnes:
## Changes
- Allow alternative CasperJS URL to be set using an environment variable for running tests on QA.
- Adds XUnit XML output from CasperJS tests back in, for pretty visualizations on Jenkins. Outputted XML is stored in the ignored `tmp/` directory in project root.
- Moves integration tests into an `integration` folder within `tests/`, to make way for unit tests to live here as well.

For review: @angaither @aaronschachter or @blisteringherb :racehorse:


angaither:
I like. :elephant: :+1:

PR #2991, opened 2014-08-20
Fixes PHP notices

aaronschachter:
@angaither Please review.

Fixes PHP notices as reported in https://jira.dosomething.org/browse/DS-233


angaither:
:+1: :shipit:

lib/modules/dosomething/dosomething_user/dosomething_user.address.inc
```diff
@@ -93,8 +93,10 @@ function dosomething_user_address_form_element_validate($form, &$form_state) {
 function dosomething_user_address_form_element_set_values($form, &$form_state, $address) {
   form_set_value($form['user_address']['address']['street_block']['thoroughfare'],
       array('value' => $address['thoroughfare']), $form_state);
-  form_set_value($form['user_address']['address']['street_block']['premise'],
+  if (isset($value['premise'])) {
```
aaronschachter: Whoops this needs to be `$address['premise']`

PR #2992, opened 2014-08-20
Clears QA IP address from flood table before running tests.

DFurnes:
## Changes

Tests were failing (after a few consecutive runs) on QA because of the flood table. This stops it from doing that by clearing the QA IP address before running tests.


aaronschachter:
:+1:

PR #2995, opened 2014-08-20
Signup Data Form - Shipment Item

aaronschachter:
Adds a `shipment_item` field to the Signup Data Form table, and makes it editable from the Signup Data Form configuration form.

Closes https://jira.dosomething.org/browse/DS-259


sergiitk:
:+1:

PR #2998, opened 2014-08-21
Remove Reportback Field tables and code

aaronschachter:
Closes https://jira.dosomething.org/browse/DS-131

This functionality was added for very few campaigns and adds a lot of complexity to the codebase that we've determined we no longer need.


angaither:
:+1: :shipit:

PR #2999, opened 2014-08-21
DS-262 Allow repeating fives in phone numbers.

DFurnes:
### Changes
- No longer fails validation on phone numbers with `555` in them.

Fixes [DS-262](https://jira.dosomething.org/browse/DS-262).

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -145,7 +145,7 @@ function dosomething_user_get_staff_only_types() {
 }
 
  /**
-  * Confirms that a specific mobile phone number is valid.
+  * Confirms that the provided string is a valid US phone number.
   *
   * A valid phone number is a number with either 10 or 11 digits
   * (as long as the first digit is a "1").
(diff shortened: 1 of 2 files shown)


aaronschachter:
:+1:

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -177,7 +177,7 @@ function dosomething_user_get_staff_only_types() {
 function dosomething_user_valid_mobile($number) {
   preg_match('#^(?:\+?1([\-\s\.]{1})?)?\(?([0-9]{3})\)?(?:[\-\s\.]{1})?([0-9]{3})(?:[\-\s\.]{1})?([0-9]{4})#', $number, $valid);
   preg_match('#([0-9]{1})\1{9,}#', preg_replace('#[^0-9]+#', '', $number), $repeat);
-  return !empty($valid) && empty($repeat) && strpos($number, '555') === FALSE;
```
aaronschachter, line 180: I suppose this logic could be changed to only invalidate when the 555 are the first three numbers and not within the last... but... i guess it depends on how many bogus numbers we'll get.  Doesn't bug me too much, but something to consider and possibly ask data team?

DFurnes, line 180: From doing a little bit of research on it yesterday, it actually seems like `555-xxxx` numbers have been available for public use post-1993. Which must suck, since I feel like nobody would think your number was real.

aaronschachter, line 180: All those members that couldn't get SMS from us :frowning:

PR #3005, opened 2014-08-21
Pass those tests!

DFurnes:
:banana: 

/cc @deadlybutter


aaronschachter:
:+1: Good work.

itsjoekent:
Just read up on the difference between the two, I've seen === before but never knew the point

PR #3009, opened 2014-08-21
DS-265 Fix issue where form validation was not showing up on T-Shirt form.

DFurnes:
Fixes [DS-265](https://jira.dosomething.org/browse/DS-265).

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
@@ -309,12 +309,5 @@
       }
     }
   }
-
-  // @TODO: remove once radio button validation is properly added to Neue
-  .error {
-    .label {
-      display: none;
-    }
-  }
 }
 


DFurnes:
For review: @tjrosario

lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
```diff
@@ -309,12 +309,5 @@
       }
     }
   }
-
-  // @TODO: remove once radio button validation is properly added to Neue
-  .error {
-    .label {
-      display: none;
```
tjrosario, line 316: Do the error messages still appear for the shirt style radio's?  This was a temp override for them because the label markup is slightly different from normal required fields.

DFurnes, line 316: Going to sidestep that by just pre-selecting one of the options on initial load. The problem with hiding the error labels is that they were disappearing on the address form above as well, so users couldn't see errors when they were happening.

tjrosario, line 316: ok +1

PR #3013, opened 2014-08-21
added support to select any default checked radios

tjrosario:
checks to see if any radios are checked by default and applies the 'selected' state to them


DFurnes:
:sailboat:

PR #3014, opened 2014-08-22
Removes campaign cache data collection for MB

deezone:
Removes logic to generate mbp requests. The campaign data was cached for use by the Message Broker system to generate digest email messages. This functionality has been superseded by https://www.dosomething.org/api/v1/content/:nid

See:
MB-36
https://jira.dosomething.org/browse/MB-36


jonuy:
`dosomething_mbp_get_campaign_cache_payload` is still being called on line 52. Are there other files where this code could potentially get called from?

deezone:
Lots of moving parts, thanks for taking a look. Related request types removed.

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
```diff
@@ -167,142 +161,3 @@ function dosomething_mbp_get_common_campaign_payload(&$payload, $params) {
...
- * Implements hook_node_insert().
- *
- * Respond to the creation of campaign nodes to produce Message Broker entries
- * in the directCacheExchange.
- *
- * @param object $node
- *   Submitted values for campaign
- */
-function dosomething_mbp_node_insert($node) {
-  if ($node->type == 'campaign' && variable_get('dosomething_mbp_send_campaign_api', FALSE) == TRUE) {
```
aaronschachter, line 221: Please kill these variables too, both from the admin form and the DB.  

You'll need to write a `dosomething_mbp_update_N` hook to delete them from the DB.

aaronschachter:
And indeed, be sure to do global searches for functions you're removing. Looks good so far.

aaronschachter:
There's also all this.... https://github.com/DoSomething/message_broker_producer/blob/master/message_broker_producer.admin.inc#L384

but I know we've been discussing refactoring that module completely.  But something to keep in mind when testing this locally.  Please make sure you've tested the entire User Registration / Signup and Reportback workflow to check any potential gotchas.

aaronschachter:
There's also no longer a need for the Admin page / form right? `dosomething_mbp_config_form`

aaronschachter:
Actually disregard - we do need it to Log MBP requests.  But that form shoudl be updated to no longer set the variable, as you've done in https://github.com/DeeZone/dosomething/commit/9bf0d3d26541c4ce9c22d7565d8459f6ade77451

aaronschachter:
@DeeZone has this been tested with the full flow? (user registration, signup, reportback, adding/editing a campaign)

deezone:
@aaronschachter 

> Please make sure you've tested the entire User Registration / Signup and Reportback workflow to check any potential gotchas.

and...

> has this been tested with the full flow? (user registration, signup, reportback, adding/editing a campaign)

Not to worry, running through testing before asking for a (thumbsup).

deezone:
@aaronschachter 
Full review complete on my local. This is ready for production with your final thumbs up.

aaronschachter:
Great!  Thanks for testing. :+1:

PR #3016, opened 2014-08-22
New campaign fields

aaronschachter:

Adds two new fields to the Campaign node type and adds them as properties to the Campaign class.

The new fields are only applicable to standard Campaigns, and are editable from a new field group called Pitch Page Additional Info.  This field group will only be visible when the Campaign type is set to `campaign`.

Closes https://jira.dosomething.org/browse/DS-264 to make the End Date checked by default for new campagins.


angaither:
more fields. oy.

Shouldn't they also be exported to [campaign info](https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.info) ?

aaronschachter:
Yup, they were: https://github.com/aaronschachter/dosomething/blob/campaign_fields/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.info

angaither:
huh. otay :+1:

PR #3018, opened 2014-08-22
Confirmation Page path change

aaronschachter:
Uses `campaigns/*/confirmation` instead of `node/*/confirmation` as a quick win for tracking Google Analytics on the Confirmation page, closing https://jira.dosomething.org/browse/DS-234

By using `campaigns`, we don't have to worry about Varnish, and also if a user navigates to `campaigns/[nid]` they'll be redirected to just regular `/campaigns`.


blisteringherb:
:+1:

PR #3020, opened 2014-08-25
Fix ds pull stage: force "yes".

sergiitk:
Fix international version. Previously fixed in #2961.

Drush only accepts `-y` before `[command]` argument.
After this fix `ds pull stage` will not annoy you with this question anymore:

```
You will destroy data from /var/www/vagrant/html/ and replace with data from dosomething@staging.beta.dosomething.org:/var/www/beta.dosomething.org/current/html//
Do you really want to continue? (y/n):
```

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -295,7 +295,7 @@ function symlinks {
 #===============================================================================
 function dev {
   echo 'Enabling modules'
-  drush en devel devel_generate -y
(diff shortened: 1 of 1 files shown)


sergiitk:
@blisteringherb Could you please review this?

aaronschachter:
:+1:

PR #3023, opened 2014-08-25
Remove unused SMS Game code

aaronschachter:
@jonuy Per our discussions around https://jira.dosomething.org/browse/DS-193, when you have time to review.

It'd be great to have a test Mobilecommons Campaign setup with mData's pointing to the staging site so we can test this on staging before we deploy it... just to sanity check that nothing breaks. cc: // @tongxiang


jonuy:
@aaronschachter The stuff you removed looks fine. I also added another commit that removes more. A test report back flow has also been hooked up with Mobile Commons for the Yeah, Science! campaign on staging.

aaronschachter:
thanks for taking out the extra stuff @jonuy !  here goes.

PR #3024, opened 2014-08-25
Shirt size variable

aaronschachter:
Closes https://jira.dosomething.org/browse/DS-275 by updating the values for the Shirt Size options.

Adds check for a `dosomething_shipment_shirt_size_options` to allow for overriding the options via drush.


angaither:
:+1:

PR #3025, opened 2014-08-25
Fixes bug in Signup Data Form Shirt collector

aaronschachter:
Fixes https://jira.dosomething.org/browse/DS-274

Instead of trying to overwrite `$form['item']` from a hidden element to a radio, it adds the hidden element as `$form['shipment_item']`.  Also fixes variable name bug.


angaither:
:+1:

PR #3026, opened 2014-08-25
Makes paraneue login.inc file translatable

aaronschachter:
@DFurnes Can you please review?

Closes https://jira.dosomething.org/browse/DS-276.

We should really do a clean sweep of all of these inc files, but this is for required strings in https://jira.dosomething.org/browse/DS-273


DFurnes:
Looks nice. :checkered_flag:

PR #3027, opened 2014-08-26
Don't recommend closed campaigns

aaronschachter:
@angaither Please review

Fixes https://jira.dosomething.org/browse/DS-278


angaither:
:+1:

PR #3032, opened 2014-09-02
DS-291 Canada user API integration

mshmsh5000:
### Notes:
- Some module functions still need block comments.
- Registration not implemented, but we do create users using the API when a local user logs in that doesn't exist remotely.
- Delete not implemented.
- Local profile sync might not cover all appropriate fields.

### DIFF
--- a/lib/modules/dosomething/dosomething_canada/dosomething_canada.info
+++ b/lib/modules/dosomething/dosomething_canada/dosomething_canada.info
@@ -2,7 +2,7 @@ name = DoSomething Canada
 description = Provides functionality specific to our Canadian friends.
 core = 7.x
 package = DoSomething
-version = 7.x-0.2
+version = 7.x-0.1
 project = dosomething_canada
 
 ; Core
--- a/lib/modules/dosomething/dosomething_canada/dosomething_canada.info
+++ b/lib/modules/dosomething/dosomething_canada/dosomething_canada.info
@@ -1,0 +1,19 @@
+name = DoSomething Canada
+description = Provides functionality specific to our Canadian friends.
+core = 7.x
+package = DoSomething
+version = 7.x-0.2
+project = dosomething_canada
+
+; Core
+dependencies[] = user
+dependencies[] = date
+
+; Contrib
+
+
+; Custom
+dependencies[] = dosomething_user
+
+; Includes
+files[] = includes/CanadaSSOClient.php
--- a/lib/modules/dosomething/dosomething_canada/dosomething_canada.install
+++ b/lib/modules/dosomething/dosomething_canada/dosomething_canada.install
@@ -1,0 +1,30 @@
+<?php
+/**
+ * @file
+ * Installation for dosomething_canada.module.
+ */
+
+/**
+ * Implements hook_install().
+ */
+function dosomething_canada_install() {
+
+  variable_set('dosomething_canada_sso_url',    getenv('DS_CANADA_SSO_URL'));
+  variable_set('dosomething_canada_sso_appid',    getenv('DS_CANADA_SSO_APPID'));
+  variable_set('dosomething_canada_sso_key', getenv('DS_CANADA_SSO_KEY'));
+
+}
+
+/**
+ * Implements hook_uninstall().
+ */
+function dosomething_canada_uninstall() {
+  $vars = array(
+    'dosomething_canada_sso_url',
+    'dosomething_canada_sso_appid',
+    'dosomething_canada_sso_key',
+  );
+  foreach ($vars as $var) {
+    variable_del($var);
+  }
+}
--- a/lib/modules/dosomething/dosomething_canada/dosomething_canada.install
+++ b/lib/modules/dosomething/dosomething_canada/dosomething_canada.install
@@ -9,8 +9,8 @@
  */
 function dosomething_canada_install() {
 
-  variable_set('dosomething_canada_sso_url',    getenv('DS_CANADA_SSO_URL'));
-  variable_set('dosomething_canada_sso_appid',    getenv('DS_CANADA_SSO_APPID'));
+  variable_set('dosomething_canada_sso_url', getenv('DS_CANADA_SSO_URL'));
+  variable_set('dosomething_canada_sso_appid', getenv('DS_CANADA_SSO_APPID'));
   variable_set('dosomething_canada_sso_key', getenv('DS_CANADA_SSO_KEY'));
 
 }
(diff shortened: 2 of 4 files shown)


mshmsh5000:
### More notes:
- Needs more logging, especially for registration/login failures to aid in debugging and other recovery.
- (Related) Clean up thrown/caught exceptions, and log appropriately.
- Document in comments all user fields that will need to be synced either way.

lib/modules/dosomething/dosomething_canada/dosomething_canada.module
```diff
@@ -0,0 +1,263 @@
...
+    }
+  }
+  catch (Exception $e) {
+    form_set_error('name', t("Sorry, we don't recognize that email and password."));
+    return FALSE;
+  }
+
+  if (!$remote_account) {
+    // Todo: link to forgot password page.
+    form_set_error('name', t('Sorry, unrecognized username or password.'));
```
aaronschachter: Minor, but we don't use username field anywhere.

lib/modules/dosomething/dosomething_canada/dosomething_canada.info
```diff
@@ -0,0 +1,19 @@
+name = DoSomething Canada
+description = Provides functionality specific to our Canadian friends.
+core = 7.x
+package = DoSomething
+version = 7.x-0.2
```
aaronschachter: Random, but why 0.2?

lib/modules/dosomething/dosomething_canada/dosomething_canada.install
```diff
@@ -0,0 +1,30 @@
...
+ * @file
+ * Installation for dosomething_canada.module.
+ */
+
+/**
+ * Implements hook_install().
+ */
+function dosomething_canada_install() {
+
+  variable_set('dosomething_canada_sso_url',    getenv('DS_CANADA_SSO_URL'));
```
aaronschachter: picky picky, but inconsistent spacing in this `hook_install`

lib/modules/dosomething/dosomething_canada/dosomething_canada.module
```diff
@@ -0,0 +1,263 @@
...
+  else {
+    $account = dosomething_canada_process_new_user($name, $pass);
+    form_set_value($form['name'], $account->name, $form_state);
+  }
+}
+
+/**
+ * Validates user_register.
+ */
+function dosomething_canada_register_validate($form, &$form_state) {
```
aaronschachter: I see this has been copied from `dosomething_uk_register_validate`, but, let's kill this to keep this module lean and clean, as the entire function is commented out.

lib/modules/dosomething/dosomething_canada/includes/CanadaSSOClient.php
```diff
@@ -0,0 +1,212 @@
...
+    if ('POST' == $method)
+    {
+      $curlOpts[CURLOPT_POST] = true;
+      $curlOpts[CURLOPT_POSTFIELDS] = $query;
+    }
+
+    $ch = curl_init($url);
+    curl_setopt_array($ch, $curlOpts);
+
+    $response = curl_exec($ch);
```
aaronschachter, line 167: If this is all working and good, then no reason to change it.  I just wonder if this code could be simplified by using `drupal_http_request` instead of `curl_exec` .. cc @sergii-tkachenko

aaronschachter:
Overall seems fine.  Just little picky things but nothing blocking merge... we'd just want to be sure to go back in and clean up the minor things (and would we ever?)

lib/modules/dosomething/dosomething_canada/dosomething_canada.module
mshmsh5000: I'll standardize these.

lib/modules/dosomething/dosomething_canada/dosomething_canada.info
mshmsh5000: Random indeed! Thanks -- changing to 0.1.

lib/modules/dosomething/dosomething_canada/dosomething_canada.install
mshmsh5000: Weird spacing from my IDE -- thanks, will fix.

lib/modules/dosomething/dosomething_canada/dosomething_canada.module
mshmsh5000: +1

lib/modules/dosomething/dosomething_canada/includes/CanadaSSOClient.php
mshmsh5000, line 167: Not for 0.1, but worth simplifying if we can get the same range of flexibility w/r/t curl options.

sergiitk, line 167: Agree with both of you!
It should be fine by now, it's just easier to debug with `curl` sometimes.
In future it's worth trying refactor this to use `drupal_http_request()`.

lib/modules/dosomething/dosomething_canada/dosomething_canada.module
```diff
@@ -0,0 +1,275 @@
...
+  array_unshift($form['#validate'], 'dosomething_canada_login_validate');
+}
+
+/**
+ * Implements hook_form_FORM_ID_alter().
+ *
+ * Form user_register.
+ */
+function dosomething_canada_form_user_register_form_alter(&$form, $form_state) {
+  array_unshift($form['#submit'], 'dosomething_canada_register_validate');
```
sergiitk, line 58: Hm. `dosomething_canada_register_validate` is not declared within this changeset. Is this right?

aaronschachter, line 58: Looks like @mshmsh5000 took out that function since it didn't do anything (it was all commented out).

However, @sergii-tkachenko I know you added code into the `dosomething_uk_login_validate` , which was added after this PR was merged.  @mshmsh5000 @sergii-tkachenko  -- Do we need a `dosomething_canada_login_validate` function?

sergiitk, line 58: Yeah.
Without it there's a chance to get unexpected PHP Fatal on production, let's say, during the cache clean.

sergiitk, line 58: I mean there must be at least a declaration of the function. It's better to leave it blank than undeclared.

mshmsh5000:
Or perhaps I'll remove the hook_user_register_form_alter().

sergiitk:
@mshmsh5000 Sure, great!

sergiitk:
@mshmsh5000 Do you know if functionality similar to #3040 is also planned for Canada?

mshmsh5000:
@sergii-tkachenko Yes, after this initial release & launch.

PR #3033, opened 2014-09-02
Drush helpers for CasperJS

DFurnes:
## Changes
- You can now call Drush commands from within CasperJS using the `casper.drush()` helper method! Just pass an array of parameters and it'll shuttle them off directly to Dries.
- Splits out PhantomCSS stuff into its own file so that `common.js` only has "helper" functions.
- Mutes helper function unless using `--verbose` flag.
## How do I test this?

Great question! Pull down this branch, provision if you haven't done that in a while, and then issue the magic incantation `ds test` into your Vagrant console. It will look like this, but about 60x slower:

screen recording 2014-09-02 at 04 09 pm


angaither:
:+1:

PR #3034, opened 2014-09-03
Users endpoint - Check for existing email

aaronschachter:
Return error when creating a User via the `api/v1/users` endpoint if the email exists already.

### DIFF
--- a/lib/modules/dosomething/dosomething_user/resources/dosomething_user.resource.inc
+++ b/lib/modules/dosomething/dosomething_user/resources/dosomething_user.resource.inc
@@ -38,8 +38,8 @@ function _dosomething_user_resource_create($account) {
   }
   $email = $account['email'];
   // Check if account exists for email.
-  if (user_load_by_mail($email)) {
-    return services_error("Email $email is already registered.");
+  if ($user = user_load_by_mail($email)) {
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_user/resources/dosomething_user.resource.inc
```diff
@@ -36,10 +36,15 @@ function _dosomething_user_resource_create($account) {
   if (!isset($account['email'])) {
     return services_error("Email is required.");
   }
+  $email = $account['email'];
+  // Check if account exists for email.
+  if (user_load_by_mail($email)) {
+    return services_error("Email $email is already registered.");
```
deezone: Would it be possible to return the uid of the existing account in the error message?

aaronschachter: Sure, but what is the reasoning? @DeeZone

deezone: A part of the import process is logging existing accounts so we can report to niche.com and not pay for the user. Having the Drupal nid would allow for generating a link to the Drupal user account in the report making the report more valuable and easy to confirm / review the existing user.

PR #3035, opened 2014-09-03
API updates (Signup and Reportback)

aaronschachter:
Provides `campaigns/[nid]/signup` and `campaigns/[nid]/reportback` endpoints.  This won't be enabled on production, but has been added for SMS flow, when we're ready.

Removes the `reportback/save` resource from `dosomething_reportback` module

### DIFF
--- a/lib/modules/dosomething/dosomething_api/dosomething_api.module
+++ b/lib/modules/dosomething/dosomething_api/dosomething_api.module
@@ -39,6 +39,7 @@ function dosomething_api_services_resources() {
         'file' => array('type' => 'inc', 'module' => 'dosomething_api', 'name' => 'resources/campaign_resource'),
         'callback' => '_campaign_resource_index',
         'access callback' => '_campaign_resource_access',
+        'access arguments' => array('index'),
         'args' => array(
           array(
             'name' => 'parameters',
--- a/lib/modules/dosomething/dosomething_api/dosomething_api.module
+++ b/lib/modules/dosomething/dosomething_api/dosomething_api.module
@@ -57,8 +57,8 @@ function dosomething_api_services_resources() {
         'file' => array('type' => 'inc', 'module' => 'dosomething_api', 'name' => 'resources/campaign_resource'),
         'callback' => '_campaign_resource_signup',
         'access callback' => '_campaign_resource_access',
-        //'access arguments' => array('signup'),
-        //'access arguments append' => TRUE,
+        'access arguments' => array('signup'),
+        'access arguments append' => TRUE,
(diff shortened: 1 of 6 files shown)


lib/modules/dosomething/dosomething_api/dosomething_api.module
```diff
@@ -51,6 +52,55 @@ function dosomething_api_services_resources() {
         ),
       ),
     ),
+    'targeted_actions' => array(
+      'signup' => array(
+        'help' => 'Signup a user for a campaign. POST to campaigns/123/signup',
```
angaither, line 57: `t()` ?

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -522,6 +522,19 @@ function dosomething_campaign_is_closed($node) {
...
+ * For given campaign $node, determines if it is active.
+ *
+ * @param obj $node
+ *   A loaded node.
+ *
+ * @return bool
+ */
+function dosomething_campaign_is_active($node) {
+  // Node must be published, and campaign status != closed.
+  return $node->status == 1 && !dosomething_campaign_is_closed($node);
```
angaither, line 534: :+1:

angaither:
Code looks okay, I'm not familiar with this stuff, but how is it disabled in prod?

aaronschachter:
It's all disabled in production because the Services endpoints are not enabled.

angaither:
Does the lobby display pull from QA?

aaronschachter:
I'm 95% sure it uses production. 

https://beta.dosomething.org/api/v1/campaigns.json?parameters[is_staff_pick]=1
https://beta.dosomething.org/api/v1/content/362.json

aaronschachter:
The "campaigns" index and "node" retrieve endpoints are enabled on prod, but we won't be enabling these new endpoints introduced in this PR: which are "campaigns/[nid]/signup" and "campaigns/[nid]/reportback."  What's enabled/disabled is stored in https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_api/dosomething_api.services.inc

lib/modules/dosomething/dosomething_api/dosomething_api.module
aaronschachter, line 57: Questionable if our affiliates will ever turn on API access, but you never know.  I was coping examples from the Services module which don't translate.  But I can add this in in the next round, as I'm looking to conslidate all services code into this one DoSomething API module.

PR #3036, opened 2014-09-03
Add some better logging for Drush actions.

DFurnes:
## Changes
- Unmutes Drush helper functions because turns out that stuff is actually useful to see in logs.
## How do I test?

Meh, just give it a once-over. This isn't rocket science.


aaronschachter:
:+1:

PR #3037, opened 2014-09-03
DS-303 Update validation to accept any country code & spacing.

DFurnes:
## Changes
- Client-side and server-side phone number validation now (optionally) accepts any three-digit country code.
- Validation no longer checks for US-centric `### ### ####` grouping of digits, to allow for alternative spacing like `#### ### ###` for our brethren across the pond.

This addresses [DS-303](https://jira.dosomething.org/browse/DS-303), although we probably want to make a second pass at this in the future to validate specific country codes for each international region.
## How do I test this?

Try to break each validation function. Client-side validation function passes test suite, and can be manually tested by entering text into the phone number field in browser. The server-side validation function should be able to be tested with Drush.

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -175,7 +175,7 @@ function dosomething_user_get_staff_only_types() {
   *   True if valid, else false.
   */
 function dosomething_user_valid_mobile($number) {
-  preg_match('#^(?:\+?1([\-\s\.]{1})?)?\(?([0-9]{3})\)?(?:[\-\s\.]{1})?([0-9]{3})(?:[\-\s\.]{1})?([0-9]{4})#', $number, $valid);
+  preg_match('#^(?:\+?([0-9]{1,3})([\-\s\.]{1})?)?\(?([0-9]{3})\)?(?:[\-\s\.]{1})?([0-9]{3})(?:[\-\s\.]{1})?([0-9]{4})#', preg_replace('#\-\s\.#', '', $number), $valid);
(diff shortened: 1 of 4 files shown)


aaronschachter:
Can you please update the unit test too ?  That should help. https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_user/dosomething_user.test#L24

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -175,7 +175,7 @@ function dosomething_user_get_staff_only_types() {
   *   True if valid, else false.
   */
 function dosomething_user_valid_mobile($number) {
-  preg_match('#^(?:\+?1([\-\s\.]{1})?)?\(?([0-9]{3})\)?(?:[\-\s\.]{1})?([0-9]{3})(?:[\-\s\.]{1})?([0-9]{4})#', $number, $valid);
+  preg_match('#^(?:\+?([0-9]{1,3})([\-\s\.]{1})?)?\(?([0-9]{3})\)?(?:[\-\s\.]{1})?([0-9]{3})(?:[\-\s\.]{1})?([0-9]{4})#', preg_replace('#\-\s\.#', '', $number), $valid);
```
aaronschachter: Was this working for you?  When I try

```
$num = '44 07708 200 305';
print dosomething_user_valid_mobile($num);
```

in `http://dev.dosomething.org:8888/devel/php`, it doesn't return TRUE... 

I'm using the # found in the validation.js

DFurnes: Yeah, oopsie. See @83e45dd.

DFurnes:
Good call. Test cases updated and passing. :100:

aaronschachter:
:+1: Hooray tests

PR #3038, opened 2014-09-03
Remove Message Broker modules from Dosomething.info

aaronschachter:
Removes a few US specific modules from the .info file, to prevent the `enable_modules.php` script turning them on when they may need to be left off.


blisteringherb:
:+1:

PR #3021, opened 2014-08-25
Fix @self alias.

sergiitk:
After refactoring #2809 `drush` site alias `@self` got broken.
For unknown reason it points to Botswana:

```
vagrant@dev:/var/www/vagrant/html/sites/default$ drush @self status
 Drupal version                  :  7.31
 Site URI                        :  http://dev.botswana.dosomething.org
 Database driver                 :  mysql
 Database username               :  root
 Database name                   :  dosomething
 Database                        :  Connected
 Drupal bootstrap                :  Successful
 Drupal user                     :  Anonymous
 Default theme                   :  paraneue_dosomething
 Administration theme            :  seven
 PHP executable                  :  /usr/bin/php
 PHP configuration               :  /etc/php5/cli/php.ini
 PHP OS                          :  Linux
 Drush version                   :  6.3.0
 Drush configuration             :
 Drush alias files               :  /home/vagrant/.drush/ds.aliases.drushrc.php
 Drupal root                     :  /var/www/vagrant/html
 Site path                       :  sites/default
 File directory path             :  sites/default/files
 Temporary file directory path   :  /tmp
```

It causes multiple issues with `bin/ds` script. The most important — it fails `bin/ds pull stage`:

```
vagrant@dev:/var/www/vagrant$ bin/ds pull stage
Pulling down db from staging
WARNING:  Using temporary files to store and transfer sql-dump.  It is recommended that you specify --source-dump and --target-dump options on the command line, or set '%dump' or '%dump-dir' in the path-aliases section of your site alias records. This facilitates fast file transfer via rsync.

You will destroy data in dosomething_botswana and replace with data from staging.beta.dosomething.org/dosomething.

You might want to make a backup first, using the sql-dump command.

Do you really want to continue? (y/n): y
ERROR 1049 (42000): Unknown database 'dosomething_botswana'
Enabling Stage File Proxy...
stage_file_proxy is already enabled.                                                                                                                                                             [ok]
There were no extensions that could be enabled.                                                                                                                                                  [ok]
Setting Stage File Proxy to staging URL...
stage_file_proxy_origin was set to "http://staging.beta.dosomething.org".                                                                                                                        [success]
Pulling down files from stage. Any affiliate sites will be destroyed
You will destroy data from /var/www/vagrant/html/ and replace with data from dosomething@staging.beta.dosomething.org:/var/www/beta.dosomething.org/current/html//
Do you really want to continue? (y/n): y
```

[...]

PR #3041, opened 2014-09-03
Clear Asset Path theme setting after pulling from stage.

DFurnes:
## Changes
- Clears Asset Path theme setting after pulling database from stage (so that local development will not be trying to pull assets from the CDN).
## How do I test?

Pull from staging. Check theme settings. Profit. (Check #3021 for Sergii's fix if having issues with database sync.)

For review: @blisteringherb

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -437,10 +437,16 @@ function pull_db {
       echo "Pulling down db from staging"
       cd $WEB_PATH
       drush -y sql-sync @ds.staging @self
+      
       echo "Enabling Stage File Proxy..."
       drush -y en stage_file_proxy
+      
       echo "Setting Stage File Proxy to staging URL..."
       drush vset stage_file_proxy_origin "http://staging.beta.dosomething.org"
+
+      echo "Unsetting Asset Path theme setting..."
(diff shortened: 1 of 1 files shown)


bin/ds
```diff
@@ -437,10 +437,16 @@ function pull_db {
...
       drush -y sql-sync @ds.staging @self
+      
       echo "Enabling Stage File Proxy..."
       drush -y en stage_file_proxy
+      
       echo "Setting Stage File Proxy to staging URL..."
       drush vset stage_file_proxy_origin "http://staging.beta.dosomething.org"
+
+      echo "Unsetting Asset Path theme setting..."
+      drush php-eval '$theme_settings = variable_get("theme_paraneue_dosomething_settings", array()); $theme_settings["asset_path"] = ""; variable_set("theme_paraneue_dosomething_settings", $theme_settings);'
```
aaronschachter, line 448: Seems ok.  If we set up the CDN module.. we shouldn't need this setting at all anymore, right? https://jira.dosomething.org/browse/DS-109

DFurnes, line 448: Think so.

aaronschachter, line 448: ship :+1:

PR #3043, opened 2014-09-03
Ds 327 canada user register hook

mshmsh5000:
We'll obviously need to declare this when we implement registration calls to the API.


sergiitk:
:+1:

PR #3044, opened 2014-09-04
Remove Grunt test task (now handled by ds utility).

DFurnes:
## Changes
- Fixes issue where `test` Grunt task was pointing to old test directory.
- Removes some extra node modules we no longer need to specify.
## How do I test?

Pull down this branch, go to the `paraneue_dosomething` directory and run `grunt`. Things should not explode.


tjrosario:
Works like a charm! :+1:

PR #3046, opened 2014-09-04
showing h1 at all breakpoints and hiding h2 at mobile breakpoint on home...

tjrosario:
# Homepage (Affiliates only)
- Reveals `<h1>` tag on all breakpoints
- Hides `<h2>` at mobile breakpoint

Resolves [DS-320](https://jira.dosomething.org/browse/DS-320)


weerd:
:+1:

PR #3047, opened 2014-09-05
Add Privacy Policy link to Registration legal copy

aaronschachter:
@sergii-tkachenko  Can you review? 

Closes https://jira.dosomething.org/browse/DS-311
- Adds variables to set the Privacy Policy and Terms of Service nodes, to avoid hardcoded node nid's.
- Makes a token available to link to the Privacy Policy in the Registration Form legal text, to allow for UK request to add it to the legal copy via translation.


sergiitk:
:+1:

aaronschachter:
Note to self to also update the `footer.tpl.php` with these variables: https://github.com/DoSomething/dosomething/blob/v0.3.1/lib/themes/dosomething/paraneue_dosomething/templates/system/partials/footer.tpl.php#L51

PR #3048, opened 2014-09-05
DRY for Terms/Privacy variable links

aaronschachter:
Adds functions `dosomething_user_get_link_terms_service` and `dosomething_user_get_link_privacy_policy`, updates footer.tpl.php per https://github.com/DoSomething/dosomething/pull/3047#issuecomment-54628600


angaither:
:+1: value added. ship it. :shipit:

PR #3049, opened 2014-09-05
Neue 4.5

DFurnes:
## Changes
- Pull in Neue 4.5.x via Bower.
- Remove duplicated Media Selector code now that it is in the pattern library.

screen shot 2014-09-05 at 11 01 01 am


weerd:
:+1:

PR #3050, opened 2014-09-05
Login Form copy variable

aaronschachter:
Adds a variable to set login form copy.

Unblocks https://jira.dosomething.org/browse/DS-354

screen shot 2014-09-05 at 11 09 58 am


DFurnes:
:hand: :wave: :hand: :wave:

PR #3051, opened 2014-09-05
User Login: Display Forgot Password variable

aaronschachter:
Closes UK request to hide the "Request Password" link: https://jira.dosomething.org/browse/DS-322

If the variable is unchecked, the "Forgot password" link will not be rendered.  Should be checked on almost all sites, hence the additional `variable_set` in `dosomething_user_update_7001`.


DFurnes:
Looks good! :sailboat:

PR #3052, opened 2014-09-05
UK Login Form copy

aaronschachter:

Programmatically sets variable this so it's good to go for next deployment.  Uses version # 7003 because of the module info file.

### DIFF
--- a/lib/modules/dosomething/dosomething_uk/dosomething_uk.install
+++ b/lib/modules/dosomething/dosomething_uk/dosomething_uk.install
@@ -17,6 +17,14 @@ function dosomething_uk_install() {
 }
 
 /**
+ * Sets Login Form copy variable.
+ */
+function dosomething_uk_update_7003() {
+  $copy = "(use your vInspired login if you have one)";
+  variable_set('dosomething_user_login_form_copy', $copy);
+}
+
+/**
  * Implements hook_uninstall().
  */
 function dosomething_uk_uninstall() {


lib/modules/dosomething/dosomething_uk/dosomething_uk.install
```diff
@@ -17,6 +17,14 @@ function dosomething_uk_install() {
 }
 
 /**
+ * Sets Login Form copy variable.
+ */
+function dosomething_uk_update_7003() {
```
sergiitk, line 22: Should you also bump the version then?

aaronschachter, line 22: This matches the current version, no?

```
version = 7.x-0.3
```

sergiitk:
:+1:

lib/modules/dosomething/dosomething_uk/dosomething_uk.install
aaronschachter, line 22: When I tried `dosomething_uk_update_7001` it would not run, I believe this is because of the module version in the .info per last comment.

sergiitk, line 22: Just for the matters of history:

> [7:46 PM] Sergii Tkachenko: but drush updb processes this update as new, right?
> [7:46 PM] Aaron Schachter: yea i ran it
>       and it worked
>       wasn't working when i used 7001
> [7:46 PM] Sergii Tkachenko: yeah, great

PR #3054, opened 2014-09-05
User Opt In Fields

aaronschachter:
@sergii-tkachenko can you please review?

Adds two fields, for opt-ing into Email and Mobile.

Adds more configuration variables, which determine whether or not to display the fields on the Registration Form via `hook_form_alter`.


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -570,6 +573,34 @@ function _dosomething_user_register_helper_text(&$form) {
...
+function _dosomething_user_register_display_fields(&$form) {
+  $var_prefix = 'dosomething_user_register_form_display';
+  // List of fields which have display variables to check.
+  $display_vars = array(
+    'opt_in_email',
+    'opt_in_sms',
+  );
+  // For each display variable:
+  foreach ($display_vars as $var) {
+    $var_name = $var_prefix . '_' . $var;
```
sergiitk: I think it'd more implicit to just do $var_name = 'dosomething_user_register_form_display_' . $var;`

lib/modules/dosomething/dosomething_user/dosomething_user.admin.inc
```diff
@@ -52,6 +52,26 @@ function dosomething_user_admin_config_form($form, &$form_state) {
...
+    '#type' => 'fieldset',
+    '#title' => t('Registration'),
+    '#collapsible' => TRUE,
+    '#collapsed' => TRUE,
+  );
+  $var_name = 'dosomething_user_register_form_display_opt_in_email';
+  $form['reg'][$var_name] = array(
+    '#type' => 'checkbox',
+    '#title' => t('Display "Opt-In Email" field'),
+    '#default_value' => variable_get($var_name, FALSE),
```
sergiitk, line 76: Same here, `variable_get('dosomething_user_register_form_display_opt_in_email', FALSE),` is just easier to read.

lib/modules/dosomething/dosomething_user/dosomething_user.admin.inc
```diff
@@ -52,6 +52,26 @@ function dosomething_user_admin_config_form($form, &$form_state) {
...
+  );
+  $var_name = 'dosomething_user_register_form_display_opt_in_sms';
+  $form['reg'][$var_name] = array(
+    '#type' => 'checkbox',
+    '#title' => t('Display "Opt-In SMS" field'),
+    '#default_value' => variable_get($var_name, FALSE),
```
sergiitk, line 84: Same here.

lib/modules/dosomething/dosomething_user/dosomething_user.admin.inc
aaronschachter, line 76: I see what you mean... the reason why I've been using a `$var_name` in these admin forms (it's in a few places now) is so it's not duplicating the `dosomething_user_register_form_display_opt_in_email` twice .. in the `$form[$var_name]` and also the `'#default_value' => variable_get($var_name, FALSE),`

sergiitk, line 76: Yes, you're right here. But on the other side, it's not harmful to put the name of the variable twice.
Maybe it's just me, but I found myself struggling to get visually what's variable are you setting in this blocks from the first sight.
So it's a risk to make a mistake and forget to change the name in both places vs. code readability.
I prefer to stick with Fowler's 'Refactoring' book when I have this doubts: http://refactoring.com/catalog/inlineTemp.html

aaronschachter, line 76: Yeah, I just always end up forgetting to change the `default_value` when cutting and pasting.  I think adding some line spacing would help here.  I would prefer to revisit this later and make all the changes at once if we decide, but yeah, this is kinda minor.  I'll add some spacing and ping you again for review, maybe that will help...

sergiitk, line 76: Yeah, totally agree!

sergiitk:
Add the newlines and fix the one from first comment, and :+1:!

PR #3056, opened 2014-09-05
Last Name Registration Form  (UK)

aaronschachter:
@sergii-tkachenko Can you please review when you have a chance?

Refs https://jira.dosomething.org/browse/DS-321
Closes https://jira.dosomething.org/browse/DS-361
- Alters the `field_last_name` to be displayed on the Registration Form.
- Adds a `dosomething_user_register_form_display_last_name` which controls whether or not to display it.
- If a last name field value exists, include it in the UK SSO signup request

### DIFF
--- a/lib/modules/dosomething/dosomething_uk/dosomething_uk.module
+++ b/lib/modules/dosomething/dosomething_uk/dosomething_uk.module
@@ -102,14 +102,18 @@ function dosomething_uk_register_validate($form, &$form_state) {
 
   // User data.
   $dob = new DateObject($values['field_birthdate'][LANGUAGE_NONE][0]['value']);
+  // Initialize Last Name to default value.
+  $last_name = 'Harkness';
+  // Check if Last Name field form value exists.
+  if (isset($values['field_last_name'][LANGUAGE_NONE][0]['value'])) {
+    $last_name = $values['field_last_name'][LANGUAGE_NONE][0]['value'];
+  }
   $user_data = array(
     'first_name' => $values['field_first_name'][LANGUAGE_NONE][0]['value'],
     'password'   => $values['pass'],
     'email'      => $values['mail'],
     'dob'        => $dob,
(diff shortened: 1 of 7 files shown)


sergiitk:
Story time.

When I was signed up for the first time on vinspired.org, it won't let me in without my zip code.

I thought: alright, I'll give you one — and filled in `CF10 5AN`.
This zip is the exact location of the The Hub, fictional place from British syfy show, Torchwood, spin-off the 'Doctor Who'. It seemed laughable to my at the moment, and very British.

This led me to use this zip everywhere. It's in all my fake users on vinspired stage, in all tests, dev functions, etc. So when it was decided to hardcode zip, I already knew which one will go there.

Yeah, Harkness, Jack. Torchwood Three's leader. You're looking at this guy:
image

lib/modules/dosomething/dosomething_uk/dosomething_uk.module
```diff
@@ -102,14 +102,18 @@ function dosomething_uk_register_validate($form, &$form_state) {
 
   // User data.
   $dob = new DateObject($values['field_birthdate'][LANGUAGE_NONE][0]['value']);
+  // Initialize Last Name to default value.
+  $last_name = 'Harkness';
```
sergiitk: So that's why Captain Harkness shouldn't be here. You should just return from this function when the last name is not set and Drupal will take care of further validations. Or leave this field blank and then vInspired API will take care of it.

sergiitk:
Everything else looks great! Good work, @aaronschachter!

lib/modules/dosomething/dosomething_uk/dosomething_uk.module
aaronschachter: good point (and great story! I am now inspired to provide more narrative PR comments)

you're right -- the form element has been made required so we don't need the check.  will remove this then merge

DFurnes:
good work

PR #3057, opened 2014-09-09
Registration Form: Display Postal Code config

aaronschachter:

Displays `field_address` on the Registration Form, but only displays it if the "Display Postal Code" variable is checked with the DS User Admin form.

Closes https://jira.dosomething.org/browse/DS-355 (last configuration variable is now in place)

Closes https://jira.dosomething.org/browse/DS-362 -- Displays Postal Code on register form if the variable is set

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.features.field_instance.inc
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.features.field_instance.inc
@@ -45,7 +45,7 @@ function dosomething_user_field_default_field_instances() {
     'label' => 'Address',
     'required' => 0,
     'settings' => array(
-      'user_register_form' => 0,
+      'user_register_form' => 1,
     ),
     'widget' => array(
       'active' => 1,
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -593,8 +593,10 @@ function _dosomething_user_register_display_fields(&$form) {
     $var_name = 'dosomething_user_register_form_display_' . $var;
     $display = variable_get($var_name, FALSE);
     $field_name = 'field_' . $var;
+    // For the postal code variable:
     if ($var == 'postal_code') {
-      // The field_address Address field contains the Postal Code element.
+      // Postal code is a form element within the field_address field.
+      // We store 'field_address' as the field_name value to unset.
       $field_name = 'field_address';
     }
     // If variable is set to NOT display:
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -608,8 +608,27 @@ function _dosomething_user_register_display_fields(&$form) {
       if ($var == 'last_name') {
         $form[$field_name]['#required'] = TRUE;
       }
+      elseif ($var == 'postal_code') {
+        $form['#after_build'][] = 'dosomething_user_address_field_postal_code_only';
+      }
     }
   }
+}
+
+/**
+ * After build function to unset all address field elements except postal code.
+ */
+function dosomething_user_address_field_postal_code_only($form, &$form_state) {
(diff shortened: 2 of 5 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.features.field_instance.inc
```diff
@@ -45,7 +45,7 @@ function dosomething_user_field_default_field_instances() {
     'label' => 'Address',
     'required' => 0,
     'settings' => array(
-      'user_register_form' => 0,
+      'user_register_form' => 1,
```
sergiitk, line 48: Could you clarify what is this for?

aaronschachter, line 48: Displays `field_address` on the user registration form.

screen shot 2014-09-09 at 1 45 31 pm

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -603,11 +608,30 @@ function _dosomething_user_register_display_fields(&$form) {
...
+      }
     }
   }
 }
 
 /**
+ * After build function to unset all address field elements except postal code.
+ */
+function dosomething_user_address_field_postal_code_only($form, &$form_state) {
+  $element = &$form['field_address'][LANGUAGE_NONE][0]['address'];
```
sergiitk, line 624: Should you check somewhere whether `dosomething_user_register_form_display_postal_code` is set?

lib/modules/dosomething/dosomething_user/dosomething_user.features.field_instance.inc
sergiitk, line 48: Oh, I see, thank you. Sometimes it's difficult to understand what's what within features exports.

lib/modules/dosomething/dosomething_user/dosomething_user.module
aaronschachter, line 624: The check takes place a few lines up in `_dosomething_user_register_display_fields`.  This function might come in handy in other places if we ever wanted a Zip/Postal code only widget, regardless of whether or not the Postal Code is collected on the Registration form, so I think it makes sense to leave that check outside of this function. 

 https://github.com/DoSomething/dosomething/pull/3057/files#diff-ad30a5d569ca2ae6d2e04223aa501104R611

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -586,15 +586,20 @@ function _dosomething_user_register_display_fields(&$form) {
...
     'opt_in_sms',
+    'postal_code',
   );
   // For each display variable:
   foreach ($display_vars as $var) {
     $var_name = 'dosomething_user_register_form_display_' . $var;
     $display = variable_get($var_name, FALSE);
+    $field_name = 'field_' . $var;
+    if ($var == 'postal_code') {
+      // The field_address Address field contains the Postal Code element.
```
sergiitk: Not the obvious why exactly you do what you do here. I think better explanation answering 'Why?' we need get Postal code separately would be very helpful here.

aaronschachter: Ok, I can add more comments... just waiting to see if anything else to change?

lib/modules/dosomething/dosomething_user/dosomething_user.module
sergiitk, line 624: Just to ensure I understand it right: `_dosomething_user_register_display_fields` is a helper function to user registration/profile form alter that describes exceptions from user fields normal display. Sounds right?

aaronschachter, line 624: Yup.  It's the function that checks the "Display Last name" "Display Opt-in Email" variables.  If these variables are not set, it will unset the relevant fields from the Registration Form.

sergiitk:
Great! That's one hell of an alter. Good job @aaronschachter! :+1:

PR #3058, opened 2014-09-09
Style cleanup

DFurnes:
## Changes
- Style cleanup & removing extra styles which were moved to Neue in DoSomething/neue#340.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
@@ -295,18 +295,6 @@
     }
   }
 }
-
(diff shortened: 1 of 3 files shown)


lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
```diff
@@ -295,30 +295,3 @@
     }
   }
 }
-
-[data-modal] {
-  // @TODO: update in Neue: _forms.scss
-  label {
-    display: block;
-    float: none;
```
weerd, line 303: This targets just `label`s in `[data-modal]`...

weerd:
:+1:

PR #3063, opened 2014-09-10
adds kik meta data for campaign pages

tjrosario:
Adds Kik meta data on all campaign pages, specifically the custom 'kik-icon' meta tag with the `<head>`:

`<link rel="kik-icon" href="campaign_hero_image.jpg">`

Resolves [DS-360](https://jira.dosomething.org/browse/DS-360)

### DIFF
--- a/lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
+++ b/lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
@@ -74,12 +74,20 @@ function dosomething_metatag_add_metatag_image_src($node, $field) {
     }
   }
   if ($image) {
-    // Add <link href="[image]" rel="kik-icon"> into the HEAD.
+    // Add social sharing thumbnail meta tag, 
+    // <link rel="image_src" href="[image]"> into the HEAD.
     $attributes = array(
+      'href' => $image, 
+      'rel' => 'image_src',
+    );
+    drupal_add_html_head_link($attributes, TRUE);
+    // Add the kik icon meta tag,
+    // <link  rel="kik-icon" href="[image]"> into the HEAD
+    $attributes_kik = array(
       'href' => $image, 
       'rel' => 'kik-icon',
     );
-    drupal_add_html_head_link($attributes, TRUE);
+    drupal_add_html_head_link($attributes_kik, TRUE);
     // Add the Facebook Open Graph tag into the HEAD.
     $element = array(
       '#tag' => 'meta',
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
```diff
@@ -74,10 +74,10 @@ function dosomething_metatag_add_metatag_image_src($node, $field) {
     }
   }
   if ($image) {
-    // Add <link img_src="[image]"> into the HEAD.
+    // Add <link href="[image]" rel="kik-icon"> into the HEAD.
     $attributes = array(
       'href' => $image, 
-      'rel' => 'image_src',
+      'rel' => 'kik-icon',
```
aaronschachter: Do we actually want to replace `<link img_src="[image]">`?? Don't we want to just an additional `<link href="[image]" rel="kik-icon">` ?

tjrosario: Well, the `image_src` tag is used for the Social Sharing.  Do we have need for this right now?  If so, we can add the `kik-icon` as an additional.

aaronschachter: Yeah I think it's good to have -- I would advise we add `kik-icon` as additional.  It's basically just additional `drupal_add_html_head_link` call

tjrosario: @aaronschachter ok both are now added

aaronschachter:
:+1:

lib/themes/dosomething/paraneue_dosomething/templates/system/html.tpl.php
```diff
@@ -19,6 +19,7 @@
 
   <title><?php print $head_title; ?></title>
   <script type="text/javascript" src="<?php print NEUE_ASSET_PATH; ?>/js/vendor/modernizr.js"></script>
+  <script src="http://cdn.kik.com/kik/1.0.9/kik.js"></script>
```
DFurnes, line 22: @tjrosario Does this need to be in the `head`? Adding another blocking HTTP request is kinda crummy.

DFurnes, line 22: It would also be nice to add some documentation on what `kik.js` does somewhere & why we need it. Even if just in the issue or commit message.

barryclark, line 22: I'm hoooping we don't need to include the js at all, @tjrosario -- think we might be able to get away with being "Kik Optimized" without it?

DFurnes, line 22: Yeah, their documentation is super vague. Unless we have deeper integrations on the roadmap, maybe we try leaving this out for now?

PR #3065, opened 2014-09-10
Campaign Collection: Dynamic parent signup

aaronschachter:
Adds a `field_automatic_parent_signup` checkbox to the Campaign Group node type.  Closes https://jira.dosomething.org/browse/DS-371
- If a Campaign Group has this checked, signing up for its child Campaign will also create a signup for the parent Campaign Group.
- If it is not checked, the additional signup will not be created.


lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.module
```diff
@@ -153,6 +153,27 @@ function dosomething_campaign_group_get_parent_nid($nid) {
 }
 
 /**
+ * Returns whether or not a Campaign Group nid requires automatic parent signup.
```
sergiitk: I believe “Whether or not” is not correct grammar form in this case. It should be just "whether".
http://english.stackexchange.com/questions/3382/whether-or-not-vs-whether

sergiitk:
:+1:
The small grammar issue is not relevant, great job @aaronschachter!

PR #3066, opened 2014-09-10
UK SSO: Map SSO `phone_number` field to the local `phone_number`.

sergiitk:
#### What's this PR do?
- Fixes an issue with the cell number on the [registration from](http://dev.dosomething.org:8888/user/register) not being sent to the SSO
- Integrates local `phone_number` field with SSO `phone_number` field
#### Testing

New users, created on SSO now have Phone Number  set to actual data provided with the local registration form:  
image

The number saved as is to the DS database:
image
- This PR is the third and the final part of Jira task [#DOS-363](https://jira.dosomething.org/browse/DS-363): "Send new registration fields to UK SSO"


aaronschachter:
:+1:

PR #3068, opened 2014-09-11
DS-360 Use site logo as metatag default for non-campaign pages

aaronschachter:

Uses `hook_preprocess_page` instead of `hook_preprocess_html` to determine whether we're on a node page with a primary image.
- If so: set the relevant metatags with the relevant node field value.
- If not: use the site logo as the default. 

Closes https://jira.dosomething.org/browse/DS-360


sergiitk:
Looks good :+1:

PR #3069, opened 2014-09-11
Campaign Group cleanup

aaronschachter:
- Cleanup Campaign Group form
- For the Hunt (first campaign group), we didn't send child campaign third-party opt-ins.  For Grandparents Gone Wired (a new campaign group), we do want it turned on.  @mikefantini and I are discussing re-working the use for when we use Campaign Group node types, so for now, the easiest thing is to restore child third-party opt-ins and then revisit making this configurable via a Drupal field (if we need it)

Related comment is here: https://jira.dosomething.org/browse/DS-373?focusedCommentId=10672&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-10672 -- For the hunt


sergiitk:
:+1:

PR #3070, opened 2014-09-11
User registration signup hotfix

aaronschachter:
This fixes https://jira.dosomething.org/browse/DS-385 although it's not a good fix because all it does is add the hidden campaign nid later in the function and I'm not sure exactly why it fixes.  

The bug is happening because there is no `$form['nid']['#value']` found in `dosomething_user_login_submit`, which `_dosomething_user_add_signup_data` adds into the Form.  The introduction of `_dosomething_user_register_display_fields` seems to have caused this bug.


sergiitk:
:+1:

aaronschachter:
:+1:

PR #3072, opened 2014-09-11
Neue 5.x

DFurnes:
## Changes
- Updates for compatibility with new Neue 5.x include method.
- Removes IE-specific CSS override file, since we're targeting these with conditional HTML classes.
- Clean up old tasks that we're no longer using.
- Updates `.video` to use new `.media-video` class name.


weerd:
Some sexy coding! :+1:

PR #3073, opened 2014-09-11
DS-386 Apply register_form alters to user_register_form only

aaronschachter:

Restores missing profile fields that `_dosomething_user_register_display_fields` was removing from the `user_profile_form`.


sergiitk:
:+1:

PR #3076, opened 2014-09-12
Signup source

aaronschachter:

Closes https://jira.dosomething.org/browse/DS-372

Adds a `source` column to the Signup table.

Adds a `?source=node/[nid]` to Campaign links found in:
- Campaign Group
- Homepage
- Fact Page

If the `source` query string is available, writes it into the Signup source property.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -304,13 +304,21 @@ function dosomething_campaign_reportback_confirmation_page_access($node) {
  *
  * @param int $nid
  *   The node $nid to return variables for.
+ * @param string $image_size
+ *   The image size to return.
+ * @param string $source
+ *   The signup source to append to the campaign query string.
  *
  * @return array
  *   An array of variables.
  */
-function dosomething_campaign_get_campaign_block_vars($nid, $image_size = '740x480') {
+function dosomething_campaign_get_campaign_block_vars($nid, $image_size = '740x480', $source = NULL) {
   $wrapper = entity_metadata_wrapper('node', $nid);
   $path = 'node/' . $nid;
+  $pretty_path = drupal_get_path_alias($path);
+  if ($source) {
+    $pretty_path .= '?source=' . $source;
+  }
   $image = NULL;
   $image_nid = $wrapper->field_image_campaign_cover->getIdentifier();
   if ($image_nid) {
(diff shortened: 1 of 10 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -304,13 +304,21 @@ function dosomething_campaign_reportback_confirmation_page_access($node) {
  *
  * @param int $nid
  *   The node $nid to return variables for.
+ * @param string $image_size
+ *   The image size to return.
+ * @param string $source
+ *   The signup source to append to the campaign query string.
```
sergiitk: Add `(optional)` before the description.

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -125,8 +135,8 @@ function dosomething_signup_admin_paths() {
  * @param int $uid
  *   Optional - the user uid who has signed up.
  *   If not provided, uses global $user->uid.
- * @param text $data
- *   Optional - serialized data.
+ * @param string $source
+ *   Optional - Signup source.
```
sergiitk: Optional parameter should be marked `(optional)`, not `Optional -`.
https://www.drupal.org/node/1354#param

lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.module
```diff
@@ -180,11 +180,13 @@ function dosomething_campaign_group_is_automatic_parent_signup($nid) {
 function dosomething_campaign_group_get_campaigns($campaign_group_node) {
   $campaigns = array();
   $image_size = '400x400';
+  // Signup source to pass to the query string in campaign links.
+  $source = 'node/' . $campaign_group_node->nid;
```
sergiitk, line 184: Does it make to avoid hardcoding `node/` to source? Is there a possibility user comes from term or other entity?

aaronschachter, line 184: Yes, there will possible sources such as taxonomy term pages, or non Drupal paths such as: `sms`, `android-app`, `ios-app`, `facebook`, `twitter`, etc.

aaronschachter, line 184: The reportback confirmation page,etc.

sergiitk:
Everything else looks good to me. Just add `(optional)` to the docs and it's good to go! :+1:

PR #3077, opened 2014-09-12
Adds Source to signups view

aaronschachter:
@sergii-tkachenko Mind reviewing this one too?

Displays source as a field in the Signups view, and allows filter by source.


sergiitk:
:+1:

PR #3078, opened 2014-09-12
DS-389 Adds signup source to reportback confirmation links

aaronschachter:
@sergii-tkachenko Do you mind reviewing?

Tracks confirmation pages as signup sources.

screen shot 2014-09-12 at 2 00 06 pm


sergiitk:
:+1:

PR #3079, opened 2014-09-12
Fix Mocha XUnit test output.

DFurnes:
# Changes
- Fixes Mocha test output on QA. The `-f` option for mocha-phantomjs is broken on the XUnit reporter. So, BASH it is.

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -523,7 +523,7 @@ function unit_tests {
   printf "\n\033[1;33mRunning JavaScript unit tests with Mocha...\033[00m\n"
   if [[ $DS_MOCHA_OUTPUT == "xunit" ]]
   then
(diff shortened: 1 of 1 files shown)


sergiitk:
As an alternative, take a look at metaskills/mocha-phantomjs#149 and metaskills/mocha-phantomjs#99.
It should be possible to set a [third-party](/visionmedia/mocha/wiki/Third-party-reporters) reporter.

DFurnes:
Definitely. Do you know of a third-party reporter that properly exports XUnit from mocha-phantomjs? I looked quickly and couldn't find an existing one, so this seemed like a quick fix.

(For context, see metaskills/mocha-phantomjs#114.)

sergiitk:
:+1:  Sure quickfix works — even maintainer agrees that's the legit way... I think we always can get back and dig deeper into the `mocha-phantomjs` customization.

PR #3083, opened 2014-09-16
DS-401: CrowdCurity

mshmsh5000:
Need to prove we own the site to test the [CrowdCurity](https://www.crowdcurity.com/) security bounty program.


sergiitk:
:+1:

PR #3084, opened 2014-09-16
Add 'uk_birthday' validator that accepts DD/MM/YYYY format.

DFurnes:
# Changes
- Add 'uk_birthday' validator that accepts DD/MM/YYYY format.
- Add 'use_uk_date_validator` Drupal setting to toggle which date format to use.


sergiitk:
:+1:

PR #3090, opened 2014-09-18
removing kik api script

tjrosario:
Removing kik script since we are not leveraging any of the API at the moment.

Resolves [DS-409](https://jira.dosomething.org/browse/DS-409)


DFurnes:
:shipit:

PR #3091, opened 2014-09-19
Uk form order

DFurnes:
# Changes
- Updates to registration form for UK international site. See [DS-410](https://jira.dosomething.org/browse/DS-410).
# How do I test?

Pull down this branch. There should be no visible changes to the registration form. Run `drush en -y dosomething_uk`, and enable the "Last Name" and "ZIP Code" fields under dosomething_user admin panel. You should now see the additional fields and "First Name" and "Birthday" should be shown full-width [like this](https://cloud.githubusercontent.com/assets/583202/4337077/19e3f300-400c-11e4-83eb-f8fd2d00cc2f.png).
# Questions
- I'm not doing post-code alters here, because it seems like that would have to happen inside dosomething_user, and that seems messy. Does anyone have any ideas if there's a clean way to alter that zip code element, but only in the registration form?


blisteringherb:
:+1:

PR #3092, opened 2014-09-19
Override logo based on dev/staging environment.

DFurnes:
# Changes
- Overrides `$vars['logo']` with an environment-specific logo on local dev and staging to make it easier to tell them apart at a glance.

I had been switching the logo manually each time I pulled down from staging, and got tired of it. :zzz:


barryclark:
:+1: lovely

PR #3093, opened 2014-09-20
Provide UK-specific integration tests.

sergiitk:
#### What's this PR do?
- Provides possibility to create affiliate-specific integration tests: `ds test --intl-integration uk`
- Includes [momentjs](http://momentjs.com/) npm library to simplify work with dates
- Provides random user generation for casper tests
- Provides UK SSO integration failtest:
  - Test that Drupal configure for the UK correctly:
    - OAuth settings must present
    - Date format must be `d/m/y`
  - Test that registration form is integrated with SSO
  - Test that user created from SSO has correct data
  - Test that user created from SSO can login

image

Your vagrant have to be updated to the last baseline.
- `drush en dosomething_uk`
- Set `dosomething_uk_oauth_url`, `dosomething_uk_oauth_key`, `dosomething_uk_oauth_secret` variables
- `drush vset date_format_short "d/m/Y - H:i"`
- Login to VM and cd to `/var/www/vagrant`
- Run `bin/ds test --intl-integration uk`
- Wait for tests to finish

I wrote UK tests in [coffeescript](http://coffeescript.org/) because:
- I was in hurry
- I think casper tests are more compact and readable on coffee
- It's 100% compatible with other test scripts, you can user all js function from `common.js`
- It's already built into `casperjs`, no additional changes is required
- Now we can compare what's more convinient to write scripts.  
  If we agree to not use coffee anymore, I'll just compile tests to pure js and do the readability refactoring

This tests are to ensure nothing gets broken during [#DS-377](https://jira.dosomething.org/browse/DS-377) and [#DS-379](https://jira.dosomething.org/browse/DS-379).

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -98,13 +98,16 @@ ds pull [stage|stage.$AFFILIATE|prod] [--db|--files]
     can be synced to avoid overwriting the html directory.
 
 ds test
-  Run unit and integration tests.
-  
+  Run unit and US integration tests.
+
   --unit
   Runs unit tests using Mocha.
 
   --integration
   Runs integration tests using CasperJS.
+
+  --intl-integration <affiliate> [test]
+  Runs affiliate-specific integration tests.
 
 ds --help
   Displays this help message.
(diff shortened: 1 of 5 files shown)


DFurnes:
Looks good! :+1:

sergiitk:
Awesome! <del>Just one thing to mention — now Zip field is not saved to db, so Postcode test will not pass. I'll merge the fix for zip later.</del> <ins>Added: fixed in #3137.</ins>

PR #3094, opened 2014-09-20
Allow editors manage content aliases.

sergiitk:
#### What's this PR do?
- Grants `editors` role a permission to see, create and edit URL paths on content edit form

Closes [#DS-412](https://jira.dosomething.org/browse/DS-412).

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.features.user_permission.inc
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.features.user_permission.inc
@@ -320,6 +320,7 @@ function dosomething_user_user_default_permissions() {
     'name' => 'create url aliases',
     'roles' => array(
       'administrator' => 'administrator',
+      'editor' => 'editor',
     ),
     'module' => 'path',
   );


aaronschachter:
Do we want all staff to be able to change URL's?  this was restricted from staff by design.

Perhaps a new role would be more appropriate?  cc @mikefantini

mikefantini:
Yeah I'd rather create a new role.

@aaronschachter and I have talked about a few other permissions that might make sense as a sort of super-editor role (like editing the homepage). I'm open to other names but basically this role sit closely with editors but have greater access - between admin and editor.

Open to suggestions on names for this role, but I'd prefer we go that route.

aaronschachter:
Maybe something like `campaigns admin` ?

aaronschachter:
Also @barryclark  -- was Damien granted administrator access? The comment in https://jira.dosomething.org/browse/DS-422 seems to suggest it.  

If so, this PR isn't really relevant anymore and should just be closed.

mikefantini:
yeah campaigns_admin could work, but only if we really restrict the permissions to campaigns relevant functionality. if we want permissions to live outside of campaigns we might want something more generic. editor_plus?

aaronschachter:
Suggestions:
`power editor`
`super editor`

Also, @barryclark @sergii-tkachenko @mshmsh5000 -- thoughts about just granting Damien `administrator` access?  Then no code changes would be needed at all...

barryclark:
@aaronschachter @sergii-tkachenko @mshmsh5000 Sounds good, Aaron.

sergiitk:
@aaronschachter @barryclark @mikefantini So i'm closing this ticket, removing this permission from UK editors and granting admins right to Damien.

PR #3095, opened 2014-09-20
Cancel temporary user accounts at the affiliate sites.

sergiitk:
#### What's this PR do?
- Cancels temporary users accounts on DoSomething UK
- There are more than on and a half million users unrelated to DoSomething UK
- To handle this I provided a module update that removes these users in batches of 50
- Criterias I used to determin wheter a user is to delete:
  - User registration source is not `dosomething_uk`
  - User is not an `editor` or `administrator`
- I tested this on my local setup — works perfect
- Enable `dosomething_uk` module
- Patch your source of this PR
- Run Drupal update

Closes [#DS-411](https://jira.dosomething.org/browse/DS-411).


barryclark:
cc @mshmsh5000 @blisteringherb @aaronschachter for a look

aaronschachter:
This looks good, but we should also be checking for any signups and reportbacks as well.  Hopefully there are none and this can ship as is...

aaronschachter:
Actually.. we should implement [hook_user_delete](https://api.drupal.org/api/drupal/modules%21user%21user.api.php/function/hook_user_delete/7) functions to account for this instead of handling thru an update.

aaronschachter:
Does this mean there are tons of users to delete on the other affiliate sites as well? cc @blisteringherb

aaronschachter:
Confirmed with @blisteringherb yesterday that all affiliate sites have the same accounts which should be deleted.  I suppose one quick fix could be to move this install hook into `dosomething_user`, but only delete the accounts if the `'dosomething_is_affiliate'` variable is set?

sergiitk:
@aaronschachter agree. Should i make this change, what do you think?

aaronschachter:
Sounds ok to me... unless we wanted to just run a manual script... but doing it within install hooks is a better way to document that "yes this script has been run."  so, sure, go for it.  also, welcome back!

lib/modules/dosomething/dosomething_user/dosomething_user.install
```diff
@@ -19,6 +19,78 @@ function dosomething_user_update_7002() {
...
+
+  // …registered with DoSomething UK…
+  $query->leftJoin(
+    'field_data_field_user_registration_source', 'f',
+    'f.entity_id = u.uid'
+  );
+  $where = db_or()
+  ->condition(
+    'f.field_user_registration_source_value',
+    'dosomething_uk',
```
sergiitk: @aaronschachter in Uk there was a logic based on `user_registration_source`, which is set specifically for UK SSO users. It's not possible to keep this logic for all affiliates, they just don't have this field set. But we can use registration date field with hardcoded release date instead. What do you think?

lib/modules/dosomething/dosomething_user/dosomething_user.install
```diff
@@ -19,6 +19,78 @@ function dosomething_user_update_7002() {
 }
 
 /**
+ * Cancels temporary users accounts at the affiliate sites.
+ */
+function dosomething_user_update_7011(&$sandbox) {
```
sergiitk: @aaronschachter Previous update is `7002`, but the module is already `v1.1`. Not sure if this should be `dosomething_user_update_7003` or `dosomething_user_update_7011`.

sergiitk:
OK, this is version working at all affiliates. @aaronschachter, need you help to figure the new logic instead of using UK's `user_registration_source`.

lib/modules/dosomething/dosomething_user/dosomething_user.install
aaronschachter: Can you clarify this: `But we can use registration date field with hardcoded release date instead.` ?

For context, the `user_registration_source` is a field we added very recently.  We used it to track Users we import from an external site, which started around the summer I think.  For the US site, the majority of users will not have a value in this field.

lib/modules/dosomething/dosomething_user/dosomething_user.install
aaronschachter: Yeah we haven't used the module version numbers consistently with our update hooks... we had to add them to prevent Features overrides.  https://github.com/DoSomething/dosomething/pull/1522/files

@sergii-tkachenko Any thoughts on what would make the most sense?  Jumping up to version 1001?  Or keeping consistent, 7011... I don't mind keeping consistent since the same situation happens in a lot of other modules...

mshmsh5000:
How long does this take to run as a single update?

aaronschachter:
Re: https://github.com/DoSomething/dosomething/pull/3095#issuecomment-57672146 ... i'm considering that we just delete all users before a certain timestamp (the date of international launch) who don't have admin/staff roles, for all sites where `dosomething_is_affiliate()` is TRUE.

sergiitk:
Blocks #3242.

sergiitk:
Now the update removes users registered before  [2014-09-03 23:00:00 UTC](http://admin.dosomething.org/jenkins/job/Beta-Deploy/484/).

sergiitk:
Tested locally on QA database: 1343 users removed in 1 minute 17 seconds. Now testing on production DB (UK).

sergiitk:
Local test with UK production database, 1601218 users to delete, manual limit 10,000.
10,000 users accounts cancelled in 15 minutes 33 seconds: 933 seconds.
Avg speed: 10,000 users / 933 seconds = 11 users per second
Estimated time to process one affiliate: 1601218 users / 11 users per second = 145,565 seconds 

It's **40 hours to process 1 affiliate**. I can't guarantee it'll be the same time on production.

sergiitk:
Using drush 1000 users canceled in 26.821 seconds. Should be good to run every minute on cron.
With this speed all users should be cancelled in 27 hours.

Will check if it will take same time when running on UK production DB.

sergiitk:
1000 users UK production: avg 26.315s. I will add a lock though. First run took 3 minutes.

sergiitk:
@aaronschachter @mshmsh5000 This PR is ready for review.

lib/modules/dosomething/dosomething_user/dosomething_user.drush.inc
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * Callback for cancel-us-users command.
+ *
+ * Removes US users from DS affiliates.
+ */
+function drush_dosomething_user_cancel_us_users() {
+  if (!dosomething_settings_is_affiliate()) {
+    return t("Available only on affiliates.");
+  }
+  $lock = variable_get('dosomething_user_cancell_us_users_lock', FALSE);
```
aaronschachter: Minor - but is there a reason there are two l's in `dosomething_user_cancell_us_users_lock` ? If its a mispelling, do you mind find/replacing it?

sergiitk: Sure, will do so.

lib/modules/dosomething/dosomething_user/dosomething_user.drush.inc
```diff
@@ -0,0 +1,100 @@
+<?php
+
+/**
+ *  Implements hook_drush_command
+ */
+function dosomething_user_drush_command() {
+  $items = array(
+    'cancel-us-users' => array(
```
sergiitk, line 8: @aaronschachter what do you think about command name `cancel-us-users`? Maybe `users-cancel-us-accounts` or `dosomething-user-cancel-us-accounts` will be better?

aaronschachter, line 8: Sure, having `accounts` in the name sounds good, but I don't really mind it as is.  I think we may want to just remove this function completely after the import is done.

sergiitk, line 8: Agree. Will leave it as is.

aaronschachter:
Code seems ok, just spelling error in the variable name is really the only (minor) change needed.

sergiitk:
It's already fixed :boom:

PR #3096, opened 2014-09-23
Test redux

DFurnes:
# Changes
- Clean up test helpers a bit.
- Stop using test fixtures for campaigns (avoid crummy performance issues).
- Update test runner to allow different stored test [configurations](https://github.com/DFurnes/dosomething/tree/test-redux/tests/config).
# Test Runner Usage

See `ds test --help` for full configuration options.

screen shot 2014-09-23 at 2 00 47 pm


bin/test
```diff
@@ -0,0 +1,199 @@
...
+// Load a configuration file from the tests/config directory:
+if(program.args.length) {
+  config = require("../tests/config/" + program.args[0] + ".json");
+} else {
+  console.log("Running tests with `default` configuration.".bold.blue)
+  config = require("../tests/config/default.json");
+}
+
+// Set or override unit tests with --unit [test]:
+if(program.mocha) {
```
blisteringherb, line 46: Should there be a space between `if` and `(`?

DFurnes, line 46: I don't think we have a proscribed style for JS. I tend to omit the space, but it's probably something we should decide one way or the other at some point.

blisteringherb:
:+1:

PR #3097, opened 2014-09-23
narrowed msg callout width

### LINKED ISSUE #2738: SMS Games Page - Floating Arrow
Can we tighten the space between this arrow and this text? It may look better with it centered while you are at it. Just ping me on hipchat with your recommendation. 

On BACKSEAT DRIVER:
screen shot 2014-07-02 at 12 50 33 pm

Mocks:
screen shot 2014-07-02 at 12 52 49 pm

tjrosario:
Made width on `.message-callout` on sms campaigns to be more narrow so the text can be closer to the arrow.

Note: enforce content restriction to max out at two lines.

image

Resolves #2738


weerd:
:+1:

PR #3100, opened 2014-09-25
Casper fail screenshots

DFurnes:
# Changes
- Adds a `casper.timestampHelper` method which saves a screenshot with timestamp and optional label.
- Take a screenshot on all test failures for help debugging.


angaither:
:+1: :ghost:

PR #3101, opened 2014-09-25
User delete quiet

DFurnes:
# Changes
- Fixes an issue where the `deleteUserWithEmail` function would die if a user account with that email did not exist, because Drush does not return JSON on errors even if `--format=json` is passed.


aaronschachter:
sure :+1:

PR #3105, opened 2014-09-25
DS-384 - Drops data column from dosomething_signup table

aaronschachter:
We added the `data` column when importing legacy signups for the Mind on my Money campaign in `dosomething_user_update_7001` and have not used the `data` column since.  

It's currently not used anywhere in the site.


angaither:
:fire: :shipit:

PR #3106, opened 2014-09-26
Make sure to wait for confirmation page to appear.

DFurnes:
# Changes
- Ensure confirmation page is loaded before testing for confirmation page content. This wasn't generally an issue, but made the test a bit unreliable.

For review: @angaither


angaither:
:+1:

PR #3108, opened 2014-09-26
Test docs and phantomcss bug

DFurnes:
# Changes
- Improved documentation for test helper functions.
- Explicitly compare generated PhantomCSS diff files, since the `compareAll()` function would sporadically skip running any visual tests.


angaither:
:+1: much clean!

PR #3109, opened 2014-09-26
Pitch Page updates

aaronschachter:
Outputs new fields and alters Pitch Page markup per https://jira.dosomething.org/browse/DS-16
- Implements `hook_forms` to render the signup button twice on the pitch page, once as `signup_button_primary` and again as `signup_button_secondary`
- Only outputs the new fields (`field_secondary_call_to_action` and `field_value_proposition`) if they have been set

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -301,10 +301,9 @@ function dosomething_campaign_preprocess_pitch_page(&$vars, &$wrapper) {
   $vars['campaign']->is_pitch_page = TRUE;
   // Check for a signup button copy override.
   $label = $vars['campaign']->variables['signup_form_submit_label'];
-  // Adds signup_button_primary variable.
-  dosomething_signup_preprocess_signup_button($vars, $label, NULL, 'primary');
-  // Adds signup_button_secondary variable.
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -301,7 +301,10 @@ function dosomething_campaign_preprocess_pitch_page(&$vars, &$wrapper) {
   $vars['campaign']->is_pitch_page = TRUE;
   // Check for a signup button copy override.
   $label = $vars['campaign']->variables['signup_form_submit_label'];
-  dosomething_signup_preprocess_signup_button($vars, $label);
+  // Adds signup_button_primary variable.
+  dosomething_signup_preprocess_signup_button($vars, $label, NULL, 'primary');
+  // Adds signup_button_secondary variable.
+  dosomething_signup_preprocess_signup_button($vars, $label, NULL, 'secondary');
```
blisteringherb: Any desire to refactor the dosomething_signup_preprocess_signup_button() function so it takes the $id as an array, and then you can call this function once with an array of primary and secondary?

aaronschachter: Hmm, that's a good idea.

aaronschachter:
@blisteringherb -- updated per your comments. mind giving another review?

blisteringherb:
:+1:

PR #3110, opened 2014-09-29
Update failing logged-in signup test based on #3109 changes.

DFurnes:
# Changes
- Fixes test for logged-in user signing up for a campaign, based on update form ID in #3109.

/cc @aaronschachter


aaronschachter:
:+1: 

Might be worth adding a test for the secondary button too... (it appears when the `field_secondary_call_to_action` is filled out).

DFurnes:
Yeah, definitely. :+1:

PR #3113, opened 2014-09-29
Bug fix in Signup Data Form for T-shirts

aaronschachter:
Fixes https://jira.dosomething.org/browse/DS-314

Theme preprocess needed to check to make sure the Signup Data Form is not submitted before preprocessing elements.


DFurnes:
:+1:

PR #3114, opened 2014-09-29
Adds School ID to User

aaronschachter:
Storin that school ID, closin out https://jira.dosomething.org/browse/DS-432.

Adds the field via features, and hides it from non-staff users.


DFurnes:
Sure :+1:

PR #3115, opened 2014-09-29
Signup Data Form: Collect school

aaronschachter:
Closes https://jira.dosomething.org/browse/DS-433
- Adds checkbox for editors to configure whether or not to collect school
- Adds `school_id` as a textfield if checked, and stores value in the User's School ID field.


DFurnes:
Sure! :elephant:

PR #3119, opened 2014-09-29
Fixes sporadic timing-related test failures.

DFurnes:
# Changes
- Adds `thenWhenReady` and `thenOpenWhenReady` classes, which wait for `js-ready` class to be added to the `<html`> element. This fixes sporadic test failures if JavaScript (in particular modal initialization) did not finish in time for tests.
- Updates tests to use these new helpers

screen shot 2014-09-29 at 4 06 31 pm

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/modal-links.tpl.php
+++ b/lib/modules/dosomething/dosomething_helpers/modal-links.tpl.php
@@ -19,7 +19,6 @@
 
 <?php if (isset($modals['faq'])): ?>
 <div data-modal id="modal-faq" role="dialog">
-  <a href="#" class="js-close-modal modal-close-button white">×</a>
   <h2 class="banner"><?php print t('FAQs'); ?></h2>
   <?php foreach ($modals['faq'] as $item): ?>
     <h4 class="faq-header"><?php print $item['header']; ?></h4>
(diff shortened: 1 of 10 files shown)


lib/modules/dosomething/dosomething_helpers/modal-links.tpl.php
```diff
@@ -19,7 +19,6 @@
 
 <?php if (isset($modals['faq'])): ?>
 <div data-modal id="modal-faq" role="dialog">
-  <a href="#" class="js-close-modal modal-close-button white">×</a>
```
mshmsh5000, line 22: What do these close buttons have to do with the timing issue?

DFurnes, line 22: Unrelated, but noticed them while debugging a test failure. The modal code generates close buttons programmatically as of [Neue v3.12](https://github.com/DoSomething/neue/releases/tag/v3.12.0) so they were just causing duplicate markup.

DFurnes, line 22: Ah, I just re-read my commit message. I originally thought they were causing the issue, but it seems like it was caused by CasperJS attempting to open another modal before the last one was closed.

DFurnes, line 22: i lied

DFurnes, line 22: To further clarify: the issue that @aa6046f was intending to fix (and @0f5dc67 fixed for realsies) was an inline script error that occurred in the `_cleanup` method that runs after closing modals. It was not actually causing tests to fail (which is what the `thenWhenReady` and `thenOpenWhenReady` helpers in @5733f14 fixed).

aaronschachter:
Looks good here :+1:

PR #3120, opened 2014-09-30
Views Data Export

aaronschachter:
Adds contrib module into repo, and enables for all sites.

Tested locally and confirmed module's working properly with the `node_signups` view.

Closes https://jira.dosomething.org/browse/DS-304

Allows views to be added for https://jira.dosomething.org/browse/DS-443

### DIFF
--- a/lib/profiles/dosomething/dosomething.info
+++ b/lib/profiles/dosomething/dosomething.info
@@ -55,6 +55,7 @@ dependencies[] = variable
 dependencies[] = varnish
 dependencies[] = view_unpublished
 dependencies[] = views
+dependencies[] = views_data_export
 dependencies[] = views_ui
 dependencies[] = xmlsitemap
 dependencies[] = xmlsitemap_custom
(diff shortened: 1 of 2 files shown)


mshmsh5000:
Most excellent :+1: :heavy_check_mark: :hamburger:

mshmsh5000:
One refinement we could make would be to enable this only on staging. Assuming the DB is always on a 24-hour delay, this should be sufficient for anyone to get what they need.

aaronschachter:
I assume you mean enabling only on QA, @mshmsh5000 ?  We'd want to then handle that through the QA import script (to keep it disabled on prod).  I mentioned some minor concerns about performance on prod in the comments on https://jira.dosomething.org/browse/DS-304.  Desmond used to not be too into Views Data Export being on and often used on legacy production.

aaronschachter:
But it's easy enough to remove it from the .info file and enable as needed.

mshmsh5000:
Yes, sorry, I clarified myself [on Jira](https://jira.dosomething.org/browse/DS-304) -- I mean QA for US, staging for others. 

And the confusion here is that our 24-hour data replication goes in the US from prod to QA, and for other sites from prod to staging. File under "would be nice to clean up."

PR #3122, opened 2014-09-30
School dropdown

aaronschachter:
Closes https://jira.dosomething.org/browse/DS-445 by providing checkbox for Users to indicate they are not affiliated with a school.

Refs https://jira.dosomething.org/browse/DS-441 -- creates a dropdown for US states for now.  Creates `dosomething_user_get_administrative_area_options` function which will be able to accept `'CA'` as a parameter for our Canada site.


DFurnes:
Looks good :+1:

PR #3123, opened 2014-09-30
DS-441 Fixes PHP notices

aaronschachter:
Refs comment in https://jira.dosomething.org/browse/DS-441

Fixes PHP notices in `addressfield` module which occur if the `$context` array is empty when being called in `dosomething_user_get_administrative_area_options`:
- Notice: Undefined index: mode in addressfield_format_address_generate() (line 194 of /var/www/beta.dosomething.org/releases/20140930191003/html/profiles/dosomething/modules/contrib/addressfield/plugins/format/address.inc)
- Notice: Undefined index: countries in addressfield_format_address_generate() (line 523 of /var/www/vagrant/html/profiles/dosomething/modules/contrib/addressfield/plugins/format/address.inc).


DFurnes:
Looks great. :checkered_flag:

PR #3130, opened 2014-10-03
Adds revision setting for message_broker_producer

### LINKED ISSUE #3126: Add revision setting to .make for message_broker_producer
Until the `message_broker_producer` is ready for a production release builds of the DoSomething.org site should be locked to a specific release. Add:

```
projects[message_broker_producer][download][revision] = "40e8d9892e378846adf5cb9b295f53c8a533f123"
```

to `/html/profiles/dosomething/drupal-org.make`.

deezone:
Fixes #3126 

Current development on `message_broker_producer` results in merges into master in the repo. These changes should not be included in the current DS builds until the module is ready for the next release -> 0.2.x. A release is used rather than the recommend "tag" by @sergii-tkachenko as the needed state for the DS build is past the last tag.


aaronschachter:
If this has been tested and confirmed to work, ship it :+1:

PR #3134, opened 2014-10-06
Address Administrative Area options variable

aaronschachter:

Sets a variable of State options to use in dropdowns where we also collect additional User addresses, to fix bug reported in comments of https://jira.dosomething.org/browse/DS-452

Adds a function `dosomething_user_set_address_administrative_area_options` to use for international sites (e.g. `drush php-eval "dosomething_user_set_address_administrative_area_options('CA')")` will populate the options for Canada.

### DIFF
--- a/lib/modules/dosomething/dosomething_shipment/dosomething_shipment.forms.inc
+++ b/lib/modules/dosomething/dosomething_shipment/dosomething_shipment.forms.inc
@@ -66,7 +66,7 @@ function dosomething_shipment_form($form, &$form_state, $config) {
   // Add Address form fields.
   dosomething_user_address_form_element($form, $form_state);
 
-  if ($config['shipment_type'] == 'shirt') {
+  if ($config['shipment_item'] == 'shirt') {
     // Add Shirt form fields.
     dosomething_shipment_shirt_form_element($form, $form_state);   
   }
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.address.inc
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.address.inc
@@ -146,7 +146,9 @@ function dosomething_user_save_address_values($values) {
 }
 
 /**
- * Removes extra values from given $options of US States.
+ * Returns options of Addressfield module's US States to remove.
+ *
+ * @see https://github.com/DoSomething/dosomething/issues/1807.
  */
 function dosomething_user_get_us_states_to_remove() {
   return array(
@@ -154,6 +156,11 @@ function dosomething_user_get_us_states_to_remove() {
   );
 }
 
+/**
+ * Removes extra values from given Addressfield Form API $options of US States.
+ *
+ * @see dosomething_user_get_us_states_to_remove().
+ */
 function dosomething_user_remove_extra_us_states(&$options) {
   $states_to_remove = dosomething_user_get_us_states_to_remove();
   // Unset the extra state options.
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.address.inc
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.address.inc
@@ -148,7 +148,7 @@ function dosomething_user_save_address_values($values) {
 /**
  * Returns list of administrative areas based on country.
  */
-function dosomething_user_get_administrative_area_options($country = 'US') {
+function dosomething_user_get_addressfield_administrative_area_options($country = 'US') {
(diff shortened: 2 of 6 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.install
```diff
@@ -19,11 +19,19 @@ function dosomething_user_update_7002() {
 }
 
 /**
+ * Executes dosomething_user_set_address_administrative_area_options().
+ */
+function dosomething_user_update_7003() {
+  dosomething_user_set_address_administrative_area_options();
```
sergiitk: Should you check `dosomething_settings_is_affiliate()` so `dosomething_user_address_administrative_area_options` doesn't get to set to `US` on the affiliates?

lib/modules/dosomething/dosomething_shipment/dosomething_shipment.forms.inc
```diff
@@ -66,7 +66,7 @@ function dosomething_shipment_form($form, &$form_state, $config) {
   // Add Address form fields.
   dosomething_user_address_form_element($form, $form_state);
 
-  if ($config['shipment_type'] == 'shirt') {
+  if ($config['shipment_item'] == 'shirt') {
```
sergiitk, line 69: Should this be fixed also in the [User Shipments](https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_shipment/dosomething_shipment.views_default.inc#L43) view?

lib/modules/dosomething/dosomething_user/dosomething_user.install
aaronschachter: Yeah.. that's not a bad idea, I will add that.  The ideal would be to pass a country code variable here, per https://jira.dosomething.org/browse/INF-17.. although i haven't quite figured out what we'd do in the UK, where there could be various countries.

sergiitk: It think it would totally depend on the campaign specifics. There are not states in most counties.

lib/modules/dosomething/dosomething_shipment/dosomething_shipment.forms.inc
aaronschachter, line 69: Yeah I don't know why it's there.  Features update is not getting rid of it.  I can cleanup that file but it's outside of the scope of this PR.

lib/modules/dosomething/dosomething_user/dosomething_user.install
aaronschachter: Ok, for now I'm going to make this change in order to set it for US, and then call the `dosomething_user_set_address_administrative_area_options` function in a update hook within the DS Canada module.

sergiitk: Ok, let me know when it's ready.

lib/modules/dosomething/dosomething_user/dosomething_user.address.inc
```diff
@@ -146,9 +146,27 @@ function dosomething_user_save_address_values($values) {
 }
 
 /**
+ * Removes extra values from given $options of US States.
+ */
+function dosomething_user_get_us_states_to_remove() {
+  return array(
+    'AA', 'AE', 'AP', 'AS', 'FM', 'GU', 'MH', 'MP', 'PW', 'VI',
```
sergiitk, line 155: Why this states? Maybe explain it in the docblock?

aaronschachter, line 155: Updated.

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -285,27 +285,23 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
...
-  $states_to_remove = array(
-    'AA', 'AE', 'AP', 'AS', 'FM', 'GU', 'MH', 'MP', 'PW', 'VI',
-  );
-
-  // Unset the extra state options.
-  foreach($states_to_remove as $state) {
-    unset($options[$state]);
-  }
-
+  //@todo: Only remove states if address country == US.
```
sergiitk: Add a whitespace before `@` and remove the colon after `todo`.

sergiitk:
:+1:

Issue #2594, opened 2014-06-18
Refactor: Move form element attributes (classes, validators) into theme template.php


aaronschachter:
Move all neue specific classes, validators out of the DS custom module form constructors and into the theme via `hook_form_alter`.

PR #3135, opened 2014-10-06
School finder front end

DFurnes:
# Changes
- **New:** Adds [School Finder](https://sites.google.com/a/dosomething.org/dosomethinguxwiki/home/product-archive/school-finder) signup data form interface for campaigns.
- **Fix:** Fixes a bug where modal skip button would appear even if "Allow Skip" was unset. ([DS-451](https://jira.dosomething.org/browse/DS-451))
- **Fix:** Fixes a reference to the wrong SCSS linting task after removing testing from Grunt task runner.
- **Enhancement:** Refactor checking for `Drupal.setting` in phone number validator, since previous solution relied on non-strict equality checking.

Fixes [DS-436](https://jira.dosomething.org/browse/DS-436), [DS-437](https://jira.dosomething.org/browse/DS-437), [DS-447](https://jira.dosomething.org/browse/DS-447), [DS-451](https://jira.dosomething.org/browse/DS-451), [DS-453](https://jira.dosomething.org/browse/DS-453).
For review: @DoSomething/front-end, @mikefantini 
# Known Issues
- School Finder will not work without disabling CORS protection in your browser. This is a [known issue](https://github.com/DoSomething/dosomething/issues/3133) waiting on a fix.
# Screencast

screencast

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -15,12 +15,12 @@ include_once 'dosomething_user.theme.inc';
  * Implements hook_preprocess_page.
  */
 function dosomething_user_preprocess_page(&$vars) {
-  $schoolFinderAPIEndpoint = variable_get('dosomething_user_school_api_endpoint', 'http://lofischools.herokuapp.com/search');
+  $school_api_endpoint = variable_get('dosomething_user_school_api_endpoint', 'http://lofischools.herokuapp.com/search');
   
   drupal_add_js(
     array('dosomethingUser' =>
       array(
-        'schoolFinderAPIEndpoint' => $schoolFinderAPIEndpoint,
+        'schoolFinderAPIEndpoint' => $school_api_endpoint,
       )
     ),
     'setting'
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -10,6 +10,23 @@ include_once 'dosomething_user.profile.inc';
 include_once 'dosomething_user_valid_address.inc';
 include_once 'dosomething_user.theme.inc';
 
+
+/**
+ * Implements hook_preprocess_page.
+ */
+function dosomething_user_preprocess_page(&$vars) {
+  $schoolFinderAPIEndpoint = variable_get('dosomething_user_school_api_endpoint', 'http://lofischools.herokuapp.com/search');
+  
+  drupal_add_js(
+    array('dosomethingUser' =>
+      array(
+        'schoolFinderAPIEndpoint' => $schoolFinderAPIEndpoint,
+      )
+    ),
+    'setting'
+  );
+}
+
 /**
  * Implements hook_init().
  */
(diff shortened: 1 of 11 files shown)


lib/themes/dosomething/paraneue_dosomething/js/campaign/SchoolFinder.js
```diff
@@ -0,0 +1,153 @@
...
+  var _ = require("lodash");
+  var Validation = require("neue/validation");
+
+  var shouldValidateForm = true;
+  var resultTpl = _.template("<li><a class='js-schoolfinder-result' data-gsid='<%- gsid %>' href='#'><span><%- name %></span> <em><%- city %>, <%- state %></em></a></li>");
+
+  var getResults = function(state, query, callback) {
+    if(_.isEmpty(query) || _.isEmpty(state)) return callback([]);
+
+    $.getJSON("http://lofischools.herokuapp.com/search?state=" + state + "&query=" + query, function(data) {
```
aaronschachter: This should be the `dosomething_user_school_api_endpoint` value, so this will work for other sites.

in php: `drupal_get('dosomething_user_school_api_endpoint')`.

http://drupal.stackexchange.com/questions/4482/how-to-send-variable-from-php-file-to-js-file-in-drupal-7

DFurnes: Oof, yup. Good catch.

angaither:
Nice GIF.

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -10,6 +10,23 @@ include_once 'dosomething_user.profile.inc';
 include_once 'dosomething_user_valid_address.inc';
 include_once 'dosomething_user.theme.inc';
 
+
+/**
+ * Implements hook_preprocess_page.
+ */
+function dosomething_user_preprocess_page(&$vars) {
+  $schoolFinderAPIEndpoint = variable_get('dosomething_user_school_api_endpoint', 'http://lofischools.herokuapp.com/search');
```
aaronschachter: This is very picky, but Drupal standards for variable naming conventions should be lowercase and underscores.

This is minor but I'd argue that since all of the Finder JS lives in Paraneue, that this hook should be called from within the theme's preprocess includes vs. the module... as well as all the additions made to the Signup Data Form... but, happy to save all that for a grand form refactor a la #2594

DFurnes: ~~Re: variable naming, I think we should be using our JavaScript code guidelines here which specify camelCase.~~ Oh, noticed you were talking about the PHP variable. Sure thing.

Re: module vs. theme; I see what you're saying. We use this pattern (doing it in the module) in a couple of other places, so let's keep that consistent until a future refactoring.

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -10,6 +10,23 @@ include_once 'dosomething_user.profile.inc';
...
 include_once 'dosomething_user.theme.inc';
 
+
+/**
+ * Implements hook_preprocess_page.
+ */
+function dosomething_user_preprocess_page(&$vars) {
+  $school_api_endpoint = variable_get('dosomething_user_school_api_endpoint', 'http://lofischools.herokuapp.com/search');
+  
+  drupal_add_js(
```
aaronschachter, line 20: Thought about this last night -- minor, but perhaps something to revisit after, we will only need the JS added for campaigns with school finders (this will be adding to every page)

aaronschachter:
Can't really speak to the JS but seems ok here :+1:

PR #3136, opened 2014-10-06
Add an option to use NFS for Vagrant rather than SSHFS.

DFurnes:
# Changes
- **Enhancement:** Adds option to use NFS for mounting Vagrant box, rather than using SSHFS, if `DS_VAGRANT_MOUNT_NFS` environment variable is set.

For review: @mshmsh5000, @sergii-tkachenko 
# Why?

SSHFS has been crashing multiple times a day for me, and I'd rather trade a little bit of VM performance for the "me" performance lost when restarting Vagrant and force-unmounting drives all day. I've been testing this on my own machine for a few days and it seems to be working better for me. If the `DS_VAGRANT_MOUNT_NFS` environment variable is not set, **this will not change anything** for anyone else.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -10,8 +10,14 @@ Vagrant.configure("2") do |config|
   # SSH Agent forwarding
   config.ssh.forward_agent = true
 
-  # SSHFS -- reverse mount from within Vagrant box
-  config.sshfs.paths = { "/var/www/vagrant" => "../dosomething-mount" }
+  if ENV['DS_VAGRANT_MOUNT_NFS']
+    # NFS
+    config.vm.network :private_network, ip: "10.11.12.13"
+    config.vm.synced_folder ".", "/var/www/vagrant", type: "nfs"
+  else
+    # SSHFS -- reverse mount from within Vagrant box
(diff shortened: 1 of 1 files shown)


sergiitk:
From my experience NFS turned out to be even more unstable and slow. I've made a wiki article about it:
[Vagrant Disc Performance Benchmark](/DoSomething/dosomething/wiki/Vagrant-Disc-Performance-Benchmark).

DFurnes:
Yeah, totally. It's seemed a lot more stable for me, but that may just be my computer being weird. Not advocating for this to be the default setting.

sergiitk:
@DFurnes it just looks like stable. Problems start when because of its asynchronicity, f.i. when you delete a folder with a lot of files, it returns an error because the last file hasn't been deleted at the moment when sshfs receives the command to delete the folder itself.

DFurnes:
I'd like to give it a try for a few more weeks. It's been a breath of fresh air to not have to restart my dev environment so often. Do you mind if I merge this, since its all toggled off behind an env flag?

sergiitk:
Sure, go for it :+1:.

PR #3137, opened 2014-10-06
UK SSO: Store Postal Code in local database.

sergiitk:
#### What's this PR do?
- Saves Postal Code to the local database during automation creation of local users from remote SSO
- <ins>ADDED: Implicitly specifies county in users' address</ins>
- `drush en dosomething_uk`, setup 
- Set `dosomething_uk_oauth_url`, `dosomething_uk_oauth_key`, `dosomething_uk_oauth_secret` variables
- Register and check your profile
  screen shot 2014-10-06 at 11 18 35 pm
- <del>it still not possible to see or edit it on the profile edit form</del>
- <del>as a side effect it hides Country: US from UK users</del>

Part of refactoring [#DS-381](https://jira.dosomething.org/browse/DS-381).

### DIFF
--- a/lib/modules/dosomething/dosomething_uk/dosomething_uk.module
+++ b/lib/modules/dosomething/dosomething_uk/dosomething_uk.module
@@ -268,12 +268,14 @@ function dosomething_uk_new_user_from_remote($name, $password, $sso_user) {
 
   $dob = new DateObject($sso_user['dob']);
   $fields = array(
-    'birthdate'  => $dob->format(DATE_FORMAT_DATE),
-    'first_name' => $sso_user['first_name'],
-    'last_name'  => $sso_user['last_name'],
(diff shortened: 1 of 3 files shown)


aaronschachter:
Code looks ok.. but seeing the `_dosomething_uk_utility_set_user_fields` makes me wonder whether or not we'll need to be implementing a similar function for affiliate sites across the board.  Would it not make sense that this function would just live in `dosomething_user` ?

sergiitk:
Yes, thinking the same.
Plus, I'm not sure if this code is right. I think, I can recall user's profile with Post Code filled in.

I'll check this and move the function to dosomething_user tomorrow.

sergiitk:
Added. When the country is specified implicitly, Post Code does appear at the users' profile edit form.
CC @aaronschachter

aaronschachter:
:+1:

PR #3138, opened 2014-10-07
Member count

aaronschachter:
## New stuff:

Adds get/set methods for a Member Count variable, which will be accessible via the DS API.
- Closes https://jira.dosomething.org/browse/DS-427
## Cleanup:

Moves custom Services resource definitions out of `dosomething_campaign` and `dosomething_user` into `dosomething_api` -- to keep all Services related code in one module.

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -98,36 +98,6 @@ function dosomething_user_permission() {
 }
 
 /**
- * Implementation of hook_services_resources().
- */
-function dosomething_user_services_resources() {
-  $resources = array();
-  $resources['users'] = array(
-    'operations' => array(
-      'create' => array(
-        'help' => 'Create a user.',
-        'file' => array(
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -1150,3 +1120,23 @@ function dosomething_user_save_school_id($school_id) {
...
+}
+
+/**
+ * Sets the Member Count variable.
+ *
+ * @param int $value
+ *   Number of members.
+ */
+function dosomething_user_set_member_count($value) {
+  //@todo Validate that $value is numeric?  Need to define usage.
```
sergiitk, line 1140: A whitespace before `@todo`.

sergiitk:
:+1:

lib/modules/dosomething/dosomething_user/dosomething_user.module
aaronschachter, line 1140: Will fix this in a later PR.

PR #3139, opened 2014-10-07
Enable System set_variable resource in DS API

aaronschachter:

Enables the built-in Services `set_variable` resource to allow other apps to authenticate and then set a Drupal variable.

Removes `dosomething_user_set_member_count` function as this variable will only be set via the API (or could just be set via `variable_get`)

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -1129,14 +1129,3 @@ function dosomething_user_save_school_id($school_id) {
 function dosomething_user_get_member_count() {
   return variable_get('dosomething_user_member_count');
 }
-
-/**
- * Sets the Member Count variable.
- *
- * @param int $value
- *   Number of members.
- */
-function dosomething_user_set_member_count($value) {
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -1129,14 +1129,3 @@ function dosomething_user_save_school_id($school_id) {
...
 }
-
-/**
- * Sets the Member Count variable.
- *
- * @param int $value
- *   Number of members.
- */
-function dosomething_user_set_member_count($value) {
-  //@todo Validate that $value is numeric?  Need to define usage.
```
sergiitk, line 1140: Add a whitespace before `@todo` and remove one before `Need` :bowtie:

sergiitk:
:+1:

lib/modules/dosomething/dosomething_user/dosomething_user.module
aaronschachter, line 1140: But I'm deleting this!

sergiitk, line 1140: Oh, my fault.

PR #3141, opened 2014-10-07
Clear cache perms to Communication Team

aaronschachter:
SMS Reportback paths are inaccessible until cache is cleared.  This allows Communication Team self-serve access.

cc @tongxiang


aaronschachter:
This permission has already been granted on production... just updating features to match it.

PR #3142, opened 2014-10-07
Remove metatag exports

aaronschachter:

Closes https://jira.dosomething.org/browse/DS-429, which has the details.  

All of the various sites should have their own metatags, and it's too crazy to store all of it in code for them all.


sergiitk:
:+1:

PR #3143, opened 2014-10-08
Member Count Variable / Token

aaronschachter:

Adds a Drupal token for the Member Count variable, closes https://jira.dosomething.org/browse/DS-458

Also adds the Member Count variable to the DS User admin form, to allow admins to override.

Preprocesses Home page title with `token_replace` to close https://jira.dosomething.org/browse/DS-454 (Allows Homepage node title to be set to "Join [site:member-count] members.")


sergiitk:
:+1:

PR #3145, opened 2014-10-08
Readable Member Count

aaronschachter:
Adds `$readable` parameter to `dosomething_user_get_member_count` function to optionally return a human readable string like "3.1 million".
- Adds it as a token, closing https://jira.dosomething.org/browse/DS-458
- Outputs it into the Campaign pitch page if set, closing https://jira.dosomething.org/browse/DS-455


sergiitk:
:+1:

PR #3149, opened 2014-10-09
DS API: Get Member Count

aaronschachter:
Closes https://jira.dosomething.org/browse/DS-444

Provides an endpoint to get the Member Count.  Must use POST, as its a custom Services resource action.  Will update wiki upon merge.

**POST** http://dev.dosomething.org:8888/api/v1/users/get_member_count.json

Result:

```
    {
       "formatted": "2,900,343",
       "readable": "2.9 million"
    }
```

### DIFF
--- a/lib/modules/dosomething/dosomething_api/dosomething_api.services.inc
+++ b/lib/modules/dosomething/dosomething_api/dosomething_api.services.inc
@@ -84,6 +84,11 @@ function dosomething_api_default_services_endpoint() {
           'enabled' => '1',
         ),
       ),
+      'actions' => array(
+        'get_member_count' => array(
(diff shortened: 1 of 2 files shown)


angaither:
:+1: 
how/where does the number get set? (or pulled from)

DFurnes:
Coooooool.

aaronschachter:
@angaither -- It's a Drupal variable. @deadlybutter is working on a MemberCounter app that sets the variable via the Drupal app's Services `set_variable` resource.  The DS User admin form also has this variable is exposed, in case we ever need to override it.

angaither:
@aaronschachter sweet!

PR #3151, opened 2014-10-09
Fact Page fixes

aaronschachter:
- Use Markdown for Intro, closes https://jira.dosomething.org/browse/DS-430
- Replace hardcoded CTA with field, closes https://jira.dosomething.org/browse/DS-431


sergiitk:
:+1:

PR #3155, opened 2014-10-10
Updates for Neue 5.1

DFurnes:
# Changes
- Update Bower manifest & minor included path change for Neue 5.1.x.
- Uses refactored Scroll Indicator class hook (with `.js-scroll-indicator` placed on parent container, rather than adding class to each individual link).

:checkered_flag:


sergiitk:
:+1:

PR #3158, opened 2014-10-14
Closed Campaign: Only display Winners section if Scholarship

aaronschachter:

Fixes https://jira.dosomething.org/browse/DS-251

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module
+++ b/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module
@@ -91,6 +91,18 @@ function dosomething_campaign_run_preprocess_text_vars(&$vars, $wrapper) {
  *   Node id of the campaign run node.
  */
 function dosomething_campaign_run_preprocess_winners(&$vars, $nid) {
+  // If this campaign has a scholarship:
+  if (!empty($vars['field_scholarship_amount'])) {
+    // The winners section should be displayed.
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module
```diff
@@ -91,6 +91,18 @@ function dosomething_campaign_run_preprocess_text_vars(&$vars, $wrapper) {
  *   Node id of the campaign run node.
  */
 function dosomething_campaign_run_preprocess_winners(&$vars, $nid) {
+  // If this campaign has a scholarship:
```
sergiitk: Maybe

``` php
if (empty($vars['field_scholarship_amount'])) {
  // Don't display the winners section.
  $vars['display_winners'] = FALSE;
  return;
}

// The winners section should be displayed.
$vars['display_winners'] = TRUE;
```

And then just continue?

sergiitk:
Besides, the comment, everything is :+1:

lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module
aaronschachter: yeah - that's cleaner.  will do

sergiitk: Great, thanks!

PR #3171, opened 2014-10-15
Critical Dries update

aaronschachter:


aaronschachter:
Ran a `ds build` locally to test and all seems well.

Issue #3159, opened 2014-10-14
Taxonomy term template


aaronschachter:
- [x] Create a tpl file for viewing a taxonomy term.
- [x] Currently when viewing certain terms, the page dies:

```
EntityMetadataWrapperException: Unknown data property field_image_campaign_cover. in EntityStructureWrapper->getPropertyInfo() (line 335 of /var/www/vagrant/html/profiles/dosomething/modules/contrib/entity/includes/entity.wrapper.inc).
```
-  [x] Only display Campaign and Grouped Campaign nodes (currently Images can be tagged with Cause, so they show up in the default view)
- [x] CSS Paint

aaronschachter:
@mikefantini Should Grouped Campaign nodes be displayed on a Cause page?  Or only Campaign nodes.

aaronschachter:
@mikefantini Also - what is the desired sort for how campaigns should be listed?

mikefantini:
@aaronschachter yeah I think it makes sense for grouped campaign nodes to be displayed as well. It gets a little funny because we sign you up for the group when you sign up for an individual campaign anyways, but I could see someone wanting to navigate to the group landing page from this page to learn more about the group.

Ideal sort is:
- status = active
- staff pick
- sponsored
- in high season
- newest -> oldest

Let me know if that's doable or needs to be simplified for v1.

aaronschachter:
As we dscussed today, i'm going to attempt to use the existing campaigns view, filtering it by the term being viewed.

aaronschachter:
@mikefantini Assuming that for Cause, we want to filter by `field_cause` , not `field_primary_cause`.

mikefantini:
yeah let's go with field_cause @aaronschachter

aaronschachter:
Finding issues with `dosomething_taxonomy_taxonomy_term_view` ... specifically `hook_taxonomy_term_view` in core.  Thinking that overriding the menu callback per http://www.appnovation.com/blog/custom-taxonomy-term-pages may be the way to implement this.

aaronschachter:
Per https://github.com/DoSomething/dosomething/issues/3159#issuecomment-59257451 -- too crazy to try to pass multiple taxonomy fields as arguments to the /campaigns view.  Using a simplified Taxonomy view, `campaigns_by_term`, for now (which allows the the same view to be used for all terms).  Current view is all published campaign and campaign groups, with prioirty:
- Staff pick first
- Date created descending

[...]

PR #3172, opened 2014-10-16
Custom taxonomy term page

aaronschachter:

This is a work in progress, getting a custom tpl ready for #3159.  It starts on a custom Taxonomy Term page callback, because the core taxonomy's `taxonomy_term_page` function loads all nodes associated with a term, which we don't want.  

We only want `campaign` and `campaign_group` nodes to display, hencing adding the new `campaigns_by_term` view to `dosomething_campaign` module.

Coming next:
- Template file for taxonomy term
- Theme function for the Campaign Group and Taxonomy term to share the same Campaign gallery markup


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.info
```diff
@@ -4,25 +4,18 @@ core = 7.x
 package = DoSomething
 version = 7.x-1.0
 project = dosomething_campaign
-dependencies[] = ctools
```
aaronschachter, line 7: All this removal is just features update doing its thing :ok_woman:

lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.pages.inc
```diff
@@ -0,0 +1,40 @@
...
+  $size = '400x400';
+  foreach ($view->result as $delta => $record) {
+    $campaign_vars = dosomething_campaign_get_campaign_block_vars($record->nid, $size, $source);
+    $campaigns[] = theme('campaign_block', $campaign_vars);
+  }
+  $build['campaigns'] = array(
+    '#theme' => 'item_list',
+    '#items' => $campaigns,
+  );
+  return $build;
```
aaronschachter, line 39: This return value get replaced later with a custom theme function, passing all relevant variables such as `description` and `campaigns` as variables to a tpl file, e.g.

```
$vars['description'] = check_markup($term->description, $term->format);
$vars['campaigns'] = $campaigns;
return theme('dosomething_taxonomy_term_page_view', $vars);
```

sergiitk:
:+1:

PR #3173, opened 2014-10-16
Appends source query string to /campaigns

aaronschachter:
Refs: https://jira.dosomething.org/browse/DS-388

Modifies `explore_campaigns` view to add Signup srouce to query string, e.g. /campaigns/comics-rescue?source=campaigns


sergiitk:
:+1:

PR #3175, opened 2014-10-16
Canada SSO: simple integration tests.

sergiitk:
#### What's this PR do?
- Provides simple integration tests for Canada SSO setup and login
- Setup `DS_CA_USER` and `DS_CA_PASS` variables in VM `/etc/environment`
- Run `ds test canada`
  image
  * Js error is not relevant to this ticket

Now SSO test contain only login, will be extended when the registration is ready.
- Part of [#DS-379](https://jira.dosomething.org/browse/DS-379): Move functionality common to UK and Canada to the new nodule.

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -32,8 +32,6 @@ program
     console.log('    test default --integration=specific/test/dir/**/*');
     console.log('    test uk');
     console.log('    test uk --casperjs=uk/sso');
-    console.log('    test canada');
-    console.log('    test canada --casperjs=canada/sso');
     console.log('');
   })
   .parse(process.argv);
--- a/bin/test
+++ b/bin/test
@@ -32,6 +32,8 @@ program
     console.log('    test default --integration=specific/test/dir/**/*');
     console.log('    test uk');
     console.log('    test uk --casperjs=uk/sso');
+    console.log('    test canada');
+    console.log('    test canada --casperjs=canada/sso');
     console.log('');
   })
   .parse(process.argv);
--- a/tests/integration/canada/sso.coffee
+++ b/tests/integration/canada/sso.coffee
@@ -1,0 +1,50 @@
+# SSO
+
+# Sometimes SSO server response takes long.
+casper.options.waitTimeout = 10000
+
+# Generate user.
+system = require 'system';
+# Temporary. Will be replaced with generated user when SSO registration is ready.
+user =
+  email: system.env.DS_CA_USER
+  password: system.env.DS_CA_PASS
+uid = false
+
+# ------------------------------------------------------------------------
+# Test login form
+
+casper.test.begin "Test that user created from SSO can login", 1, (test) ->
+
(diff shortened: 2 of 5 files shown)


bin/test
```diff
@@ -32,6 +32,8 @@ program
     console.log('    test default --integration=specific/test/dir/**/*');
     console.log('    test uk');
     console.log('    test uk --casperjs=uk/sso');
+    console.log('    test canada');
```
DFurnes: I don't think we need to keep adding each config as a new example. I think the user can be expected to look at what configs are available, rather than patching the binary each time to list them in the help text.

sergiitk: I just put it for easy copy-paste. Should I delete both lines?

tests/integration/canada/sso.coffee
```diff
@@ -0,0 +1,50 @@
...
+  # Run tests.
+  casper.run -> @test.done()
+  return
+
+# ------------------------------------------------------------------------
+# Utilities
+
+# Finds user uid on the profile edit page ans saves it to global variable.
+saveUserUid = ->
+  uid = casper.getCurrentUrl().match(/\/user\/([0-9]+)\/edit$/)[1]
```
DFurnes, line 48: This is super clever. :+1:

tests/integration/canada/sso.coffee
```diff
@@ -0,0 +1,50 @@
+# SSO
+
+# Sometimes SSO server response takes long.
+casper.options.waitTimeout = 10000
+
+# Generate user.
+system = require 'system';
+# Temporary. Will be replaced with generated user when SSO registration is ready.
+user =
+  email: system.env.DS_CA_USER
```
DFurnes, line 10: Should this be stored in `config.js`? I feel like requiring environment variables to be configured before tests can be run may not be super intuitive.

bin/test
DFurnes: I think usually examples would just include one or two "representative" samples, so I'd say so.

tests/integration/canada/sso.coffee
sergiitk, line 10: I really don't like hardcoding user and password.

Yes, it's just a temporary measure.
When the registration implementation is complete, I'll use same flow as in UK with generated users, so I'll know the password in advance.

DFurnes, line 10: Alright, sounds fine for now!

DFurnes:
Oh, `config/default.json` should also be updated to exclude these Canada tests [like so](https://github.com/DoSomething/dosomething/blob/506f171b7b9aa5bd6e66c2a367bec36531d22add/tests/config/default.json#L6).

tests/integration/canada/sso.coffee
sergiitk, line 48: :coffee: :heart:

DFurnes:
:shipit:!

PR #3178, opened 2014-10-17
Campaign gallery

aaronschachter:

Refs #3159
- Adds a Campaign Gallery template file, to share between the Campaign Group and Taxonomy Term page templates.
- Refactors Campaign Block vars to call  `node_load` instead of `entity_metadata_wrapper` to avoid Fatal exceptions when fields aren't set.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/campaign/campaign-gallery.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/campaign/campaign-gallery.tpl.php
@@ -6,10 +6,10 @@
  * - $items: The Campaign gallery items to display.
  */
 ?>
-<div>
-  <ul>
+<section class="container container--campaigns">
+  <ul class="gallery -mosaic -featured">
   <?php foreach($items as $item): ?>
-    <li><?php print $item; ?></li>
+    <li class="campaign -published"><?php print $item; ?></li>
   <?php endforeach; ?>
-  </ul>
+  </li>
 </div>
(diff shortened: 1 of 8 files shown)


lib/themes/dosomething/paraneue_dosomething/templates/campaign/campaign-gallery.tpl.php
```diff
@@ -0,0 +1,17 @@
+<?php
+/**
+ * Returns the HTML for a Campaign Gallery.
+ *
+ * Available Variables
+ * - $items: The Campaign gallery items to display.
+ */
+?>
+<section class="container container--campaigns">
+  <ul class="gallery -mosaic -featured">
```
sergiitk, line 10: Is this right? `-mosaic -featured`?

aaronschachter, line 10: Pretty sure. This is a copy/paste jam from the `campaign-group.tpl.php`... I don't name the classes :)

sergiitk, line 10: Yea, seems so. I just newer seen class names that start with dash before.

sergiitk:
:+1:

lib/themes/dosomething/paraneue_dosomething/templates/campaign/campaign-gallery.tpl.php
aaronschachter, line 10: Me neither @weerd :sake:

weerd, line 10: In fenders we trust :smile: 

@sergii-tkachenko @aaronschachter It's part of our front-end approach. So we have classes that define a "module", like the `.gallery` one above, and then we have modifier classes that start with a dash like `.-mosaic` to indicate that it's a specific type of gallery. Additional dashed classes are just extra modifiers on the main module. Hopefully that makes sense. It's a made up system, but it's been pretty great for CSSing :art:

PR #3179, opened 2014-10-17
Campaign Group cleanup

aaronschachter:
Cleanup from #3178.

The `dosomething_campaign_get_campaign_gallery` now does what `dosomething_campaign_group_get_campaigns` was doing, so we don't need it anymore.

Adds the logic to determine whether or not the Campaign Group is in pre-launch into loop which collects the nodes within the `field_campaign` values.


sergiitk:
:+1:

PR #3181, opened 2014-10-17
Taxonomy term template

aaronschachter:
Refs #3159
- Refactors `dosomething_taxonomy_term_page` to simply return the full view of a taxonomy term, allowing us to make use of built in Taxonomy term template files and preprocess hooks.
- Adds a `taxonomy-term.tpl.php` into the Paraneue theme.
  - Outputs `field_subtitle` in the header
  - Adds inline css to display cover image if set


sergiitk:
:+1:

PR #3184, opened 2014-10-17
Adds secondary page tpl to DRY

aaronschachter:
@DFurnes Can you review?  Per our convo earlier.

Refs #3159

Renames `page--campaign.tpl.php` as `page--full-width.tpl.php`, and preprocesses Campaigns, Campaign Groups, Fact Pages, and Taxonomy Terms to use it.


DFurnes:
Looks great! :shipit:

PR #3185, opened 2014-10-17
Global copy variables

aaronschachter:
Adds variables for global copy to be used within the Taxonomy Term template, and possibly other places.

Adds said copy into Taxonomy Term tpl (refs #3159)


DFurnes:
Looks good to me! :ok_hand:

PR #3195, opened 2014-10-20
Taxonomy Terms admin list

aaronschachter:
Provides table list of all terms in all vocabularies, similar to `admin/content`


sergiitk:
:+1:

PR #3196, opened 2014-10-20
Term video

aaronschachter:
Adds existing `field_video` field collection to Cause, Action Type, and Issue vocabularies.

Outputs into template if present, with same markup as the Campaign Group.


sergiitk:
:+1:

Issue #3162, opened 2014-10-14
Vocabulary fields


aaronschachter:
All user facing vocabularies need the following fields added:
- [x] `field_image_campaign_cover`
- [x] `field_subtitle`
- [x] `field_video`
- [x] `field_partners`
- [x] `field_facts`
- [x] `field_intro_title`

mikefantini:
@aaronschachter we also need:
- title
- facts

aaronschachter:
For title, we'll just be using the Term name.  Does Fact appear on every vocabulary?

aaronschachter:
@mikefantini Confirming: Assuming we want to display the Partner modals (e.g. Why we <3 Toyota) as well? with the video/image/copy fields?

mikefantini:
@aaronschachter let's start with facts just on cause and issue for now.

and yeah we'll want the partner modals on collections that have partners. If you want we can just add that to the campaign collection vocabulary and then add it to other vocabularies as needed.

I could see us wanting it for issue and holiday as well potentially.

mikefantini:
Facts should be listed immediately after the display of campaigns. They should be under the header Facts About [Title].

The display of these facts should use this style: http://neue.dosomething.org/#typography---unordered-lists.

We should also have the Sources animation that we use on the action page and on facts pages.

mikefantini:
@aaronschachter design team wants there to be a field_intro_title for these to break up page a little bit so that it's not just full bleed image above our campaign tiles. can you add this to the fields needed above?

aaronschachter:
We can add it, but there's always going to be a description in the text, which would prevent said situation...

mikefantini:
right sorry they wanted this purple header to further break up the page. The 1-2 lines of text between the header and the campaigns didn't feel like enough.

PR #3201, opened 2014-10-20
Taxonomy Term partners

aaronschachter:
- Adds `field_partners` field collection to Cause, Issue, and Action Type vocabularies
- Refactors `dosomething_taxonomy_get_partners_data` to expect the `field_partners` field array, instead of a node, allowing a taxonomy term to call the function as well (Refs #3162)
- Adds `paraneue_dosomething_get_sponsor_logos` helper function to return the `sponsor_logos` template from given `$partners_data` array (as returned by `dosomething_taxonomy_get_partners_data`)

Some of this PR works towards moving all Partners preprocessing to the  `dosomething_taxonomy_get_partners_data`, which is used by Campaigns and now Taxonomy Terms in order to start DRY.  ~~An ongoing todo has been to refactor the `node--campaign-group.tpl.php` to use the same `sponsor_logos` template.~~

^^^
**UPDATE** Refactors `campaign_group` to also use the same `sponsor_logos` theme function.


lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
```diff
@@ -105,22 +105,18 @@ function dosomething_taxonomy_get_campaigns($tid) {
...
-  $logo_style = 'wmax-423px';
+  // Intialize return array.
+  $partners = array();
 
   // Loop through field collection items.
-  foreach ($wrapper->field_partners->value() as $delta => $item) {
+  foreach ($field_partners as $delta => $item) {
 
-    $fc_item = entity_metadata_wrapper('field_collection_item', $item);
+    $fc_item = entity_metadata_wrapper('field_collection_item', $item['value']);
```
sergiitk, line 121: It's better to make a check like `if (empty($item['value'])) { continue; }` here.

lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
```diff
@@ -180,8 +179,8 @@ function dosomething_taxonomy_get_partners_vars($nid) {
   $info_count = 0;
 
   // Loop through field collection items.
-  foreach ($wrapper->field_partners->value() as $delta => $item) {
-    $fc_item = entity_metadata_wrapper('field_collection_item', $item);
+  foreach ($field_partners as $delta => $item) {
+    $fc_item = entity_metadata_wrapper('field_collection_item', $item['value']);
```
sergiitk, line 188: And here.

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -88,3 +88,28 @@ function paraneue_dosomething_get_social_networks() {
...
+      $sponsors[$delta]['name'] = $partner['name'];
+      $sponsors[$delta]['logo_url'] = $partner['logo']['url']['default'];
+    }
+  }
+  if (!empty($sponsors)) {
+    return theme('sponsor_logos', array(
+      'sponsors' => $sponsors,
+    ));
+  }
+  return NULL;
```
sergiitk: Should be a string, e.g. `''`.

PR #3202, opened 2014-10-20
Homepage sponsor admin

DFurnes:
# Changes
- Allows sponsors on the homepage to be updated on the homepage node (based on Partners taxonomy).
- Adds a new image style with expected dimensions for homepage partner images & a desaturate filter.
# Example

screen shot 2014-10-20 at 5 33 14 pm


lib/modules/dosomething/dosomething_home/dosomething_home.module
```diff
@@ -14,18 +14,15 @@ function dosomething_home_preprocess_node(&$vars) {
...
 
+  // Prepare campaign thumbnails for homepage
   $campaigns = $vars['field_campaigns'];
-
-  if (empty($campaigns)) {
-    // Fallback to old stuff.
-    $campaigns = array(362, 832, 8, 843, 345, 31, 15, 17, 850);  
-  }
-
   foreach ($campaigns as $key => $campaign) {
```
aaronschachter, line 26: Minor, but without the fallback, this will return a PHP error if `$campaigns` is not set.

DFurnes, line 26: Fair point. Should the fallback just be to create an empty array then? Falling back to arbitrary campaign NIDs seems wonky to me.

lib/modules/dosomething/dosomething_home/dosomething_home.module
```diff
@@ -60,5 +57,25 @@ function dosomething_home_preprocess_node(&$vars) {
     $thumbnails[$nid] = theme('thumbnail', $tiles);
   }
 
+  // Prepare sponsor logos for homepage
+  $partners = array();
+  $wrapper = entity_metadata_wrapper('node', $vars['node']);
```
aaronschachter, line 62: Entity metadata wrappers are kind of a beast I introduced into the codebase that I'm hoping we refactor enough to move away from. Mainly because if a field isn't set, the site breaks with that "Fatal exception" WSOD... which I didn't realize would happen so much when features get out of date (was happening on international sites regularly)

This is ok to keep for now, but can you just make sure / confirm that adding a Partner term without a logo doesn't break the homepage?

aaronschachter:
Looks great, please just confirm that the preprocessing gracefully handles exceptions when terms are added without sponsor logos.  If so, shipppps it

lib/modules/dosomething/dosomething_home/dosomething_home.module
aaronschachter, line 26: Sure, sounds good.

DFurnes:
Yep, current behavior for sponsors without images is just to leave `src` blank and print the `title` attribute:

screen shot 2014-10-21 at 10 26 30 am

aaronschachter:
:+1: SHIP IT VALUE ADDED

lib/modules/dosomething/dosomething_home/dosomething_home.module
```diff
@@ -60,5 +57,25 @@ function dosomething_home_preprocess_node(&$vars) {
...
+
+  // Loop through field collection items.
+  foreach ($wrapper->field_sponsors->value() as $delta => $item) {
```
aaronschachter, line 65: This should make sure that `isset($wrapper->field_sponsors)` first before checking.  The fear is real: QA internationl build did not revert features and is now WSOD on

EntityMetadataWrapperException: Unknown data property field_sponsors. in EntityStructureWrapper->getPropertyInfo() (line 335 of /var/www/international.dosomething.org/releases/20141021190304/html/profiles/dosomething/modules/contrib/entity/includes/entity.wrapper.inc).

DFurnes, line 65: Uff da. I'll put up a patch.

PR #3203, opened 2014-10-21
Active Campaign preview page

aaronschachter:

If a campaign is closed, you can't preview it's active state without setting its status back to Active.

This PR adds a menu callback, `node/*/active`, which will only display for Closed Campaigns.  The active campaign can be previewed from here.  This page is available to staff only.

Also changes the Closed Campaign preview page link to display only when the Campaign is active, eliminating duplicate links (the regular view page is the same as the Closed page when the Campaign is closed)

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -57,6 +57,16 @@ function dosomething_campaign_menu() {
     'type' => MENU_LOCAL_TASK,
     'weight' => 60,
   );
+  // Internal active page for staff.
+  $items['node/%node/active'] = array(
+    'title' => 'Active',
+    'page callback' => 'node_page_view',
+    'page arguments' => array(1),
+    'access callback' => '_dosomething_campaign_active_page_access',
+    'access arguments' => array(1),
+    'type' => MENU_LOCAL_TASK,
+    'weight' => 70,
+  );
   // Internal closed page for staff.
   $items['node/%node/closed'] = array(
     'title' => 'Closed',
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -246,8 +256,20 @@ function _dosomething_campaign_closed_page_access($node) {
...
+}
+
+/*
+ * Determines whether a user has access to the active page callback.
+ */
+function _dosomething_campaign_active_page_access($node) {
+  // Staff only.
+  if (!dosomething_user_is_staff()) { return FALSE; }
+
+  return (dosomething_campaign_is_closed($node));
```
sergiitk: Extra parentheses?

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -246,8 +256,20 @@ function _dosomething_campaign_closed_page_access($node) {
 
   // Only display closed page link if a campaign run exists for node.
   $run_nid = dosomething_campaign_run_get_closed_run_nid($node->nid);
-  // Closed page available if this is a campaign node with a campaign run.
-  return ($node->type == 'campaign' && $run_nid);
+
+  // Closed page available if this is an active campaign node with a campaign run.
+  $is_active = !dosomething_campaign_is_closed($node);
+  return ($node->type == 'campaign' && $run_nid && $is_active);
```
sergiitk: Better to move the expression to variable?

sergiitk:
:+1:

PR #3176, opened 2014-10-16
Canada SSO: adapt login to common flow provided in external_auth.

sergiitk:
#### What's this PR do?
- Adapts Dosomething Canada module: use login flow provided in [`external_auth`](/dosomething/drupal-external-auth)
#### How should this be manually tested?
- Enable and setup Dosomething Canada module
- Setup `DS_CA_USER` and `DS_CA_PASS` variables in VM `/etc/environment`
- Run `ds test canada`  
  image
#### What are the relevant tickets?
- Part of [#DS-379](https://jira.dosomething.org/browse/DS-379): Move functionality common to UK and Canada to the new nodule
- #3174: UK SSO: adapt login/signup to common flow provided in external_auth
- dosomething/drupal-external-auth#2

PR #3174, opened 2014-10-16
UK SSO: adapt login/signup to common flow provided in external_auth.

sergiitk:
#### What's this PR do?
- Adapts Dosomething UK module: use login/signup provided in [`external_auth`](/dosomething/drupal-external-auth)
#### How should this be manually tested?
- Enable and setup Dosomething UK module
- Run `ds test uk`  
  image
#### What are the relevant tickets?
- Part of [#DS-379](https://jira.dosomething.org/browse/DS-379): Move functionality common to UK and Canada to the new nodule.
- dosomething/drupal-external-auth#2

PR #3205, opened 2014-10-21
SSO Canada: registration

sergiitk:
#### What's this PR do?
- Utilizes Canada SSO registration in `dosomething_canada` module
- Improves and extends Canada integration tests
- Enable and setup Dosomething Canada module
- Run `ds test canada`  
  image
- Part of [#DS-379](https://jira.dosomething.org/browse/DS-379): Move functionality common to UK and Canada to the new nodule
- #3176: Canada SSO: adapt login to common flow provided in external_auth
- #3174: UK SSO: adapt login/signup to common flow provided in external_auth
- dosomething/drupal-external-auth#2
### Known integration issues

Users created not able to login at http://www.tigweb.org/:
- [ ] The User can only restore its own username (which is not same as email) using [TIG Recover Username](https://www.tigweb.org/members/retrieve.html) form. The user will receive email like the following, but it's still not possible to log in using the username and its original password. We think the issue could be in the dash sign in `ds-sergiitest5dosomethingorg`
  image
- [ ] It's not possible to recover password on TIG using both recovered username and original email
- [ ] It's not possible to login on TIG using both recovered username and original email

### DIFF
--- a/lib/modules/dosomething/dosomething_canada/includes/dosomething_canada_register_controller.inc
+++ b/lib/modules/dosomething/dosomething_canada/includes/dosomething_canada_register_controller.inc
@@ -10,15 +10,53 @@ class DosomethingCanadaRegisterController implements ExternalAuthRegisterControl
   // ---------------------------------------------------------------------
   // Instance variables
 
+  /**
+   * The registration form.
+   */
+  private $account_values;
+
+  /**
+   * The SSO controller.
+   */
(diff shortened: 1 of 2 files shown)


sergiitk:
Known integration issues added.
CC @mfurdyk @mshmsh5000

aaronschachter:
:+1:

PR #3208, opened 2014-10-21
Taxonomy Term Facts

aaronschachter:
Adds existing `field_facts` Entity Reference field, and uses existing `dosomething_fact_get_fact_field_vars` function for easy preprocessing.  Outputs into template.


lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
```diff
@@ -68,6 +68,14 @@ function dosomething_taxonomy_preprocess_taxonomy_term(&$vars) {
     }
   }
 
+  // Facts.
+  $vars['facts'] = array();
+  $wrapper = entity_metadata_wrapper('taxonomy_term', $vars['tid']);
```
sergiitk, line 73: Maybe check `$vars['tid']` or `$wrapper->field_facts` to avoid possible fatal "no such method count()"?

lib/themes/dosomething/paraneue_dosomething/templates/taxonomy/taxonomy-term.tpl.php
```diff
@@ -38,6 +38,25 @@
...
+        <ul>
+        <?php foreach ($facts['facts'] as $fact): ?>
+          <li><?php print $fact['fact']; ?> <sup><?php print $fact['footnotes']; ?></sup></li>
+        <?php endforeach; ?>
+        </ul>
+        <section class="sources">
+          <h3 class="__title js-toggle-sources"><?php print t('Sources'); ?></h3>
+          <ul class="__body legal">
+            <?php foreach ($facts['sources'] as $key => $source): ?>
+              <li><sup><?php print ($key + 1); ?></sup> <?php print $source; ?></li>
```
sergiitk: Extra parentheses after `print`?

lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
aaronschachter, line 73: I'll check for the `$wrapper->field_facts`... I think its pretty safe to assume `$vars['tid']` will exist in a `hook_preprocess_taxonomy_term`.

lib/themes/dosomething/paraneue_dosomething/templates/taxonomy/taxonomy-term.tpl.php
aaronschachter: I can make this change, its copied from the Fact Page tpl though..

sergiitk:
:+1:

PR #3210, opened 2014-10-21
Check for field existance before preprocessing homepage.

DFurnes:
# Changes
- Checks if fields exist on the homepage before going to work. This prevents errors if features do not properly revert and everything goes kablooie.

For review: @aaronschachter


aaronschachter:
Nice branch name :+1:

DFurnes:
:dancer:

PR #3213, opened 2014-10-22
Variable for dosomething_user_address_country

aaronschachter:
This prevents the Shipment Form (and other Signup Data forms that collect addresses) from defaulting to 'US' when creating address field values.

However, we're still in trouble for Teens For Jeans Canada because upon creating, users are created a default value with country "US" because of the DoSomething User field definition: https://jira.dosomething.org/browse/DS-449


blisteringherb:
:+1:

PR #3216, opened 2014-10-22
Communications team - view unpublished access

aaronschachter:
Let them test campaigns!


aaronschachter:
This permission has been granted on production, which is why the DoSomething User feature is currently overridden.  Should be set back to "default" upon next deployment.

Issue #3223, opened 2014-10-23
UK SSO: newly registered users has wrong country


sergiitk:
![image](https://cloud.githubusercontent.com/assets/672669/4759081/9db5bcca-5ae4-11e4-8974-c191f68d1a2e.png)

Issue #3217, opened 2014-10-22
Save CA as address country for DS Canada User Registration


aaronschachter:
Discussed TFJ Canada with @sergii-tkachenko .  For now, we'll keep the User Address default country to US, and the DS Canada SSO classes (#3205) will set the User's country.

This should allow the Shipment Form to collect User addresses.  Current functionality is that User's `field_address` value is set to US upon account creation, so when you access the Shipment Form (via `dosomething_user_address_form_element` -- https://github.com/DoSomething/dosomething/blob/v0.3.62/lib/modules/dosomething/dosomething_user/dosomething_user.address.inc#L52), the US states are displayed, since the existing address value is saved with US.

PR #3224, opened 2014-10-23
UK SSO: Add country field integration test.

sergiitk:
#### What's this PR do?
- Adds an integration tests that checks if the user's country is UK
- Enable and setup Dosomething UK module
- Run `ds test uk`  
#3217 — Fix address field country on DS Canada
#3223 — UK SSO: newly registered users has wrong country
### Known integration issues

This test also revealed that user's created during registration on DS UK has `US` country: #3223
The ones created during login works well:
image

ADDED: At this point the test is expected to fail the first time and pass the second time (see screenshot above).

### DIFF
--- a/tests/integration/uk/sso.coffee
+++ b/tests/integration/uk/sso.coffee
@@ -16,6 +16,8 @@ FIELD_LAST_NAME  = 'field_last_name[und][0][value]'
 FIELD_BIRTHDATE  = 'field_birthdate[und][0][value][date]'
 FIELD_POSTCODE   = 'field_address[und][0][postal_code]'
 
+USER_COUNTRY = 'UK'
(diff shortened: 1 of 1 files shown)


aaronschachter:
Does this test need https://github.com/DoSomething/drupal-external-auth/pull/3 in order to pass?  If so, doesn't mean that mean we need to change the commit to pull from in the make file?

sergiitk:
@aaronschachter it needs not only DoSomething/drupal-external-auth#3, but also new functionality in main repo + the change in the make file to pass. That's why I had to split it, it's cross-repo dependent.
I'll fix the thing in next PR.

aaronschachter:
:+1:

PR #3230, opened 2014-10-24
Canada SSO: fixes users address country.

### LINKED ISSUE #3217: Save CA as address country for DS Canada User Registration

sergiitk:
#### What's this PR do?
- Changes user's address on DS Canada so new users created during login and registration have 'CA' in country field
- Changes `dosomething_user`: `_dosomething_user_register_display_fields` now doens't unset address, but restrict access to it 
- Run tests after cloning this branch
- Test that functionality works ad before after `dosomething_user` change: `ds test default`
  screen shot 2014-10-23 at 6 42 11 pm
- Enable and setup DoSomething  Canada module
- Test Canada-specific functionality: `ds test canada`
  screen shot 2014-10-23 at 6 44 29 pm

Fixes #3217 — Save CA as address country for DS Canada User Registration

### DIFF
--- a/lib/modules/dosomething/dosomething_canada/dosomething_canada.module
+++ b/lib/modules/dosomething/dosomething_canada/dosomething_canada.module
@@ -7,7 +7,6 @@
 
 define('DOSOMETHING_CANADA_WATCHDOG', 'dosomething_canada');
 define('DOSOMETHING_CANADA_USER_SOURCE', 'dosomething_canada');
-define('DOSOMETHING_CANADA_USER_COUNTRY', 'CA');
 
 // In lieu of a future, global management solution, standardize feedback
 // messages here.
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_canada/dosomething_canada.module
```diff
@@ -7,6 +7,7 @@
 
 define('DOSOMETHING_CANADA_WATCHDOG', 'dosomething_canada');
 define('DOSOMETHING_CANADA_USER_SOURCE', 'dosomething_canada');
+define('DOSOMETHING_CANADA_USER_COUNTRY', 'CA');
```
aaronschachter: There is a `dosomething_user_address_country` variable which I added in #3213.  I suppose it doesn't hurt to have a constant here, but it feels kinda duplicated...

aaronschachter:
This looks pretty good except for the one DRY type comment. wWe'll also need to determine whether or not this needs to be updated for existing Canadian users... I'm assuming we will..

aaronschachter:
:+1:

PR #3232, opened 2014-10-24
Adds missing Signup Source to Campaign Group gallery

aaronschachter:

Something I missed when refactoring Campaign Groups and Taxonomy terms to share the same template file :fries:


sergiitk:
:+1:

PR #3236, opened 2014-10-24
Canada SSO: improve intergration tests.

sergiitk:
#### What's this PR do?
- Refactors and improves Canada SSO integration tests
- Makes Canada tests use the same test flow as in UK SSO
  image
- Enable and setup DoSomething Canada module
- Run `ds test canada`


DFurnes:
:shipit:

PR #3238, opened 2014-10-24
Adds isset checks to avoid PHP Notice undefined index

aaronschachter:
Cleanup in `dosomething_campaign_preprocess_common_vars` to avoid lots of 

```
Notice: Undefined index: partner_info in dosomething_campaign_preprocess_common_vars() (line 92 of /var/www/vagrant/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc).
```


sergiitk:
:+1:

PR #3246, opened 2014-10-27
Delete all user data upon user delete

aaronschachter:
Renames `dosomething_user_delete_all_data` as `dosomething_user_delete_all_data`, and calls it within a `hook_user_delete`.  

Will be super useful for cleaning up test data international sites.


sergiitk:
You missed another one watchdog [here](https://github.com/DoSomething/dosomething/pull/3246/files#diff-ad30a5d569ca2ae6d2e04223aa501104R1147).

aaronschachter:
Good catch!  Updated.

sergiitk:
:+1:

Issue #3242, opened 2014-10-27
Update Canada User addresses


aaronschachter:
Users currently imported have US stored in their `field_address`, which mean they'll be presented with the US states dropdown in the Shipment Form for Canada TFJ.

Update relevant Users to save `'CA'` to their `field_address` country, and confirm that the Shipment Form (test on https://canada.dosomething.org/node/1227) displays Canadian states.

PR #3249, opened 2014-10-27
Provide the affiliate code to country code mapping.

sergiitk:
#### What's this PR do?
- Provides a common way to get ISO country code of current affiliate
#3242


aaronschachter:
I like this, although it introduces some duplication in the `dosomething_user_address_country` variable which we have started to use.

Can we replace the `dosomething_user_address_country` variable calls with this function to avoid confusion?

aaronschachter:
Discussed with @sergii-tkachenko -- we'll refactor the `dosomething_user_address_country` variable with this new function `dosomething_settings_get_affiliate_country_code` in the next pull request.

Ship it value added :+1:

PR #3262, opened 2014-10-28
Profile cleanup

aaronschachter:
@DFurnes Can you review?

Kills redundant page.tpl.php's, and the Recommended Campaigns section, which is no longer displayed in the Profile per mocks.


DFurnes:
Looks good to me! :anchor:

Issue #3250, opened 2014-10-27
need config variable for "Need help locating your school" copy


mikefantini:
we'll want a config variable for this copy for the school finder in case it updates frequently as we're trying to react to questions and user confusion

wires: https://a0465d35-a-5d563b7e-s-sites.googlegroups.com/a/dosomething.org/dosomethinguxwiki/home/product-archive/school-finder/School%20Finder%20Wires%201.0.png?attachauth=ANoY7cpeKk4nt8wkj2e1iJd7IQv3zS_ByO1bU9RIjwK4pmNryd8w99ignAfWnUardavbnzJKF7Y-8nn9AMMhmpvgdQn2gcbfMyTrb_v_kQqjpdy6-aYbgz7aLV2fGftA72CohfqlRRzkAFQsdVoBo0TQLl5NNGsIyuHUHiyyLGUi71DZMj6-vTnOExs3EiR6JVCVJt7supo03kwg9Eo9mIp0OX2CgfBcf6iFSwN6L4hX-1-8RyZZxoPUIFS2TikL_Rct5Csl5WiwDrTyy_DSJ6GhbJHNFzjFvVSU3UqPvAhLZ202IUT0PTg%3D&attredirects=0

aaronschachter:
@DFurnes I can add a new variable for this into the DS User admin config form and pass the variable to the `dosomething_user_preprocess_page` function... (which i'm thinking actually would be cleaner to move into paraneue preprocess instead) .. but where is the copy displayed?  Would you be able to pick this up once I set up the variable?

aaronschachter:
@mikefantini The copy can be overwritten by an admin at `admin/config/dosomething/dosomething_user`.  If no variable is present, the current copy on the site will be used.

mikefantini:
awesome thanks

PR #3263, opened 2014-10-28
School Help Text variable

aaronschachter:
@DFurnes Can you review?

Adds a variable into the DS User admin config form, to pass #3250 over to Team Fender.


aaronschachter:
Ideally the JS would use the current copy if this variable is not set.

DFurnes:
School finder help can be overridden [here](https://github.com/DoSomething/dosomething/blob/cf1a180a67505979189e419318281db0c3578a78/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc#L373). We're just toggling visibility of that form element on and off.

aaronschachter:
Got it.  Updating PR now

aaronschachter:
@DFurnes updated

DFurnes:
:shipit:

Issue #3261, opened 2014-10-28
Profile template v1.0


aaronschachter:
New User Profile sections for v1.0: https://sites.google.com/a/dosomething.org/dosomethinguxwiki/home/product-archive/profile-template
- [x] You're Doing
  - Displays active standard campaigns which the user has signed up for (Campaign Type == 'campaign' && Campaign is not closed)
  - If no campaigns, displays copy variables
- [x] You Did -- Displays all user reportbacks.  
  - If no reportbacks, this section does not display.
- [x] Account Info, with link to Edit profile

aaronschachter:
@mikefantini @lkpttn The mocks show inconsistent header for the Prizes section between the empty and filled designs.  Is the header to be "My Swag" ?

lkpttn:
These will all be "Your _____" and swag should be prizes. So "Your Prizes".

aaronschachter:
Closing this and creating separate issues for remaining Profile sections.

PR #3268, opened 2014-10-28
User Profile: I'm Doing

aaronschachter:
First pass at the "I'm Doing Section" per #3261.
- Adds `dosomething_campaign_get_campaigns_doing()` function to return active standard Campaigns a User has signed up for, but not reported back on yet.
- Displays No Signups Header / Copy if User is not doing any campaigns
- Removes `hook_user_view` and `profile.inc` files to keep Profile preprocessing in one file (theme.inc)
- Removes unneeded address elements as per #3261
## todo:
- Display "You're Doing" campaigns as Gallery items instead of Campaign blocks
## Screenshots

profile-v1


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -323,6 +323,41 @@ function dosomething_campaign_reportback_confirmation_page_access($node) {
...
+  foreach ($signup_nids as $delta => $nid) {
+    // If they reported back, they already did it.
+    if (dosomething_reportback_exists($nid, $uid)) {
+      continue;
+    }
+    $node = node_load($nid);
+    // If the campaign is closed, they're not doing it anymore.
+    $is_closed = dosomething_campaign_is_closed($node);
+    // SMS Games should not appear in the Doing list.
+    $is_sms = dosomething_campaign_get_campaign_type($node) == 'sms_game';
```
sergiitk: The expression is string comparison so it must use `===` operator:

``` php
php > var_dump('sms_game' == true);
bool(true)
php > var_dump('sms_game' === true);
bool(false)
```

Also it'd be easier to read if the expression was in parentheses.

lib/themes/dosomething/paraneue_dosomething/templates/user/user-profile.tpl.php
```diff
@@ -3,43 +3,22 @@
...
- *     - [premise]: User's apartment, suite, etc... (string).
-*   - [address]: Array containing address information only for submitted data.
- *     - [country]: User's country of residence (string).
- *     - [state]: User's state of residence (string).
- *     - [city]: User's city of residence (string).
- *     - [zip]: User's postal zip code (string).
- *     - [street]: User's street address (string).
- *     - [premise]: User's apartment, suite, etc... (string).
- *   - [signedup]: Array containing a data for campaigns a User has signed up for.
+ * Available Variables.
```
sergiitk: Please Lowercase `Variables` and replace the dot with a colon.
See https://www.drupal.org/node/1354#templates

lib/themes/dosomething/paraneue_dosomething/templates/user/user-profile.tpl.php
```diff
@@ -3,43 +3,22 @@
...
+ *  - $title (string).
```
sergiitk: Colon after `$title`.

lib/themes/dosomething/paraneue_dosomething/templates/user/user-profile.tpl.php
```diff
@@ -3,43 +3,22 @@
...
+ *  - $subtitle (string).
```
sergiitk: Same here.

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -323,6 +323,41 @@ function dosomething_campaign_reportback_confirmation_page_access($node) {
...
+ * @return array
+ *   Array of Campaign node nid's.
+ */
+function dosomething_campaign_get_campaigns_doing($uid = NULL) {
+  if (!$uid) {
+    global $user;
+    $uid = $user->uid;
+  }
+  // Get all signups.
+  $signup_nids = dosomething_signup_get_signup_nids_by_uid($uid);
```
sergiitk, line 340: Check if `$signup_nids` is set: dosomething_signup_get_signup_nids_by_uid() could be returning `NULL`.

sergiitk:
:+1:

PR #3272, opened 2014-10-28
Deployments: flush the Redis cache before running drush.

### LINKED ISSUE #3211: International build - features not reverting
QA itnl currently broken because features not reverted, and new fields not created.  Production sites show quite a few features are not reverted as well.

sergiitk:
This PR is to avoid  a race condition state between DB and cache, which causes features recreation to fail during international deployments. See #3211.
#### What's this PR do?
- Brings new step into `ds build` that clears Redis cache **before** running any `drush` command
  - All cache is flushed during US deployments
  - Only country-specific during international deployments

Fixes #3211.

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -239,7 +239,15 @@ function build_helper {
   then
     # Flush only country-specific cache during international deployments.
     echo "Flush Redis cache on $SITE."
-    redis-cli --raw keys "$SITE:*" | xargs -n1 redis-cli del
+
+    # Solve Redis cache prefix.
+    case $DS_ENVIRONMENT in
+     "local") CACHE_PREFIX="ds" ;;
+     *) CACHE_PREFIX=$SITE ;;
+    esac
+
+    # Redis CLI connection settings must be preset.
+    redis-cli --raw keys "$CACHE_PREFIX:*" | xargs -n1 redis-cli del
   fi
 
   echo 'Adding role to admin user...'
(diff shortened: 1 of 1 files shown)


sergiitk:
CC @blisteringherb @aaronschachter @mshmsh5000 @barryclark

bin/ds
```diff
@@ -233,6 +233,17 @@ function build_helper {
 
   fi
 
+  # Before running drush it's required to flush Redis cache to avoid
+  # a race condition state between DB and cache.
+  if [[ $SITE == "default" ]]
+  then
+    # Flush everything when building US.
+    redis-cli flushall
```
mshmsh5000: Not sure if `redis-cli` alone will connect to the proper (remote) redis server on prod web heads.

sergiitk: Confirming international prod:

``` sh
dosomething@PROD.international:~$ redis-cli keys "*" | wc -l
29810
```

sergiitk: And you were right, US production cli client isn't connected to the server.
@blisteringherb and I have discussed possible ways out.

@mshmsh5000 What would you say if we skipped this step for US?
We are not sure if US prod is affected at all so it might make sense to just run the code only on intl servers.

CC @aaronschachter

sergiitk:
Tested a local international build made with `ds` script including this change. All worked well!

aaronschachter:
@sergii-tkachenko Is this good to merge?

sergiitk:
@aaronschachter not yet.

PR #3276, opened 2014-10-28
Fixes User Reset layout

aaronschachter:
Simplifies `theme_hook_suggestions` checks to just check for `page__user__%` and NOT existence of `page__user__edit` to use `page--full-width.tpl.php` on the Profile only.  

Bug was occurring because it was being applied to the User Reset page, which I had missed in the `$standard_pages` array.  This way, by default, it will use the normal `page.tpl.php` and we don't need to keep adding different User Pages into the `$standard_pages` variable.

Tested the User Reset page locally by turning off Message Broker and enabling default Drupal mail password reset.


sergiitk:
:+1:

PR #3278, opened 2014-10-29
Term intro image

### LINKED ISSUE #3266: need field_intro_image added to taxonomy collections
display should be the same as grouped campaigns

image

aaronschachter:
Closes #3266.

The Intro Image should only display if a Video is not set, like the Campaign Group template.


sergiitk:
:+1:

PR #3290, opened 2014-10-30
DS build: Revert only overridden features.

sergiitk:
This PR is to avoid possible circular dependencies when reverting features.
#### What's this PR do?
- Makes DS script revert only features that are clearly overridden instead of reverting everything


mshmsh5000:
Worth a try!

mshmsh5000:
:+1:

PR #3292, opened 2014-10-30
DS build: Add drush status.

sergiitk:
This PR makes `ds` build display drush status in order to provide additional debug information during international deployments.


mshmsh5000:
Looks good :+1:

PR #3294, opened 2014-10-30
DS Build: echo debug line with drush working directory

sergiitk:
This PR is to ensure drush commands are executed from the right directory during international deployments.


mshmsh5000:
Brilliant :+1:

PR #3295, opened 2014-10-31
Adding proximity boost to enable search phrases

### LINKED ISSUE #3209: improvements to search results
Here are a few examples of searches that don't return ideal results
- when searching for "apply scholarships", /scholarships should be my first result, not 20th
- when searching for "Nancy Lublin", top results should be /about/our-team
- when searching for "donate", /donate should be one of the top results
- when searching for "animal testing", bully text shows up and displays this content, even though the term "test" isn't anywhere on the campaign page

image

I think one thing we may want to consider is removing the boosts for campaigns and instead have all page types be weighted equally.

blisteringherb:
Removing campaign boosts and adding term proximity boosts so that content with the entire search phrase shows up closer to the top regardless of content type.

Fixes #3209 

@angaither @aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_search/dosomething_search.module
+++ b/lib/modules/dosomething/dosomething_search/dosomething_search.module
@@ -80,16 +80,9 @@ function dosomething_search_apachesolr_query_alter(DrupalSolrQueryInterface $que
     $params['q'] = '*:*';
   }
 
-  $boosts = array(
-    'bs_high_season' => '1.0',
-    'bs_field_staff_pick' => '5.0',
-    'bs_sponsored' => '4.0',
-  );
-  foreach ($boosts as $field => $boost) {
-    $params['bq'][] = "{$field}:true^$boost";
-  }
-
-  $params['bq'][] = "bundle:campaign^100.0";
+  // Add a term proximity boost so that multiple search terms
+  // are treated as one phrase.
+  $params['pf'] = 'content~2^1000';
 
   if ($query) {
     $query->addParams($params);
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_search/dosomething_search.module
```diff
@@ -80,16 +80,9 @@ function dosomething_search_apachesolr_query_alter(DrupalSolrQueryInterface $que
...
-  $boosts = array(
-    'bs_high_season' => '1.0',
-    'bs_field_staff_pick' => '5.0',
-    'bs_sponsored' => '4.0',
-  );
-  foreach ($boosts as $field => $boost) {
-    $params['bq'][] = "{$field}:true^$boost";
-  }
-
-  $params['bq'][] = "bundle:campaign^100.0";
```
aaronschachter, line 92: I'm a little clueless here, but just checking / curious: All this code is supposed to be removed too?

blisteringherb, line 92: I added it back in again for the /campaigns page, but it should be removed for regular search.

aaronschachter:
Seems ok.  I see the comments in within `dosomething_search_apachesolr_query_alter` that read as:

```
  // Add a 'q' param to apachesolr_views queries to enable the
  // elevate.xml to work properly.
```

Is this still the case?  I thought I remembered you mentioning that we don't use elevate.xml anymore?

blisteringherb:
@aaronschachter We're not using elevate at the moment, but that needs to be there so that the boosts work with views.

aaronschachter:
@blisteringherb Right on -- do you mind just updating the comments accordingly then? Then good to ship :+1:

PR #3298, opened 2014-10-31
Adds Title to Administrative Area options variable

### LINKED ISSUE #3248: Canada State label
Request from @mfurdyk -- For school modal, label should read as "Province" instead of "State" in the first part of the select school modal.

Not sure if the Addressfield module makes this change on the User Profile field.  This is a similar request to #3183... where the ideal solution would be a Canadian language to use to translate variables to.

Setting this to low priority for now.

aaronschachter:
We store Addressfield's State/Province options in a variable that `dosomething_user_set_address_administrative_area_options` sets.

This PR stores refactors this variable to be stored as a multi-dimensional array which includes the title to use, (e.g. "State", "Provice") to close #3248 for Canada TFJ


sergiitk:
:+1:

PR #3301, opened 2014-11-03
User reportbacks

### LINKED ISSUE #3293: Admin reportback create/edit form
Allow administrators ability to create a reportback entity, selecting the User Uid and the Campaign node nid that the reportback should be associated with.

Also expose these fields on the Reportback Edit form.

This will be used to add Reportbacks to the Closed state of Legacy Campaigns for which we'll be creating new beta campaign nodes.

aaronschachter:
- Adds admin only form to manually create a Reportback: fixes #3293
- Adds new view `user_reportbacks` which displays all reportbacks.  Can filter by Campaign Name and Cause.

screen shot 2014-11-03 at 11 32 36 am


sergiitk:
:+1:

PR #3302, opened 2014-11-03
Displays actual User data on profile

aaronschachter:
Fixes bug, uses correct variable to load User and display User Profile info.


sergiitk:
:+1:

PR #3305, opened 2014-11-03
Fixes race hook_update bug with message_broker_producer update

### LINKED ISSUE #3303: Message Broker / DS MBP errors
Per @DeeZone on 10/31:

Something funky going on with message_broker_producer_update_7002 not being built on staging when doing a deploy. As a result dosomething_mbp_update_7002 fails as the database table is missing.

deezone:
Fixes #3303 

`drush updb` produces a race condition between `dosomething_mbp_update_7002` and message_broker_producer_update_7002`. This fix ensures the`message_broker_producer_productions`database table is in place when running the`dosomehting_mbp` update.

### DIFF
--- a/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
+++ b/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
@@ -16,6 +16,11 @@ function dosomething_mbp_update_7001(&$sandbox) {
  */
 function dosomething_mbp_update_7002(&$sandbox) {
 
+  if (!db_table_exists('message_broker_producer_productions')) {
+    $message_broker_producer_productions_schema = drupal_get_schema_unprocessed('message_broker_producer', 'message_broker_producer_productions');
+    db_create_table('message_broker_producer_productions', $message_broker_producer_productions_schema);
+  }
+
   $productions = array();
 
   $productions[0]['machine_name'] = 'transactional_user_register';


aaronschachter:
Seems ok. Has the stage DB been manually altered so that `dosomething_mbp` schema is 7001?

deezone:
No need to alter the db, this will update in the right order with the db as is. I tested on my local by dropping the db (`drush sql-drop`) and doing a fresh `ds pull stage --db`.

Using `drupal_get_schema_unprocessed()` is a "no no" but it will get us through in "hacky style".

aaronschachter:
What I'm not crazy about is that we'll be testing this update from soup to nuts for the first time on **production**, since the staging DB has schema 7002 and has the `message_broker_producer_productions` table installed already.

The ideal test here should be to drop the `message_broker_producer_productions` table from staging DB, and set the `message_broker_producer` and `dosomething_mbp` schema versions to match production so we are actually replicating what will happen on prod.

Isn't accurate to say that your local test didn't execute the code in this PR since it would have already had the `message_broker_producer_productions` table upon pulling from stage?

deezone:
@aaronschachter I understand your concern. I checked the staging and production databases and both don't have the `message_broker_producer_productions` db table. My previous comment on the issue being about the table already existing was not true.

on the staging db:

```
dosomething@staging:/var/www/beta.dosomething.org/current/html$ mysql -u root
mysql> show tables;
...
| menu_router                                   |
| metatag                                       |
...
```

Note the missing `message_broker_producer_productions` table.

Running this update on staging and production will work. The key being that the`message_broker_producer_productions` table is created before `dosomething_mbp` tries to update the table with the production entries. That was the source of the update error and the `hook_update` failing to work.

aaronschachter:
ok, :shipit:

PR #3306, opened 2014-11-03
Fact Page intro variable

aaronschachter:
- Removes unused fields in the Fact Page content type.
- Adds admin form for a Fact Page Copy Intro variable
- Outputs Fact Page Copy Intro variable onto all Fact Page templates


sergiitk:
:+1:

PR #3309, opened 2014-11-04
Bug fix to resolve sending transactional requests to MB

### LINKED ISSUE #3303: Message Broker / DS MBP errors

deezone:
Fixes #3303

Fixes regression of `message_broker_producer_request()` not containing `$payload` values.


aaronschachter:
I'm confused. Do we also need to rollback the commit of `message_broker_producer` in the build file?

deezone:
No need to roll back, this resolves the key issue that causing transactions to fail. I discovered this regression last night after the mbp race issue was resolved.

aaronschachter:
Ok give it a go :+1:

PR #3313, opened 2014-11-04
Adding gallery field to solr index

blisteringherb:
Searching for Nancy Lublin or other staff members doesn't return any appropriate search results because the gallery fields need to be indexed.  Adding this field to the static content type will index the gallery fields in Solr.

@angaither 

Addresses #3209


aaronschachter:
:+1:

PR #3314, opened 2014-11-04
Restores peace and happiness to DS build

aaronschachter:
Applies patch in https://www.drupal.org/files/issues/services-2200269.patch.

Sort of bizarre for this to have just started happening this afternoon, but this fixes (I've ran `ds build` locally)


sergiitk:
:+1:

PR #3324, opened 2014-11-05
Drop local database before pulling stage DB

aaronschachter:
To help promote sanity when debugging "tables existing" errors.

Note: This will prompt you "Are you sure?".  I'm into leaving this warning in, to further notify a developer "Hey you're' about to lose any of your DB work" ... but, could be open to adding the `-y` option in.

```
vagrant@dev:/var/www/vagrant$ bin/ds pull stage --db
Dropping local DB
Do you really want to drop all tables in the database dosomething? (y/n): y
Pulling down db from staging
```

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -464,8 +464,10 @@ function pull_db {
     if [[ ${#alias[@]} == 1 ]]
     then
       cd $WEB_PATH
+
       echo "Dropping local DB"
-      drush sql-drop
+      drush -y sql-drop
+
       echo "Pulling down db from staging"
       drush -y sql-sync @ds.staging @self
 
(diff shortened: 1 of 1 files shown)


bin/ds
```diff
@@ -463,8 +463,10 @@ function pull_db {
   then
     if [[ ${#alias[@]} == 1 ]]
     then
-      echo "Pulling down db from staging"
       cd $WEB_PATH
+      echo "Dropping local DB"
+      drush sql-drop
```
sergiitk: It's better to change this to `drush -y sql-drop`.

sergiitk:
:+1:

PR #3325, opened 2014-11-05
Lint fix

DFurnes:
# Changes
- Updates comment style to fix linting error.
- Remove strict Ruby version requirement from Gemfile. We've been able to remove other Ruby dependencies and SCSS-Lint works fine on 1.9.3+.


tjrosario:
:+1:

PR #3327, opened 2014-11-05
Ensuring that disallowed types do not show up in search results

### LINKED ISSUE #3206: facts, action guides, campaign runs, and the homepage shouldn't show up in search results
we're currently having individual facts show up in search results. These should be removed from possible results (user gets 404'd when they try to access the pages).

Fact pages and campaigns that have facts on them however should remain.

image

The homepage is also showing up and shouldn't:

image

And action guides are showing up too but shouldn't be:

image

Neither should campaign runs:

image

or images:

image

basically the only thing I should be able to click-through to from the search results are:
- campaigns
- campaign collections
- fact pages
- static content pages
- taxonomy views (only if published)

@blisteringherb made some updates to this, let me know if you have any questions

blisteringherb:
- There is a bug in the features export for apachesolr search
  that doesn't respect the types that shouldn't be indexed. This
  will ensure that those types don't show up in search.

Fixes #3206 

@aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_search/dosomething_search.module
+++ b/lib/modules/dosomething/dosomething_search/dosomething_search.module
@@ -113,13 +113,13 @@ function dosomething_search_apachesolr_query_alter(DrupalSolrQueryInterface $que
 function dosomething_search_filter_types($type) {
   $defaults = array(
     'campaign',
-    'campaign_collection',
+    'campaign_group',
     'fact_page',
     'static_content'
   );
-  $allowed_types = variable_get('allowed_search_types', $defaults);
+  $allowed_types = variable_get('dosomething_search_allowed_search_types', $defaults);
   if (!in_array($type, $allowed_types)) {
-    return true;
+    return TRUE;
   }
 }
 
--- a/lib/modules/dosomething/dosomething_search/dosomething_search.module
+++ b/lib/modules/dosomething/dosomething_search/dosomething_search.module
@@ -79,7 +79,9 @@ function dosomething_search_apachesolr_query_alter(DrupalSolrQueryInterface $que
   // other apachesolr views.
   $query_params = $query->getParams();
   if (!isset($query_params['q'])) {
-    $params['q'] = '*:*';
+    // Set both vars so that disallowed types never
+    // show up in apachesolr views.
+    $query_params['q'] = $params['q'] = '*:*';
     $boosts = array(
       'bs_high_season' => '1.0',
       'bs_field_staff_pick' => '5.0',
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_search/dosomething_search.module
```diff
@@ -101,6 +110,19 @@ function dosomething_search_apachesolr_query_alter(DrupalSolrQueryInterface $que
...
+function dosomething_search_filter_types($type) {
+  $defaults = array(
+    'campaign',
+    'campaign_collection',
+    'fact_page',
+    'static_content'
+  );
+  $allowed_types = variable_get('allowed_search_types', $defaults);
+  if (!in_array($type, $allowed_types)) {
+    return true;
```
aaronschachter: picky picky `TRUE`

lib/modules/dosomething/dosomething_search/dosomething_search.module
```diff
@@ -101,6 +110,19 @@ function dosomething_search_apachesolr_query_alter(DrupalSolrQueryInterface $que
   }
 }
 
+function dosomething_search_filter_types($type) {
+  $defaults = array(
+    'campaign',
+    'campaign_collection',
```
aaronschachter: It's `campaign_group`

lib/modules/dosomething/dosomething_search/dosomething_search.module
```diff
@@ -101,6 +110,19 @@ function dosomething_search_apachesolr_query_alter(DrupalSolrQueryInterface $que
...
+    'fact_page',
+    'static_content'
+  );
+  $allowed_types = variable_get('allowed_search_types', $defaults);
```
aaronschachter: Is `allowed_search_types` a variable set by Apachesolr module?

lib/modules/dosomething/dosomething_search/dosomething_search.module
```diff
@@ -92,6 +94,13 @@ function dosomething_search_apachesolr_query_alter(DrupalSolrQueryInterface $que
     $params['bq'][] = "bundle:campaign^100.0";
   }
 
+  // There is a bug in the apachesolr allowed types export that
+  // always overrides the types that should be indexed in search. This
+  // code is necessary to make sure that only the allowed types are included
+  // in search results.
+  $disallowed_types = implode(' ', array_filter(array_keys(node_type_get_types()), 'dosomething_search_filter_types'));
```
aaronschachter, line 101: Is this a reason why `dosomething_search` is constantly overridden?  If so can we remove the types from the feature, since we'll just be handling it this way instead of features?

blisteringherb, line 101: They're constantly overridden because time-dependent items are included in the export.  https://www.drupal.org/node/1984502.  I attempted to use a few of these fixes, but unfortunately it's still not reliably using what's in features.

PR #3331, opened 2014-11-05
404page

itsjoekent:
Git history was broken, screwed up. This is the actual one now!

(Old for reference https://github.com/DoSomething/dosomething/pull/3310) All of those dicussion points should be resolved now.


lib/modules/dosomething/dosomething_notfound/dosomething_notfound.admin.inc
```diff
@@ -0,0 +1,11 @@
+<?php
+
+function dosomething_notfound_admin_config_form($form, &$form_state){
+
+  $form['dosomething_notfound_disable_results'] = array(
+    '#type' => 'checkbox',
+    '#title' => t('Disable relevant results'),
+    );
+  //variable_get
```
aaronschachter: Might as well get rid of this "note to self" comment.

lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
```diff
@@ -0,0 +1,80 @@
...
+  ->execute()
+  ->fetchCol(0);
+  
+  if ($result) {
+   $nid = $result[0];
+   $node = node_load($nid);
+	} // No else is needed as a default node is always created when the module installs
+
+	// Return & Load node
+	return node_view(node_load($nid), 'full');
```
aaronschachter: You can just use return `node_view($node, 'full');` here, since you set it on line 49 ^^

aaronschachter:
:+1:  SHIP IT

PR #3332, opened 2014-11-05
Check yourself before you wreck the drush updb

aaronschachter:
Per https://github.com/DoSomething/dosomething/issues/3320#issuecomment-61872005


sergiitk:
:+1:

Issue #3344, opened 2014-11-06
turn off address validation on canadian site


mikefantini:
![image](https://cloud.githubusercontent.com/assets/6330971/4940529/26e56058-65db-11e4-869d-76ae044f1b1f.png)

@DFurnes @sergii-tkachenko

PR #3345, opened 2014-11-06
Provides an option to bypass JS validation of user's postcode.

### LINKED ISSUE #3329: remove data-validate='zip' attribute on address collection for international sites
image

sergiitk:
#### What's this PR do?
- Provides a theme setting to bypass JS validation of user's postcode
- Open [Paraneue Settings](http://dev.dosomething.org:8888/admin/appearance/settings/paraneue_dosomething)
- Turn off `Validate Post Code`
  image
- Go to [Yeah, Science!](http://dev.dosomething.org:8888/campaigns/yeah-science)
- Click 'Get this banner, yo.'
- Fill in `M5H 3H1` to `Postal code` field. The field must not be highlighted in red after that
  image

Closes #3329

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -307,7 +307,9 @@ function dosomething_user_add_validation_attributes_to_address_fields($form, &$f
   $address['address']['street_block']['premise']['#attributes']['data-validate'] = 'address2';
   $address['address']['locality_block']['locality']['#attributes']['data-validate'] = 'city';
   $address['address']['locality_block']['administrative_area']['#attributes']['data-validate'] = 'state';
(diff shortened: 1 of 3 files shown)


aaronschachter:
I'm kind of leaning towards the configuration setting living in the DS User admin config form vs the Paraneue settings.  If we ever have two themes (which we could).. we'd have to duplicate this logic.  What do you think @sergii-tkachenko ?

Especially since we also need a similar setting for shutting off validation completely (#3344) which I was envisioning living within the DS User admin config form.

sergiitk:
My reasoning was that actual JS validation [is the part of paraneue_dosomething theme](https://github.com/DoSomething/dosomething/blob/dev/lib/themes/dosomething/paraneue_dosomething/js/validation/address.js#L68). So I think it makes sense to setup JS validations in theme settings.

aaronschachter:
OK :+1:

Issue #3340, opened 2014-11-06
Setup a cron job to update users address country


sergiitk:
We need to create a cron job that runs  `drush update-users-country` every minute on each affiliate. It works same as `drush cancel-us-users`, but it will fix users countries. Details in #3338 and #3242.

sergiitk:
CC @aaronschachter @mshmsh5000 @mikefantini

sergiitk:
It'd also be great to forward stdout and stderr to log file to follow progress.

sergiitk:
Added to crontab on intl production:

```
0 0 * * * cd /var/www/international.dosomething.org/current && bin/drush-intl update-users-country &>> /home/dosomething/update-users-country.log
```

PR #3338, opened 2014-11-06
Add drush task to fix users address country.

sergiitk:
Fixes #3242.

PR #3346, opened 2014-11-06
Fixing user country with drush: update only users register after intl release.

sergiitk:
Makes users country update #3338 independent from users cleanup #3095.
Ref #3340.


aaronschachter:
:+1:

PR #3350, opened 2014-11-06
Fixed css to apply style to the header instead of div

itsjoekent:
Old:
screen shot 2014-11-06 at 2 42 45 pm
New:
screen shot 2014-11-06 at 2 42 52 pm


DFurnes:
:shipit:

Issue #3349, opened 2014-11-06
Create bin script to run drush commands at every affiliate


sergiitk:
Reference #3339, #3340.

CC @aaronschachter

Issue #3322, opened 2014-11-05
get Zendesk set up on Canada


mikefantini:
**UPDATE:**

The approach is to set up Zendesk for Canada to avoid all more affiliate-only logic.

Once #3351 is merged and `dosomething_zendesk` is uninstalled on all affiliate sites (per #3349), the steps will be to:
- [x] manually enable DoSomething Zendesk module on Canada
- [x] set up its auth variables
- [x] set its dosomething_zendesk_group_id to the Canada Zendesk group.

**ORIGINAL ISSUE:**

Canada doesn't have zendesk set up so this is a dead link.

Ideally language would be "Still have issues? Email us at teensforjeanscanada@dosomething.org with your school name and provence."

![image](https://cloud.githubusercontent.com/assets/6330971/4912309/7a4f501e-649b-11e4-8556-621599434948.png)

mikefantini:
one possible route (although no idea how hard this is) would be to flag the "Still having issues..." bit to not display for Canada and then we can just include the text above in the admin copy config variables

mikefantini:
from standup the plan is to investigate turning on zendesk for Canada and have all questions from Canadian users go into the one Canadian group

aaronschachter:
Up and running on Canada prod.

PR #3351, opened 2014-11-06
Enables affiliate access to Zendesk

aaronschachter:
Gets things set up for enabling Canada Zendesk (#3322)
- Allows for a global `dosomething_zendesk_group_id` variable to be set.  If this is set, route all Zendesk tickets to this group.  We'll set this variable for Canada.
- Removes the check for affiliate in rendering the Zendesk form.  

This means we'll need to disable / uninstall the Dosomething Zendesk module on every affiliate before we push this code (blocked by #3349).

Once all affiliate sites have Zendesk uninstalled and this code is pushed, I will manually enable DoSomething Zendesk module on Canada, set up its auth variables, and set its `dosomething_zendesk_group_id` to the Canada Zendesk group.


sergiitk:
:+1:

PR #3352, opened 2014-11-06
Provides bin script that runs drush on each affiliate.

### LINKED ISSUE #3349: Create bin script to run drush commands at every affiliate

sergiitk:
#### Usage
- `vagrant ssh`
- `cd /var/www/vagrant/`
- `bin/drush-intl status --fields=uri`
  image

Closes #3349.

### DIFF
--- a/bin/drush-intl
+++ b/bin/drush-intl
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+# ==============================================================================
+# Globals
+
+# The base project directory
+BASE_PATH=`pwd -P`
+
+# The webroot directory name for the dev environment
+WEB_DIR='html'
+
+# Full webroot path
+WEB_PATH=$BASE_PATH/$WEB_DIR
+
+
+# ==============================================================================
+# Functions
+
+for SITE in botswana canada congo ghana kenya indonesia nigeria training uk
+do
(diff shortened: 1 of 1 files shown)


bin/drush-intl
```diff
@@ -0,0 +1,25 @@
...
+WEB_DIR='html'
+
+# Full webroot path
+WEB_PATH=$BASE_PATH/$WEB_DIR
+
+
+# ==============================================================================
+# Functions
+
+for SITE in botswana canada congo ghana kenya indonesia nigeria training uk
```
aaronschachter, line 19: Not necessarily for this PR, but we should really have this list as a variable somewhere.  Lots of gotchas where we miss an affiliate.

aaronschachter:
:+1: Much value added.

Can we give it a shot by disabling and uninstalling `dosomething_zendesk` ? :smile:

bin/drush-intl
sergiitk, line 19: Yes, def. We should discuss and find a common way to provide a source with this list. Will create ticket for it.

sergiitk:
Yes, @aaronschachter, let's do it!

PR #3355, opened 2014-11-06
Added video field to image feature

### LINKED ISSUE #3228: Add "Image Video" field to Image content type
For the 404 page, we want to display a Video on desktop but an Image on mobile.  We are also experimenting with Video Cover images.  It seems useful to add a new field to the Image content type to render an Image node as a video file.
- Add a "Image Video" File field to the Image content type, which allows all video formats.  The resulting machine name should be `field_image_video`.
- Export the new field via Features export  to the DoSomething Image module. You'll need to export the `field_base` and `field_instance` exportables in `drush fe dosomething_image`.

itsjoekent:
Fixes https://github.com/DoSomething/dosomething/issues/3228

(Ignore commit name typo)


lib/modules/dosomething/dosomething_image/dosomething_image.features.inc
```diff
@@ -447,6 +447,5 @@ function dosomething_image_node_info() {
       'help' => '',
     ),
   );
-  drupal_alter('node_info', $items);
```
aaronschachter, line 450: Services 7.x-2.2 started adding all of these in there.  It looks like you're running an older build - can you confirm in your admin/modules list that it shows Services 7.x-2.0?

It probably is.  This will get added back in upon any features udpates with a newer build, so please remove this file.

Also you can update your build by doing a `ds pull stage --db` (or just running `ds build` from vagrant)

aaronschachter:
Looks good except for `drupal_alter`.  Also, pls reference the Github issue # in your PR descriptino so we can track work against various issues.

itsjoekent:
@aaronschachter Ill check now for the version, also updated the PR with the issue

lib/modules/dosomething/dosomething_image/dosomething_image.features.inc
itsjoekent, line 450: I have version 7.x-3.10

aaronschachter:
SHIP IT :+1:

PR #3357, opened 2014-11-06
Fixes lock being set on premature returns.

sergiitk:
Fix a bug in lock logic introduced in #3095 and #3338.


aaronschachter:
:+1:

PR #3359, opened 2014-11-06
Fix gallery index

blisteringherb:
The snippets are displaying the labels for gallery images instead of just the image titles, this adds a view mode, "dosomething_search_index", for the field_collection entity and the features export settings for the gallery and field_gallery_item field collections.

@aaronschachter


aaronschachter:
Nice find! :+1: value added

PR #3360, opened 2014-11-07
Fires custom event on action page. Fixes #3308.

### LINKED ISSUE #3308: Track Action Page views vs Pitch Page views
**ISSUE:** We need to track unique "Action Page" views in Google Analytics.  Current limitation is that we don't have a way to distinguish page views between the Pitch Page and Action Page, since both have the same URL.

**UPDATED:** Fire a GA tracking event for every action page view, in order to specifically track action page views.  We could fire one for Pitch Page views too, for the hay of it.

**ORIGINAL thought:**

e.g. Use a query string with redirects: `/campaigns/teens-jeans?action=signup`
- When a user signs up, direct them to the campaign node with `?action=signup` appended to query string
- If a user is viewing a link with `?action=signup` and is actually not signed up (like if someone shares their `?signedup` link), redirect to URL without the query string

DFurnes:
# Changes
- Adds a [custom event](https://developers.google.com/analytics/devguides/collection/gajs/eventTrackerGuide) to the action page. See #3308 for context.
# Question
- Is there a better place for this? This method seemed mostly concerned with preprocessing variables for theming, but it seemed to be the most appropriate place I could find.

For review: @aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -282,6 +282,11 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
     $vars['reportback_link']['label'] = t("Update Submission");
   }
   $vars['reportback_form'] = $vars['content']['reportback_form'];
+
+  // Add analytics custom event on action page.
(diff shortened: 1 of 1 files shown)


aaronschachter:
I would have put it in this function too, as its the easiest place to target action page right now.  Ship it :+1: 

If we ever get around to a refactor sprint, it'd be worth refactoring this `dosomething_campaign.theme.inc` code up into specific paraneue functions for presentation and the dosomething_campaign_load properties for API properties type data.

PR #3361, opened 2014-11-07
Address validation switch

### LINKED ISSUE #3344: turn off address validation on canadian site

aaronschachter:
Fixes #3344.  Only validate addresses if `dosomething_user_validate_address` variable is set to TRUE.

`dosomething_user_update_7006` will set this to TRUE for US but not set it for all other sites.

Also adds a custom access callback to the UPS API settings page, which checks to see if the variable is TRUE.  Also removes the "administer UPS configuration" permission and just checks for "administer modules' instead, in the interest of simplifying.

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -199,6 +199,9 @@ function dosomething_user_is_staff($user = NULL) {
 function dosomething_user_form_alter(&$form, $form_state, $form_id) {
   global $user;
   $account = $user;
+  // Check if we should be validating addresses.
+  $is_validate_address_set = variable_get('dosomething_user_validate_address');
+
   switch ($form_id) {
     case 'user_login_block':
     case 'user_login':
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -123,6 +118,19 @@ function dosomething_user_admin_paths() {
...
+ * Access callback for UPS API configuration page.
+ *
+ * @see dosomething_user_menu()
+ */
+function dosomething_user_ups_api_config_access() {
+  // If administrator:
+  if (user_access('administer modules')) {
+    // Display page if we are validating addresses.
+    return variable_get('dosomething_user_validate_address');
+  }
```
sergiitk, line 130: It's better to add the return FALSE at the end to be more explicit.

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -123,6 +118,19 @@ function dosomething_user_admin_paths() {
...
 /**
+ * Access callback for UPS API configuration page.
+ *
+ * @see dosomething_user_menu()
+ */
+function dosomething_user_ups_api_config_access() {
+  // If administrator:
+  if (user_access('administer modules')) {
+    // Display page if we are validating addresses.
+    return variable_get('dosomething_user_validate_address');
```
sergiitk: And here it'd better to `return variable_get('dosomething_user_validate_address', FASLE);`

sergiitk:
Great job, just minor comments and it'd be good to go :+1:

PR #3362, opened 2014-11-07
Use module names as menu titles

aaronschachter:
Cleans up the DS config menu to make it easier for devs to know what variables are used where. 

screen shot 2014-11-07 at 9 46 02 am

Third Party variables is the odd one here, mainly because it shouldn't live on the DoSomething Signup admin page, since the `communications team` needs access to set Third Party variables but should not have access to configure the DoSomething Signup variables.


sergiitk:
:+1:

mshmsh5000:
Too late to the party, but: This is nice.

PR #3366, opened 2014-11-07
Allow Signup Data Form reset

aaronschachter:
Adds a Reset Signup Data Form to the Signup view (visible only to staff). 

Allows staff to manually reset the signup data form, allowing a user to re-submit the signup data form if they want to change their answer.


sergiitk:
:+1:

Issue #3321, opened 2014-11-05
Update "still having issues?" language on school finder


mikefantini:
Still having issues? Contact us [link] with your school name and state/province.

![image](https://cloud.githubusercontent.com/assets/6330971/4912296/573263be-649b-11e4-80a8-df7f768a67c9.png)

mikefantini:
since we're going to try adding zendesk to Canada's site per #3322 I updated the copy we want above to be state/province rather than just state

PR #3367, opened 2014-11-07
School Finder State/Province label

aaronschachter:
Updates help text per #3321.  Will display as "State" in US, but "Province" in Canada.

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
@@ -374,12 +374,19 @@ function dosomething_signup_user_signup_data_form($form, &$form_state, $signup) 
     $form['school_finder']['help_link'] = array(
       // Used for front-end autocomplete.
       '#type' => 'markup',
-      '#markup' => '<a href="#" class="secondary js-schoolfinder-help">Need help locating your school?</a></p>',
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -374,12 +374,19 @@ function dosomething_signup_user_signup_data_form($form, &$form_state, $signup)
     $form['school_finder']['help_link'] = array(
       // Used for front-end autocomplete.
       '#type' => 'markup',
```
sergiitk, line 376: I know it's not documented, but there's should be special form type called [link](https://api.drupal.org/comment/16104#comment-16104). Can you check if this is still working?

aaronschachter, line 376: I've tried it.. but when I click on "Need help locating your school?"  the actual link `href` is loaded, e.g. I am redirected to http://dev.dosomething.org:8888/%23.

sergiitk, line 376: Ok then, it's minor anyways.

aaronschachter, line 376: Right on, was worth a shot.  Ok to ship?

sergiitk:
:+1:

PR #3368, opened 2014-11-07
Grant member support access to Zendesk config

aaronschachter:
Will allow member support team to add Zendesk group id's for `campaign_group` nodes, like Grandparents Gone Wired.


sergiitk:
:+1:

PR #3371, opened 2014-11-10
Gallery template refactor / User Profile triads

aaronschachter:
Per convo in Dries hipchat room: Uses one single `gallery.tpl.php` and applies relevant `duo` and `triad` classes.

Updates User Profile to use `triad` class for the Doing section.


sergiitk:
:+1:

PR #3390, opened 2014-11-11
Fix Signup page view permissions

aaronschachter:
Needed for editors to have access to Signup Reset Form introduced in #3366


sergiitk:
:+1:

PR #3391, opened 2014-11-11
Added Video hero

### LINKED ISSUE #3333: Render Image video in header if Image video field exists
~~Blocked by #3228~~

Preprocessing should be done within`dosomething_helpers_preprocess_hero_images` and `dosomething_helpers_add_inline_css` so it can be used by the various templates which call that function.

itsjoekent:
Displays a video inside a hero if you are on desktop and the file is set in the image node (Video field made in previous PR). Also falls back to default landscape photo if the video is not set. 

Fixes #3333 
## 

@aaronschachter can you sanity check what I did?

### DIFF
--- a/lib/modules/dosomething/dosomething_image/dosomething_image.module
+++ b/lib/modules/dosomething/dosomething_image/dosomething_image.module
@@ -100,7 +100,7 @@ function dosomething_image_get_themed_image_url($nid, $ratio, $style) {
   // If the image doesn't have the appropriate crop don't return a url.
   if (isset($base_path)) {
     // For videos
-    if(!isset($style)){
+    if (!isset($style)) {
       return file_create_url($base_path);
     }
     return image_style_url($style, $base_path);
--- a/lib/modules/dosomething/dosomething_image/dosomething_image.module
+++ b/lib/modules/dosomething/dosomething_image/dosomething_image.module
@@ -101,7 +101,7 @@ function dosomething_image_get_themed_image_url($nid, $ratio, $style) {
   if (isset($base_path)) {
     // For videos
     if(!isset($style)){
-      return $base_path;
+      return file_create_url($base_path);
     }
     return image_style_url($style, $base_path);
   }
--- a/lib/modules/dosomething/dosomething_image/dosomething_image.module
+++ b/lib/modules/dosomething/dosomething_image/dosomething_image.module
@@ -78,6 +78,8 @@ function dosomething_image_get_image_url($nid, $ratio) {
       case 'thumb':
         $path = $node->field_image_thumbnail[LANGUAGE_NONE][0]['uri'];
         break;
+      case 'video':
+        $path = $node->field_image_video[LANGUAGE_NONE][0]['uri'];
     }
     return $path;
 }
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_image/dosomething_image.module
```diff
@@ -97,6 +99,10 @@ function dosomething_image_get_themed_image_url($nid, $ratio, $style) {
 
   // If the image doesn't have the appropriate crop don't return a url.
   if (isset($base_path)) {
+    // For videos
+    if(!isset($style)){
```
aaronschachter: Spacing: https://www.drupal.org/coding-standards#controlstruct

aaronschachter: Also, would it make sense to make sure to check for `$ratio == 'video'` here, in case someone doesn't send the `$style` on an image?

lib/themes/dosomething/paraneue_dosomething/templates/notfound/node--notfound.tpl.php
```diff
@@ -16,8 +16,11 @@
   <header role="banner" class="-centered notfound">
     <div class="wrapper">
       <h1 class="__title"><?php print $title ?></h1>    
-      <?php print $subtitle ?> <!-- TODO, this should get the same properties of __subtitle -->
+      <?php print $subtitle ?>
     </div>
+    <?php if(isset($video)): ?>
```
aaronschachter: Spacing again

lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
```diff
@@ -68,6 +68,9 @@ function dosomething_notfound_preprocess_node(&$vars)  {
   dosomething_helpers_preprocess_hero_images($vars);
   dosomething_helpers_add_inline_css($vars);
 
+  // Add hero video
+  $vars['video'] = $vars['hero_image']['video'];
```
aaronschachter: `$vars['hero_image']['video']` might not be set right?  If a notfound node doesn't have a video, this will produce an error.  Please check the error logs (http://dev.dosomething.org:8888/admin/reports/dblog) as you visit various notfound nodes with different content (with video, without) to make sure you aren't adding errors in there.

lib/modules/dosomething/dosomething_image/dosomething_image.module
itsjoekent: Crap, that was supposed to be ratio originally - nice catch

lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
itsjoekent: Ah it is producing an error, I'll surround it with an if isset

lib/modules/dosomething/dosomething_image/dosomething_image.module
itsjoekent: just checked, ratio related code is done in the function above 'dosomething_image_get_image_url', and the ratio for video is just 'video'

I check for style specifically because I'm not resizing the video and it would be confusing to provide that parameter

aaronschachter:
:+1:

Issue #3328, opened 2014-11-05
US states are shown on users profile page on Canada


sergiitk:
On Canada [user #1732915](https://canada.dosomething.org/user/1732915/edit) has `CA` (Canada) country in address field:

``` sql
mysql> SELECT field_address_country FROM field_data_field_address WHERE entity_id=1732915;
+-----------------------+
| field_address_country |
+-----------------------+
| CA                    |
+-----------------------+
1 row in set (0.00 sec)
```

However, the address field still shows US states:
![image](https://cloud.githubusercontent.com/assets/672669/4922541/dd8f0f1c-6512-11e4-8c28-536b00903aba.png)

Related issue: #3307
CC @aaronschachter

sergiitk:
Additional: when the form is saved, country field gets reset back to `US`. 
For example, user [user #1732917](https://canada.dosomething.org/user/1732917/edit):

``` sql
> SELECT field_address_country FROM field_data_field_address WHERE entity_id=1732917;
+-----------------------+
| field_address_country |
+-----------------------+
| US                    |
+-----------------------+
1 row in set (0.00 sec)
```

aaronschachter:
This one is nasty. I suspect it's because the `field_address` feature is configured to US:

https://github.com/DoSomething/dosomething/blob/v0.3.96/lib/modules/dosomething/dosomething_user/dosomething_user.features.field_instance.inc#L24

and 

https://github.com/DoSomething/dosomething/blob/v0.3.96/lib/modules/dosomething/dosomething_user/dosomething_user.features.field_instance.inc#L55

This would require altering the feature such that any country be saved, with the disadvantage of needing to add even more `hook_form_alter` fu**ery to make work internationally, saving the site's country code programatically instead of letting Drupal User module do its thing.

Unless we had users just select what country they live in.

sergiitk:
@aaronschachter now I'm experimenting with custom address plugin. Works good so far. I'll fix country tests as described #3393 to make to make testing easier.

PR #3395, opened 2014-11-11
Changes the way country is tested in UK and Canada.

### LINKED ISSUE #3393: Fix UK and Canada tests due to a change in profile Page
Before #3261 country has been shown at user's profile page. Now it's hidden.
The other way to check country is to look into address wrapper on profile edit page:

``` html
<div class="addressfield-container-inline locality-block country-US">
```

However, now this functionality is blocked due to the bug described in #3328.

sergiitk:
#### What's this PR do?
- Changes the way country is tested in UK and Canada due to change in profile introduced in #3261
- Closes #3393 
- This change can be considered as specs for fixing #3328

### DIFF
--- a/tests/integration/canada/sso.coffee
+++ b/tests/integration/canada/sso.coffee
@@ -121,13 +121,6 @@ login = (test) ->
 
 # Performs 3 tests to check user's profile.
 user_profile = (test) ->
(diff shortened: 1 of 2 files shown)


aaronschachter:
:+1: smart

Issue #3312, opened 2014-11-04
Donate Page: New UI Requirements


tjrosario:
**Updated description** - by @aaronschachter 11/12

No new hero pattern, just basic form modal.  I've added checklist for all items needed below.

**Original issue**

Overview:

We will need add the ability to support a new Hero pattern that includes:
- Input field for donate amount
- CTA that links to opening a Stripe payment form modal

Other page requirements:
- [x] Include a Stripe payment form modal
- [x] Add Stripe API script to `<head>`.  This should be placed before our `app.js`  
  `<script type="text/javascript" src="https://js.stripe.com/v2/"></script>`
-  [x] Config file to store API keys (public and private).  The private key should not be exposed in the repository.
  - [x] Private API Key
  - [x] Public API Key - #3413 
- [x] We will need to leverage the Stripe PHP library since we cannot expose the private key on the client.  See: https://stripe.com/docs/libraries
- [x] Create a Stripe customer based on the First Name, Last Name (and email?)
- [x] Send a Stripe email receipt 

Wireframes and requirements can be found on the product wiki:
https://sites.google.com/a/dosomething.org/dosomethinguxwiki/home/product-archive/donate-page-powerups

PR #3397, opened 2014-11-11
DoSomething Payment module + Stripe integration

aaronschachter:
Adds a new DoSomething Payment module which will provide a new custom block for Donation Form, which uses Stripe.

Also adds the Stripe PHP Library for said Donation Form into the build / libraries.

Doesn't actually do anything yet though, but sets up place to add code for #3312

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -376,6 +376,10 @@ function composer {
     echo -e "\e[4mInstalling composer dependencies for Message Broker PHPLIB...\e[0m"
     eval "/usr/local/bin/composer install"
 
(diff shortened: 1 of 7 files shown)


sergiitk:
The install file is empty. You can add uninstall function there and remove dosomething_payment_\* vars.

aaronschachter:
Yeah.. I totally meant to do that. Good catch!

aaronschachter:
@sergii-tkachenko updated!

sergiitk:
:+1: Great!

PR #3400, opened 2014-11-12
Added specific path option to the 404 page

### LINKED ISSUE #3317: Path-specific 404 page suggestions
As we kill the legacy app, it will be important to have a few bits of preset copy to provide context for the user depending on which section of the legacy app they were coming from.

For example, if a user is coming from /projects/123, we'll want to provide them with copy that is projects specific.

**Updated description**

Since we only need suggestions for 2 URL patterns, we will just add 2 copy variables for each URL pattern into the admin form. e.g `dosomething_notfound_copy_projects` and `dosomething_notfound_copy_grants`

Each copy variable should be outputted as Markdown , using `check_markup($copy, 'markdown')`

**Original description:**
Ideal scenario is that we have a solid admin interface to configure what URL pattern maps to what copy.

I could see the following fields making sense for this:
- from https://legacy.dosomething.org/
- copy

Similar to here:
image

the trick is that we would want this to accept /\* so that all /project/\* got the same copy. Also I guess with this setup we would need to make one for /projects/\* and one for /project/*

Also if someone is hitting a preset path we'll want to also give them our recommended campaigns

Mocks: 
image

itsjoekent:
/projects/\* and /grants/\* are given different results containing a unique copy (editable in /admin/config/dosomething/dosomething_notfound) accompanied by campaign recommendations.

Fixes #3317


lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
```diff
@@ -81,11 +82,23 @@ function dosomething_notfound_preprocess_node(&$vars)  {
...
-  // @TODO: Check if its path specific (Intentionally hardcoded), get campaign suggestions, return & set suggestions var to true if it is 
+  $raw_url = request_path();
+
+  if (strpos($raw_url, "grants") === 0) {
+    $vars['suggestions'] = TRUE;
+    $vars['results_info'] = variable_get('dosomething_notfound_copy_grants');
+    $vars['raw_results'] = dosomething_notfound_get_recommended_campaigns();
+    return;
+  }
+  elseif (strpos($raw_url, "projects") === 0) {
```
aaronschachter: Should just be /project*: https://github.com/DoSomething/dosomething/issues/3317#issuecomment-62595357

lib/modules/dosomething/dosomething_notfound/dosomething_notfound.admin.inc
```diff
@@ -1,10 +1,20 @@
...
-
+function dosomething_notfound_admin_config_form($form, &$form_state) {
+ 
   $form['dosomething_notfound_disable_results'] = array(
     '#type' => 'checkbox',
     '#title' => t('Disable relevant results'),
     );
+  $form['dosomething_notfound_copy_projects'] = array(
+    '#type' => 'textarea',
+    '#title' => 'Projects copy',
```
aaronschachter, line 11: Let's make each of these copy variables required.  https://api.drupal.org/api/drupal/developer!topics!forms_api_reference.html/7#textfield

lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
```diff
@@ -107,3 +120,16 @@ function dosomething_notfound_preprocess_node(&$vars)  {
   $vars['results'] = $search;
 
 }
+
+function dosomething_notfound_get_recommended_campaigns() {
+  // Get 4 recommended campaign nids for term $tid and current user.
+  $rec_nids = dosomething_campaign_get_recommended_campaign_nids();
```
aaronschachter: `dosomething_campaign_get_recommended_campaign_nids` takes a $limit parameter. If you use `dosomething_campaign_get_recommended_campaign_nids(NULL, NULL, 3)`, you won't need the `array_slice` on L134

aaronschachter:
:+1: Value added has been added Ship it

itsjoekent:
added value complete

PR #3401, opened 2014-11-12
Donate Form block

aaronschachter:
Refs #3312
- Adds `dosomething_payment_form` which submits a Payment to Stripe
- Adds a Donate Form block which renders the Payment Form with Donate variables.
- Adds a template file for the Donate Form for Paraneue magic
- Creates a `block` subdirectory in the `templates/system` directory to house said block, and our existing `block.tpl.php` override
- Adds a testing function to make this easier to test.

TODO:
- Create a Stripe customer using the form `first_name` and `last_name` fields.


lib/modules/dosomething/dosomething_payment/dosomething_payment.module
```diff
@@ -26,13 +26,191 @@ function dosomething_payment_menu() {
...
+      '#type' => 'hidden',
+      '#default_value' => $confirm_msg,
+      '#access' => FALSE,
+    );
+  }
+
+  // Check if we are in secret testing mode.
+  if ($_GET['testing']) {
+    $email = variable_get('site_mail', '');
+    $test = dosomething_payment_get_stripe_test_values($email, 71100);
```
sergiitk, line 133: 71100?

lib/modules/dosomething/dosomething_payment/dosomething_payment.module
```diff
@@ -26,13 +26,191 @@ function dosomething_payment_menu() {
...
+  }
+  catch (Exception $e) {
+    form_set_error(NULL, t("Sorry, there was an error processing your request."));
+    watchdog('dosomething_payment', $e, NULL, WATCHDOG_ERROR);
+  }
+}
+
+/**
+ * Returns array of values for Stripe client testing.
+ *
```
sergiitk, line 204: I think link to api with testing description would've been nice here, if available.

lib/modules/dosomething/dosomething_payment/dosomething_payment.module
aaronschachter, line 133: Yeah.. this is just a random number I'm using for testing. $711.00.  (My birthday is 7/11)

lib/themes/dosomething/paraneue_dosomething/templates/system/block/block--dosomething_payment--donate_form.tpl.php
```diff
@@ -0,0 +1,18 @@
+<?php
+/**
+ * Available variables:
+ * - $content: The rendered Payment Form.
+ */
```
sergiitk, line 5: Add `- $block->subject: Block title.`

lib/modules/dosomething/dosomething_payment/dosomething_payment.module
aaronschachter, line 204: I'm happy to add that in on the next PR, as I'll need to make more API calls to create a Stripe customer for the given First/Last name fields.

lib/modules/dosomething/dosomething_payment/dosomething_payment.module
sergiitk, line 133: Could you please add a comment explaining that 71100 is just a random number? I know that some payment processors use amount value in order to generate return code, for example 1000 — success, 2000 — declined by your bank, etc.

sergiitk:
:+1:  Minor comments, will be fixed in next PR as discussed with @aaronschachter.

lib/modules/dosomething/dosomething_payment/dosomething_payment.module
aaronschachter, line 133: Sure, that definitely makes sense.  I'm thinking maybe it makes sense to set the Payment value in the query string instead of hardcoding a number anyway...so you can test this by going to `node/123?testing=71100` where $711 will be the `amount` value.

aaronschachter:
Thanks @sergii-tkachenko !! Yup, I will make these changes in next round.  This helps us get closer to prototyping with this block.

PR #3405, opened 2014-11-13
Search appearance

sbsmith86:
This builds on the work @blisteringherb did on getting an image add to each search result and updating the appearance for the main search page. I made some style tweaks based on the mocks and refactored it a little to use the new neue theme functions @aaronschachter and I worked on. 

cc: @DFurnes @weerd @barryclark


lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -124,3 +124,48 @@ function paraneue_dosomething_get_sponsor_logos($partners_data = array()) {
...
+  foreach ($results as $delta => $value) {
+    switch ($value['bundle']) {
+      case 'campaign':
+        $wrapper = entity_metadata_wrapper('node', node_load($value['node']->entity_id));
+        $value['display_image'] = dosomething_image_get_themed_image($wrapper->field_image_campaign_cover->value()->nid, 'square', '100x100');
+        $value['subtitle'] = $value['fields']['sm_field_call_to_action'][0];
+        break;
+      case 'static_content':
+      case 'fact_page':
+        if (!empty($vale['fields']['sm_field_subtitle'][0])) {
```
aaronschachter: spelling

lib/themes/dosomething/paraneue_dosomething/templates/notfound/node--notfound.tpl.php
```diff
@@ -12,10 +12,10 @@
  */
 ?>
 <article>
-
+<?php dpm($variables); ?>
```
aaronschachter: EEK

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -124,3 +124,48 @@ function paraneue_dosomething_get_sponsor_logos($partners_data = array()) {
...
+
+/**
+ * For given array of search results, theme a duo gallery with them.
+ *
+ * @param array $results
+ *   Array of search results.
+ *
+ * @return string
+ */
+function paraneue_dosomething_get_search_results($results) {
```
weerd: Name seems a bit confusing. Maybe the theming part should be a separate function? So this one gets the search results and then in the code where this returns to, then send the results to a duo gallery theming function? Not sure if this is the plan and it's just a few steps away or not... it's been a while :confused:

weerd: Actually yeah, seems like that is what can happen. I would suggest this function returns an array of all the results and sends that back. Then that result set is sent for theming. Thoughts?

sbsmith86: @weerd well, what i was running into is that we need the search results on both the 404 and search page, themed the same way, and there was duplicate code that was running through array of results and theming them. So the idea I had for this function was to just have a function I could pass in a set of results to, get the content I need from the result set, and do the theming. 

I agree the naming is weird, it should be more like `paraneue_dosomething_get_search_results_gallery()`.

aaronschachter: Could split this out a smaller function like `get_search_vars` in case we want to use anywhere else.. or call within the `get_search_gallery` type function ...

sbsmith86: @aaronschachter ooo i like that. will do.

weerd: Cool. Yeah, renaming it might just be enough :)

Alternatively, which may be a bit much, but maybe `paraneue_dosomething_get_search_results_gallery()` is called by both 404 and regular search page, but then this func calls `paraneue_dosomething_get_search_results()` to collect the results, and then calls the theming on those results...? Would still be DRY and a bit more modular since then you can still retrieve results without theming as gallery if that's ever needed. Not sure if that makes any sense. Just dumping my thoughts :)

sbsmith86:
I made a lot of updates to this branch based on some answers I got at stand up and comments on the PR. Can you take another look? @aaronschachter @weerd

sbsmith86:
Oh and @deadlybutter , Please take a look!

lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
```diff
@@ -125,4 +125,9 @@ function dosomething_notfound_preprocess_node(&$vars)  {
   $vars['search'] = TRUE;
   $vars['results'] = $search;
 
+  // Theme the set of results as duo gallery.
+  $vars['search_results_gallery'] = paraneue_dosomething_get_search_gallery($vars['results']);
+
+  // Make the search form availabe to the page.
+  $vars['search_form'] = drupal_render(drupal_get_form('search_form'));
```
aaronschachter: Just checking: Does this throw PHP warnings?  I think it might.  Might be worth doing one last check in the logs to make sure all looks good http://dev.dosomething.org:8888/admin/reports/dblog

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -124,3 +124,63 @@ function paraneue_dosomething_get_sponsor_logos($partners_data = array()) {
   }
   return '';
 }
+
+/**
+ * For given array of search results, theme a duo gallery with them.
```
aaronschachter: Doesn't theme a duo gallery anymore tho :santa:

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -124,3 +124,63 @@ function paraneue_dosomething_get_sponsor_logos($partners_data = array()) {
...
+function paraneue_dosomething_get_search_vars($results) {
+  if (empty($results)) {
+    return '';
+  }
+  // Preprocess each result and store the information we need.
+  foreach ($results as $delta => $value) {
+    switch ($value['bundle']) {
+      case 'campaign':
+        $wrapper = entity_metadata_wrapper('node', node_load($value['node']->entity_id));
+        $value['display_image'] = dosomething_image_get_themed_image($wrapper->field_image_campaign_cover->value()->nid, 'square', '100x100');
```
aaronschachter: If a node doesn't have a campaign cover value, this could break hardcore. We've moved away from using `entity_metadata_wrapper` cos of it.  Since you're loading the node, I'd suggest checking for jsut the raw field value, or checking for the existence of  `$wrapper->field_image_campaign_cover` before using the `value()` function

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -124,3 +124,63 @@ function paraneue_dosomething_get_sponsor_logos($partners_data = array()) {
...
+      'link' => $value['link'],
+      'image' => $value['display_image'],
+      'description' => $value['subtitle'],
+    );
+    $result_variables[$delta] = $item;
+  }
+  return $result_variables;
+}
+
+function paraneue_dosomething_get_search_gallery($results) {
```
aaronschachter, line 192: Perhaps the block comments here?

lib/themes/dosomething/paraneue_dosomething/templates/system/patterns/media.tpl.php
```diff
@@ -18,20 +18,26 @@
   <div class="wrapper">
     <?php if (!empty($content['image'])): ?>
       <?php print $content['image']; ?>
+    <?php elseif ($use_default): ?>
+      <img src='/profiles/dosomething/themes/dosomething/paraneue_dosomething/bower_components/neue/dist/assets/images/apple-touch-icon-precomposed.png' />
```
aaronschachter: Might be cleaner to use [path_to_theme](https://api.drupal.org/api/drupal/includes%21theme.inc/function/path_to_theme/7) here.  Or additionally, set the default image within the preprocess as the `$content['image']`

aaronschachter:
Looks good here, just minor comments.

lib/themes/dosomething/paraneue_dosomething/templates/notfound/node--notfound.tpl.php
```diff
@@ -27,17 +26,14 @@
     <div class="container notfound-search-results">
       <div class="wrapper">
         <?php if (isset($results_info)): ?>
-          <p class="results-info-copy"><strong><?php print $results_info ?></strong></p>
+          <p class="results-info-copy"><?php print $results_info ?></p>
+        <?php endif; ?>
+        <?php if (!empty($search_form)): ?>
```
itsjoekent: Should if empty go under the if $search to keep it all together? Also if there is no search to display it saves the extra if check

lib/themes/dosomething/paraneue_dosomething/scss/content/_notfound.scss
```diff
@@ -2,6 +2,13 @@ header.notfound p {
   color: #fff;
 }
 
-.results-info-copy {
+.search-results-copy {
```
itsjoekent: Hmm, did you fetch upstream? Currently the code looks like this https://github.com/DoSomething/dosomething/blob/dev/lib/themes/dosomething/paraneue_dosomething/scss/content/_notfound.scss

and the php print is no longer inside a <p> as the markup generates it's own html (the node-notfound.tpl is outdated in this PR as well with this error)
https://github.com/DoSomething/dosomething/blob/dev/lib/themes/dosomething/paraneue_dosomething/templates/notfound/node--notfound.tpl.php#L30

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
sbsmith86: @aaronschachter Good to know. I will fix.

lib/themes/dosomething/paraneue_dosomething/scss/content/_notfound.scss
sbsmith86: @deadlybutter no I didn't get your changes yet, but will pull them in this morning.

lib/themes/dosomething/paraneue_dosomething/templates/notfound/node--notfound.tpl.php
sbsmith86: @deadlybutter This is a check to add the search form to the page, if the search form exists. I figured we would want to show the search form regardless if there were results or not. Is that the case?

lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
sbsmith86: @aaronschachter It was, thanks! I need to get in the habit of checking that.

aaronschachter: Ya. Usually I just separate it out into 2 lines: 

```
$form = drupal_get_form('whatever'); 
$vars['form'] = drupal_render($form);
```

lib/themes/dosomething/paraneue_dosomething/templates/notfound/node--notfound.tpl.php
```diff
@@ -27,17 +26,14 @@
...
+          <?php print $search_form; ?>
         <?php endif; ?>
         <?php if ($suggestions){ print $results; }?>
```
sbsmith86: @deadlybutter @aaronschachter Wondering what `$suggestions` vs `$search` is? In what cases would `$suggestions` be true, and what does it print and does it need to be themed like the other results?

sbsmith86:
@aaronschachter @deadlybutter I addressed a lot of your comments, please review.

I'm still unclear about what the `$suggestions` and `$results_info` functionality is, even after pulling in the most recent fixes. I talked to Luke and he said that there is a static line of text that needs to go into the template above the search form (as shown in the most recent mocks) so I have added into the template. I removed the  `$suggestions` and `$results_info` from the template for now, since they don't seem to be being used, but I can add it back if I am totally misunderstanding how all this is supposed to work.

lib/themes/dosomething/paraneue_dosomething/templates/notfound/node--notfound.tpl.php
```diff
@@ -26,18 +25,12 @@
   <?php if ($search == TRUE || $suggestions == TRUE): ?>
     <div class="container notfound-search-results">
       <div class="wrapper notfound-results">
-        <?php if (isset($results_info)): ?>
-          <?php print $results_info ?>
+        <h3 class="search-results-header">We did a quick search, here is similar content that we do have</h3>
```
aaronschachter: Needs to be in `t()` function, and also missing punctuation.

lib/themes/dosomething/paraneue_dosomething/templates/notfound/node--notfound.tpl.php
```diff
@@ -26,18 +25,12 @@
   <?php if ($search == TRUE || $suggestions == TRUE): ?>
     <div class="container notfound-search-results">
       <div class="wrapper notfound-results">
-        <?php if (isset($results_info)): ?>
-          <?php print $results_info ?>
```
aaronschachter, line 30: I believe `$results_info` is when a path-based suggestion is found (#3317).  I think we need to rename these variables to be more readable.

aaronschachter:
Looks great! SHIP IT :+1: Value added to the value

itsjoekent:
@sbsmith86 The reason behind results_info was there are three possible displays of the 404 page. 
- Just the banner. In the event a lot of traffic is going to an incorrect URL and it's affecting site performance we can disable the fancy 404 features.
- Path specific results. On the legacy app, /project and /grants still drive quite a bit of traffic. Instead of showing them search results, we give them a bit of text explaining why they are there. Further, the mock calls to instead show them campaign suggestions instead of search results (hence the suggestions variable). 
- Search results. If the URL being accessed is not path specific and search functionality is enabled, we perform a search. 

Since the last two options both had to display copy above some form of 'results' (campaign recommendations or search results), I figured it would be best to just have one variable for storing that copy, so we didn't need $search_copy, $project_copy and $grant_copy. 

It's really simple to update this, we just need to change [this line](https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module#L139) to results_info instead of search_header and remove Line 35 to Line 39 [here](https://github.com/DoSomething/dosomething/blob/dev/lib/themes/dosomething/paraneue_dosomething/templates/notfound/node--notfound.tpl.php#L35) as it's now handled by results_info. 

_phew_ Hope this helped clear things up!

sbsmith86:
@deadlybutter Thanks! I was able to track down some people last week that helped clear this all up for me. The themed 404 and search pages are on staging now, if you want to take a look.

PR #3416, opened 2014-11-14
Fix Stripe Customer / Charge requests 

aaronschachter:
Creating a Stripe Customer with the token exposes the `$customer->id` property, fixing broken payment errors.


sergiitk:
:+1:

PR #3417, opened 2014-11-14
Runs UK and Canada tests on international subdomain.

sergiitk:
#### What's this PR do?
- Makes UK and Canada integration tests run on international domain so `dosomething_settings_get_affiliate_country_code()` would return right value inside the tests
- Fixes an error in `bin/test` script with url detection
- Makes `casper.drush` take into account an affiliate currently executing test at
  image

<del>Now running tests requires an amend to of `/etc/hosts` file that assigns intl hosts to 127.0.0.1, e.g. `127.0.0.1 dev.canada.dosomething.org`.
This will be fixed in the new box build.</del>

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -141,7 +141,7 @@ function runIntegrationTests() {
   }
 
   var _casperURL = "http://dev.dosomething.org:8888";
-  if(config.verbose) {
+  if(config.url) {
     _casperURL = config.url;
   }
 
--- a/tests/scripts/drush.js
+++ b/tests/scripts/drush.js
@@ -49,6 +49,9 @@ casper.drush = function(command, json) {
     if(json === true) {
       command.push("--format=json");
     }
+
+    // URI of the drupal site run drush at. Needed to run drush on the right
+    // affiliate in the multi-site environment.
     command.push("--uri=" + url);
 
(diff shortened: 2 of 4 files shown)


tests/scripts/drush.js
```diff
@@ -49,6 +49,7 @@ casper.drush = function(command, json) {
     if(json === true) {
       command.push("--format=json");
     }
+    command.push("--uri=" + url);
```
DFurnes, line 55: Could you add a quick comment here to explain what this does? Took me a few seconds to understand why Drush needed to being passed the URL, and that the variable was coming from the global var set in other Casper file.

bin/test
```diff
@@ -141,7 +141,7 @@ function runIntegrationTests() {
   }
 
   var _casperURL = "http://dev.dosomething.org:8888";
-  if(config.verbose) {
+  if(config.url) {
```
DFurnes, line 144: Ooof, good catch. :boom:

DFurnes:
:+1:!

sergiitk:
Thank @DFurnes!

tests/scripts/drush.js
sergiitk, line 55: Ok, will add it.

PR #3418, opened 2014-11-14
Stripe Publish API key variable

aaronschachter:
Makes Stripe Publish API key a variable, and adds into into JS via DRUPAL MAGIC.


sergiitk:
Good! :+1:

PR #3419, opened 2014-11-14
Fixes race condition in UK and Canada integration tests.

sergiitk:
Fixes race condition in UK and Canada integration tests by waiting for mockup user to be completely loaded. 
Error example:
image


DFurnes:
:+1:

PR #3420, opened 2014-11-14
Donate Form: Display confirmation within block content

aaronschachter:
Redirects to current page that form is rendered on, with a `?donated=1` query string.  

If query string is present, display confirmation message instead of the Donate Form.


sbsmith86:
Looks good to me :wine_glass: :+1:

aaronschachter:
:sake:

PR #3435, opened 2014-11-17
Fixes JS error

aaronschachter:
@DFurnes wrote all of this and I should deserve no credit for this commit


DFurnes:
#shipit :shipit:

PR #3439, opened 2014-11-17
Adding query string to search form

### LINKED ISSUE #3426: display the text of the search query in the search bar on the 404 page
In this example we would display "jeans" since "actnow" is being excluded per #3318

image

sbsmith86:
Fixes #3426 . Adding the query string used in the 404 search to the text field in the search form displayed on the 404 page. 

@aaronschachter @deadlybutter

### DIFF
--- a/lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
+++ b/lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
@@ -152,5 +152,5 @@ function dosomething_notfound_form_search_form_alter(&$form, &$form_state, $form
   // Get the query string from the form state.
   $query_string = $form_state['build_info']['args']['query_string'];
   // Use the query as the default text in the search form.
-  $form['basic']['keys']['#default_value'] = t($query_string);
+  $form['basic']['keys']['#default_value'] = $query_string;
 }
--- a/lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
+++ b/lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
@@ -152,5 +152,7 @@ function dosomething_notfound_form_search_form_alter(&$form, &$form_state, $form
   // Get the query string from the form state.
   $query_string = $form_state['build_info']['args']['query_string'];
   // Use the query as the default text in the search form.
-  $form['basic']['keys']['#default_value'] = $query_string;
+  if (!empty($query_string)) {
+    $form['basic']['keys']['#default_value'] = $query_string;
+  }
 }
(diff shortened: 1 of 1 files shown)


itsjoekent:
looks good to me, I'd wait for @aaronschachter as I'm still new to Drupal

lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
```diff
@@ -130,11 +130,27 @@ function dosomething_notfound_preprocess_node(&$vars)  {
...
   // Search header text.
   $vars['search_header'] = t('We did a quick search, here is some similar content that we do have.');
 }
+
+/**
+* Implementation of hook_form_FORM_ID_alter()
+*/
+function dosomething_notfound_form_search_form_alter(&$form, &$form_state, $form_id) {
+  // Get the query string from the form state.
+  $query_string = $form_state['build_info']['args']['query_string'];
```
aaronschachter, line 153: Isn't it possible that the `query_string` wouldn't be set on just a regular page load?  Since the search form appears on every page.  Can you confirm in the error logs this isn't flooding with php notice / warnings?

lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
```diff
@@ -130,11 +130,27 @@ function dosomething_notfound_preprocess_node(&$vars)  {
...
+  // Use the query as the default text in the search form.
+  $form['basic']['keys']['#default_value'] = t($query_string);
```
aaronschachter: I don't think we'd want to translate the query string... ? cc @sergii-tkachenko

lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
sbsmith86, line 153: @aaronschachter hmmm, no i'm not seeing any errors or warnings. Maybe because I am only passing in the query string argument when rendering the form on the 404 page. But I can do a double check for the existence of the query string to be more defensive.

lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
sbsmith86: @aaronschachter i will remove.

lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
aaronschachter, line 153: It might just be part of the way a drupal `search_form` is rendered.  If errors aren't not showing up in places like the homepage and stuff, feel free to keep as is.

sbsmith86, line 153: @aaronschachter yea, using i'm using `drupal_build_form()` to render this particular search form which allows you to pass in a form state. So it probably doesn't effect other forms rendered on the page. Not seeing errors related to the search for on any other pages, but I added the double check just in case, as I do think it is important to be more defensive.

sbsmith86:
@aaronschachter is this approved?

aaronschachter:
Ship it :+1:

Issue #3440, opened 2014-11-17
Tests fails with "TypeError: callback provided to sync glob"


sergiitk:
Test script fails to run with the following error:

```
/var/www/vagrant/node_modules/glob/glob.js:67
      throw new TypeError('callback provided to sync glob')
            ^
TypeError: callback provided to sync glob
    at glob (/var/www/vagrant/node_modules/glob/glob.js:67:13)
    at /var/www/vagrant/bin/test:191:5
    at Function.forEach (/var/www/vagrant/node_modules/lodash/dist/lodash.js:3297:15)
    at getFiles (/var/www/vagrant/bin/test:186:5)
    at runIntegrationTests (/var/www/vagrant/bin/test:125:20)
    at _mochaReporter (/var/www/vagrant/bin/test:76:3)
    at /var/www/vagrant/bin/test:106:5
    at /var/www/vagrant/node_modules/shelljs/src/exec.js:107:7
    at ChildProcess.exithandler (child_process.js:646:7)
    at ChildProcess.emit (events.js:98:17)
```

Here's the link to Jenkins build failure: http://admin.dosomething.org/jenkins/job/Staging-Tests/107/console

It seems like [`isaacs/node-glob`](isaacs/node-glob) `sync` function has changed in branch `4.1`. Reverting to `4.0` branch fixes the issue.

The functions fails in the [bin/test](/DoSomething/dosomething/blob/dev/bin/test#L184) script.
CC @DFurnes

sergiitk:
Tests work now: http://admin.dosomething.org/jenkins/job/Staging-Tests/108/console.

PR #3441, opened 2014-11-17
Downgrade node-glob dependency in DS test suite due to #3440.

sergiitk:
Temporarily fixes tests failure described in #3440.


DFurnes:
:shipit:

PR #3450, opened 2014-11-18
Stripe Errors

aaronschachter:
- Uses `drupal_set_message` to display errors returned from Stripe instead of `form_set_error`.  
- Adds more optional testing query string parameters to make for easier testing.

screen shot 2014-11-18 at 9 29 50 am
screen shot 2014-11-18 at 9 30 19 am
screen shot 2014-11-18 at 9 29 18 am


sergiitk:
:+1:

PR #3452, opened 2014-11-18
Using a larger image style for thumbnails

### LINKED ISSUE #3447: thumbnails are pixelated in search results
@sbsmith86 sorry just noticed this, but the thumbnail images we're pulling in for search results look pretty pixelated.

seems to be a similar issue as #3267 

again as in that instance, we do have high res square images available

image

sbsmith86:
Fixes #3447. 

@aaronschachter @weerd @DFurnes


DFurnes:
:fireworks:

weerd:
"Now with more awesome!" :+1:

PR #3455, opened 2014-11-18
force a leading zero on exp month if value is less than 10

### LINKED ISSUE #3438: Donate Form: Expiration month error
You can enter "5" instead of "05" as expiration month.  The form will shake with an error, but it is still submitted.

I advise we either allow a single digit expiration month.. or potentially have a dropdown...

tjrosario:
Adds a leading zero to exp month if value is less than ten.  This check happens upon keyup after a 200ms delay so that it's not too invasive to the user.

Note: would be ideal to add this support to the validation library.

Resolves #3438

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/validation/donate.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/validation/donate.js
@@ -36,16 +36,6 @@ define(function(require) {
   }
 
   /**
-  * Validates string for at most two digits
-  *
-  * @param {String}   string  String to be test against regex
-  * @return {boolean}
-  */
-  function isMaxTwoDigits(string) {
-    return (/^\d{1,2}$/).test(string);
-  }
-
-  /**
   * Validates string for at least three digits
   *
   * @param {String}   string  String to be test against regex
@@ -62,7 +52,6 @@ define(function(require) {
   * @return {boolean}
   */
   function isValidMonth(string) {
-    //return (/^(0[1-9]|1[0-2])$/).test(string);
     return (/^([1-9]|1[0-2])$/).test(string);
   }
 
@@ -157,7 +146,7 @@ define(function(require) {
   * @return {boolean}
   */
   function validateExpMonth(string, done, success, failure) {
-    if ( isValidMonth(string) && isMaxTwoDigits(string)) {
+    if ( isValidMonth(string)) {
       return done({
         success: true,
         message: success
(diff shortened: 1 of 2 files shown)


DFurnes:
Can we update the [validation function](https://github.com/tjrosario/dosomething/blob/122ef11495464d340740516c2fe3e77a27bed3cc/lib/themes/dosomething/paraneue_dosomething/js/validation/donate.js#L54-L56) to allow single-digit months? Unless Stripe is requiring a two-digit month? Adding the padded zero with JavaScript just seems really jarring to me.

I think you should be able to validate by checking the input after passing it through `parseInt`.

tjrosario:
@DFurnes no problem.  I changed the validation to accept single digits and check if it's a valid month (1-12)

lib/themes/dosomething/paraneue_dosomething/js/validation/donate.js
```diff
@@ -52,7 +62,8 @@ define(function(require) {
   * @return {boolean}
   */
   function isValidMonth(string) {
-    return (/^(0[1-9]|1[0-2])$/).test(string);
+    //return (/^(0[1-9]|1[0-2])$/).test(string);
```
DFurnes: Remove commented line?

lib/themes/dosomething/paraneue_dosomething/js/validation/donate.js
```diff
@@ -36,6 +36,16 @@ define(function(require) {
   }
 
   /**
+  * Validates string for at most two digits
+  *
+  * @param {String}   string  String to be test against regex
+  * @return {boolean}
+  */
+  function isMaxTwoDigits(string) {
```
DFurnes: Is this necessary? It looks like your `isValidMonth` regex should already be ensuring this.

lib/themes/dosomething/paraneue_dosomething/js/validation/donate.js
```diff
@@ -52,7 +62,8 @@ define(function(require) {
   * @return {boolean}
   */
   function isValidMonth(string) {
-    return (/^(0[1-9]|1[0-2])$/).test(string);
+    //return (/^(0[1-9]|1[0-2])$/).test(string);
+    return (/^([1-9]|1[0-2])$/).test(string);
```
DFurnes: Also might be worth adding an optional `0` on the first group in the regex for overachievers. ;)

(Current regex will fail on `01` as an input, for example.)

tjrosario: @DFurnes yep good catches.  Updated the regex for optional leading zero.

DFurnes:
Looks good! :shipit:

PR #3457, opened 2014-11-18
Expose hidden SMS Game signup form fields

aaronschachter:
Removes `#access => FALSE` from the SMS Game signup form, so we can run Optimizely tests which can set different story_id's.


sergiitk:
:+1:

PR #3458, opened 2014-11-18
Added filtering to notfound search

### LINKED ISSUE #3318: search results on 404 page
We'll want to split up parts of the URL, run a SOLR query, and display the results to the user.

Ex. /blog/puppies-running should do a SOLR query of "puppies running" and provide the user the top 4 search results for that query.

Mocks:
image

itsjoekent:
Creates a form for an admin to input search words to filter (comma separated)

@aaronschachter if you have a better idea for implementation let me know, since this should be less than 50 words I don't think it will be that slow

Last part of #3318 remaining ( Fixes https://github.com/DoSomething/dosomething/issues/3318 )

### DIFF
--- a/lib/modules/dosomething/dosomething_notfound/dosomething_notfound.admin.inc
+++ b/lib/modules/dosomething/dosomething_notfound/dosomething_notfound.admin.inc
@@ -22,7 +22,7 @@ function dosomething_notfound_admin_config_form($form, &$form_state) {
   $form['dosomething_notfound_search_filter'] = array(
     '#type' => 'textarea',
     '#title' => 'Search filter words',
-    '#description' => 'Comma seperate all filter words (eg: blog,actnow,etc)', 
+    '#description' => 'Comma separated list of all words to exclude from the search string (e.g. blog, actnow)', 
     '#default_value' => variable_get('dosomething_notfound_search_filter'),
     );
   return system_settings_form($form);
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_notfound/dosomething_notfound.admin.inc
```diff
@@ -19,5 +19,11 @@ function dosomething_notfound_admin_config_form($form, &$form_state) {
     '#default_value' => variable_get('dosomething_notfound_copy_grants'),
     '#required' => TRUE,
     );
+  $form['dosomething_notfound_search_filter'] = array(
+    '#type' => 'textarea',
+    '#title' => 'Search filter words',
+    '#description' => 'Comma seperate all filter words (eg: blog,actnow,etc)',
```
aaronschachter: Spelling.  This is to exclude the words from the search right?  I would suggest we specifically say that, maybe something like:
Comma separated list of all words to exclude from the search string (e.g. blog, actnow)

aaronschachter:
Just minor spelling mistake / copy change, then ship it  :+1: value added to the holistic value

itsjoekent:
@aaronschachter all fixed

aaronschachter:
:+1: Ship it value added to the intrinsic value

itsjoekent:
added more value

PR #3459, opened 2014-11-18
Updates to hide or show caption field in reportback.

### LINKED ISSUE #3453: Reportback Caption JS visibility
~~Blocked by #3284~~

Only display the "caption" field in the Reportback Form when the user has uploaded an image.

For now, the user cannot update the caption, and can only enter the caption when uploading the image.

weerd:
Resolves #3453

Only hides the caption field is there's not already a previously submitted reportback image. If there is one, it doesn't affect the caption so user can go back and add a caption to prior image :)

@aaronschachter 

CC: @DFurnes


aaronschachter:
ok by me :+1:

DFurnes:
Godspeed. :boat:

PR #3465, opened 2014-11-18
Adding a no results state to 404 pages

### LINKED ISSUE #3428: no search results found on 404 page
if there are no results available, the display should be:
- header
- copy: "We couldn't find what you were looking for, but try searching our site for something similar." treatment would be same as copy above search bar on current 404 page
- search bar
- header: "Or join a campaign" (h3)
- recommended campaigns (same as on the path specific 404 page)

sbsmith86:
I refactored the code a little to basically set a default version of the page to take into account when there were no results found, or if the path they hit happened to also be a term we are filtering, like /actnow, for example. This way the 404 page will always show something, unless of course, we disable the 404 functionality in rare cases. 

Fixes #3428 

@aaronschachter @deadlybutter

### DIFF
--- a/lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
+++ b/lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
@@ -154,9 +154,8 @@ function dosomething_notfound_preprocess_node(&$vars)  {
 */
 function dosomething_notfound_form_search_form_alter(&$form, &$form_state, $form_id) {
   // Get the query string from the form state.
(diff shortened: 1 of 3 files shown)


aaronschachter:
Looks good :+1: 

Not needed for merge, but we may want to create variables to store all the copy to avoid making code changes for variables like `results_copy`.

sbsmith86:
@aaronschachter great point. I'll create an issue for that.

itsjoekent:
works for me :D

Issue #3285, opened 2014-10-30
Reportback Gallery View


aaronschachter:
- [x] Displays all Reportback Images and their captions.  
- [x] Should be accessible via endpoint to make paginated Ajax requests.
- [x] Filter by Reportback File status (see #3389)
- [x] Allow random sort

aaronschachter:
@julielorch Assuming the gallery is sorted by most recent first?

PR #3467, opened 2014-11-18
Reportback Files API endpoint

aaronschachter:
Refs #3285

Creates API endpoint for getting Reportback Files, which will be used in the new Reportback Gallery to dynamically load more Reportback Files as users scroll through the gallery.  Borrows heavily from built-in Services functions for paging and page size.

Also enables `dosomething_api` enterprise wide by making it a dependency in the DS profile.  This is because all Campaigns will require this endpoint for displaying this new gallery in the Prove It sections.

Will update the wiki once merged.

Example URLs:

```
http://dev.dosomething.org:8888/api/v1/reportback_files.json?parameters[nid]=362

http://dev.dosomething.org:8888/api/v1/reportback_files.json?parameters[nid]=362&page=10&pagesize=2
```

### DIFF
--- a/lib/modules/dosomething/dosomething_api/dosomething_api.info
+++ b/lib/modules/dosomething/dosomething_api/dosomething_api.info
@@ -6,3 +6,4 @@ dependencies[] = services
 features[ctools][] = services:services:3
 features[features_api][] = api:2
 features[services_endpoint][] = drupalapi
+mtime = 1415889493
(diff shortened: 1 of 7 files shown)


lib/modules/dosomething/dosomething_api/dosomething_api.info
```diff
@@ -6,3 +6,4 @@ dependencies[] = services
 features[ctools][] = services:services:3
 features[features_api][] = api:2
 features[services_endpoint][] = drupalapi
+mtime = 1415889493
```
aaronschachter, line 9: I have no idea why this was added. Simply did a `drush fu dosomething_api` to add the `reportback_files` resource as enabled.

sergiitk:
Great.

sergiitk:
:+1:

PR #3399, opened 2014-11-12
Reportback form cropping

sbsmith86:
@aaronschachter Wanted to send a pull request so you can start check out some of the module level code I added. 

:warning:  Most of this code (especially the js) is not production ready!! Just want to start getting eyes on this so we can move it to staging to start testing on devices.

PR #3469, opened 2014-11-19
Adds DS Reportback admin config form

aaronschachter:
Gets a variable in place for turning Reportback cropping on/off.  Variable doesn't do anything yet, but will be used in #3399


sergiitk:
:+1:

PR #3471, opened 2014-11-19
404 copy variables

### LINKED ISSUE #3470: Create variables to manage copy on 404 page
As mentioned by @aaronschachter in https://github.com/DoSomething/dosomething/pull/3465. 

We should a Drupal variable to manage the `$results_copy` so that it won't require code updates to change. 

We have variables managing path specific copy, but we should also create variables for no results found and search results. 

Currently: 
Search: We did a quick search, here is some similar content that we do have.
No Results found: "We couldn't find what you were looking for, but try searching our site for something similar."

sbsmith86:
Fixes #3470 

@aaronschachter created those variables for $results_copy you mentioned.


aaronschachter:
:+1:

PR #3473, opened 2014-11-19
added top margin to container titles on campaign collections 

### LINKED ISSUE #3241: sources treatment on collections should match other places on the site
The spacing is off around sources for collections, especially between Sources and What You Get.

Let's replicate spacing on the campaign template: 
image

Here is what it looks like now on collections:
image

tjrosario:
Made spacing consistent when `container__title` immediately follows sources

Resolves #3241


lib/themes/dosomething/paraneue_dosomething/scss/content/_sources.scss
```diff
@@ -31,6 +31,9 @@
     padding-top: 18px;
   }
 
+  & + .container__title.inline--alt-color {
```
weerd: Do you need the `.inline--alt-color` class? Should this just apply to `.container__title` and not specifically ones that have the alt color?

Issue #3380, opened 2014-11-10
add title field to taxonomy collections


mikefantini:
please add to cause, action type, and issue

aaronschachter:
I'd suggest this is more of a Title Override, so there are less fields for editors to fill out.  If a value exists for Title Override, use it, otherwise use the regular Taxonomy term.

mikefantini:
@aaronschachter so this is an optional field?

aaronschachter:
Yup, that's the idea.

PR #3476, opened 2014-11-19
Added title field to taxonomy

itsjoekent:
Adds a title field to action type, cause and issue. Fixes half of #3380

@aaronschachter


aaronschachter:
:+1:

PR #3477, opened 2014-11-19
Added title preprocessing

### LINKED ISSUE #3380: add title field to taxonomy collections

itsjoekent:
Fixes #3380

@aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
+++ b/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
@@ -66,11 +66,18 @@ function dosomething_taxonomy_preprocess_taxonomy_term(&$vars) {
   $gallery = dosomething_campaign_get_campaign_gallery($campaigns, $size, $source);
   $vars['campaign_gallery'] = $gallery;
 
+  // Set title field to name if there is nothing set.
+  if (empty($vars['field_title'])) {
+    $vars['field_title'][0]['value'] = $vars['name'];
+  }
+  $vars['title'] = $vars['field_title'][0]['value'];
+
   // Term fields:
   $text_fields = array(
     'intro_title',
     'subtitle',
   );
+
   foreach ($text_fields as $label) {
     $field = "field_{$label}";
     if (isset($vars['content'][$field])) {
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
```diff
@@ -66,11 +66,18 @@ function dosomething_taxonomy_preprocess_taxonomy_term(&$vars) {
...
+    $vars['field_title'][0]['value'] = $vars['name'];
+  }
+  $vars['title'] = $vars['field_title'][0]['value'];
+
   // Term fields:
   $text_fields = array(
     'intro_title',
     'subtitle',
   );
+
```
itsjoekent, line 81: Woops, made a new line here on accident when debugging - want me to remove this or no big deal?

lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
```diff
@@ -66,11 +66,18 @@ function dosomething_taxonomy_preprocess_taxonomy_term(&$vars) {
   $gallery = dosomething_campaign_get_campaign_gallery($campaigns, $size, $source);
   $vars['campaign_gallery'] = $gallery;
 
+  // Set title field to name if there is nothing set.
+  if (empty($vars['field_title'])) {
+    $vars['field_title'][0]['value'] = $vars['name'];
+  }
+  $vars['title'] = $vars['field_title'][0]['value'];
```
aaronschachter: This is kind of confusing. Could it go like this?

```
$vars['title'] = $vars['name'];
if (!empty($vars['field_title'])) {
  $vars['title'] = $vars['field_title'][0]['value'];
}
```

Should then also add:

```
drupal_set_title($vars['title']);
```

which sets the title in the `<HEAD>`

Also, you'll need to change the `taxonomy-term.tpl.php` to output `$title`.

lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
aaronschachter, line 81: Not really a biggie.. although you could do the check for title here too by adding it in here (since its basically doing what your checking for with the `empty` / setting the variable

lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
itsjoekent: Okie dokes!

aaronschachter:
:+1: awesome!

itsjoekent:
added value

PR #3481, opened 2014-11-19
fixed spacing between sources and headlines on collections

### LINKED ISSUE #3241: sources treatment on collections should match other places on the site

tjrosario:
Resolves #3241


weerd:
LOL :+1:

Issue #3474, opened 2014-11-19
Responsive featured mosaic image (reasons you won't believe)


aaronschachter:
- [x] Refactor campaign gallery template to use gallery.tpl.php with `mosaic` layout (deprecating the Campaign Gallery tpl)
- [x] Refactor campaign gallery tiles to use a new `tile.tpl.php` (deprecating the Campaign Block tpl)
- [x] Move the `dosomething_campaign_get_campaign_gallery` function into paraneue as theme function. For now, hardcode two sizes: `400x400` for tiles, and use `768x768` for the featured image (which is always the first). the featured image is only needed if a `$feautred = TRUE` parameter is set.
- [x] To unblock #3267,  add a data-attribute to the `<img>` , set to the path to the source of the large 768x768 image. Still output the img in the tile as the 400x400.
- [x] Replace `campaign_gallery.tpl.php` in Taxonomy Term template with Mosaic `gallery.tpl.php`
- [x] Replace `campaign_gallery.tpl.php` in Campaign Group template with Mosaic `gallery.tpl.php`
- [x] Delete `campaign-gallery.tpl.php` and relevant theme functions

weerd:
For adding the src to the large image:

```
<img src="...path-to-small-image-here..." data-src-large="...path-to-large-image-here..." />
```

PR #3485, opened 2014-11-19
Paraneue Tile pattern template

aaronschachter:
Groundwork for #3474: Adds a Tile template (`tile.tpl.php`) and renders the Taxonomy Campaign Gallery using the paraneue `gallery.tpl.php` and said new `tile.tpl.php`


sergiitk:
:+1:

PR #3492, opened 2014-11-20
Stripe form double bind

tjrosario:
- Provides a workaround to prevent the form "submit" event to fire twice when creating a stripe token
- Disabled autocomplete on credit card fields

### DIFF
--- a/lib/modules/dosomething/dosomething_payment/dosomething_payment.module
+++ b/lib/modules/dosomething/dosomething_payment/dosomething_payment.module
@@ -132,6 +132,7 @@ function dosomething_payment_form($form, &$form_state, $config = array()) {
       'data-validate' => 'cvv',
       'data-validate-required' => '',
       'data-stripe' => 'cvc',
+      'autocomplete' => 'off',
     ),
   );
   $form['exp_month'] = array(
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_payment/dosomething_payment.module
```diff
@@ -132,6 +132,7 @@ function dosomething_payment_form($form, &$form_state, $config = array()) {
       'data-validate' => 'cvv',
       'data-validate-required' => '',
       'data-stripe' => 'cvc',
+      'autocomplete' => 'off',
```
DFurnes, line 135: Is there a particular reason we're disabling autocomplete here? I hate entering my credit card manually.

/lazy

DFurnes:
Looks fine! :ship:

lib/modules/dosomething/dosomething_payment/dosomething_payment.module
tjrosario, line 135: It's just a matter of security, especially with the cvv.  I followed the conventions of the Stripe embedded form where they disable it on all fields.

PR #3495, opened 2014-11-20
Making a variable for suggested campaigns copy

### LINKED ISSUE #3486: remove recommended campaigns header copy for 404 page
we're just going to put the copy in the section above rather than giving these recommended campaigns a header. this will be for both for a 404 search page with no results found and for a 404 path specific page

sbsmith86:
Fixes #3486 . Also updated the mark up for path specific pages vs other pages so we use H3 when needed vs just the p tag.

@aaronschachter @deadlybutter

### DIFF
--- a/lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
+++ b/lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
@@ -98,6 +98,7 @@ function dosomething_notfound_preprocess_node(&$vars)  {
     $vars['notfound_page_type'] = "path-specific";
     // Sets the copy above the results.
     $vars['results_copy'] = check_markup($copy, 'markdown');
+    $vars['suggestions_header'] = variable_get('dosomething_notfound_copy_suggested_campaigns');
     // Raw results is themed in preprocessing and stored as $campaign_results.
     $vars['raw_campaign_results'] = dosomething_notfound_get_recommended_campaigns();
     return;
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
```diff
@@ -98,6 +98,7 @@ function dosomething_notfound_preprocess_node(&$vars)  {
     $vars['notfound_page_type'] = "path-specific";
     // Sets the copy above the results.
     $vars['results_copy'] = check_markup($copy, 'markdown');
+    $vars['suggestions_header'] = variable_get('dosomething_notfound_copy_suggested_campaigns');
```
aaronschachter, line 101: Let's add a default in case the variable isn't set: [variable_get](https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/variable_get/7). That way, you won't have to set the variable manually on stage/prod (and protects against an instance where the variable is deleted by accident).

lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
```diff
@@ -144,7 +145,7 @@ function dosomething_notfound_preprocess_node(&$vars)  {
   $search_form = drupal_get_form('search_form');
   $vars['search_form'] = drupal_render($search_form);
   $vars['results_copy'] = variable_get('dosomething_notfound_copy_no_results');
-  $vars['suggestions_header'] = t('Or join a campaign');
+  $vars['suggestions_header'] = variable_get('dosomething_notfound_copy_suggested_campaigns');
```
aaronschachter, line 148: Same here.

lib/modules/dosomething/dosomething_notfound/dosomething_notfound.module
sbsmith86, line 101: @aaronschachter Makes sense! However, they just decided to kill this copy all together, so I'm just going to update this branch so it removes it :)

PR #3498, opened 2014-11-20
Adds missing Campaigns container

aaronschachter:
Missed this in #3485, as the Taxonomy Term now uses the `gallery.tpl.php` instead of `campaign-gallery.tpl.php`.  I'm working on refactoring the theme layer to remove the `campaign-gallery.tpl.php` completely (need to update Campaign Group template next)


weerd:
:+1:

Issue #3267, opened 2014-10-28
featured campaign on collections is super pixelated


mikefantini:
## The fix:

See #3474 - We need JS which handles mobile / desktop transition; on desktop, dynamically switch the image src to the data-attribute src defined to replace the 400x400 with the 768x768
## The bug:

Using `400x400` preset for all Campaign Galleries makes the Featured image pixellated.

![image](https://cloud.githubusercontent.com/assets/6330971/4810900/ba407dd2-5eb6-11e4-8f1a-7a216968f185.png)

![image](https://cloud.githubusercontent.com/assets/6330971/4810908/c729db1a-5eb6-11e4-9f85-b523d58c7be7.png)

example http://staging.beta.dosomething.org/volunteer/animals

mikefantini:
happening on the homepage too

aaronschachter:
My guess here is that the `$size` parameter we are passing when we preprocess these galleries with `dosomething_image_get_image` is too small.

barryclark:
@aaronschachter @tjrosario mentioned that this is more of a back-end issue, would you be able to tackle this?

aaronschachter:
This will be handled in #3474.  Assigning to @weerd, as the JS will be what fixes this.

aaronschachter:
@weerd This issue is now ready for JS-ification. I've added the changes in #3474 for the Taxonomy Term template, and am now working on applying them over to the Campaign Group template as well.

weerd:
JS-ification sequence commencing!

PR #3499, opened 2014-11-20
Campaign Tile image: data-src-large attribute

aaronschachter:
Adds the `768x768` image URL as a `data-src-large` attribute to allow fenders to begin fix for #3267.

Note: This will only appear upon cache clear, as the DS Image cache will need to be truncated to see the new attributes.

Output:
screen shot 2014-11-20 at 12 50 00 pm

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
@@ -225,14 +225,14 @@ function paraneue_get_gallery_item_content($nid, $source = NULL) {
     // Presets for the node's Cover Image display.
     $ratio = 'landscape';
     $default = '400x400';
-    $high_res = '768x768';
+    $large = '768x768';
     // Get the Cover Image node nid.
     $nid_image = $node->field_image_campaign_cover[LANGUAGE_NONE][0]['target_id'];
     // Get its high-res URL.
-    $high_res_url = dosomething_image_get_themed_image_url($nid_image, $ratio, $high_res);
+    $large_url = dosomething_image_get_themed_image_url($nid_image, $ratio, $large);
     // Set attributes to pass to the themed image.
     $attributes = array(
-      'data-src-large' => $high_res_url,
+      'data-src-large' => $large_url,
     );
     $image = dosomething_image_get_themed_image($nid_image, $ratio, $default, $node->title, $attributes);
   }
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -214,16 +214,29 @@ function paraneue_dosomething_get_search_gallery($results) {
...
   if ($source) {
     $link .= '?source=' . $source;
   }
+
+  // If the node has a Cover Image:
   if (!empty($node->field_image_campaign_cover)) {
+    // Presets for the node's Cover Image display.
+    $ratio = 'landscape';
+    $default = '400x400';
+    $high_res = '768x768';
```
weerd: Nit-picky, but I'd name it `$large`... cause I foresee `$high_res` or `$high_def` being even larger and for 2x retina displays as we add more in future...

aaronschachter: :fries: ok will change

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -214,16 +214,29 @@ function paraneue_dosomething_get_search_gallery($results) {
...
+    // Get the Cover Image node nid.
     $nid_image = $node->field_image_campaign_cover[LANGUAGE_NONE][0]['target_id'];
-    $image = dosomething_image_get_themed_image($nid_image, 'landscape', $size, $node->title);
+    // Get its high-res URL.
+    $high_res_url = dosomething_image_get_themed_image_url($nid_image, $ratio, $high_res);
```
weerd: Is ratio always landscape even though asking for square? I forget how this image request in Drupal works :confounded:

weerd:
Apart from comments, looks good :dancer: :+1:

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
aaronschachter: The ratio corresponds to the field to pull the file from on the Image node.  There is Landscape, Square, Thumbnail, and now Video, too.  This could easily be swapped out for the Square field, I was basically following what was already in there.  For Cover Images, each Image node needs both a Square and Landscape file uploaded in order to display, so .. we can change this to Square field.  I can do in the next round of changes.

PR #3503, opened 2014-11-20
Adds call to action into preprocessing & template

### LINKED ISSUE #3383: add call to action to taxonomy collections
## Tasks
- [x] Add existing field `field_call_to_action` to the Cause, Issue, and Action Type Taxonomy vocabularies, and `drush fu dosomething_taxonomy` to export the fields into the DoSomething Taxonomy feature.
- [x] Template changes:  (`taxonomy-term.tpl.php`)
  - [x] New cta div like Static Content, where the button is hardcoded to say "Do it" and links to the Campaign Gallery on the page (e.g. `<a href="#gallery">` or whatever the div is called).  Example: http://staging.beta.dosomething.org/about
  - [x] If a value exists for `field_call_to_action`, output it above the button.
## Description

same display as static content, but should only appear on the bottom of the page above the footer

we'll want this to link to back up to the campaign tiles. I think it probably makes sense for this to take you right to the campaign tiles as opposed to the intro title, but open to what you think.

and does this then mean that we won't need the link field? if we're getting rid of that I say we get rid of the button field as well and just go with DO IT for all of these. but again let me know what you think

itsjoekent:
Fixes https://github.com/DoSomething/dosomething/issues/3383

@aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
+++ b/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
@@ -105,6 +105,11 @@ function dosomething_taxonomy_preprocess_taxonomy_term(&$vars) {
     $vars['zendesk_form'] = drupal_get_form('dosomething_zendesk_form', $vars['term']);
   }
 
+  // Call to action.
+  if (!empty($vars['field_call_to_action'])) {
+    $vars['call_to_action'] = $vars['field_call_to_action'][0]['value'];
+  }
+
 }
 
 /**
--- a/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
+++ b/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
@@ -73,6 +73,7 @@ function dosomething_taxonomy_preprocess_taxonomy_term(&$vars) {
   $text_fields = array(
     'intro_title',
     'subtitle',
+    'call_to_action',
   );
 
   foreach ($text_fields as $label) {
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
```diff
@@ -105,6 +105,11 @@ function dosomething_taxonomy_preprocess_taxonomy_term(&$vars) {
     $vars['zendesk_form'] = drupal_get_form('dosomething_zendesk_form', $vars['term']);
   }
 
+  // Call to action.
+  if (!empty($vars['field_call_to_action'])) {
+    $vars['call_to_action'] = $vars['field_call_to_action'][0]['value'];
```
aaronschachter: This variable should be set by adding `call_to_action` within the `$text_fields` array to keep this consistent.

lib/themes/dosomething/paraneue_dosomething/templates/taxonomy/taxonomy-term.tpl.php
```diff
@@ -41,8 +41,19 @@
     </section>
   <?php endif; ?>
 
+  
+  <div class="cta">
+    <div class="wrapper">
+      <?php if (isset($call_to_action)): ?>
+        <h2 class="__message"><?php print $call_to_action; ?></h2>
+      <?php endif; ?>
+      <a href="#gallery" class="btn">Do it</a>
```
aaronschachter: Needs to be translated: `<?php print t("Do it"); ?>`

aaronschachter:
:+1:

PR #3504, opened 2014-11-20
Use mosaic gallery tpl for Campaign Group

aaronschachter:
Refs #3474

Uses the `gallery.tpl.php` and `tile.tpl.php` functions for the Campaign Group gallery.

The only thing that's off here is that the`<li>` no longer have the "campaign" or "-published" / "-unpublished" classes.  I am adding the "-unpublished"  class into the `<article>` and not the `<li>`... as its way trickier to get it in the `<li>` .  

Need to look into this with @weerd to figure out whether or not we can use this approach instead of setting the class on the `<li>`

Example: What happens is the gray background is not applied since the "-unpublished" class doesn't live on the `<li>`.

screen shot 2014-11-20 at 4 06 49 pm

Once this is approved and merged I will remove the `campaign-gallery.tpl.php` template and functions to close out #3474 and DRY

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.module
+++ b/lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.module
@@ -25,29 +25,15 @@ function dosomething_campaign_group_preprocess_node(&$vars) {
       ),
     );
 
-    // Initialize variable assuming Campaign Group is pre-launch state.
-    $pre_launch = TRUE;
-    $vars['campaign_gallery'] = NULL;
-    // If campaign have been added to this Campaign Group:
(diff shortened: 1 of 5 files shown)


aaronschachter:
@weerd To look into whether or not we can go this route with Campaign Groups (currently unpublished campaigns don't show the gray BG because it expects the `-unpublished` class on the `<li>` not the `<article>`)

weerd:
Took a look and seems like it should be fine to remove the `campaign` class on the `<li>` and move the `-unpublished` from the `<li>` to the `.tile` :)

aaronschachter:
Great!  Thanks @weerd ! Merging.

weerd:
post-merge :+1:

PR #3505, opened 2014-11-20
Taxonomy gallery show more

weerd:
Addresses the "greater than 5 campaigns" show more issue in ##3265. This provides a new JS object where you can pass a container element, a collection of 1 or more items inside container and it will hide them, and then show them via a "Show More" button. 

This also will work smoothly if there happens to be more than one mosaic gallery on a page and hide extra items on all of the galleries. @mikefantini said it's highly unlikely, something like 0.00001%, but that means there's still A CHANCE!

@DFurnes 

CC: @barryclark @aaronschachter @sbsmith86


lib/themes/dosomething/paraneue_dosomething/js/revealer/Revealer.js
```diff
@@ -0,0 +1,58 @@
...
+   * via an interactive "Show More" button.
+   * 
+   * @param {jQuery} $container The container element object that encloses the items to hide.
+   * @param {jQuery} $content   A collection of objects to hide.
+   * @param {string} category   Specify a category type for the content to provide more unique behavior.
+   */
+  var Revealer = function($container, $content, category) {
+    this.$container = $container;
+    this.$content   = $content;
+    this.$button    = $("<button class=\"btn tertiary\">Show More</button>");
```
aaronschachter: If we wanna get picky picky, this Show More should be translated for our pals in Congo.

DFurnes:
Looks great! :+1:

lib/themes/dosomething/paraneue_dosomething/js/revealer/Revealer.js
weerd: Hmmm, ok. Will try it!

PR #3506, opened 2014-11-20
Fixed CTA location

### LINKED ISSUE #3383: add call to action to taxonomy collections

itsjoekent:
confirmed with @mikefantini 

@aaronschachter 

fixes https://github.com/DoSomething/dosomething/issues/3383

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/taxonomy/taxonomy-term.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/taxonomy/taxonomy-term.tpl.php
@@ -41,16 +41,6 @@
     </section>
   <?php endif; ?>
 
-  
-  <div class="cta">
-    <div class="wrapper">
-      <?php if (isset($call_to_action)): ?>
-        <h2 class="__message"><?php print $call_to_action; ?></h2>
-      <?php endif; ?>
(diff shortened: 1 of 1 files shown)


aaronschachter:
:+1: COMPUTER SCIENCE

itsjoekent:
YES

PR #2527, opened 2014-06-13
DoSomething config admin menu group

aaronschachter:
Creates a `admin/config/dosomething` menu to place our custom configuration forms.

screen shot 2014-06-13 at 1 08 09 pm

PR #3508, opened 2014-11-21
Add tag as array

### LINKED ISSUE #3507: Campaign transactional email missing additional tags
The `campaign_signup` email messages should have the "drupal_campaign_signup" tag as well as the campaign ID as a tag. The campaign ID tag is missing.

deezone:
Fixes #3507
- `campaign_signup` transactional email messages were missing additional message tags. The last tag addition was overwriting previous tags as it needed to be added as an array item.
- Also fixes admin path to point to config page

### DIFF
--- a/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
+++ b/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
@@ -10,7 +10,7 @@
  */
 function dosomething_mbp_menu() {
 
-  $items['admin/config/services/dosomething-mbp'] = array(
+  $items['admin/config/dosomething/dosomething_mbp'] = array(
     'title' => 'DoSomething MBP',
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
```diff
@@ -10,7 +10,7 @@
  */
 function dosomething_mbp_menu() {
 
-  $items['admin/config/dosomething/dosomething_mbp'] = array(
+  $items['admin/config/services/dosomething-mbp'] = array(
```
aaronschachter: This should be kept in the dosomething menu group to keep all of our custom admin pages in the same place. (See #2527 and #3362) Please update the .info path instead.

aaronschachter:
Bug fix looks fine, pls just keep the menu path as is per comment.

aaronschachter:
:+1:

PR #3512, opened 2014-11-21
Remove campaign gallery tpl

aaronschachter:
Removes now deprecated `campaign-gallery.tpl.php`, since the Taxonomy Term and Campaign Group template files make use of the `gallery.tpl.php` pattern template.

Also cleans up PHP Notices.


sergiitk:
:+1: Great

PR #3513, opened 2014-11-21
Taxonomy tweaks

tjrosario:
- removed right margin from compacted mixin
- ensure columned sections do not get compacted


weerd:
:+1:

PR #3518, opened 2014-11-21
Fixes non-working reportback cache clear

aaronschachter:
Found this bug while testing the Campaign Closed reportback gallery.


sergiitk:
:+1:

PR #3521, opened 2014-11-21
Update to remove styles now placed in Neue.

weerd:
Removing styles that have been moved into Neue! :boom: 

@DFurnes


DFurnes:
:runner:

PR #3525, opened 2014-11-24
Allow admins to create dummy reportbacks for Close Gallery

### LINKED ISSUE #3514: need the ability to add reportbacks without associating them to a real user/First Name
We're having a hard time knowing which user to associate past reportbacks to using https://beta.dosomething.org/admin/users/reportbacks/add.

This makes it hard to add reportbacks for old campaigns we are putting in the closed state.

And we can't just use one fake user because First Name is required and First Name gets displayed on the closed state.

cc: @lkpttn

aaronschachter:
Closes #3514 

The `admin/users/reportbacks/add` form introduced in #3469 is working fine, however the SQL query which displays promoted Reportbacks was grabbing the User uid from the File, and not the Reportback.  This was causing the admin name to display instead of the user associated with the reportback.


sergiitk:
:+1:

Issue #3468, opened 2014-11-18
some campaigns are showing up multiple times in search


mikefantini:
![image](https://cloud.githubusercontent.com/assets/6330971/5098435/0dada186-6f54-11e4-91e4-02a8b8ba2e1f.png)

![image](https://cloud.githubusercontent.com/assets/6330971/5098440/1534b9bc-6f54-11e4-9b03-b3dca590319e.png)

and I confirmed that these are all the same node (nid=56 for crisis crew)

blisteringherb:
I'm seeing the correct results showing up. The cron job may have run and fixed the issue.
![image](https://cloud.githubusercontent.com/assets/28132/5100423/2d6ff454-6f74-11e4-932d-cd32c70aaabb.png)

mikefantini:
@blisteringherb if you go to page 2 or page 3 of the results you'll see what I mean. Sorry I should have been more clear about that

blisteringherb:
@mikefantini It looks like there were two instances of the index.  I removed one and reindexed, and now things seem to look good.

mikefantini:
@blisteringherb hmmm now seeing multiples of Senior Socials and Ban Binaries on pages 2 and 3 of a "bullying" search:

![image](https://cloud.githubusercontent.com/assets/6330971/5107762/4c69bddc-6fd2-11e4-83ed-ad86d395d53a.png)

![image](https://cloud.githubusercontent.com/assets/6330971/5107771/653eb678-6fd2-11e4-9b94-3db1bf6ff4b6.png)

blisteringherb:
I've taken a look in the index itself and each campaign is indexed only once, so it seems that this is possibly an issue with the pagination.  I'd suggest trying to print out all results on one page to see what's generated there.

mikefantini:
assigning over to @sbsmith86. let me know if you have any questions

mikefantini:
no longer seeing this on bully text or crisis crew, but still seeing this with senior socials and donate on the 3rd page of a "bullying" search

![image](https://cloud.githubusercontent.com/assets/6330971/5173896/7fe38dc4-73fa-11e4-950c-0e4da775404b.png)

mikefantini:
once prod started acting properly again this was resolved

sbsmith86:
@mikefantini Sweet! :candy:

PR #3531, opened 2014-11-24
Dup search results bug

sbsmith86:
Fixes duplicate search results showing #3468 and fixing the `get_search_vars()` function so it sets default variables for anything that is not a campaign or fact page. @aaronschachter

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
@@ -158,6 +158,7 @@ function paraneue_dosomething_get_search_vars($results) {
           'image' => $value['display_image'],
           'description' => $value['subtitle'],
         );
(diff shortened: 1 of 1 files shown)


aaronschachter:
The code looks okay, but I don't follow how this fixes it ?

sbsmith86:
@aaronschachter we were looping through each result and checking the bundle type. If it was a campaign or fact page, we would set an array `item` with the variables needed for theming and continue on with the loop. But if the result wasn't one of those bundle types, then the `item` array wouldn't be updated with new values for that result. Instead, it used the values stored in the `item` array already, causing what appeared to be duplicated search results. 

This fix, goes adds the all the variables need for theming each result only when the variables for that result get processed. And using the default allows us to take into consideration all results. 

I hope that makes sense :disappointed_relieved:

aaronschachter:
makes sense! :+1:

PR #3533, opened 2014-11-24
Video poster

### LINKED ISSUE #3424: 404 page video needs a poster image
The `video` tag on the 404 page needs a `poster="…"` attribute to show an image while the video file is loading. Ideally we want this to be the same as the first frame of the video.

Otherwise the user will just see a black background while the video is loaded.

/cc @mikefantini

itsjoekent:
Adds a poster field to the Image node & prints the URL inside the notfound header.

Fixes #3424 

@aaronschachter


aaronschachter:
Why does this need to be a seperate field?  Couldnt' we use the existing landscape or square fields?

itsjoekent:
On staging the content of those fields is different than what would be the videos poster. For example, Badger: 
screen shot 2014-11-24 at 3 27 55 pm

Why this is, well I have no idea.

itsjoekent:
The only way we could re-use the field would be landscape images being the first frame of the video, which is totally do-able but we'd need to change some content around it seems

aaronschachter:
I kind of like using Landscape images instead... only because the Landscape field would never be displayed right?

itsjoekent:
if we did landscape it would be displayed for the amount of time it takes to load the video - on my home connection thats basically nothing but for others it might be a few seconds

but if the landscape field is the first video frame, its not an issue. it just means we need to redo some content

aaronschachter:
Discussed with @DFurnes .  let's keep as a separate field.  But do you mind exporting it to DS Image instead of DS Not found?  You basically need to remove all field_base/field_instances code from DS Notfound, clear cache, and then re-export to DS Image.  Picky, but it's cleaner to keep all of teh Image node's fields in the one feature module.

aaronschachter:
I AM IDIOT. Ship it :+1:

itsjoekent:
:laughing: 

value added

lib/modules/dosomething/dosomething_notfound/dosomething_notfound.features.field_base.inc
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+/**
+ * Implements hook_field_default_field_bases().
+ */
+function dosomething_notfound_field_default_field_bases() {
+  $field_bases = array();
+
+  // Exported field_base: 'field_video_poster'
+  $field_bases['field_video_poster'] = array(
```
aaronschachter, line 14: Here is the culprit.

lib/modules/dosomething/dosomething_notfound/dosomething_notfound.info
```diff
@@ -7,11 +7,14 @@ dependencies[] = dosomething_campaign
 dependencies[] = dosomething_static_content
 dependencies[] = entityreference
 dependencies[] = features
+dependencies[] = image
 dependencies[] = node
 dependencies[] = strongarm
 dependencies[] = text
 features[ctools][] = strongarm:strongarm:1
 features[features_api][] = api:2
+features[field_base][] = field_video_poster
```
aaronschachter, line 16: And here.  Dammit I wasn't an idiot.

PR #3535, opened 2014-11-24
Kill video poster

itsjoekent:
Kills duplicate variable @aaronschachter


aaronschachter:
:+1:

itsjoekent:
removed value, or was it ever value to begin with?

Issue #3183, opened 2014-10-17
International sites need to be able to customize phone number format placeholder


DFurnes:
Validation for the phone number field (such as on user login) currently validates any domestic or international number with any country code. The placeholder `(555) 555-5555` that we give to users, however, is not always country-appropriate.

We need a way for each site to use country-appropriate date placeholders, either through translation (which is what we're doing on the UK site through the "UK English" translations), based on the site region settings, or through the admin panel.

/cc @aaronschachter @sergii-tkachenko

aaronschachter:
@DFurnes Is the string wrapped in a `t` function?  If so, this can just be done via the admin translate UI.

DFurnes:
@aaronschachter Yep, that's what we're doing on the UK site:

![translation interface on UK](https://cloud.githubusercontent.com/assets/583202/4703148/5d9603e4-5868-11e4-8f31-de9ffbc8a26e.png)

The string for the placeholder `(555) 555-5555` isn't showing up as translatable on other English-speaking sites, such as Canada, presumably because they don't have any other languages set up to translate to:

![translation interface on Canada QA](https://cloud.githubusercontent.com/assets/583202/4703168/853d31b0-5868-11e4-9953-8d7edf93e513.png)

I see there's an option to add a [custom language](https://www.drupal.org/node/24593), but I want to run that by someone before we do it to make sure that's the correct way of handling this.

aaronschachter:
Hmmm. That I'm not sure of.  I don't see a English (Canadian) language to set as the default language, which would be ideal.  

It's possible that this could be an override within the Canada settings.php, like this: https://github.com/DoSomething/dosomething/blob/v0.3.55/config/settings.php#L114

.. but not sure if that gets too into crazy town either (or.. is it just the cleanest?)

@sergii-tkachenko @blisteringherb Any thoughts?

sergiitk:
@aaronschachter confirming that there's no `en-ca` language in Drupal.

sergiitk:
@aaronschachter let's try with settings override. I think there shouldn't be too much translation exceptions and the fact there's no special language in Drupal proves this point.

aaronschachter:
@blisteringherb @sergii-tkachenko  Any thoughts as to how to add this to Canada's settings.php particularly?  And to keep it tracked within code?

blisteringherb:
@aaronschachter Maybe we can have an additional include statement in the settings.php that includes a settings file that's kept in version control.

[...]

PR #3553, opened 2014-11-25
Stringoverrides

### LINKED ISSUE #3545: Change DS Canada log in modal header
In the log in modal, change the header from "Log in to get started!" to "Log in to your DoSomething or TakingITGlobal account to get started!".

image

to 

image

aaronschachter:
Makes everything in the world translatable via `admin/config/regional/stringoverrides`.

Here's the catch.  I can't save records here with the `$conf['locale_custom_strings_en']['']` setting in settings.php.  Removing it makes things work, but it means we have to manually add the string back in after deployment.  @sergii-tkachenko Anything I'm missing or does that seem right?

Also allows for easy Canadian translation to allow an admin use UI to fix #3545 and #3183.  

screen shot 2014-11-25 at 5 37 32 pm

### DIFF
--- a/config/settings.php
+++ b/config/settings.php
@@ -110,10 +110,6 @@ if (is_readable('sites/default/settings.'. $environment .'.php')) {
   include_once('sites/default/settings.'. $environment .'.php');
 }
 
-$conf['locale_custom_strings_en'][''] = array(
-   'Registration successful. You are now logged in.' => "You've created an account with DoSomething.org."
- );
-
 $conf['optimizely_id'] = '747623297';
 
 $conf['dosomething_is_affiliate'] = FALSE;
--- a/lib/profiles/dosomething/dosomething.info
+++ b/lib/profiles/dosomething/dosomething.info
@@ -47,6 +47,7 @@ dependencies[] = redirect
 dependencies[] = redis
 dependencies[] = search
 dependencies[] = services
+dependencies[] = stringoverrides
 dependencies[] = strongarm
 dependencies[] = taxonomy_access_fix
 dependencies[] = uuid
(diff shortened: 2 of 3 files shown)


sergiitk:
Looks good to me. But maybe we should enable this module only on Canada?

aaronschachter:
It doesn't work without removing the `$conf['locale_custom_strings_en']` from settings.php though...

aaronschachter:
Records don't save in the UI on `admin/config/regional/stringoverrides`

aaronschachter:
Per https://www.drupal.org/project/stringoverrides#troubleshooting

> Check your settings.php for any $conf["locale_custom_strings_en"] overrides, these conflict with the administration interface

Removing it from settings.php seems like the easiest path of resistance, vs messing with various settings.php configs to enable this on Canada only.  thoughts @sergii-tkachenko @mshmsh5000  ?

It does mean we need to manually set this in each site if we go this route.

mshmsh5000:
I haven't audited the settings.php configs to see how much we'll need to recreate after enabling the module. But assuming it's not insane, I'm happy to do any one-time manual config if it means we get String Overrides to work. It's a much more intuitive interface for non-devs to use.

aaronschachter:
I agree, I think it's going to make things easier.

Upon deployment, all we need to do is create an entry for:

`Registration successful. You are now logged in.`

overridden as:

`You've created an account with DoSomething.org.`

It actually might be best for this message to be site specific per affiliate sites anyway,. e.g. `You've created an account on congo.dosomething.org`

If we deploy tomorrow (Wednesday) someone will need to set this up manually (although nothing will break if it's not done, it will simply display the default message).  If it goes deploys on Monday, I can set it.

PR #3555, opened 2014-11-26
Canada: allow response code 302.

sergiitk:
Larvel-user-api also could response `< HTTP/1.1 302 Moved Temporarily` on success.


aaronschachter:
:+1:

mshmsh5000:
:+1:

aaronschachter:
For future PR, I could see this as a setting within the external auth module, where you can configure what codes to allow via db to avoid hardcoding...

PR #3573, opened 2014-12-01
Binding touch events to image not document

### LINKED ISSUE #3571: Scrolling disabled when cropping is turned on
Can't scroll the modal when cropping images is turned on.

sbsmith86:
Of course, if you bind events to the whole document and then preventDefault() you will turn on scrolling completely. Updated the code so that the touch events controlling dragging and resizing are binded to the image crop container only. Fixes #3571 

@DFurnes


DFurnes:
Ooof! :shipit:

PR #3575, opened 2014-12-01
Check Reportback uid and permissions upon save

aaronschachter:
Checks for permission `edit any reportback` to allow creating a new role for Robot Users, which could potentially only have the `edit any reportback` permission and various other system permissions.


sergiitk:
:+1:

PR #3576, opened 2014-12-01
Fixes #3574

### LINKED ISSUE #3574: Bug in Profile Doing links
Profile links break when viewing the profile at `user/5`, the links point to `user/node/1424`.  

These should be aliases too.

aaronschachter:
Fixes inconsistent links between viewing profile on `user` and `user/[uid]`

Sets the full site URL as the `link` variable, as the search vars that use this template use the full site URL too.


sbsmith86:
looks good!

PR #3577, opened 2014-12-02
Signup API endpoint

aaronschachter:
Enables the `campaigns/[nid]/endpoint` and allows `uid` and `source` to be passed as Signup properties.

Adds new permission `edit any signup` which is checked when a request is made when the uid does not match the logged in user's uid.

```
POST /api/v1/campaigns/1273/signup HTTP/1.1
Host: dev.dosomething.org:8888
Content-Type: application/json
Accept: application/json
X-CSRF-Token: efkrSDe0fJg6sqseLcIVv7XFxTRb3npxmHFrapRru6U
Cache-Control: no-cache

{ "uid": 15, "source": "api jam" } 
```

Response: Signup sid, if success.

```
[
    "1307"
]
```


sergiitk:
:+1:

PR #3578, opened 2014-12-02
Renames CanadaSSOClient.php to dosomething_canada_sso_client.inc.

sergiitk:
Drupal autoloading classes [must have `.inc` extension](https://www.drupal.org/node/608152#d7-autoloading).
Also it's uncommon to use camel case file naming in Drupal inc files.


aaronschachter:
:+1:

PR #3582, opened 2014-12-02
Introduces common DosomethingUserRemote abstract class.

sergiitk:
#### What's this PR do?
- Splits `dosomething_uk_sso_user.inc` to fields common and UK-specific fields
- Creates new abstract class of a remote user in `dosomething_user` to extend other remote users from (e.g. Canada)
#### Testing

screen shot 2014-12-03 at 12 40 55 am
image


lib/modules/dosomething/dosomething_uk/includes/dosomething_uk_register_controller.inc
```diff
@@ -74,6 +68,13 @@ class DosomethingUkRegisterController implements ExternalAuthRegisterController
       );
     }
 
+    // Determines whether the user agreed to receive e-mail newsletter.
+    if (!empty($values['field_opt_in_email'][LANGUAGE_NONE][0]['value'])) {
+      $user_data['contactable'] = TRUE;
```
aaronschachter, line 73: A user potentially could opt-out of email, and/or SMS.  Would it make sense to store this as just `opt_in_email` to keep it consistent?

lib/modules/dosomething/dosomething_uk/includes/dosomething_uk_register_controller.inc
```diff
@@ -63,7 +57,7 @@ class DosomethingUkRegisterController implements ExternalAuthRegisterController
     // @todo: use normal value when the international phone validator is ready.
     if (!empty($form_state['input']['field_mobile'][LANGUAGE_NONE][0]['value'])) {
       $phone = $form_state['input']['field_mobile'][LANGUAGE_NONE][0]['value'];
-      $this->remote_account->setPhoneNumber($phone);
+      $user_data['phone_number'] = $phone;
```
aaronschachter, line 60: Any reason to rename as `'phone_number'`vs keeping as `mobile` to keep things consistent?

lib/modules/dosomething/dosomething_uk/includes/dosomething_uk_sso_user.inc
```diff
@@ -152,43 +82,51 @@ class DosomethingUkSsoUser {
...
-      $data['dob']
+      $data['first_name'],
+      $data['dob'],
+      (isset($data['phone_number']) ? $data['phone_number'] : FALSE)
     );
+
+    // Set UK-specific fields.
+    $instance->last_name   = $data['last_name'];
+    $instance->postcode    = $data['postcode'];
+    $instance->screen_name = strtolower($instance->getFirstName()) . time();
```
aaronschachter: Potentially call this username? I'm just not crazy about having properties that seem to match with the Drupal User, but are named differently... but definitely open to hear the reasoning...

lib/modules/dosomething/dosomething_uk/includes/dosomething_uk_register_controller.inc
sergiitk, line 73: 1. SMS switch is still not implemented from the UK side
2. `contactable` Is actual name of field in the remote user. The idea is to represent the structure of the remote user.
   image

aaronschachter:
Code looks okay, just questions around naming conventions.

lib/modules/dosomething/dosomething_uk/includes/dosomething_uk_register_controller.inc
sergiitk, line 60: Same as in `contactable`, `phone_number` is the actual name of the remote field.

sergiitk:
Discussed with @aaronschachter: revisit common fields in `DosomethingUserRemote`.

sergiitk:
image
image

@aaronschachter could you please re-review?

aaronschachter:
:+1:

Issue #3449, opened 2014-11-18
dosomething_search_index view mode causes overridden Features


aaronschachter:
Need to Features Update  (`drush fu`) for numerous features which are appearing overridden, to include the `dosomething_search_index` view mode added in #3359 

![screen shot 2014-11-18 at 8 20 07 am](https://cloud.githubusercontent.com/assets/1236811/5087922/c2cc9288-6efb-11e4-8a54-a758816b7fbc.png)

aaronschachter:
#4766

sergiitk:
Thanks @aaronschachter.

PR #3585, opened 2014-12-03
Taxonomy Term campaigns view sort

aaronschachter:
Displays all active campaigns first, before any closed campaigns.

Also adds the `dosomething_search_index` view mode to DS Campaign per #3449.


sergiitk:
:+1:

PR #3586, opened 2014-12-03
Link images to content in recommended campaign galleries.

### LINKED ISSUE #3580: Images on the recommended campaigns block on 404 page and no search results aren't clickable
These should click through to the appropriate campaign:

image

image

sbsmith86:
Fixes #3580 . Linking recommended campaign gallery images to the content.


aaronschachter:
This is where this starts to get a little tricky.  The title is a link, which makes that clickable.  Does it make sense that if a URL is set, the whole entire block is just a link? cc @mikefantini

aaronschachter:
Code looks ok, i think we just need to clairify whether or not the entire block should just be a link.

mikefantini:
so just to clarify, if there is a link for the triad gallery, both the image and the title should be linked @sbsmith86 @aaronschachter

aaronschachter:
:+1:

PR #3587, opened 2014-12-03
Rescoping header CSS on not found pages

### LINKED ISSUE #3536: BUG: Styling on 404 headers broken on node pages.
Fix issue introduced when I scoped 404 styling to the .page-notfound class. This class is not added to node pages, so if you hit a notfound page node directly, it breaks some styles.

sbsmith86:
If you hit a node url directly it does not use the .page-notfound class so we can't scope all the css to these pages. This fixes #3536 

@DFurnes @weerd


weerd:
Sounds good!

PR #3588, opened 2014-12-03
Get variables for campaign groups in search

### LINKED ISSUE #3584: search results display for campaign groups
when a campaign group shows up in search results we should pull in:
- cover image
- title
- call to action

currently we are only pulling in the title

ex. when I search Halo and Tune In and Take Action comes up it is missing this key info

image

sbsmith86:
Fixes #3584


aaronschachter:
:+1: epic win value added

PR #3592, opened 2014-12-04
Wrapping up image swapping script for high res feature tile images.

### LINKED ISSUE #3267: featured campaign on collections is super pixelated

weerd:
Resolves #3267

@DFurnes @sbsmith86 @tjrosario 

Dave, took a slightly different approach than needing to remove the data role after image swap. Instead, saving all feature tiles to an array; if array empty, then images swapped, otherwise they've already been swapped and added to the "completed" array and thus no action required :)

All changed `$galleries` var to `$mosaicGalleries` so it's more descriptive of the actual jQuery collection obtained.

CC: @aaronschachter @barryclark

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/bower.json
+++ b/lib/themes/dosomething/paraneue_dosomething/bower.json
@@ -20,7 +20,7 @@
     "tests"
   ],
   "dependencies": {
-    "neue": "DoSomething/neue#~5.3.3",
+    "neue": "DoSomething/neue#~5.3.4",
     "jquery": "1.8.3",
     "mailcheck": "~1.0.3",
     "html5shiv": "~3.7.0",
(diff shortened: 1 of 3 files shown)


sbsmith86:
This looks good to me, im still getting familiar with prototype stuff, but I don't see anything glaring.

DFurnes:
Looks great! :ship:

weerd:
Sweet! Also updated the version of Neue required to 5.3.4 :)

PR #3593, opened 2014-12-04
Canada: Provides integration test for the phone field.

sergiitk:
#### What's this PR do?
- Provides integration test for the phone field
- Fixes phone field import on remote user syndication during login
#### Tests result

image
image
image


DFurnes:
Looks good!

Issue #3597, opened 2014-12-05
Reportback Files: Add uid_reviewer and reviewed timestamp


aaronschachter:
Not critical, but would be useful to store who reviewed each image and at what time.
- `review_uid`
- `review_timestamp`
- `review_source` -- could be the path of page, or `ios` for the iOS RB Reviewer app

Issue #3564, opened 2014-12-01
Reportback Flag form


aaronschachter:
- [x] Add the "Flag reportback" form functionality, which sets Flagged Reason, and whether or not to delete files, to the new Reportback admin view.
- [x] Store Flagged and Flagged Reasons on Reportback if flagged
- [x] Delete files if "Delete" selected

PR #3608, opened 2014-12-09
ReportbackFile Entity

aaronschachter:
- Adds ReportbackFileEntity class to facilitate issues like #3564 and #3597
- Fixes access callback for `admin/users/rb`

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -26,6 +26,20 @@ function dosomething_reportback_entity_info() {
     'module' => 'dosomething_reportback',
     'views controller class' => 'EntityDefaultViewsController',
   );
+  $info['reportback_file'] = array(
+    'label' => t('Reportback File'),
+    'base table' => 'dosomething_reportback_file',
+    'entity keys' => array(
+      'id' => 'fid',
+      'label' => 'fid',
+    ),
+    'entity class' => 'ReportbackFileEntity',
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -237,15 +251,22 @@ function dosomething_reportback_review_page_access($entity) {
   if (!user_access('view any reportback')) {
     return FALSE;
   }
+  // If no entity was passed, that's fine -- we'll display all reportbacks.
+  if ($entity == NULL) {
```
sergiitk, line 255: Did you mean `is_null()` or `!empty()`?

aaronschachter, line 255: Hmm yeah, i'm checking for `is_null`.  change?

sergiitk, line 255: Unless you want also meant a case when `$entity = FALSE` because `FALSE == NULL`:

``` php
php > var_dump(FALSE == NULL);
bool(true)
```

aaronschachter, line 255: Ehh, I mean.. i'm ok to leave it :) I'm specifically passing `NULL` as the argument for the `$entity` in `dosomething_reportback_menu`. and then Nodes and Taxonomy Terms for the other callbacks where this function is called.

sergiitk:
:+1:

PR #3614, opened 2014-12-09
Reportback Flag form - front end

aaronschachter:
Heaven help me I wrote jQuery today. :christmas_tree: 

Refs #3564.

Adds form elements for the Flag reportback form, which are only displayed when "flagged" is checked.

Doesn't actually write any values yet, as the first priority is to get a prototype up for campaigns team to test.

screen shot 2014-12-09 at 3 42 11 pm

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/js/reportback_review_form.js
+++ b/lib/modules/dosomething/dosomething_reportback/js/reportback_review_form.js
@@ -1,0 +1,21 @@
+(function ($) {
+  $(document).ready(function(){
+
+    // Hide all flag forms to start.
+    $('.flag-form').hide();
+
+    // Listen for changes to Flagged checkboxes.
+    $(':input[name*="status"]').click(function() {
+      // Find the next flagForm within the awesome Drupal markup.
+      var flagForm = $(this).parent().parent().parent().next('.flag-form');
+      // If "flagged" was checked:
+      if ($(this).val() == 'flagged') {
+        // Display the flagForm.
+        flagForm.show();
+      }
+      else {
+        flagForm.hide();
+      }
+    });
+  });
+}(jQuery));
--- a/lib/modules/dosomething/dosomething_reportback/js/reportback_review_form.js
+++ b/lib/modules/dosomething/dosomething_reportback/js/reportback_review_form.js
@@ -4,10 +4,10 @@
     // Hide all flag forms to start.
     $('.flag-form').hide();
 
-    // Listen for changes to Flagged checkboxes.
-    $(':input[name*="status"]').click(function() {
-      // Find the next flagForm within the awesome Drupal markup.
-      var flagForm = $(this).parent().parent().parent().next('.flag-form');
+    // Listen for changes to Status checkboxes.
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_reportback/js/reportback_review_form.js
```diff
@@ -0,0 +1,21 @@
+(function ($) {
+  $(document).ready(function(){
+
+    // Hide all flag forms to start.
+    $('.flag-form').hide();
+
+    // Listen for changes to Flagged checkboxes.
+    $(':input[name*="status"]').click(function() {
```
sergiitk: Is this possible to select these using classname instead?

lib/modules/dosomething/dosomething_reportback/js/reportback_review_form.js
```diff
@@ -0,0 +1,21 @@
...
+      // Find the next flagForm within the awesome Drupal markup.
+      var flagForm = $(this).parent().parent().parent().next('.flag-form');
```
sergiitk: Maybe better to provide the selector argument to `parent()`?

lib/modules/dosomething/dosomething_reportback/js/reportback_review_form.js
aaronschachter: It's pretty nasty code, since I'm using the `$form['rb_files']['#tree'] = TRUE;`, each id and name is uniqiue.  Here's what the markup looks like:

```
<input type="radio" id="edit-rb-files-1462-status-approved" name="rb_files[1462][status]" value="approved" class="form-radio">
```

screen shot 2014-12-09 at 3 49 48 pm

lib/modules/dosomething/dosomething_reportback/js/reportback_review_form.js
aaronschachter: I never really ever write JS... i'm not sure what you mean by this?

lib/modules/dosomething/dosomething_reportback/js/reportback_review_form.js
sergiitk: Still, `.form-radio` selector with a filter afterwards will makes search much faster.

aaronschachter: Ok. like this?

```
$('.form-radio:input[name*="status"]').click(function() {
```

Seems to work locally...

sergiitk: Yes, great!

lib/modules/dosomething/dosomething_reportback/js/reportback_review_form.js
sergiitk: jQuery [`parent()`](http://api.jquery.com/parent/) function accepts selector argument, for example:

``` js
$( "p" ).parent( ".selected" )
```

sergiitk:
:+1:

PR #3620, opened 2014-12-10
Member index

### LINKED ISSUE #3603: DS API: Retrieve User uid by mobile
Should only accessible by a logged in admin user.

Unfortunately the built-in Drupal `user` resource does not allow querying by fields attached to a User, which is super annoying.  We'll need to build our own `index` operation under our custom `users` resource and allow querying by mobile.

aaronschachter:
Fixes #3603: Allows an logged in administrator to query by email or mobile via DS API.

Also adds an access argument to the `_member_resource_access` function, to restrict usage to admin's. 

Returns array of User objects that match the parameters, but only returns their uid.  The parameters query is required.

Usage:

**GET** `http://dev.dosomething.org:8888/api/v1/users.json?parameters[mobile]=2125551000`

Response:

```
[
    {
        "uid": "5"
    }
]
```

If there are no Users that match, an empty array is returned.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/dosomething_api.info
+++ b/lib/modules/dosomething/dosomething_api/dosomething_api.info
@@ -6,4 +6,4 @@ dependencies[] = services
 features[ctools][] = services:services:3
 features[features_api][] = api:2
 features[services_endpoint][] = drupalapi
-mtime = 1417467369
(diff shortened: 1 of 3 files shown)


sergiitk:
:+1:

PR #3622, opened 2014-12-10
Reportback cropping updates

sbsmith86:
I made some style updates to the cropping interface. 
- Changed coloring
- Styled cursor to indicated moving or resizing
- Changed the initial and minimum size of the box to be half the height of the image. 
  
  @weerd @DFurnes


weerd:
:+1:

PR #3624, opened 2014-12-10
Fixes #3594

### LINKED ISSUE #3594: 404 Search results repeating words
For @deadlybutter:

If a url is something like dosomething.org/facts/facts-about-nothing it will pull "facts" out twice for the search. Preferably would limit it to just one of each unique word.

screen shot 2014-12-05 at 11 40 56 am

itsjoekent:
Fixes #3594 

Prevents duplicate search terms in the 404 page (eg: /facts/facts-dogs now only returns facts dogs)

@aaronschachter


aaronschachter:
NICE PHP :+1: Value added

itsjoekent:
OH YEAH

PR #3625, opened 2014-12-10
Added check to 404 video poster

### LINKED ISSUE #3559: PHP Notice in Notfound tpl
Notice: Undefined variable: video_poster in include() (line 21 of /var/www/vagrant/lib/themes/dosomething/paraneue_dosomething/templates/notfound/node--notfound.tpl.php).

itsjoekent:
Verifies there is a poster image for the 404 video before printing it.

Fixes #3559 

@aaronschachter

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/notfound/node--notfound.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/notfound/node--notfound.tpl.php
@@ -18,7 +18,8 @@
       <?php print $subtitle ?>
     </div>
     <?php if (isset($video)): ?>
-      <video src=<?php print $video ?> autoplay="" loop="" poster=<?php if(isset($video_poster)){ print $video_poster; } ?> ></video>
+      <?php print $video_poster ?>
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/templates/notfound/node--notfound.tpl.php
```diff
@@ -18,7 +18,7 @@
       <?php print $subtitle ?>
     </div>
     <?php if (isset($video)): ?>
-      <video src=<?php print $video ?> autoplay="" loop="" poster=<?php print $video_poster ?> ></video>
+      <video src=<?php print $video ?> autoplay="" loop="" poster=<?php if(isset($video_poster)){ print $video_poster; } ?> ></video>
```
aaronschachter: Kinda picky, but does it make sense to only output poster= if the variable is set? Also should the value be in quotes? Same with the Video src?  I don't know how all this new HTML5 stuff works / what is best practice :older_man:

itsjoekent: fixinz

itsjoekent:
@aaronschachter updated the codez

aaronschachter:
:+1:

PR #3637, opened 2014-12-12
Setting up initial JS for Reportbacks

sbsmith86:
@weerd Here is my first pass at just setting up the js. I don't remember where we landed on using the prototype pattern for the Reportback.js, but feel free to refactor!

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/campaign/ImageUploadBeta.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/campaign/ImageUploadBeta.js
@@ -1,0 +1,51 @@
+define(function(require) {
+  "use strict";
+
+  var $ = require("jquery");
+
+  /**
+   * Controls reportback behavior and interations.
+   */
+  var ImageUploadBeta = function() {
+    this.$uploadButton = $(".js-image-upload-beta");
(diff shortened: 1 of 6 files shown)


lib/themes/dosomething/paraneue_dosomething/js/campaign/ImageUploadBeta.js
```diff
@@ -0,0 +1,51 @@
...
+  };
+
+  ImageUploadBeta.prototype.init = function () {
+    this.imageCropInit();
+  };
+
+  /**
+   * Opens a modal after the user uploads a file. Initiates cropping.
+   */
+  ImageUploadBeta.prototype.imageCropInit = function () {
```
weerd, line 23: Not sure if you have other plans for this init, but seems like `init` and `imageCropInit` could be combined into just `init`? Not completely sure if you have specific plans as to why you separated them :)

weerd:
Looking good :D :+1:

lib/themes/dosomething/paraneue_dosomething/js/campaign/ImageUploadBeta.js
sbsmith86, line 23: @weerd yea im thinking they don't need to be separate either.

PR #3639, opened 2014-12-15
Fixes #3617.

### LINKED ISSUE #3617: can't submit "not affiliated with school" on mobile
using the school finder, a user can't submit "Not affiliated with a school" when using their mobile device.

It seems as a user is able to submit the form on mobile if they actually select a school and the user seems to be able to submit "Not affiliated with a school" on desktop.

This is happening on both staging and beta

DFurnes:
# Changes
- Fixes #3617. Submit button was un-tappable when focus was triggered on Mobile Safari. By not programmatically focusing the submit button, we avoid that issue.


weerd:
:tropical_drink: :+1:

PR #3640, opened 2014-12-16
Issue3523 mbp international settings

### LINKED ISSUE #3523: Transaction settings in dosomething_mbp
Add admin interface to support **default** settings for:
- Transaction settings by `$origin` / event type (member_signup, password_reset, campaign_signup, campaign_reportback)
  ex. `dosomething_mbp_member_signup_template`
  - template_name (dropdown selection of existing Mandrill templates)
    https://mandrillapp.com/api/docs/templates.JSON.html#method=list
  - filter by label: `message-broker`, `<affiliate code>`
  - *|SUBJECT|*
  - *|FROM_NAME|*
  - *|FROM_ADDRESS|*
    (list current Mandrill value, updates via Madrill API call) https://mandrillapp.com/api/docs/templates.JSON.html#method=update
  - `message_broker_producer_production` - `machine_name`

**Default `merge_var` values**
- *|FNAME|* ("Doer")
- Settings will be overwritten if the value exists in the `dosomething_mbp_request() $parms` call.
- [ ] Load default values via `hook_update_NN() for
  - member_signup
  - password_reset
  - campaign_signup
  - campaign_reportback

See:
https://github.com/DoSomething/dosomething/issues/3464
https://github.com/DoSomething/dosomething/issues/3464#issuecomment-63687017

deezone:
Fixes #3523 

Moves from hard coded template setting in transactional payloads to dynamic setting by "Application ID" (Affiliate country code) and available templates in Mandrill.
- Adds admin setting for "Application ID" (Affiliate country code) that's connected to the same setting in `message_broker_producer` module.
- Adds admin setting for Mandrill API key.
- Uses cURL call to Mandrill API to lookup templates by  "Application ID" (Affiliate country code).
- Adds admin interface to manage template assignment to email transaction types: `user_registration`, `user_password`, `campaign_signup`, `campaign_group_signup` and `campaign_reportback` by "Application ID" (Affiliate country code) .
- Uses template assignments to define 'template` value in message payloads replacing hard code / US specific settings.

**To test**:
- run `drush updb` to get system vars installed via `hook_install`
- go to `/admin/config/dosomething/dosomething_mbp`
  - Add "Application ID", ex: "CA". Currently only template entries for "CA" and "GB".
  - Add Mandrill Key setting
- Save form to be presented with "Transactional Template" settings for each transaction type
- Create test transaction to trigger transactional messages using selected template.

**Update**: Template entries in Mandrill for US, CA, UK and IN. Stilll need to add entries for the other affiliates. Switched from "GB" to "UK" as that's the official `county_code` value.

### DIFF
--- a/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.admin.inc
+++ b/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.admin.inc
@@ -58,22 +58,22 @@ function dosomething_mbp_config_form($form, &$form_state) {
       $form['dosomething_mbp_templates']['dosomething_mbp_current_application'] = array(
         '#markup' => $current_application_markup,
       );
-      $user_registration_template = variable_get('dosomething_mbp_user_registration_template', '');
-      $form['dosomething_mbp_templates']['dosomething_mbp_user_registration_template'] = array(
+      $user_register_template = variable_get('dosomething_mbp_user_register_template', '');
+      $form['dosomething_mbp_templates']['dosomething_mbp_user_register_template'] = array(
         '#type' => 'select',
         '#title' => t('User Registration'),
         '#options' => $mandrill_templates['select'],
-        '#default_value' => $user_registration_template,
+        '#default_value' => $user_register_template,
         '#description' => 'The Mandrill template to use for User Registration transactional messages.',
         '#required' => TRUE,
         '#weight' => 70,
       );
-      $password_reset_template = variable_get('dosomething_mbp_password_reset_template', '');
-      $form['dosomething_mbp_templates']['dosomething_mbp_password_reset_template'] = array(
+      $user_password_template = variable_get('dosomething_mbp_user_password_template', '');
+      $form['dosomething_mbp_templates']['dosomething_mbp_user_password_template'] = array(
         '#type' => 'select',
         '#title' => t('Password Reset'),
         '#options' => $mandrill_templates['select'],
-        '#default_value' => $password_reset_template,
+        '#default_value' => $user_password_template,
         '#description' => 'The Mandrill template to use for Password Reset transactional messages.',
         '#required' => TRUE,
         '#weight' => 80,
--- a/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.admin.inc
+++ b/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.admin.inc
@@ -64,11 +64,40 @@ function dosomething_mbp_config_form($form, &$form_state) {
         '#title' => t('User Registration'),
         '#options' => $mandrill_templates['select'],
         '#default_value' => $user_registration_template,
-        '#description' => 'The Mandrill template to use for User Registration.',
+        '#description' => 'The Mandrill template to use for User Registration transactional messages.',
         '#required' => TRUE,
-        '#weight' => 60,
+        '#weight' => 70,
       );
-
+      $password_reset_template = variable_get('dosomething_mbp_password_reset_template', '');
+      $form['dosomething_mbp_templates']['dosomething_mbp_password_reset_template'] = array(
+        '#type' => 'select',
+        '#title' => t('Password Reset'),
+        '#options' => $mandrill_templates['select'],
+        '#default_value' => $password_reset_template,
+        '#description' => 'The Mandrill template to use for Password Reset transactional messages.',
+        '#required' => TRUE,
+        '#weight' => 80,
+      );
+      $campaign_signup_template = variable_get('dosomething_mbp_campaign_signup_template', '');
+      $form['dosomething_mbp_templates']['dosomething_mbp_campaign_signup_template'] = array(
+        '#type' => 'select',
+        '#title' => t('Campaign Signup'),
+        '#options' => $mandrill_templates['select'],
+        '#default_value' => $campaign_signup_template,
+        '#description' => 'The Mandrill template to use for Campaign Signup transactional messages.',
+        '#required' => TRUE,
+        '#weight' => 90,
+      );
+      $campaign_reportback_template = variable_get('dosomething_mbp_campaign_reportback_template', '');
+      $form['dosomething_mbp_templates']['dosomething_mbp_campaign_reportback_template'] = array(
+        '#type' => 'select',
+        '#title' => t('Campaign Reportback'),
+        '#options' => $mandrill_templates['select'],
+        '#default_value' => $campaign_reportback_template,
+        '#description' => 'The Mandrill template to use for Campaign Reportback transactional messages.',
+        '#required' => TRUE,
+        '#weight' => 100,
+      );
     }
 
   }
--- a/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.admin.inc
+++ b/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.admin.inc
@@ -58,51 +58,56 @@ function dosomething_mbp_config_form($form, &$form_state) {
       $form['dosomething_mbp_templates']['dosomething_mbp_current_application'] = array(
         '#markup' => $current_application_markup,
       );
+      $template_var = 'dosomething_mbp_user_register_template_' . variable_get('message_broker_producer_application_id');
       $form['dosomething_mbp_templates']['dosomething_mbp_user_register_template'] = array(
         '#type' => 'select',
         '#title' => t('User Registration'),
         '#empty_option' =>t('-- Select Mandrill Template --'),
         '#options' => $mandrill_templates['select'],
-        '#default_value' => variable_get('dosomething_mbp_user_register_template') . '_' . variable_get('message_broker_producer_application_id'),
+        '#default_value' => variable_get($template_var),
         '#description' => 'The Mandrill template to use for User Registration transactional messages.',
         '#required' => TRUE,
         '#weight' => 70,
       );
+      $template_var = 'dosomething_mbp_user_password_template_' . variable_get('message_broker_producer_application_id');
       $form['dosomething_mbp_templates']['dosomething_mbp_user_password_template'] = array(
         '#type' => 'select',
         '#title' => t('Password Reset'),
         '#options' => $mandrill_templates['select'],
-        '#default_value' => variable_get('dosomething_mbp_user_password_template', '') . '_' . variable_get('message_broker_producer_application_id'),
+        '#default_value' => variable_get($template_var),
         '#description' => 'The Mandrill template to use for Password Reset transactional messages.',
         '#required' => TRUE,
         '#weight' => 80,
       );
+      $template_var = 'dosomething_mbp_campaign_signup_template_' . variable_get('message_broker_producer_application_id');
       $form['dosomething_mbp_templates']['dosomething_mbp_campaign_signup_template'] = array(
         '#type' => 'select',
         '#title' => t('Campaign Signup'),
         '#empty_option' =>t('-- Select Mandrill Template --'),
         '#options' => $mandrill_templates['select'],
-        '#default_value' => variable_get('dosomething_mbp_campaign_signup_template') . '_' . variable_get('message_broker_producer_application_id'),
+        '#default_value' => variable_get($template_var),
         '#description' => 'The Mandrill template to use for Campaign Signup transactional messages.',
         '#required' => TRUE,
         '#weight' => 90,
       );
+      $template_var = 'dosomething_mbp_campaign_group_signup_template_' . variable_get('message_broker_producer_application_id');
       $form['dosomething_mbp_templates']['dosomething_mbp_campaign_group_signup_template'] = array(
         '#type' => 'select',
         '#title' => t('Campaign Group Signup'),
         '#empty_option' =>t('-- Select Mandrill Template --'),
         '#options' => $mandrill_templates['select'],
-        '#default_value' => variable_get('dosomething_mbp_campaign_group_signup_template', '') . '_' . variable_get('message_broker_producer_application_id'),
+        '#default_value' => variable_get($template_var),
         '#description' => 'The Mandrill template to use for Campaign Group Signup transactional messages.',
         '#required' => TRUE,
         '#weight' => 90,
       );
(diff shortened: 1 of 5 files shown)


aaronschachter:
@DeeZone There is a function @sergii-tkachenko added, [`dosomething_settings_get_affiliate_country_code`](https://github.com/DoSomething/dosomething/blob/v0.3.147/lib/modules/dosomething/dosomething_settings/dosomething_settings.module#L74), which already returns a two character country code. Can we use this instead to avoid repetition / potential errors?

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.admin.inc
```diff
@@ -9,11 +9,208 @@
...
+      $current_application_markup = t('Current Application ID: "') . variable_get('message_broker_producer_application_id', '') . t('" as defined in the message_broker_producer') . l(t('Producer Configuration settings.'), 'admin/config/services/message-broker-producer/mq-settings') . t(' Producer Configuration settings or above in the configuration settings.');
+      $form['dosomething_mbp_templates']['dosomething_mbp_current_application'] = array(
+        '#markup' => $current_application_markup,
+      );
+      $user_register_template = variable_get('dosomething_mbp_user_register_template', '');
+      $form['dosomething_mbp_templates']['dosomething_mbp_user_register_template'] = array(
+        '#type' => 'select',
+        '#title' => t('User Registration'),
+        '#options' => $mandrill_templates['select'],
+        '#default_value' => $user_register_template,
```
aaronschachter: Because this is the only place you use this variable, I think it'd be a bit more readable to just set the `#default_value` to `variable_get('dosomething_mbp_user_register_template', '');`.  

I know this format is used in other forms.. but in those forms, we set the variable value to just the string, `dosomething_mbp_user_register_template`, to be used by the form ID and the variable name.

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.admin.inc
```diff
@@ -9,11 +9,208 @@
...
   return system_settings_form($form);
 }
+
+/**
+ * Format template data for presentation based on returned values from Mandrill.
+ *
+ * @parm string $application_ID
+ *   Defined in message_broker_producer, a unique identifier of the site.
+ */
+function dosomething_mbp_templates($application_ID) {
```
aaronschachter: Can we rename this to be more descriptive?  Perhaps something like `dosomething_mbp_get_mandrill_template_options` ?  As is, it looks like its' implementing a non-existent `hook_templates`.

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.admin.inc
```diff
@@ -9,11 +9,208 @@
...
+
+  $mandrill_templates['select'][0] = '-- Select Mandrill Template --';
```
aaronschachter: The better option here is to implement `"#empty_option"` within your form array: http://drupal.stackexchange.com/questions/13958/empty-value-select-does-not-appear-for-select-fields-with-states-requireme

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.admin.inc
```diff
@@ -9,11 +9,208 @@
...
+        'name' => $mandill_template->name,
+        'subject' => $mandill_template->subject,
+        'from_email' => $mandill_template->from_email,
+        'from_name' => $mandill_template->from_name,
+        'updated_at' => $mandill_template->updated_at,
+      );
+    }
+  }
+  else {
+    drupal_set_message('dosomething_mbp_get_template_data - No Mandrell templates found.', 'error');
```
aaronschachter: Spelling.

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.admin.inc
```diff
@@ -9,11 +9,208 @@
...
+}
+
+/**
+ * Gather Mandrill template details with the label specific to the affiliate
+ * (target_ID).
+ *
+ * @parm string $application_ID
+ *   Defined in message_broker_producer, a unique identifier of the site.
+ */
+function dosomething_mbp_get_template_data($application_ID) {
```
aaronschachter: Picky, but using the capitalized `ID` in the variable name is not very Drupal best practicey.  If we leave it in, it opens the door to more of it. Can we pls rename to `$application_id` ?

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.admin.inc
```diff
@@ -9,11 +9,208 @@
...
+    curl_setopt($ch, CURLOPT_HTTPHEADER, array('Content-Type: application/json','Content-Length: ' . strlen($json)));
+    $result = curl_exec($ch);
+    curl_close($ch);
+    $decoded = json_decode($result);
+  }
+  else {
+    drupal_set_message('dosomething_mbp_mandrill_curl - Mandrill key not set.', 'error');
+  }
+
+  return $decoded;
```
aaronschachter, line 219: If there are no results, `$decoded` will not be set and produce PHP notices/warnings.  Should be set to an empty array in the beginning of the function.

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
```diff
@@ -105,42 +105,43 @@ function dosomething_mbp_get_transactional_payload($origin, $params = NULL) {
...
   switch ($origin) {
     case 'user_register':
+      $payload['template'] = variable_get('dosomething_mbp_user_register_template');
       $payload['birthdate'] = $params['birthdate'];
       $payload['subscribed'] = 1;
       if (isset($params['mobile'])) {
         $payload['mobile'] = $params['mobile'];
       }
       $payload['merge_vars']['FNAME'] = $params['first_name'];
-      $payload['email_template'] = 'mb-general-signup';
```
aaronschachter, line 114: This seems a little dangerous..  Have you tested this code with transactionals from a clean staging DB?

Will this break things when the variables are not set (as will be the case when this code is deployed)? Do we need an install hook to set the relevant variables?

deezone:
@aaronschachter See `hook_install` addition. Tested with clean staging db, works including toggling setting when application_id / affiliate country code is changed..

**Issue**, the Mandrill key needs to be set outside of the files under version control. I'm guessing this goes in 'settings.php`? One for each of the affiliate sites as well as the main US site.

Suggestions?

deezone:
**Note to self**: Possibly need to add template settings for:
- Sender Name
- Sender email
- Subject

These values can be managed through the Mandrill interface as default values but could be administered in the Drupal admin interface as an override. Perhaps just list settings as read only via values returned in the curl call that collects the template name values.

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
```diff
@@ -108,3 +108,65 @@ function dosomething_mbp_update_7003(&$sandbox) {
...
+
+   variable_set('message_broker_producer_application_id', 'US');
+
+   variable_set('dosomething_mbp_user_register_template_US', 'mb-general-signup-us');
+   variable_set('dosomething_mbp_user_password_template_US', 'mb-password-reset-us');
+   variable_set('dosomething_mbp_campaign_signup_template_US', 'mb-campaign-signup-us');
+   variable_set('dosomething_mbp_campaign_reportback_template_US', 'mb-campaign-report-back-us');
+   variable_set('dosomething_mbp_campaign_group_signup_template_US', 'mb-group-campaign-signup-us');
+
+   variable_set('dosomething_mbp_user_register_template_UK', 'mb-general-signup-affiliate-uk');
```
aaronschachter: Why store all the variables for each in affiliate in all DB?  I get that its a workaround, but.. it's kind of messy.

Since we have to set the Mandrill key manually for each, maybe it makes sense to move this all into settings.php... let's discuss

deezone:
@aaronschachter Perhaps a less "messy" solution in `hook_install` that adds affiliate specific template settings.

deezone:
@aaronschachter Thoughts? ^^

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
```diff
@@ -108,3 +108,28 @@ function dosomething_mbp_update_7003(&$sandbox) {
   }
 
 }
+
+/**
+ * Creates production entry in message_broker_producer for user profile updates
+ */
+ function dosomething_mbp_update_7004(&$sandbox) {
+
+   $affiliate = variable_get('message_broker_producer_application_id', 'US');
```
aaronschachter: is this not the same as the country code?  DRY https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_settings/dosomething_settings.module#L74

aaronschachter:
Also just checking - this has been tested and is functional?  Does all one need to do is pull down this branch and run drush updb in order to test?

deezone:
Confirmed, after latest round of changes which includes your suggestion to use `dosomething_settings_get_affiliate_country_code()`:

Before update:

```
mysql> select * from variable where name like '%dosomething_mbp%';
+---------------------+-------+
| name                | value |
+---------------------+-------+
| dosomething_mbp_log | i:1;  |
+---------------------+-------+
1 row in set (0.00 sec)
```

after update:

```
mysql> select * from variable where name like '%dosomething_mbp%';
+------------------------------------------------+-------------------------------------+
| name                                           | value                               |
+------------------------------------------------+-------------------------------------+
| dosomething_mbp_campaign_group_signup_template | s:27:"mb-group-campaign-signup-us"; |
| dosomething_mbp_campaign_reportback_template   | s:26:"mb-campaign-report-back-us";  |
| dosomething_mbp_campaign_signup_template       | s:21:"mb-campaign-signup-us";       |
| dosomething_mbp_log                            | i:1;                                |
| dosomething_mbp_user_password_template         | s:20:"mb-password-reset-us";        |
| dosomething_mbp_user_register_template         | s:20:"mb-general-signup-us";        |
+------------------------------------------------+-------------------------------------+
```

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
deezone, line 114: `dosomething_mbp_update_7004()` will take care of setting the necessary variables based on the country code of the site - US and each of the affiliate sites.

Tested with a clean staging DB resulting in the expected template settings which would be different for each site depending on the country code system variable. 

Could use a test run with an affiliate staging db. Didn't have any luck with `ds pull stage.uk --db` -> `ERROR 1049 (42000): Unknown database 'dosomething_uk'`.

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
```diff
@@ -108,3 +108,45 @@ function dosomething_mbp_update_7003(&$sandbox) {
...
+   variable_set('message_broker_producer_rabbitmq_vhost', 'dosomething');
+
+   $affiliate = dosomething_settings_get_affiliate_country_code();
+   $affiliate_lc = strtolower($affiliate);
+
+   if ($affiliate == 'US') {
+     variable_set('dosomething_mbp_user_register_template', 'mb-general-signup-us');
+     variable_set('dosomething_mbp_user_password_template', 'mb-password-reset-us');
+     variable_set('dosomething_mbp_campaign_signup_template', 'mb-campaign-signup-us');
+     variable_set('dosomething_mbp_campaign_reportback_template', 'mb-campaign-report-back-us');
```
aaronschachter, line 136: To keep this consistent within the codebase, I'd vote that we name this template as `reportback` and not `report-back` if possible

aaronschachter:
Reviewed with @sergii-tkachenko and @DeeZone.

Approach will be to avoid using variable names for the Mandrill tempaltes to use, and instead hardcode templates per event / country code like so:

```
/**
* Fill this out with real comments
 *
@param string $event_name
  Could be "campaign_signup", "password_reset", etc
*
*/
function dosomething_mbp_get_template_name($event_name) {
  $country_code = dosomething_settings_get_affiliate_country_code();
  return 'mb-' . $event_name . '-' . $country_code;
}
```

Advantages of this is that there are less variables / code to maintain, and we can easily swap out templates by renaming them in Mandrill.

aaronschachter:
NOTE: Before deployment to production itnl, we will need to manually set the Rabbit MQ password `message_broker_producer` variable, in order to avoid errors.  

**We should test this process on Intl QA before deploying!!!**

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.admin.inc
```diff
@@ -9,11 +9,212 @@
...
   return system_settings_form($form);
 }
+
+/**
+ * Format template data for presentation based on returned values from Mandrill.
+ *
+ * @parm string $application_id
+ *   Defined in message_broker_producer, a unique identifier of the site.
+ */
+function dosomething_mbp_get_mandrill_template_options($application_id) {
```
aaronschachter, line 128: We're also going to lose this function per https://github.com/DoSomething/dosomething/pull/3640#issuecomment-70723782

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.admin.inc
```diff
@@ -9,11 +9,212 @@
...
+}
+
+/**
+ * Gather Mandrill template details with the label specific to the affiliate
+ * (target_ID).
+ *
+ * @parm string $application_id
+ *   Defined in message_broker_producer, a unique identifier of the site.
+ */
+function dosomething_mbp_get_template_data($application_id) {
```
aaronschachter, line 151: Also getting rid of this per https://github.com/DoSomething/dosomething/pull/3640#issuecomment-70723782

aaronschachter:
Related, we no longer need the `application_id` variable, as `dosomething_settings_get_affiliate_country_code` provides this value

deezone:
Replaced by:
https://github.com/DoSomething/dosomething/pull/3763

PR #3643, opened 2014-12-16
Removed second video poster print

itsjoekent:
Fixes this bug: http://staging.beta.dosomething.org/admin/reports/event/592622
(Don't know of any Github issues for it)

@aaronschachter


aaronschachter:
:+1:

PR #3649, opened 2014-12-17
Adds prove it buttons to the figure template

### LINKED ISSUE #3376: User Profile: I'm Doing buttons
Add "Prove It" buttons to each Figure for the "I'm Doing" section.

@weerd suggested that these be added as a separate `<div>` instead of including within the `description`.  @weerd -- are there Neue updates needed for this?

itsjoekent:
Fixes #3376 

@aaronschachter

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
@@ -549,6 +549,7 @@ function paraneue_dosomething_preprocess_user_profile(&$vars) {
         'title' => $value['link'],
         'image' => $image,
         'description' => $value['call_to_action'],
+        'prove_it' => '../' . $value['path_alias'] . '#prove',
       );
       $doing_items[$delta] = paraneue_dosomething_get_gallery_item($item, 'figure');
     }
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
@@ -550,7 +550,8 @@ function paraneue_dosomething_preprocess_user_profile(&$vars) {
         'title' => $value['link'],
         'image' => $image,
         'description' => $value['call_to_action'],
-        'link' => $base_url . '/' . $value['path_alias'],
+        'url' => $base_url . '/' . $value['path_alias'] . '#prove',
+        'button_text' => t("Prove It"),
       );
       $doing_items[$delta] = paraneue_dosomething_get_gallery_item($item, 'figure');
     }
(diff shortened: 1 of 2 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
```diff
@@ -549,6 +549,7 @@ function paraneue_dosomething_preprocess_user_profile(&$vars) {
         'title' => $value['link'],
         'image' => $image,
         'description' => $value['call_to_action'],
+        'prove_it' => '../' . $value['path_alias'] . '#prove',
```
aaronschachter: A user profile is visible on both `/user/` and `/user/[uid]`.. this link will break in one of those scenarios.  We can use the `l` function with a [fragment](https://api.drupal.org/comment/12934#comment-12934) or use [`global $base_url`](https://api.drupal.org/api/drupal/developer!globals.php/global/base_url/7) to create the full path.

I can also see us having calling this something like `footer` vs `prove_it`, where the `footer` is set to the entire link.  So other galleries that use this template could put other things into the footer section. Does that make sense?

itsjoekent: Point 1 - Ah, that would be a problem. I'll just swap in the base URL and append the campaign url

Point 2 - Totally get what you're saying, just not sure footer is the right word. Maybe `campaign_url` ?

itsjoekent:
@aaronschachter updated it, thoughts?

lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
```diff
@@ -549,12 +550,12 @@ function paraneue_dosomething_preprocess_user_profile(&$vars) {
         'title' => $value['link'],
         'image' => $image,
         'description' => $value['call_to_action'],
+        'link' => $base_url . '/' . $value['path_alias'],
```
aaronschachter: Let's set a `button_text` variable that displays what copy to display in the button.  this should be

```
function paraneue_dosomething_preprocess_user_profile(&$vars) {
  if ($doing = $vars['doing']) {
    foreach ($doing as $delta => $value) {
      $image = dosomething_image_get_themed_image($value['nid_image'], 'landscape', '300x300');
      $item = array(
        'url' => $value['path_alias'],
        'title' => $value['link'],
        'image' => $image,
        'description' => $value['call_to_action'],
        'button_text' => t("Prove It"),
      );
```

aaronschachter: We'll change the `url` to contain full path, with the `#prove` fragment

aaronschachter:
:+1:

PR #3651, opened 2014-12-17
Flagged and flagged_reason columns

aaronschachter:
Refs #3564

Adds `flagged` and `flagged_reason` columns to the `dosomething_reportback` table.  Writes the values upon setting a Reportback File to flagged.

This ditches the use of the Flag module for flagging Reportbacks, moving to a simpler custom implementation.


sergiitk:
:+1:

PR #3653, opened 2014-12-18
Delete Reportback Images

aaronschachter:
Refs #3564 - allows staff to delete a Reportback image from the new review form.  This is necessary for when users upload things we don't want hosted on our server.

Also cleans up the Reportback Files and related `file_managed` Files upon deleting a Reportback.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback_file.inc
+++ b/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback_file.inc
@@ -58,7 +58,7 @@ class ReportbackFileEntity extends Entity {
     if ($this->status === 'flagged') {
       $reportback = reportback_load($this->rbid);
       $reportback->flag($values['flagged_reason']);
-      if ($values['delete'] == 1) {
+      if ($values['delete']) {
         $this->deleteFile();
       }
     }
--- a/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback_file.inc
+++ b/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback_file.inc
@@ -71,7 +71,9 @@ class ReportbackFileEntity extends Entity {
    */
   public function deleteFile() {
     $file = file_load($this->fid);
-    return file_delete($file);
+    if ($file) {
+      return file_delete($file);
+    }
   }
 }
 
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback_file.inc
```diff
@@ -50,10 +58,23 @@ class ReportbackFileEntity extends Entity {
     if ($this->status === 'flagged') {
       $reportback = reportback_load($this->rbid);
       $reportback->flag($values['flagged_reason']);
+      if ($values['delete'] == 1) {
```
sergiitk: `if ($values['delete'] == 1)` is exactly the same as `if ($values['delete'])`.
Did you mean `if ($values['delete'] === 1)`?

lib/modules/dosomething/dosomething_reportback/js/reportback_review_form.js
```diff
@@ -17,5 +17,15 @@
         flagForm.hide();
       }
     });
+
+    // Listen for changes to Delete checkboxes.
+    $(':input[name*="delete"]').click(function() {
+
+      // If "flagged" was checked:
+      if ($(this).is(':checked')) {
+        alert(Drupal.t("Deleting this image cannot be undone.\n\nOnly do this if you really, really mean it!"));
```
sergiitk, line 26: Maybe it would be better to use [`confirm()`](http://www.w3schools.com/jsref/met_win_confirm.asp)?

lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback_file.inc
aaronschachter: Hmm. The form will send a 0 if not. Although this function will be used by the API too. Would you agree its best practice to specifically check for whether or not the data sent contains a value of 1 sent for delete?

sergiitk: I think simple `if ($values['delete'])` is enough.

lib/modules/dosomething/dosomething_reportback/js/reportback_review_form.js
aaronschachter, line 26: Definitely. Will change

aaronschachter, line 26: Discussed with @sergii-tkachenko -- since the Cancel button doesn't actually uncheck the checkbox yet, going to keep as `alert` for now.

sergiitk, line 26: @aaronschachter and I agreed to assign the confirm dialog to the submit button in future and leave this alert for now.

sergiitk:
:+1:

lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback_file.inc
mshmsh5000: `if ($values['delete'])` will be true if the passed value is a string. `if ($values['delete'] == 1)` will be false for a string, unless the string is `"1"`. That's the only time `== 1` is worth specifying. Truthiness!

PR #3656, opened 2014-12-18
Reportback File teaser view mode

aaronschachter:
Displays a Reportback File teaser on the Reportback entity view.

Adds checks to see if a File exists before returning images.


sergiitk:
:+1: Cool

PR #3659, opened 2014-12-18
Edit Reportback Review

aaronschachter:
Displays the Reportback Review form on the Reportback view (e.g. `reportback/123`)

Allows for easy editing of a specific Reportback. Doesn't display the quantity / why participated for each Reportback File, since its the same for each.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
@@ -101,8 +101,13 @@ function dosomething_reportback_files_form($form, &$form_state, $status = NULL, 
     'status' => $status,
   );
 
+  // Display mode for rendering each Reportback File.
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
```diff
@@ -100,12 +100,28 @@ function dosomething_reportback_files_form($form, &$form_state, $status = NULL,
   $params = array(
     'status' => $status,
   );
-  if (isset($entity->nid)) {
+
+  // Display mode for rendering each Reportback File.
+  $view_mode = 'full';
+
+  // If we are viewing a reportback entity:
+  if (isset($entity->rbid) && arg(0) == 'reportback') {
```
sergiitk: Just in case, it's better to use `===`.

sergiitk:
Looks good, one little comment.

PR #3660, opened 2014-12-18
User Profile - Reportback permalinks

aaronschachter:
Links the Impact description to the Reportback `reportback/[rbid]` if the logged in user has access to `view any reportback`.


sergiitk:
:+1:

PR #3669, opened 2014-12-19
Fixes bug in Reportback Files review endpoint

aaronschachter:
Missed this when changing the function parameters in #3651.  Fixes the `reportback_files/[fid]/review` endpoint


sergiitk:
:+1:

PR #3670, opened 2014-12-19
Switching class placed on upload field based on theme setting

sbsmith86:
We are supporting both reportback experiences for the time being, so this puts the correct class on the image upload field to support the different js scripts. 

@DFurnes @aaronschachter


aaronschachter:
:+1:

PR #3671, opened 2014-12-19
Reportback File tweaks

aaronschachter:
- Don't display "count" on the Review form when viewing a single Reportback 
- Display total number of files uploaded in Reportback File

screen shot 2014-12-19 at 4 52 58 pm


sergiitk:
:+1:

PR #3673, opened 2014-12-22
Reportback modal cropping

sbsmith86:
*Links up the cropping functionality into the image upload modal and populates the main reportback form with the cropping coordinates.

*Moved cropping interface styles to the crop modal sass.

@weerd

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/campaign/ImageUploadBeta.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/campaign/ImageUploadBeta.js
@@ -17,6 +17,7 @@ define(function(require) {
   var ImageUploadBeta = function($button) {
     this.$uploadButton   = $button;
     this.$cropModal      = $("#modal--crop");
+    this.$reportbackForm = $("#dosomething-reportback-form");
     this.$imageForm      = this.$cropModal.find("#dosomething-reportback-image-form");
     this.$imagePreview   = this.$cropModal.find(".image-preview");
     this.imageValues     = {};
@@ -98,6 +99,7 @@ define(function(require) {
    */
   ImageUploadBeta.prototype.cropInit = function(files) {
     var ImageCrop = require("campaign/ImageCrop");
+    var _this = this;
 
     if (/^image/.test( files[0].type)) {
       var reader = new FileReader();
@@ -108,7 +110,7 @@ define(function(require) {
         origImage.src = this.result;
         // Remove any cropping already on the page before initiating.
         ImageCrop.dsCropperDestroy();
-        ImageCrop.dsCropperInit(origImage);
+        ImageCrop.dsCropperInit(origImage, _this.$reportbackForm);
       };
     }
   };
--- a/lib/themes/dosomething/paraneue_dosomething/js/campaign/ImageUploadBeta.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/campaign/ImageUploadBeta.js
@@ -114,23 +114,13 @@ define(function(require) {
   };
 
   /**
-   * Return an object containing the information gathered in the modal form.
+   * Returns the value of the caption field.
    * Should only be used after the form has been submitted and the modal has closed.
-   *
-   * Includes:
-   * - Caption
-   * - @TODO: x, y, width, height for cropping.
    *
    * @return {jQuery} An object of form values.
    */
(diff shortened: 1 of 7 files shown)


lib/themes/dosomething/paraneue_dosomething/js/campaign/ImageUploadBeta.js
```diff
@@ -76,22 +72,57 @@ define(function(require) {
...
+  /**
+   * Initiates cropping on a file.
    *
-   * @return {jQuery} An object of form values.
+   * @param {jQuery} files  An object representing a file.
    */
-  ImageUploadBeta.prototype.getImageValues = function() {
-    var caption = this.$imageForm.find("#caption").val();
+  ImageUploadBeta.prototype.cropInit = function(files) {
+    var ImageCrop = require("campaign/ImageCrop");
```
weerd, line 101: Is there a reason why this is _required_ here and not up top? Isn't it all getting compiled together anyhoo, so might as well just require it at top with everything else... unless I'm missing something?

sbsmith86, line 101: @weerd I figured we only needed it if cropping was enabled. Since this function only gets called when cropping is enabled, i put it here.

lib/themes/dosomething/paraneue_dosomething/scss/content/_modal-crop.scss
```diff
@@ -55,4 +55,52 @@
       max-width: 100%;
     }
   }
+
+  // Styles for cropping.
+  .image-crop-container {
+    position: relative;
+    display: inline-block;
+
+    .crop-box {
```
weerd: Not 100% sure but could you do the whole `.module > .wrapper` stuff we discussed on friday here?

lib/themes/dosomething/paraneue_dosomething/scss/content/_modal-crop.scss
```diff
@@ -55,4 +55,52 @@
...
+      }
+    }
+
+    .resize-handle-wrapper {
+      position: relative;
+      width: 100%;
+      height: 100%;
+    }
+
+    .resize-handle {
```
weerd, line 77: Also here with the `.resize-handle > .wrapper` :)

lib/themes/dosomething/paraneue_dosomething/scss/content/_modal-crop.scss
sbsmith86: @weerd sure i will try renaming some stuff. For this one it is interesting because it is actually acting like a container since the main purpose of this div is to create a bounding box for cropping. Maybe I can make it `.crop-bounding-box` or something.

lib/themes/dosomething/paraneue_dosomething/js/campaign/ImageUploadBeta.js
weerd, line 101: Cool. I think that makes sense. I saw you removed it from at top of the **app.js** file. I figured since this all gets combined into a minified file it gets added regardless and probably more readable at top. Although I get the require stuff confused at times, so maybe @DFurnes can shed some light, haha. Fine to merge for now with this cause I don't know what the answer is :)

sbsmith86, line 101: @weerd ok cool. Yea I like require because it allows you to load js when you actually need it instead of all the time. I could even possibly require the js and then assign a callback function that runs the code that uses it only after the file has actually been added to the page. This way you don't add more js then you ever need. Happy to hear other thoughts on this though!

weerd, line 101: Let's add it as is and if we need to change later no biggie :) It was more a question than an actual request, haha!

lib/themes/dosomething/paraneue_dosomething/scss/content/_modal-crop.scss
weerd: `.crop-bounding-box` makes more sense to me :) Nice name!

sbsmith86:
@weerd updated class names! Let me know what you think.

weerd:
Awesomeness :+1:

Issue #3652, opened 2014-12-17
Refactor Flagged Reportbacks


aaronschachter:
#3651 adds new columns to the Reportback entity to store flagged/flagged_reason that way, instead of the Flag module.
- [x] Update hook needs to check all existing `flagged_reportback` flags, and set the `flagged` and `flagged_reason` values accordingly
- [x] Update `dosomething_reportback_get_reportback_total_by_nid` accordingly

aaronschachter:
The Reportback flags have been migrated... but the flagged reason has not been.

aaronschachter:
`dosomething_reportback_update_7018` ended up joining the Reason on the wrong field (the Flagging id instead of the Reportback rbid).  This is a very good example of why we're removing the Flag functionality altogether, because the Flagging as a separate entity is a pain.

PR #3690, opened 2015-01-07
Update flagged Reportbacks

aaronschachter:
Refs #3652

Updates Reportbacks which have been flagged using the `flagged_reportback` Flag with the new way (properties on the Reportback and Reportback File entities).


sergiitk:
How many reportbacks are there to update?

aaronschachter:
871 reportbacks have been flagged, per the latest prod snapshot on QA.

`SELECT * FROM flagging WHERE fid =2`

sergiitk:
:+1:

PR #3691, opened 2015-01-07
Adds show more functionality to gallery

### LINKED ISSUE #3372: Paraneue Gallery: "Show more" functionality
The `paraneue_get_gallery` should support an option to specify whether or not a "Show more" link should display.

If set to display, the gallery should hide/display additional items within the template.

itsjoekent:
Fixes #3372

@aaronschachter 
- added show more parameter to gallery functions
- added data role to galleries which should show more
- updated taxonomy & user profiles to use this

@weerd 
- added data role to the template
- updated app.js to add revealer based on data role

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/app.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/app.js
@@ -46,6 +46,7 @@ define("app", function (require) {
 
 
     var $mosaicGalleries = $body.find(".gallery.-mosaic");
+    console.log($mosaicGalleries);
 
     /**
      * Show hide large galleries.
--- a/lib/themes/dosomething/paraneue_dosomething/js/app.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/app.js
@@ -44,18 +44,15 @@ define("app", function (require) {
       Donate.init();
     }
 
-
-    var $mosaicGalleries = $body.find(".gallery.-mosaic");
-    console.log($mosaicGalleries);
+    var $revealGalleries = $body.find("ul[data-role='show_more']");
 
     /**
      * Show hide large galleries.
-     * Only applies to taxonomy pages for the time being.
      */
-    if ($body.hasClass("page-taxonomy") && $mosaicGalleries.length > 0) {
+    if ($revealGalleries.length > 0) {
       var galleryList = [];
 
-      $mosaicGalleries.each(function (index, gallery) {
+      $revealGalleries.each(function (index, gallery) {
         var $gallery = $(gallery);
         var $tiles = $gallery.find("li");
 
@@ -65,6 +62,7 @@ define("app", function (require) {
       });
     }
 
+    var $mosaicGalleries = $body.find(".gallery.-mosaic");
 
     /**
      * Swap low resolution images for feature tiles in mosaic gallery
--- a/lib/themes/dosomething/paraneue_dosomething/js/app.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/app.js
@@ -44,8 +44,7 @@ define("app", function (require) {
       Donate.init();
     }
 
-    var $revealGalleries = $body.find("ul[data-role='show_more']");
-
+    var $revealGalleries = $body.find('[data-show-more="true"]');
     /**
      * Show hide large galleries.
      */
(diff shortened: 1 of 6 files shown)


aaronschachter:
How many show at a time?

itsjoekent:
5, that's what it already was in the JS

aaronschachter:
yeah, so that will be weird. i think it needs to change based on whatever gallery style is being rendered. @weerd any thoughts on best way to code this? per the gallery style name?

itsjoekent:
yeah, it was normally meant for mosaic which is 5, but on user profile it should be 6

lib/themes/dosomething/paraneue_dosomething/js/app.js
```diff
@@ -44,17 +44,15 @@ define("app", function (require) {
       Donate.init();
     }
 
-
-    var $mosaicGalleries = $body.find(".gallery.-mosaic");
+    var $revealGalleries = $body.find("ul[data-role='show_more']");
```
weerd: You probs don't need the `ul` on there. Would make it more generic to apply to more things if it's just the `[data-role='show_more']` :)

lib/themes/dosomething/paraneue_dosomething/templates/system/patterns/gallery.tpl.php
```diff
@@ -9,11 +9,12 @@
  * - $items: An array of items to display. Each $item in $items contains:
  *   - $item['class']: The class to apply to the li (string).
  *   - $item['content']: Themed item content (string).
+ * - $roles: Data roles to be applied (string).
  *
  * @see paraneue_dosomething_get_gallery()
  */
 ?>
-<ul class="gallery -<?php print $layout; ?> <?php print $classes; ?>">
+<ul class="gallery -<?php print $layout; ?> <?php print $classes; ?>" data-role="<?php print $roles; ?>">
```
weerd: @deadlybutter so this all looks good. Just a note with data roles... you typically name the "role" part something, and it's 1 value per role. So in the above, you shouldn't really add a list of roles to a data-role. For this scenario it could be something like `data-show-more="true"` or something like that... And then with jquery you can grab the role and check it. Grab it with one of jquery's helpers, like `$('some-container').data('show-more');  // will retrieve 'true'`

itsjoekent: gotcha, thanks!

itsjoekent:
@weerd could you also give your input on @aaronschachter question?

weerd:
Oh. Hmmm, maybe it could be a parameter that gets passed through; or check if it's a mosaic gallery we're dealing with then retrieve 5 otherwise 6. Could also be some sort of switch statement that sets the amount depending on the context? 

```
switch (galleryType) {
  case mosaic:
    count = 5;
    break;
  case other-type:
    count = 6;
    break;
  case yet-another-type:
    count = 3;
    break;
  default:
    count = 1000; // teehee, so many!
    break;
}
```

itsjoekent:
All updated @aaronschachter / @weerd

aaronschachter:
Seems ok by me. @weerd -- can you pls review these changes?

itsjoekent:
@aaronschachter @weerd all good now, i think

lib/themes/dosomething/paraneue_dosomething/js/app.js
```diff
@@ -44,26 +44,27 @@ define("app", function (require) {
...
+    if ($revealGalleries.length > 0) {
       var galleryList = [];
 
-      $mosaicGalleries.each(function (index, gallery) {
+      $revealGalleries.each(function (index, gallery) {
         var $gallery = $(gallery);
         var $tiles = $gallery.find("li");
+        var isMosaicGallery = $gallery.hasClass('-mosaic');
+        var initialItems = isMosaicGallery.hasClass('-mosaic') ? 5 : 6;
+        var stepItems = isMosaicGallery.hasClass('-mosaic') ? 8 : 6;
```
weerd: Gah! No need to check if isMosaicGallery has a class... that should be a boolean...

weerd:
:+1: :shipit:

weerd:
@DFurnes be aware that after you merge in your Neue changes, this will potentially break if `mosaic` gallery has been renamed to something else. Update the jquery collection query :)

PR #3693, opened 2015-01-08
Terms API endpoint

aaronschachter:
New endpoints needed for the [iOS Reviewer app](https://github.com/dosomething/ios-RBReviewer):
- **GET** `/api/v1/terms`: Currently returns all Cause terms (only vocabulary needed for the endpoint at this time)
- **GET** `/api/v1/terms/[tid]/inbox`: Returns Reportback Files with status `pending` for a given taxonomy term.  Only accessible for users with permission to `view any reportback`

Also:
- Adds `dosomething_api_get_reportback_files` function to DRY loading data for the `campaigns/[nid]/gallery` and `terms/[tid]/inbox` endpoints
- Returns `why_participated` and `uid` properties in the response if the user has access to `view any reportback`


sergiitk:
Looks good :+1:

Issue #3689, opened 2015-01-07
Reportback Flags cleanup


aaronschachter:
- [x] Remove all references to `flagged_reportback` Flag
- [x] Delete the Promoted flag:
  - [x] Remove the Promoted flag links from the current Reportback view
  - [x] Delete the Editorial Caption and Weight fields, update features
  - [x] Delete the Promoted flag, update features
  - [x] Delete the `node/*/gallery` page functionality

aaronschachter:
Locally the Promoted flag seemed to delete, but it still exists on staging and production.  I'm finding that the `field_image_description` has not been dropped from the Promoted flag however, which could potentially cause issues.

aaronschachter:
Promoted flag deleted per #3713

PR #3695, opened 2015-01-09
Deletes promoted flag

aaronschachter:
Refs #3689 - Removes now deprecated functionality for promoting Reportbacks.


sergiitk:
:+1:

PR #3696, opened 2015-01-09
Enables application/x-www-form-urlencoded request parsing

aaronschachter:
Fixes the logout POST in the Reviewer iOS app, which was resulting in a 406.  See http://drupal.stackexchange.com/questions/62349/post-request-trying-to-make-a-session-authentication-in-the-rest-server-return-a


sergiitk:
:+1:

PR #3698, opened 2015-01-09
Reportback script

### LINKED ISSUE #3633: Display promoted and approved Reportbacks in Campaign template
- [x] Add `reportback_files` into the Paraneue preprocess for the Campaign Action Page. Should display X # of Promoted files, and then Y # of Approved files after
- [x] Render each `reportback_file` with a tpl
- [x] Load more approved files with ajax request

### LINKED ISSUE #3663: Prove it gallery total count
The gallery should know how many total results there are, to not display "Show More" when there's actually nothing left to show or nothing to show at all.

Upcoming PR will introduce a `dosomething_reportback_get_reportback_files_query_count` function.

weerd:
Resolves #3633
Resolves #3663

@aaronschachter @sbsmith86 

@sbsmith86 this doesn't fully include merging in with your code. Not sure if I'll do that in this PR or do it in a quick separate one. Just FYI!

Some styling fixes still need to be addressed as well as some IE testing. However, the main functionality of loading more gallery items at a time up to the total for a campaign is all set, as well as the layout adjustments and responsiveness is all done.

cc: @DFurnes


lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -281,6 +281,7 @@
       <div class="wrapper">
         <?php if( $show_new_reportback ): ?>
 
+          <?php //dpm($variables); ?>
```
aaronschachter: :dragon_face:

aaronschachter:
Bender approved :+1: (except for the dpm comment)

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
weerd: doh!

aaronschachter:
@weerd - Does this also close out #3663 ?

weerd:
Good catch! Yep it does.

DFurnes:
Looks really nice. :+1:

sbsmith86:
This looks good to me to. Let me know when you are ready to merge my stuff in. The only things I see missing in Reportback.js that I added was initiating to ImageUpload.js and an even handler. I don't think there will be that many issues however my files are in a different place now so just need to make sure everything links up.

lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
```diff
@@ -18,7 +18,9 @@
 
 .campaign--action .container--prove,
 .campaign--sms .container--do {
-  background: $purple asset-url("images/campaign-pattern-default.png") repeat center 63px;
+  // @TODO: See if this needs to be removed with v2 of Reportback Prove it section.
+  // background: $purple asset-url("images/campaign-pattern-default.png") repeat center 63px;
+  background-color: $yellow;
```
DFurnes: Are we launching this yellow background color before Reportback v2? (Wondering if this should be hidden behind a feature flag.)

weerd: Good catch!

PR #3701, opened 2015-01-12
Gallery File fid's variable

aaronschachter:
Creates an admin variable to set the default placeholder images when a Campaign does not have enough Reportback Images to display.

screen shot 2015-01-12 at 10 26 05 am

For the sake of simplicity, the variable stores a comma separated list of File fid's.  To theme these images, use the new `dosomething_campaign_get_default_gallery_image_urls` function, which [explodes](http://php.net/manual/en/function.explode.php) the variable and returns an array of image URLs to each file (if it exists).

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -815,3 +815,27 @@ function dosomething_campaign_form_campaign_node_form_after_build($form, &$form_
   drupal_add_js ($path . "/js/campaign_node_form.js");
   return $form;
 }
+
+/**
+ * Returns array of Image URL's for the default placeholder images in a Campaign Gallery.
+ *
+ * @param string $size
+ *   The preset to theme the image with.
+ *
+ * @return array
+ */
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -815,3 +815,27 @@ function dosomething_campaign_form_campaign_node_form_after_build($form, &$form_
...
+ *   The preset to theme the image with.
+ *
+ * @return array
+ */
+function dosomething_campaign_get_default_gallery_image_urls($size = '300x300') {
+  $urls = array();
+  // Variable name which holds a comma separated list of default File fid's.
+  $var_name = 'dosomething_campaign_default_gallery_fids';
+  // Extract the fid's.
+  $fids = explode(',', variable_get($var_name, array()));
```
sergiitk: I think it should be

``` php
$fids = explode(',', variable_get($var_name, ''));
```

aaronschachter: Makes sense, i'll make that change.

sergiitk:
:+1:, just make the change discussed.

PR #3706, opened 2015-01-12
Automatically flag Images for flagged Reportbacks

aaronschachter:
If a reportback is flagged, a user can still upload more images.  This PR sets each subsequent image status to `flagged` when the user uploads a new image to a flagged Reportback.


sergiitk:
:+1:

PR #3709, opened 2015-01-12
Fixes Flagged bug

aaronschachter:
The Review form sends through `delete` as 0... so need to check if its TRUE or not.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback_file.inc
+++ b/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback_file.inc
@@ -66,7 +66,7 @@ class ReportbackFileEntity extends Entity {
         $reason = $values['flagged_reason'];
       }
       $reportback->flag($reason);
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback_file.inc
```diff
@@ -66,7 +66,7 @@ class ReportbackFileEntity extends Entity {
         $reason = $values['flagged_reason'];
       }
       $reportback->flag($reason);
-      if (isset($values['delete'])) {
+      if (isset($values['delete']) && $values['delete']) {
```
sergiitk: Isn't `isset($values['delete']) && $values['delete']` is the same as `!empty($values['delete'])`?

aaronschachter: Good catch!  Sure is. Updated.

sergiitk:
:+1:

PR #3713, opened 2015-01-12
Deletes Promoted flag for real

aaronschachter:
`dosomething_reportback_update_7019` never ended up deleting the Promoted flag, and I missed it. The issue seems to be because the Image Description field was never properly deleted from the Promoted flag.

This PR adds a new hook, `dosomething_reportback_update_7020`, which basically is identical to 7019 except it deletes the Image Description field correctly, allowing the Promoted flag to be deleted.  This is pretty gross but manually changing the schema in the DB doesn't feel like the right way to handle either...

What do you think @sergii-tkachenko ?


sergiitk:
:+1:

PR #3716, opened 2015-01-13
Don't encode Reportback / RB File user input on entity load

### LINKED ISSUE #3704: Reportback Form: Special characters weirdness
Only filter text when displaying a reportback, not saving it.  Otherwise...

Who are you? I&#039;m batman!  

screen shot 2015-01-12 at 12 12 49 pm

aaronschachter:
Fixes #3704 - which was occurring when a User updates a Reportback that contains special characters in the Caption or Why Participated.  The form was being loaded with encoded input, thus the user would save the encoded values if they didn't go back in and fix the weird looking characters.


sergiitk:
:+1:

PR #3717, opened 2015-01-13
Flagged Reason migrate fix

### LINKED ISSUE #3652: Refactor Flagged Reportbacks

aaronschachter:
Closes #3652

Fixes error in `dosomething_reportback_update_7018`, which queried by the incorrect column when gathering the Flagged reasons.  This is a great example of why it makes more sense to simplify the build and store flaggings as a column on the reportback itself. 

Also exposes Review Source when viewing the Reportback File entity, to better spot check migrated Flagging records.


sergiitk:
:+1:

PR #3719, opened 2015-01-13
Use bundled Ruby Sass to compile.

DFurnes:
# Changes
- Use bundled Ruby Sass to compile. :art:

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/Gemfile
+++ b/lib/themes/dosomething/paraneue_dosomething/Gemfile
@@ -1,3 +1,4 @@
 source 'https://rubygems.org'
 
+gem 'sass', '~>3.4.0'
 gem 'scss-lint', '0.31.0'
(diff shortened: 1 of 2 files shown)


sergii-tkachenko:
:+1:

sergiitk:
It changes the error from

```
Running "sass:prod" (sass) task
Warning:
You need to have Ruby and Sass installed and in your PATH for this task to work.
More info: https://github.com/gruntjs/grunt-contrib-sass
 Use --force to continue.
```

to

```
Running "sass:prod" (sass) task
Warning:
bundleExec options set but no Bundler executable found in your PATH.
More info: https://github.com/gruntjs/grunt-contrib-sass
 Use --force to continue.
```

DFurnes:
Yep –– and Ruby Sass is actually a no-go because it requires 1.9.3 and we're stuck with 1.8.7 on the VM. I'm looking into jumping back to Libsass. Got everything compiling correctly with the latest Susy RC, but running into issues with the binary not compiling correctly on staging (Ubuntu).

DFurnes:
So this PR is :-1: for now. For local dev, you can just compile things from your local machine if you have Ruby 1.9.3+ installed, and it should all be good. Sorry for the troubles.

Issue #3707, opened 2015-01-12
Migrate Signups and Reportbacks to Campaign Run nid


aaronschachter:
Admin form that lets you select a Campaign node, and then are presented with dropdowns of campaign runs associated with that node.  If a Campaign run node exists, you are able to archive the Signups and Reportbacks to the selected Campaign Run nid.

Upon submit, bulk update the Signup and Reportback nid properties to the Campaign Run nid instead of the Campaign nid, thus opening up the Campaign for a second run.

aaronschachter:
This is almost complete, the form submit function just needs to use the Drupal Batch API in order to prevent timeouts.

aaronschachter:
Ran migrate form on QA for Comeback Clothes:

Archived 36149 Signups and 2795 Reportbacks from 362 to 2341.

Ran in about 30 seconds.  Not going to bother with the Batch API, as this is simply an update, not an entity save or anything like that. Closing this issue as completed.

aaronschachter:
http://staging.beta.dosomething.org/node/362/archive

http://qa.dosomething.org/node/362/archive

PR #3721, opened 2015-01-13
Archive Activity form

aaronschachter:
Refs #3707

Creates a page callback and form to bulk migrate a Campaign's Activity (Signups + reportback) to a Campaign Run associated with it, thus opening the Campaign up for a second run.

The Archive page will only display for Campaign nodes for which Campaign Runs exist.

This form won't be submitted any time too soon.  last TODO left is to add batch processing into the form submit, to avoid timeouts for campaigns with large activity numbers.

screen shot 2015-01-13 at 3 18 38 pm


sergiitk:
:+1:  That's a great job for only 1 hour. You're fast!

aaronschachter:
:racehorse:

PR #3725, opened 2015-01-14
Finder neue 6.0

DFurnes:
# Changes
- Restores lazy-loading behavior on the homepage.
- Updates campaign finder to use Neue 6.0 gallery markup.

### DIFF
--- a/lib/modules/dosomething/dosomething_home/dosomething_home.module
+++ b/lib/modules/dosomething/dosomething_home/dosomething_home.module
@@ -53,7 +53,7 @@ function dosomething_home_preprocess_node(&$vars) {
       }
 
       $tiles['status'] = TRUE;
-      $tiles['image'] = '<img src="' . $tiles['image_url'] . '">';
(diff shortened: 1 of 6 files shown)


lib/modules/dosomething/dosomething_home/dosomething_home.module
```diff
@@ -53,7 +53,7 @@ function dosomething_home_preprocess_node(&$vars) {
       }
 
       $tiles['status'] = TRUE;
-      $tiles['image'] = '<img src="' . $tiles['image_url'] . '">';
+      $tiles['image'] = '<img src="data:image/gif;base64,R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs=" data-src="' . $tiles['image_url'] . '">';
```
sbsmith86, line 56: Is there a reason why this image is hardcoded now?

DFurnes, line 56: Aye. It's a placeholder 1x1 spacer gif so that we can lazy-load in the images without flashing a broken image tag. (Set as a data-URL just so we don't need to make an extra HTTP request for a 72 byte image.)

sbsmith86, line 56: Ahh gotcha. Cool!

weerd:
:+1:

PR #3736, opened 2015-01-14
Use latest Neue 6.0 beta, and restores mosaic gallery views.

DFurnes:
# Changes
- Pulls in the latest beta of Neue 6.0, and restores mosaic gallery views to the homepage and campaign taxonomy page. :art:


weerd:
:shipit:

PR #3738, opened 2015-01-15
Fix display issue with staff pick flag in Finder.

DFurnes:
# Changes
- Fix display issue with staff pick flag in Finder.


weerd:
tableflip

:shipit:

PR #3739, opened 2015-01-15
Fix display issues on 404 page.

DFurnes:
# Changes
- Fixes display issues on 404 pages by moving to new Neue 6.0 container pattern.
- Removes unnecessary overrides; only one minor override on header pattern remains! :tada: 
- Removes unnecessary empty string from `autoplay` and `loop` boolean attributes.


sbsmith86:
Looks good! Yay for refactoring!

weerd:
Clean code with pizazz! :dancer: :+1:

PR #3743, opened 2015-01-15
Update campaign group template for Neue 6.0.

DFurnes:
# Changes
- Updates the "campaign group" template with new Neue 6.0 container pattern.


weerd:
:dancer: :+1:

PR #3745, opened 2015-01-15
No source mappy

DFurnes:
# Changes
- Disables source maps (having trouble rendering properly with node-sass beta).
- Fixes issue where CSS would not be overwritten if file already existed (which caused issues on watch task).


weerd:
:+1:

PR #3748, opened 2015-01-16
Clean up sources on taxonomy page.

DFurnes:
# Changes
- Place sources inside a `container__body` so they have appropriate spacing.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/taxonomy/taxonomy-term.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/taxonomy/taxonomy-term.tpl.php
@@ -49,25 +49,26 @@
 
   <section class="container -padded">
     <div class="wrapper">
-      <div class="container__body -narrow">
-        <?php if (!empty($facts)): ?>
+      <?php if (!empty($facts)): ?>
+        <div class="container__body -narrow">
           <h2 class="inline--alt-color"><?php print $facts_header; ?></h2>
           <ul>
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/templates/taxonomy/taxonomy-term.tpl.php
```diff
@@ -49,25 +49,26 @@
...
           <h2 class="inline--alt-color"><?php print $facts_header; ?></h2>
           <ul>
             <?php foreach ($facts['facts'] as $fact): ?>
               <li><?php print $fact['fact']; ?> <sup><?php print $fact['footnotes']; ?></sup></li>
             <?php endforeach; ?>
           </ul>
-
-          <section class="sources">
+        </div>
+        <div class="container__body -narrow">
```
weerd, line 61: So why does the "sources" block need to be enclosed in a separate `.container__body` as well if it relates to the information above? Can't it just be placed inside the `.container__body` for the facts list with some margin above it.

DFurnes, line 61: It can, but then we need to have specific overrides to add that top margin in each scenario. I think it's better to work with the same pattern we use for layout elsewhere. As for semantic concerns, that's why they're placed in the same `<section>`. :)

DFurnes:
Verbal :+1: from @weerd!

PR #3751, opened 2015-01-20
Grunt cleanup

DFurnes:
# Changes
- Cleans up Grunt configuration: 
  - Call multitasks in cases where there aren't more than one task per plugin.
  - Move temporary Node-Sass code out of Gruntfile.
  - Remove unused plugins.
- Fix display issue with admin tabs on widescreen displays.
- Remove empty sass files.


lib/themes/dosomething/paraneue_dosomething/package.json
```diff
@@ -9,23 +9,17 @@
   "devDependencies": {
     "autoprefixer-core": "^4.0.2",
     "css-mqpacker": "^2.0.0",
-    "grunt": "~0.4.1",
```
weerd, line 12: we no longer need Grunt? haha Is Grunt added globally to the server?

weerd:
I think it looks good :+1:

lib/themes/dosomething/paraneue_dosomething/package.json
DFurnes, line 12: Yep. That was a workaround for running `grunt-cli` locally from a ways back.

Issue #3647, opened 2014-12-16
outstanding redirects needed


mikefantini:
for some reason the following redirects didn't get created

https://docs.google.com/a/dosomething.org/spreadsheets/d/1xrpAdq1JOqGZDptgq_TAP47cpVwXIWUaMxeSCNYWIII/edit#gid=1821480830

can we follow the same process with the previous redirects? i.e. use the legacy URL alias and redirect to the beta canonical

mshmsh5000:
Script is done and updated; will wait until a non-Friday to run.

mshmsh5000:
Record of the redirects that were created: https://gist.github.com/mshmsh5000/bb6c30a0295854328bb5

PR #3753, opened 2015-01-20
#3647: Slight mod to previous script for legacy redirects

mshmsh5000:
If we need to do this _one more time_, I'll create a more generalized script. But this is a one-off just for #3647.


aaronschachter:
:+1:

PR #3761, opened 2015-01-20
Fix display issue with secondary navigation on mobile.

DFurnes:
# Changes
- Removed unnecessary wrapping div so that `list-style` applies correctly to the `<ul>`.


weerd:
:+1:

PR #3763, opened 2015-01-20
Issue3523 mbp international settings v2

### LINKED ISSUE #3523: Transaction settings in dosomething_mbp

deezone:
Fixes #3523 

Replaces:
https://github.com/DoSomething/dosomething/pull/3640
- Assigns Mandrill template name based on the `$origin` of the request:
  - `user_register`
  - `user_password`
  - `campaign_signup`
  - `campaign_group_signup`
  - `campaign_reportback`
- and the `dosomething_settings_get_affiliate_country_code()`.

All templates have been created and named in MailChimp based on this naming pattern. The MailChimp templates have been exported to Mandrill.

### DIFF
--- a/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
+++ b/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
@@ -13,7 +13,6 @@ function dosomething_mbp_install() {
   variable_set('message_broker_producer_rabbitmq_host', getenv('DS_MB_RABBITMQ_HOST'));
   variable_set('message_broker_producer_rabbitmq_port', getenv('DS_MB_RABBITMQ_PORT'));
   variable_set('message_broker_producer_rabbitmq_username', getenv('DS_MB_RABBITMQ_USERNAME'));
-  variable_set('message_broker_producer_rabbitmq_password', getenv('DS_MB_RABBITMQ_PASSWORD'));
   variable_set('message_broker_producer_rabbitmq_vhost', getenv('DS_MB_RABBITMQ_VHOST'));
 
   // Define the Message Broker application_id as the affiliate country code
--- a/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
+++ b/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
@@ -5,6 +5,36 @@
  */
  
 /**
+ * Implements hook_install().
+ */
+function dosomething_mbp_install() {
+
+   // Default Message Broker RabbitMQ connection settings
+   variable_set('message_broker_producer_rabbitmq_host', getenv('DS_MB_RABBITMQ_HOST'));
+   variable_set('message_broker_producer_rabbitmq_port', getenv('DS_MB_RABBITMQ_PORT'));
+   variable_set('message_broker_producer_rabbitmq_username', getenv('DS_MB_RABBITMQ_USERNAME'));
+   variable_set('message_broker_producer_rabbitmq_password', getenv('DS_MB_RABBITMQ_PASSWORD'));
+   variable_set('message_broker_producer_rabbitmq_vhost', getenv('DS_MB_RABBITMQ_VHOST'));
+
+}
+
+/**
+ * Implements hook_uninstall().
+ */
+function dosomething_canada_uninstall() {
+  $vars = array(
+    'message_broker_producer_rabbitmq_host',
+    'message_broker_producer_rabbitmq_port',
+    'message_broker_producer_rabbitmq_username',
+    'message_broker_producer_rabbitmq_password',
+    'message_broker_producer_rabbitmq_vhost',
+  );
+  foreach ($vars as $var) {
+    variable_del($var);
+  }
+}
+
+/**
  * Deletes dosomething_mbp_send_campaign_api system variable
  */
 function dosomething_mbp_update_7001(&$sandbox) {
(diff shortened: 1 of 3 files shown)


aaronschachter:
Looks way cleaner, awesome job @DeeZone! All looks good, just two questions: 

1 - How does the name change in Mandrill work with deploying this?  Did you just take copies of the current templates as rename them with the country codes? Emails are currently working right? :rabbit:  

2 - Just double checking - have you tested for local US and international?

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
```diff
@@ -108,3 +108,30 @@ function dosomething_mbp_update_7003(&$sandbox) {
...
+   variable_set('message_broker_producer_rabbitmq_username', 'dosomething');
+   variable_set('message_broker_producer_rabbitmq_vhost', 'dosomething');
+
+   $affiliate = dosomething_settings_get_affiliate_country_code();
+
+   variable_set("dosomething_mbp_user_register_template", 'mb-user-register-' . $affiliate);
+   variable_set("dosomething_mbp_user_password_template", 'mb-user-password-' . $affiliate);
+   variable_set("dosomething_mbp_campaign_signup_template", 'mb-campaign-signup-' . $affiliate);
+   variable_set("dosomething_mbp_campaign_reportback_template", 'mb-campaign-reportback-' . $affiliate);
+   variable_set("dosomething_mbp_campaign_group_signup_template", 'mb-campaign-group-signup-' . $affiliate);
```
sergiitk: Why do you still need these?

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
```diff
@@ -148,10 +145,12 @@ function dosomething_mbp_get_transactional_payload($origin, $params = NULL) {
       $payload['merge_vars']['IMPACT_NUMBER'] = $params['impact_number'];
       $payload['merge_vars']['IMPACT_NOUN'] = $params['impact_noun'];
       $payload['merge_vars']['REPORTBACK_IMAGE_MARKUP'] = $params['image_markup'];
-      $payload['email_template'] = 'mb-campaign-report-back';
       $payload['email_tags'][] = 'drupal_campaign_reportback';
       break;
 
+    default:
+      watchdog('dosomething_mbp', 'Invslid dosomething_mbp_get_transactional_payload() $origin value: ' . $origin, WATCHDOG_ERROR);
```
sergiitk: Wrong usage of `watchdog()`, see https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/watchdog/7

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
```diff
@@ -184,3 +183,20 @@ function dosomething_mbp_get_common_campaign_payload(&$payload, $params) {
   $payload['merge_vars']['CAMPAIGN_TITLE'] = $params['campaign_title'];
   $payload['merge_vars']['CAMPAIGN_LINK'] = $params['campaign_link'];
 }
+
+/**
+ * Define template name to use for transaction based on the event and the
+ * affiliate code.
+ *
+ * @param string $event_name
+ *   NamExample: "campaign_signup", "password_reset", etc
```
sergii-tkachenko: Why don't keep dashes, not underscores?

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
```diff
@@ -4,14 +4,44 @@
...
  */
  
- /**
+/**
+ * Implements hook_install().
+ */
+function dosomething_mbp_install() {
+
+   // Default Message Broker RabbitMQ connection settings
+   variable_set('message_broker_producer_rabbitmq_host', getenv('DS_MB_RABBITMQ_HOST'));
```
deezone: @sergii-tkachenko Do the environment variables get set in `/etc/environment` on our Vagrant box? Is this the same for the production server as well?

deezone:
@aaronschachter Yes, copies made of the existing Mandrill templates with entries for each affiliate as well as "us" - 9 countries x 5 templates = 45 new templates created. I'll remove the current templates once the new deployment starts to use the new / duplicated templates from the current templates.

I worked with @sergii-tkachenko to create tests of both us and the related affiliate sites on my local. The deployment process is:
- `bin/drush-intl vset --yes message_broker_producer_rabbitmq_password K*&%$@!**`
  Necessary to set password through drush script to keep password out of version control
- Set environment variables for RabbitMQ configuration : @sergii-tkachenko has suggested this can be done on the production server:

```
DS_MB_RABBITMQ_HOST="10.241.0.14"
DS_MB_RABBITMQ_PORT="5672"
DS_MB_RABBITMQ_USERNAME="dosomething"
DS_MB_RABBITMQ_VHOST="dosomething"
```
- enable `message_broker_producer` and `dosomething_mbp` modules as part of the affiliate site deployment: perhaps this can be done by updating `/html/profiles/dosomething/dosomething.info`?

aaronschachter:
@DeeZone That's an good idea, but you'd need to test it out first:
- get a clean affiliate db
- add them to the .info
- run a `ds build` for international

sergiitk:
:+1:

PR #3767, opened 2015-01-20
Reportback Image Rotation

sbsmith86:
@DoSomething/front-end 

This PR is for implementing the ability for users to rotate and crop an image. I had to refactor/reorganize the code to get this all to work, so if it would be easier, I can schedule a time for us to review the code as a group since this PR might be hard to read. It might even be easier to look at the files as a whole instead of looking at individual changes.

I also pinned on a couple of quick bug fixes: 
- Was using an undefined event in the getCoords() function that was breaking the script on FF, IE 10, and safari...and it was just a dumb mistake.
  - Making sure the button in the modal is always enabled even after the user submits the form and closes the modal. So if they change the file and reopen the modal the can still use the button.


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -283,18 +286,33 @@ function dosomething_reportback_form_save_file(&$form_state) {
...
-// Function to crop image.
-function dosomething_reportback_image_crop($image, $x, $y, $width, $height) {
-  // Crop the image.
-  $result = image_crop($image, $x, $y, $width, $height);
+/**
+ * Rotate and crops an image based on user input.
+ */
+function dosomething_reportback_image_edit($image, $x, $y, $width, $height, $rotate) {
+  // Rotate the image and then crop it if rotation was successful.
+  if ($rotate) {
```
weerd, line 302: should `$rotate` default to `false` or `null` in the parameters list?

`function dosomething_reportback_image_edit($image, $x, $y, $width, $height, $rotate = false)`

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -283,18 +286,33 @@ function dosomething_reportback_form_save_file(&$form_state) {
...
+    // Crop the image.
+    if ($result) {
+      $result = image_crop($image, $x, $y, $width, $height);
+    }
+  }
+  // If rotation not defined, just crop the image.
+  else {
+    $result = image_crop($image, $x, $y, $width, $height);
+  }
+
```
weerd, line 305: So this seems like the `$result = image_crop($image, $x, $y, $width, $height);` is repeated. Could the logic just go:
- if $rotate, then first rotate the image.
- then if $result, then crop the image.

Don't think you need to separate it with the `if/else` stuff.

sbsmith86, line 305: @weerd  If we aren't rotating the image, then we still need to crop it. `$result` just holds a true or false value indicating the image manipulation was done. 

If I understand your example correctly, if $rotate is not defined, then $result wouldn't get set and the crop would not happen.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
sbsmith86, line 302: fixed!

lib/themes/dosomething/paraneue_dosomething/scss/content/_modal-crop.scss
```diff
@@ -91,4 +99,4 @@
       }
     }
   }
-}
+}
```
sbsmith86: added a new line here.

lib/themes/dosomething/paraneue_dosomething/js/campaign/ImageCrop.js
```diff
@@ -200,27 +167,166 @@ define(function(require) {
...
+
+      $cropContainer.css({
+        height: height,
+      });
+
+      setupCrop(width,height);
+    }
+
+    // Actually do the rotation and translation.
+    $image.css({
```
DFurnes, line 257: `$.css` adds vendor prefixes automatically! :coffee:

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
DFurnes, line 305: :+1: to removing the extra `else` block.

weerd, line 305: Sorry, meant something more like this:

```
if ($rotate) {
  $result = image_rotate($image, $rotate);
}

$result = image_crop($image, $x, $y, $width, $height);
```

If rotate is set, then do the rotation and save to $result, otherwise just crop and save to $result. It'll still get cropped regardless of $rotate or not :)

sbsmith86, line 305: @weerd @DFurnes Hmm yea I had it like that before but then I was thinking. What if rotation failed for some reason? If we try to rotate and it fails, `$result` would be false but we would still try to crop it which would then either also fail (worst case) or it would crop the image without the rotation. It's not the worst out come but what do you think?

weerd, line 305: If it fails we probs have worse issues to deal with, haha :) Not sure though... @DFurnes ?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -283,18 +286,33 @@ function dosomething_reportback_form_save_file(&$form_state) {
...
-function dosomething_reportback_image_crop($image, $x, $y, $width, $height) {
-  // Crop the image.
-  $result = image_crop($image, $x, $y, $width, $height);
+/**
+ * Rotate and crops an image based on user input.
+ */
+function dosomething_reportback_image_edit($image, $x, $y, $width, $height, $rotate) {
+  // Rotate the image and then crop it if rotation was successful.
+  if ($rotate) {
+    $result = image_rotate($image, $rotate);
```
DFurnes: Actually, followup question – should this be assigned to `$image`? Seems like otherwise it would just be overwritten when going to crop it in the next step.

sbsmith86: @DFurnes the `image_rotate()` and `image_crop` functions take and image file, apply the crop or rotation, save the image file, then returns either true or false indicating it is successful. So after it rotates, the image file is saved with the rotation, and then we load the image file again, and do the crop.

weerd:
Great work @sbsmith86 :dancer: Looks good overall :+1:

PR #3776, opened 2015-01-21
IE8 fix for reportback gallery extended gallery layout on desktop.

weerd:
@DFurnes regarding fix for weird issue with RespondJS mediaMatch working correctly in IE8 & IE9.


DFurnes:
:ship: :earth_americas: :see_no_evil:

Issue #3769, opened 2015-01-20
persistent nav missing on mobile


mikefantini:
staging:
![image](https://cloud.githubusercontent.com/assets/6330971/5828695/56187c36-a0d5-11e4-8c8c-b5f88153f040.png)

prod:
![image](https://cloud.githubusercontent.com/assets/6330971/5828713/8d5dc0d4-a0d5-11e4-8029-a1560fb9071b.png)

DFurnes:
[This guy](https://github.com/DoSomething/dosomething/blob/59d33b71cd2ef930e58b99e69f4ec4a5b176dd48/lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback.scss#L1) seems to be the culprit. Is that safe to unhide @weerd?

weerd:
Yep, sorry!

PR #3777, opened 2015-01-21
Update to restore the page nav after it was removed for testing on IE.

weerd:
@DFurnes @mikefantini Nav should be restored after this update, tee hee :)


DFurnes:
(╯°□°）╯︵ ┻━┻

weerd:
Resolves #3769

PR #3778, opened 2015-01-21
Image sizes

### LINKED ISSUE #3758: duo gallery should be 150 x 150
placeholder images are correct size but campaign page thumbnails don't seem to be:

image

DFurnes:
# Changes
- Adds `.-medium` modifier to media pattern, which sets image container to 150px.
- Increase image sizes so they fit properly in new Neue galleries (figure now uses 400px image, media uses 300px image, both scaled to fit on the client).

Fixes #3758.


weerd:
Groovy! :dancer: :+1:

Issue #3764, opened 2015-01-20
need more space between paragraph text and inline buttons


mikefantini:
![image](https://cloud.githubusercontent.com/assets/6330971/5827871/5bb67688-a0cc-11e4-9441-0d04d983a08d.png)

to re-create, report back on two campaigns

mikefantini:
another example on Sandbox Campaign with Location Button:

![image](https://cloud.githubusercontent.com/assets/6330971/5828198/4c0a6dea-a0cf-11e4-8a11-80f9b3383c80.png)

PR #3783, opened 2015-01-21
Fix spacing for button on t-shirt page.

DFurnes:
# Changes
- Fixes spacing for button on t-shirt page. References #3764.


weerd:
Is it enclosed in `<p>` just to help add a top/bottom margin?

DFurnes:
Yep, add same standard spacing that text blocks get.

weerd:
Could we use a adjacent child selector deal... feels weird that it's a button class, but needs to have a wrapping `<p>` just for proper spacing...

something like:

```
p + .button {  margin-top: $base-spacing; }
```

DFurnes:
That's not a bad idea. I also wonder if there's a place for an `.inline-form` type element that could house buttons and other form fields within text blocks.

weerd:
Yeah, that could be useful :)

DFurnes:
Closing, we'll use the `.form-actions.-inline` pattern added in Neue 6.0-beta4! :dancer:

PR #3785, opened 2015-01-22
File upload support for IE8/9

aaronschachter:
From jam session with @weerd


weerd:
Yaaaas! I can haz custom file input markup! :+1: :dancers:

PR #3789, opened 2015-01-22
restrict cropbox to never be smaller than 480px

sbsmith86:
A user should not be able to crop an image under 480x480 so this just sets the minimum crop size to be that.

There will be another PR later to restrict user's from even uploading a file smaller than 480x480.

@DoSomething/front-end

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/campaign/ImageCrop.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/campaign/ImageCrop.js
@@ -46,7 +46,7 @@ define(function(require) {
   };
 
   /**
-   * Returns a values equal to the minimum allowed crop size scaled down to fit the image.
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/js/campaign/ImageCrop.js
```diff
@@ -45,6 +46,13 @@ define(function(require) {
   };
 
   /**
+   * Returns a values equal to the minimum allowed crop size scaled down to fit the image.
```
weerd: small typo `values -> value` :)

sbsmith86: whoops good catch.

weerd:
Looks good :+1:

PR #3791, opened 2015-01-23
Client side validation of image size

sbsmith86:
@DoSomething/front-end 

Adds some client side validation to check that an image is the correct size and shows and error message if it does not. The placement and style of the message might change.


DFurnes:
:ship:!

PR #3794, opened 2015-01-23
Enables dosomething message broker globally.

sergiitk:
Enables `dosomething_mbp` on US and international.
This enables `message_broker_producer` by the dependency.


deezone:
:+1:

PR #3797, opened 2015-01-23
Removed caption logic to reportback js

sbsmith86:
@weerd I moved the image caption show/hide logic into Reportback.js because the logic to show it after a user submits an image. It was in ImageUpload.js which we aren't using in the new reportback gallery. I needed to move it now, because the caption wasn't showing and a user won't be able to submit a form if they don't enter a caption. You had a TODO in there to do this. 

I haven't removed it from ImageUpload.js yet completely, since we are still supporting both experiences at the moment.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
@@ -43,35 +43,44 @@ define(function(require) {
      * @param  {jQuery} $container  Reportback container object.
      */
     init: function ($container) {
-      var _this = this;
-      
       this.reportbackContainer = $container;
       this.$reportbackWrapper  = this.reportbackContainer.find(".wrapper");
       this.$reportbackForm     = this.reportbackContainer.find("#dosomething-reportback-form");
       this.$uploadButton       = this.reportbackContainer.find(".js-image-upload-beta");
+      this.$captionField       = this.$reportbackForm.find(".form-item-caption");
       this.$initialGallery     = $container.find(".gallery--reportback");
-      
+
       this.itemsPrefetched = $container.data("prefetched");
       this.itemsTotal      = $container.data("total") - this.itemsPrefetched;
       this.nid             = $container.data("nid");
       this.offset          = this.itemsPrefetched;
-      
+
       if (this.itemsPrefetched < this.itemsTotal) {
         this.apiUrl = this.buildApiUrl();
         this.enableViewMore();
         this.enableResponsive();
       }
-      
-      // Initializes image upload modal.
-      var imageUpload = new ImageUploadBeta(this.$uploadButton);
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
```diff
@@ -43,36 +43,45 @@ define(function(require) {
...
+
+      this.imageUploadInit();
+    },
+
+
+    imageUploadInit: function() {
+      var _this = this;
       var imageUpload = new ImageUploadBeta(this.$uploadButton);
-      
+      var submittedImage = this.$reportbackForm.find(".submitted-image").length === 0 ? false : true;
```
sergiitk: Maybe it would be easier to reverse the condition?

``` js
var submittedImage = this.$reportbackForm.find(".submitted-image").length > 0;
```

sbsmith86: @sergii-tkachenko interesting. why is that easier?

weerd: Good call @sergii-tkachenko  seems a little cleaner and returns a boolean :)

weerd: It eliminates the need for a ternary. Just simpler logic.

sergiitk: 1. Because you won't have this reverse clause statement there `false : true;`
2. Because it also covers a case when `.length` returns `undefined` or `null`.

weerd: You could even simplify it further:

``` js
var submittedImage = this.$reportbackForm.find(".submitted-image").length;
```

If the length is `0` then it will be a falsey value... if it's any number above `0` then it will be truthy.

sbsmith86: @sergii-tkachenko @weerd cool. will update.

sergiitk: @weerd in this case it will not be boolean and may confuse when debugging. But another one ways is using double negation:

``` js
var submittedImage = !!this.$reportbackForm.find(".submitted-image").length;
```

sergiitk: I'll stick with `> 0` though — it's easier to read.

weerd: Gotcha, ok :)

weerd:
Apart from that one edit, looks great! :dancer: :+1:

PR #3798, opened 2015-01-26
Neue 6.0.0 beta3

DFurnes:
# Changes
- Updates for [Neue v6.0 beta3](https://github.com/DoSomething/neue/releases/tag/v6.0.0-beta3).
  - Remove waypoints & tabs patterns that were added to Neue.
  - Replace `.tabs` class usage with new `.waypoints` pattern.
  - Add Filament Group's [fixed-sticky](https://github.com/filamentgroup/fixed-sticky) polyfill for sticky nav.
  - Updates markup for CTA & message callout pattern.
- Updates class naming for sponsors container on homepage to match convention used elsewhere for styling specific instances of a pattern.


DFurnes:
Re: 5b44656, some of those overrides should definitely get merged back into the `.waypoints.-primary` pattern in Neue. I also ran into a specificity issue where `.waypoints.-primary` styles were overriding `.waypoints--action` styles unless I added some specificity to the rule `.waypoints.waypoints--action` so perhaps something to think about if we start running into this more often.

lib/themes/dosomething/paraneue_dosomething/js/app.js
```diff
@@ -28,6 +26,10 @@ define("app", function (require) {
   require("Analytics");
   require("tiles");
 
+  $(document).ready(function() {
```
sbsmith86, line 29: should this be in the document.ready() below?

DFurnes, line 29: Good call, merged!

weerd:
I think it looks good :) :+1:

PR #3804, opened 2015-01-26
Provides additional settings and debug messages to DS MBP installation.

sergiitk:
#### What's this PR do?
- Sets additional Drupal message during `dosomething_mbp` install stage to ensure its execution
- Specifies explicit `message_broker_producer` dependency
- Automatically sets rabbitmq password from env variable


deezone:
:+1:

PR #3807, opened 2015-01-27
waiting for image to load for mobile. fixes #3803

### LINKED ISSUE #3803: All iPhone pics that I uploaded triggered the "too small" error message
Steps to replicate:
1. Report back to a campaign on staging via iPhone
2. Upload a pic that's bigger than 480x480
3. Get an the error message that says your pic is too small and should be bigger than 480x480

sbsmith86:
### The problem

On mobile, when the validation ran the image had not yet been loaded and so the function thought that the dimension of the image was 0x0, causing an error to be thrown that the image was not large enough.
### Changes

This PR creates an image object from the file reader result and then waits for it to load using `image.onload = function() {}` . It then refactors the code a bit so that we pass the image created to various functions instead of the file reader result object. 

@DoSomething/front-end

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/campaign/ImageUploadBeta.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/campaign/ImageUploadBeta.js
@@ -1,5 +1,3 @@
-// @TODO - Move to new folder.
-
 define(function(require) {
   "use strict";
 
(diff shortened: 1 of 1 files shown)


DFurnes:
:+1: :shipit:

barryclark:
:rocket:

PR #3810, opened 2015-01-27
Fixing scoping issue, adding semicolon after function declaration.

sbsmith86:
Fixing some small bugs I found when compiling JS. @weerd @DFurnes


DFurnes:
:+1:

weerd:
So it was scoping to the parent object instead of the submit element?

weerd:
:+1:

sbsmith86:
@weerd `_this` wasn't defined in the `enableSubmit()` function and didn't need to be. I could use the global context of `this`

weerd:
Ah, gotcha :) Good catch!

PR #3811, opened 2015-01-27
Adjusts figure description height

### LINKED ISSUE #3806: Prove It button placement should be independent of detail height
On the user profile, we should have all the "Prove It" buttons be the same height. I think this is a bug right now, but it still seems to be dictated by number of lines.

upload

cc: @DFurnes @mikefantini

itsjoekent:
Fixes #3806 

screen shot 2015-01-27 at 2 15 23 pm


weerd:
:+1:

PR #3812, opened 2015-01-27
Implements ExternalAuthUserPassController on UK and Canada.

### LINKED ISSUE #3749: Canada forgot password for remote TIG accounts
Currently TIG-only user neither can restore its password on DS Canada (because the account to restore password to is created on first successful login) nor create new DS Canada account (because its email will be checked on TIG for the uniqueness).

The solution would be to lookup user in TIG DB by its email during password recovering and automatically create account DS Canada if it exists remotely.

CC @barryclark @mshmsh5000

sergiitk:
#### What's this PR do?
- Automatically creates user from remote source when forgot password requested and the user has no local account
#### Test results

UK:
screen shot 2015-01-27 at 3 04 25 pm

Canada:
screen shot 2015-01-27 at 3 06 40 pm

Fixes #3749


aaronschachter:
:+1:

PR #3817, opened 2015-01-27
Updates to grab full reportback data, such as captions and limiting number of prior reportbacks to show

weerd:
Now with captions!

This is now dynamically pulling in the captions added to each reportback, and showing it for the latest reportback submitted, as per the design mocks.

Also allows for specifying the number of prior reportbacks to show in new interface. Current setting is 5, which includes the latest submission (large image w/ caption), and 4 other reportbacks which are styled as tiny cubes in a row below the large entry.

@aaronschachter @sbsmith86 

CC: @barryclark @DFurnes

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -98,13 +98,23 @@ function dosomething_reportback_form($form, &$form_state, $entity = NULL) {
       // Output reportback images with date last updated.
       $form['reportback_submissions']['reportback_file']['#attributes']['is_supersized'] = FALSE;
 
-      $reportback_prior_images = array_reverse($entity->getThemedImages('300x300'));
-      $reportback_latest_image = array_shift($reportback_prior_images);
+      $reportback_submissions_data = array();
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -98,13 +98,23 @@ function dosomething_reportback_form($form, &$form_state, $entity = NULL) {
...
+      $reportback__submissions_count = 5;
+
+      foreach ($entity->fids as $index => $fid) {
+        $reportback_submissions_data[$index] = reportback_file_load($fid);
+        $reportback_submissions_data[$index]->image = dosomething_image_get_themed_image_by_fid($fid, '300x300'); 
+      }
+
+      $reportback_prior_submissions = array_reverse($reportback_submissions_data);
+      $reportback_prior_submissions = array_slice($reportback_prior_submissions, 0, $reportback__submissions_count);
+      $reportback_latest_submission = array_shift($reportback_prior_submissions);
```
weerd, line 112: Left this as separate array function calls because I figured it would be more readable. However, I can refactor and combine moreso if desired. :dango:

aaronschachter:
Sweet PHP :+1:

weerd:
Thanks! :dancer:

PR #3821, opened 2015-01-28
Adds profile link and UID to Zendesk users

### LINKED ISSUE #3565: pull in link to user profile and uid into Zendesk
When a user submits a a help ticket, the link to the user's profile and the user's uid should be easily accessible on the zendesk ticket.

itsjoekent:
Fixes #3565

@aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_zendesk/dosomething_zendesk.module
+++ b/lib/modules/dosomething/dosomething_zendesk/dosomething_zendesk.module
@@ -352,7 +352,7 @@ function dosomething_zendesk_verify_zendesk_user($email, $client = NULL) {
       return dosomething_zendesk_create_zendesk_user($email, $client);
     }
     // Update user with profile fields if missing.
-    if (!$search->results[0]->user_fields->profile_url) {
+    if (!isset($search->results[0]->user_fields->profile_url)) {
       $uid = user_load_by_mail($email)->uid;
       $profile_fields = dosomething_zendesk_get_profile_fields($uid);
       $id = $search->results[0]->id;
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_zendesk/dosomething_zendesk.module
```diff
@@ -351,6 +351,18 @@ function dosomething_zendesk_verify_zendesk_user($email, $client = NULL) {
...
+    $profile_fields = dosomething_zendesk_get_profile_fields($uid);
+    $id = $search->results[0]->id;
+    $updated_user = array(
+      'id' => $id,
+      'user_fields' => array(
+        'profile_url' => $profile_fields['profile_url'],
+        'profile_uid' => $profile_fields['profile_uid'],        
+      ),
+    );
+    $client->users()->update($updated_user);
```
aaronschachter, line 367: Can we check to see if the ZD User already has the fields set, vs setting every time? These values will never change, so its unnecessary overhead to post the update every time someone submits the form.

lib/modules/dosomething/dosomething_zendesk/dosomething_zendesk.module
```diff
@@ -351,6 +351,20 @@ function dosomething_zendesk_verify_zendesk_user($email, $client = NULL) {
       // Create a zendesk user.
       return dosomething_zendesk_create_zendesk_user($email, $client);
     }
+    // Update user with profile fields if missing.
+    if (!$search->results[0]->user_fields->profile_url) {
```
aaronschachter: Hmm. What does this look like if not set, does it throw an error? Or does each user return empty value here? (Do you need to check for `isset`)

itsjoekent: Just tested it and it worked fine but it did add an error to the log 

Notice: Undefined property: stdClass::$profile_url in dosomething_zendesk_verify_zendesk_user() 

Ill change it to isset and see what happens

itsjoekent: isset works fine pushing change now

aaronschachter:
:+1:

PR #3828, opened 2015-01-28
Wreck it fix it

DFurnes:
# Changes
- Applies the ol' `isset()` magic remedy. :pill: 

For review: @aaronschachter @DoSomething/front-end


aaronschachter:
:+1:

PR #3833, opened 2015-01-28
Approved as default status on RB Form

### LINKED ISSUE #3823: set default status of reportbacks in inbox to approved
this wouldn't actually assign the approved status to the reportbacks until the form was submitted, it would just allow for quicker moderating and less clicks by reviewers

so now the only clicks required in the inbox would be:
1. if a reviewer wanted to change the status from approved on any reportbacks 
2. submitting the form.

aaronschachter:
Fixes #3823


sergiitk:
:+1:

Issue #3834, opened 2015-01-28
Neue issues on canada.dosomething.org


barryclark:
Did we deploy the latest version of neue updates to Canada earlier? Seeing the problems below on campaigns:

![image](https://cloud.githubusercontent.com/assets/692762/5948933/0e5fe3b4-a717-11e4-8ca1-c68d773b06f6.png)

![image](https://cloud.githubusercontent.com/assets/692762/5948940/15a948ea-a717-11e4-806f-2abf837ccd27.png)

barryclark:
Assigning @DFurnes but might be one for @sergii-tkachenko to tackle if this is a deployment issue.

DFurnes:
Not seeing on my computer. Could Patrice double-check that it wasn't just a caching issue?

![screen shot 2015-01-28 at 6 07 56 pm](https://cloud.githubusercontent.com/assets/583202/5949136/bcb756a8-a718-11e4-8889-00cfee29f9d1.png)

![screen shot 2015-01-28 at 6 07 58 pm](https://cloud.githubusercontent.com/assets/583202/5949140/c24233d6-a718-11e4-9b96-1febc88f2ad4.png)

DFurnes:
Was able to reproduce the carousel issue on Firefox, will get a patch ready for tomorrow.

Is it possible the header issue could have just been a network thing? The image being used for Canadian Teens For Jeans is fairly large (about 1MB).

mfurdyk:
Whoa that's crazy big @DFurnes. I see the one on the US side is also 1 MB so it's the standard campaign image we had received... maybe the vamps have a rider requiring only highest quality jpg with no compression, but that's a hefty addition to the page size ;) Happy to compress or maybe have you flag internally to approve?

DFurnes:
Yikes, that's true! :grimacing: Updated both sites with a compressed version!

DFurnes:
Both issues have been patched, just waiting on next deploy!

PR #3837, opened 2015-01-29
Fix display issue where carousel was breaking out of grid.

DFurnes:
# Changes
- Fixes a display issue where carousel in "Prove It" section was breaking out of the grid in Firefox.

References #3834. For review: @DoSomething/front-end


sbsmith86:
:+1:

Issue #3792, opened 2015-01-23
Canada: Sync pre-SSO registered users via API


mshmsh5000:
Use list sent by Michael Furdyk. We will not try to sync their passwords across systems, but we'll still achieve the main goal: Sync email, so that TIG can contact them.

sergiitk:
The task consist from the following:
- [x] Create `dosomething_canada` drush script that posts users to the [laravel-user-api](/DoSomething/laravel-user-api)
- [x] Convert the list to csv or json
- [x] Test the the drush script locally
- [x] Deploy the changes to production
- [x] Run the script
- [x] Send the results to @mfurdyk
- [x] Remove drush script from DS codebase

sergiitk:
CC @mfurdyk

PR #3839, opened 2015-01-29
Creates a drush job to post missing TIG users to the remote user API.

sergiitk:
Ref #3792.


aaronschachter:
:+1:

sergiitk:
Note: please don't run this on QA.

PR #3842, opened 2015-01-29
Reportback placeholders

### LINKED ISSUE #3664: Theme Gallery placeholders
**UPDATE**

~~Blocked by #3694~~

This issue is to theme the placeholder images stored in #3694  if not enough approved images exist.

**ORIGINAL Description**

@mikefantini we are blocked on this. what is the decision?

My vote is that a campaign cannot be published without an editor uploading 6 placeholder reportback images (to inform the user of what we're expecting), which others seemed to be in agreement with during Innovation Mtg around this.

We could potentially add these images with a `'placeholder` status.  The gallery could not populate with "real" images until it has at least 6 approved (to avoid complicating the gallery logic of "I have 2 approved so display 4 placeholders").  Hoping to ship simplest solution first and then we can always refine later.

weerd:
Resolves #3664

_Now With Placeholders!_

The Reportback Gallery now handles a variety of scenarios with what is available from Promoted and Approved  Reportbacks and when needed filling in with Placeholders. See below:

*Promoted RB Count and Approved RB Count indicate how many of each respective type are available for a campaign in the database.

| Promoted RB Count | Approved RB Count | Placeholders Used |
| :-: | :-: | :-: |
| 0 | 20 | 0 |
| 2 | 20 | 0 |
| 0 | 0 | 6 |
| 2 | 0 | 4 |
| 2 | 3 | 1 |
| 6 | 0 | 0 |

@aaronschachter @mikefantini 

CC @barryclark @sbsmith86

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
@@ -400,8 +400,9 @@ function paraneue_dosomething_preprocess_reportback_files(&$vars) {
 
   $reportbacks_gallery['items'] = array();
 
+
   // Promoted Reportbacks:
-  // =====================
+  // ========================================================================
   // Set params to retrieve random Promoted Reportbacks for this nid.
   $params = array();
   $params['nid'] = $vars['nid'];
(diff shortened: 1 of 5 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
```diff
@@ -399,45 +399,72 @@ function paraneue_dosomething_preprocess_reportback_files(&$vars) {
...
 
-  // Get the total number of approved reportbacks.
+  // Approved Reportbacks:
+  // ========================================================================
+  // Set params to get the total number of Approved Reportbacks for this nid.
   $params['status'] = 'approved';
   if (isset($params['random'])) {
     unset($params['random']);
   }
+  $reportbacks_gallery['total_approved'] = dosomething_reportback_get_reportback_files_query_count($params);
```
aaronschachter, line 429: This is fine for now, but I've been thinking it may make sense to start storing totals in cache somewhere, so we're not constantly counting totals. I need similar functionality for the RB Reviewer app.  This more of a note than anything to change now though.

aaronschachter:
Ship it, value added :+1:

lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
weerd, line 429: Cool. Yeah, storing those in cache would be rather useful. Very much agree :)

PR #3843, opened 2015-01-29
Remove unused images from theme.

DFurnes:
# Changes
- Removes a bunch of images that we don't use. The majority of these are pattern resources that have since been moved to Neue.


weerd:
:sheep: IT!

PR #3852, opened 2015-01-30
Canada: remove post-missing-tig-users drush task.

### LINKED ISSUE #3792: Canada: Sync pre-SSO registered users via API

sergiitk:
Introduced in #3839.
Closes #3792.


aaronschachter:
:+1: :fire: :fire: :fire:

PR #3856, opened 2015-02-02
Reportback form: Query by Primary Cause

aaronschachter:
Alters the `dosomething_reportback_get_reportback_files_query` function to query the `field_data_field_primary_cause` table instead of `taxonomy_index`, to avoid situations where Reviewers may be reviewing the same reportbacks - as a campaign can only have 1 taxonomy term tid assigned to its Primary Cause.


sergiitk:
:+1:

PR #3858, opened 2015-02-02
Grants org friend access to the active page callback

### LINKED ISSUE #3822: give org friend role access to active tab
use case: we're taking Comeback Clothes and putting it back into the active state and need sponsor to approve updated language.

aaronschachter:
Fixes #3822

@mikefantini - Note: The Org friend will not see the tabs, but will be able to access the active page directly via URL (http://staging.beta.dosomething.org/node/362/active)

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -279,11 +279,7 @@ function _dosomething_campaign_closed_page_access($node) {
  * Determines whether a user has access to the active page callback.
  */
 function _dosomething_campaign_active_page_access($node) {
-  global $user;
-  // Grant access if you're  staff OR you're an org friend:
-  $has_access = dosomething_user_is_staff() || in_array('org friend', $user->roles);
-  if (!$has_access) {
-    // No active page for you.
+  if (!user_access('view any unpublished content')) {
     return FALSE;
   }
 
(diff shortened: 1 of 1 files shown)


mikefantini:
@aaronschachter sounds good!

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -279,9 +279,15 @@ function _dosomething_campaign_closed_page_access($node) {
  * Determines whether a user has access to the active page callback.
  */
 function _dosomething_campaign_active_page_access($node) {
-  // Staff only.
-  if (!dosomething_user_is_staff()) { return FALSE; }
+  global $user;
+  // Grant access if you're  staff OR you're an org friend:
+  $has_access = dosomething_user_is_staff() || in_array('org friend', $user->roles);
```
sergiitk: You need to add `dosomething_user` to `dosomething_campaign`'s dependencies.

sergiitk: In addition, do you think it worth to create dedicated `dosomething_user_is_staff_and_org_friend` function or extend `dosomething_user_is_staff` with `$include_org_friends` argument?

aaronschachter: So, an org friend is definitely not staff. (Staff can see all user info).

The only permission the org friend currently has is to `view any unpublished content`. What do you think about simplifying the query to check for that permission, instead of checking for both Staff OR Org friend?

sergiitk: Yes @aaronschachter, this is great idea!

PR #3859, opened 2015-02-02
Neue 6.0 beta4

DFurnes:
# Changes

Another messy one. This should be the last.
- Overrides theme functions to properly render `.field-label`, `.text-field`, `.text-field.-search`, `.select`, `.option` patterns from Form API.
- Uses new `list` pattern (and `.with-lists` wrapper for Markdown content) to apply list styles where necessary, and `-compacted` and `-concatenated` modifiers where necessary. This replaces the old `list-compacted` override we were using.
- Uses new `form-actions` pattern introduced in this release.
- Uses renamed `container__block` instead of `container__body`.

Ready for review! :cactus: 


lib/themes/dosomething/paraneue_dosomething/includes/form.inc
```diff
@@ -13,6 +13,133 @@ function paraneue_dosomething_form_alter_base(&$form, &$form_state, $form_id) {
...
 
 /**
+ * Implements theme_form_element().
+ */
+function paraneue_dosomething_form_element($variables) {
+  $element = &$variables['element'];
+  $output = theme_form_element($variables);
+
+  // If rendering a radio button, create rendered element for label
+  if($element['#type'] == 'radio') {
```
weerd: Better to use `===`?

lib/themes/dosomething/paraneue_dosomething/includes/form.inc
```diff
@@ -13,6 +13,133 @@ function paraneue_dosomething_form_alter_base(&$form, &$form_state, $form_id) {
...
+    $variables['rendered_element'] = $element['#children'] . '<span class="option__indicator"></span>' . "\n";
+    $output = theme('form_element_label', $variables);
+  }
+
+  // If rendering a checkbox, create rendered element for label
+  if($element['#type'] == 'checkbox') {
```
weerd: Same as above.

lib/themes/dosomething/paraneue_dosomething/includes/form.inc
DFurnes: In PHP too? Seems like we generally use `==`.

lib/themes/dosomething/paraneue_dosomething/templates/system/html.tpl.php
```diff
@@ -19,7 +19,7 @@
   <meta name="crowdcurity-site-verification" content="voJ0lyI8JXV5Ab1XAYMaoQj7WsoScgkfk9TLG0dA">
 
   <title><?php print $head_title; ?></title>
-  <script type="text/javascript" src="<?php print NEUE_ASSET_PATH; ?>/dist/modernizr-neue.js"></script>
+  <script type="text/javascript" src="<?php print DS_ASSET_PATH; ?>/dist/modernizr.js"></script>
```
weerd, line 22: :dancer:

lib/themes/dosomething/paraneue_dosomething/includes/form.inc
weerd: PHP also uses `===` to check type.

aaronschachter: @sergii-tkachenko has been reviewing PR's pushing for the `===` too

weerd:
Overall looks grand! :100: :sheep: IT!

lib/themes/dosomething/paraneue_dosomething/includes/form.inc
DFurnes: :+1: Will change!

PR #3860, opened 2015-02-02
Closed Gallery fix

aaronschachter:
The closed gallery data is set via `dosomething_reportback_get_reportback_files_query`, which was only returning `uid` if the logged in user has access to `view any reportback`.

Instead, expose the uid regardless of permission.  Fixes bug.


sergiitk:
:+1:

PR #3861, opened 2015-02-02
Reportback updates part 1

weerd:
Removed the `placeholder.tpl.php` file now that the Reportback placeholders will have captions; best to template them using the same markup as other placeholders. Added the hardcoded placeholder captions with some temporary lorem ipsum which needs to be replaced.

Also adds styles for the **Gigantor** button and cleans up other styles within the reportback interface.

@DFurnes @sbsmith86 

CC: @aaronschachter

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback.scss
@@ -8,7 +8,7 @@
     padding: ($base-spacing / 4);
 
     @include media($larger) {
-      padding: 10px;
+      padding: ($base-spacing / 2);
     }
 
     img {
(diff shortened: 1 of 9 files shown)


lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback.scss
```diff
@@ -176,13 +180,32 @@
...
     height: auto;
+    margin: ($base-spacing / 2) 0 0;
+    padding: 0;
+    text-transform: lowercase;
     width: auto;
   }
 }
 
+.reportback__submissions-latest {
+  background-color: #fff;  // remove
```
weerd, line 193: Just FYI, was purposely left here. Next few PRs will deal with items within this element and will be addressed then.

DFurnes:
Looks good to me! :ship:

PR #3868, opened 2015-02-03
Replaces PHP short-tag in Action Guide template which caused E_PARSE on PHP 5.9

### LINKED ISSUE #3865: Broken Campaigns on beta3
Broken

Per hipchat convo -- seems to happen only beta3... seems php5.5 syntax?

```
Warning: Illegal string offset 'safe_value' in dosomething_campaign_load() (line 128 of /var/www/beta.dosomething.org/releases/20150129215555/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc).
```

sergiitk:
Fixes #3865.


aaronschachter:
Looks good :+1: 

Should we also add a line ending to avoid the `E_PARSE: syntax error, unexpected end of file` per Hip thread?

sergiitk:
No, the end of file error message is caused by the mismatch of PHP open tags.

Issue #3790, opened 2015-01-23
Smoother cropping selection


sbsmith86:
Moving and resizing the crop box can be jumpy if not doing it really slow. Need to implement a solution using `requestAnimationFrame()` for a smoother interaction.

PR #3875, opened 2015-02-03
Bind move event to the document.

sbsmith86:
## The problem

One issue causing a rough animation when resizing the cropbox, was that I was binding the move event to the cropbox. That meant if the user's mouse left the cropbox while they were dragging the mouse to resize it, the animation would stop. The user was having to move the mouse really slowly to stay on the cropbox as they resized it.

Addresses #3790 . I will have another PR moving all the animations to use `requestAnimationFrame()`, but this should serve as a quick win. 
## Changes

Binds the mouse move event to the `$(document)` so the users mouse can be anywhere as the move.

@DoSomething/front-end

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/images/ImageCrop.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/images/ImageCrop.js
@@ -128,9 +128,7 @@ define(function(require) {
     $cropBox.off("touchend mouseup");
 
     // Bind move events to the document.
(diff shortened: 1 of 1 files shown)


DFurnes:
This event is only bound when cropper is active, right?

sbsmith86:
@DFurnes yea, this whole script only runs if cropping is enabled. And, the resizing, and moving of the cropbox only happens when the users is also holding the mouse down.

DFurnes:
Cool, works for me! :ship:

PR #3876, opened 2015-02-04
Campaign Count totals

### LINKED ISSUE #3846: Reportback Totals variables
We're calculating # of Approved in the Campaign template, and the RB Reviewer app displays total counts of Inbox for each Taxonomy Term. These numbers should ideally be written every time they need to change, vs every page load.

Ref: https://github.com/DoSomething/dosomething/pull/3842/files#r23847294

aaronschachter:
Refs #3846
- Displays Reportback totals for campaigns
- Updates totals per entity upon Review Form submission
- Renders count as link to the relevant view if non-zero

TODO:
- [x] Function to reset all campaign totals
- [x] Create `dosomething_campaign_get_campaigns` function to avoid `node_load` to get Campaign title

totals-2


sergiitk:
:+1:

PR #3885, opened 2015-02-04
Adds dosomething_campaign_get_campaigns function

aaronschachter:
Added to avoid `node_load` for each Campaign in the new `admin/users/rb` page.

Refs TODO in #3876


sergiitk:
:+1:

PR #3887, opened 2015-02-04
Refactor dosomething_helpers_set_variable

aaronschachter:
Refactors `dosomething_helpers_set_variable` so a loaded entity is no longer passed, but instead just the `entity_type` and `entity_id` are used as parameters.

This prevents unnecessary `node_load` calls when setting the Count totals for Campaign nodes and Taxonomy terms in `dosomething_reportback_get_reportback_files_query_count`.


sergiitk:
:+1:

PR #3895, opened 2015-02-04
Reportback change photo button

sbsmith86:
Refreshes the page when a user clicks "change photo"

@weerd @DFurnes

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/images/ImageUploadBeta.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/images/ImageUploadBeta.js
@@ -106,6 +106,13 @@ define(function(require) {
       event.preventDefault();
       Modal.close(false);
     });
+
+    // Close the modal when a user wants to change their photo.
+    this.$cropModal.find(".-change").click(function(event) {
+      event.preventDefault();
+      window.location.href = "#prove";
+      window.location.reload(true);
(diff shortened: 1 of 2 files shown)


weerd:
For a temp fix, looks good! :+1:

lib/themes/dosomething/paraneue_dosomething/js/images/ImageUploadBeta.js
```diff
@@ -106,6 +106,13 @@ define(function(require) {
       event.preventDefault();
       Modal.close(false);
     });
+
+    // Close the modal when a user wants to change their photo.
+    this.$cropModal.find(".-change").click(function(event) {
+      event.preventDefault();
+      window.location.href = "#prove";
```
DFurnes, line 113: You can close an open modal like this! :smile: 

``` js
var Modal = require("neue/modal");
Modal.close();
```

weerd, line 113: Problem is we're having some issues needing to clear out the file-input file, so we're temporarily doing a page refresh to reset until we can do a proper fix.

DFurnes, line 113: Ohhh, tricky.

PR #3898, opened 2015-02-05
Reportback Permalink fix

aaronschachter:
The RB permalink page was broken, because it renders the `dosomething_reportback_files_form` for the Reportback entity.  The new changes to the form introduced in #3876 were attempting to set count variables for the Reportback entity.  We only need count variables per node and taxonomy term, so this adds checks to only set the count variables if an `$entity_type` has been defined.

Also fixes PHP notices in the Reportback Totals page, and Fact Page list block.


sergiitk:
:+1:

PR #3900, opened 2015-02-06
Reportback cropbox size

### LINKED ISSUE #3864: Make the crop box bigger
Crop box should default to 50% of the size of the image. However it'll need to have a minimum size of 480x480.

sbsmith86:
## Fixes #3864

The cropbox was being set to start at 480x480. But if an image was really big, then as 480x480 box was too small to start with. 
## Changes

This PR makes some code style and comment updates, but mostly it adds a function `getCropBoxSize()` that calculates a larger starting size for the cropbox. 

@DoSomething/front-end


DFurnes:
:ship:!

weerd:
Nice! :sheep: it!

Issue #3884, opened 2015-02-04
Updates to crop box handle


barryclark:
- [x] unicode icon
- [x] double size of grab box for copping
- [x] move location of grab box to bottom right hand corner instead of hanging off the outside of the crop box

barryclark:
From #3774 - The area to trigger resizing the cropbox on desktop is causing the mouse cursor to change to the resize cursor too early. However, the resize area needs to be large on mobile to account for thumbs triggering the event. So we need to change the styling so that the resize area is smaller/larger based on screen size.

weerd:
@sbsmith86 looking for a useful unicode arrow. 

This might work: ⬊ `U+2B0A` (diagonal down arrow)

weerd:
We could also have @dave add an Icomoon arrow into out Neue icon font, and incorporate it that way. Was chatting with him about it and we could easily add this:
![screen shot 2015-02-04 at 2-1 02 50 pm](https://cloud.githubusercontent.com/assets/105849/6047338/2be662f4-ac77-11e4-9e40-260c5078f168.png)

And it default to the unicode diagonal down arrow if font fails for whatever reason ;)

sbsmith86:
@weerd @DFurnes That would be awesome.

weerd:
Cool. Yeah, @sbsmith86 for the time being add in the unicode and then we can replace once a Neue update goes through with the new icon :)

PR #3901, opened 2015-02-06
Style updates to the resize handle

sbsmith86:
## Fixes Issue #3884

Moves the resize handle in to the bottom right corner with a blue background and an south east arrow icon to indicate to the user that is where they need to grab to resize the box.

@DoSomething/front-end


DFurnes:
:shipit:!

weerd:
:+1:

PR #3903, opened 2015-02-06
Updating placeholder text with approved langauge.

### LINKED ISSUE #3850: Content for placeholder image captions.
Upcoming PR will include captions on placeholders that will be hard-coded for the time being. 

The language used is currently _lorem ipsum_ so we need to replace that will specific language unless it happens that the majority of our users speak 1st-century BC Latin, and these passages from the nonsensical translation of _Cicero's_ text is somehow _on brand_ ;)

-Cheers!

image

weerd:
## Fixes #3850

Add new approved placeholder text to replace the _1st-century BC Latin_ sentence fragment we were using temporarily.

@DoSomething/front-end 

For reference:
image


DFurnes:
:checkered_flag: :chart_with_upwards_trend:

PR #3907, opened 2015-02-08
Show preview image of correct image file

### LINKED ISSUE #3892: preview shows wrong image if user first uploads an image less than 480 x 480
If the user attempts to upload an image smaller than 480x480 ( an error message is generated), and then attempts to upload an image larger than 480x480, the thumbnail preview shows the original smaller attempted uploaded image instead.
1. Register or sign in to DoSomething.org and go to http://staging.beta.dosomething.org/campaigns/hunt-day-2-trash-scavenger-hunt
2. Try to upload an image that is smaller than 480x480.
3. After the error message appears, try to upload an image that is 480x480 or larger.
4. Click Done.

The thumbnail preview displays the original attempted uploaded image, instead of the latter image.

sbsmith86:
## Problem

The crop preview logic was tied to a Modal close event that was being subscribed to whether the image was valid or not. So, if a user first uploaded an image that was too small, that image was stored and bound to the modal close event which called the crop preview logic and this image didn't get updated when the user submitted another image that actually passed validation. 
## Changes - Fixes #3892

Only subscribes to the Modal close event, and calls the crop preview code if the image passes validation.

@DoSomething/front-end

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/images/ImageUploadBeta.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/images/ImageUploadBeta.js
@@ -56,6 +56,7 @@ define(function(require) {
           image.src = result;
 
           image.onload = function() {
+            var image = this;
             var isValid = _this.validImage(this);
             var $errorMsg = _this.$uploadButton.parent().siblings(".error");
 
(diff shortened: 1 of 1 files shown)


DFurnes:
:+1:

PR #3917, opened 2015-02-09
Updates view_unpublished contrib module

aaronschachter:
I've ran a build locally and verified that we no longer need the patch per latest update of the View Unpublished module.


sergiitk:
:+1:

PR #3922, opened 2015-02-09
user_name should be username in MBP params

### LINKED ISSUE #3921: DS MBP PHP Notice: Undefined index: username
Happens on password reset.

Notice: Undefined index: username in dosomething_mbp_get_transactional_payload() (line 127 of /var/www/beta.dosomething.org/releases/20150209191210/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module).

deezone:
Fixes #3921 

`dosomething_mbp_get_transactional_payload()` is looking for `username` rather that `user_name`.

**See**:
- https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module#L127
- https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_user/dosomething_user.module#L411


aaronschachter:
:+1:

PR #3923, opened 2015-02-09
Updated reportback image flow.

sbsmith86:
## Problem

Needed to rework the reportback image flow to deal with various cases of the user not fully submitting the modal form and closing out of the modal.
## Changes
- Only creates the crop preview image if the user hits the "done" button in the modal.
- If the modal closes without the user clicking "done", we clone and reset the file field so that the user can reupload the image
- Creates an edit photo button if the user hits done that allows the user to open the modal again and reset their crop.
- Made some updates to the crop preview logic that was causing the preview to break if the modal was submitted more than once.

@DoSomething/front-end


DFurnes:
Looks good to me! :boat:

weerd:
Yar, looks good :dancer:

PR #3927, opened 2015-02-10
Reportback ie behavior

### LINKED ISSUE #3699: Reportback Upload IE Support
Hey there, @mikefantini @barryclark @weerd @DFurnes !

IE8 and 9 don't support the code we are using to allow a user to upload an image and see a preview of the image before we send it to the backend. Without being able to see the preview image there is no way for us to support the cropping functionality. 

So we were discussing ways to change or support the image upload flow for IE users who can not crop images. 

The options thrown out were: 
- We could not pop open the modal, but somehow indicate to the user that file has been uploaded (by showing the filename, for example). Then we would do some sort of auto crop ( I think this is the best option.)
  - We could try to find a polyfill that would allow us to give broswers that don't have it, the support we need to implement cropping. This may or may not be feasible. 
  - Not supporting reportback on IE 8 or 9 all together. 

So I think we just need to settle on a solution and make sure everyone is on the same page.

weerd:
## Fixes #3699

This script provides a fallback interface for the file upload on the Reportback 2.0 when a browser does not support the File API.

The fallback interface will show a more streamlined interface (AFTER activating the input to add a photo) with the file name selected along with a link to _change photo_ if the user would like to select a different image before submitting.

image

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/images/ImageUploadFallback.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/images/ImageUploadFallback.js
@@ -4,26 +4,37 @@
 define(function(require) {
   "use strict";
 
-  var $ = require("jquery");
+  var $                    = require("jquery");
+  var _                    = require("lodash");
+  var fallbackUploadTplSrc = require("text!images/templates/fallback-upload-interface.tpl.html");
 
   var ImageUploadFallback = {
+    /**
+     * Initialize the fallback file input script.
+     * @param  jQuery Object $input    Button input from form.
+     * @param  String        container Class name for container that should house fallback content.
+     * @return void
+     */
+    init: function ($input, container) {
+      _this = this;
+      
+      console.log($input);
 
-    init: function ($button, $context) {
-      this.$container      = $context;
-      this.$uploadButton   = $button
+      // this.$fileInput = $input;
+      
+      // this.button.$original = this.$fileInput.parent();
+      // this.button.id = this.$fileInput.attr('id');
+      // this.button.name = this.$fileInput.attr('name');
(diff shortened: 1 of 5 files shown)


lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
```diff
@@ -61,13 +62,11 @@ define(function(require) {
         this.enableResponsive();
       }
 
-      if (Modernizr.filereader) {
-        this.imageUploadInit();
-      }
+      Modernizr.filereader ? this.imageUploadInit() : ImageUploadFallback.init(this.$uploadButton, '.reportback__submissions');
```
DFurnes: I'd argue an `if`/`else` block would be a bit more readable here...

lib/themes/dosomething/paraneue_dosomething/js/images/ImageUploadFallback.js
```diff
@@ -0,0 +1,135 @@
...
+  "use strict";
+
+  var $                    = require("jquery");
+  var _                    = require("lodash");
+  var fallbackUploadTplSrc = require("text!images/templates/fallback-upload-interface.tpl.html");
+
+
+  var ImageUploadFallback = {
+
+    component: {
```
DFurnes, line 14: I like this! :clap:

DFurnes:
:+1: aside from the one code style comment.

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
weerd: Yeah... was debating how readable that was. Will update!

weerd:
Cooool! Fire away!

Issue #3874, opened 2015-02-03
Calculate Reportback Totals upon insert and update


aaronschachter:
Updated description:
- [x] Update the pending total upon Reportback File insert for its campaign, and its primary cause tid.
- [x] Upon review (Reportback File save), update the totals for all statuses for the Reportback's campaign node and the campaign's primary tid.

Original description:

Refs #3846 
- Drush command to run `dosomething_reportback_reset_all`
- Schedule Jenkins job to regularly update saved Reportback Total helper variables.

aaronschachter:
@sergii-tkachenko Is this something you'd be able to take on? I'd be a little lost on scheduling the Jenkins Job.

It may make sense to alternate calculating the totals for the `taxonomy_term` and the `node` entities (vs doing them all in one job)

sergiitk:
Yes, I can do that.

PR #3928, opened 2015-02-10
Update Campaign RB Totals upon Reportback File Save 

aaronschachter:
Per convo with @sergii-tkachenko today re: #3874

Instead of relying on the Review Form to calculate Entity Reportback totals, calculate the relevant total per saving a Reportback File.

This PR calculates the total for the node reported back on. Following PR will update the node's Cause term totals.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback_file.inc
+++ b/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback_file.inc
@@ -169,6 +169,26 @@ class ReportbackFileEntityController extends EntityAPIController {
   }
 
   /**
+   * Overrides save() method.
+   *
+   * Updates count variables for corresponding nodes and taxonomy terms.
+   */
+  public function save($entity, DatabaseTransaction $transaction = NULL) {
+    if (DOSOMETHING_REPORTBACK_LOG) {
+      watchdog('dosomething_reportback_file', 'save:' . json_encode($entity));
+    }
+    $status = NULL;
+    if (isset($entity->is_new)) {
+      $op = 'insert';
+      $status = 'pending';
+    }
+    parent::save($entity, $transaction);
+    // Update the pending count for the node reportback was saved for.
+    $reportback = reportback_load($entity->rbid);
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback_file.inc
```diff
@@ -169,6 +169,26 @@ class ReportbackFileEntityController extends EntityAPIController {
...
   /**
+   * Overrides save() method.
+   *
+   * Updates count variables for corresponding nodes and taxonomy terms.
+   */
+  public function save($entity, DatabaseTransaction $transaction = NULL) {
+    if (DOSOMETHING_REPORTBACK_LOG) {
+      watchdog('dosomething_reportback_file', 'save:' . json_encode($entity));
+    }
+    $status = NULL;
```
sergiitk, line 180: I think it would be better to move this and the next `if` statement below `parent::save()` function. Nothing here is set that is affecting, `parent::save()`, but related to the count reset.

sergiitk:
Looks good, just one note above.

lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback_file.inc
aaronschachter, line 180: If I remember correctly, I think checking for `$entity->is_new` won't work after calling `parent::save()`.  We use this in the `ReportbackEntityController`'s `save` function as well.. but i'll check to make sure.

sergiitk, line 180: Yes, this can be a game changer, didn't know that. So this PR is good if your check passes.

aaronschachter, line 180: Just checked, and `entity->is_new` is not set after the save. Ok to leave in?

sergiitk, line 180: Yes, this is good then.

PR #3929, opened 2015-02-10
Reportback Access fix

aaronschachter:
Fixes nasty WSOD when viewing a Reportback entity which has been deleted.


sergiitk:
:+1:

PR #3936, opened 2015-02-11
Remove username from password_reset MBP payload

### LINKED ISSUE #3921: DS MBP PHP Notice: Undefined index: username

deezone:
Fixes #3921 

Remove `username` from mbp password_reset payload as it's not used by the Mandrill template.


aaronschachter:
:+1:

PR #3937, opened 2015-02-11
Remove EXIF data from file

### LINKED ISSUE #3909: Squishy preview image on mobile
- Image itself was distorted but took up the whole area
- Image was cut off, so only took up a portion of the area

sbsmith86:
## Problem

When images are taken with a phone or digital camera, EXIF data is attached to the image file which specifies image orientation. An image taken with a phone in "portrait" view, will actually create a landscape image, but the EXIF data tells the phone to view it as a portrait image. This was messing up the cropping and preview logic because it is heavily based on an images raw width and height, which was being flipped flopped on phones when the image was displayed. This caused the squishy image problem.
## Changes - Fixes #3909

Please see the comments in the code for more information, but basically, i'm removing the EXIF data on images entirely, allowing us to work with the RAW image on the front end. But since users can rotate images, if a portrait image appears landscape to the user, they can just rotate it back it's correct orientation.

@DoSomething/front-end 
/cc @mikefantini

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/images/ImageUploadBeta.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/images/ImageUploadBeta.js
@@ -71,7 +71,6 @@ define(function(require) {
 
             // When the user submits the modal form, send crop values to the form.
             _this.$imageForm.submit(function(event) {
-              console.log("close");
               // @TODO: This is a temporary solution until Neue is updated with capacity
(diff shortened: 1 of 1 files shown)


DFurnes:
Looks good to me. I don't think this is a bad solution at all. :+1:

weerd:
Awesome-sauce :dancer: Nice work! :+1:

PR #3940, opened 2015-02-11
Prefer square-crop over poster on mobile video tiles.

DFurnes:
# Changes
- Prefer nested images over the poster when setting fallback image on mobile. See DoSomething/neue#430.

Still considering whether there's a better solution for feature-detecting this!


weerd:
Looks good! :+1:

PR #3941, opened 2015-02-11
Updates Views contrib module

aaronschachter:
Fixes https://www.drupal.org/node/2424403 SA-CONTRIB-2015-039 - Views - Multiple vulnerabilities


sergiitk:
Can you also check if features with views are not overridden?

aaronschachter:
Yup, they all look ok after running a `ds build`.

sergiitk:
:+1:

PR #3942, opened 2015-02-11
Reportback rotate fix

### LINKED ISSUE #3935: Picture rotates in only 180 degree increments after clicking on Change Photo.
1. Log in to http://staging.beta.dosomething.org/campaigns/hunt-day-2-trash-scavenger-hunt
2. Click the Add Photo blue +.
3. Select a picture, and then click "Change Photo".
4. Click the Add Photo blue + again.
5. Select a picture, and then click "Rotate Photo".

The image rotates in 180 degree increments instead of the usual 90.

putting this in as a post-launch issue @barryclark @sbsmith86

sbsmith86:
## Problem

If user closed the modal and then clicked add a new photo, the previously bound click event to the modal was being re-initiated when it opened again so it was firing twice.
# Changes - Fixes #3935

Unbind any click event on the rotate button before turning it on to reset it. Can probably do with a refactor later to just set degrees to 0 when the modal opens. 

I also cleaned up some js linting errors I was getting. 

@DoSomething/front-end

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/images/ImageCrop.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/images/ImageCrop.js
@@ -279,7 +279,6 @@ define(function(require) {
    * @param {number} degrees    How many degrees to rotate the image.
    */
   var rotateImage = function($image, degrees) {
-    console.log(degrees);
     var width          = $image.width();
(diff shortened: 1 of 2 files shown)


DFurnes:
Looks good! :+1:

PR #3943, opened 2015-02-11
Reportback caption fix take 2

### LINKED ISSUE #3527: Reportback Caption character count
Determine character count, and then set up client-side validation to validate it.

Related to #3484 - SMS Reportbacks too

weerd:
## Fixes #3527

This should properly fix the caption field validation in the modal. Second try's the charm! This will temporarily break the submission of reportbacks until #3911 is completed (which is happening shortly).

@DoSomething/front-end


DFurnes:
Rad! :shipit:

sbsmith86:
good

PR #3948, opened 2015-02-11
unbind click event when modal closes

sbsmith86:
Need to unbind the click event on the "crop" button when the modal closes so it doesn't fire multiple times when the modal re-opens.

@DoSomething/front-end


weerd:
Sneaky little events! :+1:

PR #3949, opened 2015-02-11
unbind the click event right before rebinding it

sbsmith86:
Quick little fix to make sure the event get unbinded and rebinded each time. 

@weerd


weerd:
:sheep: iiiiiiit!

PR #3951, opened 2015-02-11
Fix of PNG images

sbsmith86:
@barryclark a quick fix to allow png images.


barryclark:
In person :+1: :smile:

PR #3958, opened 2015-02-12
Footnote fact

DFurnes:
# Changes
- Fixes two more instances of old "sources" markup. :grimacing: 


weerd:
:+1:

PR #3964, opened 2015-02-12
Reportback ie bug

sbsmith86:
This is really ugly and not and ideal solution. We are using browser sniffing to detect if we are on ios safari, which is the only browser that takes into account EXIF data. 
- If we are on IOS then, it reads the file as an ArrayBuffer and does all the EXIF stripping.
- If we are not, then it reads it as a dataurl and continues normally. 

This will need a major refactor as it is repeating a lot of code.

@DoSomething/front-end

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/images/ImageUploadBeta.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/images/ImageUploadBeta.js
@@ -49,51 +49,102 @@ define(function(require) {
       }
 
       var fr = new FileReader();
-      fr.readAsArrayBuffer(this.files[0]);
-      fr.onload = function () {
-        var blob = _this.getStrippedBlob(this.result);
-        var urlCreator = window.URL || window.webkitURL;
-        var imageUrl = urlCreator.createObjectURL(blob);
-        var image = new Image();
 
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/js/images/ImageUploadBeta.js
```diff
@@ -49,51 +49,102 @@ define(function(require) {
...
+              if (!$errorMsg.length) {
+                var $error = $("<div class='messages error'>Please upload a larger photo. Min. size is 480px by 480px.</div>");
+                $error.insertAfter(_this.$uploadButton.parent());
+              }
             }
+          };
+        };
+      }
+      else {
+        $("reportback__submissions form-wrapper").prepend("<p>" + Modernizr.exiforientation + "</p>");
```
DFurnes, line 101: Should this line be removed?

DFurnes:
Seems like a good temporary fix to me!

lib/themes/dosomething/paraneue_dosomething/js/images/ImageUploadBeta.js
sbsmith86, line 101: yes, good catch!

PR #3970, opened 2015-02-12
Adds preprocessing to the User Profile form

### LINKED ISSUE #3150: Hide timezone so users can't change it
Causes issues with campaign dates documented here: https://trello.com/c/pnzlbUt8

### LINKED ISSUE #3805: updates to User Profile Edit form (/user/uid/edit)
@deadlybutter after speaking with @aaronschachter it seems like these edits are a nice compromise in terms of updates we can make to the edit profile view (/user/uid/edit).

Sorry easiest way to make the edits was to just hack in browser so I took screen grabs of that, let me know if this rearrangement makes sense.

Zoomed out view:
image

Regular zoom components:
image
image
image

aaronschachter:
Fixes #3805 
Fixes #3150 

screen shot 2015-02-12 at 4 41 51 pm


sergiitk:
:+1:

Issue #3735, opened 2015-01-14
Remove Reportback feature flags / conditional logic


aaronschachter:
Remove checkboxes and relevant conditionals for:
- [x] DS RB Module Enable cropping
- [x] Paraneue "Show New Reportback"

aaronschachter:
This should be split up into 2 issues: one for the Reportback Form (to go to @sbsmith86) and the other for the Gallery template (to go to @weerd )

aaronschachter:
This is being handled by #3636

PR #3972, opened 2015-02-13
Neue 6.0 RC1

DFurnes:
# Changes
- Update Neue dependency & make required markup changes.
- Adds Carousel to the app until we remove old Report Back functionality in  #3735.
- Removes the code used to generate "order" classes, since we no longer use those with the new grid layout.

/cc @deadlybutter


weerd:
Nice nice :) :+1:

sbsmith86:
:100:

PR #3974, opened 2015-02-13
Fixes scholarship callout on SMS campaigns.

### LINKED ISSUE #3408: Official Rules needs to show up on SMS games and scholarship handwritten font alignment needs fixing
example https://www.dosomething.org/campaigns/science-sleuth (https://beta.dosomething.org/node/3478/edit)

@lkpttn please add mock of what this should look like here. thanks!

DFurnes:
# Changes
- Fixes bad alignment of scholarship callout on SMS campaigns.
- Moves message callout into form so that we don't have to use absolute positioning.
- Adds "Official Rules" link to the SMS games page template.

Fixes #3408.


weerd:
:sheep: it!

Issue #3636, opened 2014-12-12
Remove old Prove It code


aaronschachter:
Once we go live:
- [x] Remove the `reportback_image` array code in `dosomething_campaign.theme.inc`
- [x] Delete the `field_image_reportback_gallery` field from the DoSomething campaign feature / db
- [x] Remove old Step 4 from template
- [x] Clean up and remove old Reportback module form code.
- [x] Remove JS logic to check for whether using new or old Reportback.
- [x] Remove feature flag for enabling cropping.
- [x] Remove feature flag for toggling to show the new Reportback.

aaronschachter:
Assigning over to @weerd to kill the old Prove it section.

PR #3975, opened 2015-02-13
Deletes deprecated Reportback Gallery entity reference field

aaronschachter:
Refs #3636 - Removes the now depcreated field we used to upload User Reportbacks (which required creating Image nodes)


weerd:
:+1:

PR #3979, opened 2015-02-13
Update header markup on user profile page.

DFurnes:
# Changes
- Update header markup on user profile page.


weerd:
:+1:

PR #3980, opened 2015-02-13
Fix some display issues with figure on profile page.

DFurnes:
# Changes
- Fix some display issues with figure on profile page & updates old template variables to use new naming on the combined template.


weerd:
:+1:

PR #3981, opened 2015-02-13
Update to selection year options and adding some basic styles.

weerd:
Update to specify 13-25 year range correctly. Annnnd some basic Neue 6.0 styles to make it look nicer.

@DFurnes @sbsmith86 

CC: @mikefantini


DFurnes:
Radical! :snowboarder:

PR #3986, opened 2015-02-17
Fix issues with fact page layout.

### LINKED ISSUE #3771: fix spacing between facts on 11 facts pages
image

on prod we have a call to action which makes space between 5 and 6 not be an issue, but space between 10 and 11 would still cause problems

DFurnes:
# Changes
- Fixes #3771. Cleans up spacing issues on fact page. Also refactors template to remove excessive looping, instead printing two separate groups of facts.


sbsmith86:
:+1:

PR #3994, opened 2015-02-17
Turn off modal-container click event when the modal opens

### LINKED ISSUE #3953: crop modal shouldn't close when you mouse-up™ outside the modal
when a user goes to crop the image, if they drag their mouse outside of the modal and mouse-up™ the modal closes, even though this was definitely not the intention of the user.

we saw this issue a lot in user testing

I see the value in having global modals close when users click outside of the modal, but since the click is starting in the modal in this instance it shouldn't close

sbsmith86:
## Problem

While a user resizes the cropbox, if they mouse up outside of the modal, the modal closes.
## Changes - Fixes #3953

Turn off the click event tied to the `.modal-container` that calls the close handler in Modal.js, when the modal opens. I feel like most modal experiences allow the user to click outside of the modal to close it, so it seems to make since to just turn it off in this particular case. 

@DoSomething/front-end

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/images/ImageUploadBeta.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/images/ImageUploadBeta.js
@@ -27,7 +27,7 @@ define(function(require) {
     this.imageValues     = {};
     this.cropEnabled     = (typeof Drupal.settings.dsReportback.cropEnabled !== "undefined") ? Drupal.settings.dsReportback.cropEnabled : false;
     this.readyToSave     = false;
-
+    this.$body           = $("body");
     this.init();
   };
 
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/js/images/ImageUploadBeta.js
```diff
@@ -208,11 +208,16 @@ define(function(require) {
...
+
     Modal.open(this.$cropModal,
       {
         animated: false,
       }
     );
+
+    // Turn off the click event that closes
+    // the modal when a user clicks outside of the modal.
+    this.$body.off("click", ".modal-container");
```
DFurnes, line 220: This will turn off ability to close any modal by clicking the overlay until the page refreshes... this is a fine quick-fix, but we should have a more robust solution for the future. Can you add a `@TODO` to refactor?

Issue #3995, opened 2015-02-17
Spin up DoSomething Brazil (on hold)


barryclark:
Initial checklist for review @aaronschachter @sergii-tkachenko @blisteringherb @mshmsh5000:
- [ ] Initialize the Brazil site from the most recent US snapshot
- [ ] Copy over campaign images
- [ ] Clean out non-staff users
- [ ] Clean out report backs
- [ ] Create accounts for Brazil admin team

sergiitk:
What's the timeline for this @barryclark?

aaronschachter:
Why would we not just start from a clean database?

sergiitk:
@aaronschachter I think so we won't have to move the existing campaigns.

angaither:
closing #global

PR #3996, opened 2015-02-17
Adding brazil as an option for deployment and country code

blisteringherb:
Adding Brazil to lists of countries where necessary.

Addresses #3995 


angaither:
what superb php :elephant:

aaronschachter:
It sounds like we've landed on the decision to build out Brazil as part of the main DS app, so perhaps close this one? cc @mshmsh5000 @sergii-tkachenko @barryclark

mshmsh5000:
:+1: let's clean it up.

PR #3999, opened 2015-02-17
Using closeButtonOnly option when modal opens

### LINKED ISSUE #3953: crop modal shouldn't close when you mouse-up™ outside the modal

sbsmith86:
## Fixes #3953

Disable close on the modal overlay so users can't accidentally close the modal when resizing the crop box. 

@DoSomething/front-end


DFurnes:
Ship it! :shipit:

PR #4007, opened 2015-02-18
Caption bug fix

### LINKED ISSUE #3962: Caption required bug
Updating an existing reportback without changing the image always gives a "Caption field" is required error. Unless you choose to add another photo.

weerd:
## Fixes #3962

There are three potential scenarios this address:
- old Reportback exists, but was submitted prior to caption addition and thus doesn't have a caption, so we add a placeholder (only happens if user is updating one of these old reportbacks).
- new Reportback exists, and caption exists, so grab it and pass it to the form caption field.
- no reportback exists because user hasn't submitted anything, so leave the caption field blank.

This should fix the bug we were seeing with users trying to update a recent reportback in Reportbacks v2.0, but they were getting error messages that the caption field needs to be filled out.

@aaronschachter 

CC: @sbsmith86  @mikefantini


sbsmith86:
Nice :dancers:

PR #4009, opened 2015-02-18
Quick perf wins

DFurnes:
# Changes

A few quick performance wins & miscellaneous cleanup (see commits). :athletic_shoe: 

For review: @DoSomething/front-end @aaronschachter

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/patterns/_tile.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/patterns/_tile.scss
@@ -1,5 +1,7 @@
 // Tile pattern enhancements
 .tile {
+  transform: translateZ(0);
+
   img[data-src] {
     opacity: 0;
     transition: opacity 0.5s;
--- a/lib/themes/dosomething/paraneue_dosomething/template.php
+++ b/lib/themes/dosomething/paraneue_dosomething/template.php
@@ -51,6 +51,8 @@ function paraneue_dosomething_css_alter(&$css) {
   unset($css[drupal_get_path('module', 'date') . '/date_api/date.css']);
   unset($css[drupal_get_path('module', 'ctools') . '/css/ctools.css']);
   unset($css[drupal_get_path('module', 'views') . '/css/views.css']);
+  unset($css[drupal_get_path('module', 'field_group') . '/field_group.field_ui.css']);
+  unset($css[drupal_get_path('module', 'addressfield') . '/addressfield.css']);
 }
 
 /**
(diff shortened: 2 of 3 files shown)


DFurnes:
@aaronschachter @mshmsh5000 I assume we're safe to remove the `<meta name="crowdcurity-site-verification">` tag since we're all set up and verified, but just wanna double-check.

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_tile.scss
```diff
@@ -1,5 +1,7 @@
 // Tile pattern enhancements
 .tile {
+  transform: translateZ(0);
```
weerd, line 3: yaaas!

lib/themes/dosomething/paraneue_dosomething/template.php
```diff
@@ -51,6 +51,8 @@ function paraneue_dosomething_css_alter(&$css) {
   unset($css[drupal_get_path('module', 'date') . '/date_api/date.css']);
   unset($css[drupal_get_path('module', 'ctools') . '/css/ctools.css']);
   unset($css[drupal_get_path('module', 'views') . '/css/views.css']);
+  unset($css[drupal_get_path('module', 'field_group') . '/field_group.field_ui.css']);
+  unset($css[drupal_get_path('module', 'addressfield') . '/addressfield.css']);
```
weerd, line 55: These were getting added? @_@

weerd:
:+1:

mshmsh5000:
@DFurnes I think we can remove the tag now.

lib/themes/dosomething/paraneue_dosomething/template.php
```diff
@@ -92,6 +94,7 @@ function paraneue_dosomething_js_alter(&$js) {
 
   // Force removal of files that we can't seem to kill with exclude[js] array
   unset($js['profiles/dosomething/modules/contrib/devel/devel_krumo_path.js']);
+  unset($js['profiles/dosomething/modules/contrib/google_analytics/googleanalytics.js']);
```
aaronschachter, line 97: Porque?  If we don't want this we can just uninstall the module...

aaronschachter, line 97: But we use the module to track the different analytics in each affiliate site

DFurnes, line 97: Not removing the module, just removing some extra Drupal-specific JS that I don't believe we're using (such as custom rules for some lightbox plugin and CTools things). It still adds the Analytics snippet based on the `UA-XXXXXX-XX` site ID in the module settings.

aaronschachter, line 97: Ah ok. So its the .module that adds the snippet? http://cgit.drupalcode.org/google_analytics/tree/google_analytics.module?id=9bd941b#n328

DFurnes, line 97: Yep! The actual tracking snippet is added there as an inline script (which then makes its own async request for Google's analytics code from their CDN).

aaronschachter:
Bender approved :+1:

PR #4014, opened 2015-02-18
Updating placeholder text, errr... text.

### LINKED ISSUE #3952: change caption placeholder text
since it gets cuts off on mobile let's just make it "60 characters or less"

img_1027

weerd:
## FIxes #3952

Updated the caption text so it's shorter and fits nicely on narrow viewports.

I also updated the template so if we need further changes to the text it happens only in one place, even though we use the caption placeholder in two inputs. DRY that shiznit.

@DFurnes @sbsmith86


sbsmith86:
oh snap :dancer:

PR #4022, opened 2015-02-19
Reportback change photo

### LINKED ISSUE #3896: Change photo button should pull up a file input and change the photo in the modal
When a user clicks on the change photo button, they should be able to select a file and the new file should pop into the modal and the new file should be submitted with the reportback. 

Need to investigate how we are going to update the reportback form with the new file selected since there are security issues involved changing the value of a file field on the fly.

sbsmith86:
## Fixes #3896
- Moves the input button from the reportback form into the modal when the modal opens so that a user can change their image right from the modal. Moves the input button back into the form when they are done with it. 
-  I had to rework the error message handling to account for a user changing their photo in the modal and that photo being invalid.
- Slight update to the input style in the modal so that the cursor shows as a pointer.

@DoSomething/front-end


lib/themes/dosomething/paraneue_dosomething/js/images/ImageUpload.js
```diff
@@ -69,18 +69,36 @@ define(function(require) {
...
 
   /**
+   * Detatches and moves the current upload button
+   * @param {jQuery}  The element to move the button to.
+   */
+  ImageUpload.prototype.moveUploadButton = function($el) {
+    var $button = this.$uploadButton.detach();
+
+    $button.appendTo($el);
+  };
```
weerd, line 99: Nice function :dancer:

weerd:
Loooooks snazzy! Nice work :+1:

DFurnes:
Looks good to me! :shipit:

Issue #4018, opened 2015-02-19
Enable the use of an alternate background and text color in the Reportback section.


weerd:
Once #4017 is completed, use the new variables to customize the Reportback section with a background color and corresponding text color.

PR #4031, opened 2015-02-20
Wrapping up updating both backend and frontend with new custom settings.

### LINKED ISSUE #4017: Update Campaign Custom Settings
Add a **Alt text color** field.

This field will affect the color of:
- Pics or It Didn't Happen
- field_reportback_copy
- view more link

*update: We decided this would be a hex field

weerd:
## Fixes #4017  #4018

This update changes the name of the `alt_color` custom setting to `sponsor_color` and adds two new settings, `alt_text_color` and `alt_background_color`.

Update also applies these changes to the front-end to allow settings to take affect.

@aaronschachter 
@DoSomething/front-end


aaronschachter:
Some of the code in `dosomething_helpers. theme.inc`seems like it could be DRYed... but fine for now. Looks ok to me otherwise :+1: nice install hooking

weerd:
@aaronschachter indeed :) Definitely could use some cleanup at a later date.

PR #4039, opened 2015-02-23
Permalink feature flag

### LINKED ISSUE #4037: Permalink feature flag
Only display the permalink page if feature flag var is set

angaither:
Created `dosomething_display_reportback_permalink` variable to start work on permalink display.
If var is set, redirect to permalink page instead of old confirmation page. 

Fixes #4037

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -292,7 +292,14 @@ function dosomething_reportback_form_submit($form, &$form_state) {
     $form_state['redirect'] = $redirect;
   }
   if ($display_permalink) {
-    $form_state['redirect'] = 'somelink';
+    $form_state['redirect'] = array(
+      'reportback/' . $values['rbid'],
+      array(
+        'query' => array(
+          'rbid' => $values['fid'],
+        ),
+      ),
+    );
   }
 }
 
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -286,9 +286,13 @@ function dosomething_reportback_form_submit($form, &$form_state) {
   dosomething_reportback_save($values);
 
   // Redirect to confirmation.
-  if (!$is_admin_form) {
+  $display_permalink = variable_get('dosomething_display_reportback_permalink');
+  if (!$is_admin_form && $display_permalink) {
     $redirect = dosomething_campaign_get_confirmation_path($values['nid']);
     $form_state['redirect'] = $redirect;
+  }
+  if ($display_permalink) {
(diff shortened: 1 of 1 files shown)


angaither:
Actually @aaronschachter the path `reportback/rbid` is rendering in the admin view, is that something that is used in your reportback approval process? Should I create a new menu hook for this?

angaither:
I don't know if it would break all the things, but maybe for admins the url could be `admin/reportback/<rbid>`

aaronschachter:
I'm not tied to any of the paths, but current menu paths are as follows:
- `reportback/[rbid]` -- Yup this is the admin view of the Reportback and all its Files, renders in admin theme. Defined in the ReportbackEntity class.
- `rb/[fid]` -- Admin view of a Reportback File fid (displays Reviewer information), renders in Paraneue. Defined in the ReoportbackFileEntity class.

The reportback reviewer views do reference both of these paths.  Creating a new menu hook sounds easy enough for now, but just want to make sure we kill stuff we would deprecate too (likely the `/rb/fid` path)

angaither:
Yes, I was thinking of changing the `reportback/<rbid>` to `admin/rebortback<rbid>`
And using `reportback/<rbid>` as the permalink view

Which would mean this PR is ooooook (upon approval)

aaronschachter:
Sounds good :+1:

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -286,10 +286,21 @@ function dosomething_reportback_form_submit($form, &$form_state) {
...
+  if (!$is_admin_form && $display_permalink) {
     $redirect = dosomething_campaign_get_confirmation_path($values['nid']);
     $form_state['redirect'] = $redirect;
   }
+  if ($display_permalink) {
+    $form_state['redirect'] = array(
+      'reportback/' . $values['rbid'],
+      array(
+        'query' => array(
+          'rbid' => $values['fid'],
```
aaronschachter: Should this be `fid => $values['fid']` ?

angaither: :+1:

aaronschachter:
value added ship it :+1:

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -286,10 +286,21 @@ function dosomething_reportback_form_submit($form, &$form_state) {
   dosomething_reportback_save($values);
 
   // Redirect to confirmation.
-  if (!$is_admin_form) {
+  $display_permalink = variable_get('dosomething_display_reportback_permalink');
```
aaronschachter, line 289: Will you add a checkbox to toggle on/off? So we can test staging deploys without this feature. Also.. late feedback but would it be cleaner to prefix the variable with the module that defines it? e.g. `dosomething_reportback_display_permalink`

PR #4041, opened 2015-02-23
Adding in reportback divider and soem style updates.

### LINKED ISSUE #4020: Add dividing line in Reportback to make sections in interface more apparent.
Based on the design mocks, need to add dividing horizontal rule that separates the sections visually based on designs.

weerd:
## Fixes #4020

Adds a divider between the crop preview/mini gallery and the form below it. The divider will also use the custom background color setting if it is specified in the administrative interface for the campaign custom settings.

Had to readjust the markup slightly and add some wrapping divs around the form input elements so that the padding is added to the `reportback__submissions` and `reportback__inputs` containers instead of the immediate child div of the form. This helps so that the `<hr>` added can span the full width of the form container. Also cleaned up the markup slightly to wrap the crop related hidden inputs into the `reportback__submissions` container.

Lastly, did some style cleanup to help deal with old captions that are too long and or long words, so there's no visual overflow.

@DoSomething/front-end

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -142,20 +142,40 @@ function dosomething_reportback_form($form, &$form_state, $entity = NULL) {
   }
 
   // Hidden fields to collect cropping information.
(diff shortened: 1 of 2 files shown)


DFurnes:
Looks great! :boat:

Issue #3540, opened 2014-11-25
Reportback Permalink page (bender)


aaronschachter:
A page which displays a User's Reportback Image and Caption, and includes a Signup form / modal for its associated Campaign.  Also being referred to as a sharecard within the biz.
- [ ] Menu and page callback
- [ ] Determine path. e.g. `dosomething.org/rb/234`

aaronschachter:
@mikefantini @julielorch I'd suggest that if a Reportback File has been flagged, the permalink view of it will result in a Page Not Found. thoughts?

mikefantini:
@aaronschachter definitely makes sense to me for first implementation

mikefantini:
per meeting with @angaither and @aaronschachter we decided URL structure should be /rb/[rbid]#[fid] - this will set us up for a future where we want the permalink to host all images of a reportback and allow the user to either carousel through or use thumbnails to see different images

PR #4043, opened 2015-02-23
Updated reportback view url to admin/reportback.

angaither:
Change the current `reportback` url to `admin/reportback/[rbid]`

Updated the link where it's currently used.

This change is needed to allow for the creation of the user-facing `reportback/[rbid]`links.

Refs #3540

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -194,12 +194,12 @@ function dosomething_reportback_menu() {
     //'access callback' => 'dosomething_reportback_access',
     'access arguments' => array('administer modules'),
   );
-  $items['reportback/%reportback'] = array(
+  $items['admin/reportback/%reportback'] = array(
     'title' => 'Reportback',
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -194,12 +194,12 @@ function dosomething_reportback_menu() {
...
+  $items['admin/reportback/%reportback'] = array(
     'title' => 'Reportback',
-    'page callback' => 'dosomething_reportback_view_entity',
-    'page arguments' => array(1),
-    'access callback' => 'dosomething_reportback_access',
-    'access arguments' => array('view', 1),
+    'page callback' => 'dosomething_reportback_admin_view_entity',
+    'page arguments' => array(2),
+    // 'access callback' => 'dosomething_reportback_access',
+    'access arguments' => array('administer modules'),
```
aaronschachter: Campaigns team still need to access this view, and they don't have access to administer modules.  I believe changing the access arguments to `view any reportback` could do the trick...

aaronschachter:
Just the one access callback fix then good to go :+1:

PR #4060, opened 2015-02-24
Use standard header pattern on homepage, rather than wonky custom one.

DFurnes:
# Changes
- Use [Header](http://neue.dosomething.org/#header) pattern on homepage, rather than wonky custom one. It's the same thing but less code.

# Screenshots
#### Before:

screen shot 2015-02-24 at 2 58 25 pm
#### After:

screen shot 2015-02-24 at 2 58 14 pm


weerd:
Siiiii. :sheep: it!

Issue #4058, opened 2015-02-24
Old People


sergiitk:
### Description

AKA 26+ Club. An "old person" (26 and 26+) should:
#### User Registration
- [x] Should not be added to Mobile Commons
- [x] **Not be added** to the DoSomething Members list in MailChimp (`mailchimp_list_id` not equal to - ListID: `f2fab1dfd4` (for reference: `71893` - public facing ID)) and the **DoSomething New Members** grouping / **UserRegistration** interest group (ID# 10613)
  - [x] User **should be added** to the Old People / 26+ Club MailChimp list `mailchimp_list_id` **`a27895fe0c`** (for refernece `80825` - public facing ID) and the "General" interest_grouping -> **General Interest Group** (ID# 9613), mailchimp_group_name: **"Web Registrations"** (625)
- [x] Send a confirmation transactional email to the old person confirming they have been subscribed to the 26+ Club list. 
  - [x] Send **template name `mb-user_welcome-26plus_v1-0-0`** in message payload. https://trello.com/c/rrrHgiZC
- [ ] Adjust web page message after successful 26+ user registration from the general message "Registration successful. You are now logged in." to something supplied by @barryclark 
#### Campaign Signup
- [x] Should not be added to Mobile Commons
- [x] **Not be added** to the DoSomething Members list in MailChimp (`mailchimp_list_id` not equal to  `f2fab1dfd4` (for reference `71893` - public facing ID) and the **Campaign Specific campaign** grouping / interest group (values come from Drupal admin.
#### Mobile Commons
- [x] <del> Move Mobile Commons Signup Submissions to Message Broker</del> #3840
- [x] <del>Turn off Mobile Commons in MB payloads for persons older than 26 on registration and signups</del> #4061
#### MailChimp
- [x] Make MailChimp List IDs manageable in Drupal on "Third Party Opt-Ins" page #3841
- [x] Extend Drupal payloads with MailChimp List IDs
- [x] Replace MB MailChimp List ID setting in [`mbc-registration-email`](/DoSomething/mbc-registration-email/blob/master/MBC_UserRegistration.class.inc#L339) with the value from payload. Related issue: https://github.com/DoSomething/mbc-registration-email/issues/45
- [x] Change MailChimp List ID in Drupal payloads for 26+ Clud during registrations
- [x] Turn off MB for 26+ Club on signups
[...]

PR #4061, opened 2015-02-24
Turn off Mobile Commons subscriptions for the 26+ Club

sergiitk:
As specified in #4058, this PR turns off Mobile Commons subscriptions for persons older than 26 years old.


sergii-tkachenko:
CC @DeeZone

deezone:
Based on https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_user/dosomething_user.module#L494 looks good. 

+1

PR #4069, opened 2015-02-25
Reportback view more fixes

### LINKED ISSUE #3848: Slight bug in Reportback Ajax script if requestSize variable changed.
Nothing major, but if requestSize variable is changed to grab a larger amount there seems to be a slight bug in the logic. Just need to test and update.

### LINKED ISSUE #3908: View more link persists when there’s no more images to view

weerd:
## Fixes #3908

This update fixes the logic with the **View More** button for the Reportback Gallery. There was a slight bug in the logic that was affecting the script knowing whether there were some last stragglers in the amount of Approved RBs in stock. 

Also tested the script using different `bundleSize` and `requestSize` values and works great now. Increased the `requestSize` to 60, so the script gets a larger **JSON** file from the API.

@DoSomething/front-end

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
@@ -29,7 +29,7 @@ define(function(require) {
 
     bundleSize: 12,
 
-    requestSize: 20,
+    requestSize: 60,
 
     inventory: [],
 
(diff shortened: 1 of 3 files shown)


lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
```diff
@@ -29,7 +29,7 @@ define(function(require) {
 
     bundleSize: 12,
 
-    requestSize: 20,
+    requestSize: 60,
```
DFurnes, line 32: :clap:

DFurnes:
Looks great! :ship:

PR #4073, opened 2015-02-26
More cleanup

DFurnes:
# Changes
- Remove client-side code from old reportback interface –– no sense in making people download code & styles for things that will never be rendered.
- Little bit of user profile cleanup. Remove unused classes.
- Minor cleanup on campaigns template.


weerd:
Looks good! :+1:

PR #4076, opened 2015-02-26
Provides MailChimp list id in MB registration and signup payloads.

sergiitk:
#### What's this PR do?
- Adds `mailchimp_list_id` to signup and registration Message Broker payloads
- Provides a [Third Party Integrations](dev.dosomething.org:8888/admin/config/dosomething/opt_in) configuration to set `dosomething_signup_mailchimp_general_list_id`
- Presets the `dosomething_signup_mailchimp_general_list_id` variable from the environment var `DS_MB_MAILCHIMP_LIST_ID`

A part of #4058.

**Related**:
https://github.com/DoSomething/mbc-registration-email/issues/45

### DIFF
--- a/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
+++ b/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
@@ -109,8 +109,9 @@ function dosomething_mbp_get_transactional_payload($origin, $params = NULL) {
   // Payload specific to each transaction type.
   switch ($origin) {
     case 'user_register':
-      $payload['birthdate'] = $params['birthdate'];
-      $payload['subscribed'] = 1;
+      $payload['mailchimp_list_id'] = $params['mailchimp_list_id'];
+      $payload['birthdate']         = $params['birthdate'];
+      $payload['subscribed']        = 1;
       if (!empty($params['mobile'])) {
         $payload['mobile'] = $params['mobile'];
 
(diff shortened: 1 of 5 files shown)


sergiitk:
Tested it: `transactional_user_register`, `transactional_campaign_signup` and `transactional_campaign_group_signup` now contain `mailchimp_list_id`.

sergiitk:
MB emails with changed payloads work as before:
image

angaither:
:+1: :elephant:

deezone:
Is there a related issue for the MailChimp list ID for old people? Will the old people ID have an admin form field as well?

sergiitk:
I want to introduce the changes gradually, so this is nothing a first part of refactoring. This extends the payloads I mentioned above with `mailchimp_list_id`. The next step will be to make MB us this changes. After that I'll introduce additional configuration for 26+ Club. All this described in #4058.

deezone:
Sounds good!

Related MB consumer that will use the `mailchimp_list_id` setting:
https://github.com/DoSomething/mbc-registration-email/issues/45

deezone:
Running through tests on my local I've discovered how fragile missing MailChimp settings are. The condition here: https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_signup/dosomething_signup.module#L568 fails silently when no values are set. The result is no message being sent to MB. There needs to be some kind of feedback to ensure missing settings are intentional.

I think there needs to be some error messages / feedback around this condition:

https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_signup/dosomething_signup.module#L568

PR #4080, opened 2015-02-26
Restart func should be grepping for a different message

blisteringherb:
Restart php5-fpm function was looking for the incorrect message to determine if php should be restarted after deployment.  This fixes the incorrect string.

@barryclark @mshmsh5000


mshmsh5000:
:+1: :beer:

PR #4086, opened 2015-02-26
Progress on fixing image upload issue.

### LINKED ISSUE #3712: Reportback Cropping: image_gd_create() error
There is a bug that sometimes gets shown when cropping and image. It seems to be tied to when the image cropping script things that the crop area selected does not fully cover the image. 
image_gd_create

weerd:
## Fixes #3712

My suspicions were correct! After troubleshooting for a bit, found the `dosomething_reportback_image_edit($image, $x, $y, $width, $height, $rotate = NULL)` requires values for the `x`, `y`, `width`, `height`, that then get passed to `image_crop($image, $x, $y, $width, $height)`.

So, I'm passing along default `value` values for the hidden inputs. An alternative could be to set the defaults for the parameters in the function, but not sure what the best call is here. 

Either solution would work. I've also tested this using the technique described in the issue to reproduce the error and the error is gone :dancer: 

@DoSomething/front-end 
@angaither 
@barryclark

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -145,22 +145,27 @@ function dosomething_reportback_form($form, &$form_state, $entity = NULL) {
   $form['reportback_submissions']['crop_x'] = array(
     '#weight' => 20,
     '#type' => 'hidden',
+    '#default_value' => 0,
   );
   $form['reportback_submissions']['crop_y'] = array(
     '#weight' => 21,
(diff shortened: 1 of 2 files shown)


barryclark:
Nice one, @weerd!

angaither:
@weerd when does this happen? don't you always send in the values once the user crops?

weerd:
@angaither  Nope. Not if the API is not supported (example: IE8/9). There's no crop modal for browsers that don't support the File API, thus we just get the file and send it through without obtaining any of the `x, y, width, height, rotation` info.

angaither:
This works! :+1:

PR #4088, opened 2015-02-27
Update to fix broken animation when unveiling gallery items in Reportback gallery.

### LINKED ISSUE #4087: Animation on Ajax reportback gallery item reveal is broken.
Not sure when this happened, but it be broke. 

@mikefantini There was another issue for this, but I can't seem to find it.

weerd:
## Fixes #4087

Updates and fixes the unveiling animation bug on reportback gallery items after clicking _View More_. Also adds a class to the items, so if we want to add additional customizations to the animation, we can just do it within the css declaration :dancer: 

@DoSomething/front-end


DFurnes:
Woot! :shipit:

PR #4090, opened 2015-02-27
Updating preview gallery order of items.

### LINKED ISSUE #3934: preview images w/ multiple reportbacks should show most recent first
related to #3829

I thought that issue wasn't working properly, but I realized it is I had just reported back with more than 5 images and so it seemed like my new image was replacing the old one, not adding to my gallery

weerd:
## Fixes #3934

Quick fix to where the latest submission image get added to in the mini gallery after a user crops a new image.

@DoSomething/front-end


DFurnes:
:ship:

PR #4092, opened 2015-02-27
Add Image API rule to ensure images are always compressed JPEGs.

DFurnes:
# Changes
- Use the [ImageCache Actions](https://www.drupal.org/project/imagecache_actions) module to convert all header (`1440x810` image style) and tile images (`400x400` image style) to compressed JPEGs. :gift: 
# Why?

This was previously unenforced, so we ended up with a couple really big images (think 1MB) when people would upload PNGs. That's not great. By letting Drupal handle image compression, we can upload ginormous high-quality source images, and know that they'll always be nicely packaged up for users.

For review: @aaronschachter @angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_image/dosomething_image.features.inc
+++ b/lib/modules/dosomething/dosomething_image/dosomething_image.features.inc
@@ -161,7 +161,7 @@ function dosomething_image_image_default_styles() {
     'name' => '400x400',
     'label' => '400x400',
     'effects' => array(
-      2 => array(
+      3 => array(
         'label' => 'Scale and crop',
         'help' => 'Scale and crop will maintain the aspect-ratio of the original image, then crop the larger dimension. This is most useful for creating perfectly square thumbnails without stretching the image.',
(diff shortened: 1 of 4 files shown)


angaither:
I _think_ you have to add the new module somewhere else as well. 

Otherwise :+1:

aaronschachter:
@angaither you are correct -- if this contrib module is going to be enabled sitewide on all affiliate sites, it should be added to the dependencies in the DoSomething installation profile in `dosomething.info`

aaronschachter:
Ref: https://github.com/DoSomething/dosomething/blob/dev/bin/ds#L266

aaronschachter:
@DFurnes After making the change to the `dosomething.info` file, you can test the auto-enabling by pulling down staging db and running `ds build`

DFurnes:
Added to installation profile. Tested with `ds pull stage --db && ds build` and everything looks great!

aaronschachter:
Sounds good! Ship it :+1:

PR #4093, opened 2015-02-27
Update to fix odd IE8 bug with labels and inputs in Reportback fallback.

### LINKED ISSUE #3866: In IE8, the user is unable to change photo.
IE8/Windows 7 - I uploaded an image, was presented with the option to edit photo, but then nothing happens.

Seemed to work in IE9

weerd:
## Fixes #3866

So as it turns out if an `<input>` sits inside of a `<label>`, then you don't need a `for` attribute on the label that allows you to click the label to select the input. It will naturally behave that way. You only need the `for` attribute to match the `id` of the input to allow this behavior if the `<input>` sits outside the `<label>`.

In modern browsers, if you happen to add a `for` attribute to the label and an input sits inside of it, it's not a big deal and should still behave the same way; click the label and it selects the input. However, IE8 apparent :hankey: the bed and needs to not have any `for` attribute in the label if the input sits inside the label.

themoreyouknow

We also had the same `id` on the `<label>` and `<input>` which was a boo boo, so that wasn't helping things.

Also removed some extraneous commas because my snazzy new IDE was getting mad at them.

@DoSomething/front-end

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/form.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/form.inc
@@ -166,7 +166,7 @@ function paraneue_dosomething_file($variables) {
     return '<label class="gigantor" id="' . $id . '">' . $message . '<input' . drupal_attributes($element['#attributes']) . ' /></label>';
   }
   else {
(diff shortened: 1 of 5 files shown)


DFurnes:
:boat: :boat: :boat:

PR #4097, opened 2015-03-02
Reportback request animation frame

### LINKED ISSUE #3790: Smoother cropping selection

### LINKED ISSUE #3963: crop box doesn't go to the edge of the image
http://screencast.com/t/cI4oSKBPlJ

### LINKED ISSUE #4032: some rb images have black band on them
A few examples in the excluded section of Teens for Jeans reportbacks: https://beta.dosomething.org/node/1144/rb/reviewed?status=3

Here is one specific example: https://beta.dosomething.org/reportback/59664

sbsmith86:
## Changes

Refactors the ImageCrop script to use `window.requestAnimationFrame()` for better performance and smoother animations. Adds a requestAnimationFrame polyfill.

Fixes #3790  - Moving to requestAnimationFrame() creates a smoother animation by separating the tasks of calculating the new position of the cropbox, and redrawing the cropbox at the new position. 

Fixes #3963 - We keep track of the last position of the cropbox to do the animation now, so we can use that to set a boundary by fixing the position. So, if the new position of the box will be outside of the crop area, then don't allow it to go there.

Fixes #4032 - Similar to above, we can track if the user is resizing the box too big and stop the animation before redrawing the box, so that the cropbox never breaks the crop area. This should stop black bars being generated in reportback images. 

@DoSomething/front-end

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/bower.json
+++ b/lib/themes/dosomething/paraneue_dosomething/bower.json
@@ -34,7 +34,8 @@
     "respond": "~1.4.2",
     "almond": "~0.2.9",
     "jquery.iecors": "*",
-    "filament-sticky": "~0.1.4"
+    "filament-sticky": "~0.1.4",
+    "raf.js" : "0.0.4"
   },
   "resolutions": {
     "jquery": "1.8.3"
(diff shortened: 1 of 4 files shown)


DFurnes:
:+1:! :shipit:!

weerd:
Awesomeness! :dancer: :+1:

PR #4100, opened 2015-03-02
Update to provide fallback placeholders if no placeholders uploaded to site.

### LINKED ISSUE #3956: Use default hardcoded placeholder images as fallback.
If there are no placeholder `fids` available, then we need to provide a final fallback of the DS Logo filler so that the placeholders have some image content for the outputted markup.

weerd:
### Fixes #3956

Updated `dosomething_campaign_get_default_gallery_image_urls()` so it returns `false` if no default gallery images found, which is easier to test against in a conditional.

This allows falling back to local placeholder files if no placeholders have been uploaded to the site.

@aaronschachter 

CC: @barryclark @DFurnes


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -879,19 +879,28 @@ function dosomething_campaign_form_campaign_node_form_after_build($form, &$form_
...
+    // Loop through each one:
+    foreach ($fids as $fid) {
+      // Make sure an image exists:
+      if ($url = dosomething_image_get_themed_image_url_by_fid((int) $fid, $size)) {
+        $urls[] = $url;
+      }
     }
+    return $urls;
   }
-  return $urls;
```
aaronschachter, line 894: Perhaps it'd be better to explicitly return `FALSE` here?

aaronschachter:
Looks good except for one minor comment.

aaronschachter:
@weerd For future PR's can you please assign a reviewer?

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
weerd, line 894: The return `FALSE` happens but only if the `$fids` variable is `NULL`. There's a conditional there to decide that.

weerd, line 894: I might be misunderstanding what you mean tho :camel:

weerd, line 894: I could change it to this and might be more readable:

``` php
if (!is_null($fids)) {
    $urls = array();

    // Extract the fid's.
    $fids = explode(',', $fids);

    // Loop through each one:
    foreach ($fids as $fid) {
      // Make sure an image exists:
      if ($url = dosomething_image_get_themed_image_url_by_fid((int) $fid, $size)) {
        $urls[] = $url;
      }
    }
}

return false;
```

aaronschachter, line 894: Ah yeah, I missed line 886.  I think this is more readable.  Picky picky but it should be `FALSE`.

weerd, line 894: K. I'll update it. Yeah, I forget that PHP likes `FALSE` all in caps but JavaScript likes `false` all in lowercase. whatchagonnado :sheep:

weerd:
@aaronschachter updated!

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -879,19 +879,27 @@ function dosomething_campaign_form_campaign_node_form_after_build($form, &$form_
...
+
+  return false;
```
aaronschachter: just caps lock this `FALSE` and then ship it

aaronschachter:
:+1:

PR #4104, opened 2015-03-03
Store scaled minimum in a separate variable

sbsmith86:
Fixes a small bug where the minimum wasn't being a reset when the image changed by storing the global minimum (480x480) in a different variable than the scaled minimum (480 scaled down based on the image size) so it doesn't get overwritten. 

@DoSomething/front-end


DFurnes:
:ship:

weerd:
:+1:

PR #4108, opened 2015-03-04
Adds get reset url resource to API

### LINKED ISSUE #4062: Password Reset Endpoint
An endpoint (perhaps `/api/v1/users/password_reset?email=<taget email>` is needed to request a password reset link for a specific Drupal user account by email address (or perhaps UID).

The password reset endpoint will be used by the `mbc-user-import`script. A password reset link is needed to send as part of the message content in the transactional email message for newly created Drupal users.

An example being the daily niche.com user import.

**Related**:
https://trello.com/c/Qsywpzzo
https://github.com/DoSomething/mbc-user-import/blob/master/MBC_userImport.class.inc#L128
https://github.com/DoSomething/mbc-user-import/issues/27
https://github.com/DoSomething/mb-toolbox/issues/29

itsjoekent:
Fixes #4062 

Creates targeted ation at /users/<uid>/password_reset_url

Required for
https://github.com/DoSomething/mb-toolbox/issues/29

@aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
@@ -9,7 +9,7 @@ function _member_resource_defintion() {
         'file' => array(
           'type' => 'inc',
           'module' => 'dosomething_api',
-          'name' => 'resources/member.resource',
+          'name' => 'resources/member_resource',
         ),
         'callback' => '_member_resource_create',
           'args' => array(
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_api/resources/member_resource.inc
```diff
@@ -52,13 +52,35 @@ function _member_resource_defintion() {
...
+            'name' => 'uid',
+            'optional' => FALSE,
+            'source' => array('path' => 0),
+            'type' => 'int',
+            'description' => 'The uid of the user whose password reset url we are getting',
+          ),
+        ),        
+        'callback' => '_member_resource_reset',
+        'access callback' => '_member_resource_access',
+        'access arguments' => array('password_reset_url'),
```
aaronschachter, line 81: I don't see any changes to the `_member_resource_access` function. This endpoint should only be accesible to administrators, as public exposure could allow anyone to access the URL for any UID and access their account.

aaronschachter, line 81: Oh, I see, by default it checks for administrator access.  I think this would be a little cleaner if we actually didn't pass an argument to the access callback, and made $op = NULL by default... what do you think @deadlybutter ?

itsjoekent, line 81: I just followed how all of the other resources were doing it, the member resource access was already implemented here: https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_api/resources/member_resource.inc#L69

As for making it null by default... Yeah it makes sense, since only a few functions don't require admin. Do you want me to refactor this in a separate PR?

aaronschachter, line 81: Up to you, can merge this or just make the change in here. Code looks ok to ship :+1:

itsjoekent, line 81: I'll do it in a sep. PR, want to get the MB toolbox issue tied with this out of the way first -- ill come back after that to refactor and fix up the perm issue i have assigned

PR #4113, opened 2015-03-05
Allow an override for the getImage reportback file.

### LINKED ISSUE #4112: Reportback file getImage
So far this only returns images sized `300x300` this needs to be larger for the permalink page, it should allow for overrides.

angaither:
Fixes #4112


aaronschachter:
:+1:

Issue #3541, opened 2014-11-25
Reportback Permalink page front-end


aaronschachter:
Blocked by #3540, waiting on finished designs. CC @julielorch

sbsmith86:
Closed with PR #4114

PR #4114, opened 2015-03-05
Reportback Confirmation Page

sbsmith86:
Addresses #3541. Giving the confirmation page some #fender love :blush: 
- Added a new stylesheet for the permalink page
- Layout of template file to follow Neue patterns
- Use the "full_width" page layout for permalink pages

@DoSomething/front-end 
/cc @angaither @barryclark


weerd:
Nice work! :+1:

lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
```diff
@@ -0,0 +1,58 @@
...
+    &.-purple {
+      background: $purple;
+    }
+
+    .-white {
+      background: $white;
+    }
+
+    > .wrapper {
+      @include span(16);
```
DFurnes, line 18: How come this is necessary to override here?

weerd, line 18: I think cause it's a like a "card" sitting in the middle of the page. But come to think of it, maybe a css module `.card` class would work better here than overriding page stuff...?

Instead of handling everything in a top-level `.reportback__permalink` class that sits on the article up high, just target the card content and style specifically. Not sure. What says team fender?

sbsmith86, line 18: The mobile design has spacing around the white background, the default display spanned the white container 100%. 

image

DFurnes, line 18: Hmm. Yeah, I think maybe a `card` pattern might be the way to go here. Let's merge this and potentially refactor later.

DFurnes:
:+1: Looks good!

lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
sbsmith86, line 18: @weerd hmm i dunno because there is some styling that is not just specific to the card, like the header alignment on mobile, and the purple background.

weerd, line 18: Oh gotcha. Yeah, I guess that stuff would need to be overridden then. Sorry, I hadn't seen the mobile view yet @_@ Just going by what I remembered for the desktop view :dancer:

sbsmith86:
@weerd @DFurnes ok, i will add a quick todo an merge, I can refactor with the non-owners version.

PR #4118, opened 2015-03-05
the rbid is not set in the values array on new reportbacks.

### LINKED ISSUE #4117: Reportback confirmation Error
image

Error on staging when loading the confirmation page.

angaither:
use the saved reportback id from `dosomething_reportback_save` to redirect. 
Fixes #4117


sbsmith86:
:+1:

Issue #4123, opened 2015-03-05
Don't print name if there's no caption


angaither:
Old reportbacks don't have captions, so it looks odd when you print the name without a caption

![](https://i.imgur.com/v8nCx1K.png)

weerd:
It's like you've signed the picture :)

sbsmith86:
lolz good catch!

angaither:
@mikefantini should be proud of my QA

mikefantini:
:clap:

Issue #4121, opened 2015-03-05
Add scholarship qualification text


angaither:
This needs display if the incentive field is set on the campaign page.

sbsmith86:
@lkpttn @mikefantini is there a quick mock of what/where this should be? Is it just right next to the subtitle? like "Subtile. [incentive text]"?

mikefantini:
yeah you got it @sbsmith86 it's just the second sentence of the subtitle.

PR #4126, opened 2015-03-06
Permalink template updates

sbsmith86:
Add incentive language as the second sentence of the subtitle. #4121 
Only print the caption if it exists (old reportbacks don't have 'em). #4123 

@DoSomething/front-end


angaither:
that is some fending :+1:

Issue #2523, opened 2014-06-13
Campaign Runs - Report backs should be attached to campaign signups (?)


barryclark:
This issue is not confirmed for build yet, but serves as a placeholder for discussion.

@aaronschachter's notes here: https://gist.github.com/aaronschachter/cf17cb60d820a6f02c5c

aaronschachter:
The history here is that while we have a singular table of reportbacks for both web and mobile, we do not have a table that stores both mobile and web signups.

If there was one, it would make sense to associate reportbacks with signups.  We would then be only storing signups per campaign run, vs. needing to worry about the run for both signup and reportback tables.  This also makes for clean querying: wanting to find all users who signed up but did not reportback could happen within the single signups table.

Suggested course of action:
- For each mobile signup, create a record in signup table
- Update reportbacks to track what signup it belongs to

angaither:
don't think we are going this route, closing (cc @aaronschachter)

aaronschachter:
Makes sense not to re-engineer this in Phoenix if nothing is broken -- we do go this route in the Northstar API via the Campaign Activity object returned in `users/_id/:id/campaigns`. 

One thing to keep in mind if/when we ever look to support multiple Campaign Runs in Northstar -- it could be tricky since the current Phoenix process for archiving a Campaign updates the Campaign id of all of its Signups and Reportbacks to the closed Campaign Run id.

PR #4127, opened 2015-03-06
Adds Signup/Reportback objects as properties to Campaigns class

aaronschachter:
screen shot 2015-03-06 at 7 44 45 am

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -248,6 +248,15 @@ function dosomething_campaign_load($node, $public = FALSE) {
     unset($campaign->variables);
   }
 
+  // Return current user activity for this campaign:
+  $campaign->signup = NULL;
+  if ($sid = dosomething_signup_exists($campaign->nid)) {
+    $campaign->signup = signup_load($sid);
+  }
+  $campaign->reportback = NULL;
+  if ($rbid = dosomething_reportback_exists($campaign->nid)) {
+    $campaign->reportback = reportback_load($rbid);
+  }
   return $campaign;
 }
 


angaither:
@aaronschachter is there an issue for this?
this just returns data for the current user?

aaronschachter:
Yup, the issue is referenced in the commit message (https://github.com/DoSomething/dosomething/issues/4119) -- and yes, just interested in current user's activity.

angaither:
:+1: code looks ok
but I'm just very confused why this endpoint would return this data, shouldn't it be it's own endpoint?
calling for a node should return the node data, not user data.... and if it is returning user data, it should be requested in the url, not just the current user

aaronschachter:
Definitely open to suggestions / convo about this but the logic behind this is that anytime a user is viewing a Campaign, we need to know both if they are signed up for it and if they reported back.  We do this while viewing a Campaign on the web as well. I suppose apps which use the API could be make 3 separate GET requests, but that's 3 GET's per Campaign View.. and just seems like unnecessary steps if they could be returned in one shot...

sergiitk:
I agree with @angaither. It is strange to see current user's data returned by the endpoint correspondent to content management. I'd suggest to create `/users/current/*` endpoints for exposing current user related information.

aaronschachter:
But each signup and reportback is node specific, so what would those URL's look like @sergii-tkachenko  ?

sergiitk:
Something like `/users/current/campaign/1918.json`. This can return information user specific info for campaign id 1918, such as reportbacks and signups.

aaronschachter:
@sergii-tkachenko @angaither  I think the `current` is throwing me off here, because we will eventually start having Signups and Reportbacks for Campaign Run nodes.

Maybe lose the `current` for something like `users/activity?nid=362` ? 

Where the nid can be a Campaign, or a Campaign Run...

angaither:
That sounds good @aaronschachter 
Also, isn't the idea that the user api should be handling this anyways?

sergiitk:
@aaronschachter yes, sounds great!

aaronschachter:
Ok, I'll close this and open a new one. @angaither I believe storing signups/reportbacks in User API is the end goal but building this into Drupal for now to unblock mobile app development.

angaither:
yeah, the only thing blocking this in northstar the fact that there aren't users in the api.
But it's currently storing signups/reportbacks

aaronschachter:
Ohh @sergii-tkachenko I get the `current` now. So it would be

`users/current/activity?nid=362`
and 
`users/328342/activity?nid=362`

sergiitk:
:dancer: Yes!

aaronschachter:
This reminds me of how it'd be pretty nice to refactor Signups and Reportbacks into one "Activity" type table, to avoid separate Signup and Reportback id's: #2523

aaronschachter:
Might be worth revisiting this idea again. Per slack convo in the #api room: 

> When reviewing how we determine whether or not a user is signed up for a campaign, Evan (our former iOS mentor) was surprised the current user’s activity not included in the payload of a GET campaign?id=123. 
> The way the mobile app has to work without this data exposed in a campaign payload, is by querying for all of the current user’s activity upon app launch, then storing it in memory. we then traverse this array of activities (called a `DSOCampaignSignup`) when viewing a campaign. this is not ideal, because the user may have signed up on the web after launching the app.
> a 2nd scenario is 1 query to the campaign for the data, and 2nd API query to determine whether or not the user is signed up, which also isn’t ideal.

PR #4129, opened 2015-03-06
Provides users/:uid/activity API endpoint

### LINKED ISSUE #4119: API - Get Signup sid and Reportback rbid for Current User per Campaign
Check for current user and whether or a Signup / Reportback entity exists, and return relevant sid/rbid's.

aaronschachter:
Fixes #4119 to access a given (or the current) user's activity for a specified nid.

**GET** http://dev.dosomething.org:8888/api/v1/users/8/activity.json?nid=1619

**GET** http://dev.dosomething.org:8888/api/v1/users/current/activity.json?nid=1619

screen shot 2015-03-06 at 9 49 04 am

### DIFF
--- a/lib/modules/dosomething/dosomething_api/dosomething_api.info
+++ b/lib/modules/dosomething/dosomething_api/dosomething_api.info
@@ -6,4 +6,4 @@ dependencies[] = services
 features[ctools][] = services:services:3
 features[features_api][] = api:2
 features[services_endpoint][] = drupalapi
-mtime = 1425416358
+mtime = 1425655516
--- a/lib/modules/dosomething/dosomething_api/dosomething_api.services.inc
+++ b/lib/modules/dosomething/dosomething_api/dosomething_api.services.inc
@@ -124,6 +124,11 @@ function dosomething_api_default_services_endpoint() {
           'enabled' => '1',
         ),
       ),
+      'relationships' => array(
+        'activity' => array(
+          'enabled' => '1',
+        ),
+      ),
       'targeted_actions' => array(
         'password_reset_url' => array(
           'enabled' => '1',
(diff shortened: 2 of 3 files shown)


sergiitk:
:+1:

aaronschachter:
Minor discussion question -- this brings up https://github.com/DoSomething/dosomething/issues/2523 again. Note the response is not 2 separate properties for a Signup object and a Reportback object, but an object which contains relevant properties from both.

Any thoughts on whether or not this endpoint perhaps be renamed to Signups, e.g. `users/current/signups?nid=23` ? And Reportbacks become properties of a Signup?  It feels weird to have an "Activity" class which is the combination of both, only because "Activity" feels very non-descriptive... 

cc @sergii-tkachenko @jonuy @angaither @weerd

jonuy:
I think "activity" as a combo of the two is actually ok. It'd feel weirder for me if "reportbacks" were a property of "signups".

aaronschachter:
You should not be able to reportback without signing up, so it makes sense to me that reportbacks would exist as a property on a signup. it'd also mean we just archive one table (signups) vs two .

We do get the performance gain by having them as 2 separate tables tho.

jonuy:
I thought reportbacks and signups were on two different tables already and essentially acting as two different resources. That a reportback should only be created after a signup seems like logic that apps should be responsible for. Like with the current state of SMS, reportbacks are getting submitted without a corresponding signup.

aaronschachter:
@jonuy yes -- they are currently built out as two diff tables / entities.  https://github.com/DoSomething/dosomething/issues/2523 is more of a discussion point of: "Shouldn't it be a global / product rule that every Reportback have a corresponding Signup"

PR #4132, opened 2015-03-06
Allow non-owners to see the permalink page.

### LINKED ISSUE #4065: Permalink page title
After the rb is submitted, the title should say
"Awesome Work"
subtitle of "We got your pics"

@mikefantini is this language the same on every campaign?

### LINKED ISSUE #4066: Expose the permalink page to non owners.
Add `problem_fact` and `solution_fact` to the card.
Hide the why_participated section from non-owners.

### LINKED ISSUE #4106: Permalink var tokens
The permalink copy created needs to support tokens
Refs #4085

angaither:
This looks like more code than it is. 
I removed the if statement that only allowed owners & staff to see permalink pages. 
Added `token_replace` to set the owners name as the title for anon users.
Updated variable name from `is_current` to `is_owner`

Fixes #4106 
Fixes #4065 
Fixes #4066 

cc @sbsmith86

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -193,7 +193,7 @@ function dosomething_reportback_menu() {
     'page callback' => 'dosomething_reportback_view_entity',
     'page arguments' => array(1),
     //'access callback' => 'dosomething_reportback_access',
-    'access arguments' => array('view own reportback'),
+    'access arguments' => array('access content'),
   );
   $items['admin/reportback/%reportback'] = array(
     'title' => 'Reportback',
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -407,45 +407,42 @@ function dosomething_reportback_admin_view_entity($entity) {
  * Callback for /rbf/FID page.
  */
 function dosomething_reportback_view_entity($entity) {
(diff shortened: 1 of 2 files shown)


aaronschachter:
Doesn't this fail because of permissions? https://github.com/DoSomething/dosomething/blob/v0.4.27/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module#L196

aaronschachter:
Oh, I guess it would work... but we should the access callback to just something like `access content` or whatever the default for something like campaigns is?

angaither:
@aaronschachter yeah, I was going to export permissions in another PR
there's a permission already created that's `view any reportback`

aaronschachter:
Ok sounds good. It sounds like we need a new permission for `review any reportback`. All of the Review forms and the Review endpoint all use the `view any reportback` permission: e.g. https://github.com/DoSomething/dosomething/blob/v0.4.27/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module#L203 and https://github.com/DoSomething/dosomething/blob/v0.4.27/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module#L298

angaither:
Oooooh, okay, well I can update the menu hook to just say access content then, I just thought you built those special for me!

aaronschachter:
:\ sorry, `view any reportback` is super old. Actually we might be able to just rename it to `review any reportback`, since we're going public.

angaither:
cool, I updated the access callback perms, I'll make a new issue to update those permissions

aaronschachter:
Rockin :+1:

Issue #4067, opened 2015-02-25
Update title on permalink pages


angaither:
When a non-owner views the permalink page, the title should update.

it should say
"Join `[first_name]` and DoSomething.org"
subtitle:  `cta`

angaither:
closed with #4132

PR #4134, opened 2015-03-06
Fendering non-owner permalink page

### LINKED ISSUE #4131: Add 'do it' button to page
This button will direct people to the campaign pitch page

@sbsmith86 I'm not sure about button styles and stuff, but you would link to 
`drupal_get_path_alias('node/' . $node->nid)`

sbsmith86:
- Refactor permalink template to use a `card` pattern. 
- Updated template to work for the non-owner version of the permalink page. Addresses #4067 
- Added cta to template. Fixes #4131 

@DoSomething/front-end also, should this card pattern be in Neue? Halp.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
@@ -31,13 +31,18 @@
       padding: $base-spacing;
     }
 
-    &.-border {
+    &.-right {
       border-top: 2px solid $light-gray;
+      padding: 0px;
 
       @include media($tablet) {
         border-left: 2px solid $light-gray;
         border-top: none;
       }
+    }
+
+    .card__copy {
+      padding: $base-spacing;
     }
   }
 
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
@@ -12,6 +12,7 @@
   }
 }
 
+//@TODO - move into /patterns/card.scss
 .card {
   background: $white;
   margin: 0 $base-spacing / 2;
@@ -31,7 +32,7 @@
       padding: $base-spacing;
     }
 
-    &.-right {
+    + .card__column {
       border-top: 2px solid $light-gray;
       padding: 0px;
 
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
@@ -9,42 +9,34 @@
     &.-purple {
       background: $purple;
     }
+  }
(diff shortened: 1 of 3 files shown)


lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
```diff
@@ -9,44 +9,41 @@
...
+      @include span(6 of 12);
+      float: none;
+      display: table-cell;
+      vertical-align: top;
+      padding: $base-spacing;
+    }
 
-        &.-border {
-          border-top: 2px solid $light-gray;
+    &.-right {
```
DFurnes: Would this make more sense  as a `.card__column + .card__column`?

weerd:
Think overall looks good. Only thing I'm not keen on is the `.-right` modifier class. It implies layout location, but not sure if it should. For example, on mobile, is it on the "right" side? I'm assuming it likely stacks, so the name can be confusing. I would go with something that doesn't imply it's positioning, like giving each column a modifier like, `.-alpha` & `.-beta`, or some other better name.

Minor nit-picking :) Not crucial.

lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
weerd: :+1: to what @DFurnes suggests :)

DFurnes:
Looks nice! :+1: Let's also move `.card` pattern into a `scss/patterns/card.scss` directory with other patterns (not necessary for merge).

sbsmith86:
cool cool. Yea I will make that `-right` change!

weerd:
:dancer:

lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
```diff
@@ -9,44 +9,42 @@
...
+      @include span(6 of 12);
+      float: none;
+      display: table-cell;
+      vertical-align: top;
+      padding: $base-spacing;
+    }
 
-        &.-border {
-          border-top: 2px solid $light-gray;
+    + .card__column {
```
sbsmith86, line 35: @DFurnes @weerd How does this look?

weerd, line 35: Do you need the `&` in there?

```
& + .card__column { .. }
```

weerd, line 35: nevermind. I'm dumb. haha

sbsmith86, line 35: @weerd no, tried it both ways, didn't seem necessary. Also, it's similar to something like this. https://github.com/DoSomething/dosomething/blob/v0.4.27/lib/themes/dosomething/paraneue_dosomething/scss/patterns/_promotion.scss#L89-L94

PR #4137, opened 2015-03-09
Reportback final cleanup

### LINKED ISSUE #4107: Remove original Reportback Form code
Remove anything we don't need/use anymore. Here is one instance: 

https://github.com/DoSomething/dosomething/blob/v0.4.21/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.theme.inc#L12

weerd:
### Fixes #4107

Removes old code no longer in use now that Reportback v2 has launched.

@aaronschachter @DFurnes

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
@@ -8,6 +8,8 @@
  * - $classes: Additional classes passed for output (string).
  * - $campaign_creator: HTML for the Campaign Creator link/modal.
  */
+
+dsm($variables);
 ?>
 
 <header role="banner" class="header -hero header--action <?php print $classes; ?>">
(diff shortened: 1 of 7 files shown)


DFurnes:
:+1: from me!

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -263,7 +263,7 @@
 
     <div class="wrapper">
       <div class="container__block">
-        <h3 class="heading -beta inline-alt-text-color"><?php print t('Pics or It Didn’t Happen'); ?></h3>
+        <h3 class="heading -beta inline-alt-text-color"><?php print t('Pics or It Didn&rsquo;t Happen'); ?></h3>
```
aaronschachter, line 266: Not a biggie, but what's the reason behind this change? Kinda makes translation a little more difficult.

aaronschachter:
Looks good except I think encoding the apostrophe makes things less readable (and not sure why we'd need to encode it)

weerd:
So, it's a right apostrophe. Not a normal one, and always best to encode those. While not as readable it's in general better and actually makes it's more obvious that it's a special apostrophe character. It's not even available on the keyboard. It's a special HTML entity.

aaronschachter:
Dang, you fancy. Alright. :+1:

PR #4144, opened 2015-03-09
Fix automated test issues

DFurnes:
# Changes
- Back in the green! :traffic_light: Fixes some issues that were causing test failures.
- Removes reportback (Step 4) tests, since they aren't compatible with the new reportback interface.
- Fixes permission issue caused by change to `dosomething_signup_create` function, which now requires an active user session.

screen shot 2015-03-09 at 3 44 54 pm

For review: @aaronschachter


aaronschachter:
:+1:

PR #4146, opened 2015-03-09
Fixes the version of vagrant box in code.

sergiitk:
Specifies explicitly the version of vagrant box in code so it's controlled programmatically and we can test other released versions in branches.


blisteringherb:
:+1:

PR #4148, opened 2015-03-10
Require tip fields on campaigns

### LINKED ISSUE #4147: Make all of the Tip fields required on the campaign template
Re: https://trello.com/c/XAO9XYpa/30-step-three-merge-tag

Often campaigns team don't fill them out, resulting in the variable not being filled out correctly in transactional email messages.

image

@angaither is this quick to tackle?

angaither:
Fixes #4147

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.features.field_instance.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.features.field_instance.inc
@@ -1823,7 +1823,7 @@ function dosomething_campaign_field_default_field_instances() {
     'entity_type' => 'node',
     'field_name' => 'field_post_step_copy',
     'label' => 'After Copy',
-    'required' => 0,
+    'required' => 1,
     'settings' => array(
       'text_processing' => 1,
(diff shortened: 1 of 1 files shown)


aaronschachter:
@angaither -- Does this break things when creating/editing a SMS Campaign?

angaither:
well, it won't let you save without those fields, requested by @barryclark and @mikefantini not sure if it's needed or not.

aaronschachter:
If this PR doesn't allow you to save a SMS Campaign Form, we cannot ships it. 

If required, since we are hiding the form elements on the SMS Campaign Form, we'd need a validate hook to check to see if it's Campaign Campaign and not a SMS Campaign, and then check to see if the fields were inputted, which yeah, is gross.

angaither:
It lets you save the campaign, the fields just have to be filled out

barryclark:
Mike just mentioned that this won't allow you to submit the form on SMS campaigns. Doh. Apologies, wasn't aware of that @angaither. We might have to close this and just make sure that the campaigns team are aware they need to fill these out.

PR #4149, opened 2015-03-10
Set permalink as cached content.

### LINKED ISSUE #4138: Add caching to reportback permalink
set the contents of the permalink page into cache.

angaither:
- Adds content of themed permalink to cache
- This cache table already exists and can be cleared with drushy-boy

Fixes #4138


aaronschachter:
:+1:

Issue #4143, opened 2015-03-09
Allow users to share the permalink on the confirmation page. 


sbsmith86:
- [x] Link up facebook sharing
- [x] Link up twitter sharing
- [x] Link up tumblr sharing

angaither:
twitter metatag

```
<meta name="twitter:card" content="summary_large_image">
<meta name="twitter:site" content="@nytimes">
<meta name="twitter:creator" content="@SarahMaslinNir">
<meta name="twitter:title" content="Parade of Fans for Houston’s Funeral">
<meta name="twitter:description" content="NEWARK - The guest list and parade of limousines with celebrities emerging from them seemed more suited to a red carpet event in Hollywood or New York than than a gritty stretch of Sussex Avenue near the former site of the James M. Baxter Terrace public housing project here.">
<meta name="twitter:image:src" content="http://graphics8.nytimes.com/images/2012/02/19/us/19whitney-span/19whitney-span-articleLarge.jpg">
```

angaither:
oh, nice [twitter has a validator](https://cards-dev.twitter.com/validator)

sbsmith86:
@angaither sweet! so are these things you would set on your end?

angaither:
@sbsmith86 yeah, I'm working on the twittahz now.

aaronschachter:
FYI @angaither I remember I had to submit our site to twitter to get the Cards working for campaign pages: https://github.com/DoSomething/dosomething/issues/1678#issuecomment-48954404

Unsure if we'll get Twitter Card Magic right off the bat or if there's anything else we'll need to submit.

angaither:
[According to their docs](https://dev.twitter.com/cards/overview) it looks like we should be able to get the share card with just the meta tags. :pray:

mikefantini:
[...]

PR #4155, opened 2015-03-10
Twitter sharez

angaither:
first stab at getting twitter meta data in the permalink page.
this might need some tuning/field changes, but I would like to get this on stage to start testing the share card. 

Refs #4143


aaronschachter:
:+1:

PR #4159, opened 2015-03-11
Get file from reportback entity, if no query string.

### LINKED ISSUE #4157: Error on permalink pages with no FID
URL: http://staging.beta.dosomething.org/reportback/1186
`Fatal error: Call to a member function getImageURL() on a non-object in /var/www/beta.dosomething.org/releases/20150310213454/lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module on line 128`

angaither:
Fixes #4157


DFurnes:
Seems reasonable. :dancer:

PR #4164, opened 2015-03-11
Upload files via Reportback endpoint

### LINKED ISSUE #4151: Reportback API must support uploading files
Investigating using base 64 encoded strings, similar to the Services's default File CREATE resource.

Planning on building this as extra data values to pass into the current `campaigns/:nid/reportback`, `file` and `filename`.  If neither exist, inspect the `file_url` for a given URL string (used by SMS Reportbacks)

cc @jonuy

aaronschachter:
Closes #4151

Allows passing `file` and `filename` values to the `campaigns/:nid/reportback` endpoint to upload a file that has been base64 encoded.

Still supports uploading via remote image url (`image_url`) as well.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -220,9 +220,10 @@ function _campaign_resource_signup($nid, $values) {
  * @param array $values
  *   The reportback data to post. Expected keys:
  *   - uid: The user uid (int).  Optional, uses global $user if not set.
- *   - file: Base 64 encoded file string to save.
+ *   - file: Base64 encoded file string to save.
  *   - filename: The filename of the file provided as file.
  *   - file_url: The URL of the reportback file to save (used if no file/filename exist).
+ *   - caption: The caption for the Reportback File if provided.
  *   - quantity (int).
  *   - why_participated (string).
  *   - num_participants (int).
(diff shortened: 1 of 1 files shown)


aaronschachter:
cc @jonuy @sergii-tkachenko

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -234,10 +236,24 @@ function _campaign_resource_reportback($nid, $values) {
   }
   $uid = $values['uid'];
 
-  // Create a file from the $file_url.
-  $file = dosomething_reportback_save_file_from_url($nid, $uid, $values['file_url']);
+  $file = NULL;
+
+  if (isset($values['file']) && isset($values['filename'])) {
```
sergiitk: Maybe change `isset()` to `!empty()`?

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -220,7 +220,9 @@ function _campaign_resource_signup($nid, $values) {
  * @param array $values
  *   The reportback data to post. Expected keys:
  *   - uid: The user uid (int).  Optional, uses global $user if not set.
- *   - file_url: The URL of the reportback file.
+ *   - file: Base 64 encoded file string to save.
```
sergiitk: It's `Base64`, one word.

sergiitk:
Looks good to me, just some minor comments.

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
angaither: does this need to return errors in an else if both are not set?

aaronschachter: The else checks for a `file_url` and throws an error there if it doesn't exist.

Issue #4095, opened 2015-03-02
Kill current confirmation page


angaither:
Once we go live with the new permalink, we can remove code for the old confirmation page.
- [ ] remove the if that checks for the variable `dosomething_reportback_display_permalink`
- [ ] [remove the menu hook for the confirmation page](https://github.com/DoSomething/dosomething/blob/70bd75b65779ba70220610492942d2a29d9cb2be/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module#L30-L38)
- [ ] [remove menu hook callback](https://github.com/DoSomething/dosomething/blob/8853999ec664c142c5ae76419a996a6c66bafe4b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.pages.inc#L8-L16)
- [ ] [tshirt code](https://github.com/DoSomething/dosomething/search?utf8=%E2%9C%93&q=shirt&type=Code)
- [ ] [confirmation page tpl](https://github.com/DoSomething/dosomething/blob/8853999ec664c142c5ae76419a996a6c66bafe4b/lib/themes/dosomething/paraneue_dosomething/templates/campaign/reportback-confirmation.tpl.php)

Putting this in the milestone after the page goes live.

angaither:
@aaronschachter can you think of anything else I'm missing?

aaronschachter:
@angaither -- I'll take a look around.. one question is do we still display the Reportback Confirmation Message field on the Campaign node somewhere in the new page?

@mikefantini Do we want to kill T-Shirt form code completely, or should we leave it? There are actually quite a few variables here.

mikefantini:
@aaronschachter right now we don't have plans to display the reportback confirmation message field anywhere (confirmed by Ben), but I'm hesitant to lose that data just yet.

also re: the t-shirt form code I'm all for removing it completely.

aaronschachter:
I think there should be a separate issue from removing the T-Shirt form and variables. We should not kill the DoSomething Reward module/tables completely because we will need it for the mobile app.

angaither:
Going to close this issue for now, since currently sms campaigns redirect to this campaign.

Issue #4162, opened 2015-03-11
add feature flag for share language and CTA on confirmation page


mikefantini:
we don't want any of these to show until the Permalink - shares on confirmation page milestone is closed:
- Social header on reportback permalink page.
- Social copy on reportback permalink page.
- Social cta on reportback permalink page.
- Social buttons

![image](https://cloud.githubusercontent.com/assets/6330971/6599641/568197a6-c7e1-11e4-8343-ba1a2c5c2b2b.png)

PR #4166, opened 2015-03-11
Removed redirect to old confirmation page.

angaither:
Doesn't remove the confirmation page, just removes the redirect to the page after submitting a reportback
Refs #4095

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -320,22 +320,16 @@ function dosomething_reportback_form_submit($form, &$form_state) {
   // Save values into reportback.
   $rid = dosomething_reportback_save($values);
 
-  // Redirect to confirmation.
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -320,22 +320,16 @@ function dosomething_reportback_form_submit($form, &$form_state) {
   // Save values into reportback.
   $rid = dosomething_reportback_save($values);
 
-  // Redirect to confirmation.
-  $display_permalink = variable_get('dosomething_reportback_display_permalink');
```
aaronschachter, line 324: Should we kill this variable from the db and config form as well?

angaither, line 324: I'm going to have another pr to change that to `display_share_form` today
#4162

aaronschachter:
:+1:

PR #4169, opened 2015-03-12
Permalink sharing take two

angaither:
This includes the frontend of the share that was previously merged & unmerged in
As well as a feature flag to hide sharing if merged into prod.

@DoSomething/front-end 
@aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
@@ -22,11 +22,11 @@ function dosomething_reportback_admin_config_form($form, &$form_state) {
     '#default_value' => variable_get('dosomething_reportback_log', FALSE),
     '#description' => t("Logs Reportback activity. This should be disabled on production."),
   );
-  $form['dosomething_reportback_display_permalink'] = array(
+  $form['dosomething_reportback_display_permalink_share'] = array(
     '#type' => 'checkbox',
-    '#title' => t('Enable permalink page.'),
-    '#default_value' => variable_get('dosomething_reportback_display_permalink', FALSE),
-    '#description' => t("Displays the permalink/confirmation page instead of the confirmation page after submitting a reportback."),
+    '#title' => t('Enable sharing on the permalink page.'),
+    '#default_value' => variable_get('dosomething_reportback_display_permalink_share', FALSE),
+    '#description' => t("Displays the permalink share form/copy"),
   );
   return system_settings_form($form);
 }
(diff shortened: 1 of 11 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -327,3 +327,24 @@ function paraneue_dosomething_get_themed_placeholder_reportbacks($count = 6) {
 
   return $placeholders;
 }
+
+// @TODO - NOT DONE YET - Just a first pass a creating these share URLs.
+function paraneue_dosomething_facebook_intent($link) {
+  $facebook_app_id = '1374507732873066';
```
aaronschachter: Just checking -- is this safe to put here?

DFurnes:
Shippy ship ship. :shipit:

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
```diff
@@ -22,11 +22,11 @@ function dosomething_reportback_admin_config_form($form, &$form_state) {
     '#default_value' => variable_get('dosomething_reportback_log', FALSE),
     '#description' => t("Logs Reportback activity. This should be disabled on production."),
   );
-  $form['dosomething_reportback_display_permalink'] = array(
+  $form['dosomething_reportback_display_permalink_share'] = array(
     '#type' => 'checkbox',
-    '#title' => t('Enable permalink page.'),
-    '#default_value' => variable_get('dosomething_reportback_display_permalink', FALSE),
```
aaronschachter, line 28: Its OCD, but can we delete any variables we are no longer using in an update N hook?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -443,6 +443,17 @@ function dosomething_reportback_view_entity($entity) {
     $subtitle = ($is_owner) ? $copy_vars['owners_rb_subtitle'] . " " . $incentive : $node->call_to_action;
     $link = '/' . drupal_get_path_alias('node/' . $node->nid);
 
+    // Set twitter text.
+    $twitter_text = token_replace(variable_get('dosomething_campaign_permalink_confirmation_owners_social_network_copy'), array('node' => $node));
```
aaronschachter, line 447: Just curious -- is this `owners_social_network_copy` variable only used as Twitter message ?

sbsmith86, line 447: So for twitter, it has to be used explicitly in the intent link. For Facebook, I think it is used in the og tag, and for Tumblr, I think we will have to use it in the intent link as well.

angaither, line 447: no, it's for facebook and tumblr as well

angaither:
@aaronschachter added setting for facebook id & an update hook to remove the unused setting

aaronschachter:
:+1:

PR #4171, opened 2015-03-12
Adding facebook sharing

sbsmith86:
Adding facebook share link and og tags. 

@DoSomething/front-end @angaither @aaronschachter


angaither:
:elephant: :shipit: value added. +1 :dancer:

PR #4174, opened 2015-03-12
non-owner permalink updates

### LINKED ISSUE #4153: Handle display of solution fact/statement/supporting statement
`If (fact) { print fact; }
elseif (statement) { print statement; }`

`if (supporting statement) { print supporting statement; }`

### LINKED ISSUE #4173: Add bottom border back on social cta module
I originally removed the bottom border so that blue hover background on the social media buttons would go right to the edge, but if the text in the card is to short it creates this effect: 

image

So, I need to add it back.

sbsmith86:
## Changes
- Commenting helper functions I made for social
- If there is a solution fact, show that, otherwise show the solution statement. Always should the solution supporting statement, if it exists. Fixes #4153 
- Add the bottom border back to the share cta. Fixes #4173 

@DoSomething/front-end


DFurnes:
:shipit:!

PR #4179, opened 2015-03-13
Issue copy paster errorz

### LINKED ISSUE #4176: Update issue in permlink share copy
@mikefantini says

> it looks like cause is being pulled in and not issue; again I used [node:issue] but if that's wrong let me know or I can create an issue for this token to pull from a different field

@lkpttn  says

> also, lower case the issue

angaither:
Fixes #4176


aaronschachter:
:+1:

PR #4182, opened 2015-03-13
Token replace social share copy

### LINKED ISSUE #4181: Token replace social share variable
from @mikefantini 

> it seems like the title token isn't working properly, although I may have used the wrong one. Let me know if we need an issue or if I used the wrong token.

angaither:
Fixes #4181


aaronschachter:
:+1:

PR #4184, opened 2015-03-13
display secondary cta above cta button

### LINKED ISSUE #4183: use secondary call to action above Do It button on non-owner permalink page
Text above the button should pull from field_secondary_call_to_action

image

sbsmith86:
Display the secondary cta above the cta button on non-owner permalink pages so that it doesn't display the same thing as the subtile of the page. Fixes #4183 

@DoSomething/front-end


DFurnes:
:+1:

PR #4193, opened 2015-03-13
Get the most recently uploaded rb item by default.

### LINKED ISSUE #4192: original photo shows on confirmation page if user just updates submission w/o new photo
steps to recreate:
- report back
- update your submission w/ a new photo
- update your submission a third time w/o a new photo

Do this and the original photo shows, which is slightly confusing. If this is more than like an hour to fix let's leave it, but if it's a quick fix, let's do it.

angaither:
Fixes #4192


aaronschachter:
:+1: value added

Issue #4197, opened 2015-03-16
Permalink sign up button should also work for un-authenticated users


angaither:
Currently the signup is only created if the user is signed in, because only the markup for the modal link is returned with `dosomething_signup_get_signup_button`

angaither:
Fixes with #4198

PR #4198, opened 2015-03-16
Permalink as pitch page

### LINKED ISSUE #4115: Beta RB permalink should store the Reportback File fid as Signup Source
When a beta signs up for a campaign via the alpha's permalink for rbf/[fid], the beta's signup source should be stored as the rbf/[fid], to track the number of betas the alpha has gotten to join the campaign.

Currently sources as stored as the canoical paths of the referring entity:
`node/[nid]`
`taxonomy/term/[tid]`
`campaigns`

### LINKED ISSUE #4154: non-owner permalink: button should read "SIGN UP" when it serves as pitch
just making a note that once we update the non-owner permalink to serve as the actual campaign pitch page, we need to change the button copy from DO THIS to SIGN UP

### LINKED ISSUE #4190: Redirect to action page after signup
After a user signs up from a permalink, the user should be redirected to the campaign action page

angaither:
This turns the `do it` button on a permalink page into the `Sign Up` button
Currently, this only supports Authenticated Users (and shouldn't be merged just yet)

This also includes some light refactoring and of course, ever present :space_invader: fixes

Fixes #4154 
Fixes #4190 
Fixes #4115

@aaronschachter I created #4197 for the anon users, but it doesn't seem like that get_signup_button is the way to go for that. do you have any feels on how to handle that?

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -167,7 +167,7 @@ function dosomething_campaign_scholarship_block_content() {
       unset($rows[$key]['data']['src']);
       $rows[$key]['data']['amount'] = '$' . $scholarship['amount'];
       $rows[$key]['data']['deadline'] = $scholarship['deadline'];
-      $rows[$key]['data']['image'] = "<img src=" . $rows[$key]['data']['image'] . "</img>"; 
+      $rows[$key]['data']['image'] = "<img src=" . $rows[$key]['data']['image'] . "</img>";
     }
 
     $vars['header'] = array('nid', 'title', 'cta', 'image', 'path', 'pretty path', 'staff pick', 'amount', 'deadline');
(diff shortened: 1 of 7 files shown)


aaronschachter:
Seems ok. What is the issue in using `get_signup_button` to display the login/reg form for anon users, @angaither ?

angaither:
@aaronschachter the thing I've been stuck on is getting the redirect after logging in to work.
I added [this](https://github.com/DoSomething/dosomething/pull/4198/files#diff-6cd2bd158ecf1b2419b491fc3dae7ceaR519) line to do the redirect, but the [user module redirect](https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_user/dosomething_user.module#L549) after logging in overrides the redirect that I set in this hook.

aaronschachter:
@angaither Does it work to change the User module's redirect to only execute based on whether or not they were on a Campaign page or a Reportback ?

angaither:
@aaronschachter no, for some reason `dosomething_reportback_signup_submit`  isn't getting executed at all on a registration/sign in. it's only getting called when the user is already logged in. 
I've tried scoping the redirected to not run on reportbacks, and even commenting it out entirely (when I do that it redriects to `/user`)

angaither:
ahhhhhHA,
so I think the custom handler is getting added to only the top login/register form
if I click the `login` button on the top of the page, the redirect work, but the actual signup button that pulls up the modal doesn't work. 
the modal login forms must be handled differently...

angaither:
ok, now this works cc @aaronschachter

aaronschachter:
Looks good! :+1:

PR #4201, opened 2015-03-16
Don't convert to table display until we are tablet. Fixes #4196

### LINKED ISSUE #4196: share cta region runs over on mobile
iphone 5:

image

sbsmith86:
I only need the `.card__column` to display as a table at the tablet breakpoint so that I can use `table-footer-group` to display the cta at the bottom of the card instead of the top on mobile.

@DoSomething/front-end


DFurnes:
:+1:

PR #4211, opened 2015-03-16
Always load fids from a reportback.

### LINKED ISSUE #4203: Permalink page error in metatag module
`Fatal error: Call to a member function getImageURL() on a non-object in /var/www/vagrant/lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module on line 128`

http://staging.beta.dosomething.org/reportback/685

angaither:
Metatag module was attempting to call `file->getImageURL` on not a file.
Because of _science_ `reportback_load` doesn't always grab the fids from the entity.
This calls the function in the reportback class to manually get the fids.

`¯\_(ツ)_/¯`
Fixes #4203

### DIFF
--- a/lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
+++ b/lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
@@ -121,8 +121,11 @@ function dosomething_metatag_get_permalink_vars() {
   $rbid = arg(1);
   $fid = ($_GET['fid']);
   $rb_entity = reportback_load($rbid);
-  if (!$fid) {
-    $fid = $rb_entity->fids[0];
+  if (!$fid && !empty($rb_entity->fids)) {
+    $fid = array_pop($rb_entity->fids);
+  }
+  else {
+    $fid = array_pop($rb_entity->getFids());
   }
   $file = reportback_file_load($fid);
   $image = $file->getImageURL('large');
--- a/lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
+++ b/lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
@@ -121,10 +121,7 @@ function dosomething_metatag_get_permalink_vars() {
   $rbid = arg(1);
   $fid = ($_GET['fid']);
   $rb_entity = reportback_load($rbid);
-  if (!$fid && !empty($rb_entity->fids)) {
-    $fid = array_pop($rb_entity->fids);
-  }
-  else {
+  if (!$fid) {
     $fid = array_pop($rb_entity->getFids());
   }
   $file = reportback_file_load($fid);


lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
```diff
@@ -121,8 +121,11 @@ function dosomething_metatag_get_permalink_vars() {
   $rbid = arg(1);
   $fid = ($_GET['fid']);
   $rb_entity = reportback_load($rbid);
-  if (!$fid) {
-    $fid = $rb_entity->fids[0];
+  if (!$fid && !empty($rb_entity->fids)) {
+    $fid = array_pop($rb_entity->fids);
+  }
+  else {
+    $fid = array_pop($rb_entity->getFids());
```
aaronschachter, line 125: Is there a fix somewhere in https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc#L29 that would help make this more readable / understandable / non-science ?

angaither, line 125: the entity actually contains the fids at the end of the `__construct` function, which is why I'm just very confused.

angaither, line 125: two dpms, one from the end of `__construct` and the other inside `dosomething_metatag_get_permalink_vars`

aaronschachter, line 125: Yeah, that's really strange.  I'm taking a look too to see if i can find any clues.

angaither, line 125: Right?! I've been banging my head against the wall on this one today. I haven't been able to figure out why it happens only sometimes. It's _very_ odd.

aaronschachter, line 125: So I find that if i hardcode the rbid to an integer in `dosomething_metatag_get_permalink_vars` and call it within `devel/php`, I see the fids array when i output it.  So i think there is something real weird happening taking place calling the `reportback_load` in a `hook_preprocess_page`.

I also tried outputting `vars` in there, to see if the loaded reportback was already in there, but it's not. i think if the template was using some default Entity template type stuff vs a theme function, we'd get those variables for free in the `$vars`.   The theme function seems to be using `$rb` as just an array, not a loaded `reportback` entity..

aaronschachter, line 125: It seems at least a more readable fix could be to always `array_pop($rb_entity->getFids());` and avoid the `if`/`else`... but it would def be nice to have this in the one singular place, whether it be a property on a Reportback or a function you always need to call on a loaded Reportback, and to not really chalk this one up to "who knows whats going on"..

angaither:
@aaronschachter updated it to call `array_pop($rb_entity->getFids());` in the metatag module

aaronschachter:
Ok to ship as a hotfix... the ideal would be to figure out how to theme this as a proper entity to be able to access entity variables in the preprocess vars. It's going to be very cumbersome to always have to reload the entity(ies) anytime we want to preprocess this page elsewhere.

angaither:
I agree, but this is blocking deployment, and I've already spent a full day on it

aaronschachter:
Yep, I hear that. Can you create an issue referring this PR for refactoring the permalink template so it can properly be preprocessed?

PR #4212, opened 2015-03-16
Updates to permalink page on mobile

### LINKED ISSUE #4209: updates to mobile confirmation page
- [ ] add in Social cta on reportback permalink page.
- [ ] remove Social header on reportback permalink page.
- [ ] remove Social copy on reportback permalink page.

sbsmith86:
## Fixes #4209
- Show share cta on mobile 
- Hide share copy on mobile.

@DoSomething/front-end

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
@@ -36,11 +36,14 @@
   }
 
   .card__copy {
-    .-share {
-      display: none;
+    &.-share {
+      h3:last-of-type,
+      p:last-of-type {
+        display: none;
 
-      @include media($tablet) {
-        display: block;
+        @include media($tablet) {
+          display: block;
+        }
       }
     }
   }
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
@@ -34,19 +34,6 @@
       }
     }
   }
-
-  .card__copy {
-    &.-share {
-      h3:last-of-type,
-      p:last-of-type {
-        display: none;
-
-        @include media($tablet) {
-          display: block;
-        }
-      }
-    }
-  }
 }
 
 //@TODO - Move to Neue as a modifier for the cta pattern
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
@@ -26,14 +26,6 @@
       border-top: 2px solid $light-gray;
     }
 
-    > .wrapper {
-      display: none;
-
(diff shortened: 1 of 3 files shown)


lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
```diff
@@ -26,22 +26,23 @@
...
-        display: block;
+        padding: $base-spacing;
       }
     }
+  }
 
-    .social-icon {
-      padding: $base-spacing / 2;
+  .card__copy {
+    .-share {
```
DFurnes: We need modifiers to be directly attached to a class to avoid naming conflicts, so I'd change this to `&.-share` and then maybe style any `h3` or `p` children.

sbsmith86: @DFurnes oh i see, I thought the scoping the `.card__copy` was enough. I can update!

DFurnes: It probably wouldn't be an issue, but the danger is if we nested something inside the `.card__copy` that also had a `.-share` modifier, it would also inherit these styles. So better safe than sorry! :smile:

sbsmith86:
@DFurnes updated!

lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
```diff
@@ -41,7 +33,19 @@
         padding: $base-spacing;
       }
     }
+  }
+
+  .card__copy {
+    &.-share {
+      h3:last-of-type,
```
DFurnes: Hmm, I'm a little wary about using `:last-of-type` here in case we were to add more markup in sometime later. I'm wondering if actually adding some global utility classes like `.show-only-mobile`, `.show-only-tablet` might be a better approach? (Sorry to be so picky/indecisive.)

DFurnes: We seem to have a decent amount of situations where we want things to hide/appear on mobile (such as homepage headers) that don't really make sense to make specific CSS rules for each scenario. Perhaps something to merge in for a quick fix now, and then refactor at the Neue level in the near future.

sbsmith86:
@DFurnes I like that idea of the .show-only classes! I think that would be great value added.

lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
weerd: device agnostic names tho?

DFurnes: :japanese_ogre: 

Yes!

sbsmith86: @weerd @DFurnes I think I can give this a shot in Neue tomorrow.

DFurnes:
Fab! Much cleaner. :ship:

PR #4215, opened 2015-03-16
Updating to allow for property set in plain text.

### LINKED ISSUE #4214: With new environment PHP throwing 'illegal string offset' error
errors

Need to check for value existing and if so then assign it, otherwise default due to plain text values for some fields.

weerd:
### Fixes #4214

If the contents of the `$value` variable contain an array then check for the `safe_value` otherwise default to setting the property as just the plain text `$value`.

@aaronschachter @sergii-tkachenko


aaronschachter:
:+1:

sergiitk:
Awesome!

PR #4217, opened 2015-03-17
Changing reserved variable names to unique names

### LINKED ISSUE #4216: IE8 share bug
IE8 doesn't like that we have a variable named `top` in our small script that allows use to open the share links in a pop up window. So I need to rename that variable in order to get the pop up script to work. 

Nice little read up on variable names you shouldn't use http://www.nczonline.net/blog/2007/06/03/javascript-variable-names-you-shouldn-t-use/

sbsmith86:
## Fixes #4216

We can't use variable names like `top` in IE8 since it is a reserved name in the outer most window object, and IE8 doesn't ignore it like other browsers. 

@DoSomething/front-end

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/app.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/app.js
@@ -106,15 +106,15 @@ define("app", function (require) {
 
     $('body').on('click', 'a.js-share-link', function(event) {
       event.preventDefault();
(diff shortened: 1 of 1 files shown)


DFurnes:
The more you know! :dizzy: Ship it!

weerd:
Yeah, good to know for future work :) Excellent find!

PR #4222, opened 2015-03-17
Need to fallback if error loading image.

weerd:
Not actually sure if returning `NULL` is ideal here; maybe a different fallback would be more ideal? This cropped up specifically due to issues with the **Stage File Proxy** module not properly loading files. This would cause errors in the logs since the `$image` variable was empty, and thus results in this notice:

```
Notice: Trying to get property of non-object in dosomething_image_get_themed_image_by_fid() (line 147 of /var/www/dev.dosomething.org/lib/modules/dosomething/dosomething_image/dosomething_image.module).
```

However, because it's now returning `NULL`, doesn't really provide much feedback on the front-end that the image link is broken. Any ideas on a better fallback would be appreciated.


aaronschachter:
One idea is a placeholder image like the DS logo, or possibly have our own little broken sad face image or something to use (to differentiate from any scenarios where we legitimately would want to display the DS logo)...

weerd:
Yeah, that sounds good to me. @lkpttn any suggestions for an image to use? We have a logo we could use via Neue https://github.com/DoSomething/neue/blob/dev/assets/images/logo.png

lib/modules/dosomething/dosomething_image/dosomething_image.module
```diff
@@ -139,7 +139,12 @@ function dosomething_image_get_themed_image_by_fid($fid, $style, $alt = '') {
   if (!$file) {
     return NULL;
   }
+
   $image = image_load($file->uri);
+  if (!$image) {
+    return NULL;
```
sergiitk: `@return` statement should be fix in order to represent possible `NULL` value in return.

aaronschachter:
This could potentially be used as the default for all of these `dosomething_image_get` functions, not just `dosomething_image_get_themed_image_by_fid`

lkpttn:
I can't fully grasp the context of where this is, is it a failure to load a reportback image?

sergiitk:
Maybe provide default pic instead of null? It would cover cases when null response from this function is not expected, as in [`dosomething_campaign_preprocess_closed_reportback_gallery()`](https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc#L392)?

weerd:
@lkpttn 

Success on image load:
pasted_image_at_2015_03_17_11_40_am_720

Error on image load:
pasted_image_at_2015_03_17_11_40_am_360

angaither:
I think we should use the @mikefantini image on images.ds.org :new_moon_with_face:

weerd:
@sergii-tkachenko yeah, it seems we're all in agreement on a fallaback image from the `/images` directory in the DS App :)

mikefantini:
@lkpttn maybe let's go with a photo of the puppet + some short copy explaining that there's no image? 

only trick is copy would have to be somewhat legible for the smallest sizes

@weerd can you provide the sizes we would need to use this for? Also would we need square and landscape versions?

weerd:
Let's start with square for now and make more as needed. I think it basically ranges from `480 x 480` to `100 x 100`. Basically what we see in the reportback gallery example above.

We only need one image at `480 x 480` but keeping in mind that it can get resized down to `100 x 100`. Responsive sites, yo!

weerd:
Using this for now until @lkpttn gets me a new image :dancer: 

iyh7g

weerd:
Example output:

image

@mikefantini :) Once I get a new image from @lkpttn I can replace the "yuno"..... unless..... "yuno" works ;)

weerd:
@aaronschachter @sergii-tkachenko I've updated the code. Using the "yuno" image for now and won't merge until @lkpttn passes me a new official one to use. Just wanted to get the code fix up so y'all could check it out!

lib/modules/dosomething/dosomething_image/dosomething_image.module
```diff
@@ -139,16 +139,23 @@ function dosomething_image_get_themed_image_by_fid($fid, $style, $alt = '') {
...
-    'file' => array(
-      '#theme' => 'image_style',
-      '#style_name' => $style,
-      '#path' => $image->source,
-      '#alt' => $alt,
-    ),
-  );
-  return drupal_render($content);
+  if (!$image) {
+    return '<img src="' . DS_ASSET_PATH .'/dist/images/no-image-found.jpg" title="No Image Found" />';
```
sergiitk: Maybe [`theme_image()`](https://api.drupal.org/api/drupal/includes!theme.inc/function/theme_image/7) instead?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -99,7 +99,6 @@ function dosomething_reportback_form($form, &$form_state, $entity = NULL) {
       $reportback_submissions_data[$index] = reportback_file_load($fid);
       $reportback_submissions_data[$index]->image = dosomething_image_get_themed_image_by_fid($fid, '300x300');
     }
-
```
sergiitk, line 102: Revert, this file is not related.

sergiitk:
:+1:

sergiitk:
U DA MAN, DIEGO! :boom:

weerd:
thanks for the `revert`ing tip @sergii-tkachenko!

lkpttn:
pixelsloth

weerd:
Merging with new image!

Issue #4210, opened 2015-03-16
FB window should close after I post


mikefantini:
When sharing from the confirmation page, the FB window should close after I post to FB. This is how it works when I share from Twitter.

Currently, the window is staying open and reloading the permalink page after I post to FB. So I then have two windows (or two tabs on mobile) of the permalink page.

This then causes further issues if I try to share on another social channel after I share on FB (on mobile it just updates the "share tab" but it doesn't open it so the user is not made aware).

As a side note, Tumblr keeps the second window/tab open and just does it's celebratory confetti thing with no next step.

mikefantini:
@sbsmith86 just went through the FB share flow on buzzfeed and it has the behavior we are looking for here.

sbsmith86:
@mikefantini : so i have a fix for the fb/window issue going to staging. It doesn't look like we are going to be able to easily close the popup window on mobile but it does on desktop. However, I made an update that opens a new window everytime one of those social buttons are pressed so that we don't run into the issue where the window doesn't seem to open when going back to share on another network.

mikefantini:
got it thanks @sbsmith86. I'll play with it and let you know if I have any adjustments, but what you described sounds like a smart solution

mikefantini:
@sbsmith86 - weird so the window is now closing on desktop but not on mobile

is the intended experience that I stay on FB on mobile vs. being taken back to DS? I could see either making sense, but I guess it makes more sense to have one window stay open of DS and one window remain on FB if we can't close that second one out.

sbsmith86:
@mikefantini yea so, I can't get the second window to close on mobile without some real hackery. So the user would have to go back to the original tab, but since each share button opens a new window they shouldn't have an issue sharing on a different network if they choose.

PR #4224, opened 2015-03-17
Using a different intent link for facebook

sbsmith86:
Updated the intent link to use /sharer/share.php instead. Hopefully that will resolve the redirect issue #4210 

@DoSomething/front-end


DFurnes:
Let's give it a go! :ship:

PR #4226, opened 2015-03-17
Renames dosomething_mobilecommons_* variables.

sergiitk:
This PR renames `dosomething_mobilecommons_*` variables to `dosomething_signup_mobilecommons_*` because `dosomething_mobilecommons` module does not exists anymore.
This needed for consistency of further development, such as introducing new similar variables as part of in #4058.


deezone:
:+1:

PR #4227, opened 2015-03-17
Open a new window everytime

sbsmith86:
## Changes

Add `_blank` to the window so that a new popup is produced everytime. Otherwise (on mobile safari), if a user clicks one share button and goes through that process, when they go back to the original reportback tab and click to share again the page loads in the same window but doesn't come into focus which is confusing.

@DoSomething/front-end


DFurnes:
:ship:!

angaither:
:house: :mailbox_with_no_mail:

PR #4233, opened 2015-03-18
Anchor card cta at bottom of the card

### LINKED ISSUE #4194: anchor buttons to the bottom of the permalink
image

image

sbsmith86:
## Fixes #4194

Made the height of the wrapper of the card column with the text & cta to be 100%. This way, when the cta moves to the bottom of the column on tablet, it will always be at the very bottom of the card no matter how much text is in the copy element above it. 

@DoSomething/front-end

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
@@ -11,10 +11,20 @@
     }
   }
 
-  .card__column > .wrapper {
-    @include media($tablet) {
-      display: table;
-      border-collapse: collapse;
+  .card {
+    height: 100%;
+  }
+
+  .card__column {
+    height: 100%;
+
+    > .wrapper {
+      height: 100%;
+
+      @include media($tablet) {
+        display: table;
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback-permalink.scss
```diff
@@ -11,10 +11,20 @@
...
+
+  .card__column {
+    height: 100%;
+
+    > .wrapper {
+      height: 100%;
+
+      @include media($tablet) {
+        display: table;
+        border-collapse: collapse;
```
weerd, line 26: Is the `border-collapse` necessary? I thought you really only need this if actually using a `<table>` in the markup?

sbsmith86, line 26: @weerd yea, we need it. Without it, the border around the cta doesn't show when it gets set to `display: table-footer-group`

weerd, line 26: Ahhh, ok!

weerd:
:+1:

PR #4236, opened 2015-03-18
Add link to permalink from user profile

### LINKED ISSUE #4218: Access permalink share page from user profile
Admins currently have a link to the admin view, that should remain, but for a user it should be replaced with the link to the public facing permalink

angaither:
Links x verbs nouned to the public permalink page for non-admins.
Leaves the existing link (for admins) to admin reportback there.
Fixes #4218


aaronschachter:
:+1:

PR #4239, opened 2015-03-18
Move non-owner cta to correct spot in tpl

### LINKED ISSUE #4194: anchor buttons to the bottom of the permalink

sbsmith86:
## Fixes #4194

I forgot to move the cta for non-owners to the right place in the template, and target it the same as the share cta in the styling for the permalink page.

@DoSomething/front-end


weerd:
hokay! :+1:

PR #4241, opened 2015-03-18
Updates CTools to 1.7.

sergiitk:
Fixes https://www.drupal.org/node/2454909.

Build worked well, site seems to be as usual.


aaronschachter:
:+1: thanks for looking into this!

PR #4243, opened 2015-03-18
Signup Group endpoints for Inviter app prototype

### LINKED ISSUE #4238: Signups API endpoint
New API endpoint needed for Mobile app prototyping.  required functionality:
- Post a signup with only a UID and a given source. If the given source is a signup entity e.g. `signup/123` then load Signup 123 and find its corresponding campaign nid. Endpoint should create the signup recordfor the given UID and the source's nid.
- GET all signups with a given source. e.g. GET `signups?source=signup/123` but also include the alpha signup 123 within result set

Background:
The mobile app is looking to introduce functionality that allows Users to join a Signup Group. Current spec is as follows:
- An alpha signs up for a campaign and receives an Invite Code. This is their signup sid.
- The alpha invites betas to signup for a campaign using their Invite Code. The beta signup is psoted with the alpha source, `signup/123`.
- All  betas who have signed up with `signup/123` or whose signup ID is actually 123 (the alpha) are provided with an aggratgate view of reportback items and overall quantity

Desired behavior is that there is a view within the mobile app where a user can simply enter an INvite Code (without viewing the campaign it belongs to).  Entering the Invite Code will signup the user for the campaign its associated with, and then present the campaign (with a view of all users who have used the same invite code)

aaronschachter:
Closes #4238 (details are listed in that issue).
- POST to new `api/v1/signups` endpoint with only a `source` string, if the `source` is in the format `signup/[sid]`. The `dosomething_signup_get_target_nid_for_source` function checks a given `source` string and gets the relevant node nid it is used to create the signup for.
- GET to new `api/v1/signups` endpoint with a parameters array:
  -- If a `signup/[sid]` string is passed to `parameters[source]`, return all Signups with that `signup/sid` as the source, including the alpha sid.

I know it's kind of weird to return the alpha signup as well as the betas, and we can clean this up with perhaps a better function name / endpoint if we go this route. Also open to different suggestions. 

The main purpose of adding this is not to make it perfect, as it won't be used in production for a long time, but to prototype whether or not the Alpha > Beta Signup Group per campaign flow will work in the Inviter app.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
@@ -91,3 +91,35 @@ function _signup_resource_create($data) {
     services_error($e);
   }
 }
+
+/**
+ * Callback for Signups index.
+ *
+ * @param array $parameters
+ *   Array passed within query string. Possible keys:
+ *   - source (string).
+ *   - nid (int).
+ *
+ * @return mixed
+ *   Object of the newly created user if successful. String if errors.
+ */
+function _signup_resource_index($parameters) {
+  // Initialize output.
+  $index = array();
+
+  $query = db_select('dosomething_signup', 's');
+  $query->fields('s');
+  if (isset($parameters['source'])) {
+    $query->condition('source', $parameters['source']);
+  }
+  if (isset($parameters['nid'])) {
+    $query->condition('nid', $parameters['nid']);
+  }
+
+  // @todo: We still need to figure out how to include the alpha.
+  // This is only returning the betas that used the alpha signup as source.
+
+  // Return max of 25 records for now.
+  $query->range(0, 25);
+  return $query->execute()->fetchAll();
+}
--- a/lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
@@ -97,21 +97,30 @@ function _signup_resource_create($data) {
  */
 function _signup_resource_index($parameters) {
   // Initialize output.
-  $index = array();
+  $result = array();
+  $alpha_sid = NULL;
 
   $query = db_select('dosomething_signup', 's');
-  $query->fields('s');
+  $query->fields('s', array('sid'));
   if (isset($parameters['source'])) {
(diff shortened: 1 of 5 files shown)


aaronschachter:
Sample request:
`http://dev.dosomething.org:8888/api/v1/signups.json?parameters[source]=signup/54`

Output:

```
[  
   {  
      "sid":"54",
      "uid":"101",
      "nid":"362",
      "timestamp":"1394654653",
      "signup_data_form_timestamp":"1397683873",
      "signup_data_form_response":"1",
      "why_signedup":null,
      "source":null
   },
   {  
      "sid":"55",
      "uid":"106",
      "nid":"362",
      "timestamp":"1394659408",
      "signup_data_form_timestamp":"1397684404",
      "signup_data_form_response":"1",
      "why_signedup":null,
      "source":"signup/54"
   },
   {  
      "sid":"1621",
      "uid":"5",
      "nid":"362",
      "timestamp":"1426706731",
      "signup_data_form_timestamp":null,
      "signup_data_form_response":null,
      "why_signedup":null,
      "source":"signup/54"
   }
]
```

aaronschachter:
cc @jonuy

lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
```diff
@@ -0,0 +1,126 @@
...
+ */
+function _signup_resource_index($parameters) {
+  // Initialize output.
+  $result = array();
+  $alpha_sid = NULL;
+
+  $query = db_select('dosomething_signup', 's');
+  $query->fields('s', array('sid'));
+  if (isset($parameters['source'])) {
+    $query->condition('source', $parameters['source']);
```
angaither: maybe I'm missing it somewhere else in the pr here, but make sure to clean the user input before querying with it.

angaither:
@aaronschachter other than the possible security concern :+1: looks ok for now/prototype

aaronschachter:
Muchas gracias, @angaither ! Will be sure to strip the `source` param before merge.

lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
aaronschachter: True blue

PR #4245, opened 2015-03-19
Proposed solution for #4244.

### LINKED ISSUE #4244: PHP 5.5 Illegal string offset warnings
```
Warning: Invalid argument supplied for foreach() in element_children() (line 6487 of /var/www/dev.dosomething.org/html/includes/common.inc).
Warning: Illegal string offset '#children' in drupal_render() (line 5952 of /var/www/dev.dosomething.org/html/includes/common.inc).
Warning: Illegal string offset '#children' in drupal_render() (line 5962 of /var/www/dev.dosomething.org/html/includes/common.inc).
Warning: Illegal string offset '#children' in drupal_render() (line 6000 of /var/www/dev.dosomething.org/html/includes/common.inc).
Warning: Illegal string offset '#printed' in drupal_render() (line 6007 of /var/www/dev.dosomething.org/html/includes/common.inc).
DoSomething.org
```

image

sergiitk:
Fixes #4244.
See http://stackoverflow.com/questions/14123628/drupal-force-regions-to-always-render

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/template.php
+++ b/lib/themes/dosomething/paraneue_dosomething/template.php
@@ -104,7 +104,7 @@ function paraneue_dosomething_js_alter(&$js) {
 function paraneue_dosomething_page_alter(&$page) {
   // Ensure we have a 'page_bottom' region
   if (!isset($page['page_bottom'])) {
-    $page['page_bottom'] = array('region' => 'page_bottom');
+    $page['page_bottom'] = array();
   }
 
   paraneue_dosomething_page_alter_login($page);


sergiitk:
@DFurnes, as I understand `page_bottom` region should be responsible for modal dialogs, see `paraneue_dosomething_page_alter_login()` and `paraneue_dosomething_page_alter_register()`. Please ensure this works right.

DFurnes:
Yep, that's correct. Have you tested that modals are still properly added to the page with this change?

sergiitk:
Yes. Errors disappeared and the modals still work.

DFurnes:
Cool, let's ship it then! :shipit:

PR #4247, opened 2015-03-19
Add markup/styles for SMS first names Optimizely test.

### LINKED ISSUE #3568: add beta name fields for SMS game A/B test
# Tasks
- [x] Create and output a second SMS Game Signup Form with extra input values for names, which is hidden on the page. ideally using same form constructor with parameter set to include names?
- [x] Second SMS Game Signup Form needs to add additional beta names to submitted MobileCommons request

The Optimizely test can hide the regular Form and unhide the hidden Form to A/B it up.
# Overview

we want to run an A/B test where we ask alphas for the first name of their beta friends

we would need to pass this first name along to mobile commons to be associated with the beta

[mobile mocks](https://s3.amazonaws.com/uploads.hipchat.com/34218/350105/42fQwILikUk4mG0/Mobile%20SMS.png)
[desktop mocks](https://s3.amazonaws.com/uploads.hipchat.com/34218/350105/kg5HrkAu4cMln8t/SMS%20Game.png) -> "First Name" should be the placeholder text in the alpha's first name field.

cc: @jonuy @Bladt @weerd

DFurnes:
# Changes
- Adds markup and styles for SMS first names Optimizely test. Closes #3568.
- Add hidden fields & whitelist them to be passed to Mobile Commons. We only validate them client-side, so Drupal doesn't need to know whether or not A/B test is functioning on page.
- Removes CSS column groups, instead floating individual fields as half-width blocks. This allows us to have adjacent fields side-by-side rather than stacked vertically (needed for friend name & number fields to be side-by-side in the A/B variant).

For review: @aaronschachter @DoSomething/front-end @jonuy 
# Snippet

``` js
var $ = jQuery;
var $form = $("#dosomething-signup-sms-game-form");
$form.addClass('optimizely-sms-form');
$form.find(".form-item-alpha-first-name").remove();
$form.find("#edit-group1").html($("#ab-test-sms-names").html()); 
```

(You can test this by copying and pasting into the web console on staging or your dev environment!)
# Updated Example Payload

The Node application will receive the exact same payload, but may include the extra `beta_first_name_#` fields if the test is activated on the page. For example:

```
array(10) { ["alpha_first_name"]=> string(4) "Dave" ["alpha_mobile"]=> string(10) "2035830429" ["beta_mobile_0"]=> string(10) "2025555692" ["beta_mobile_1"]=> string(10) "2025555672" ["beta_mobile_2"]=> string(10) "2025455692" ["beta_first_name_0"]=> string(3) "Jim" ["beta_first_name_1"]=> string(3) "Don" ["beta_first_name_2"]=> string(3) "Bob" ["story_id"]=> string(6) "176153" ["story_type"]=> string(6) "176155" }
```
# Screenshots
#### Before:

screen shot 2015-03-19 at 11 24 48 am
#### After (with A/B test activated):

screen shot 2015-03-19 at 11 25 42 am

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
@@ -167,13 +167,7 @@ function dosomething_signup_sms_game_form($form, &$form_state, $node, $num_frien
   $form['group1'] = array(
     '#type' => 'fieldset',
     '#attributes' => array(
-      'class' => array('-columned first-group'),
-    ),
-  );
-  $form['group2'] = array(
-    '#type' => 'fieldset',
-    '#attributes' => array(
-      'class' => array('-columned second-group'),
+      'class' => array('campaign-sms-numbers'),
     ),
   );
   $form['group1']['alpha_mobile'] = array(
@@ -189,13 +183,7 @@ function dosomething_signup_sms_game_form($form, &$form_state, $node, $num_frien
     ),
   );
   for ($i = 0; $i < $num_friends; $i++) {
-    // Default to second fieldset.
-    $group = 'group2';
-    if ($i == 0) {
-      // Unless its the first beta.
-      $group = 'group1';
-    }
-    $form[$group]['beta_mobile_' .$i] = array(
+    $form['group1']['beta_mobile_' .$i] = array(
       '#type' => 'textfield',
       '#required' => TRUE,
       '#title' => 'Friend\'s Cell Number',
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
@@ -183,6 +183,11 @@ function dosomething_signup_sms_game_form($form, &$form_state, $node, $num_frien
     ),
   );
   for ($i = 0; $i < $num_friends; $i++) {
+    $form['group1']['beta_first_name_' .$i] = array(
+      '#type' => 'hidden',
+      '#title' => 'Your Friend\'s Name',
+    );
+
     $form['group1']['beta_mobile_' .$i] = array(
       '#type' => 'textfield',
       '#required' => TRUE,
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -776,11 +776,16 @@ function dosomething_signup_sms_game_multi_player_signup_request($values) {
     'beta_mobile_0',
     'beta_mobile_1',
     'beta_mobile_2',
+    'beta_first_name_0',
+    'beta_first_name_1',
+    'beta_first_name_2',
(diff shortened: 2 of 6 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
```diff
@@ -189,13 +183,12 @@ function dosomething_signup_sms_game_form($form, &$form_state, $node, $num_frien
...
-    // Default to second fieldset.
-    $group = 'group2';
-    if ($i == 0) {
-      // Unless its the first beta.
-      $group = 'group1';
-    }
-    $form[$group]['beta_mobile_' .$i] = array(
+    $form['group1']['beta_first_name_' .$i] = array(
+      '#type' => 'hidden',
+      '#title' => 'Your Friend\'s Name',
```
aaronschachter, line 188: `t()` ?

aaronschachter:
code looks ok to me, just one minor going global comment

lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
DFurnes, line 188: This field isn't actually doing anything besides telling Drupal to whitelist sending `'beta_first_name` fields along to Mobile Commons. The form fields used for the A/B test are hard-coded in the `<div id="ab-test-sms-names" />` markup in the template.

DFurnes, line 188: Happy to update anyways, but it's more of a developer convenience ("what is this for?") than anything. I don't even think Drupal creates `<label>` elements for hidden form fields.

jonuy:
@aaronschachter @DFurnes do you guys know if in the end this POSTs to our Node app? I'm only remembering now that SMS games with `mobilecommons_opt_in_path` and `mobilecommons_friends_opt_in_path` filled in behave differently than games with `sms_game_mp_story_id` and `sms_game_mp_story_type`. These beta player names are intended for the latter - the MP games - and so would need to POST to the Node app.

aaronschachter:
`sms_game_mp_story_id` and `sms_game_mp_story_id` variables set  == posts to node app

`mobilecommons_opt_in_path` and `mobilecommons_friends_opt_in_path` set == posts to mobilecommons

DFurnes:
But in both cases, the payload sent to either the relevant Node app or Mobile Commons is passed through `dosomething_signup_get_mobilecommons_friends_vars`, right? I couldn't find an alternative for Node requests, but just wanna make sure.

jonuy:
@DFurnes It seems like this `dosomething_signup_sms_game_multi_player_signup_request` in the dosomething_signup module is actually where the request to the Node app gets made.

DFurnes:
Now targets the correct multiplayer game method. Also updated PR (above) with new example payload.

aaronschachter:
Looks ok here :+1:

jonuy:
Also think it looks ok :+1:

DFurnes:
:dancer:

PR #4248, opened 2015-03-19
Header gallery data

### LINKED ISSUE #4187: Retrieve reportback data for header image gallery.
Get 4 promoted reportbacks for the header gallery, with appropriate fallbacks.
- [x] get 4 promoted reportbacks
- [x] if not enough promoted, fallback to approved reportbacks
- [x] if new campaign or no reportbacks, don't show the gallery

weerd:
### Fixes #4187

Did some refactoring of the preprocess Reportbacks output code now that there are 2 spots on a page where they could show up. There is now a **Reportback Showcase** gallery, apart from the normal Reportback gallery in the _Prove It_ section.

This required DRYing up the code a bit to allow it to be more versatile and work for both galleries.

Fixed an issue where PHP was squawking because the `theme_image()` was missing the `attributes` parameter (apparently the function doesn't have it default to anything).

_Note:_
I also did a quick update to upgrade the image style requested for reportbacks. They were set to `300x300`, but I bumped them to `400x400`. They actually need to be `480x480` but will address that in another issue since it requires a new image style to be created.

@aaronschachter @sergii-tkachenko @DFurnes

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
@@ -397,91 +397,6 @@ function paraneue_dosomething_preprocess_node_campaign(&$vars) {
 
 
 /**
- * Get total number of Reportbacks by status (promoted, approved, etc).
- *
- * @param $nid  Campaign node id.
- * @param $status Reportback status (promoted, approved, etc).
- * @return int
- */
-function paraneue_dosomething_preprocess_reportbacks_total_by_status($nid, $status) {
-  $params = array();
-  $params['nid'] = $nid;
-  $params['status'] = $status;
-
-  return dosomething_reportback_get_reportback_files_query_count($params);
-}
-
-
-/**
- * Get a specified number of themed Promoted Reportbacks.
- *
- * @param $nid  Campaign node id.
- * @param $count Number of Promoted Reportbacks to retrieve.
- * @return array
- */
-function paraneue_dosomething_preprocess_themed_promoted_reportbacks($nid, $count, $random = FALSE) {
-  $promoted_items = array();
-
-  $params = array();
-  $params['nid'] = $nid;
-  $params['status'] = 'promoted';
-  $params['random'] = $random;
-
-  $promoted_results = dosomething_reportback_get_reportback_files_query_result($params, $count);
-
-  foreach ($promoted_results as $rb_file) {
-    $rb_file->image = dosomething_image_get_themed_image_url_by_fid($rb_file->fid, '400x400');
-    $promoted_items[] = paraneue_dosomething_get_gallery_item((array) $rb_file, 'photo');
-  }
-
-  return $promoted_items;
-}
-
-
-/**
- * Build Reportback collection based on Promoted & Approved Reportback counts.
- *
- * @param $reportbacks
- * @param $desired_count
- * @param bool $use_placeholders
- * @return bool
- */
-function paraneue_dosomething_preprocess_reportback_assimilation ($reportbacks, $use_placeholders = FALSE) {
-
-  $fill_count = $reportbacks['desired_count'] - $reportbacks['total_promoted'];
-
-  // Absolutely no Approved Reportbacks to fulfill the fill count needed.
-  if ($reportbacks['total_approved'] === 0) {
-    if ($use_placeholders) {
-      $placeholders = paraneue_dosomething_get_themed_placeholder_reportbacks($fill_count);
-      $reportbacks['items'] = array_merge($reportbacks['items'], $placeholders);
-      $reportbacks['prefetched'] = 0;
-    }
-    else {
-      return FALSE;
-    }
-  }
-
-  // Some but not enough Approved Reportbacks to fulfill the fill count needed.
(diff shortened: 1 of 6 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
```diff
@@ -382,87 +382,132 @@ function paraneue_dosomething_preprocess_node_campaign(&$vars) {
...
- * Adds $gallery_items array of themed Reportback Files for given node preprocess $vars.
+ * Get total number of Reportbacks by status (promoted, approved, etc).
+ *
+ * @param $nid  Campaign node id.
+ * @param $status Reportback status (promoted, approved, etc).
+ * @return int
  */
-function paraneue_dosomething_preprocess_reportback_files(&$vars) {
-  $reportbacks_gallery = &$vars['reportbacks_gallery'];
+function paraneue_dosomething_preprocess_reportbacks_total_by_status($nid, $status) {
```
aaronschachter: This is minor, but the function name throws me off because i'm expecting this to preprocess an input array of variables, like most preprocess functions do.

I can see us using this helper function all over the place, not just in the theme. What do you think about renaming moving it into the `dosomething_reportback.module` file?

lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
```diff
@@ -382,87 +382,132 @@ function paraneue_dosomething_preprocess_node_campaign(&$vars) {
...
-  $reportbacks_gallery['items'] = array();
 
+/**
+ * Get a specified number of themed Promoted Reportbacks.
+ *
+ * @param $nid  Campaign node id.
+ * @param $count Number of Promoted Reportbacks to retrieve.
+ * @return array
+ */
+function paraneue_dosomething_preprocess_themed_promoted_reportbacks($nid, $count, $random = FALSE) {
```
aaronschachter: Same here, usually a preprocess function takes variables by reference and messes with them.  I'd recommend just renaming this as `paraneue_dosomething_get_themed_promoted_reportbacks` ?

lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
weerd: Sounds good to me! Yeah, I was kinda debating whether I needed to leave that in the preprocess file or not :shrug:. Will update!

lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
```diff
@@ -382,87 +382,132 @@ function paraneue_dosomething_preprocess_node_campaign(&$vars) {
...
-
+/**
+ * Build Reportback collection based on Promoted & Approved Reportback counts.
+ *
+ * @param $reportbacks
+ * @param $desired_count
+ * @param bool $use_placeholders
+ * @return bool
+ */
+function paraneue_dosomething_preprocess_reportback_assimilation ($reportbacks, $use_placeholders = FALSE) {
```
aaronschachter: Same criticism of using `preprocess` in function name.  Also, `assimilation` is kind of.. well, wtf is this, as far as readability.  Gentle suggestion of renaming to something like `public_gallery` or something, but up to you. When you read the function comments, it makes sense, but if we use this in other places around the theme, I don't think the `assimilation` name will help indicate what's going on in the code.

lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
weerd: Should I move it to the **includes/helpers.inc** file then? Since not really a preprocess?

aaronschachter: Sounds like a good spot for it.

aaronschachter:
Code seems ok, just minor suggestions to help readability / maintenance.

DFurnes:
Looks pretty good to me. Agreed with @aaronschachter's comments. :+1:

lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
weerd: LOL. I was being creative with the function name :) Will tone down my creativity, haha.

weerd:
@aaronschachter would it be best to put the other two functions `paraneue_dosomething_get_themed_promoted_reportbacks` and the `soon-to-be-renamed-assimilation-function` in the **helpers.inc**? I figure that's probably the best place for those...

aaronschachter:
Sure, @weerd, makes sense to me. Could potentially add a new `gallery.inc` but might make it harder to find stuff (too many files?)

weerd:
@aaronschachter so thanks to your suggestion, it actually highlighted that in the **helpers.inc** there was already a _very_ similar function I created a while back for the **approved** reportbacks! So I consolidated the two functions (getting themed promoted and getting themed approved reportbacks) into a single function you pass the **status** type to :) DRYness.

Also moved the other function into the **dosomething_reportback.module** file.

Seems much cleaner now! Lemme know if it looks better.

CC: @DFurnes

aaronschachter:
very nice :+1:

PR #4251, opened 2015-03-19
Move "Why signed up" field into elements wrapper. Fixes #4229.

### LINKED ISSUE #4229: Signup Reason and Signup Reason Label alignment is off on signup data form
related to #4202 Signup Reason and Signup Reason Label is still a bit off

image

image

DFurnes:
# Changes
- Moves "Why signed up" field into elements wrapper with other form fields. Fixes #4229.

# Screenshot

ahhhhhhh that's soo much better


sbsmith86:
:+1:

Issue #4255, opened 2015-03-19
logged in users don't get signed up for campaign from permalink


mikefantini:
Steps to reproduce:
- go to permalink as an authenticated user who _isn't_ already signed up for the campaign
- click sign up

If the user _is already signed up_ for the campaign, the functionality works as expected.

angaither:
The contents of the entire permalink page are`cache_set` as html, this was causing the `dosomething_signup_form` and the `login`||`signup` form to not respond.
Maybe we can try adding that in with JS?
Or perhaps we need to set the cached data to a serialized array of all of the vars that are loaded and still call the theme function on the variables.

PR #4257, opened 2015-03-19
Remove permalink cache temporarily.

angaither:
This was causing the signup button to be cached, and not signing people up for campaigns.
just commenting out the cache lines, till I figure out how to not cache that part or something
Refs #4255


aaronschachter:
:+1:

Issue #4110, opened 2015-03-04
Refactor Campaigns API endpoint


aaronschachter:
Currently returns duplicate records, which is weird.

We'll want to query by various taxonomy fields on the template, and also by status, etc.  Curious about whether or not the Restful module could handle this nicely https://www.drupal.org/project/restful

aaronschachter:
... or if this is potentially a job for a SOLR API query, which @mshmsh5000 had mentioned potentially setting up for querying Reportbacks by as well.  Food for thought re: our API discussion today.

aaronschachter:
If we need to / can get RESTful up and running to handle Reportback / Reportback Item / File posting, and also relevant Reportback Items by Campaign, #4260 would provide a very easy solution to improving our `/campaigns` endpint.

PR #4260, opened 2015-03-19
Restful module

aaronschachter:
**UPDATE** 3/23/15 6:42 EST

This should be ok to merge and keep disabled on staging/production for now. Still need to sanity check that enabling does not mess with any current business critical API requests we make to ds.org

---

The endpoints for a User GET or PATCH don't exist yet. There are default endpoints that exist in Services, however it would require posting the data in the crazy `field_first_name[LANGUAGE_NONE][0]['value']` type format.  So, that's not fun.

We can create our own custom endpoints to massage the data like we want to, like i've been doing for `campaigns/:nid/reportback` or `campaigns/:nid/signup`, but that's not exactly fun either.

I wanted to look into the [RESTful module](https://www.drupal.org/project/restful) as the internet seems to say nice things about it. Once I figured out the weird cTools plugin stuff, it was actually pretty rad to see the `users` and `campaigns` endpoints provide nicely formatted objects, working with filtering out of the box.

screen shot 2015-03-19 at 1 08 05 pm

Note: The RESTful module defines the `api` endpoint, and all of its settings, on Drupal variables which we could export with Strongarm. The enabling/disabling of endpoints is done via these plugins, not as `services_endpoint` exportables like we use currently.

I'm going to keep commenting on this PR with both good things and bad things I see as I explore this option a little bit more.  So far, a bad thing is what a Field Collection looks like out of the box, but this is probably customizable.  Also not sure how to get the Files post working correctly just yet (right now we cheat in the `campaigns/:nid/reportback` endpoint by requiring a base64 filestring.


aaronschachter:
Note that you can actually create different objects based on the v1.0. So there can be a v1.1 with a different format for a property.

What seems kinda weird is having a half this / half that approach.. like if we kept the custom Services endpoints we have this far .... however we could create a new `v1.0` API endpoint and eventually the plain `v1` could be depcreated.

ideally if we go this route we need to port over the Signup and Reportback `targeted_actions`  as just POST signup and POST reportback. but what's a little tricky there is you also need to POST a file.

aaronschachter:
Able to send `PATCH` requests to update single fields.

screen shot 2015-03-19 at 2 03 28 pm

aaronschachter:
Also finding this "Discovery resource" pretty awesome. It generates API documentation as it finds each resource you declare.  About to mess with their built in File upload resource.

screen shot 2015-03-19 at 2 04 55 pm

DFurnes:
This looks amazing. Nice find @aaronschachter! :book: 

Looks like it follows the [JSON API](http://jsonapi.org) pseudo-standard out of the box which is pretty cool too.

aaronschachter:
Thanks @DFurnes ! yeah its pretty neat so far. Will keep posting progress, figuring out how to nest Resources in https://github.com/aaronschachter/dosomething/commit/c9d8472a80b291e08b559602e5fbf5e2e65628fd

screen shot 2015-03-19 at 3 12 12 pm

aaronschachter:
Ah ha. To get the nesting stuff to work for a custom entity, you need to define the relationships for the custom entity via `hook_entity_property_info`: https://github.com/aaronschachter/dosomething/commit/6ff95ecc1281201dca30fcacf90bbaaf374d92fc

**NOTE** If this is not set and you try to set the `resource` when declaring `public_fields`, you get a 502 Bad Getway WSOD :skull: 

screen shot 2015-03-19 at 5 12 47 pm

angaither:
this looks :ok_hand: much easier & closer to what we wanted

aaronschachter:
Yeah so far it's been so much better than wrestling with the `hook_services_resources` [stuff](https://github.com/DoSomething/dosomething/blob/v0.4.55/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc#L3), and the output looks 10000x than anything i've built custom! Definitely leaning towards this new module and endpoint (`api/v1.0`) for beginning new mobile app development (cc @jonuy )

Speaking of custom stuff, I'm going to try to add some advanced functionality today, mainly custom endpoints for things like `reportback-files/:id/review` or a Reportback endpoint which accepts a File and then creates a relevant Reportback Item.  Or customizing the nested entities so it doesn't bring back the entire Campaign object when referencing it from a `reportbacks/:id` endpoint.

DFurnes:
Any idea why this module encourages including minor version in the URL?

It seems strange, since the only thing that API clients would presumably care is major (breaking) version bumps, and it seems like it would be tedious and of relatively little benefit to maintain compatibility for minor API versions `v1.0`, `v1.1`, `v1.2`, etc.

aaronschachter:
Not sure, I haven't looked into that and just been focused on how to make it work for us.  I suppose we could define the plugins to just be `v1` and not `v1.0` but i haven't tried it yet. If the `.inc` file for each resource isn't named correctly, things breaks.

folderz

They use [Github](https://github.com/RESTful-Drupal/restful) instead of their [d.o project page](http://drupal.org/project/restful). to manage issues and all documentation, so I'm sure we could ask :)  It's definitely a new module and is being actively developed (commits every day)

aaronschachter:
@DFurnes Looks like you don't have to specify a minor verison # in the URL, this will work. `http://dev.dosomething.org:8888/api/v1/signups` or even this `http://dev.dosomething.org:8888/api/signups` depending on the module settings:

RESTful config

Having the different versions helps support different devices, but seems to just be an option. Per their documentation "The versions are at the resource level, for more flexibility and control." So you can have signups 1.4 but reportbacks 1.2

aaronschachter:
Continuing the journey... one inconvenient bug is if when you name your `inc` file incorrectly. Upon cache clear you get lots of errors. So far i've been them by disabling and re-enabling the `restful` module. This only when for developing locally, but just a "gotcha" to document if we bring this into our repo.

```
require_once(/var/www/dev.dosomething.org/html/profiles/dosomething/modules/dosomething/dosomething_api/plugins/restful/node_user/1.0/node_user__1_0.inc):[warning]
failed to open stream: No such file or directory plugins.inc:475
PHP Fatal error:  require_once(): Failed opening required '/var/www/dev.dosomething.org/html/profiles/dosomething/modules/dosomething/dosomething_api/plugins/restful/node_user/1.0/node_user__1_0.inc' (include_path='.:/usr/share/php:/usr/share/pear') in /var/www/dev.dosomething.org/html/profiles/dosomething/modules/contrib/ctools/includes/plugins.inc on line 475
Drush command terminated abnormally due to an unrecoverable error.   [error]
Error: require_once(): Failed opening required
'/var/www/dev.dosomething.org/html/profiles/dosomething/modules/dosomething/dosomething_api/plugins/restful/node_user/1.0/node_user__1_0.inc'
(include_path='.:/usr/share/php:/usr/share/pear') in
/var/www/dev.dosomething.org/html/profiles/dosomething/modules/contrib/ctools/includes/plugins.inc,
```

aaronschachter:
[This one was a major GOTCHA](https://github.com/aaronschachter/dosomething/commit/c7da27c48f7edde6f93754cbbc08a4f15ffb0c9f). Out of the box, I can't figure out a way to **GET** `/reportback_items` filtering by a Campaign NID (its a nested join on the `dosomething_reportback` table to the `node` table).
The main GOTCHA was that i was attempting to use functionality that was not included in the latest release (1.0-rc4) that I saw on Github. So this pins the RESTful module to the latest commit, which supposedly has Field Collection support but still doesn't look like it when viewing the Campaign response.

[This commit](%28https://github.com/aaronschachter/dosomething/commit/c7da27c48f7edde6f93754cbbc08a4f15ffb0c9f%29) allows querying Reportback Items by nid and status. The multi-value id=23,45 doesn't seem to work here, but possibly cos we're extending `RestfulDataProviderDbQuery` and not the usual `DoSomethingReportbackItemsResource`. Joining on the resource to get the nested entity doesn't seem to work in this ProviderDbQuery either, as far as I can tell.

e.g. 
`http://dev.dosomething.org:8888/api/v1.0/gallery?filter[campaign]=362&filter[status]=promoted`

 It's essentially doing some of the heavy lifting we're doing in [here](https://github.com/DoSomething/dosomething/blob/v0.4.56/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module#L953).  If we keep this `gallery` resource, we'll need to join on the Primary Cause field to query by Cause term id's.  My hope is htat I can figure out a way for the **GET** `reportback-items` to use this function instead of the default GET (which doesn't allow filtering by Campaign nid or cause, at least as far as I can see)

aaronschachter:
Trying to link the ReportbackItems `fid` to the `files` resource is resulting in a 502.

When I change the `fid` type to `file` in the https://github.com/DoSomething/dosomething/blob/v0.4.58/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module#L134 it fixes the 502.. but the returned `file` is null.  

This type of stuff doesn't seem to happen much on any of the node joins, but more in our custom entities is where I've been getting burned. (https://github.com/DoSomething/dosomething/pull/4260#issuecomment-83818519) When we first created the Reportback File table, we had not turned on File Entity yet (hence the integer field)

aaronschachter:
Here's a sample of a Campaign using the newly added Assets resource (which exports Image node).
screen shot 2015-03-23 at 1 00 11 pm

aaronschachter:
[This](https://github.com/aaronschachter/dosomething/commit/fc86c6a6e92da5b0b5f8708263c3ae29d0a38a92) is similar spot to https://github.com/DoSomething/dosomething/pull/4260#issuecomment-85172604 that I've been stuck on. The multi-value taxonomy fields are returned beautifully, but for some reason the single-value fields (Primary Cause, Primary Action Type) are returning NULL.

y u no work single-value

The good news, is that adding these taxonomy fields as public fields gets pretty sweet Campaign Finder type functionality out of the box on the **GET** `campaigns` resource:

Get all campaigns with Action Type == Donate Something:
`http://dev.dosomething.org:8888/api/v1/campaigns?filter[action_type]=7`

Get all campaigns tagged as "dance" in the Tags field:
`http://dev.dosomething.org:8888/api/v1/campaigns?filter[tags]=147`

Get all campaigns tagged as "Environment" in the Cause multi-select field:
`http://dev.dosomething.org:8888/api/v1/campaigns?filter[cause]=4`

I think it's safe to say that using RESTful will definitely close out #4110.

Also worth noting that RESTful provides a `count` and `next` properties for these paginated requests (cc @weerd )
paginate

aaronschachter:
This PR is in decent shape for others to pull down and test to mess around with the api endpoint it creates.  it also may be useful to browse the [module documentation](https://github.com/restful-drupal/restful) as well 

I've prevented the `ds build` script from enabling automatically, so in order to test you must follow these steps:
- pull down this branch
- run a `ds build` ( to add the RESTful module into the relevant `html` dir)
- `drush -y en restful`
- Clear cache and start messing around at `http://dev.dosomething.org:8888/api/` 

I'd love for others to take this for a spin before we merge it. I believe it should be safe to merge.. **not** necessarily safe to turn on in staging. I haven't fully tested if there are any conflicts with our current custom Services endpoints when enabling this module and its plugins.

I'm still trying/hoping to figure out the funkiness in issues I've found, but I also want to be sure tomorrow to spend time documenting of the current approach I started when building out our endpoints (customized Services endpoints), and some pros and cons for if we continue that route, or go this route (with RESTful). 

cc @angaither @sergii-tkachenko @weerd @DFurnes

aaronschachter:
I attempted defining all of the plugins as `2.0` to hopefully just completely start on a separate endpoint but it broke stuff. The module may not starting at 2.0 without a 1.0, but I didn't go into it too deep because  I spent a lot of the day trying to figure out the nested resources stuff.

The nested resources aren't necessarily a BLOCKER for Mobile App development, but... it would be bad news to have rely on always making multiple requests when one should do it.  I'm hoping it's just something really dumb I'm missing... or, something we can open an issue about in the RESTful repo...

aaronschachter:
Other filtering examples using public fields:

Reportback Items with Caption:
`http://dev.dosomething.org:8888/api/v1/reportback_items?filter[caption]=Test`

Reportbacks with Why Participated:
`http://dev.dosomething.org:8888/api/v1/reportbacks?filter[why_participated]=Test`

Reportbacks with Quantity >= 50
`http://dev.dosomething.org:8888/api/v1/reportbacks?filter[quantity][value]=50&filter[quantity][operator]=%22%3E=%22`

Reportbacks with Quantity >= 50 for Campaign 362
`http://dev.dosomething.org:8888/api/v1/reportbacks?filter[quantity][value]=50&filter[quantity][operator]=%22%3E=%22&filter[campaign]=362`

[Documentation](https://github.com/RESTful-Drupal/restful/blob/64c4ba0aee1767a5e82cf94f7ed0d453a69a3286/docs/api_url.md#applying-a-query-filter)

drewish:
I spent some time looking at the todos on the campaign's single value term fields. I didn't get to the bottom of the problem but it was a good excuse to dig through your code and the restful module's. 

One thing I thought was interesting was changing `field_primary_cause` to allow 2 values it went from:

```
primary_cause: null
```

to:

```
primary_cause: [
  null
],
```

Bumping it to unlimited gave the same result. So it seem like there might be something besides it just being a single value term field.

lib/modules/dosomething/dosomething_api/plugins/restful/campaigns/1.0/DoSomethingCampaignsResource.class.php
```diff
@@ -0,0 +1,96 @@
...
+       'resource' => array(
+          'image' => 'assets',
+        ),
+    );
+
+    //@todo This isn't working! Returns NULL.
+    $public_fields['primary_cause'] = array(
+      'property' => 'field_primary_cause',
+       'resource' => array(
+          'primary_cause' => 'terms',
```
drewish, line 40: Should be: `'cause' => 'terms',`

lib/modules/dosomething/dosomething_api/plugins/restful/campaigns/1.0/DoSomethingCampaignsResource.class.php
```diff
@@ -0,0 +1,96 @@
...
+          'cause' => 'terms',
+        ),
+    );
+
+    // @todo This doesn't work either.
+    // It's not liking something about the single value taxonomy :(
+    $public_fields['primary_action_type'] = array(
+      'property' => 'field_primary_action_type',
+       'resource' => array(
+          'primary_action_type' => 'terms',
```
drewish, line 56: Should be: `'action_type' => 'terms',`

drewish:
I'm tired enough that I don't have a good explanation for the why behind those changes. I just did some debugging and followed the flow into down to [this line](https://github.com/RESTful-Drupal/restful/blob/7.x-1.x/plugins/restful/RestfulEntityBase.php#L515) where the test was failing. Comparing the values on the two fields showed one had `primary_action_type` and the other `action_type` but the key in the `$resource` variable didn't change between them. Making them match got the fields displaying correctly.

drewish:
Oh duh... `action_type` is the vocabulary name which is also a bundle. Took me that long to remember D7 added string names for the vocabs instead of just having a vid like in D6.

drewish:
I'd been putting my fixes for this over on https://github.com/DoSomething/dosomething/pull/4360

drewish:
I'm going to close this since it should have all been picked up by my other PR.

PR #4266, opened 2015-03-20
The first slash was blocking the url alias

### LINKED ISSUE #4254: URL alias not maintained when registering from permalink
maintain url alias when registering for DS from a permalink

to reproduce:
- go to permalink page as an anonymous user
- click sign up
- register as a new user

image

angaither:
Fixes #4254


aaronschachter:
:+1:

PR #4271, opened 2015-03-20
Stathat

angaither:
Adds stathat module
send donation stats to stathat

### DIFF
--- a/lib/modules/dosomething/dosomething_payment/dosomething_payment.module
+++ b/lib/modules/dosomething/dosomething_payment/dosomething_payment.module
@@ -81,7 +81,7 @@ function dosomething_payment_block_view($delta = '') {
  * Form constructor for creating a Payment.
  */
 function dosomething_payment_form($form, &$form_state, $config = array()) {
-  // This JS is needed to create a Stripe token, which is used for API posts. 
+  // This JS is needed to create a Stripe token, which is used for API posts.
   drupal_add_js('https://js.stripe.com/v2/');
   // Add the API publish key to JS settings.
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_payment/dosomething_payment.module
```diff
@@ -210,6 +210,10 @@ function dosomething_payment_form_submit($form, &$form_state) {
   $values['amount'] = $values['amount'] * 100;
   // Send Payment request to Stripe API.
   if (dosomething_payment_post_payment_stripe($values)) {
+    // Send stathat info.
+    if (module_exists('stathat')) {
+      stathat_send_ez_count('drupal - Donate', 1);
```
aaronschachter, line 215: Minor, but I'm not sure if the module handles when the ez key hasn't been entered. Is the plan to manually enter the key on stage and prod?

aaronschachter:
Code looks good, just need to check for minor edge case where form may be submitted before key is entered (or if erased)..

lib/modules/dosomething/dosomething_payment/dosomething_payment.module
angaither, line 215: yarrr, it just wouldn't send which isn't worse than what's happening now, right?

angaither:
wouldn't really want to hardcode the key into code...

aaronschachter:
Yah, think its fine for now since it's just Donate Page, i'm just not sure if the module throws an error/watchdog if the key doesnt exist, if it gets wiped out by chance.

PR #4272, opened 2015-03-20
Header gallery styles

### LINKED ISSUE #4186: RB header gallery default to off
we're going to be testing the efficacy of this via optimizely, so we'll want to default this to be turned off across the site and then via optimizely we will toggle the gallery on.

we'll then be able to see how much of an impact it has on conversion rate

### LINKED ISSUE #4188: Style header gallery images
Add rotation transform for each image, either by adding a class or using an `nth` child approach.

weerd:
### Fixes #4186

This should wrap up the styles for the new **showcase gallery** in the header. Adds some modified patterns which will need to be moved into Neue. Also pulled out the `.photo` pattern and placed into `/patterns` directory for easier reference when moving to Neue.

The rotation with `nth-child` aims to rotate `odd` and `even` gallery items plus a could extra `nth-child` definitions to add some variety in between :dancer: 

The `.-hidden` class is only temporary and will be used for easy Optimizely tests to toggle between showing or not showing the showcase gallery.

@DoSomething/front-end

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
@@ -346,7 +346,7 @@ function paraneue_dosomething_build_reportbacks_gallery_collection ($reportbacks
       $reportbacks['prefetched'] = 0;
     }
     else {
-      return FALSE;
+      return NULL;
     }
   }
 
--- a/lib/themes/dosomething/paraneue_dosomething/scss/patterns/_container.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/patterns/_container.scss
@@ -1,0 +1,9 @@
+.container {
+  &.-showcase {
+    background-color: #000;
+
+    > .wrapper {
+      padding: $base-spacing 0;
+    }
+  }
+}
--- a/lib/themes/dosomething/paraneue_dosomething/scss/patterns/_container.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/patterns/_container.scss
@@ -1,5 +1,5 @@
 .container {
-  &.-hidden {
+  &.optimizely-hidden {
     display: none;
   }
 
--- a/lib/themes/dosomething/paraneue_dosomething/scss/patterns/_container.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/patterns/_container.scss
@@ -1,4 +1,8 @@
 .container {
+  &.-hidden {
+    display: none;
+  }
+
   &.-showcase {
     background-color: #000;
 
--- a/lib/themes/dosomething/paraneue_dosomething/scss/patterns/_gallery.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/patterns/_gallery.scss
@@ -1,0 +1,28 @@
+.gallery {
+
+  &.-shuffle {
+    li {
+      margin: 0;
+      position: relative;
+
+      &:nth-child(odd) {
+        top: 7px;
+        transform: rotate(-1deg);
+      }
+
+      &:nth-child(even) {
+        top: -5px;
+        transform: rotate(1deg);
+      }
+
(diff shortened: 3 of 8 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -346,7 +346,7 @@ function paraneue_dosomething_build_reportbacks_gallery_collection ($reportbacks
       $reportbacks['prefetched'] = 0;
     }
     else {
-      return FALSE;
+      return NULL;
```
weerd, line 349: @aaronschachter I made a quick change here cause returning `FALSE` apparently is no good if using `isset()` in template cause it sees it as the variable having a value (which makes sense). So I changed it to return `NULL` so the conditional in template works correctly.

aaronschachter, line 349: ok

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_container.scss
```diff
@@ -0,0 +1,13 @@
+.container {
+  &.-hidden {
```
DFurnes: Can we consider renaming this `.optimizely-hidden` or something similar to denote that it's for a temporary A/B test?

weerd: Sure. Makes sense :) Updating.

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_gallery.scss
```diff
@@ -0,0 +1,28 @@
+.gallery {
+
+  &.-shuffle {
```
DFurnes, line 3: This is fabulous. :100:

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_photo.scss
```diff
@@ -0,0 +1,29 @@
...
+
+    @include media($larger) {
+      padding: ($base-spacing / 2);
+    }
+
+    img {
+      width: 100%;
+    }
+
+    .__copy {
```
DFurnes, line 15: Doesn't have to be in this PR, but can you update this to `.photo__copy` sometime?

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign--pitch.tpl.php
```diff
@@ -7,6 +7,9 @@
  * - $classes: Additional classes passed for output (string).
  * - $campaign_scholarship: Scholarship amount (string).
  */
+
+//dpm($variables);
```
DFurnes: Remove this bugger?

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_photo.scss
weerd, line 15: Ah jes. I can do it in the near future when I move the pattern over to Neue :)

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign--pitch.tpl.php
```diff
@@ -7,6 +7,9 @@
...
+//dpm($reportbacks_showcase);
```
DFurnes: And this bugger? :bug:

DFurnes:
Really nice! :ship: Few teeeny tiny comments, but otherwise all good!

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign--pitch.tpl.php
weerd: Argh!

PR #4273, opened 2015-03-20
Updates to incorportate mobile styles.

weerd:
Needed to add mobile styles for the **showcase gallery**!

These gallery styles for a `.-quartet` will need to be reconciled somehow with the Neue styles. For this particular version, on small/medium devices we want it to go 2 items per row and 4 per row at desktop. Not sure what the ideal alternative would be.

@DoSomething/front-end


DFurnes:
This might be a new gallery style? Something like `.quartet-extra`?

(You're better at naming than me. :wink:)

DFurnes:
:+1: to merge, but definitely something to patternize!

weerd:
Cool. Will do when moved to Neue!

PR #4274, opened 2015-03-20
Add missing require for jQuery.

DFurnes:
# Changes
- Add missing require for jQuery, fixes `$ is undefined` error.


weerd:
:sheep: it!!!

PR #4275, opened 2015-03-23
Update to enabled admins to click on button on Reportbacks to edit their status.

### LINKED ISSUE #4219: Admins should see a link to admin permalink page from gallery
admins viewing images in the gallery should be able to clickthrough to reportback permalink to exclude/flag the image

```
  link image to `/admin/reportback/<rbid>`
}
```
- [x] add link for gallery items output via PHP
- [x] add link for gallery items inserted via AJAX when viewing more.

weerd:
### Fixes #4219

This adds a nice little hover effect on each Reportback (both the Prove It gallery and the showcase gallery up at top of page) that shows an edit button admins can click on to head to the administration page in the CMS for that reportback.

image

_Note_: This addresses Reportbacks output via PHP, however it does not address Reportbacks being viewed via the View More since they are added via JavaScript and need a little somethin', somethin' extra to get that working.

@angaither 
@DoSomething/front-end 

CC: @mikefantini

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
@@ -286,8 +286,13 @@ function paraneue_dosomething_get_themed_reportbacks_by_status($campaign_id, $st
 
   $reportback_results = dosomething_reportback_get_reportback_files_query_result($params, $count);
 
+  $admin_link = user_access('view any reportback');
+
   foreach ($reportback_results as $reportback) {
     // @Todo: DRY. Potentially grab this data from the API endpoint to truly DRY?
+    if ($admin_link) {
+      $reportback->admin_link = '/admin/reportback/' . $reportback->rbid;
+    }
     $reportback->image = dosomething_image_get_themed_image_url_by_fid($reportback->fid, '400x400');
     $reportback_items[] = paraneue_dosomething_get_gallery_item((array) $reportback, 'photo');
   }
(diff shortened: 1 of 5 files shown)


lib/themes/dosomething/paraneue_dosomething/templates/system/patterns/photo.tpl.php
```diff
@@ -1,4 +1,9 @@
 <figure class="photo -stacked -framed">
+  <?php if (user_access('view any reportback')): ?>
+    <div class="admin-edit">
+      <a class="button -secondary" href="/admin/reportback/<?php print $content['rbid'] ?>">Edit Status</a>
```
angaither: can we pass this in from the module? both the user_access and the link?
then in the template we would just say 

```
<?php if (!is_null($admin_link): ?>
 print $admin_link;
<?php endif; ?>
```

weerd: Sure :)

DFurnes:
:+1: :art:

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -286,8 +286,13 @@ function paraneue_dosomething_get_themed_reportbacks_by_status($campaign_id, $st
 
   $reportback_results = dosomething_reportback_get_reportback_files_query_result($params, $count);
 
+  $admin_link = user_access('view any reportback');
```
angaither: if `admin_access`, but leave the link `admin_link`

angaither:
:+1:

Issue #4276, opened 2015-03-23
User login events > northstar


angaither:
Send user login events into Northstar, either via api.ds.org or directly into northstar

angaither:
refs #4411

PR #4285, opened 2015-03-24
Admin northstar inital setup

angaither:
Adds new `dosomething_northstar` module
Currently only has an admin form to add settings. 
Add guzzle as a new library in `drupal-org.make`

Initial ground work for #4276

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.admin.inc
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.admin.inc
@@ -1,0 +1,53 @@
+<?php
+
+/**
+ * @file
+ * Zendesk admin settings.
+ */
+
+/**
+ * System settings form for northstar config.
+ */
+function dosomething_northstar_config_form($form, &$form_state) {
+   $form = array();
+
+  $form['northstar'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Northstar Settings'),
+  );
+  $form['northstar']['dosomething_northstar_url'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Northstar URL'),
+    '#required' => TRUE,
+    '#default_value' => variable_get('dosomething_northstar_url', 'http://northstar.dev'),
+  );
+  $form['northstar']['dosomething_northstar_port'] = array(
+    '#type' => 'textfield',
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_northstar/dosomething_northstar.admin.inc
```diff
@@ -0,0 +1,53 @@
+<?php
+
+/**
+ * @file
+ * Zendesk admin settings.
```
aaronschachter: Not needed for merge, but let's correct this at some point.

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.info
```diff
@@ -0,0 +1,6 @@
+name = DoSomething Northstar
+description = Provides connection to our Norhtstar user database.
```
aaronschachter: OCD spelling

lib/profiles/dosomething/drupal-org.make
```diff
@@ -253,6 +253,10 @@ projects[stage_file_proxy][subdir] = "contrib"
 ;libraries[dsapi][download][type] = "git"
 ;libraries[dsapi][download][url] = "git@github.com:DoSomething/dsapi-php.git"
 
+; Guzzle
+libraries[guzzle][download][type] = "git"
+libraries[guzzle][download][url] = "https://github.com/guzzle/guzzle.git"
```
aaronschachter, line 258: Worth setting to a release number? Per recent burns with Stripe API...

angaither, line 258: conflicted on that, I don't want to get super behind on the version, as they release new stuff frequently. Maybe these transactions is something we could add monitoring into?

aaronschachter, line 258: Ok, nothing we need for now to get setup, but def something to keep in mind.

aaronschachter:
Shippy :+1:

lib/profiles/dosomething/drupal-org.make
DFurnes, line 258: +1 to locking to a specific tag. Seems inevitable that we'll get burned when they release a major version otherwise.

PR #4287, opened 2015-03-24
Reportback &-persands

### LINKED ISSUE #4286: & isn't showing properly in captions
https://trello.com/c/s2PgyHlH

earliest we're able to confirm as working was 3/13. seems to have broken shortly after that

Hilary estimates about 15% of reportbacks are using this for Comeback Clothes and it would look good for sponsor to be able to promote these.

angaither:
@mikefantini this fixes the `caption` & `why_participated` in the owner/non-owner permalink as well as the admin view of the permalink

Fixes #4286


sergiitk:
:+1:

mikefantini:
yeah let's check the gallery as well (both showcase gallery and step 4 gallery, although I assume they will output the same caption display)

angaither:
@mikefantini @sergii-tkachenko  ok, added gallery stuffs here as well

PR #4295, opened 2015-03-26
Fixes #4291.

### LINKED ISSUE #4291: Invalid argument supplied for foreach() node--campaign.tpl.php:212
When indexing all content using solr, some campaigns produce the following error:

```
Invalid argument supplied for foreach() node--campaign.tpl.php:212
```

Looks like sometimes `$do` variable isn't set in `node--campaign.tpl.php:212`

sergiitk:
Some campagins don't have `$do` variable set, which produces error described in #4291:
`Invalid argument supplied for foreach() node--campaign.tpl.php:212`

Steps to reproduce error:
- `drush solr-mark-all`
- `drush solr-index`

This PR changes the template so it checks if `$do` variable present first.


DFurnes:
:shipit:

PR #4296, opened 2015-03-26
Reindexes Solr automaticall after ds pull stage.

### LINKED ISSUE #4290: Reindex Solr after ds pull stage
Extend ds script: reindex solr after pulling the database from staging.

sergiitk:
Automatically reindex Solr after `ds pull stage`.
Takes about 20 seconds, which is nothing compared to pulling and importing the database itself.
Closes #4290.


angaither:
:+1:

Issue #4282, opened 2015-03-24
Add helper variable progress to campaign


angaither:
this field displays the total progress of the number of x things y'ed

if left blank, use the total number of approved reportback quanity goes there
otherwise, use the number in the new total progress field

once a new reportback is approved, add to the progress value set in the campaign node (if it's overridden we are adding to that overridden value)

aaronschachter:
@angaither FYI that we're using `dosomething_helper_variables` to store the current Reportback count by status of each Campaign and Cause's. This issue is also really similar to #4051 which is an open issue to store current count of Signups, which suggests also using the helper variables or @mshmsh5000 mentions storing it in Redis in that thread.

Fields could work here, it would just be a different implementation from the current Reportback Status counts. I used the helper variables specifically for the Reportback stuff to avoid `node_save` upon each Reportback File create/update, and to avoid adding 4 fields to the Campaign node / Cause vocab.

aaronschachter:
This is also perhaps a duplicate of #3899 ?

angaither:
Ah, yes. it does look like a dup of #3899, however this need to be able to potentially be override, like the totals in a campaign run.

aaronschachter:
My $0.02 is that we should keep the build simpler by simply not allowing overrides.

If that's not an option, we could always build an edit form for the Helper variables, to avoid storing all this stuff in fields.  Keeping in fields has its pros and cons but I think the biggest con would be a frequent `node_save`.

angaither:
The case for overrides comes from campaigns with external partners, where we hear count numbers higher than than what we have in the rb_quanitity, like H&M reports we got 30K more pounds of clothes, but only 10k was reported back to us.

aaronschachter:
Ah, I see. Thanks for clarifying / documenting!  Those are the sort of scenarios where it can be easy to forget why we built something a certain way when we look back at it in the future :watch:

angaither:
Closed with #4320

Issue #3899, opened 2015-02-06
Sum variable for Campaign Quantity


aaronschachter:
Store a new calculated helper variable `sum_quantity` for each campaign node, which calculates the sum of all Reportbacks where flagged = 0 and also have no pending Reportback Files.  Call this function within `dosomething_reportback_reset_all` (Refs #3874)

Refs: https://trello.com/c/fRDF3nFy cc @mikefantini

mikefantini:
@aaronschachter what does the "and also have no pending Reportback Files" mean?

Does it only calculate the quantity sum if there are no reportbacks in the inbox?

aaronschachter:
Reportbacks which have any RB Files set to `pending` status should not have their quantity counted in the sum.  Once the pending file status is set to either Approved/Promoted/Excluded, its reportback quantity will be included in the sum (upon next calculation).

mikefantini:
ok great. wanted to make sure that's what it meant.

angaither:
closed with #4317

PR #4302, opened 2015-03-27
Set progress as default on campaign variables page

angaither:
This code is not ready for prod, but is a TEST for @mikefantini to see how updating the `rb_quanity` would work on each reportback approval.

I _intentionally_ left in messages/debug statements to help visualize what is going on when this code runs.
I would like to deploy this to staging to play around with this. 

Refs #3899 Refs #4282


sergiitk:
:+1:

PR #4306, opened 2015-03-27
Add stathat.php to drupal-org.make to allow stats to be logged.

### LINKED ISSUE #4298: Stathat from donate page not reporting
It's unclear if no one is donating, or if this is just not making it to stathat.

angaither:
I added the stathat module before, but not the php library, so stats weren't getting logged.
silly...
Fixes #4298

### DIFF
--- a/lib/profiles/dosomething/drupal-org.make
+++ b/lib/profiles/dosomething/drupal-org.make
@@ -275,6 +275,10 @@ libraries[predis][download][url] = "https://github.com/nrk/predis.git"
 ; Pin to 0.8 release, which still has the lib directory.
 libraries[predis][download][revision] = "4123fcd85d61354c6c9900db76c9597dbd129bf6"
 
+; Stathat
+libraries[stathat][download][type] = "get"
+libraries[stathat][download][url] = "https://www.stathat.com/downloads/stathat.php"
+
 ; Stripe PHP
 libraries[stripe-php][download][type] = "git"
 libraries[stripe-php][download][url] = "https://github.com/stripe/stripe-php"


lib/profiles/dosomething/drupal-org.make
```diff
@@ -275,6 +275,10 @@ libraries[predis][download][url] = "https://github.com/nrk/predis.git"
 ; Pin to 0.8 release, which still has the lib directory.
 libraries[predis][download][revision] = "4123fcd85d61354c6c9900db76c9597dbd129bf6"
 
+; Stathat
+libraries[stathat][download][type] = "get"
```
sergiitk, line 279: Is `get` a valid type? Did you mean `file`?
http://cgit.drupalcode.org/drush_make/plain/README.txt?id=refs/heads/6.x-2.x

angaither, line 279: yes, it worked
https://www.drupal.org/node/1432374

```
; To include a code library external to the Drupal project, such as
; jQuery UI or TinyMCE, you can include them using the 'libraries'
; attribute, specifying at least the download 'type' and 'url'.
; Note that any referenced third-party libraries *must* be listed at
; http://drupal.org/packaging-whitelist
libraries[jquery.cycle][download][type] = get
libraries[jquery.cycle][download][url] = http://malsup.com/jquery/cycle/release/jquery.cycle.zip?v2.99
libraries[jquery.cycle][destination] = libraries
```

sergiitk, line 279: Ok then. This is strange though, they didn't include this type to `drush topic docs-make`.

sergiitk:
:+1:

PR #4309, opened 2015-03-30
Fixes #4305 by using proper tag for paragraph spacing rule.

### LINKED ISSUE #4305: not enough padding btwn know it image and problem
image

happening on http://staging.beta.dosomething.org/node/1199; seems fine with video (see https://www.dosomething.org/node/362)

cc: @lkpttn

DFurnes:
# Changes
- Fixes #4305 by using proper tag for `p + p` paragraph spacing rule. I think this makes semantic sense to use a `p` tag here as well, but it might be worth pondering if there's a better long-term solution for cases like this.


sbsmith86:
I think `p` works here.

DFurnes:
:dancer:

PR #4310, opened 2015-03-30
Add a dollars amount & track that in stathat.

### LINKED ISSUE #4208: StatHat coverage for donations
1. Simple counter for each successful donation submission
2. Incrementing value for total donations collected

The first will be a health check for this integration in production. The second will be gravy, but worth tracking.

angaither:
Fixes #4208


sergiitk:
:+1:

PR #4317, opened 2015-03-30
Cron all the things

angaither:
#### What this pr does

This cron will pull reportbacks updated in the past hour and update the quantity count for those nodes.
#### Known issue

I still think we will see higher numbers than the reality from over adding (numbers may look higher than they actually are)
(eg) If a user updates a reportback that quantity number will get added to the total as many times as the user updates the reportback. 
#### N.B.

I looked into pulling the quantity from the log, but that still isn't a difference of the past quantity and the current quantity. 
Until we have endpoints from the dashboard to set these numbers correctly every so often, I think this is our best option. 

Refs #4282
Refs #3899

@mshmsh5000 @mikefantini @joshcusano

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.cron.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.cron.inc
@@ -1,0 +1,25 @@
+<?php
+
+/**
+ * @file
+ * Provides cron job/drush command for dosomething_reportback.module.
+ */
+
+/**
+ * Implements hook_cron().
+ */
+function dosomething_reportback_cron() {
+  // Get all reportbacks updated in the past hour.
+  $results = db_query("SELECT nid, sum(quantity) as quantity
+                     FROM dosomething_reportback
+                     WHERE from_unixtime(updated) >= date_add(now(), INTERVAL '-1' hour)
+                     GROUP BY nid;");
+
+  // Updated reportback counts accordingly.
+  foreach ($results as $result) {
+    $previous_progress = dosomething_helpers_get_variable('node', $result->nid, 'sum_rb_quanity');
+    $updated_progress = (int) $previous_progress + (int) $result->quantity;
+    dosomething_helpers_set_variable('node', $result->nid, 'sum_rb_quanity', $updated_progress);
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.cron.inc
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+/**
+ * Implements hook_cron().
+ */
+function dosomething_reportback_cron() {
+  // Get all reportbacks updated in the past hour.
+  $results = db_query("SELECT nid, sum(quantity) as quantity
+                     FROM dosomething_reportback
+                     WHERE from_unixtime(updated) >= date_add(now(), INTERVAL '-1' hour)
```
sergiitk, line 15: Maybe it'd be better to calculate the date string in PHP?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.cron.inc
```diff
@@ -0,0 +1,25 @@
...
+                     GROUP BY nid;");
+
+  // Updated reportback counts accordingly.
+  foreach ($results as $result) {
+    $previous_progress = dosomething_helpers_get_variable('node', $result->nid, 'sum_rb_quanity');
+    $updated_progress = (int) $previous_progress + (int) $result->quantity;
```
sergiitk, line 21: Good idea!

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.cron.inc
angaither, line 15: curious why you think that?
I would think that it would be better to calculate the date in SQL, so that you know the `now()` is from that db server and not the webserver as those things might not be in sync.

sergiitk, line 15: I thought it'd be better for performance: DB servers usually are more busy than webheads running cron jobs. But yes, your have legit point about time being in sync.

sergiitk:
:+1:

PR #4322, opened 2015-03-31
This was saving `progress` and `sum_rb_quantity`

### LINKED ISSUE #4321: double saving progress values
when you save campaign helpers page it's saving fields as their given name, as well as the variable set manually.

angaither:
only one is needed.
fixes #4321


sergiitk:
:+1:

PR #4323, opened 2015-04-01
Neue 6.4

DFurnes:
# Changes
- Include Neue 6.4, which is now [published on NPM](https://www.npmjs.com/package/dosomething-neue). This allows us to simplify our front-end packaging infrastructure quite a bit, which will be really nice as we break things into more modular pieces.
- Use [Webpack](http://webpack.github.io) for builds, rather than Require.js/Almond. Webpack allows us to use NPM package information to intelligently bundle dependencies, rather than writing shims for everything. It also handles multiple entry points/bundles much more gracefully.
- Update jQuery to 1.11.0 :tada: and use updated version of Admin Menu module for jQuery 1.9+ support. We're no longer on a 2 year old release, great!


sbsmith86:
yay :fireworks: for debug mode for grunt :+1:

weerd:
Fantabulous! :sheep: it! 

Goodbye Bower, you were great while you lasted :dancer:

Issue #4308, opened 2015-03-30
Persistent sign up button on pitch page.


sbsmith86:
Create a persistent sign up button on the pitch page.

sbsmith86:
@mikefantini There is now a feature flag on staging at `Appearance-> Settings -> Paraneue DoSomething -> Persistent Signup Nav` to toggle this feature

sbsmith86:
Closing. We are moving forward with this in #4364

PR #4327, opened 2015-04-01
Persistent Sign up button on the pitch pages

sbsmith86:
## Addresses #4308

I added a feature flag that can be used to turn on the "persistent sign up nav" on the pitch page. It introduces a new look for the cta pattern and I am using a `-persistent` class as a modifier. Also added a `.-persistent-beta` class to the `campaign--pitch` container so that I can use it to hide/show various elements on the pitch page when the persistent nav is turned on.  

The idea is to get this on staging so we can test what is feels like on actual devices and then, if we decide to move forward, run an optimizely test to gauge performance. 

@DoSomething/front-end

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
@@ -131,8 +131,9 @@
 
 }
 
-.campaign--pitch.-persistent-beta {
-  .cta {
+.campaign--pitch.-persistent {
+  .cta.-persistent {
+    display: block;
     background-color: $black;
     border: none;
     position: sticky;
@@ -173,19 +174,23 @@
       clear: none;
       text-align: left;
       margin-bottom: 0px;
-      font-size: 18px;
+      font-size: $font-regular;
       float: none;
       vertical-align: middle;
 
       @include media($medium) {
         width: auto;
-        font-size: 24px;
+        font-size: $font-medium;
         float: left;
         padding: 0px;
         display: inline-block;
-        padding: 12px 0px 0px 60px;
+        padding: ($base-spacing/2) 0px 0px ($base-spacing * 2.5);
       }
     }
+  }
+
+  .cta {
+    display: none;
   }
 
   // Need to hide this cloned element that is injected by the fixed-sticky js.
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
@@ -131,66 +131,76 @@
 
 }
 
-.campaign--pitch.-persistent {
+.campaign--pitch {
   .cta.-persistent {
-    display: block;
-    background-color: $black;
-    border: none;
-    position: sticky;
-    top: 0;
-    z-index: 100;
-
-    // Fallback for position:sticky, through Filament Group's fixed-sticky
-    &.fixedsticky-on {
-      position: fixed;
-      width: 100%;
-      max-width: 1440px;
-    }
-
-    > .wrapper {
-      @include clearfix;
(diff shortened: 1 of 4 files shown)


lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
```diff
@@ -130,3 +130,81 @@
   }
 
 }
+
+.campaign--pitch {
+  .cta.-persistent {
+    display: none;
+  }
+
+  &.-persistent-beta {
```
DFurnes: Can you switch this to `.optimizely-*` naming convention we've used for other A/B test classes?

sbsmith86: sure thing.

lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
```diff
@@ -130,3 +130,81 @@
...
+        }
+      }
+
+      .cta__message {
+        display: table-cell;
+        @include span(7 of 12);
+        color: $white;
+        clear: none;
+        text-align: left;
+        margin-bottom: 0px;
```
DFurnes: Can you kill the `px` on zero values for consistency with styles written elsewhere?

lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
```diff
@@ -130,3 +130,81 @@
...
+        font-size: $font-regular;
+        float: none;
+        vertical-align: middle;
+
+        @include media($medium) {
+          width: auto;
+          font-size: $font-medium;
+          float: left;
+          padding: 0px;
```
DFurnes: Likewise.

lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
```diff
@@ -130,3 +130,81 @@
...
+        @include media($medium) {
+          width: auto;
+          float: left;
+          display: inline-block;
+        }
+      }
+
+      .cta__message {
+        display: table-cell;
+        @include span(7 of 12);
```
DFurnes, line 184: Let's move this mix-in to the top of the rule, so it's clear `table-cell` still gets applied & isn't being overridden.

lib/themes/dosomething/paraneue_dosomething/scss/content/_campaign.scss
```diff
@@ -130,3 +130,81 @@
...
+        float: none;
+        vertical-align: middle;
+
+        @include media($medium) {
+          width: auto;
+          font-size: $font-medium;
+          float: left;
+          padding: 0px;
+          display: inline-block;
+          padding: ($base-spacing/2) 0px 0px ($base-spacing * 2.5);
```
DFurnes: The `* 2.5` here seems a bit suspect to me. My vote would be to keep even multiples of our base-spacing value unless there's a good reason to fudge it.

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign--pitch.tpl.php
```diff
@@ -26,6 +26,19 @@
     </div>
   </header>
 
+  <?php if (isset($campaign->secondary_call_to_action)): ?>
+    <div class="cta -persistent js-fixedsticky fixedsticky">
```
DFurnes: Is the `.fixedsticky` class here necessary?

DFurnes:
Some nit-picking, but looks great! :neckbeard: :ship:

sbsmith86:
@DFurnes Thanks for the review! Made those updates.

PR #4334, opened 2015-04-02
Update copy on reportback participation field

### LINKED ISSUE #4325: update copy of rb question
from "Why is this important to you?" to "Why is this campaign important to you?"

campaigns team has seen some confusion about this when a scholarship is attached to a campaign so we're hoping to make it a bit more specific, plus this subtle copy change will hopefully garner better anecdotes to share externally.

we shouldn't have to change anything on the backend because we're still looking for the same info, just asking in a slightly different way.

cc: @barryclark can you please help assign this?

sbsmith86:
## Fixes #4325

Updated the prompt for the "why I participated field" for added clarity.

:sake: :wine_glass: :beers: 

@DoSomething/front-end


DFurnes:
:dancer: :100: :ship: :anchor: :palm_tree: :cocktail:

PR #4340, opened 2015-04-03
Show progress on action page under feature flag

### LINKED ISSUE #4330: display numeric progress towards goal
related to #4288, it would be great to display the numerical progress towards the goal that @angaither has been working on somewhere on the action page so we can easily test on staging.

this would need to be feature flagged because obviously we don't want this showing up yet on prod, but it would be awesome if you output this number somewhere on the action page (no preference, really wherever is easiest for you to add and remove).

@angaither if there are any other details you think @sbsmith86 needs for this that would be awesome

sbsmith86:
## Fixes #4330

Display the new progress variable that @angaither exposed under a feature flag so that it can be tested. This is not the final location/design for this :warning: 

@DoSomething/front-end

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/theme-settings.php
+++ b/lib/themes/dosomething/paraneue_dosomething/theme-settings.php
@@ -51,6 +51,10 @@ function paraneue_dosomething_form_system_theme_settings_alter(&$form, &$form_st
       '#title' => t('Persistent Signup Nav'),
       '#description' => t('Toggles the cta on the campaign pitch page to show as a persistent nav')
     ),
+    'show_campaign_progress' => array(
+      '#title' => t('Show campaign progress'),
+      '#description' => t('Toggles the display of campaign progress on the action page')
+    ),
   );
 
   foreach ($flags as $name => $flag) {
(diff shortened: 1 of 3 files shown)


angaither:
loookin cool. :+1: 
how do you set these theme settings?

sbsmith86:
@angaither in `theme-settings.php`!

angaither:
but there's not a checkbox or something? normally we make it a checkbox, so it isn't a code push to turn it on/off

lib/themes/dosomething/paraneue_dosomething/theme-settings.php
```diff
@@ -51,6 +51,10 @@ function paraneue_dosomething_form_system_theme_settings_alter(&$form, &$form_st
       '#title' => t('Persistent Signup Nav'),
       '#description' => t('Toggles the cta on the campaign pitch page to show as a persistent nav')
     ),
+    'show_campaign_progress' => array(
```
sbsmith86, line 54: I just add new settings to the flag array here @angaither ,  and down below the array is used to create the checkboxes.

angaither, line 54: oh sweet, nice!

PR #4344, opened 2015-04-05
Improve build speed with shallow clones

cjcodes:
Adds a [--shallow-clone](http://api.drush.org/api/drush/commands!make!make.drush.inc/function/make_drush_command/master) parameter to the drush make command, which should improve dev box build time.

I'm not really that familiar with your build process, but I hope this helps in some way...

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -135,7 +135,7 @@ function build {
 
   # Do the build
   echo 'Running drush make...'
-  drush make --prepare-install --no-gitinfofile --no-cache "$BUILD_MAKEFILE" "$WEB_DIR"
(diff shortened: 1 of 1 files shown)


DFurnes:
Faster is always better! :runner:

cjcodes:
Oops. I suppose this is only available in drush 7.x. What version are you dudes running?

sergiitk:
@cjcodes we're using drush 6:
https://github.com/DoSomething/ansible-dosomething/blob/master/files/composer.json

sergiitk:
We will update it at some point though. Assigning this to myself.

sergiitk:
Drush is updated to 7.0.0 everywhere.
Just tested the build on Vagrant, everything works great. I didn't notice any speed improvements, but it's good for traffic and disc space anyways.

Thank you for contribution @cjcodes!
You can merge this now! :+1:

PR #4345, opened 2015-04-06
Remove check for cta to display sign up button

### LINKED ISSUE #4336: Show sign up button with persistent nav turned off
The sign up button in the header on campaign pages is being removed on campaigns with no secondary call to action and the persistent nav feature flagged turned off. It should display even without the secondary CTA in the right place. 

Might not be an issue if we remove the feature entirely.

sbsmith86:
## Fixes #4336

Fixed a bug where, when the persistent nav setting is turned on, the sign up button would only display if there was a secondary call to action. Now, the sign up button will display even if there is no secondary call to action. 

Also, switched `import` statements to `require` statements in `lib.js` for a quick fix for a hoisting issue causing jquery not to be ready before libraries need to use it. 

@DoSomething/front-end

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/lib.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/lib.js
@@ -7,16 +7,16 @@ window.$ = jQuery;
 import 'mailcheck';
 
 // jQuery Once plugin
-import 'jquery-once';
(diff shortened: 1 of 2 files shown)


DFurnes:
:+1: :ship:

PR #4346, opened 2015-04-06
Linting fixes and patched UMD modules

DFurnes:
# Changes
- Fixes a bunch of linting issues that had cropped up over time. Also disables the `quotmark` JSHint linter to enforce double vs. single-quotes for now.
- Remove unused Features module that had a few unused feature tests. We should be using Modernizr tests in almost all cases.
- Uses forked versions of some jQuery dependencies with patched UMD snippet to include CommonJS support. This fixes an issue where `import`s were getting hoisted and `window.jQuery` was undefined for the plugins.


sbsmith86:
:shipit:

PR #4348, opened 2015-04-06
Tumblr bug

### LINKED ISSUE #4337: tumblr dialogue never opens up on permalink
image

cc: @barryclark can you please assign this out? not sure if it's fender or bender task

sbsmith86:
## Fixes #4337

Using a new tumblr share url, also, we were passing in the actual image tag as the reportback image and not the image URL which was breaking the dialog. 

@DoSomething/front-end cc @angaither


DFurnes:
Cool! :+1:

PR #4351, opened 2015-04-07
Wrap sign up button rendering in conditionals so it only prints onces

### LINKED ISSUE #4349: secondary sign up button isn't showing up on pitch for logged in users
image

it is working for anonymous users
image

sbsmith86:
## Fixes #4349

Wrap rendering the secondary sign up button in conditionals so that the `render` function is only called once.


angaither:
:+1:

PR #4353, opened 2015-04-07
Allow `npm link` for dosomething-neue.

DFurnes:
# Changes
- This allows us to work on local Neue copy and see changes reflected in the Vagrant box, without having to compile and copy the whole directory. Read more: [npm link](https://docs.npmjs.com/cli/link).
# Discussion

As far as I can tell, there's no easy way to do this for multiple NPM packages without manually whitelisting each one. Open to any suggestions though if people have ideas!

For review: @DoSomething/front-end @sergii-tkachenko @blisteringherb

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -24,14 +24,13 @@ Vagrant.configure("2") do |config|
     config.vm.network :private_network, ip: "10.11.12.13"
     config.vm.synced_folder ".", "/var/www/dev.dosomething.org", type: "nfs"
 
-    # Allow `npm link`
-    if File.exists?(File.expand_path(File.join("/usr/local/lib/node_modules/dosomething-neue")))
+    # Allow `npm link` for Neue
+    if File.exists?("/usr/local/lib/node_modules/dosomething-neue")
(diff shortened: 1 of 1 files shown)


sergiitk:
Should I just install `dosomething-neue` into the box?

DFurnes:
No, this is for local development – i.e. seeing changes made in your local dev copy of Neue reflected in styles/scripts on the DS app.

Vagrantfile
```diff
@@ -23,6 +23,15 @@ Vagrant.configure("2") do |config|
     # NFS
     config.vm.network :private_network, ip: "10.11.12.13"
     config.vm.synced_folder ".", "/var/www/dev.dosomething.org", type: "nfs"
+
+    # Allow `npm link`
+    if File.exists?(File.expand_path(File.join("/usr/local/lib/node_modules/dosomething-neue")))
```
sergiitk: Why do you need `File.join` here?

sergiitk: Maybe `File.exists?` is enough?

DFurnes: Yep, good call.

PR #4354, opened 2015-04-07
Use patched UMD branches for a couple of modules.

DFurnes:
# Changes
- Uses patched UMD branch for a couple of modules. I had previously patched these on the `master` branch, but I think NPM might be caching that old code. This seems to fix it.


sbsmith86:
:+1:

PR #4358, opened 2015-04-08
Style updates to persistent nav

sbsmith86:
## Address #4308

Making various updates to the persistent nav styles so that the alignment and spacing is managed better at all breakpoints. 

@DoSomething/front-end


DFurnes:
:shipit:

PR #4359, opened 2015-04-08
Decrease spacing on desktop for the persistent nav

sbsmith86:
## Changes

Reduces spacing on the persistent nav on desktop.

@DoSomething/front-end


DFurnes:
:ship:

Issue #4223, opened 2015-03-17
Reportback entity is not properally loaded in preprocess_page


angaither:
The reportback entity loads properly within the `__construct` function, however when it's in a `hook_preprocess_page` it is missing any fids associated with it

two dpms, one from the end of `__construct` and the other inside `dosomething_metatag_get_permalink_vars`
![](http://cl.ly/image/1C072q3r0H3u/Screen%20Shot%202015-03-16%20at%205.06.34%20PM.png) 

> from @aaronschachter 
> So I find that if i hardcode the rbid to an integer in `dosomething_metatag_get_permalink_vars` and call it within `devel/php`, I see the fids array when i output it.  So i think there is something real weird happening taking place calling the `reportback_load` in a `hook_preprocess_page`.
> I also tried outputting `vars` in there, to see if the loaded reportback was already in there, but it's not. i think if the template was using some default Entity template type stuff vs a theme function, we'd get those variables for free in the `$vars`.   The theme function seems to be using `$rb` as just an array, not a loaded `reportback` entity..

I've tried typecasting the rbid as an int before loading it, but it doesn't seem to make a difference

Refs #4211

PR #4360, opened 2015-04-09
Restful

drewish:
These are @aaronschachter's changes from #4260 along with my additions.


jonuy:
I can help with the acceptance testing here. My lenses are colored by the mobile app needs, so that'll be the direction I'm approaching it at.

As far as I'm concerned, I think we can do an initial merge of Restful work when these criteria are fulfilled:
- Get campaigns for an interest group via API
  - The details of which are being covered [here](https://trello.com/c/Sgyt2coY/32-campaigns-data-from-drupal-api)
- Campaigns can be marked as being discoverable on the mobile app
  - Also covered in the above Trello card.

If I can make a GET request to `www.dosomething.org/api/v1/campaigns?filter[tags]=1,2,3,4&on_mobile_app=true` (or something similar) to get relevant campaigns, then I'm sold. Support for these two will at least help with unblocking mobile app engineers from doing work in the next sprint.

There's certainly more work that can continue to be done here afterwards around file uploads and some other new features we've been considering for the mobile app that we can discuss separately.

@drewish for now, let me know if you need clarification on more things. Otherwise I'm ready to help run some testing on my local dev too so that we can move to get this merged in.

drewish:
I got a file uploads mostly working. Need to figure out permissions on it. It works roughly as:

```
curl -X "POST" "http://dev.dosomething.org:8888/api/file-upload" \
    -H "X-CSRF-Token: 3QYKfYSZjkHut73lD-qvy2fFklRU0opj5SJA0GyPDTk" \
    -F "foo=@myfilename"
```

drewish:
I pulled the file related changes out of this. It should be ready for review.

drewish:
Wanted to drop a couple of sample URLs from the [Trello card](https://trello.com/c/Sgyt2coY/32-campaigns-data-from-drupal-api) in here as some basic documentation.

This is how you'd get campaigns withe the `field_on_mobile_app` is true with a single action type taxonomy term:

```
http://dev.dosomething.org:8888/api/v1.0/campaigns
  ?filter[on_mobile_app]=1
  &filter[action_type]=10
```

Here's how you'd find campaigns with one of two action type taxonomy terms:

```
http://dev.dosomething.org:8888/api/v1.0/campaigns
  &filter[action_type][operator]=IN
  &filter[action_type][value][0]=10
  &filter[action_type][value][1]=11
```

It's not pretty but it does seem to work correctly.

jonuy:
@drewish I've pulled this branch down onto my local box. It's been a while since I've been in our Drupal-land, what do I need to do in order to actually test this? Are there modules I'll need to enable? Databases to update? etc.

Also, for the `v1.0` part of the URL, could we change that to be just `v1` instead to match the pattern we've set in other URL paths?

As for the code side of things... @angaither would you be able to be the code reviewer for this one?

drewish:
@jonuy regarding the `/v1.0` prefix I'm happy to dig in and try and see if we can get it to use `/v1` but I thought I remembered @aaronschachter mentioning something about it in his PR. Here's what he had to say about it https://github.com/DoSomething/dosomething/pull/4260#issuecomment-83752064 :

> Note that you can actually create different objects based on the v1.0. So there can be a v1.1 with a different format for a property.
> 
> What seems kinda weird is having a half this / half that approach.. like if we kept the custom Services endpoints we have this far .... however we could create a new v1.0 API endpoint and eventually the plain v1 could be depcreated.

He raises a good point that it's probably best to have services and restful working on different paths rather than fighting over the same prefix. If the plan is to port the functionality over eventually we could change it to `/v2.0` so it's clear it works differently.

lib/modules/dosomething/dosomething_api/plugins/restful/campaigns/1.0/DoSomethingCampaignsResource.class.php
```diff
@@ -0,0 +1,98 @@
...
+      'property' => 'field_call_to_action',
+    );
+
+    $public_fields['on_mobile_app'] = array(
+      'property' => 'field_on_mobile_app',
+    );
+
+    // This is so wtf.
+    // If I name it as 'cover_image' it doesn't work.
+    // If I name it as just 'image' it works.
```
angaither: maybe update this to @todo: as why doesn't `cover_image` work :wink:

lib/modules/dosomething/dosomething_api/plugins/restful/assets/1.0/DoSomethingAssetsResource.class.php
```diff
@@ -0,0 +1,62 @@
...
+      $output = array();
+      foreach ($value as $item) {
+        $output[] = $this->imageProcess($item);
+      }
+      return $output;
+    }
+    return array(
+      'id' => $value['fid'],
+      // @todo This should direct to the new files resource.
+      'self' => file_create_url($value['uri']),
```
angaither, line 54: are you returning image urls here? we have a custom function for that that returns the image cached version
`dosomething_image_get_themed_image_by_fid`
https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_image/dosomething_image.module#L137

angaither, line 54: not sure if that will help, of if you're returning multiple image styles here...

lib/modules/dosomething/dosomething_api/plugins/restful/files/1.0/DoSomethingFilesResource.class.php
```diff
@@ -0,0 +1,70 @@
...
+      'filesize' => $value['filesize'],
+      'width' => $value['width'],
+      'height' => $value['height'],
+      'styles' => $value['image_styles'],
+    );
+  }
+
+  protected function fileProcess($value) {
+    // This feels wrong because you'd expect the file entity to be loaded already.
+    $value = (array) file_load($value);
```
angaither, line 58: likewise we have a custom reportback file load function `reportback_file_load`

https://github.com/DoSomething/dosomething/blob/34745773ccd6e396a0380b86c7574559b3e4dc4f/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module#L399

angaither, line 58: Also, we were also seeing issues with the entity not being loaded properly in places it should have been #4223

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.features.field_instance.inc
```diff
@@ -1685,6 +1685,55 @@ function dosomething_campaign_field_default_field_instances() {
...
+  // Exported field_instance: 'node-campaign-field_on_mobile_app'
+  $field_instances['node-campaign-field_on_mobile_app'] = array(
+    'bundle' => 'campaign',
+    'default_value' => array(
+      0 => array(
+        'value' => 0,
+      ),
+    ),
+    'deleted' => 0,
+    'description' => 'Check this for the campaign to appear on the mobile apps.',
```
angaither, line 1697: `t()`

angaither:
@drewish looks pretty good, but there's a couple times in here where you have code commented out, if it's unused, please remove it before merge! :100:

jonuy:
I think I did something wrong with the install/make on my local box because on a first run the `on_mobile_app` field was there, but on re-install it wasn't. Either way, on the second go around, I just manually added it to the Campaign content type and tested from there.

I successfully tested with queries against campaigns that had the `on_mobile_app` flag set and different tags set. These all worked :smile: :
- `http://dev.dosomething.org:8888/api/v1.0/campaigns`
- `http://dev.dosomething.org:8888/api/v1.0/campaigns?filter[on_mobile_app]=1`
- `http://dev.dosomething.org:8888/api/v1.0/campaigns?filter[on_mobile_app]=1&filter[tags]=491`
- `http://dev.dosomething.org:8888/api/v1.0/campaigns?filter[tags][operator]=IN&filter[tags][value][0]=491&filter[tags][value][1]=97&filter[on_mobile_app]=1`

Only suggestions:
- Maybe move the `on_mobile_app` flag into the **Taxonomy and Discovery** section on the Campaign edit page. At least that's where I saw it on my first install. I totally could've messed something up there too
- Have the Restful module enabled by default. Again not sure if I messed something up when I tried installing, but I had to manually flip the switch on for the module.

jonuy:
Oh, also to the `v1.0` comment I made earlier... tl;dr - let's just leave it as `v1.0` for the time being.

From a client's perspective, having to do `v1` on some paths and `v1.0` on others can be confusing. Internally for us, I imagined the distinction we could make between what services handles and what restful handles would be at that one level past the `v1.0` part - ie: `.../campaigns` or `.../reportbacks`, etc.

But depending on how the client code's done, it really does just boil down to a URL path string change. So yea, limited effort on the client side + your reasons above + the effort it might take on the Drupal side to find a way to change it = doesn't seem worth it.

angaither:
@jonuy I feel like we could handle the url path on the fastly side as well

jonuy:
@angaither true! :star:

jonuy:
@drewish Any questions on the comments above? Feel free to make any edits based on the feedback and then just merge it on in when ready.

lib/modules/dosomething/dosomething_api/plugins/restful/campaigns/1.0/DoSomethingCampaignsResource.class.php
drewish: Yeah good question, that was one of @aaronschachter's open questions. I'll look into it.

lib/modules/dosomething/dosomething_api/plugins/restful/files/1.0/DoSomethingFilesResource.class.php
drewish, line 58: In regards to your first comment about using `reportback_file_load` I'm not sure these files are limited to report backs. The name makes me think that function is specific to that type of file. Am I wrong?

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.features.field_instance.inc
drewish, line 1697: I thought the feature pulled out those strings for translation down here: https://github.com/DoSomething/dosomething/pull/4360/files#diff-81d7f21ce5de35802663f0371af1d1d1R3082

drewish:
Hey sorry I wasn't getting the notifications for some reason. I'll try to sort that out. 

@jonuy 

> Maybe move the on_mobile_app flag into the Taxonomy and Discovery section on the Campaign edit page. At least that's where I saw it on my first install. I totally could've messed something up there too

That's exactly where it should be. Might try a `drush fr dosomething_campaign` after you check out the branch?

> Have the Restful module enabled by default. Again not sure if I messed something up when I tried installing, but I had to manually flip the switch on for the module.

@aaronschachter had left it off for safety sake so it could be merged but not activate the new code. I'm happy to go either way. @angaither do you have strong feelings?

lib/modules/dosomething/dosomething_api/plugins/restful/files/1.0/DoSomethingFilesResource.class.php
drewish, line 58: Or maybe we can use that because the only files users upload are report back files?

angaither, line 58: correct, users can only upload reportback files

angaither:
@drewish we can leave it off so it doesn't affect international sites

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.features.field_instance.inc
angaither, line 1697: :+1:

jonuy:
@drewish Cool, yea it seems I may have botched something on my install where the flag wasn't showing up. And leaving it off by default particularly for international also sounds good.

Seems like a couple changes remain that Andrea pointed out: removing unused commented out code, adding an `@todo` note, using `reportback_file_load`, and also possibly using `dosomething_image_get_themed_image_by_fid`. Pending those, you're welcome to merge it on in when ready! :)

lib/modules/dosomething/dosomething_api/plugins/restful/assets/1.0/DoSomethingAssetsResource.class.php
drewish, line 54: Oh I think he was referring to having it link to something like `http://dev.dosomething.org:8888/api/v1.0/assets/1145`. Though that made me realize I should give a close look at how all the assets/files stuff fits together.

drewish:
I removed the one comment about cover_image since it did seem to work. The rest of the todos I'd like to leave until I can talk to Aaron about how he envisioned them working. I think this should be good to go for now but will definitely need some follow up.

PR #4270, opened 2015-03-20
Overrides MailChimp settings for 26+ Club.

sergiitk:
#### What's this PR do?
- Introduces 3 new setting overrides for 26+ Club:
  1. `dosomething_signup_mailchimp_26plusclub_list_id`: Overrides `mailchimp_list_id`
  2. `dosomething_signup_mailchimp_26plusclub_grouping_id_register`: Overrides `mailchimp_grouping_id` for registrations
  3. `dosomething_signup_mailchimp_26plusclub_group_name_register`: Overrides `mailchimp_group_name` for registrations
- Extends and imporves [Third Party Opt-Ins](http://dev.dosomething.org:8888/admin/config/dosomething/opt_in) configuration interface:
  image
- Turns off MBP for 26+ Club signups
- Overrides settings for 26+ Club registrations
#### Any background context you want to provide?

New user payload before:

``` php
array (size=11)
  'activity' => string 'user_register' (length=13)
  'email' => string 'sergii+test-032015-7@dosomething.org' (length=36)
  'uid' => string '1704217' (length=7)
  'email_template' => string 'mb-user-register-US' (length=19)
  'merge_vars' => 
    array (size=2)
      'MEMBER_COUNT' => string '3 million' (length=9)
      'FNAME' => string 'Sergii' (length=6)
  'mailchimp_list_id' => string '* hidden *' (length=10)
  'birthdate' => string '580780800' (length=9)
  'subscribed' => int 1
  'email_tags' => 
    array (size=1)
      0 => string 'drupal_user_register' (length=20)
  'activity_timestamp' => int 1426871199
  'application_id' => string 'US' (length=2)
```

New user payload after:

``` php
array (size=13)
  'activity' => string 'user_register' (length=13)
  'email' => string 'sergii+test-032015-7@dosomething.org' (length=36)
  'uid' => string '1704233' (length=7)
  'email_template' => string 'mb-user-register-US' (length=19)
  'merge_vars' => 
    array (size=2)
      'MEMBER_COUNT' => string '3 million' (length=9)
      'FNAME' => string 'Sergii' (length=6)
  'mailchimp_list_id' => string '* hidden-2 *' (length=10)
  'birthdate' => string '580780800' (length=9)
  'subscribed' => int 1
  'mailchimp_grouping_id' => string '9613' (length=4)
  'mailchimp_group_name' => string 'Web Registrations' (length=17)
  'email_tags' => 
    array (size=1)
      0 => string 'drupal_user_register' (length=20)
[...]

PR #4361, opened 2015-04-09
Improves dosomething_signup_update_7019().

sergiitk:
This is to make few improvements into `dosomething_signup_update_7019()`:
- Add `DS_MB_26PLUSCLUB_ENABLED` env variable to set preset "Enable 26+ Club"
- Renames other env variables in the hook install according to actual variable name change

Ref #4270.


deezone:
:+1:

Issue #4049, opened 2015-02-24
Campaign gallery API endpoint improvements


aaronschachter:
Per convo w @mikefantini 
- [x] Return `promoted` images as well as `approved` in the Gallery endpoint
- [x] Add a new query parameter to filter by status (will allow us to only retrieve promoted files)
- [x] Add a new query parameter to filter by File fid's (can exclude multiples by comma, e.g. `?exclude=12,890,23412`

This would need corresponding changes on the front-end side to alter the logic gathering Gallery data to display.

weerd:
Collecting a bunch of tasks that need to be completed to smoothly upgrade the API endpoints for the campaign gallery:

- [ ] Add a feature flag to toggle using `/api/v1/campaigns/{nid}/gallery.v2.json` so we don't break everything.

angaither:
@weerd is this something that you did with the reportback api already?

weerd:
Not the "excluding" portion of it. We could close this and re-open an issue for that feature alone if it's something we want later on?

aaronschachter:
Didn't we want the exclusion for preventing duplicates on the web gallery? e.g. reportback item 23 is promoted and selected for the 6. it can easily show up agian in the approved section (displayed chronologically after the 6 random promoted)

mikefantini:
@weerd is looking into this as he tackles #5170

PR #4366, opened 2015-04-10
Experimental oop api

weerd:
Here's a lot of the progress with the `/reportbacks` and `/reportback-items` endpoints.

Example Reportback Retrieve

``` javascript
// REQUEST: http://dev.dosomething.org:8888/api/v1/reportbacks/1209.json

// RESPONSE:
{
  data: {
  id: "1209",
  created_at: "1427226250",
  updated_at: "1428442705",
  flagged: false,
  quantity: 12,
  why_participated: "Because I like doing some things!",
  items: {
    total_retrieved: 2,
    data: [
      {
        id: "2633",
        caption: "Some awesome caption here.",
        uri: "http://dev.dosomething.org:8888/api/v1/reportback-items/2633",
        media: {
        uri: "http://dev.dosomething.org:8888/sites/default/files/styles/480x480/public/uid_170415.jpeg",
        type: "image"
        },
        created_at: "1427226250",
        status: "approved"
      },
      {
        id: "2634",
        caption: "Some other awesome caption here.",
        uri: "http://dev.dosomething.org:8888/api/v1/reportback-items/2634",
        media: {
        uri: "http://dev.dosomething.org:8888/sites/default/files/styles/480x480/public/uid_170412.jpeg",
        type: "image"
        },
        created_at: "1427226585",
        status: "approved"
      }
    ]
  },
    campaign: {
      id: "1321",
      title: "Hunt Day 5: Band-Aid Brigade"
    },
    user: {
      id: "100005"
    }
  }
}
```

Example Reportback Items Collection

``` javascript
// REQUEST: http://dev.dosomething.org:8888/api/v1/reportback-items.json?campaigns=345,1321&count=2&random=true

// RESPONSE:
{
  total_retrieved: 2,
  data: [
    {
      id: "233",
      caption: "Monuments",
      uri: "http://dev.dosomething.org:8888/api/v1/reportback-items/2313",
      media: {
        uri: "http://dev.dosomething.org:8888/sites/default/files/styles/480x480/public/reportbacks/1321/uid_1713.jpg",
        type: "image"
      },
      created_at: "1423695826",
      status: "approved",
      reportback: {
        id: "1838",
        created_at: "1422458628",
        updated_at: "1424298714",
        flagged: false,
        quantity: 15,
        why_participated: "Good times! Yo!"
      },
      campaign: {
        id: "1321",
        title: "Hunt Day 5: Band-Aid Brigade"
      },
      user: {
        id: "100001"
      }
    },
    {
      id: "984",
      caption: "DoSomething? Just did!",
      uri: "http://dev.dosomething.org:8888/api/v1/reportback-items/984",
      media: {
        uri: "http://dev.dosomething.org:8888/sites/default/files/styles/480x480/public/reportbacks/1321/uid_129-nid_1321-0.jpg?itok=S20O7GvV",
        type: "image"
      },
      created_at: "1408659318",
      status: "approved",
      reportback: {
        id: "1747",
        created_at: "1408659318",
        updated_at: "1408659318",
        flagged: false,
        quantity: 1,
        why_participated: "test"
      },
      campaign: {
        id: "1321",
        title: "Hunt Day 5: Band-Aid Brigade"
      },
      user: {
        id: "100002"
      }
    }
  ]
}
```

Along with the new resources created in Services, there are also 3 new classes created in the **dosomething_reportback** module: `ReportbackTransformer`, `Reportback` and `ReportbackItem`.

`Reportback` and `ReportbackItem` are subclasses of `ReportbackTransformer` which allows all the data transformations to be dictated in one class via an assortment of methods. So regardless of which endpoint is hit, and whether retrieving a single item or a collection of items all the data is always returned in a standard format as expected. 

The data in the JSON response is also properly typed so no `0 or 1` when it should be a boolean `false or true`, and where expected numbers are integers instead of strings (except for ids).

There's still a bit more work to be done on the `/reportback.json` index method, and also allowing for and including pagination data in the responses. More to come on that, but this PR was already getting ginormous ;)

@jonuy @drewish @angaither 

CC: @DeeZone @DFurnes 

*Note: There currently is no Feature for this endpoint, so you will need to manually toggle it on within the Services Admin.


lib/modules/dosomething/dosomething_api/dosomething_api.module
```diff
@@ -4,12 +4,34 @@
...
-include_once 'resources/reportback_resource.inc';
-include_once 'resources/term_resource.inc';
+module_load_include('inc', 'dosomething_api', 'dosomething_api.features');
+module_load_include('inc', 'dosomething_api', 'resources/campaign_resource');
+module_load_include('inc', 'dosomething_api', 'resources/member_resource');
+module_load_include('inc', 'dosomething_api', 'resources/reportback_resource');
+module_load_include('inc', 'dosomething_api', 'resources/reportback_item_resource');
+module_load_include('inc', 'dosomething_api', 'resources/reportback_file_resource');
+module_load_include('inc', 'dosomething_api', 'resources/signup_resource');
+module_load_include('inc', 'dosomething_api', 'resources/term_resource');
```
angaither: I would vote to leave these at `include_once` since we are just including files. it's the [standard we've been using to include files](https://github.com/DoSomething/dosomething/search?utf8=%E2%9C%93&q=include_once)

weerd:
The nice thing about passing off a lot of the logic to dedicated classes is that if we decide to switch from Services Module to RESTful Module, there isn't much overhaul. In whatever fashion RESTful module addresses resources and catches query parameters, we just pass those along by creating a new instance of the respective class (either `Reportback` or `ReportbackItem`) and let it do it's thing ;)

lib/modules/dosomething/dosomething_api/dosomething_api.module
weerd: Hokey dokey! :dancer:

deezone:
First time taking a look at what you're up to. Thrilled to see `reportback-items.json` as I can think of a great application in the Message Broker generated transactional campaign signup and digest messages.

Nice work!

weerd:
thanks @DeeZone! :)

lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
```diff
@@ -0,0 +1,111 @@
...
+    ),
+
+  );
+
+  return $reportback_item_resource;
+}
+
+
+function _reportback_item_resource_access() {
+  // Temp universal access for now.
```
angaither: just add `@todo` here to make this easier to find

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -959,16 +1003,30 @@ function dosomething_reportback_get_reportback_files_query($params = array()) {
...
   if (isset($params['status'])) {
     $query->condition('rbf.status', $params['status']);
   }
+
   if (isset($params['nid'])) {
-    $query->condition('rb.nid', $params['nid']);
+    if (is_array($params['nid'])) {
+      $query->condition('rb.nid', $params['nid'], 'IN');
+    } else {
+      $query->condition('rb.nid', $params['nid']);
```
angaither, line 1021: is there any case this could fail if nid is a string, or an int? does there need to be additional validation here, or is that elsewhere and I'm missing it?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.reportback.inc
```diff
@@ -0,0 +1,119 @@
...
+    // Not ideal that this is NULL instead of FALSE but due to how logic happens in original query function. It should be updated!
+    // Logic currently checks for isset() instead of just boolean, so won't change until endpoints switched.
+    $filters['random'] = $parameters['random'] === 'true' ? TRUE : NULL;
+
+    // @TODO: Need to flesh this out. Temporarily disabled.
+    // $query = dosomething_reportback_get_reportbacks_query_result($filters, $filters['count']);
+    // $reportbacks = services_resource_build_index_list($query, 'reportbacks', 'rbid');
+    $reportbacks = null;
+
+    if (! $reportbacks) {
```
angaither: I don't think we typically put a space before the bang!

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
weerd, line 1021: Hmmm, so like if a query parameter is passed for `?campaigns=sometext` instead of `?campaigns=362`? I think MySQL just ignores it, but we should probably validate for it. I can do that within one of the classes. I'll add it as a TODO to add a method in the `RBTransformer` class, since we'll likely need it for more than just `nid`.

lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
weerd: K.

angaither:
looking good, just a couple of style questions

NOIIICE @weerd

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
weerd, line 1021: Actually, I just need to modify the `formatData()` Transformer method cause it's already doing some stuff in there.... the tricky part is that we can now pass a single int as `?campaigns=362` or a series of ints as `?campaigns=362,1301,293`... The `formatData()` methods takes these comma separated values and turns them into an array of values that then get passed into the query. So to validate, I'd need to check if each of these are actually "numbers" if wanting numbers otherwise toss it out... Maybe have to think it through a bit.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.reportback.inc
weerd: Jeff Way is subconsciously affecting my code style :dancers:

DFurnes: (╯°□°）╯︵ ┻━┻

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
angaither, line 1021: totally, I would just make sure that it doesn't cause a failure, and if it doesn't fail it can just be added as a todo

weerd, line 1021: Yeah, I've tried it a bit before and it doesn't fail... just give unexpected results a bit, since it defaults to grabbing items without any filter on it (for like items for a specific nid). No failure per se.

weerd:
References #4049

Issue #4364, opened 2015-04-09
Optimizely A/B/C/D tests on pitch page


sbsmith86:
Set up pitch page to work for 4 different optimizely tests. 
- [x] Persistent nav with sign up count cta
- [x] Persistent nav with secondary cta
- [x] Current implementation with sign up count cta
- [x] Current implementation with secondary cta

cc @mikefantini @lkpttn

sbsmith86:
Here is the code needed to run the optimizely tests. @mikefantini 

### Current Implentation - with secondary cta

Do nothing! 

### Current Implementation - with count

`jQuery(".campaign--pitch").find(".cta").removeClass("optimizely-hide-count").addClass("optimizely-show-count");`

### Persistent Nav - secondary cta

`jQuery(".campaign--pitch").addClass("optimizely-persistent");`

### Persistent Nav - with count

`jQuery(".campaign--pitch").addClass("optimizely-persistent").find(".cta").removeClass("optimizely-hide-count").addClass("optimizely-show-count");`

sbsmith86:
Going to close this now, since testing has been done.

DFurnes:
Should we make an issue to remove the Optimizely classes now that tests are over?

sbsmith86:
@DFurnes  Hmm not sure if testing is done, I should have been more clear. Closing this issue because the tests have been set up. 

@mikefantini are we done with these tests? If so, I can create the issue to clean up the optimizely work.

DFurnes:
Oh herpa derp. :grimacing:

mikefantini:
@sbsmith86 @DFurnes tests are still running. We should have results by early next week and then I'll create an issue for cleanup at that time.

DFurnes:
Sounds good!

PR #4367, opened 2015-04-10
Set up pitch template for Optimizely testing

sbsmith86:
## Ref #4364

We are setting up the campaign pitch page to have 4 Optimizely tests run (outlined in #4364). Changes in this PR: 
- Remove theme setting that was controlling the persistent CTA. It was just there for review purposes. Now we needed on the page so that it can be toggled on/off for the testing.
- Allows us to test both the cta as it is now, and the persistent version with or without the sign up count included.

@DoSomething/front-end


DFurnes:
Shippy. :ship:

weerd:
Late to the convo :( but yeah looks good!

PR #4375, opened 2015-04-13
Reportback endpoints updates

weerd:
Addresses allowing you to pass multiple status types to get a collection of results:

```
/api/v1/reportback-items.json?campaign=362&status=promoted,approved
```

*Note: for the time being, a campaign needs to be specified due to how a legacy function works.

Also does some cleanup and fixes a bug or two, in addition to adding `total` for a collection to being adding data for pagination.

@jonuy @angaither

CC: @drewish @DFurnes

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.reportback_transformer.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.reportback_transformer.inc
@@ -60,12 +60,15 @@ abstract class ReportbackTransformer {
       'id' => $data->rbid,
       'created_at' => $data->created,
       'updated_at' => $data->updated,
-      'flagged' => (int) $data->flagged ? TRUE : FALSE,
       'quantity' => (int) $data->quantity,
     );
 
-    if ($data->why_participated) {
+    if (isset($data->why_participated)) {
       $output['why_participated'] = $data->why_participated;
+    }
+
+    if (isset($data->flagged)) {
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.reportback_transformer.inc
```diff
@@ -7,6 +7,20 @@ abstract class ReportbackTransformer {
     module_load_include('inc', 'services', 'services.module');
   }
 
+
```
angaither, line 10: add some comments here about what's going on

angaither:
other than the double for loop, which I have some questions about, lookin :+1:

weerd:
Yeah, wasn't keen on the double for loop :-| If you have an alternative idea, I'm down for trying it out :)

It's partly due to the prior function that was created that caches the total reportbacks by status type for a campaign id `dosomething_reportback_get_reportback_total_by_status()`. It only accepts a single status type passed, which makes sense for also how that data is being store in the DB.

So I needed to use the double for loop, since we can request multiple campaigns and multiple statuses for reportback items for each of those campaigns. So we loop through the number of campaign ids passed and for each we collect the count for the different types of status indicated.

Hopefully that makes more sense.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.reportback_transformer.inc
weerd: There is a comment above the code. Should I be more detailed?

PR #4377, opened 2015-04-14
Fixes indexing tokens on password reset page.

sergiitk:
Restrict indexing pages when reset password token is set.
Fixes https://www.crowdcurity.com/dosomething/reports/40.


sergiitk:
Hm, it looks like I also need to remove `<meta name="robots" content="NOODP" />` tag. Reworking.

angaither:
:+1:

PR #4379, opened 2015-04-14
Disable reportback form when image is invalid

### LINKED ISSUE #4054: Disable reportback form submission when image is invalid
When a user uploads an image that is not a valid size or file type, submission of the whole reportback form should be disabled until the user uploads a valid image.

sbsmith86:
## Fixes #4054

Disables the submit button on the reportback form if the user uploads an image that is not valid. 

@DoSomething/front-end


weerd:
Looks cool :) :+1:

PR #4380, opened 2015-04-14
Updates to add pagination to reportback-items endpoint.

### LINKED ISSUE #4381: Add pagination for Reportback Items collection endpoint
Paginate all the things!

weerd:
## Fixes #4381

:sparkles: _Now with PAGINATION!_ :sparkles: 

Request:

```
http://dev.dosomething.org:8888/api/v1/reportback-items.json?&campaigns=362,1321&status=promoted&count=4&page=3
```

Response:

```
{
  pagination: {
    total: 20,
    per_page: 4,
    current_page: 3,
    total_pages: 5,
    prev_uri: "http://dosomething.org/api/v1/reportback-items.json?&campaigns=362,1321&status=promoted&count=4&page=2",
    next_uri: "http://dosomething.org/api/v1/reportback-items.json?&campaigns=362,1321&status=promoted&count=4&page=4"
  },
  retrieved_count: 4,
  data: [
    {...},
    {...},
    {...},
    {...}
  ]
}
```

Something's to note... currently (I think only on staging and local, maybe production too), the value that gets calculated for the `total` is based off of what counts are in cache. These counts get updated when Reportbacks are submitted for a campaign, etc... However, I believe there's an issue with them not updating properly if reportbacks are deleted (which we've done a bit on staging). So on local testing, the `total` and `total_pages` can be a little skewed since the number reported for the calculation are off.

The code could probs also benefit from a refactor pass, but wanted to get something up. Might also make sense to rip out the pagination methods into their own class specifically for API stuff :dancer: 

I place all the pagination data in a `pagination` object. Hopefully that's cool :) Also we should discuss the property name differences between Northstar and this, so we can agree upon some standard naming...

@uy @angaither @drewish

CC: (might be of interest to) @DFurnes @barryclark @DeeZone

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.reportback_transformer.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.reportback_transformer.inc
@@ -14,6 +14,7 @@ abstract class ReportbackTransformer {
    * @param $data
    * @param $parameters
    * @param $endpoint
+   *
    * @return null|string
    */
   protected function getNextPageUri($data, $parameters, $endpoint) {
@@ -36,6 +37,7 @@ abstract class ReportbackTransformer {
    * @param $data
    * @param $parameters
    * @param $endpoint
+   *
    * @return null|string
    */
   protected function getPrevPageUri($data, $parameters, $endpoint) {
@@ -57,6 +59,7 @@ abstract class ReportbackTransformer {
    *
    * @param $parameters
    * @param $endpoint
+   *
    * @return string
    */
   protected function getPathParameters($parameters, $endpoint) {
@@ -84,6 +87,7 @@ abstract class ReportbackTransformer {
    * each status requested.
    *
    * @param array $parameters Query parameters.
+   *
    * @return int Total number of reportback items.
    */
   protected function countItems($parameters) {
@@ -101,6 +105,7 @@ abstract class ReportbackTransformer {
 
   /**
    * @param string $data Single or multiple comma separated data items.
+   *
    * @return string or array
    */
   protected function formatData($data) {
@@ -120,6 +125,7 @@ abstract class ReportbackTransformer {
    * @param $parameters
    * @param $count
    * @param $endpoint
+   *
    * @return array
    */
   protected function paginate($parameters, $endpoint) {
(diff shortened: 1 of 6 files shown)


weerd:
Debating whether it might make more sense to pull `count` out of the `pagination` object and have it as meta data at the same level as the `data` object. 

If we don't want/need pagination object (for example if you request a random assortment of 20 items... can't really have pagination when asking for random) then we would lose out on the `count` data. Just thinking out loud. Would appreciate feedback.

weerd:
Ok! Based on a chat with @jonuy regarding above comment, removed `count` from inside pagination and moved as sibling property of `data` array, and renamed it to `retrieved_count` so it doesn't get confused with the `count` query parameter!

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.reportback_transformer.inc
```diff
@@ -9,6 +9,77 @@ abstract class ReportbackTransformer {
 
 
   /**
+   * Get the URI for the next page in the collection of data.
+   *
+   * @param $data
+   * @param $parameters
+   * @param $endpoint
```
sergiitk, line 16: You missed a newline between `@param` and `@return`. Pls fix here and everywhere below.

weerd, line 16: K :)

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.reportback_transformer.inc
```diff
@@ -44,6 +120,48 @@ abstract class ReportbackTransformer {
...
+  /**
+   * Get the offset for requests based on specified page number and count of items.
+   *
+   * @param $page
+   * @param $count
+   *
+   * @return int
+   */
+  protected function setOffset($page, $count) {
+    $page = ($page > 0) ? $page -= 1 : $page;
```
sergiitk: This looks strange. In case when `$page > 0 == FALSE`, you just setting `$page` to `$page`, which looks rudimentary.
Would the following achieve what you wanted?

``` php
if ($page > 0) {
  $page -= 1;
}
```

sergiitk:
Code looks :+1:. I'd suggest to test it very careful :)

PR #4382, opened 2015-04-15
Remove Alternate BG FID Field on campaigns

### LINKED ISSUE #4026: Remove the Alt BG File fid field.
let's hold off on working on this until next standup at earliest

sbsmith86:
## Fixes #4026

Removes the alternate background FID field on the custom settings page on campaigns.


angaither:
I think we have some of these set in the DB, should we remove them as well?

sbsmith86:
@angaither ah. Does the update hook i put in remove them?

angaither:
@sbsmith86 let's try

angaither:
I tried it locally, and don't think it removes variables that are already set I think you'll have to do something like this to remove all instances of the variable

```
  db_delete('dosomething_helpers_variable')
      ->condition('entity_type', ’node')
      ->condition('name', ‘alt_bg_fid')
      ->execute();
```

sbsmith86:
@angaither updated! ran the hook locally and check for instances of that field, looks like it was removed.

angaither:
@sbsmith86 cool :+1:

sbsmith86:
Thanks for the help! @angaither

Issue #3311, opened 2014-11-04
DS Build is broken


aaronschachter:
New unseen error, occurring in `ds build` as of Tuesday afternoon 11/4

cap aborted!
SSHKit::Command::Failed: cd '/var/www/beta.dosomething.org/releases/20141104201434'; /var/www/beta.dosomething.org/releases/20141104201434/bin/ds build stdout: Nothing written
cd '/var/www/beta.dosomething.org/releases/20141104201434'; /var/www/beta.dosomething.org/releases/20141104201434/bin/ds build stderr: Nothing written
/var/lib/jenkins/.rvm/gems/ruby-2.0.0-p195/gems/sshkit-1.3.0/lib/sshkit/command.rb:94:in `exit_status='
/var/lib/jenkins/.rvm/gems/ruby-2.0.0-p195/gems/sshkit-1.3.0/lib/sshkit/backends/netssh.rb:142:in`block (4 levels) in _execute'
/var/lib/jenkins/.rvm/gems/ruby-2.0.0-p195/gems/net-ssh-2.8.0/lib/net/ssh/connection/channel.rb:551:in `call'
/var/lib/jenkins/.rvm/gems/ruby-2.0.0-p195/gems/net-ssh-2.8.0/lib/net/ssh/connection/channel.rb:551:in`do_request'
/var/lib/jenkins/.rvm/gems/ruby-2.0.0-p195/gems/net-ssh-2.8.0/lib/net/ssh/connection/session.rb:561:in `channel_request'
/var/lib/jenkins/.rvm/gems/ruby-2.0.0-p195/gems/net-ssh-2.8.0/lib/net/ssh/connection/session.rb:465:in`dispatch_incoming_packets'
/var/lib/jenkins/.rvm/gems/ruby-2.0.0-p195/gems/net-ssh-2.8.0/lib/net/ssh/connection/session.rb:221:in `preprocess'
/var/lib/jenkins/.rvm/gems/ruby-2.0.0-p195/gems/net-ssh-2.8.0/lib/net/ssh/connection/session.rb:205:in`process'
/var/lib/jenkins/.rvm/gems/ruby-2.0.0-p195/gems/net-ssh-2.8.0/lib/net/ssh/connection/session.rb:169:in `block in loop'
/var/lib/jenkins/.rvm/gems/ruby-2.0.0-p195/gems/net-ssh-2.8.0/lib/net/ssh/connection/session.rb:169:in`loop'
[...]

PR #4386, opened 2015-04-15
Services module security update to 3.12.

sergiitk:
See https://www.drupal.org/node/2471879

In addition, we no longer need the patch, it has been included to the module:
https://www.drupal.org/node/2200269


sergiitk:
Build works well.

mshmsh5000:
:+1: Looks good and like :beer:

sergiitk:
Note: patch introduced in #3311.

PR #4396, opened 2015-04-20
Update reportback totals on cron

### LINKED ISSUE #4326: Progress count updates
The signup count doesn't update correctly if someone updates their reportback.

One idea is: 
Stores the total quantity of a reportback as well as new row called `updated_quantity` to store the difference of the original quantity to the updated quantity. 

Another idea is to have a cron run nightly to update the numbers that just queries the database. 

This is needed to get the total reportback totals for goals/progress

TODO: spec out the query that is needed
does this work if somebody reduces their quantity?
how will this work with overrides?
would we need to remove all vars previously set?

### LINKED ISSUE #4335: set progress/signup count variables for all campaigns
Currently, in order to see the signup/progress counts you have to visit the campaign helpers variable page, we should create an update hook or cron job to set this for all nodes, before we go live with this feature.

angaither:
Add a cron job to run the reportback totals every hour, remove duplicate counting problems. 

However, this removes campaign the ability to override totals. I think that will have to be a different solution. @mikefantini 

Fixes #4335, Fixes #4326

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.variable.inc
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.variable.inc
@@ -165,6 +165,7 @@ function dosomething_helpers_variable_form($form, &$form_state, $node) {
       '#title' => t('Progress'),
       '#description' => t('The total approved reportback quantity so far is: ' . $rb_progress),
       '#default_value' => $vars['progress'] ?: $rb_progress,
+      '#disabled' => TRUE,
     );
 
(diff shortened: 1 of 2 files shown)


mikefantini:
so @angaither this works for the majority of campaigns, but we would still need another solution for when we want campaign leads to be able to adjust the number manually (i.e. in Comeback Clothes or Teens for Jeans)?

sergiitk:
Code looks good to me :+1:

angaither:
@mikefantini correct

PR #4400, opened 2015-04-21
Remove test class

sbsmith86:
I accidentally did not remove the optimizely test class when I committed my latest changes. So removing that now. Sorry! 

d1959bfd692

@DoSomething/front-end


weerd:
That optimizely class is so persistent!! :+1:

DFurnes:
wocka wocka

PR #4408, opened 2015-04-22
Removed cta message scope to account for different mark up

### LINKED ISSUE #4403: # of ppl doing and secondary CTA are both showing at the same time
image

sbsmith86:
## Fixes #4403

The persistent nav and the regular cta have different markup so I needed to remove scoping specific to the `cta__message` element and only target the state classes.

@DoSomething/front-end


DFurnes:
:+1:

PR #4412, opened 2015-04-23
Update twitter meta tags to use photo card

### LINKED ISSUE #4404: Change permalink Twitter share to be the Photo Card type
This doesn't need to change the size of the image, it's just more performant to have them served through imagecache.

sbsmith86:
## Fixes #4404

Updates the twitter meta tags on the permalink page to use the photo card instead of the summary card. :8ball:


angaither:
:+1:

PR #4416, opened 2015-04-23
Mobile signup count

### LINKED ISSUE #4394: signup count getting overwritten
For campaigns with SMS signups, we need to be able to manually add in those SMS signups for now.

When I first put the "real" signup number in, it displays properly. But then web signups should add to this higher number as they come in (updated intermittently). However this number is just getting overwritten with the web signup updates.

i.e. on https://beta.dosomething.org/node/362/custom-settings I changed it from 9302 to 45321 and then later it went to 9304.

angaither:
Created two signup count vars per node
`mobile_signup_count` is user input, can be updated
`web_signup_count` is automatically generated, and updated on a cron 1/hour

The two are added together to display on the optimizly test on the pitch page. 
Fixes #4394 

@sbsmith86 @mikefantini


sbsmith86:
woot! :+1:

sergiitk:
:+1:

PR #4418, opened 2015-04-23
Refactors and improves Services member create callback.

### LINKED ISSUE #4413: User create API: set default country
When creating a user using Services API, set the country in the address field.

sergiitk:
#### What's this PR do?
- Refactors `_member_resource_create()` to use universal `dosomething_user_set_fields()` utility instead of custom cycle through fields
- Automatically detects and sets users country based on affiliate country code
- Fixes PHP notices produced by optional fields

Fixes #4413

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
@@ -164,21 +164,13 @@ function _member_resource_create($account) {
 
   // List of available properties to save:
   $fields = array(
-    'birthdate'  => $account['birthdate'],
+    'birthdate' => $account['birthdate'],
     'first_name' => $account['first_name'],
-    'country'    => dosomething_settings_get_affiliate_country_code(),
+    'country' => dosomething_settings_get_affiliate_country_code(),
+    'last_name' => !empty($account['last_name']) ? $account['last_name'] : NULL,
+    'mobile' => !empty($account['mobile']) ? $account['mobile'] : NULL,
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_api/resources/member_resource.inc
```diff
@@ -161,29 +161,27 @@ function _member_resource_create($account) {
...
+
+  // Optional fields:
+  if (!empty($account['last_name'])) {
+    $fields['last_name'] = $account['last_name'];
   }
+  if (!empty($account['mobile'])) {
+    $fields['mobile'] = $account['mobile'];
+  }
+  if (!empty($account['user_registration_source'])) {
+    $fields['user_registration_source'] = $account['user_registration_source'];
```
angaither: might be cleaner to set these in the array above

```
'last_name' =>  $account['last_name'] ?: NULL
'mobile' => $account['mobile'] ?: NULL
```

sergiitk: It still would produce the notice. The actual magic is in `empty()` statement, that checks presence of the array element without generating a notice. What is possible, is to wrap the `if` statements into one string, like this:

``` php
'last_name' => !empty($account['last_name']) ? account['last_name'] : NULL,
```

But I don't see this much cleaner than the original solution.

angaither: it just seems less broken up when it's all one array

angaither:
:+1:

PR #4419, opened 2015-04-23
Export Modal API for displaying campaign signup form.

### LINKED ISSUE #4415: prompt user after sign up for address collection is not prompting user right after sign up
see staging.beta.dosomething.org/campaigns/thumb-wars

@DFurnes would this be for you?

we're launching 1 in 3 of us May 1st and would need this

DFurnes:
# Changes
- Fixes #4415. After moving from RequireJS to Webpack for script bundling, we can't require modules from the browser, so we need to export `window.DSModal` to access that API from a script tag.


weerd:
:+1:

PR #4422, opened 2015-04-24
update twitter image tag

sbsmith86:
Updating `twitter:image` metatag to `twitter:image:src` because twitter has bad documentation. 

@blisteringherb @angaither


angaither:
:+1:

PR #4424, opened 2015-04-24
Upgrades mocha to 1.14.0.

sergiitk:
Fixes http://admin.dosomething.org/jenkins/job/Staging-Tests/307/consoleFull:

```
TypeError: 'undefined' is not a function (evaluating 'mocha.useColors(config.useColors)')
```


DFurnes:
:+1:

PR #4425, opened 2015-04-24
Expose validation object for unit tests.

DFurnes:
# Changes
- Fixes error where `Validation` object was not available from scripts.

For review: @sergii-tkachenko


sergiitk:
:+1:

PR #4432, opened 2015-04-28
Kudos resource

### LINKED ISSUE #4431: Add Kudos API resource for full CRUD capabilities
To enable the full use of all HTTP verbs when creating, retrieving and deleting Kudos records for Reportback Items, need to create a unique resource for Kudos instead of using the _targeted actions_ option on the Reportback Items resource.

weerd:
### Fixes #4431

This PR adds a new **Kudos** module and also creates a new **Kudos API Resource**. The module sets up a `dosomething_kudos` database table which will contain all kudos records that are attached to specific reportback items.

There is more work to be done, but this PR sets up a lot of the groundwork. The `/api/v1/kudos` basic endpoints work and can successfully create a kudos record (or multiple records at once) and delete individual records. 

Just to note, the newly added abstract `Transformer` class residing in the `dosomething_api` module will set up universal transformer methods that the resource classes can extend. This abstract class will be fleshed out in following PRs.

@jonuy @angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
@@ -165,5 +165,5 @@ function _reportback_item_resource_kudos($fid, $user_id, $term_ids) {
   );
 
   // @TODO: temporarily returning back the header body data for testing.
-  return $parameters;
+  return dosomething_kudos_create($parameters);
 }
--- a/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
@@ -80,50 +80,6 @@ function _reportback_item_resource_definition() {
 
     ),
 
-    'targeted_actions' => array(
-
-      'kudos' => array(
-        'help' => 'Register a kudos action being applied to a reportback item.',
-        'file' => array(
-          'type' => 'inc',
-          'module' => 'dosomething_api',
-          'name' => 'resources/reportback_item_resource',
-        ),
-        'callback' => '_reportback_item_resource_kudos',
-        'args' => array(
-          array(
-            'name' => 'fid',
-            'description' => 'The reportback item file ID.',
-            'optional' => FALSE,
-            'type' => 'int',
-            'source' => array(
-              'path' => 0
-            ),
-          ),
-          array(
-            'name' => 'user_id',
-            'description' => 'The ID of the user executing a kudos.',
-            'optional' => FALSE,
-            'type' => 'int',
-            'source' => array(
-              'data' => 'user_id',
-            ),
-          ),
-          array(
-            'name' => 'term_ids',
-            'description' => 'An array of kudos term IDs.',
-            'optional' => FALSE,
-            'type' => 'array',
-            'source' => array(
-              'data' => 'term_ids',
-            ),
-          ),
-        ),
-        'access callback' => '_reportback_item_resource_access',
-      ),
-
-    ),
-
   );
 
   return $reportback_item_resource;
@@ -154,16 +110,4 @@ function _reportback_item_resource_index($campaigns, $status, $count, $random, $
 function _reportback_item_resource_retrieve($fid) {
   $reportbackItems = new ReportbackItem;
   return $reportbackItems->show($fid);
-}
-
-
-function _reportback_item_resource_kudos($fid, $user_id, $term_ids) {
-  $parameters = array(
-    'fid' => $fid,
-    'uid' => $user_id,
-    'tids' => $term_ids,
-  );
-
-  // @TODO: temporarily returning back the header body data for testing.
-  return dosomething_kudos_create($parameters);
 }
--- a/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.install
+++ b/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.install
@@ -1,0 +1,53 @@
+<?php
+/**
+ * @file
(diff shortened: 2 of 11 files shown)


lib/modules/dosomething/dosomething_kudos/dosomething_kudos.install
```diff
@@ -0,0 +1,53 @@
...
+ */
+function dosomething_kudos_schema() {
+  $schema = array();
+  $schema['dosomething_kudos'] = array(
+    'description' => 'Table of kudos interactions on reportback items.',
+    'fields' => array(
+      'kid' => array(
+        'description' => 'The primary identifier for a kudos record.',
+        'type' => 'serial',
+        'unsigned' => TRUE,
```
angaither: These should all be 
`unsigned => FALSE`
which is be the default already, so you can remove all the references to `unsigned`

angaither:
The module should also be added to `dosomething.info`  and the `dosomething.make` files

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.install
weerd: Oop! I got it backwards :disappointed_relieved:

aaronschachter:
@weerd Did you check out RESTful at all? It provides a lot of these CRUD endpoints out of the box...

lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
```diff
@@ -165,5 +165,5 @@ function _reportback_item_resource_kudos($fid, $user_id, $term_ids) {
   );
 
   // @TODO: temporarily returning back the header body data for testing.
```
aaronschachter: Is this comment still valid?

weerd: Ah nope. That whole targeted action is getting removed out of there the reportback items endpoint. 
I was planning on cleaning that up on next PR, but will update the TODO note to reflect the change. That endpoint still works, but will be removed. This was done prior to realizing I wouldn't be able to use a targeted action to complete all functionality for kudos.

aaronschachter:
Ooh, one thing that's missing. You can automatically enable the Services endpoints by exporting them into the `dosomething_api` via Features.  If they are enabled in your local DB, run `drush fu dosomething_api` to update the feature -- you'll see the new endpoints added to the file.

Once all is said and done, pull down the staging db and run a `ds build` to do a full test of enabling the module and the endpoints.

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.install
```diff
@@ -0,0 +1,49 @@
...
+        'type' => 'int',
+        'not null' => TRUE,
+        'default' => 0,
+      ),
+    ),
+    'primary key' => array('kid'),
+    'indexes' => array(
+      'fid' => array('fid'),
+      'uid' => array('uid'),
+      'tid' => array('tid'),
```
aaronschachter, line 43: Just a sanity check: the `tid` index is intentional? For querying  all kudos records for a certain `tid`?

weerd:
Oh yeah. I was gonna do the Features stuff in a separate PR once I had the Kudos stuff at full functionality, but I can do that in this PR if it makes more sense?

aaronschachter:
Oh, ok. It's think it's fine to keep it in a separate PR -- didn't know it was on the way.

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.install
weerd, line 43: Yep yep :) They can be searched for by taxonomy id.

weerd:
@aaronschachter went ahead and just removed that unused targeted action for kudos in reportback-items resource :)

weerd:
@angaither @aaronschachter also made the updates to the `dosomething.info` and `dosomething.make` files. Check those out to make sure they look correct :grimacing:

aaronschachter:
Looks good to me, did you test with a clean stage db and `ds build` ?

weerd:
Oh, I didn't do that since I didn't set up the Feature in this PR. Should I do that test anyhoo and then turn the stuff on manually?

aaronschachter:
Yeah, it's worth sanity testing that the module is enabled and the table shows up.

weerd:
The module doesn't seem to be turned on during the rebuild but it doesn't break the build :)

lib/profiles/dosomething/dosomething.info
```diff
@@ -80,6 +80,7 @@ dependencies[] = dosomething_campaign_run
 dependencies[] = dosomething_fact
 dependencies[] = dosomething_fact_page
 dependencies[] = dosomething_image
+dependencies[] = dosomething_kudos
```
aaronschachter, line 83: This is the line that should be enabling it by default..

aaronschachter:
That's weird, @weerd. If you enable the module manually, does it work properly? e.g. Table appears, etc

weerd:
Yep. It's all good if done manually. Works fine; can create kudos, retrieve them, etc.

aaronschachter:
Hmm, then something is up.. .either with this module or possibly with the `ds build` function that checks to auto-enable modules... 
It's not the end of the world if we have to enable manually but it's def a new issue..

aaronschachter:
I ran `ds build` locally with these changes and it enabled ok. Did it work on staging? @weerd

weerd:
Haven't deployed to stage yet, but I can now to try it :)

weerd:
Looks good on Staging; was properly enabled. Might just be a "my local is slightly borked" kinda thing. :grimacing:

PR #4434, opened 2015-04-28
Send users to northstar

### LINKED ISSUE #4276: User login events > northstar

angaither:
When a user logs into dosomething, send that user into Northstar.
- Locally in vagrant to test you have to add `10.0.2.2 northstar.dev`  to `/etc/hosts`
  - there's a PR DoSomething/ansible-vagrant/pull/3 to do that
- Northstar module needs to be enabled

Fixes #4276

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -80,41 +80,41 @@ function dosomething_northstar_build_ns_user($user, $form_state) {
     'addr_zip'      => $user->field_address[LANGUAGE_NONE][0]['postal_code'],
   );
 }
+//@TODO: use this once we can use guzzle.
+// class Northstar {
 
-class Northstar {
+//   public $client;
 
-  public $client;
+//   public function __construct()
+//   {
 
-  public function __construct()
-  {
+//     $this->client = $library;
+//     $base_url = NORTHSTAR_URL;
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -28,14 +34,114 @@ function dosomething_northstar_menu() {
...
+      'X-DS-REST-API-Key' => NORTHSTAR_APP_KEY,
+      'Content-Type' => 'application/json',
+      'Accept' => 'application/json',
+      ),
+    );
+
+  return $client;
+}
+//@TODO: use this once we can use guzzle.
+// class Northstar {
```
blisteringherb, line 114: Did you want to include this commented out code or just add it after guzzle is avail?

blisteringherb:
You mentioned the Northstar module needs to be enabled, should it be included in the makefile?

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
angaither, line 114: Yeah, I mean might as well leave it in, as it the DREAM code we can use when guzzle is available.

angaither:
@blisteringherb I don't want this enabled on stage, since we don't have a corresponding QA for northstar
so I left it out of the makefile

blisteringherb:
:+1:

PR #4438, opened 2015-04-30
Update signup cron count query

### LINKED ISSUE #4437: web signups not automatically updating for # of ppl doing display
# of ppl doing display should sum:
- Mobile Count (manually input from admin)
- Web signups (automatically updated as new signups come in)

This summation is working properly, however the web signups number is not updating.

Custom settings tab for node 362 on prod:
image

Signups tab for node 362 on prod:
image

What the optimizely test would display on the pitch page:
image

This should be showing 14,753

related to #4394

angaither:
Since we aren't overriding the web signup count, we no longer need to do a time based query to look up and override the old var
Fixes #4437

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/includes/dosomething_signup.cron.inc
+++ b/lib/modules/dosomething/dosomething_signup/includes/dosomething_signup.cron.inc
@@ -13,14 +13,11 @@ function dosomething_signup_cron() {
   // Get all reportbacks updated in the past hour.
   $results = db_query("SELECT nid, count(nid) as total_signups
                       FROM dosomething_signup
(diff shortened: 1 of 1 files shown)


blisteringherb:
:+1:

lib/modules/dosomething/dosomething_signup/includes/dosomething_signup.cron.inc
```diff
@@ -13,14 +13,11 @@ function dosomething_signup_cron() {
   // Get all reportbacks updated in the past hour.
```
aaronschachter, line 13: Update / remove comment?

PR #4442, opened 2015-05-04
Issue4105 reportback image processor

### LINKED ISSUE #4105: Generate themed Reportback images upon Reportback Form submit
Both the web Reviewer forms and the Reviewer app chug along when Drupal is processing a high-res Reportback image for the first time.  Ideal fix is to asychronously hit the URL's of the processed images upon save, to speed up load time for reviewers.

[Stack Overflow suggests](http://drupal.stackexchange.com/questions/110449/non-blocking-drupal-http-request) that `drupal_http_request` cannot make asynchronous requests and PHP generally doesn't hang that way. Adding custom JS to ping the URL or creating a Message Broker task are both potential solutions that come to mind...

This will mostly affect staff and volunteers, but will be user facing at times when viewing the gallery.

deezone:
### Fixes #4105
- Addition of `imageProcessingQueue` queue to support collection of report backs and their image uploads. The entries will be used to trigger the consumer mbc-image-processor.
- Removes `userCampaignActivityQueue`, cruft, not used.

### DIFF
--- a/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
+++ b/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
@@ -121,6 +121,17 @@ function dosomething_mbp_update_7006(&$sandbox) {
 }
 
 /**
+ * Updates user_register production to:
+ * - include imageProcessingQueue
+ * - remove userCampaignActivityQueue
+ * from transactional_campaign_reportback production.
+ */
+function dosomething_mbp_update_7007(&$sandbox) {
+  db_drop_table('message_broker_producer_productions');
+  dosomething_mbp_install_productions();
+}
+
+/**
  * Common function for install and update of related DoSomething specific
  * message_broker_producer produciton settings.
  */
--- a/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
+++ b/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
@@ -188,8 +188,8 @@ function dosomething_mbp_install_productions() {
   $productions[4]['queues'][] = 'transactionalQueue';
   $productions[4]['queues'][] = 'loggingQueue';
   $productions[4]['queues'][] = 'activityStatsQueue';
-  $productions[4]['queues'][] = 'userCampaignActivityQueue';
   $productions[4]['queues'][] = 'userAPICampaignActivityQueue';
+  $productions[4]['queues'][] = 'imageProcessingQueue';
   $productions[4]['routing_key'] = 'campaign.report_back.transactional';
   $productions[4]['status'] = 1;
 
--- a/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
+++ b/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
@@ -125,8 +125,12 @@ function dosomething_mbp_update_7006(&$sandbox) {
  * - include imageProcessingQueue
  * - remove userCampaignActivityQueue
  * from transactional_campaign_reportback production.
+ *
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
```diff
@@ -188,8 +199,8 @@ function dosomething_mbp_install_productions() {
   $productions[4]['queues'][] = 'transactionalQueue';
   $productions[4]['queues'][] = 'loggingQueue';
   $productions[4]['queues'][] = 'activityStatsQueue';
-  $productions[4]['queues'][] = 'userCampaignActivityQueue';
```
angaither, line 191: I think we still want this `userCampaignActivityQueue` right? 
that's the one that supports the mongo collection that stores the user activity stream

deezone, line 191: Nope.

The `loggingQueue`: https://github.com/DoSomething/mbc-logging/blob/master/app.js#L30 supports the activity stream for logging and `activityStatsQueue` feeds the dashboard via `mbp-externalApplications-dashboard`: https://github.com/DoSomething/mbp-externalApplications-dashboard. 

I created `userCampaignActivityQueue` quite awhile ago for a different purpose that's no longer needed. This is an effort to clean things up a bit while adding other, unrelated functionality for image processing.

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
```diff
@@ -121,6 +121,17 @@ function dosomething_mbp_update_7006(&$sandbox) {
...
 
 /**
+ * Updates user_register production to:
+ * - include imageProcessingQueue
+ * - remove userCampaignActivityQueue
+ * from transactional_campaign_reportback production.
+ */
+function dosomething_mbp_update_7007(&$sandbox) {
+  db_drop_table('message_broker_producer_productions');
+  dosomething_mbp_install_productions();
```
aaronschachter, line 135: Dropping the table seems a little... dangerous. Is there a chance this will mess things up in prod when this script is running, and the table doesn't exist?

An alternative approach could be to just delete the record and insert one (instead of killing the entire table)

deezone, line 135: This is how it's been done all the way back to `function dosomething_mbp_update_7004(&$sandbox) {` based on @sergii-tkachenko initial commit. The goal is to reset the table values regardless of user interaction from the last update to the settings in code.

Is it worth adding in a `db_table_exists()`?

aaronschachter, line 135: If it's worked for other deploys, then no biggie.  I just meant using `db_delete` and `db_insert` instead of completely dropping the table.

While we're in here, might as well mention I got these errors upon running a `ds build --install`:

```
Missing argument 1 for dosomething_mbp_update_7006(), called in      [warning]
/var/www/dev.dosomething.org/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
on line 29 and defined in dosomething_mbp_update_7006() (line 113 of
/var/www/dev.dosomething.org/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install).
```

Probably not critical if everything's okl, but, seems somewhat relevant.  Would be cleaner to not get those errors upon install.

deezone:
@aaronschachter Removed &$sandbox as a param to hook_update_N() to get past the PHP warning. What do you think?

aaronschachter:
If tested and all is good, shippit :+1:

deezone:
Confirmed, updb with a copy of the staging site rebuilds the transactional_campaign_reportback Message Broker production as expected without errors/warnings.

PR #4457, opened 2015-05-05
Allow for null reportback status.

### LINKED ISSUE #4454: Only count non-flagged reportbacks
Totally forgot to add this `AND` to the [query](https://github.com/DoSomething/dosomething/blob/b7b36a5c6364b71feab18d88800af88b50f61480/lib/modules/dosomething/dosomething_signup/includes/dosomething_signup.cron.inc)
Currently counting all reportbacks, approved or not :grimacing:

### LINKED ISSUE #4456: Update flagged/non-flagged reportbacks
Do a wee bit of surgery on the db to 
1. Change the default `flagged` in the reportback table from `0->null`
2. Update all non-reviewed reportbacks from `0->null`

Refs #4454

angaither:
Update current records to match new schema.

Fixes #4454 
Fixes #4456


aaronschachter:
:+1:

PR #4458, opened 2015-05-05
Update user permalink name to say your friend.

### LINKED ISSUE #4455: non-owner permalink title if user first name does not exist
If there is no first name attached to the user who submitted a reportback let's have the top of the non-owner permalink page read "Join DoSomething.org" instead of "Join [user:first_name] and DoSomething.org"

image

This is a consideration for providing the permalink to users via SMS after they have submitting a reportback via SMS.

cc: @angaither @jonuy @tongxiang 

This work should get done before @tongxiang you push the feature live.

angaither:
fixes #4455 if we want to go with this.


mikefantini:
yeah let's just go with this @angaither

aaronschachter:
:+1:  May want to make a variable if we have to change again, but looks like great PHP

Issue #1488, opened 2014-03-24
ds script doesn't check .bash_profile before appending aliases


mshmsh5000:
This should check that this string doesn't already exist in `.bash_profile`:

`echo 'alias codercs="phpcs --standard=$WEB_PATH/profiles/dosomething/modules/contrib/coder/coder_sniffer/Drupal/ruleset.xml --extensions=php,module,inc,install,test,profile,theme"' >> ~/.bash_profile`

![codercs](https://f.cloud.github.com/assets/149811/2500762/c2fe9352-b365-11e3-99e8-16cca7afeb1b.png)

PR #4459, opened 2015-05-06
Removes `codercs` alias.

sergiitk:
It'd be better to setup this with ansible, if needed. Current version adds this alias to `.bashrc` on each build.
Ref #1488.


aaronschachter:
Ok. :+1: 

We haven't done much Coder testing for a long time :disappointed:

PR #4463, opened 2015-05-07
Update views module

aaronschachter:


sergiitk:
I'd suggest to run `drush features-update-all` and looks for changed `*.views_default.inc` files to make sure this update doesn't bring structural changes to the view export format.

aaronschachter:
Very good call! @sergii-tkachenko! I forgot to check that. I've run it and there are no changes to any of the Views files.

sergiitk:
Awesome! :+1:

PR #4464, opened 2015-05-07
Updates to cleanup and separate concerns for the Kudos API endpoint functionality.

weerd:


lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -8,25 +8,14 @@ include_once 'dosomething_kudos.kudos.inc';
...
-
-  foreach($parameters['tids'] as $id) {
-    $values['tid'] = $id;
-
-    $records[] = db_insert('dosomething_kudos')->fields($values)->execute();
-  }
-
-  return $records;
+function dosomething_kudos_create($values) {
+  return db_insert('dosomething_kudos')->fields($values)->execute();
```
aaronschachter, line 18: Discussed this w @weerd. We're going to go with the approach of creating a Kudos Entity to handle these CRUD functions, to keep consistent with how entities like Signups, Reportbacks, Reportback Files, Shipments, Rewards, etc work.

Ideally the `Kudos` class created can be used for the definition in `hook_entity_info`. Consistent naming convention would call the class `KudosEntity` ... but, we can worry about that once you get there...

PR #4466, opened 2015-05-07
Updated reportback quanity to work same as signups

### LINKED ISSUE #4453: Reportback count override
With #4396 we lost the ability to override reportback counts
Because of campaigns with 3rd party sponsors, we need this, for now, going to build it like #4320 where the helper text displays the total count in the database, but you can add extra values and the two will be added together for the sum

angaither:
now it shows db total in the desc, and will be added to whatever is in the field
Fixes #4453

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -354,8 +354,8 @@ function dosomething_campaign_preprocess_pitch_page(&$vars, &$wrapper) {
   // Moar $tagline.
   $vars['tagline'] .= t('Any cause, anytime, anywhere.');
 
-  // Get sign up total and add it to cta
   $campaign_progress = (int) dosomething_helpers_get_variable('node', $vars['nid'], 'sum_rb_quantity_override') + (int) dosomething_helpers_get_variable('node', $vars['nid'], 'sum_rb_quanity');
+  // Get sign up total and add it to cta
   $signup_count = (int) dosomething_helpers_get_variable('node', $vars['nid'], 'web_signup_count') + (int) dosomething_helpers_get_variable('node', $vars['nid'], 'mobile_signup_count');
   $vars['signup_cta'] .= t('Join @signup_count people doing this.', array(
       '@signup_count' => number_format($signup_count, 0, '', ','),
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -355,7 +355,7 @@ function dosomething_campaign_preprocess_pitch_page(&$vars, &$wrapper) {
   $vars['tagline'] .= t('Any cause, anytime, anywhere.');
 
   // Get sign up total and add it to cta
```
aaronschachter, line 358: Picky picky, but can we change this to help make the code more readable? This is not signup total.  Might be nice to update the comment with a description of why this is happening.

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.variable.inc
```diff
@@ -149,23 +149,14 @@ function dosomething_helpers_variable_form($form, &$form_state, $node) {
...
-    // counts submitted directly to the third party, but haven't been
-    // reported back on our site
-    if ($vars['progress'] > $rb_progress) {
-      dosomething_helpers_set_variable('node', $node->nid, 'sum_rb_quanity', $vars['progress']);
-    }
     // Get the progress from the set variable, else do a query for it.
     $rb_progress = dosomething_helpers_get_variable('node', $node->nid, 'sum_rb_quanity') ?: dosomething_reportback_get_reportback_total_by_nid($node->nid);
 
-    $form['goals']['sum_rb_quanity'] = array(
+    $form['goals']['sum_rb_quanity_override'] = array(
```
aaronschachter: This is picky picky too but the OCD in me cannot not point out this spelling mistake. Not sure how much of a lift it'd be to rename...

angaither: oh yeah, it's a bit late... for the original one

angaither: unless I do an update hook for all of them

aaronschachter: My 0.02 is that it would be worth it. Doesn't need to be in this PR.. but I don't think we should get into the habit of letting spelling mistakes slide, it makes the code less maintainable.

angaither: cool made an issue for it #4468

Issue #4445, opened 2015-05-04
Goal animation logic 


sbsmith86:
Add logic on the backend that delivers one animation file to the campaign template based on goal progress. 

If current progress is: 
- Between 0 and 20% of goal, use animation 1
- Between 21% and 40% of goal, use animation 2
- Between 41% and 60% of goal, use animation 3
- Between 61% and 80% of goal, use animation 4
- Between 81% and 99% of goal, use animation 5
- Goal is met, use animation 6

mikefantini:
@sbsmith86 looks good to me!

angaither:
no longer doing this with animations

PR #4467, opened 2015-05-07
[hold off on reviewing] Goal animation logic

### LINKED ISSUE #4461: Set animation file status to 1 when saved
When animation files are uploaded and saved on the campaign configuration page, they have a status of `0` in the database so they will be removed when the next cron job runs. So we need to work out a way to set the status to 1. 

/cc @aaronschachter @angaither

sbsmith86:
## Fixes #4461 and #4445
- Changes goal animation config fields to be text fields that take in FID values as discussed in #4461 
- Creates two helper functions - one that gets the current range of progress, and one that gets the animation file that corresponds with that progress range.
- Creates a variable that the template can use to print the correct animation file. 

@angaither @aaronschachter


angaither:
@sbsmith86 can we close this one out now?

sbsmith86:
@angaither  yea

PR #4469, opened 2015-05-07
Campaigns query

### LINKED ISSUE #4110: Refactor Campaigns API endpoint

aaronschachter:
Refactors existing `dosomething_campaign_get_campaigns` method to `dosomething_campaign_get_campaigns_query_result($params)`, similar to the `dosomething_reportbck_get_reportbacks_query_result`.

Possible `$params` keys:
- `is_staff_pick` (int)
- `on_mobile_app` (int)
- `term_ids`: array of term id's.  Can be any field, queries the `taxonomy_index` table to lookup whether the nid is associated with the given term.

Updates existing GET `/campaigns` endpoint accordingly:

```
http://dev.dosomething.org:8888/api/v1/campaigns.json?is_staff_pick=1
http://dev.dosomething.org:8888/api/v1/campaigns.json?on_mobile_app=1
http://dev.dosomething.org:8888/api/v1/campaigns.json?term_ids=4,5
```

Closes #4110

cc @deadlybutter this will break the lobby dashboard query once pushed to prod...

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -12,12 +12,26 @@ function _campaign_resource_definition() {
         'access arguments' => array('index'),
         'args' => array(
           array(
-            'name' => 'parameters',
+            'name' => 'is_staff_pick',
             'optional' => TRUE,
-            'type' => 'array',
-            'description' => 'Parameters',
-            'default value' => array(),
-            'source' => array('param' => 'parameters'),
+            'type' => 'int',
+            'source' => array('param' => 'is_staff_pick'),
+            'default value' => NULL,
+          ),
+          array(
+            'name' => 'on_mobile_app',
+            'optional' => TRUE,
+            'type' => 'int',
+            'source' => array('param' => 'on_mobile_app'),
+            'default value' => NULL,
+          ),
+          array(
+            'name' => 'term_ids',
+            'description' => 'Taxonomy term ids to filter by.',
+            'optional' => TRUE,
+            'type' => 'string',
+            'source' => array('param' => 'term_ids'),
+            'default value' => NULL,
           ),
         ),
       ),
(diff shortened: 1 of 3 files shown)


aaronschachter:
@jonuy This approach will allow us to query by any taxonomy term, so the query string should not change whether or not we query on `field_tags` or a new Interest Group taxonomy field.

jonuy:
@aaronschachter Great! Is it safe to assume that these params can also be combined? Like...

```
http://dev.dosomething.org:8888/api/v1/campaigns.json?on_mobile_app=1&term_ids=4,5
```

itsjoekent:
roger dodger

thanks for the heads up @aaronschachter

aaronschachter:
@jonuy Yea.. safe to assume although i'm finding the `term_ids` filter isn't working as expected. Looking into it.

aaronschachter:
Fixed. @weerd Mind reviewing?

weerd:
Checking now :)

weerd:
I might vote for:

```
http://dev.dosomething.org:8888/api/v1/campaigns.json?is_staff_pick=true
http://dev.dosomething.org:8888/api/v1/campaigns.json?on_mobile_app=true
```

Instead of using `1` to denote what would be meant as true. Just feels more legible. We're already using `?random=true` in reportback stuff ;) I don't think it'll mean a code change, but it's nice that the data type would match what the intent is.

weerd:
Actually, maybe slight code change in the resource to designate a boolean. Plus it's probs better since the resource w/ it being indicated as an `int` could mean any number technically... but a boolean is a boolean ;)

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -12,12 +12,26 @@ function _campaign_resource_definition() {
...
         'args' => array(
           array(
-            'name' => 'parameters',
+            'name' => 'is_staff_pick',
             'optional' => TRUE,
-            'type' => 'array',
-            'description' => 'Parameters',
-            'default value' => array(),
-            'source' => array('param' => 'parameters'),
+            'type' => 'int',
```
weerd: See comment in convo. Suggesting turning this to `boolean`

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -153,36 +167,19 @@ function _campaign_resource_access($op = 'view', $args = array()) {
...
-  $index = array();
-
-  // Get the results of the explore_campaigns view.
-  $view = views_get_view('explore_campaigns');
-  $view->execute();
-
-  $filter_staff_pick = FALSE;
-  if (isset($parameters['is_staff_pick']) && $parameters['is_staff_pick']) {
-    $filter_staff_pick = TRUE;
+function _campaign_resource_index($is_staff_pick = NULL, $on_mobile_app = NULL, $term_ids = NULL) {
```
weerd, line 170: If turning `is_staff_pick` & `on_mobile_app` to a booleans, then switch these to default to `false` :)

weerd:
Looks good overall, just the change regarding the `int`s to `boolean`s!

aaronschachter:
can do

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
aaronschachter, line 170: The default should be no value at all though, returning all campaigns regardless of whether or not they are staff picks, on mobile app, etc.

aaronschachter, line 170: The params are optional.

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
aaronschachter: _cough_ https://github.com/DoSomething/dosomething/blob/v0.4.104/lib/modules/dosomething/dosomething_api/resources/reportback_resource.inc#L65

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
weerd, line 170: Gotcha.

Issue #4444, opened 2015-05-04
Time left in campaign


angaither:
we want to show the amount of time left in the hot phase.
For now, we will be using the high season end date

@mikefantini can you add the specific rules here of how the date/time left is displayed

mikefantini:
@angaither:
- If 8+ days remain, deadline will be displayed as a date (i.e. Ends April 30)
- If 3-7 days remain, deadline will be displayed as days remaining (i.e. 5 days left or ends in 5 days)
- If 1-2 days remain, deadline will be displayed as hours remaining (i.e. 24 hours left or ends in 24 hours)

@angaither also ideally the end date is set so that # of hours counts down from 23:59:59 EST on the end date so the # of hours remaining is the same for all users. I know we have seen a problem with some users who have a different timezone set in the past, but most if not all users should have Timezone set to UTC which I _think_ would resolve the issue? Not sure though.

![image](https://cloud.githubusercontent.com/assets/6330971/8488301/a859858c-20df-11e5-8ce0-b4a8e437be3b.png)

itsjoekent:
@angaither am I doing the bender for this issue and then handing it off to @sbsmith86 or am I doing both?

sbsmith86:
@deadlybutter I don't think you need to reassign to me. Once the bender part is done, the only fender bit is to print the variable in the template here: https://github.com/DoSomething/dosomething/blob/d7807f18823383c30b3f6b4f277e7ba75b6f6270/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php#L60

itsjoekent:
ah gotcha, thanks!

PR #4470, opened 2015-05-08
Added time_left as var on pitch page theme.

angaither:
Start of work for #4444

This will still need to be displayed in many different date formats, however we are holding on that for now.


aaronschachter:
Ok :+1: 

Should it be `$vars['time_left']` or not yet?

angaither:
@aaronschachter eh, not yet, we have to build the whole which way should I show the date.

PR #4474, opened 2015-05-11
Updates to awesome-ize Reportbacks API structure.

### LINKED ISSUE #4473: Create universal Transformer Class for API on DS app
Will be creating a Transformer class within the API module that can be extended by other module classes when creating specific transformers for those modules (ie: ReportbackTransformer, CampaignTransformer, KudosTransformer, etc.)

weerd:
### Fixes #4473

After some discussion with @aaronschachter we're going to move towards a more standard class based structure for the API work. There is now a universal API Transformer _abstract_ class within the **dosomething_api** module that can be extended by other transformer classes in other modules. The Transformer abstract class contains a variety of methods that the module transformer classes can harness. 

Module specific transformer classes should follow this directory structure and naming scheme: 

```
module_name/includes/ModuleTransformer.php
```

The Reportbacks API endpoints have been updated to follow this new structure which should lay the groundwork for more API work moving forward using Services module.

@aaronschachter 

CC: @angaither @jonuy @DFurnes @drewish


lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -83,39 +99,19 @@ protected function getPathParameters($parameters, $endpoint) {
 
 
   /**
-   * Calculate total number of Reportback items for the specified campaigns by
-   * each status requested.
-   *
-   * @param array $parameters Query parameters.
+   * @param $ids
```
aaronschachter: These functions should include documentation, esp around the datatypes of parameters, and their use.. https://www.drupal.org/coding-standards/docs#functions

lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -83,39 +99,19 @@ protected function getPathParameters($parameters, $endpoint) {
...
+  protected function getReportbackItems($ids) {
+    $filters = array(
+      'fid' => $this->formatData($ids),
+    );
 
-    if (count($array) > 1) {
-      return $array;
-    }
+    // Obtaining all Reportback items.
+    $query = dosomething_reportback_get_reportback_files_query_result($filters, 'all');
```
aaronschachter, line 134: Is this reportback supposed to be here? Isn't this class the standard Transformer class?

lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -127,10 +123,10 @@ protected function formatData($data) {
    *
    * @return array
    */
-  protected function paginate($parameters, $endpoint) {
+  protected function paginate($total, $parameters, $endpoint) {
```
aaronschachter, line 156: Same here, needs documentation.

lib/modules/dosomething/dosomething_reportback/includes/ReportbackTransformer.php
```diff
@@ -105,20 +114,4 @@ protected function transform($reportback) {
...
-    $filters = array(
-      'fid' => $this->formatData($ids),
-    );
-
-    // Obtaining all Reportback items.
-    $query = dosomething_reportback_get_reportback_files_query_result($filters, 'all');
-
-    return services_resource_build_index_list($query, 'reportback-items', 'fid');
-  }
 }
```
aaronschachter: I've seeing these a lot, worth fixing. Github is complaining - best practice is to leave a blank line at the end of a file.

angaither: we should maybe add an `.editorconfig` to the repo so we don't have to worry about these

lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -170,11 +166,12 @@ protected function setOffset($page, $count) {
 
   /**
    * @param array $items Collection of item objects retrieved data.
+   * @param string $method Name of method to use on each array item.
    *
    * @return array
    */
-  protected function transformCollection($items) {
-    return array_map(array($this, 'transform'), $items);
+  protected function transformCollection($items, $method = 'transform') {
```
aaronschachter, line 208: Super minor, but calling the variable `method` made me think we're passing an actual function name here. What do you think about renaming it as `$op` ? It's commonly found in the [Drupal universe](https://api.drupal.org/api/drupal/modules%21node%21node.api.php/function/hook_node_access/7).

lib/modules/dosomething/dosomething_api/includes/Transformer.php
weerd, line 134: So yeah, I'm open to suggestions here. While yes this is the standard Transformer class, reportbacks are such a ubiquitous part of the DS experience that I foresee other module transformers that extend the main transformer needing access to something like this... Campaigns for example would probably need a method of retrieving reportback items data, and ideally not repeat the code in multiple places.

Thoughts?

lib/modules/dosomething/dosomething_reportback/includes/ReportbackTransformer.php
DFurnes: +9999 `.editorconfig`! :)

lib/modules/dosomething/dosomething_api/includes/Transformer.php
weerd, line 208: But we are actually passing a function/method name there ;)

lib/modules/dosomething/dosomething_api/includes/Transformer.php
aaronschachter, line 134: Hmm, none come to mind right, because yeah I agree: I can see Campaigns, Terms, Users, all needing reportback that are related to themselves.  It just feels a little funky being in the general Transformer. I think fine to keep here for now, but something to passively think about and perhaps never endi up changing.

aaronschachter:
Code looks ok to me, just needs documentation per comments. :+1:

lib/modules/dosomething/dosomething_api/includes/Transformer.php
weerd, line 134: hahahahah. Yeah, I hear you though and agree. We could also potentially keep it in the ReportbackTransformer and this Transformer somehow instantiates the ReportbackTransformer class to get access to those reportback functions, but not sure if that starts to get too hairy. Definitely open to a future refactor once we get some of these other transformers in play and see how they interact.

PR #4477, opened 2015-05-11
Replace html title with tokens.

### LINKED ISSUE #4439: site name won't read member count token
ideally using [site:member-count] in this instance would pull in the actual member count rounded to the hundred thousands digit (i.e. like it does on the pitch page footer message)

https://beta.dosomething.org/admin/config/system/site-information

image

image

angaither:
Fixes #4439 

@mikefantini I think on production/stage right now the setting is `['site:member-count']` and it will need to be changed to `[site:'member-count-readable']`


aaronschachter:
:+1:

PR #4486, opened 2015-05-12
Always be flaggin

### LINKED ISSUE #4485: Reportback flagging does not set flagged value to 0
introduced by #4457

after a reportback is reviewed, if approved, it's status is not set to 0.

angaither:
reportback files that were approved, were not getting the `flagged` value set to 0. This fixes that, as well as an update hook to fix the ones stuck in the `null` status 
Fixes #4485 

:flags:

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc
+++ b/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc
@@ -197,16 +197,18 @@ class ReportbackEntity extends Entity {
   /**
    * Flags the reportback with the given $reason string.
    *
+   * @param string $status
+   *   The status of the reprotback file (approved, flagged, excluded, promoted)
    * @param string $reason
    *   Text to store in the flagged_reason property.
    */
-  public function flag($reason = NULL) {
-    if (is_null($reason)) {
-      $this->flagged = 0;
-    }
-    else {
+  public function setFlagged($status = NULL, $reason = NULL) {
+    if ($status === 'flagged') {
       $this->flagged = 1;
       $this->flagged_reason = $reason;
+    }
+    else {
+      $this->flagged = 0;
     }
     return entity_save('reportback_file', $this);
   }
--- a/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc
+++ b/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc
@@ -200,9 +200,14 @@ class ReportbackEntity extends Entity {
    * @param string $reason
    *   Text to store in the flagged_reason property.
    */
-  public function flag($reason) {
-    $this->flagged = 1;
-    $this->flagged_reason = $reason;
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback_file.inc
```diff
@@ -58,17 +58,16 @@ class ReportbackFileEntity extends Entity {
       // Store that instead.
       $this->review_source = $values['source'];
     }
-    if ($this->status === 'flagged') {
```
aaronschachter, line 61: This conditional was added as a sanity check: If an app posts a Review with a non-`flagged` status but still sends a `flagged_reason` -- it's assuming it actually shouldn't be flagged.  Any reason to take it out?

angaither, line 61: Yeah, so the `flag` function is now called on every reportback file review, and sets the value to `0` or `1` 
since the default is now null

angaither, line 61: the `dosomething_reportback.flagged` was getting set to null, and only changing to `1` if it was flagged, when ideally if it wasn't flagged, it would update to 0

aaronschachter, line 61: With this logic, isn't it the case that you could never unflag a Reportback? Since we don't allow editing the Flagged Reason?

angaither, line 61: I didn't know that was a requirement.
is there a different place I should put this code?

aaronschachter, line 61: Maybe renaming the function to `setFlagged` instead of `flag`.. and moving that status conditional there? If the status is `flagged`, then it should be 1 else set it to 0 ?  Not sure without testing but I think that would still allow reviewers the ability to correct any mistakes..

angaither, line 61: @aaronschachter updated

lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc
```diff
@@ -197,12 +197,19 @@ class ReportbackEntity extends Entity {
   /**
    * Flags the reportback with the given $reason string.
    *
+   * @param sting $status
+   *   The status of the reprotback file (approved, flagged, excluded, promoted)
```
aaronschachter, line 201: sting

aaronschachter:
:+1:

lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc
angaither, line 201: ha exactly ;)

PR #4493, opened 2015-05-13
Problem share images

### LINKED ISSUE #4471: ImageMagick experiments for drawing problem fact images
Computers can do the work for us.
http://www.imagemagick.org/Usage/draw/#text

problemfact
### Size

471x248px
Background: #23B7FB  
### Fonts

Proxima Nova 24pt #FFFFFF
Proxima Nova Bold 16pt  #337D9F
### Margins

20px on all sides.

angaither:
This cron job will update all active campaigns and create an image for the problem statement of that campaign
Fixes #4471
cc @sbsmith86 

OH, this also requires imagemagick DoSomething/ansible-dosomething/issues/15


lib/modules/dosomething/dosomething_campaign/dosomething_campaign_problem_shares/create-images.sh
```diff
@@ -0,0 +1,34 @@
...
+  -pointsize 24 \
+  caption:"$2" \
+  $TEMP_FILE_2 \
+
+composite \
+  $TEMP_FILE_2 \
+  -gravity northwest \
+  -geometry +20+50 \
+  $TEMP_FILE_1 \
+  $FINAL_FILE #TODO - should be the name of the file with node id.
```
sbsmith86: oh i guess this TODO doesn't need to be here anymore. eh.

aaronschachter:
Has this been tested with a `ds build`? Doesn't the new module need to be added to the .info and make file to install correctly? 

Also, are we ok to add the font files into github open source land?

angaither:
@aaronschachter we can't add this to the install file without the imagemagick stuff in stage/production yet

cc @DFurnes @sbsmith86 re: font-files

sbsmith86:
@angaither @aaronschachter we have them in github in the neue repo, so following that precedent it should be fine. @DFurnes do you see any issue with this?

DFurnes:
I don't think so. The font files are fully accessible to download over HTTP (to render the page in the browser), so there shouldn't be any extra harm in having them on GitHub.

aaronschachter:
Code looks fine, I am unsure if the module will be found without being added into the make file.

angaither:
@aaronschachter yeah, that's fine, there's not anything we can do with this till we get imagemagick on those environments anyways

PR #4494, opened 2015-05-13
Campaign transformer and class

### LINKED ISSUE #4482: Campaign API endpoint -- index fields
The campaigns index probably shouldn't return a full view of every campaign... but it definitely needs to return:
- Nid / ID
- Title
- Call to Action
- Cover Image (format?)

weerd:
### Fixes #4482

Establishes a slightly new approach to how we extract Campaign node data, and further, how that node data is transformed when returning data for API requests.

There is now a new `Campaign` class in the **dosomething_campaign** module. It updates on what was done before with returning a campaign `stadClass` object, but allows it to have a slew of custom methods for extracting data and making code more maintainable. When a node id is passed to the class upon instantiation, it will load the node and return a customized Campaign class object ready for use in different contexts (one of which will be the API).

Instantiate using:

```
$campaign = new Campaign($id);
```

The `Campaign` class has numerous `protected` methods that it uses to get associated data for a campaign to flesh out the kind of data we would expect to get back for a campaign object. Currently this object only collects the necessary data that is going to be used for the campaign, but additional methods will/can be added to collect more information (such as action guide content, faqs and other main campaign page content, etc).

We're defining a modified structure for the returned campaign class object that follows a more organized hierarchical structure that also mimics the structure for response data from the API. This allows us to have to do less cleanup in the `transformCampaign()` method of the `Transformer` class. Along those lines, after much discussion with @DFurnes and @aaronschachter  we're removing the need for unique property names for ids (such as `nid`, or `tid`). Since there's a much more hierarchical structure to the object, the property `id` will refer to the parent. This saves time with the need to change every `nid`, `tid` in the transformer for the sake of the API and just makes life easier.

A couple new helper utility functions have also been added to the _dosomething_helpers_ module:
- `dosomething_helpers_extract_field_data()`
  - Extract Drupal field data, accounting for single or multiple values and language. Also return both `raw` and `formatted` values if available.
- `dosomething_helpers_issetor()` 
  - Checks if a value is set and if so returns it otherwise returns a default (which can be specified)
### Examples
- [Example output for the Campaign class object](https://gist.github.com/weerd/580ac04fa53b7ec21a06)
- [Example output for the Transformed Campaign API repsonse return](https://gist.github.com/weerd/c9641363235a45dd589f)

_NOTE_: I'm still updating DocBlocks across the methods.


lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -164,6 +187,25 @@ function _campaign_resource_access($op = 'view', $args = array()) {
   return services_error(t('Campaign node @nid is not active.', array('@nid' => $node->nid)), 403);
 }
 
+
+function _campaign_resource_alt_access() {
+  // @TODO: Temp universal access for now.
+  // Need to rethink how access happens within these callbacks
+  // for new approach.
```
aaronschachter, line 194: Def worth considering using `node_access`... https://api.drupal.org/api/drupal/modules%21node%21node.module/function/node_access/7

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
```diff
@@ -0,0 +1,285 @@
...
+  protected function getTagline() {
+    return dosomething_helpers_extract_field_data($this->node->field_call_to_action);
+  }
+
+
+  /**
+   * @return string
+   */
+  protected function getTimeCommitment() {
+    // @TODO: I've renamed "active_hours" to "time_commitment" because it sounds more straightforward; but appreciate feedback.
```
aaronschachter, line 256: My only concern here, like we've talked about in chat, is it confusing to know what field the property is pulling from.  I know there's that open issue around renaming fields..

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
```diff
@@ -0,0 +1,285 @@
...
+  protected function getStatus() {
+    return dosomething_helpers_extract_field_data($this->node->field_campaign_status);
+  }
+
+
+  protected function getType() {
+    return dosomething_helpers_extract_field_data($this->node->field_campaign_type);
+  }
+
+}
```
aaronschachter: :crying_cat_face:

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
weerd, line 256: Yep. There might be at first, but this is proposed to be changed...  there's an issue for it #4506. Particularly for the API side, `active_hours` doesn't seem to mean much... are their `inactive_hours`? Time commitment just felt more straightforward for API clients, but if it's going to be too confusing I can change back.

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -10,6 +10,21 @@ include_once('dosomething_helpers.theme.inc');
...
+/**
+ * Utility function to check if a variable is set, and return it
+ * if it is, otherwise return a default value.
+ *
+ * @param $var
+ * @param null $default
+ *
+ * @return array|object|string|integer|null
+ */
+function dosomething_helpers_issetor(&$var, $default = NULL) {
```
aaronschachter: Minor, but this name just feels weird to me.. like its a mispelling of `setter`. Also is passing the variable by reference on purpose?

weerd: Yeah, wasn't too fond of the name, but couldn't think of an alternative. I found this approach online and liked it, and they used the reference but wasn't too sure what the benefit was. Was hoping someone else had a more specific answer. Also naming suggestions?

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
weerd: Why sad cat? haha

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
weerd, line 194: Cool cool. Wasn't planning on getting to that in this PR. Need to update Reportback resource as well and will use this. Gonna do campaign, reportback, and kudos access stuff in a single PR.

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
aaronschachter: The little NO sign is Github complaining that theres no empty line at the end of the file

lib/modules/dosomething/dosomething_campaign/includes/CampaignTransformer.php
```diff
@@ -0,0 +1,47 @@
...
+   */
+  protected function transform($campaign) {
+    $data = array();
+
+    $data += $this->transformCampaign($campaign);
+
+    return $data;
+  }
+
+}
```
aaronschachter: Same

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
weerd, line 256: We've chatted about it, but to clarify with others, `call_to_action` is changing to `tagline` because we're already doing that on the front-end and design side. A "call to action" is not actually data, but a design pattern that applies to a few things on the front-end, so to remove confusion there was a renaming. Now just trying to bring that into the backend as well.

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
weerd: Dammit! I need to add the editorconfig file :grimacing:

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
```diff
@@ -0,0 +1,285 @@
...
+    }
+
+    return NULL;
+  }
+
+
+  protected function getTaxonomyTerm($tid) {
+    $data = array();
+
+    $taxonomy = taxonomy_term_load($tid);
```
aaronschachter, line 238: This is something to consider. The loaded `$node` object will already have the taxonomy term object loaded. What I'm not sure of is whether or not this call will perform a secondary `taxonomy_term_load` for each term.  I suppose we can leave it for now and if performance is an issue, this would def be something to look into...

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
weerd: Actually file exists. I think it's my editor that's not adhering to it. :cry:

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
aaronschachter: Yeah I don't think we need the reference.  how about just `dosomething_helpers_isset` ?

lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -217,16 +217,123 @@ protected function transformCollection($items, $method = 'transform') {
...
+      if ($data->facts['problem']) {
+        $output['facts']['problem'] = $data->facts['problem']['fact'];
+      }
+
+      if ($data->facts['solution']) {
+        $output['facts']['solution'] = $data->facts['solution']['fact'];
+      }
+
+      if ($data->facts['sources']) {
+        $output['facts']['sources'] = $data->facts['sources'];
```
aaronschachter, line 272: Checking out the response in https://gist.github.com/weerd/580ac04fa53b7ec21a06

The fact source doesn't have the `raw` / `formatted` array like some of the other markdown values do.  Is this on purpose?

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
weerd, line 238: The `$node` object doesn't give you the loaded taxonomy term :( It only gives you an array of `tid`s. I believe that before in your `stdClass` campaign object, when using the "entity wrapper" stuff it would autoload the taxonomy term object if it wasn't already.

aaronschachter:
Little comments but overall looks great!  Awesome work!  :+1:

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
aaronschachter, line 194: This one is important to not forget about, open a GH issue?

lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -217,16 +217,123 @@ protected function transformCollection($items, $method = 'transform') {
...
+      }
+
+      $output['type'] = $data['type'];
+      $output['dark_background'] = $data['dark_background'];
+
+      return $output;
+    }
+
+    if ($data['type'] === 'video') {
+      // @TODO: include video transformation code here!
```
aaronschachter, line 333: Might be worth creating a GH issue for this too ,to not leave as as TODO forever, although the only current use we'd have for it is just displaying the video on the web using this class instead of the current implementation.

lib/modules/dosomething/dosomething_api/includes/Transformer.php
weerd, line 272: Nope. it's how it was already returned from the facts module :( It just returns an array of the combined sources and default to using the formatted value. Wasn't sure if it'd be a good idea to modify the fact module function and not break stuff @_@ Maybe something to update though... what say you?

lib/modules/dosomething/dosomething_api/includes/Transformer.php
weerd, line 333: Good call. I'll create a GH issue for it!

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
aaronschachter, line 238: Oh, I was thinking about the loaded preprocess vars.  Nevermind! :factory:

lib/modules/dosomething/dosomething_api/includes/Transformer.php
aaronschachter, line 272: ohh. hm. yeah that sounds like it'd break stuff.  Umm ideally we should update the fact stuff to be consistent, because from the API perspective, i'd expect this to not be HTML since it doesn't have that array. I don't think it should block merging this PR, perhaps just a new issue?

weerd, line 272: I could do something like this, to prep it for a future fix and not break the API once that fix happens:

```
facts: {
  problem: "lorem ipsum...",
  sources: {
    [ 
      formatted:  "<p>formatted text here...</p> " 
    ],
    [ 
      formatted:  "<p>formatted text here...</p> " 
    ]
  }
},
```

Add as nested array. Once we can update the facts module to provide both we add the `raw` item to the array :)

aaronschachter, line 272: Ooh, smarty. sounds good

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
weerd, line 194: :+1: Will open as soon as I merge so I can reference the lines in the code :)

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
weerd: Sounds good. Updated and pushing up soon.

lib/modules/dosomething/dosomething_api/includes/Transformer.php
weerd, line 272: Updated and pushing up soon!

PR #4495, opened 2015-05-14
Kill 'has new content' check box in content search

### LINKED ISSUE #4489: new content issue when paginating through content search
when you click on a new page on https://beta.dosomething.org/admin/content/search the results go from 5291 to 632 (as of the writing of this issue). The reason for this huge drop in available content is that "Has new content" is selected when you clickthrough to other pages and there is nothing a user can do about that.

see https://trello.com/c/MbH8YJh3 for another example.

ideally "has new content" would just never be selected for a user as she pages through the content

angaither:
Fixes #4489


aaronschachter:
:+1:

PR #4508, opened 2015-05-15
Programatically finding full path to font files

sbsmith86:
Updated the imagemagick script to find the full path of the font files and image directory programatically so we don't need to worry about it in different environments.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign_problem_shares/create-images.sh
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign_problem_shares/create-images.sh
@@ -2,23 +2,24 @@
 
 TEMP_FILE_1=tmp1.png
 TEMP_FILE_2=tmp2.png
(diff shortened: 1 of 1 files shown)


angaither:
:+1:

DFurnes:
I think you might be able to use the `DS_APP_ROOT` environment variable to find the path. Does using `find` have a performance hit if it has to search for the file each time?

sbsmith86:
@DFurnes not sure we can use that variable in the shell script...

DFurnes:
Yeah, neither am I. I think it's set when provisioning boxes, so should be available from any shell script but not entirely sure. @sergii-tkachenko or @blisteringherb?

sergiitk:
@DFurnes, no this variable will not be available on staging and production boxes.
See https://github.com/DoSomething#5

DFurnes:
Okay, never mind! :grin: :+1:

PR #4513, opened 2015-05-18
Redirect to user profile, after search.

### LINKED ISSUE #4498: remove unnecessary click on /admin/users
update https://beta.dosomething.org/admin/users so that when admins search via UID, email, or mobile they don't have to click through the user's profile but instead get immediately taken to the profile they are looking for.

each of these fields should only only yield one result anyway, and so basically we are taking out an unnecessary click.

in the event that we do have multiple entries, just take the user to the first result.

cc: @namimody

angaither:
Fixes #4498


aaronschachter:
Ooh, fancy :+1:

Issue #4504, opened 2015-05-15
Update imagemagick script to use current site path as parameter


sbsmith86:
Update the imagemagick script to take in the base path of the current site as a parameter. Update the cron job to pass in that path when it runs the job.

PR #4518, opened 2015-05-20
Fixing cron and image script to work in correct directory

sbsmith86:
Updates to the cron and image magic script work in the correct directory in each environment. 

Ref #4504


angaither:
:+1:

Issue #4514, opened 2015-05-18
Under 13 User Flow


namimody:
Trello: https://trello.com/c/0movvETM/1376-under-13-account-sign-ups
Related: Issue #932 and Issue #931

**Users who are under 13 years old should:**
- be allowed to register for the site
- be given user accounts
- have their information stored in our database
- receive appropriate transactional emails
- should NOT have their information passed to MailChimp
- should NOT have their information passed to Mobile Commons

**User Examples:**
UID: 2389099

User showed up in Mailchimp (unsubscribed on May 6)
User never gave mobile number, so not in Mobile Commons.
For context, they are signed up for many campaigns - it is unclear how they initially joined DS.
++++++++
UID: 2392236

User shows up in MailChimp.
User entered in Mobile Commons and would receive next weekly broadcast.
For context, they are signed up for Comeback Clothes - it is unclear how they initially joined DS.
++++++++
UID: 2748641 (Nami test account)

User does not show up in MailChimp.
User receives transactional emails for signing up/reporting back for a campaign.
User entered in Mobile Commons and would receive next weekly broadcast.
For context, this account was created by clicking "log in" in the top right corner of the site, and on Chrome.
++++++++
UID: 2748914 (Nami test account 2)

User does not show up in MailChimp.
User never gave mobile number, so not in Mobile Commons.
For context, this account was created by signing up for Support Board on Safari.

**What Needs to Happen:**
- Under 13 users should consistently not be added to MailChimp
- Under 13 users should not be added to Mobile Commons

mikefantini:
user 2392236 has been opted in to the general Mobile Commons subscription list AND into individual campaign specific subscriptions (see subscriptions table on https://dosomething.mcommons.com/profiles/173868141)

interestingly user 2748641 was created in Mobile Commons but was not added to any specific subscription lists. I would have expected her to be signed up to Do Something Master campaign as a General DoSomething Web Sign Up https://dosomething.mcommons.com/campaigns/5091/opt_in_paths/164905.

**update**: this user signed up for a campaign that is tied to a mobile commons opt-in path that has ended, which would explain why she was not added to any specific subscription lists https://dosomething.mcommons.com/campaigns/130689

anyways, if a user is under 13 s/he should never be passed into Mobile Commons or MailChimp

@angaither do you mind taking a look at this?

mikefantini:
did a little more digging...

created user 1704889 on staging. that user is properly flagged as being "under 13".

[...]

PR #4519, opened 2015-05-21
To help debug issues seen around under 13 people

angaither:
Allows staff members to see the under 13 field, just not edit it.
Refs #4514


aaronschachter:
:+1:

PR #4525, opened 2015-05-21
Saving images as png files

### LINKED ISSUE #4522: compression on problem share image on FB
image

sbsmith86:
## Fixes #4522

Facebook applies high compression on jpgs so I switched to saving the images as .png files instead. 

73a90a74b40


angaither:
:+1:

PR #4528, opened 2015-05-21
Campaign class transformer docs

weerd:
Adding/updating DocBlocks for a variety of `Campaign` class and `Transformer` class methods. Also doing a bit of clean up on some methods and removing one no longer in use.

@aaronschachter 

CC: @uy @angaither


aaronschachter:
:+1:

PR #4530, opened 2015-05-22
Fix for small bug.

weerd:
Really quick fix for retrieving the id, since `Reportbacks` & `ReportbackItem` classes not updated with new approach, and returning `null` for campaign `id`.

@aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -234,7 +234,7 @@ abstract class Transformer {
    */
   protected function transformCampaign($data) {
     $output = array(
-      'id' => $data->id,
+      'id' => $data->id ?: $data->nid,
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -234,7 +234,7 @@ protected function transformCollection($items, $method = 'transform') {
    */
   protected function transformCampaign($data) {
     $output = array(
-      'id' => $data->id,
+      'id' => $data->id ?: $data->nid,
```
aaronschachter, line 237: Had to look up this syntax cause i'd never seen it before. seems fine. somewhere in the internet i feel like it discourages shorthand syntax, because it makes the code less readable... can't seem to find it though.  Seems fine for now tho

aaronschachter:
:+1:

Issue #4497, opened 2015-05-14
Sort by flagged/promoted reasons


angaither:
This actually is probably more useful for promoted

You can filter/sort by the flagged/promoted reasons 

Refs #4496

PR #4531, opened 2015-05-22
Added sort by flagged reasons

angaither:
Exposed the flagged reasons to views, as easy as this was I think we can keep `promoted` & `promoted_reason` the same as this... thoughts @aaronschachter 

Refs #4497


aaronschachter:
Sounds good :+1:

PR #4533, opened 2015-05-22
Adding initial access control to a few resources.

### LINKED ISSUE #4521: Expand on access control measures for resource access in API
https://api.drupal.org/api/drupal/modules%21node%21node.module/function/node_access/7

Need to update the following resources:
- [x] reportbacks
- [x] campaigns
- [x] kudos

weerd:
### Fixes #4521

Added some initial access control for main _view_ and _index_ of resources.

Also fixed a couple bugs that were logging warnings.

@aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -234,7 +234,7 @@ abstract class Transformer {
    */
   protected function transformCampaign($data) {
     $output = array(
-      'id' => $data->id ?: $data->nid,
+      'id' => isset($data->id) ? $data->id : $data->nid,
       'title' => $data->title,
     );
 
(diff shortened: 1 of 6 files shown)


lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -234,7 +234,7 @@ protected function transformCollection($items, $method = 'transform') {
    */
   protected function transformCampaign($data) {
     $output = array(
-      'id' => $data->id ?: $data->nid,
+      'id' => isset($data->id) ? $data->id : $data->nid,
```
weerd, line 237: @aaronschachter changed to use the longer form because it was throwing a notice that the property was undefined... I thought it was registering as `null` but I was incorrect. So need the longer form to check if set first!

aaronschachter:
:+1:

PR #4535, opened 2015-05-27
Promoted reasons

### LINKED ISSUE #4496: Promoted options
Like having flagged reasons, we should build 'promoted reasons' checkboxes

Options include
- [ ] photo
- [ ] quote 
- [ ] high impact  
- [ ] sponsored

angaither:
Add in promoted reasons checkboxes

Fixes #4496

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
@@ -96,6 +96,12 @@ function dosomething_reportback_files_form($form, &$form_state, $status = NULL, 
     'inappropriate' => t('Inappropriate'),
     'unrealistic' => t('Unrealistic quantity'),
   );
+  $promoted_options = array(
+    'photo' => t('Good photo'),
+    'quote' => t('Good quote'),
+    'sponsored' => t('Good for sponsor'),
+    'impact' => t('High impact')
+  );
 
   $params = array(
     'status' => $status,
@@ -178,6 +184,10 @@ function dosomething_reportback_files_form($form, &$form_state, $status = NULL, 
         '#options' => $status_options,
         '#default_value' => $status,
       ),
+      'promoted_reason' => array(
+        '#type' => 'checkboxes',
+        '#options' => $promoted_options,
+      ),
       'flagged_reason' => array(
         '#type' => 'checkboxes',
         '#options' => $flagged_options,
@@ -230,11 +240,14 @@ function dosomething_reportback_files_form_submit($form, &$form_state) {
   drupal_set_message(t("Updated."));
 }
 
+
+
 /**
  * Themes the dosomething_reportback_files_form constructor.
  */
 function theme_dosomething_reportback_files_form($variables) {
   $form = $variables['form'];
+
   $view_mode = $form['view_mode']['#value'];
 
   $rb_files = element_children($form['rb_files']);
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
```diff
@@ -209,16 +219,13 @@ function dosomething_reportback_files_form_submit($form, &$form_state) {
...
     $rbf = reportback_file_load($fid);
 
-    // Loop through the flagged_reason values:
-    foreach ($values['flagged_reason'] as $reason_key => $reason_value) {
-      // Form sends unchecked as a 0.
-      if (!is_string($reason_value)) {
-        unset($values['flagged_reason'][$reason_key]);
-      }
+    // Set flagged or promoted reasons.
+    if ($values['status'] == 'promoted') {
```
aaronschachter: @sergii-tkachenko usually flags these to use `===`

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
```diff
@@ -209,16 +219,13 @@ function dosomething_reportback_files_form_submit($form, &$form_state) {
...
+      dosomething_reportback_set_promte_flag_reasons($values, 'promoted_reason');
```
aaronschachter: Can you fix spelling?

aaronschachter:
Just confirming, so once a Reportback Item is promoted, the whole Reportback is considered promoted? If an editor makes a mistake, will this remove the promoted flag upon changing the Reportback Item status?

lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc
```diff
@@ -199,19 +199,25 @@ class ReportbackEntity extends Entity {
    *
    * @param string $status
    *   The status of the reprotback file (approved, flagged, excluded, promoted)
-   * @param string $reason
-   *   Text to store in the flagged_reason property.
+   * @param array $values
+   *   Array storing values for flagged/promoted reasons.
    */
-  public function setFlagged($status = NULL, $reason = NULL) {
+  public function setFlagged($status = NULL, $values = NULL) {
```
aaronschachter: Should we rename this, since it sets Promoted as well as Flagged?  Maybe something like `setFlags` ?

angaither:
@aaronschachter yeah, it works the same way flagged does, the `promoted` value will toggle between `0/1`

lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc
angaither: yeah, I thought about that, but it's not really `flags` I couldn't come up with a good name for it...

aaronschachter: Yeah.. hmm. `setProperties`... `setExtras`... `setFlaggedPromoted` ... ?

angaither: loving `setFlaggedPromoted`

aaronschachter:
:+1:

aaronschachter:
@angaither  - Random thought around this came to mind. What if a reportback has two items we want to promote, but for different reasons?

Once a reportback item is flagged, the reportback is ignored entirely, so there would never be the opportunity to update the flagged reason upon adding additional litems.  With this its different... wouldnt' the last promoted item always overwrite the reason?

PR #4539, opened 2015-05-29
Feature update to add campaign retrieve single endpoint.

weerd:
Feature to activate the single Campaign retrieve endpoint, as well as activating endpoints for the Kudos.

Use `drush -y fr dosomething_api` to update after rebasing code.

@aaronschachter


aaronschachter:
:+1:

PR #4542, opened 2015-06-01
Set campaign status to active

### LINKED ISSUE #4540: Write and update hook to activate campaigns
Quite a few campaigns have `campaign_status = NULL` 
all instances of this should be updated to `campaign_status = active`

angaither:
Fixes #4540


aaronschachter:
:+1:

PR #4545, opened 2015-06-01
Fix to restore reportbacks admin table.

weerd:
Fixes broken Reportbacks Admin view, for seeing list of Camapigns and numbers associated with their reportback counts and statuses.

@aaronschachter


aaronschachter:
:+1:

PR #4553, opened 2015-06-02
Kudos entityification

### LINKED ISSUE #4580: Implement Kudos API
Add index, retrieve, create and delete functionality.

weerd:
### Fixes #4580

Wrapping up initial run on Kudos API and making Kudos an entity.

@aaronschachter 

CC: @jonuy @DFurnes @chloealee


lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -436,7 +436,7 @@ protected function transformReportbackItem($data) {
    */
   protected function transformUser($data) {
     return array(
-      'id' => $data->uid,
+      'id' => isset($data->id) ? $data->id : $data->uid,
```
aaronschachter, line 439: This is where shorthand can def make code less readable... what's goin on here?

lib/modules/dosomething/dosomething_kudos/includes/KudosController.php
```diff
@@ -0,0 +1,79 @@
...
+  /**
+   * Overrides create() method in EntityAPIController
+   *
+   * @param array $values
+   * @return bool
+   *
+   * @throws Exception
+   */
+  public function create(array $values = []) {
+    $values += ['is_new' => TRUE];
```
aaronschachter: What's the purpose of overriding `create` here?

lib/modules/dosomething/dosomething_kudos/includes/KudosController.php
```diff
@@ -0,0 +1,79 @@
...
+      return TRUE;
+    }
+
+    return FALSE;
+  }
+
+
+  /**
+   * Overrides delete() method in EntityAPIController.
+   * Provides better feedback for API.
```
aaronschachter: Can you also specify which lines the override is for here too?  I'm just wondering what the specifics are for why there aren't any calls to the parent functions in these...

lib/modules/dosomething/dosomething_kudos/includes/KudosTransformer.php
```diff
@@ -0,0 +1,141 @@
...
+      'reportbackitem_id' => $this->formatData($parameters['reportbackitem_ids']),
+      'count' => (int) $parameters['count'],
+    ];
+
+    $results = dosomething_kudos_get_kudos_query($filters, $filters['count']);
+
+    if (!$results) {
+      return [
+        'error' => [
+          'message' => 'No kudos for you. Move along.',
```
aaronschachter: This reads like a permission denied error to me, not a results not found.

lib/modules/dosomething/dosomething_api/includes/Transformer.php
weerd, line 439: So this could use an eventual refactor pass. Sometimes when the data is passed to the method it's property name is `id`, other times it's `uid`... it's the same situation that happened when transforming a Campaign:

https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_api/includes/Transformer.php#L237

lib/modules/dosomething/dosomething_kudos/includes/KudosTransformer.php
weerd: Hehe. Oops. I was channeling my inner Soup Nazi...

lib/modules/dosomething/dosomething_kudos/includes/KudosController.php
weerd: So the `create()` method for entity, doesn't return an ideal value to help then provide an appropriate response for the API. At least I was having trouble with it. Might be something we can come back to and refactor to re-use the method if a different approach works better :)

weerd:
@aaronschachter fixed up the `create()` method in the `KudosController` class. You were correct and I was kinda overriding and not using the parent create method for no real reason. Modified it, but left the `create()` method in the `KudosController` so it can then proceed with saving and returning a custom boolean value.

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.kudos.inc
```diff
@@ -1,33 +1,33 @@
 <?php
```
aaronschachter: Is this file just not used anymore?  Time to kill?

lib/modules/dosomething/dosomething_kudos/includes/Kudos.php
```diff
@@ -0,0 +1,62 @@
...
+
+    // User
+    $user = ['id' => $this->entity->uid];
+    $this->user = (object) $user;
+  }
+
+
+  /**
+   * Get taxonomy term node data from provided id.
+   * @TODO: Potentially extract code to dosomething_helpers since duplicate code with Campaign.php
```
aaronschachter, line 47: That's fine but can you create an issue for this so it doesn't get lost in the TODOs?

aaronschachter:
Can you also clarify why delete is being overridden?  And why we can't use the parent `delete()` there?

weerd:
Made an issue :)
https://github.com/DoSomething/dosomething/issues/4583

PR #4562, opened 2015-06-04
Quick update to property name.

### LINKED ISSUE #4561: Change "item" to "reportback_item" in API response.
Now that a few different endpoint are coming around, such as `/campaigns` and `/kudos`, having nested objects in the respective objects for the endpoints labeled as `item` doesn't give enough context to what that item is...

```
// Single Kudos (no context)
{
  data: {
    id: "4",
    term: {
      id: "9",
      name: "something here"
    },
    item: {
      id: "647"
    },
    user: {
      id: "1203"
    }
  }
}

vs

// Single Kudos (with context)
{
  data: {
    id: "4",
    term: {
      id: "9",
      name: "something here"
    },
    reportback_item: {
      id: "647"
    },
    user: {
      id: "1203"
    }
  }
}
```

weerd:
### Fixes #4561

This fix will help give more context when `reportback-items` object is nested in other objects other than the `reportbacks`... for example, when they are nested in a `kudos` response object. See associated issue for code example.

@aaronschachter 

CC: @jonuy @angaither @DFurnes @chloealee 

PS> Will update the wiki shortly!

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -382,7 +382,7 @@ abstract class Transformer {
     if ($data->items) {
       $items = $this->getReportbackItems($data->items);
 
-      $output['items'] = array(
+      $output['reportback_items'] = array(
         'total' => count($items),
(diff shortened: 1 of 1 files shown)


aaronschachter:
:+1:

PR #4563, opened 2015-06-04
Add problem shares to make file/ds.info

### LINKED ISSUE #4558: Add problem shares modules to make file
We deployed problem shared to prod but had to turn `dosomething_campaign_problem_shares` on manually. We should add this to the make file so it gets turn on automatically.

angaither:
Fixes #4558


aaronschachter:
:+1:

PR #4586, opened 2015-06-09
Passing the -l param to drush to fix cron

blisteringherb:
In order for links and image paths to be correctly indexed by search, drush needs the -l or --uri params.


ghost:
:+1:

PR #4600, opened 2015-06-23
Send user updates to aurora

### LINKED ISSUE #4549: Undefined Northstar user attributes
Notice: Undefined index: und in dosomething_northstar_build_ns_user() (line 75 of /var/www/staging.beta.dosomething.org/releases/20150601213157/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module).

Only send attributes if they exist, in this case, there was no mobile number set.

### LINKED ISSUE #4570: Forward user updates from Drupal to Northstar
When users update their profile on Drupal, ensure these updates are also reflected on Northstar.

Guzzle should be available to use soon: #4411

angaither:
On use profile saves, this sends the user array into northstar.
Also refactored user registration to use guzzle to send requests. 

Fixes #4549
Fixes #4570

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -43,6 +43,9 @@ function dosomething_northstar_libraries_info() {
   return $libraries;
 }
 
+/**
+ * Send user registration events to northstar.
+ */
 function dosomething_northstar_register_user($form_state) {
   global $user;
   // Build a user that Northstar expects.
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -107,41 +148,39 @@ function _dosomething_northstar_build_http_client() {
...
-//       ));
-//   }
-
-// }
+class Northstar {
+
+  protected $client;
+
+  public function __construct() {
+    $this->client = $library;
```
aaronschachter: How is `$library` defined? In line 163?

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -107,41 +148,39 @@ function _dosomething_northstar_build_http_client() {
...
+    $base_url = NORTHSTAR_URL;
+    if (getenv('DS_ENVIRONMENT') === 'local') {
+      $base_url .=  ":" . NORTHSTAR_PORT;
+    }
+    $version = NORTHSTAR_VERSION;
+
+    $library = libraries_load('guzzle');
```
aaronschachter: Need to ensure that `libraries_load` function returns TRUE before calling below.

angaither: updated!

aaronschachter:
:+1:

PR #4609, opened 2015-06-23
Campaign api update

### LINKED ISSUE #4554: Missing Campaign reportback noun/verb properties at GET campaigns/:id
Up to you how to return these, but we'll need these for display in the mobile app:
- `reportback_verb`
- `reportback_noun`

### LINKED ISSUE #4564: Add campaign `high_season` field to the campaigns endpoint
Each campaign has a `high_season` field showing when a campaigns hot phase starts and ends. Can you add this to the campaigns endpoint?

weerd:
### Fixes #4564

Also provides some clean up and fixes some warnings that were being reported in the \* Recent log messages\* in Drupal Admin.

The `timing` output includes both `high_season` and `low_season` if available and all are output as ISO-8601, which is the [recommended standard](http://apiux.com/2013/03/20/5-laws-api-dates-and-times/) for APIs.
This also provides a way to request either a _teaser_ or _full_ (more could be added) extent of data returned for the campaign object response. This helps keep the response trimmed down if all the campaign info is technically not needed. Could be fine-tuned further if needed.

@aaronschachter @sbsmith86 

CC: @jonuy @angaither @DFurnes

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -232,102 +232,115 @@ abstract class Transformer {
    *
    * @return array
    */
-  protected function transformCampaign($data, $display = 'teaser') {
+  protected function transformCampaign($data) {
     $output = array(
       'id' => isset($data->id) ? $data->id : $data->nid,
       'title' => $data->title,
     );
 
+
     if ($data instanceof Campaign) {
-      if ($display === 'full') {
 
+      if (isset($data->tagline)) {
         $output['tagline'] = $data->tagline;
+      }
+
+      if (isset($data->created_at)) {
         $output['created_at'] = $data->created_at;
+      }
+
+      if (isset($data->updated_at)) {
         $output['updated_at'] = $data->updated_at;
+      }
+
+      if (isset($data->time_commitment)) {
         $output['time_commitment'] = $data->time_commitment;
-        // $output['type'] = $data->type; //@TODO: Should type be included? Consider there is an SMS Campaign type...
+      }
 
-        if ($data->status) {
-          // @TODO: Should the status default to "active"?
-          $output['status'] = $data->status;
-        }
+      // $output['type'] = $data->type; //@TODO: Should type be included? Consider there is an SMS Campaign type...
 
-        if ($data->cover_image) {
-          foreach ($data->cover_image as $key => $image) {
-            if (!is_null($image)) {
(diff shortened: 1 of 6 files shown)


lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -238,19 +238,33 @@ protected function transformCampaign($data) {
       'title' => $data->title,
     );
 
+
```
jonuy: :space_invader:

weerd: (╯°□°）╯︵ ┻━┻

jonuy:
Just curious, in `Transformer.php`, what's the difference in logic that you're accounting for by using `isset` for all the checks?

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
```diff
@@ -22,54 +24,70 @@ public static function get($id) {
...
+        $timing = $this->getTiming();
+        $this->timing = $timing;
+      }
 
-      $action_types_data = $this->getActionTypes();
-      $this->action_types = [
-        'primary' => $action_types_data['primary'],
-        'secondary' => $action_types_data['secondary'],
+      $reportback_info = $this->getReportbackInfo();
+      $this->reportback_info = [
```
jonuy: I may be wrong, but doesn't `$this->getReportbackInfo()` already format its returned array in the same way you're doing right here?

jonuy: The difference that I think this does make is that if `copy` or `noun` or whatever is actually not set in the return from `getReportbackInfo()`, then `$this->reportback_info` has that property set to... NULL? Or does it throw an exception trying to access a property that doesn't exist?

weerd:
Yeah, I had to change it up a slight bit. Since the `display` is now being handled in the `Campaign` class (figured it best in there cause that's where the processing is happening, so if we only want teaser info, then why collect all of it only to show some). Meaning, properties now (if in teaser mode) just straight up don't show up instead of prior being set to `NULL`. So now need to check if they're set when in the Transformer which processes the info from the Campaign object.

aaronschachter:
@weerd -- Not seeing this on staging, has this been pushed to stage?

Also don't see a sample response, but question/confirming: the High Season contains two dates as it's a date range (start and end) ?

weerd:
Got merged in EOD. Staging deploy happened a little while ago!

PR #4613, opened 2015-06-24
Fixes PHP Fatal error: Class 'Transformer' not found during ds install

sergiitk:
The error caused by incorrect dependencies.


sergiitk:
CC @weerd.

DFurnes:
Tested on my local. Fixed it! :ship:

PR #4619, opened 2015-06-25
Update to fix Kudos response object.

weerd:
Fixes Kudos response object to represent the following:

```
data: {
  id: "1",
  term: {
    id: "9",
    name: "make something"
  },
  reportback_item: {
    id: "647"
  },
  user: {
    id: "12930"
  }
}
```

*Note: Currently the `transform()` method on the `KudosTransformer` isn't doing much other than passing the `$kudos` object straight through. However, this doesn't mean that transforming stuff won't be needed at a later date. Just FYI!

@jonuy


jonuy:
:+1:

Issue #4606, opened 2015-06-23
Graph


angaither:
create a graph in d3? that tracks the goal progress as it advances

sbsmith86:
I'm looking at using [Chart.js](http://www.chartjs.org/) for this. It is lightweight, because we only have to load in the script for the chart type we want to use (~12kb) and seems to have a polyfill for IE8 support. 

@mshmsh5000 @angaither @mikefantini @namimody @DoSomething/front-end

angaither:
:+1:

mshmsh5000:
:smiley_cat: Sounds good for a prototype. And the docs include guidance for the data array format that will come from the back end.

sbsmith86:
@angaither is it

sbsmith86:
The graph is pretty much done but right now it is using dummy data. I am going to leave this issue open until #4607 is done and I can pull in real data and finish this graph up.

PR #4627, opened 2015-06-29
Hot Module Graph

sbsmith86:
This PR creates a prototype of the graph that will be used in the hot module. For now, it uses dummy data and displays on the action page only if the hot module feature flag is on. We will eventually be using a different template for campaigns in the hot phase but wanted to get this in front of some eyes. Ref #4606 

@DoSomething/front-end


DFurnes:
Looking good! :smile:

PR #4628, opened 2015-06-29
Progress log table

### LINKED ISSUE #4605: Add a table to record the history of campaign quantity progress
`dosomething_reportback_progress_log` table to store campaign goal total hour by hour

on cron store the total quantity towards the goal 

| id | campaign_id | timestamp | quantity |
| --- | --- | --- | --- |
| 1 | 363 | 1435075581 | 24 |
| 2 | 363 | 1435076581 | 40 |

angaither:
Fixes #4605

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
@@ -255,7 +255,10 @@ function dosomething_reportback_schema() {
         'default' => 0,
       ),
     ),
-    'primary key' => array('id', 'nid'),
+    'primary key' => array('id'),
+    'indexes' => array(
+      'nid' => array('nid'),
+    ),
   );
   // Add a custom cache table.
   $schema['cache_dosomething_reportback'] = drupal_get_schema_unprocessed('system', 'cache');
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
```diff
@@ -228,6 +228,35 @@ function dosomething_reportback_schema() {
...
+        'not null' => TRUE,
+      ),
+      'timestamp' => array(
+        'description' => 'The Unix timestamp of the time this quantity was stored.',
+        'type' => 'int',
+        'not null' => TRUE,
+        'default' => 0,
+      ),
+    ),
+    'primary key' => array('id', 'nid'),
```
sergiitk: Why both `'id'` and `'nid'`?

angaither: so it's easier to query search, it looked like need both, when I was testing old records were getting overwritten on each cron

sergiitk: They you should declare these as an index: https://api.drupal.org/api/drupal/includes!database!schema.inc/group/schemaapi, see `indexes`
Primary index is nor for search, but for unique identification of each row.

sergiitk: Auto-incremental `id` is sufficient for this purpose.

sergiitk:
:+1:

PR #4635, opened 2015-06-30
Fix typo that caused override to appear not to save.

### LINKED ISSUE #4632: Progress Update Does Not Save
I try to enter a number into the "progress update" field but it does not save in the field or reflect in the "total approved quantity."

angaither:
Fixes #4632


aaronschachter:
:+1:

PR #4647, opened 2015-07-01
Update entity_save from reportback_file to reportback

### LINKED ISSUE #4629: Incorrect entity_save ?
Question brought up by eagle eye @sergii-tkachenko in meeting around #4529

In https://github.com/DoSomething/dosomething/blob/v0.5.10/lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc#L218

Should `return entity_save('reportback_file', $this);` actually be `return entity_save('reportback', $this);`?

angaither:
Fixes #4629


mshmsh5000:
:+1: Let's test it on stage.

PR #4649, opened 2015-07-01
Kudos response update

### LINKED ISSUE #4650: Update Kudos response to group by taxonomy term and add total count
Sample desired output:

```
...
kudos: {
  data: [
    {
      term: {
        id: "5",
        name: "hellz yeah!"
      },
      kudos_items: {
        total: 2,
        data: [
          {
            id: "37",
            user: {
              id: "1920"
            }
          },
          {
            id: "45",
            user: {
              id: "1080"
            }
          }
        ]
      }
    },
    {
      term: {
        id: "17",
        name: "no way!"
      },
      kudos_items: {
        total: 1,
        data: [
          {
            id: "47",
            user: {
              id: "1080"
            }
          }
        ]
      }
    }
  ]
},
...
```

weerd:
### Fixes #4650

This PR makes a few updates to the Kudos response object to make it easier to work with (for mobile apps in particular) while also being more compacted. It adds some new _dosomething_kudos_ module helper methods to allow for sorting and grouping once a Kudos collection is retrieved.

@jonuy @aaronschachter 

CC: @DFurnes @angaither @chloealee

### DIFF
--- a/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
+++ b/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
@@ -44,6 +44,10 @@ function dosomething_kudos_build_kudos_query($params) {
   $query
     ->entityCondition('entity_type', 'kudos');
 
+  if (isset($params['fid'])) {
+    $query->propertyCondition('fid', $params['fid'], '=');
+  }
+
   return $query;
 }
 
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -65,3 +69,88 @@ function dosomething_kudos_get_kudos_query($params = [], $count = NULL, $start =
...
+  $data = dosomething_kudos_collect_by_taxonomy_term($data);
+  $data = dosomething_kudos_get_totals_by_taxonomy_term($data);
+  return $data;
+}
+
+
+/**
+ * Sort kudos collection by taxonomy term id.
+ *
+ * @param $a
```
aaronschachter: Defining these as types would be helpful

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -65,3 +69,88 @@ function dosomething_kudos_get_kudos_query($params = [], $count = NULL, $start =
...
+  if (!$result) $result = $a->id - $b->id;
+
+  return $result;
+}
+
+
+/**
+ * Collect all kudos in collection and organize them by taxonomy term.
+ *
+ * @param $data
```
aaronschachter: Same

aaronschachter:
Code looks fine, but it'd be nice to get more details on the input params to better understand what's happening.

weerd:
Cool cool. Will update!

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -65,3 +69,88 @@ function dosomething_kudos_get_kudos_query($params = [], $count = NULL, $start =
...
+/**
+ * Sort kudos collection based on taxonomy term and then
+ * collect and group them by term.
+ *
+ * @param array $data
+ *
+ * @return array
+ */
+function dosomething_kudos_sort($data) {
+  usort($data, 'dosomething_kudos_sort_by_taxonomy_term');
```
aaronschachter, line 89: @weerd Sorry if I'm missing something obvious, but I don't get why we're sorting by taxonomy term here.

Doesn't the grouping by term happen via the new line you added to `dosomething_reportback_get_reportback_files_query`:
`$query->addExpression('GROUP_CONCAT(DISTINCT k.kid)', 'kids');` ?

weerd:
So yeah, the group concat will collect all the Kudos ids for a RB item, however while all of those are unique kudos ids, there could be multiple of the same term, so what the above function does, is after sorting them in taxonomy term order, it groups them all up based on the term... if you check the issue this PR fixes it shows an example output :)

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -65,3 +69,94 @@ function dosomething_kudos_get_kudos_query($params = [], $count = NULL, $start =
...
+  $data = dosomething_kudos_group_by_taxonomy_term($data);
+  $data = dosomething_kudos_get_totals_by_taxonomy_term($data);
+  return $data;
+}
+
+
+/**
+ * Sort kudos collection by taxonomy term id using a value comparison function.
+ *
+ * @param array $a Single kudos object.
```
aaronschachter: Just clarifying: Kudos object == an instance of the `Kudos` class?  (One could also read this as a result row in `dosomething_kudos`... a Kudos taxonomy term)

weerd: gotcha. ok. I'll clarify :smile:

PR #4657, opened 2015-07-02
Upating response to clean it up and fix a bug.

weerd:
This PR updates the response to allow for specifying whether the Kudos response should include the `reportback_item` object which provides the `id` of the RB item the kudos belongs to. Normally when hitting the `/reportback-items` endpoint, the index of RB items in collection already has the RB item id, so the kudos collection doesn't need it and is thus not included to allow for a less redundant response.

However, if hitting the `/kudos` endpoint, then the index collection of kudos in response need the RB item  info, and thus this update adds it :)

Also updates the retrieve method on `/reportback-items` to now properly show an error if a non-existent RB item id is passed to the `/reportback-items/:id.json` endpoint.

@uy @aaronschachter


aaronschachter:
:+1:

Issue #4607, opened 2015-06-23
Output quantity JSON to hot phase module


angaither:
return a json object to the campagin in hot module phases with the quantity total for each day

```
select quanity 
from dosomething_quanity_log
where nid = X
group timestamp by dates
and n.timestamp is max where timestamps 
```

mikefantini:
@angaither this is so we can (more) easily populate the graph yeah?

angaither:
@mikefantini yup

sbsmith86:
@angaither All I need for the graph is 4 data points. One for each day we are showing on the x-axis. Is it possible to have a JSON object that holds just the four dates we want to show on the graph, and the goal progress on each of those dates (set as null if the date is in the future). So something like: 

```
{
  dates: ["JAN 1", "JAN 10", "JAN 21", "JAN 31"],
  value: [3000, 4500, 9000, null],
}
```

angaither:
@sbsmith86 hmm, originally when we talked about this we were going to send the latest timestamp from each date, because how do we pick which four dates we are going to use?
What happens when we run a campaign for a week?

sbsmith86:
@angaither yea, we still need that data to aggregate it into these 4 data points because the graph only shows 4 points. Could we determine the dates by getting using the hot phase start and end dates, divide by 4, and then calculate dates based on that interval? If the hot phase is only for a week, then we could still determine 4 dates in between start and end.

angaither:
oh boy

mikefantini:
@sbsmith86 can we sync on this? I thought we were going to be adding to the graph regularly, but only having markers at those 4 points in the date range (i.e. 4 markers within the one month).

I think the clean lines were just for ease on @lkpttn's part

![image](https://cloud.githubusercontent.com/assets/6330971/8432331/673ff5dc-1f0d-11e5-80eb-e4d8924cd1ec.png)

sbsmith86:
@mikefantini Ah yea, lets sync. You are right and I think I confused things talking about aggregating stuff. But regardless, I definitely need data about those four dates, so there will be some logic around figuring out those dates programmatically, either on the backend or the frontend.

mshmsh5000:
I think we can keep this on the front end. If you've got the and date and start date, you can divide by four and round up to get the two interim dates.

sbsmith86:
Just wrapping up this convo. I made a custom graph type that will allow us to go ahead and pull in all of the data for each day of the month like originally planned!

[...]

Issue #4602, opened 2015-06-23
Output campaign author information


angaither:
- [x]  ensure that staff members (editors) can change authors on the node
- [ ] output author information, only on campaigns in a hot phase (name, photo, job title)

mikefantini:
just a note: staff photos should be added on the campaign lead's profile on https://www.dosomething.org/user/[uid]/edit

@lkpttn just a heads up may need some of your assistance when we're ready to implement this

mikefantini:
final should look like this:

![image](https://cloud.githubusercontent.com/assets/6330971/8487931/60c4fa14-20dd-11e5-89b7-36c7ed25333a.png)

sbsmith86:
Confirmed that editors can change authors on the node.

PR #4660, opened 2015-07-06
Add progress log data

angaither:
Refs #4607
Refs #4602 

@sbsmith86 I've added the progress log data to the page, not sure if/how we want to json-ify if. 
can we can work on that together?


sbsmith86:
@angaither sure! I basically need two arrays, one with the dates, and one with the progress for each date. One thing we will also need to work through is how to send the data to the frontend, since these arrays need to be accessible in the JS, maybe in `Drupal.settings`?

aaronschachter:
Looks ok :+1:

Issue #4446, opened 2015-05-04
Hot phase progress copy


angaither:
There's a little cta when a campaign is in the hot phase that changes over time.
@mikefantini does this change over time, or when it's % complete near goal?

@mikefantini add the formula for this here, and we can decide if this should be fields or if it can be hardcode & pulled from the campaign.

mikefantini:
@angaither idea is that this would be tied to progress towards the goal.

definitely not final copy, but hoping it will look something like this https://docs.google.com/document/d/1o6Pw0zaszsrqzBTgagq89yzjdc74-4GtvzQ5VJMwMmk/edit

mikefantini:
@angaither thoughts on making this a group of fields in https://www.dosomething.org/admin/config/dosomething/dosomething_campaign under Standard Campaigns?

Suggested group name: "Hot Phase Progress Copy"

Hierarchy of this on https://www.dosomething.org/admin/config/dosomething/dosomething_campaign should probably be Standard Campaigns -> Action Page -> Hot Phase -> Hot Phase Progress Copy. (see the need for "Hot Phase" group on this config page in #4637) 

we'd need fields for 0-20%, 21-40%, 41-60%, 61-80%, 81-99%, and 100+%

maybe field names could be in the format of "0-20% Progress Copy" or "First Quintile Progress Copy"?

we need the following variables to work for these fields:
- [campaign_title]
- [reportback_noun]
- [reportback_verb]
- [campaign_goal]

also just to be clear this is the copy that appears in the speech bubble:

![image](https://cloud.githubusercontent.com/assets/6330971/8455116/44f2075e-1fd0-11e5-87ff-a30be8da36a0.png)

mikefantini:
we need the following variables to work for these fields:

- [campaign_title]
- [reportback_noun]
- [reportback_verb]

also just to be clear this is the copy that appears in the speech bubble:

![image](https://cloud.githubusercontent.com/assets/6330971/8455116/44f2075e-1fd0-11e5-87ff-a30be8da36a0.png)

angaither:
@mikefantini great, thanks for adding details.
I have a group already for "Goals and Hot module" it's just a standard grouping not underneath standard campaigns, but I can move it under  Standard Campaigns -> Action Page

mikefantini:
@deadlybutter once this is set up on the backend please reassign to @sbsmith86

itsjoekent:
[...]

PR #4661, opened 2015-07-06
Adds saftey net to the goal

itsjoekent:
For #4446 -- Fixes bug where sometimes the goal isn't defined for a campaign 

@angaither


angaither:
:+1:

Issue #4642, opened 2015-06-30
Make all exposed properties without values be NULL


weerd:
References #4641

PR #4663, opened 2015-07-06
Updating campaign response to include all properties and null values if property has no value.

weerd:
This PR addresses #4641 and #4642 for the `/campaigns` endpoint response. Returns all properties for campaign response object and assigns `NULL` to properties that do not have a value.

There is still a `display` value which can be passed (either `full` or defaults to `teaser`), which will dictate the extent of information provided. But with either display mode, the expected properties that each display type should have are always shown.

@uy @aaronschachter


aaronschachter:
Looks ok here :+1:

PR #4664, opened 2015-07-06
Add jsonified campaign progress

angaither:
@sbsmith86 do you want to check this branch out and see if it works for you?

Here's what I'm seeing locally on campaign nodes... I didn't add new reportbacks so the number is the same for all the days... but you get the point


sbsmith86:
@angaither Yea! Just checked it out and that is what I am seeing, but it looks like the dates and quantities arrays are actually strings?

screen shot 2015-07-06 at 5 24 36 pm

angaither:
@sbsmith86 let's see what we can do about that

angaither:
:cool: :cool: looks like `JSON.parse()` worked!

DFurnes:
Hmm, yeah would be great if we could include each one as an actual array, rather than string encoded. Looks fine other than that though! :+1:

PR #4665, opened 2015-07-07
Updates to fix bug in response.

weerd:
Update to the campaign response broke the expected output for `facts` property. This fix restores it to prior expected output!

@jonuy @aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -314,7 +314,9 @@ abstract class Transformer {
 
         $output['staff_pick'] = $data->staff_pick;
 
-        $output['facts'] = $data->facts;
+        $output['facts']['problem'] = $data->facts['problem'] ? $data->facts['problem']['fact'] : NULL;
+        $output['facts']['solution'] = $data->facts['solution'] ? $data->facts['solution']['fact'] : NULL;
+        $output['facts']['sources'] = $data->facts['sources'] ?: NULL;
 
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -314,7 +314,9 @@ protected function transformCampaign($data) {
 
         $output['staff_pick'] = $data->staff_pick;
 
-        $output['facts'] = $data->facts;
+        $output['facts']['problem'] = $data->facts['problem'] ? $data->facts['problem']['fact'] : NULL;
+        $output['facts']['solution'] = $data->facts['solution'] ? $data->facts['solution']['fact'] : NULL;
+        $output['facts']['sources'] = $data->facts['sources'] ?: NULL;
```
aaronschachter: I think shorthand makes this less readable. Drupal standards recommend always using `{`'s: https://www.drupal.org/coding-standards#controlstruct

aaronschachter:
Code looks fine, but also not readable per comment. Not needed for merge, but something to consider..

weerd:
Would it be cooler if I switch it to match the other shorthands:

```
$output['facts']['sources'] = $data->facts['sources'] ? $data->facts['sources'] : NULL;
```

kind of dig the compactness of the shorthand version so all the `facts` properties kinda stick together in the code ;)

aaronschachter:
That looks better at least because it matches the other ones.

weerd:
K!

PR #4668, opened 2015-07-08
hide social buttons when hot module is enabled

### LINKED ISSUE #4654: hide problem share prompt + social buttons when a campaign is hot
image

In `node--campaign.tpl.php` we need to update the conditional logic around displaying the problem shares in the "Know it" section so it only displays if the campaign is **not** in the hot phase. 

So anywhere we are checking `<?php if ($show_problem_shares): ?>`  is true. 

We should also make sure the campaign is not in the hot phase, so something like .

`<?php if ($show_problem_shares &&  !$hot_module_enabled) : ?>`

chloealee:
Fixes #4654


sbsmith86:
:dancer: ! looks good.

PR #4669, opened 2015-07-08
Allows editor to change reportback status if they made a mistake

### LINKED ISSUE #4621: changing a file's status should change the rb's flagged status
once a reportback is flagged, it always remains flagged, even if an editor made a mistake in flagging it and then changed it to be approved, promoted, excluded, etc.

we should allow for editors to change their minds (provided obviously they didn't delete an image).

so when a file's status changes from flagged to promoted the reportback reportback_flagged should change from 1 to 0

itsjoekent:
Fixes #4621 

also forgot to configure git, author is screwed up. oops!

@angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
@@ -281,6 +281,22 @@ function dosomething_reportback_set_promote_flag_reasons(&$values, $reason_type)
       }
     }
 
+    // Clear existing reasons
+    $alternate_reason = '';
+    if ($reason_type == 'promoted_reason') {
+      $alternate_reason = 'flagged_reason';
+    }
+    else {
+      $alternate_reason = 'promoted_reason';
+    }
+
+    foreach ($values[$alternate_reason] as $reason_key => $reason_value) {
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
```diff
@@ -418,5 +418,3 @@ function dosomething_reportback_count_page() {
 
   return $output;
 }
-
-
```
angaither, line 422: can you remove these changes?

angaither:
Looks ok, just want to double check, when you flag something, then unflag it, the reportback `flagged` boolean goes from 1->0

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
itsjoekent, line 422: looks like my editor auto did that, weird

itsjoekent, line 422: but I still have line 421 as a blank line. da fuq

itsjoekent:
correct, that is what happens.

PR #4670, opened 2015-07-09
Updates Drupal Core from 7.35 to 7.38.

sergiitk:
See https://www.drupal.org/SA-CORE-2015-002.
Build test passed.


angaither:
:+1:

PR #4671, opened 2015-07-09
Hot Module Graph updates

sbsmith86:
Update Graph.js so that it uses real data. Finishes #4606. 

@DoSomething/front-end


DFurnes:
Cooooooool. :clap:

PR #4674, opened 2015-07-09
Changes end date to use proper value

### LINKED ISSUE #4637: Hot phase share copy
need a field in https://www.dosomething.org/admin/config/dosomething/dosomething_campaign for what the share prompt should be during the hot phase

Suggested field name "Hot Phase Share Prompt".

### LINKED ISSUE #4673: High season end date not outputting correctly
image

from @sbsmith86 in slack

_i think the timing variable is off, i set the high season for the space jam campaign to end July 31st, and it is outputting that it ends July 1st._

itsjoekent:
Fixes #4673 

Changes the end date to use value2 instead of value


sbsmith86:
:dancer: !

PR #4675, opened 2015-07-10
Reportback response all the things

### LINKED ISSUE #4484: Include kudos info in reportback-items data returned from the API
For endpoints that return reportback-item data, also include kudos information in the returned data.

**Relevant endpoints:**
- `/api/v1/reportbacks/:rbid`
- `/api/v1/reportback-items`
- `/api/v1/reportback-items/:rb-item-id`

**Example return:**

_Only a suggestion. Feel free to add more data or re-organize as you see fit._

```
{
  data: {
    ...
    kudos: {
      kudos_id:  {
        count: Number,
        users: [id1, id2, ..., idN]
      }
    }
  }
}
```

### LINKED ISSUE #4490: API Reportback endpoint returns empty object instead of error message when not valid.
Slight bug with the response. Just need to update so if no item found the logic correctly understands that and sends through the error message instead of an object full of `null` values.

### LINKED ISSUE #4641: Expose all object properties in resources even if doesn't have a value.

### LINKED ISSUE #4642: Make all exposed properties without values be NULL

weerd:
# Updates!
##### Fixes #4484 (resolved earlier but this updates it to be a bit cleaner)

This PR updates to both the `Reportback` and `ReportbackItem` classes along with their respective transformers and controller classes.

More of the logic has been moved into the main `Reportback` and `ReportbackItem` classes, so that the returned instance has a lot of the properties that would otherwise be needed for both the API and the main Drupal app; data is more consistent in both for what to expect returned. The transformers now handle more of _just_ the transformations needed for API related data (processing and building it out for JSON response), rather than doing much heavy logic.

This also allows for easier passing of messages between the different classes, including the transformers, allowing for easier building of response objects of the API between the different types.

Also does a bunch of code cleanup and comment block updates as well as adding a few `@TODO`'s that will be turned into GitHub issues. 

_Apologies for the large PR! Had to do a bunch of work that would have killed the site if done in increments :(_

@aaronschachter @angaither  @DFurnes

CC: @jonuy @mshmsh5000


aaronschachter:
Yeahh this is pretty tough to review. Have you tested with a clean staging DB? Checked error logs for notices/warnings?

weerd:
Yeah, I've been testing/currently testing to make sure. Just wanted to get it up asap since i know it's a lot. I'll make notes here if I see any major things as I continue to test, but so far everything is working.

aaronschachter:
Sweet, thanks for testing. Looks good here  :+1:

weerd:
OK!
Here.
We.
Go!

conan walker texas ranger lever

PR #4676, opened 2015-07-10
hot module enabled when high season is current month

### LINKED ISSUE #4672: Ensure hot module is only enabled when the dates are correct
Hot module is enabled feature is only enabled when the dates are correct on that particular node. 

in `dosomething_campaign.module`

```
if (variable_get('dosomething_campaign_enable_hot_module', NULL) == 1 && !empty($vars['field_high_season']) && date_check(high season value is current month))) 
```

chloealee:
Fixes #4672 

Enables hot module when high season is current month.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -715,9 +715,12 @@ function dosomething_campaign_add_shipment_form_vars($node) {
 function dosomething_campaign_is_hot_module($vars) {
   $first_day_of_month = mktime(0,0,0,date('n'),1,date('Y'));
   $last_day_of_month = mktime(23,59,59,date('n'),date('t'),date('Y'));
-  $highSeasonStartDate = explode("-", $vars['field_high_season'][0]['value']);
-  $comp_date = mktime(0,0,0,$highSeasonStartDate[1],$highSeasonStartDate[2],$highSeasonStartDate[0]);
+  $high_season_start_date = explode("-", $vars['field_high_season'][0]['value']);
+  $comp_date = mktime(0,0,0,$high_season_start_date [1],$high_season_start_date [2],$high_season_start_date [0]);
 
+  // Hot module is enabled feature is only enabled when high season dates exist and start in the current month.
+  // @todo: Assumes that high season starts and ends in the same month.
+  // This logic will need to change if end of high season is in a different month than when it starts.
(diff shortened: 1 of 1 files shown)


chloealee:
@angaither this is under the assumption that the high season starts and ends in the same month, now that campaigns launch and end (?) on set dates. If this is not the case though, I'll need to update the logic to check the end date too - let me know!

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -713,7 +713,12 @@ function dosomething_campaign_add_shipment_form_vars($node) {
  * @return bool
  */
 function dosomething_campaign_is_hot_module($vars) {
-  if (variable_get('dosomething_campaign_enable_hot_module', NULL) == 1 && !empty($vars['field_high_season'])) {
+  $first_day_of_month = mktime(0,0,0,date('n'),1,date('Y'));
+  $last_day_of_month = mktime(23,59,59,date('n'),date('t'),date('Y'));
+  $highSeasonStartDate = explode("-", $vars['field_high_season'][0]['value']);
+  $comp_date = mktime(0,0,0,$highSeasonStartDate[1],$highSeasonStartDate[2],$highSeasonStartDate[0]);
+
+  if (variable_get('dosomething_campaign_enable_hot_module', NULL) == 1 && !empty($vars['field_high_season']) && (($comp_date >= $first_day_of_month) && ($comp_date <= $last_day_of_month))) {
```
angaither, line 724: add a comment above to make it more clear what we are testing here

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -713,7 +713,12 @@ function dosomething_campaign_add_shipment_form_vars($node) {
  * @return bool
  */
 function dosomething_campaign_is_hot_module($vars) {
-  if (variable_get('dosomething_campaign_enable_hot_module', NULL) == 1 && !empty($vars['field_high_season'])) {
+  $first_day_of_month = mktime(0,0,0,date('n'),1,date('Y'));
+  $last_day_of_month = mktime(23,59,59,date('n'),date('t'),date('Y'));
+  $highSeasonStartDate = explode("-", $vars['field_high_season'][0]['value']);
```
angaither: code style, for php we don't use camelCase

angaither:
As far as the logic for the campaigns starting and ending in the same month, I think that will be fine for this initial launch..but maybe add that as a @todo in the comments just to note that we will need to change it in the future.

chloealee:
got it, thanks @angaither ! Updated to get rid of camelCase, add comments for clarity, and added the @todo.

angaither:
:+1:

PR #4677, opened 2015-07-13
Always sets the status to pending when updated

### LINKED ISSUE #4625: Updating RB should reappear in inbox
Updated:

Any change to a reportback (new image, changed quantity, changed caption, changed why participated) should go back into the inbox and status should equal pending.

However, we don't want the quantity on the campaign to go below what it was prior to update so we're going to hold off on this until we can more robustly address this issue.

---

Submitted RB. Then approved. The total approved quantity showed correctly. 

Updated the quantity on the reportback, but was not given the opportunity to approve again. However, total approved quantity did update.

Related: https://github.com/DoSomething/dosomething/issues/4225#issuecomment-116768048

itsjoekent:
Fixes #4625 

@angaither


angaither:
:+1:

PR #4678, opened 2015-07-13
Add override for reportback delete page.

### LINKED ISSUE #4623: Can't Delete Reportbacks
I can get to this page but the button is unclickable.

image

DFurnes:
# Changes

Fixes #4623. Sadly, I don't think we can use default page template on this page without some kind of individual override, since it's technically a type `page__reportback` which we override to the "full-width" page template for.


weerd:
:+1:

PR #4679, opened 2015-07-13
Add Optimizley custom event when user is signed up for campaign.

### LINKED ISSUE #4617: Need JS Event for Action Pageviews Count in Optimizely
@DFurnes as we chatted about today -- I think we will still need an event to count action pageviews for Optimizely purposes. Please let me know if there are any other details I can provide - thanks!

DFurnes:
# Changes

Fixes #4617. This will let us use campaign signups (rather than just "signup attempts") as a metric for Optimizely tests. See Optimizely's help article for [implementation details](https://help.optimizely.com/hc/en-us/articles/200039925-Custom-event-goals).

For review: @DoSomething/front-end @angaither


angaither:
:+1:

PR #4680, opened 2015-07-13
Updates reportback total to use override

### LINKED ISSUE #4652: Override Feature Input Doesn't Register
On Senior Socials on staging (http://staging.beta.dosomething.org/node/9/), there is one approved report back, the total approved quantity correctly registers 15. However, inputting 1000 into the override field and running cron doesn't change the TAQ in any of the following places:

http://staging.beta.dosomething.org/campaigns/senior-socials
http://staging.beta.dosomething.org/node/9/custom-settings
http://staging.beta.dosomething.org/node/9/rb/reviewed

TAQ should say 1015.

itsjoekent:
Fixes #4652 

Updated the `dosomething_reportback_get_reportback_total_by_nid` function to also add `sum_rb_quantity_override` to the quantity. Also re-added a blank line that I accidentally removed in a previous PR. 

@angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -884,15 +884,7 @@ function dosomething_reportback_get_reportback_total_by_nid($nid, $type = 'quant
  *  Sum of all reportback quantities plus the override.
  */
  function dosomething_reportback_get_reportback_total_plus_override($nid) {
-   // Select all reportbacks for given $nid.
-   $query = db_select('dosomething_reportback', 'rb')->condition('nid', $nid);
-   // Get quantity of all reportbacks
-   $query->fields('rb', array('quantity'));
-   $query->addExpression('SUM(quantity)', 'quantity');
-   $result = $query->execute()->fetchAll();
-   $quantity = $result[0]->quantity;
-   dosomething_helpers_set_variable('node', $nid, 'sum_rb_quanity', $quantity);
-   return $quantity + (int) dosomething_helpers_get_variable('node', $nid, 'sum_rb_quantity_override');
+   return dosomething_reportback_get_reportback_total_by_nid($nid) + (int) dosomething_helpers_get_variable('node', $nid, 'sum_rb_quantity_override');
  }
 
 
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -864,7 +864,7 @@ function dosomething_reportback_get_reportback_total_by_nid($nid, $type = 'quant
     $query->fields('rb', array('quantity'));
     $query->addExpression('SUM(quantity)', 'quantity');
     $result = $query->execute()->fetchAll();
-    $quantity = (int) dosomething_helpers_get_variable('node', $nid, 'sum_rb_quantity_override') + (int) $result[0]->quantity;
+    $quantity = $result[0]->quantity;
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -864,7 +864,7 @@ function dosomething_reportback_get_reportback_total_by_nid($nid, $type = 'quant
     $query->fields('rb', array('quantity'));
     $query->addExpression('SUM(quantity)', 'quantity');
     $result = $query->execute()->fetchAll();
-    $quantity = $result[0]->quantity;
+    $quantity = (int) dosomething_helpers_get_variable('node', $nid, 'sum_rb_quantity_override') + (int) $result[0]->quantity;
```
angaither: I know there are a couple of places in the code where these two vars are already added together. This will cause the result to be double if we don't remove it from there. 
Also I'm a little nervous about having this query return the result of the summary and not having the two counts just added together in the campaign.theme.inc file...

lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc
```diff
@@ -333,7 +333,7 @@ class ReportbackEntityController extends EntityAPIController {
     }
 
     parent::save($entity, $transaction);
-
+
```
angaither: bah! extra spaces! Install [editor config](https://github.com/sindresorhus/atom-editorconfig) to avoid this stuffs

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
itsjoekent: Is campaign theme the only place this is needed? If so I can just add this in there.

angaither: Yeah, should be. I think the one that is displayed on the `dosomething.helpers` form already says the two totals will be added together in the theme

lib/modules/dosomething/dosomething_reportback/includes/dosomething_reportback.inc
itsjoekent: as I noted in the PR description, this used to be there but I accidentally removed it before. I guess I'll remove it again haha

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
itsjoekent: okie ill move it

itsjoekent:
@angaither good to go now?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -875,6 +875,28 @@ function dosomething_reportback_get_reportback_total_by_nid($nid, $type = 'quant
...
+ *  Sum of all reportback quantities plus the override.
+ */
+ function dosomething_reportback_get_reportback_total_plus_override($nid) {
+   // Select all reportbacks for given $nid.
+   $query = db_select('dosomething_reportback', 'rb')->condition('nid', $nid);
+   // Get quantity of all reportbacks
+   $query->fields('rb', array('quantity'));
+   $query->addExpression('SUM(quantity)', 'quantity');
+   $result = $query->execute()->fetchAll();
+   $quantity = $result[0]->quantity;
```
angaither: you could just call the other function `dosomething_reportback_get_reportback_total_by_nid` instead of copying the query here.

itsjoekent: right, crap

angaither:
:+1:

PR #4683, opened 2015-07-14
Updates problem fact help text

### LINKED ISSUE #4590: update problem fact help text
Because of problem shares, instead of "Limit: 135 characters" we need it to be "Limit: 117 characters (for the purposes of sharing this fact via social, this MUST be under 117 characters." 

image

itsjoekent:
Fixes #4590 

@angaither


angaither:
:+1:

PR #4687, opened 2015-07-15
Add setting helper.

DFurnes:
# Changes
- Adds `setting()` helper method, which gracefully handles failure if key does not exist & allows easy
  fallback to a default value. Example usage:

``` js
   import setting from './utilities/setting';

   $url = setting('dosomethingSearch.solrURL', 'http://search.dosomething.org');
```

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/utilities/setting.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/utilities/setting.js
@@ -1,0 +1,7 @@
+import dotty from 'dotty';
+
+function setting(key, defaultValue) {
+  return dotty.get(Drupal.settings, key) || defaultValue;
+}
+
+export default setting;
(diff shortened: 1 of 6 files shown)


lib/themes/dosomething/paraneue_dosomething/js/utilities/setting.js
```diff
@@ -0,0 +1,7 @@
+import dotty from 'dotty';
```
weerd, line 1: `dotty`?

weerd:
Nice :) +1

lib/themes/dosomething/paraneue_dosomething/js/utilities/setting.js
DFurnes, line 1: Yup! It's a [nice little helper](https://www.npmjs.com/package/dotty) for getting nested properties from objects.

PR #4688, opened 2015-07-15
Add Dotty to dependencies.

DFurnes:
# Changes

Whoops, missed adding [Dotty](https://www.npmjs.com/package/dotty) to dependencies in #4687.


weerd:
:+1:

PR #4691, opened 2015-07-15
Updates campaign form script

### LINKED ISSUE #4319: admin fields no longer hidden when campaign_type = SMS game
image

The following groups should be hidden:
- plan it
- do it
- prove it

itsjoekent:
Fixes #4319 

Updated the script to no longer be ID based like it previously was (cause of it breaking). New method is sort of hacky but was the best solution me & @DFurnes could come too. Currently we find the title for each field group, if it matches a 'campaign only' the code traverses up the DOM until it finds a form-wrapper class and adds that to an array. 

@DFurnes

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -19,7 +19,6 @@ define('DOSOMETHING_CAMPAIGN_PIC_STEP_HEADER', 'Snap a Pic');
  */
 function dosomething_campaign_form_campaign_node_form_alter(&$form, &$form_state, $form_id) {
   unset($form['field_active_hours'][LANGUAGE_NONE]['#options']['_none']);
-  $form['title']['#description'] = t('Title - i.e. "Protect your Pills", "Birthday Mail", or "Teens for Jeans" <br/><strong> Limit: 20 characters. </strong>');
   // Add in campaign close info.
   dosomething_campaign_run_add_campaign_run_info($form, $form_state);
   // Use #after_build to add JS even on validation errors: https://drupal.org/node/1253520#comment-4881588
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -19,7 +19,6 @@ define('DOSOMETHING_CAMPAIGN_PIC_STEP_HEADER', 'Snap a Pic');
  */
 function dosomething_campaign_form_campaign_node_form_alter(&$form, &$form_state, $form_id) {
   unset($form['field_active_hours'][LANGUAGE_NONE]['#options']['_none']);
-  $form['title']['#description'] = t('Title - i.e. "Protect your Pills", "Birthday Mail", or "Teens for Jeans" <br/><strong> Limit: 20 characters. </strong>');
```
itsjoekent, line 22: No idea why I deleted this, re-adding it.

lib/modules/dosomething/dosomething_campaign/js/campaign_node_form.js
```diff
@@ -1,33 +1,47 @@
...
+      if ($('#edit-field-campaign-type-und-sms-game').is(':checked')) {
+        toggleSections(false);
       }
     });
 
+    // Toggle the visibility of field sets in campaignSelections
+    function toggleSections(show) {
+      campaignSections.forEach(function($element, index, array) {
+        if (show) {
+          $element.show();
```
itsjoekent, line 38: I feel like this could be simplified into 

```
/** 
 * state String can be either show or hide
 **/
function(state) { 
   ...
   $element[state]();
   ...
}
```

instead of a show parameter people would pass in a string containing 'show' or 'hide', but not sure if it really matters

DFurnes:
:+1: :ship: :shipit: :anchor: :rocket: :racehorse:

PR #4692, opened 2015-07-16
ESLint

DFurnes:
# Changes

Swaps [ESLint](http://eslint.org) for JSHint, and adopts rules from [AirBnB's JavaScript style guide](https://github.com/airbnb/javascript). This wraps up some discussions made on Documentation Day™.

I've updated a few common files to pass the new linting rules, and added `/* eslint-disable */` directives to the top of less frequented files. Rather than mess with too much working code, we'll "lazy-load" the new code style by removing the ignore directive & formatting a file when it is touched next.
### Why ESLint?

JSHint has been lagging behind with support for new ES6 features, and seems to completely explode when it encounters JSX. Luckily, Airbnb provides their `.eslintrc` with their style guide, so it was a very quick & easy switch to make.
### Thoughts?

This is all just a first pass, so it's 100% open to comment and improvements. I'm sure we'll find things we want to change as we use these new rules & can update them as we go along!


lib/themes/dosomething/paraneue_dosomething/js/campaign/HotModule.js
```diff
@@ -1,3 +1,13 @@
+/* eslint-disable */
```
sbsmith86, line 1: Do we have to disable eslint on any script that isn't written in ES6 / will that mean that these scripts won't be linted at all?

DFurnes, line 1: Nope! I did add the disable flag to a lot of files to save some time upfront, but the idea is that we can just remove that flag as soon as we start modifying a file.

The new linting rules will enforce use of `let` and `const` instead of `var` & disallow old AMD-style requires, but otherwise it's same-old same-old.

sbsmith86, line 1: Gotcha!

sbsmith86:
Looks good to me

DFurnes:
Sweet, merging! :steam_locomotive:

PR #4697, opened 2015-07-16
Fix linting errors in Donate form code.

DFurnes:
# Changes

Fixed up a few minor errors in donation code so we can start linting it.
- Remove side-effects (event binding) from constructor and move into an `init()` method for the `StripeForm`. This fixes the code smell of [using the constructor without storing the result](http://eslint.org/docs/rules/no-new.html).
- Allow the selector for donation form to be overridden as a parameter to `init($donateForm)`, rather than hard-coding in this file. This follows the same convention we're using elsewhere.


weerd:
:+1:

PR #4704, opened 2015-07-17
Campaign class bug fix

weerd:
Quickie fix based on a warning in Drupal error logs. If only 1 tag assigned to a campaign, the `foreach` would crap out because it was not an array and simply a string. So the returned data is now cast to an array to allow foreach to run fine even if a single item.

Also some comment cleanup and updates.

@aaronschachter


aaronschachter:
:+1:

PR #4705, opened 2015-07-20
Adds register & login custom events

### LINKED ISSUE #4589: Add register & login custom events
We want to trigger a custom event in Google Analytics on the subsequent page load after registering or logging in. :chart_with_upwards_trend: 

/cc @mikefantini @Bladt @namimody

itsjoekent:
None of the signup / login functions return anything so you can't add a drupal_add_js there (also the redirects can break add_js). To workaround this I added into the user session whether they were loggin in or registering. Then in the preprocess I add the Google Analytics JS w/ the value of that session var (and unset it).

Fixes #4589 

@DFurnes @angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -15,6 +15,10 @@ include_once 'dosomething_user.theme.inc';
  */
 function dosomething_user_preprocess_page(&$vars) {
   $school_api_endpoint = variable_get('dosomething_user_school_api_endpoint', 'http://lofischools.herokuapp.com/search');
+  if ($_SESSION['log_login']) {
+    dosomething_helpers_add_analytics_event("login", $_SESSION['log_login']);
+    unset($_SESSION['log_login']);
+  }
   drupal_add_js(
     array('dosomethingUser' =>
       array(
@@ -228,11 +232,12 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
       // Helper text & additional data.
       _dosomething_user_add_signup_data($form);
       $form['#submit'][] = 'dosomething_user_login_submit';
-
+      $_SESSION['log_login'] = "login";
     break;
 
     case 'user_profile_form':
     case 'user_register_form':
+      $_SESSION['log_login'] = "register";
       // Gather helper text.
       _dosomething_user_register_helper_text($form);
       // Conditionally displays certain fields based on configuration settings.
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -15,6 +15,10 @@ include_once 'dosomething_user.theme.inc';
  */
 function dosomething_user_preprocess_page(&$vars) {
   $school_api_endpoint = variable_get('dosomething_user_school_api_endpoint', 'http://lofischools.herokuapp.com/search');
+  if ($_SESSION['log_login']) {
```
angaither: I think @mshmsh5000 @sheydari-ds or @blisteringherb might have some feelings on how this works with fastly

DFurnes:
Seems good to me. Not sure if there's a more Drupal-y way of pushing & popping "flash" data to the session, or if we want to add one to `dosomething_helpers` if not?

lib/modules/dosomething/dosomething_user/dosomething_user.module
itsjoekent: interesting, didnt consider that

mshmsh5000: I'm not sure exactly what `log_login` tracks -- any comments we can add here? -- but the basic rule is that, when external cache is present, un-sessioned (unauthenticated) requests will almost never reach the back end.

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -228,7 +232,7 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
       // Helper text & additional data.
       _dosomething_user_add_signup_data($form);
       $form['#submit'][] = 'dosomething_user_login_submit';
-
+      $_SESSION['log_login'] = "login";
```
blisteringherb: Can we call this something like dosomething_user_log_login, just so that we know it's something we've put in via a dosomething module?

itsjoekent: yep good idea

lib/modules/dosomething/dosomething_user/dosomething_user.module
itsjoekent: only authenticated users will ever have this so _shouldn't_ be an issue

itsjoekent:
@DFurnes from what I've been reading this seems to be the most drupal-y way

angaither:
:+1:

PR #4708, opened 2015-07-21
Updates to bring kudos in line with standard class code.

weerd:
# Update

Fixing up the Kudos class to implement similar methods across classes (like those in Reportback, ReportbackItem, etc) to make approaches for retrieving data more standardized and easy! 

Also fixes up query methods to use a much more efficient approach and minimize the number of SQL queries from multiple (depending on number of Kudos items) down to 1 or 2 queries total per request.

@aaronschachter @angaither @DFurnes 

CC: @uy @mshmsh5000 @deadlybutter

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -27,51 +27,6 @@ abstract class Transformer {
 
 
   /**
-   * Get all kudos from specified array of ids.
-   *
-   * @param $ids array
-   * @return array|null
-   * @deprecated since version... because I said so!
-   * @TODO: Remove once method no longer in use.
-   */
-  protected function getKudos($ids) {
-    if ($ids) {
-      $kudos = [];
-
-      foreach((array) $ids as $id) {
-        $kudos[] = Kudos::get($id);
-      }
-
-      return $kudos;
-    }
-
-    return NULL;
-  }
-
-
-  /**
-   * Get all kudos for specified reportback item id.
-   *
-   * @param $reportback_item_id string
-   * @return array|null
-   */
-  protected function getKudosByReportbackItemId($reportback_item_id) {
-    $filters = [
-      'fid' => $reportback_item_id,
-    ];
-
-    // Get array of all kudos ids for specified reportback item.
-    $results = dosomething_kudos_get_kudos_query($filters);
-
-    if ($results) {
-      return $this->getKudos($results);
-    }
-
-    return NULL;
-  }
-
-
-  /**
    * Get the URI for the next page in the collection of data.
    *
    * @param array $data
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -435,10 +390,15 @@ protected function transformReportbackItem($data) {
       'created_at' => $data->created_at,
     ];
 
-    // Get kudos data from comma separated values in data, if any.
-    // @TODO: Implement new method of retrieving Kudos; Kudos::get([$ids])
-    $kudos = $this->getKudos($data->kudos);
-    $output['kudos']['data'] = $kudos ? dosomething_kudos_sort($kudos) : [];
+    try {
+      $kudos = Kudos::get($data->kudos);
```
DFurnes, line 394: Would it make more sense to have `Kudos::get` just return an empty array here if no kudos exist, rather than throwing and then catching an exception?

weerd, line 394: Yeah, I was debating on that... was wondering what others thought.

weerd, line 394: I can update it.

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -57,15 +56,15 @@ function dosomething_kudos_build_kudos_query($params) {
 
 /**
  * @param array $params
```
angaither: can we get some example params in this here comment!

lib/modules/dosomething/dosomething_kudos/includes/KudosTransformer.php
```diff
@@ -3,31 +3,28 @@
...
-   * @param $parameters
+   * Display collection of the specified resource.
    *
+   * @param array $parameters Filter parameters to limit collection based on specific criteria.
    * @return array
    */
   public function index($parameters) {
     $filters = [
-      'reportbackitem_id' => $this->formatData($parameters['reportbackitem_ids']),
+      'fid' => $this->formatData($parameters['reportbackitem_ids']),
```
angaither, line 13: weren't we moving away from using drupaly id's?

weerd, line 13: We are, but this is what gets passed into the query, which involves the column names etc all still containing drupal ids. I've attempted to find a balance that makes sense as to where I need to switch back to Drupal ids... I could go a step farther and it only switches back to Drupal IDs at the point the query is built in https://github.com/DoSomething/dosomething/pull/4708/files#diff-2af8eaa58f7c9cf6c95554e8bb26bdfeL42

lib/modules/dosomething/dosomething_api/includes/Transformer.php
weerd, line 394: @DFurnes and I have decided to leave as is for now and maybe update down the line if there is a specific need.

lib/modules/dosomething/dosomething_kudos/includes/KudosTransformer.php
angaither, line 13: got it got it

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
weerd: oh! sure. forgotz

lib/modules/dosomething/dosomething_api/includes/Transformer.php
DFurnes, line 394: :+1:

angaither:
:+1:

PR #4709, opened 2015-07-21
Initial stab at a React refactoring of the School Finder.

DFurnes:
# Prototype: Don't merge!

This is an initial stab at refactoring the School Finder to be built in [React.js](http://facebook.github.io/react/). React is great for the school finder, because it helps us render the interface without constantly juggling state transitions with messy jQuery code.

I also started to build a little "component library" of React wrappers for some of our existing patterns that could easily be reused throughout the codebase. These components could also easily be something we bundle with Neue in the future.

This is _not_ functionally complete, but the basics are there so feel free to take a look see! :smile: 

/cc @DoSomething/front-end


angaither:
:hocho: for the stabbing

DFurnes:
This is likely never happening. :wave:

PR #4711, opened 2015-07-21
fix date bug

### LINKED ISSUE #4707: Hot module displays even when date is set to not now
The hot module displays, on campaigns when the date is set always. It should only show up when the date is set to the current timeframe

https://trello.com/c/F2AuCAGV/78-the-module-hot-does-not-work-for-some-campaigns

chloealee:
Fixes #4707 

Fix date to only enable hot module if today is between high season start and end dates.


angaither:
:+1: awesome!

Issue #4610, opened 2015-06-23
adding UTM parameters to internal links


mikefantini:
we're looking for all links to pages on the site to be appended by [UTM parameters](https://support.google.com/analytics/answer/1033867?hl=en). this will allow us to do better analysis on the efficacy of various marketing efforts (both external and internal).

this issue relates specifically to links from one page on the site to another (i.e. homepage -> campaign page).

can we append a convention like 
?utm_source=dosomething&utm_medium=site&utm_campaign=node_1141 to the end of internal links?

this convention would follow the following pattern:
- utm_source=dosomething
- utm_medium=site
- utm_campaign = node_nid

I can think of three cases currently on user facing pages that have internal links that don't have nid - profiles, /campaigns, and search. In this case we could go with utm_campaign=profile, utm_campaign=campaigns, utm_campaign=search. Not sure if 404 pages could/would use the nid of the page_not_found, but it currently does not. @Bladt for these edge cases just knowing that they come from the profile and search gives us all the info we really need right? another approach would be to have it be profile_uid (i.e. profile_329195) and search_search_term (i.e. search_bullying) but I think that will just make the more obvious question we'd want to answer harder to get at.

are there other examples where we have user facing pages that have internal links that don't have an nid?

another question is how this should effect the source field we store on the signups table both moving forward and retroactively?

also potentially relevant is seeing [other examples](https://docs.google.com/a/dosomething.org/spreadsheets/d/1O02BK6D6ZwRgweu6N_5mMspW2lLRYz4KOt9iEQoind0/edit?usp=sharing) of how we'll be using UTM parameters with partners.

@Bladt did I miss anything here?

please include thoughts @aaronschachter @angaither @sergii-tkachenko @mshmsh5000 @Bladt @joshcusano 

related to #4575

Bladt:
@mikefantini The scholarships page and campaign collection pages would also be internal links to campaigns. The collections have NIDs, right? But not about/scholarships?

Neither of these cases are pressing. We can, at present, track conversions tied to collections and /scholarships by looking at them as landing pages. E.g. we are currently tracking which campaigns visitors sign up for if they enter on the /scholarship page.

aaronschachter:
Yes this is what the `source` column on the Signups table was aiming to capture. The query string param is appended via code. For scholarships, we could add it if the scholarships page displayed scholarship campaigns dynamically (see #2825).

[...]

PR #4712, opened 2015-07-21
Hot Module: Social Stuff

### LINKED ISSUE #4447: Goal share buttons
Add twitter, facebook, and tumblr share buttons to the campaign template that share the goal statement.

sbsmith86:
This PR fixes #4447 and addresses #4610.
- Adds goal share buttons to the hot module with a message callout. 
- Adds a helper function `dosomething_helpers_utm_parameters` that creates a query string of utm parameters that can be customized that can then be attached to any url.
- Addes a helper function `dosomething_helpers_utm_share_paths` that creates an array of paths to be shared that have the correct UTM parameters attached for each desired social network.
- Creates the intent links for the facebook, twitter, and tumblr buttons and passes them as variables to the template (NOTE: added a todo to move this work into another helper function since we now are doing it in a couple of places

@angaither @DoSomething/front-end

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -221,7 +221,7 @@ function dosomething_campaign_preprocess_hot_module(&$vars, &$wrapper) {
     'picture' => $vars['hero_image']['desktop'],
     'caption' => $goal_copy . "? Let's make it happen!",
   );
-  $vars['hot_module_fb_link'] = dosomething_helpers_get_facebook_intent('feed_dialog', $share_paths['facebook'], $fb_options);
+  $vars['hot_module_fb_link'] = dosomething_helpers_get_facebook_intent($share_paths['facebook'], 'feed_dialog', $fb_options);
 
   // Twitter
   $tweet = "Be a part of @dosomething reaching " . $goal_copy . "- who's ready to rumble?! " . $share_paths['twitter'];
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -204,27 +204,34 @@ function dosomething_campaign_preprocess_hot_module(&$vars, &$wrapper) {
   $vars['author_name']  = $first_name . ' ' . $last_name;
   $vars['author_title'] = check_plain($wrapper->author->field_job_title->value());
   if($author->picture) {
-      $vars['author_image'] = dosomething_image_get_themed_image_by_fid($author->picture->fid, '300x300'); 
+    $vars['author_image'] = dosomething_image_get_themed_image_by_fid($author->picture->fid, '300x300');
   }
 
   // Share links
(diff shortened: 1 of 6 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -117,9 +117,10 @@ function dosomething_campaign_preprocess_hot_module(&$vars, &$wrapper) {
   $end_date = new DateTime($node->field_high_season[LANGUAGE_NONE][0]['value2']);
   $difference = $today->diff($end_date, true);
   $days_left = $difference->d;
+  $readable_end_date = $end_date->format('F') . ' ' . $end_date->format('j') . $end_date->format('S');
```
angaither, line 120: I could see this being a useful function

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -203,8 +204,37 @@ function dosomething_campaign_preprocess_hot_module(&$vars, &$wrapper) {
...
+
+  // Twitter
+  $tweet = "Be a part of @dosomething reaching " . $goal_copy . "- who's ready to rumble?! " . $share_paths['twitter'];
+  $vars['hot_module_tw_link'] = dosomething_helpers_get_twitter_intent($tweet);
+
+  // Tumblr
+  $tumblr_options = array(
+    'posttype' => 'photo',
+    'content' => $vars['hero_image']['desktop'],
+    'caption' => "Think you can make @dosomething reach " . $goal_copy . "? Prove it: " . $share_paths['tumblr'],
```
angaither, line 234: maybe we should make variables for some of this copy?

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.share.inc
```diff
@@ -26,7 +26,7 @@ function dosomething_helpers_share_redirect() {
  *
  * @return string   The facebook intent url.
  */
-function dosomething_helpers_get_facebook_intent($link, $type = NULL, $custom_options = NULL) {
+function dosomething_helpers_get_facebook_intent($type = NULL, $link, $custom_options = NULL) {
```
angaither: let's switch type and link around to the way it was originally
you want to leave arguments with defaults at the end, so you can leave arguments with defaults out of a func call

sbsmith86: @angaither ah ok. I thought this would be more readable but that makes more sense.

angaither:
:cool: the other comments are just suggestions and not needed for merge, the only thing needed is switching the arg order

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
sbsmith86, line 234: @angaither hmm maybe, not sure how much it changes! I might wait to ask Julie and fantini about that.

angaither, line 234: yeah, for sure not something we have to do for merge, but I know copy on twitter looks stale if everyone is tweeting the same thing

PR #4715, opened 2015-07-21
Campaign class update

weerd:
# Update

Final cleanup to `Campaign` classes, similar to recent update to `Kudos` classes. This brings standard methods for data access so all (Campaign, Kudos, Reportback & ReportbackItem) work in a very similar fashion.

Also includes some quick comment cleanup :)

@DFurnes @aaronschachter @angaither 

CC: @jonuy @mshmsh5000 @deadlybutter


DFurnes:
Seems good to me! :+1:

PR #4716, opened 2015-07-21
Hot module cross browser

### LINKED ISSUE #4695: Graph: Cross Browser Issues
The graph is showing up as blank in IE8, 9, and Safari. 

See trello for screenshots: https://trello.com/c/v4P7UsBb/77-hot-module-looks-different-in-different-browsers

sbsmith86:
## Addresses #4695
- Updated the format of the dates passed to the front end because the The pattern `yyyy-MM-dd` that being returned from PHP isn't an officially supported format for Date constructor in JS. This was causing some `Invalid Date` errors in Safari and IE9 that was stopping the graph from being rendered.
- Added a check for the support of the `setLineDash()` HTML5 Canvas function which is only supported in IE10+. So now in IE9 the goal line will be solid, not dashed. 

@DoSomething/front-end


weerd:
Cooool :+1:

PR #4717, opened 2015-07-22
Fixes minor time left bugs

### LINKED ISSUE #4714: Display of readable timing goal is off
The end date that is set is displaying incorrectly on the hot module front end

see: https://trello.com/c/uxM85STK/80-the-timing-is-off

itsjoekent:
Fixes #4714 

Issue has me blabbering about how I discovered the bugs and what I took care of. (including a very weird error Im encountering trying to save the timing information). 

@angaither


angaither:
:+1: 

Looks ok, but can you make a new issue with the bug you discovered, so we have a history of it. 
I don't think it's a big deal since I don't think the end date will be updating, but for SAFTEY.

Issue #4529, opened 2015-05-22
Reportback flagged status


angaither:
It seems like some flagged reportback items aren't getting set as flagged in the db.

This query on prod returns 2,257 rows
the file is set as flagged, but the reportback isn't 

```
SELECT rb.rbid, rb.nid, rb.flagged, rbf.status, rb.flagged_reason, FROM_UNIXTIME(rbf.reviewed)
FROM dosomething_reportback as rb
INNER JOIN dosomething_reportback_file as rbf on rb.rbid = rbf.rbid
WHERE rbf.status = 'flagged'
AND (rb.flagged = 0 or rb.flagged is null);
```

every test I've done locally (flagging with reasons/without reasons) correctly sets the status from null->0

@aaronschachter are you able to replicate this issue?

mikefantini:
@namimody

angaither:
This has been happening since 1/27/2015
it seems to be happening with more regularity lately, but still haven't been able to reproduce it.

mikefantini:
@angaither did anything change on 1/27/2015? it's also possible that that's just when the campaigns team started really reviewing reportbacks

angaither:
@mikefantini Yeah, the [commits](https://github.com/DoSomething/dosomething/commits/dev?page=35) during that time are all related to setting up the reportback stuff, so it looks like that's when reviews started

mikefantini:
3193 rbs were flagged but are showing up as 0 or null in the reportback_flagged column of the table.

aaronschachter:
Linking #4621

mikefantini:
@mshmsh5000 assigning over to you

angaither:
on staging we get 4 results 

| rbid | nid | flagged | status | flagged_reason | FROM_UNIXTIME(rbf.reviewed) |
| --- | --- | --- | --- | --- | --- |
| 9 | 43 | 0 | flagged | NULL | NULL |
| 184 | 1261 | 0 | flagged | NULL | NULL |
| 290 | 15 | 0 | flagged | NULL | NULL |
| 1292 | 1159 | 0 | flagged | irrelevant | 2015-05-14 18:46:25 |

angaither:
We have recreated the issue!
On staging, @mikefantini and I went though a page of 25 or so that were in the inbox, and reviewed them all at once, after hitting save, there are now 7 reportbacks on staging where the file status is `flagged`, and the flagged boolean is still 0. 

[...]

PR #4719, opened 2015-07-22
Added update query to set flagged status to 1

### LINKED ISSUE #4658: promoted reason/flagged not being saved
- [reportback in question](https://www.dosomething.org/admin/reportback/48592) - RBID 48592
- [video](http://screencast.com/t/hkdyVf1Zze2x) watch at 0:21
- [campaign](https://www.dosomething.org/campaigns/mirror-messages) - NID 7

potentially related to #4529

### LINKED ISSUE #4713: write a script that retroactively fixes flagging issue
Refs #4529 

For the hot module write a script, or a cron, or a jenkins job that goes checks the flagged status and fixes it.
We can do this to only monitor the 4/5 campaigns that are in the hot module or, just all the campaigns. Probably makes more sense to do it for all of them.
This kind of papers over the issue of why it's happening, and we wouldn't have examples anymore... but it should retroactively "fix" the issue. 

cc @mshmsh5000 @sheydari-ds @julielorch @mikefantini

angaither:
Fixes #4713
Refs #4529 

This will fix numbers in the dashboards, and local variables for the hot module.
This doesn't fix #4658

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.cron.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.cron.inc
@@ -33,4 +33,17 @@ function dosomething_reportback_cron() {
       ->execute();
   }
 
+
+  /**
+   * Search for reportback item that have been flagged, but the reportback was not flagged.
+   * @see https://github.com/DoSomething/dosomething/issues/4713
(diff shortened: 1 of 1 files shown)


aaronschachter:
:+1:

PR #4721, opened 2015-07-23
Fixes old typo that caused user var not to get set.

angaither:
[Remember when you had me remove `$whatever`](https://github.com/DoSomething/dosomething/pull/4640#discussion-diff-33679755)
I only half did that. This is the other half.


blisteringherb:
:+1:

PR #4722, opened 2015-07-23
Adds promoted/flagged reasons to the reportback log

### LINKED ISSUE #4658: promoted reason/flagged not being saved

itsjoekent:
Fixes #4658 
- Adds new column to the table 
- Adds the flagged or promoted reason
- Also fixes a bug that was partially still remaining from #4621 where the reason stays in the reportback after you change it to approved

This is what the DB looks like in my dev for reference 

screen shot 2015-07-23 at 11 29 02 am

@angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.cron.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.cron.inc
@@ -40,7 +40,7 @@ function dosomething_reportback_cron() {
    * @see https://github.com/DoSomething/dosomething/issues/4529
    */
 
-  $results = db_query("UPDATE dosomething_reportback as rb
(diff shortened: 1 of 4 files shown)


angaither:
I also added [this cron](https://github.com/DoSomething/dosomething/blob/dev/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.cron.inc#L43) yesterday. Can you update it so it also saves flagged/promoted reason if needed.

Refs #4658

PR #4725, opened 2015-07-23
Fixing small bug when generating URI for single resource and adding URI property to other resources.

weerd:
# Update

This PR fixes a bug with how the `uri` property on some resources was being output. There was a wrong reference to the `id` and thus the `uri` value was not being generated correctly. This PR also adds the `uri` property to the rest of the `Campaigns, Kudos, Reportbacks, ReportbackItems` resources so all have data regarding how to access them via their respective individual URIs.

The PR also does some extra cleanup and adds some indentation formatting to the _ReportbackController.php_ file which somehow was lost before.

@aaronschachter @DFurnes

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -293,6 +293,30 @@ abstract class Transformer {
 
 
   /**
+   * Transform Kudos data and prepare for API response.
+   *
+   * @param object $data A Kudos object containing the following properties:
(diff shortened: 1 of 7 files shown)


aaronschachter:
:+1:

weerd:
hehe, I'm working on one more update to this PR and then will request another review @aaronschachter :) Sorry, forgot to mention it was still a work in progress! But thanks for the initial check :dancer: Have the code review is completed now :)

DFurnes:
:+1:

PR #4726, opened 2015-07-23
DoSomething Global module.

sergiitk:
#### What's this PR do?
- Provides feature-based `dosomething_global` module dedicated for translation and internationalization functionality
- Installs Brazilian Portuguese language via `dosomething_global` module

image

Note: module is not enabled by default. I will provide special build argument to do so.


angaither:
I think you also need to add this to `dosomething.info`

angaither:
But I guess that would enable it... so maybe not?

sergiitk:
Yes it would. I intentionally kept it out of the info file.

angaither:
:+1: :ship:

PR #4733, opened 2015-07-23
Graph: Scale math

### LINKED ISSUE #4727: Graph: Goal line leaves chart
Removing the y-axis scale messed up where the goal line gets placed on the graph. If the highest data point is no where near the goal, then the graph doesn't scale to show the goal line. 

Probably need some scaling math that takes into consideration the highest data point so far and scales based on that and the goal.

sbsmith86:
So, I removed the scale math in a previous pull request, but this caused the graph to not scale correctly when the goal was way higher than the highest quantity amount. 

This PR updates the logic around building the y-axis scale so that, if we are pass the goal, we should scale the goal based on the highest data point. But, if we aren't at the goal, the goal point is the highest point on the graph and we should scale around that instead. Fixes #4727 

@DoSomething/front-end


DFurnes:
:+1:

PR #4734, opened 2015-07-24
Select the max total from the max timestamp.

### LINKED ISSUE #4730: Hot module json not sending correct max
The query that selects the max date from the hot module is actually not returning the most recent timestamp

```
 SELECT quantity, max(from_unixtime(timestamp)) as maxdate
 FROM dosomething_reportback_progress_log
 WHERE nid = :nid
 GROUP BY date(from_unixtime(timestamp)); 
```

angaither:
Fixes #4730 

d0536065bf0


cjcodes:
Worth noting that this is dependent upon an ever-increasing value in the quantity column.

In Engrish, this does: `SELECT the max quantity FROM each day GROUPED BY the day`
**NOT**: `SELECT the last quantity of a day GROUPED BY the day`

The latter would require a subquery and is thus not as efficient.

:+1: :shipit:

PR #4737, opened 2015-07-24
adding all new mobile app fields

### LINKED ISSUE #4723: Add date fields to campaign template for mobile app support
The mobile app needs campaigns to show up for a month, so it needs a specific start/end date. This doesn't alway align with (and will conflict) with the hot module if we use the high season field.

Steps: 
- [x] Add `field_mobile_app_date` - this should have both a start and end date
- [x] Create a new field group called `Mobile App`
- [x] Move the checkbox for `field_on_mobile_app` into the mobile app field group
- [x] Add the `field_mobile_app_date` to the api response for a campaign

@chloealee Once you've added the field to the node, you will have to export them using drupal features!
Also, the fields will need to be exposed in the api @weerd has been working on

chloealee:
Addresses most of the issues in issue #4723 

added all new mobile app fields and moved `field_on_mobile_app` with @weerd !


angaither:
One small note, update the feature to say the mobile app is collapsed automatically!

weerd:
I thought that was a jQuery specific thing that @aaronschachter  added to help make it more readable?

angaither:
:+1:

PR #4742, opened 2015-07-27
Creates ds-global script.

sergiitk:
#### What's this PR do?

Creates a bash script for setting up Global:
- Enables `dosomething_global` module
- Serves as a placeholder for additional instructions and migration commands which must be applied to existing database to set it up as Global


blisteringherb:
Shouldn't the dosomething_global module be enabled in the dosomething profile? Or is this just a temporary stopgap until we've moved everything to global?

sergiitk:
@blisteringherb Yes, it should be eventually. However, if I think it is reasonable to keep Global functionality out of production for some time. There will be major changes, like URL patterns and multiple languages that can break existing US user experience.

Using this script we will be able to keep track of required steps to make US site Global. It is useful both on Global test environment @sheydari-ds is working on and inside vagrant, after `ds pull stage`.

blisteringherb:
:+1:

PR #4754, opened 2015-07-28
creates convertDate helper function and adds field_mobile_app_date to…

### LINKED ISSUE #4723: Add date fields to campaign template for mobile app support

chloealee:
… api response for a campaign.

Fixes #4723

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
+++ b/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
@@ -145,8 +145,7 @@ class Campaign {
 
         $this->services = $this->getServices();
 
-        $mobile_app = $this->getMobileAppDate();
-        $this->mobile_app = $mobile_app;
+        $this->mobile_app = $this->getMobileAppDate();
       }
 
       $this->reportback_info = $this->getReportbackInfo();
--- a/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
+++ b/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
@@ -374,12 +374,12 @@ class Campaign {
    */
 
     protected function getMobileAppDate() {
-      $timezone = new DateTimeZone('UTC');
-
       $timing = [];
       $timing['dates'] = (dosomething_helpers_extract_field_data($this->node->field_mobile_app_date));
 
-      dosomething_helpers_convert_date($timing);
+      foreach ($timing['dates'] as $key => $date) {
+        $timing['dates'][$key] = dosomething_helpers_convert_date($date);
+      }
 
       return $timing;
   }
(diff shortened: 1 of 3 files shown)


angaither:
@chloealee looks like this branch needs a rebase against dev!

chloealee:
@angaither thanks for that and for sitting with me to rebase!

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
```diff
@@ -363,6 +366,25 @@ protected function getMobileCommonsData() {
...
 
+   /**
+   * Get the start and end dates for when campaign will be displayed on the mobile app if available.
+   * Dates formatted as ISO-8601 datetime.
+   *
+   * @return array
+   */
+
+    protected function getMobileAppDate() {
+      $timezone = new DateTimeZone('UTC');
```
weerd: I think this line for the `$timezone` can probably just go inside the `dosomething_helpers_convert_date()` function :)

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -50,6 +50,20 @@ function dosomething_helpers_menu() {
 }
 
 /**
+ * Utility function to convert dates to ISO-8601 datetime.
+ */
+function dosomething_helpers_convert_date($timing) {
```
weerd: You can probably change up the variable names in this function now that it's a generic helper function. The `$timing` and `$season` variables were very specific to extracting the `high_season` and `low_season` dates, but if this is now more generic the variables should be as well :)

Might also change name to `dosomething_helpers_convert_date_to_iso()` to make it obvious to what it's converting the date to OR, pass a default parameter that is the format, such as:

```
dosomething_helpers_convert_date($date, $format = 'iso-8601') {
 if ($format === 'iso-8601') {
  ... do yo thang...
 }
}
```

chloealee:
thanks, @weerd ! Updated all the changes we just reviewed together :)

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
```diff
@@ -144,6 +144,9 @@ private function build($data, $display = 'teaser') {
         $this->timing = $timing;
 
         $this->services = $this->getServices();
+
+        $mobile_app = $this->getMobileAppDate();
+        $this->mobile_app = $mobile_app;
```
weerd: You can probably consolidate this down to one thing since you're not doing anything else w/ the `$mobile_app` variable :)

weerd:
One small comment, other than that looks great! :+1:

PR #4755, opened 2015-07-28
Hot module: Add safety checks for null values

### LINKED ISSUE #4751: can't submit initial photo for reportback
http://screencast.com/t/xiDL5vntmP5D

to reproduce:
- go to campaign page
- sign up for campaign
- try to report back
- file upload dialog box opens up, but once you select a file cropping modal doesn't open

only happening on staging. seems to be working fine on production.

also adding another photo to an existing reportback seems to be working fine as well. it's just the initial photo that doesn't seem to be working.

cc: @julielorch

sbsmith86:
## Fixes #4751

Doing some better "defensive coding". A new campaign won't have any data tied to it. In this case, the hot module script was trying to iterate over values that were `null` and this was breaking the js on the page. This adds some checks for `null` values before we try to do things like `forEach()` over an array. And, allows for those functions to return empty arrays, so that the graph will still render, but with no data on it. 

@DoSomething/front-end

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/campaign/HotModule.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/campaign/HotModule.js
@@ -73,13 +73,15 @@ var createDateLabels = function(dates) {
   var dateLabels = [];
   var monthNames = ["Jan", "Feb", "Mar", "April", "May", "June", "July", "Aug", "Sept", "Oct","Nov", "Dec"];
 
-  dates.forEach(function(date, key) {
-    date = new Date(date);
-    var month = date.getMonth();
-    var day = date.getDate();
+  if (dates) {
+    dates.forEach(function(date, key) {
+      date = new Date(date);
+      var month = date.getMonth();
+      var day = date.getDate();
 
-    dateLabels[key] = monthNames[month] + " " + day;
-  });
+      dateLabels[key] = monthNames[month] + " " + day;
+    });
+  }
 
   return dateLabels;
 };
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/js/campaign/HotModule.js
```diff
@@ -73,13 +73,15 @@ var createDateLabels = function(dates) {
...
   var monthNames = ["Jan", "Feb", "Mar", "April", "May", "June", "July", "Aug", "Sept", "Oct","Nov", "Dec"];
 
-  dates.forEach(function(date, key) {
-    date = new Date(date);
-    var month = date.getMonth();
-    var day = date.getDate();
-
-    dateLabels[key] = monthNames[month] + " " + day;
-  });
+  if (dates) {
```
DFurnes, line 66: It might be cleaner to make `dates` have a default value of `[]` so the `.forEach` just becomes a no-op.

``` js
function createDateLabels(dates = []) {
 // ...
}
```

DFurnes:
One (subjective) suggestion, otherwise LGTM!

DFurnes:
At some point, it would be good to remove the `eslint-disable` from this file and start linting it as well.

lib/themes/dosomething/paraneue_dosomething/js/campaign/HotModule.js
sbsmith86, line 66: :+1: oh smart smart.

sbsmith86:
@DFurnes fixed the linting issues!

PR #4757, opened 2015-07-28
Updates reportback quantity logic

### LINKED ISSUE #4748: live query was resetting value set by cron, causing quantity displayed to be off
[Bake Sales for Babies](http://staging.beta.dosomething.org/node/1467/rb/reviewed?status=All&flagged_reason=All&promoted_reason=All&order=quantity&sort=desc) should have a Total Reportback Quantity of 3917 but it's stated as 717. The 717 seems to be excluding [this reportback](http://staging.beta.dosomething.org/admin/reportback/1172) but not sure why since it's set as promoted and doesn't seem to have had any changes to it.

[Baby It's Cold Inside](http://staging.beta.dosomething.org/node/1581/rb/reviewed) has a reportback total of 100 even though there are no reviewed reportbacks. This matches the quantity that is in the inbox.

[Wildlife Cards](http://staging.beta.dosomething.org/node/689/rb/reviewed) has a reportback total of 888 even though there are no reviewed reportbacks. This matches the quantity that is in the inbox.

Found these just reviewing some campaigns that have old reportbacks attached to them so unfortunately not exactly sure how to reproduce. 

However, given these discrepancies, it's worth someone manually checking that the math currently checks out for the Total Reportback Quantity on the campaigns that are going hot  August 1st so that we feel confident that this number is starting off correctly.
- Visionary Volunteer - https://www.dosomething.org/node/2089/rb
- Storm Drain Street Art - https://www.dosomething.org/node/3743/rb
- Justice for Elephants - https://www.dosomething.org/node/1536/rb (is this Lead the Stampede?)
- Get Lucky is a new campaign so we should be fine here

itsjoekent:
Fixes #4748 

[2:41 PM] Joe Kent: and I added a condition to the query in dosomething_reportback_get_reportback_total_by_nid

[2:42 PM] Joe Kent: also made other places DRY by using that function instead of copying the addition code

@angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -738,7 +738,7 @@ function dosomething_reportback_views_pre_render(&$view) {
   // Scope this to the reportback files node view.
   if ($view->name === 'reportback_files' && arg(0) === 'node') {
     $nid = $view->args[0];
-    $quantity = dosomething_reportback_get_reportback_total_plus_override($nid);//1
+    $quantity = dosomething_reportback_get_reportback_total_plus_override($nid);
     // Add total quantity to view.
     $view->attachment_before = '<h2> Total reportback quantity: ' . $quantity . '</h2>';
   }
@@ -891,7 +891,8 @@ function dosomething_reportback_get_reportback_total_by_nid($nid, $type = 'quant
 
   // Select all reportbacks for given $nid.
   $query = db_select('dosomething_reportback', 'rb')->condition('nid', $nid);
-  $query->condition('promoted', 0); // Prevents non-reviewed RB's from being added to total
+  // Prevents non-reviewed RB's from being added to total
+  $query->isNotNull('promoted');
 
   if ($filter_flagged) {
     $query->condition('flagged', 0);
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -738,7 +738,7 @@ function dosomething_reportback_views_pre_render(&$view) {
   // Scope this to the reportback files node view.
   if ($view->name === 'reportback_files' && arg(0) === 'node') {
     $nid = $view->args[0];
-    $quantity = (int) dosomething_helpers_get_variable('node', $nid, 'sum_rb_quantity') + (int) dosomething_helpers_get_variable('node', $nid, 'sum_rb_quantity_override');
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -891,6 +891,7 @@ function dosomething_reportback_get_reportback_total_by_nid($nid, $type = 'quant
 
   // Select all reportbacks for given $nid.
   $query = db_select('dosomething_reportback', 'rb')->condition('nid', $nid);
+  $query->condition('promoted', 0); // Prevents non-reviewed RB's from being added to total
```
angaither: @deadlybutter this will only select reportbacks that aren't promoted though...

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -738,7 +738,7 @@ function dosomething_reportback_views_pre_render(&$view) {
   // Scope this to the reportback files node view.
   if ($view->name === 'reportback_files' && arg(0) === 'node') {
     $nid = $view->args[0];
-    $quantity = (int) dosomething_helpers_get_variable('node', $nid, 'sum_rb_quantity') + (int) dosomething_helpers_get_variable('node', $nid, 'sum_rb_quantity_override');
+    $quantity = dosomething_reportback_get_reportback_total_plus_override($nid);//1
```
angaither: remove that extra `//1`

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
itsjoekent: oh right, can we make it 0 _or_ 1 then?

angaither: You should search for where it is null.
`$query->isNull('promoted')`

Also, code style, the comment should go above the line

itsjoekent: if we search by null then you get reportbacks inside the inbox. thats why I did 0 (and now suggesting 0 or 1) because only reportbacks not in the inbox have one of those values

angaither: sorry sorry, should be isNotNull

itsjoekent: oh, yeah that would work

itsjoekent:
@angaither I think it's good now!

angaither:
:+1:

PR #4764, opened 2015-07-30
fixes invalid argument in foreach block for mobile app date

### LINKED ISSUE #4763: Invalid argument supplied for foreach() in Campaign class
When extracting Mobile App start and end dates, there is a bug when trying to pass through `null` through foreach() block if no dates are set. 

```
Warning: Invalid argument supplied for foreach() in Campaign->getMobileAppDate() (line 379 of /var/www/staging.beta.dosomething.org/releases/20150728191855/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php).
```

chloealee:
Refs #4754 

Fixes #4763

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
+++ b/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
@@ -376,12 +376,10 @@ class Campaign {
       $timing = [];
       $timing['dates'] = (dosomething_helpers_extract_field_data($this->node->field_mobile_app_date));
 
-      if ($timing['dates'] != null) {
+      if (isset($timing['dates'])) {
         foreach ($timing['dates'] as $key => $date) {
           $timing['dates'][$key] = dosomething_helpers_convert_date($date);
         }
-      } else {
-        $timing['dates'] === null;
       }
 
       return $timing;
--- a/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
+++ b/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
@@ -376,8 +376,12 @@ class Campaign {
       $timing = [];
       $timing['dates'] = (dosomething_helpers_extract_field_data($this->node->field_mobile_app_date));
 
-      foreach ($timing['dates'] as $key => $date) {
-        $timing['dates'][$key] = dosomething_helpers_convert_date($date);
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
```diff
@@ -376,8 +376,12 @@ protected function getMobileAppDate() {
       $timing = [];
       $timing['dates'] = (dosomething_helpers_extract_field_data($this->node->field_mobile_app_date));
 
-      foreach ($timing['dates'] as $key => $date) {
-        $timing['dates'][$key] = dosomething_helpers_convert_date($date);
+      if ($timing['dates'] != null) {
```
angaither: does if(isset($timing['dates'])` work here, we tend to use that as the standard check

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
```diff
@@ -376,8 +376,12 @@ protected function getMobileAppDate() {
...
+        foreach ($timing['dates'] as $key => $date) {
+          $timing['dates'][$key] = dosomething_helpers_convert_date($date);
+        }
+      } else {
+        $timing['dates'] === null;
```
angaither: when you're assigning a var, one `=` will do 
do you need this though, if it isn't set shouldn't it already be null?

angaither:
Also, can you ref which issue this is fixing

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
weerd: the method @angaither suggests, `isset()` will actually check if the variable is `null` or even exists. So it might be possible that the variable might not even exist (which is happening in the bug I'm working on right now), and `isset()` helps check for both scenarios :)

chloealee: ahhh got it - way more efficient! Will change it, thank you both!

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
chloealee: true - I will delete!

weerd:
:+1:

angaither:
:+1:

PR #4765, opened 2015-07-30
Fixing bug with helper isset function

### LINKED ISSUE #4759: Undefined index notice in Campaign class
When extracting MailChimp and MobileCommons data, there is a bug with handling `null` when those items aren't present in the Campaign.

```
Notice: Undefined index: mailchimp_grouping_id in Campaign->getMailChimpData() (line 356 of /var/www/dev.dosomething.org/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php).
```

weerd:
### Fixes #4759

There was a bug where if the variables for the `mailchimp` and `mobile_commons` data don't exist in the `$variables` array, then PHP will throw a few notices for an _undefined index_.

This PR updates to use the `dosomething_helpers_isset()` function to pass the data in and output it or `null`.

It also updates the `dosomething_helpers_isset()` function to accept either a string or an array, or object, with optional parameter (`$key`) to pass to specify which item in array or property in object to test.

@DFurnes @angaither @aaronschachter 

CC: @chloealee

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
+++ b/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
@@ -347,8 +347,8 @@ class Campaign {
    */
   protected function getMailChimpData() {
     return [
-      'grouping_id' => $this->variables['mailchimp_grouping_id'],
-      'group_name' => $this->variables['mailchimp_group_name'],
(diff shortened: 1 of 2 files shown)


DFurnes:
:+1:!

PR #4766, opened 2015-07-30
Fixes overridden status of dosomething_campaign_run and dosomething_fact_page features

sergiitk:
Synchronizes DoSomething Features codebase with production.


sergiitk:
The only Feature left is overridden `dosomething_search`, but it requires individual change. I'd recommend first to update the Solr module.

angaither:
:+1:

PR #4769, opened 2015-07-31
Decode twitter text

### LINKED ISSUE #4624: Twitter Permalink Share Shows Jumbled Copy Where Apostrophe Should Be
When a campaign name has an apostrophe, the Twitter copy shared from the permalink shows jumbled copy instead.

image

sbsmith86:
## Fixes #4624

Runs the twitter text we are using for permalink shares through `htmlspecialchars_decode` because `token_replace` sanitizes the string and twitter was displaying the sanitized string to the user.


angaither:
:+1:

PR #4773, opened 2015-08-03
Hot fix for hot module total numbers

angaither:
This `sum_rb_quantity` value is getting set correctly on cron, but when the`dosomething_reportback_get_reportback_total_by_nid` query runs, it removes the value for some campaigns, there's a bug in the query that I will work on later, but this is just to get the hot module live now.

Refs #4748


sergiitk:
:+1:

PR #4776, opened 2015-08-03
Adding Brazil and Mexico admin roles and perms

blisteringherb:
Brazil and Mexco now have separate roles to allow editors from those countries to add content without the ability to publish.  This PR adds the new roles to dosomething_user and updates several of the features that hold the user perms for the functionality they'll need to access.

I didn't put this into the dosomething_global feature because that won't be enabled on production for some time, and these roles need to be used immediately.


angaither:
:+1:

sergiitk:
Awesome @blisteringherb!

PR #4777, opened 2015-08-03
Fix bug in how drupal reads the schema for RB table.

### LINKED ISSUE #4756: Flagged status should be null by default
Before a reportback is reviewed the `flagged` and `promoted` status in the `dosomething_reportback` table should be `NULL`

Not sure what is causing newly submitted reportbacks to have a status of 0 by default, but that shouldn't be the case

weerd:
### Fixes #4756

Drupal core's `drupal_get_schema()` function seems to retrieve the original schema for an entity and not the updated schema which proved to be a problem with the **flagged** status.

@angaither 

CC: @DFurnes

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
@@ -61,8 +61,13 @@ function dosomething_reportback_schema() {
       'flagged' => array(
         'description' => 'Whether the Reportback has been flagged.',
         'type' => 'int',
-        'not null' => TRUE,
-        'default' => 0,
+        // The following field settings were edited specifically to set 'not null' to FALSE and 'default' to NULL.
+        // While dosomething_reportback_update_7024() already updates this, the drupal_get_schema() function
+        // seems to pull from this original schema instead of the updated one and this was causing problems with
+        // Reportback flagging status.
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
```diff
@@ -61,8 +61,13 @@ function dosomething_reportback_schema() {
       'flagged' => array(
         'description' => 'Whether the Reportback has been flagged.',
         'type' => 'int',
-        'not null' => TRUE,
-        'default' => 0,
+        // The following field settings were edited specifically to set 'not null' to FALSE and 'default' to NULL.
+        // While dosomething_reportback_update_7024() already updates this, the drupal_get_schema() function
+        // seems to pull from this original schema instead of the updated one and this was causing problems with
+        // Reportback flagging status.
+        // @see https://github.com/DoSomething/phoenix/issues/4756
```
angaither, line 68:
```
/**
 *
 */
```

nit-picky, but should be block comments

angaither:
other than that :+1:

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
angaither, line 68: I was wrong!

PR #4778, opened 2015-08-03
Adding SEO changes for front page and campaigns pages

blisteringherb:
- Source doc here - https://goo.gl/MzYSo9
- Enabling metatag_views module
- Exporting metatag node and global settings to metatag module

@angaither


angaither:
@blisteringherb add the module to `drupal-org.make` file too https://github.com/DoSomething/phoenix/blob/dev/lib/profiles/dosomething/drupal-org.make

blisteringherb:
The metatag_views module is part of the metatag module package so it doesn't get downloaded separately.  Does it still need to be added to the drupal-org.make file?

angaither:
ahh, i see :+1:

PR #4781, opened 2015-08-03
Custom GA events for hot module shares

### LINKED ISSUE #4644: custom GA event for hot shares
needed for #4447.

related to #4544

sbsmith86:
Fixes #4644 - Add custom click events for Hot module shares. 

@DoSomething/front-end - Also, wondering if it is worth discussing potential refactoring of how analytics is implemented. It seems weird that every new thing we want to track needs it's own hard coded event. This file could potentially end up growing bigger and bigger. One idea could be coming up with some standard class names and data attributes to use and then creating a function that sends listens to click events on that class and sends different info to GA based on what data attributes it has?


DFurnes:
Yeah, 1000% agreed. :+1: to ship this for now though!

PR #4768, opened 2015-07-31
Enables simple campaign translation for Global project

sergiitk:
#### What's this PR do?
- Installs [Entity Translation](https://www.drupal.org/project/entity_translation) and [Title](https://www.drupal.org/project/title) module
- Replaces Campaign's title with translatable entity field
- Makes "Call to action" field translatable

CC @sheydari-ds

PR #4784, opened 2015-08-04
Installs translation modules

sergiitk:
This PR just downloads [Entity Translation](https://www.drupal.org/project/entity_translation) and [Title](https://www.drupal.org/project/title) module
Modules will be installed and enabled as a dependency for DoSomething Global feature module.
On the contrary to #4768, it will not affect US users.


sergiitk:
I updated Title module to dev branch because of an issue with table layout on content type manage fields page.
Here's alpha:

image

And dev:

image

angaither:
:+1:

PR #4791, opened 2015-08-04
Give not null a default, causing sql errors.

### LINKED ISSUE #4785: Reportback change log isn't displaying
http://staging.beta.dosomething.org/admin/reportback/2873

There's no change log displaying for recently submitted reportbacks

angaither:
Re-run reason update, since it was never added.
Fixes #4785


DFurnes:
Worked! :+1: 

Got this error, but seems unrelated. Will add to another issue.

```
SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'node-4-sum_rb_quantity' for key 'PRIMARY'
```

Issue #4797, opened 2015-08-05
reportback inbox submission query retrieval issue


mikefantini:
watched Jess review 25 FIDs on node 2805. here is [a doc](https://docs.google.com/a/dosomething.org/spreadsheets/d/1R1EOY3TEI1_eJ_otK6DftkTicaTQeNBf7g4R5t8zYqc/edit?usp=sharing) tracking what took place in terms of which FIDs remained in the inbox, which went into the reviewed tab, and which were never actually reviewed but went into the reviewed tab anyway.

mikefantini:
this is happening because when an editor clicks save on the inbox we think it is querying for the most recent files while the form is being submitted, instead of just the data that was submitted (i.e. the 25 or less FIDs that were in the inbox)

steps to replicate:
1. load rb inbox
2. have a user load up a campaign
3. have that user submit a rb
4. wait until rb is submitted
5. review rbs that are in the inbox

when you do this if all rbs that were reviewed (if <25 total) will have the proper status, but the new rb that was submitted will also appear to be reviewed as approved. 

if we don't have approved as default selected in the inbox, and there are less than 25 fids in the inbox, then the user will get an error that he/she made an illegal choice (because he or she would have tried to submit the new fid without selecting a status)

we are changing the order from newest to oldest in the inbox and removing the default approved selected in the inbox to unblock the reviewing of campaigns.

and @weerd is going to try to debug to determine why when an editor clicks save on the inbox it is querying for the most recent files while the form is being submitted, instead of just the data that was submitted

cc: @angaither

mikefantini:
see related slack convo: https://dosomething.slack.com/archives/phoenix/p1438966108001280

mikefantini:
even though user gets errors now, the reviewed rbs still go through properly.

error message user gets just to be clear (happening on node 5459, 25 errors because the inbox is full I imagine):

![image](https://cloud.githubusercontent.com/assets/6330971/9205335/9974a71c-402f-11e5-8e7e-1a04d4cd9f44.png)

mikefantini:
to be clear, when we submit the inbox, it's always trying to grab additional reportbacks outside of the 25 that the user is actually reviewing. a key distinction based on changes that we made:
- when we had reportbacks listed by newest in the inbox, it was attempting to review any reportbacks that were submitted between the time a campaign lead loaded the inbox and submitted it
[...]

PR #4800, opened 2015-08-05
Added new param to files query, can order by asc/desc

angaither:
Reversed the order of the inbox.
Refs #4797


weerd:
Oh man this looks beautiful! :dancer: :+1:

PR #4802, opened 2015-08-05
Don't pre-select approved on rb file reviews.

angaither:
Refs #4797


weerd:
Holy bugs Batman! :bug: :+1:

PR #4814, opened 2015-08-06
Reportback data cleanup.

### LINKED ISSUE #4779: clean up reportback file flagged status
Need to write an update hook that searches for any file/reportback that's pending and updates the flagged status from `0->null`

this could affect the total campaign progress number
Refs #4756

angaither:
Created these two queries as a function in case we need to re-run them again. 

Fixes #4779


weerd:
Looks good! :+1: :sake:

PR #4840, opened 2015-08-13
Adds progress copy override

### LINKED ISSUE #4638: progress copy override
need a field added to the campaign template to override the default progress update

Suggestion: add this field to custom settings under the Goal Settings group with the name "Hot Phase Progress Copy Alternate" help text "Add copy here if you'd like to override the default progress copy in the hot module. Note: The default copy changes as progress increases. Override copy is static. If you create override copy, you'll need to either delete the copy or change it yourself for this text to update."

this would override the copy in #4446 and again is the copy in the speech bubble:

image

itsjoekent:
Fixes #4638 
- Adds progress copy override admin field
- Updates progress copy logic 

@angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -204,7 +204,7 @@ function dosomething_campaign_preprocess_hot_module(&$vars, &$wrapper) {
   if (isset($override_copy)) {
     $vars['progress_copy'] = $override_copy;
   }
-  else if ($rb_percent <= 20) {
+  elseif ($rb_percent <= 20) {
     $vars['progress_copy'] = variable_get('dosomething_campaign_0_20_progress_copy');
   }
   elseif ($rb_percent <= 40) {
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -200,7 +200,11 @@ function dosomething_campaign_preprocess_hot_module(&$vars, &$wrapper) {
     $rb_percent = 0;
   }
 
-  if ($rb_percent <= 20) {
+  $override_copy = variable_get('dosomething_campaign_progress_copy_override');
+  if (isset($override_copy)) {
+    $vars['progress_copy'] = $override_copy;
+  }
+  else if ($rb_percent <= 20) {
```
blisteringherb: Since 'elseif' and 'else if' aren't evaluated exactly the same in all contexts, and it's 'elseif' on line 206 below, it probably makes sense to make this 'elseif' for consistency.

http://php.net/manual/en/control-structures.elseif.php

blisteringherb:
:+1:

PR #4852, opened 2015-08-17
Adds review fix

### LINKED ISSUE #4847: two files from same rb (one flagged one approved) causes incorrect flagged status
Steps to recreate:
- submit reportback
- add another image to the same reportback
- as admin, flag the first file
- as admin, approve the second file
- save the inbox
- check the db and the first file will appear to have status = flagged but flagged status = 0

thanks @DFurnes for figuring this out you goddamn champion

refs #4529 

key query:

```
SELECT rb.rbid, rb.nid, rb.flagged, rbf.status, rb.flagged_reason, FROM_UNIXTIME(rbf.reviewed)
FROM dosomething_reportback as rb
INNER JOIN dosomething_reportback_file as rbf on rb.rbid = rbf.rbid
WHERE rbf.status = 'flagged'
AND (rb.flagged = 0 or rb.flagged is null);
```

itsjoekent:
Fixes #4847 

Problem was the following,
- Two reportback items call the setFlaggedPromoted option (items are looped through in the form submit)
- The second item is 'approved', which in the old setFlaggedPromoted function meant the flagged & promoted fields were set to 0, **regardless of there previous variable**. (this is why it was bugged)

Solution,
- Added a second step of verification in the 'else' (where approves are handled) to make sure a previous promoted or flagged field wasn't overridden 

@angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
@@ -237,7 +237,6 @@ function dosomething_reportback_files_form_submit($form, &$form_state) {
 
     $rbf = reportback_file_load($fid);
 
-    // Set flagged or promoted reasons.
     if ($values['status'] === 'promoted') {
       dosomething_reportback_set_promote_flag_reasons($values, 'promoted_reason');
     }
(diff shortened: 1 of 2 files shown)


angaither:
the logic looks :ok_hand: but can you add some comments to explain this logic.

itsjoekent:
@namimody can we spend a few minutes tomorrow going over this / testing it?

namimody:
@deadlybutter, yep! want to plan for 10?

itsjoekent:
sure

PR #4856, opened 2015-08-18
Improves ds pull stage script.

sergiitk:
Adds feature revert and cache clear commands after downloading stage database. This prevents Drupal from falling into fatal errors in case of inconsistencies between codebase and features code.


angaither:
:+1:

PR #4858, opened 2015-08-18
Re-pre-approved reportbacks, don't required status.

### LINKED ISSUE #4797: reportback inbox submission query retrieval issue

angaither:
Fixes #4797

@weerd is also a goddamn champion. 

When a status was required on reportbacks, it was causing drupal's default validation hook to run, which re-checks all the data to see that an option was selected 
This was causing two things: 
- the 'an illegal choice was made' error 
- the get the latest reportbacks query to run and causing rbs to be approved without a member of the campaigns team reviewing the reportback
### TO TEST
1. View a campaign with many reportbacks in the inbox and review/approve those
2. Submit a reportback while the Save method on the inbox is still running

The reportback that was submitted should NOT have been reviewed. 
This was previously fixed by disabling the `approved` radio button. This PR adds that back in and doesn't cause incorrect reportbacks to be approved.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
@@ -169,9 +169,7 @@ function dosomething_reportback_files_form($form, &$form_state, $status = NULL, 
 
   foreach ($result as $record) {
     $rb_path = 'admin/reportback/' . $record->rbid;
-    // @TODO: temporarily disable pre-approved to unblock campaigns team from reviewing reportbacks.
-    // Refs #4797
-    // $status = 'approved';
+    $status = 'approved';
     if (isset($record->status) && $record->status != 'pending') {
       $status = $record->status;
     }
@@ -192,7 +190,6 @@ function dosomething_reportback_files_form($form, &$form_state, $status = NULL, 
       'status' => array(
         '#type' => 'radios',
         '#title' => t('Status'),
-        '#required' => TRUE,
         '#options' => $status_options,
         '#default_value' => $status,
       ),


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
```diff
@@ -169,9 +169,7 @@ function dosomething_reportback_files_form($form, &$form_state, $status = NULL,
 
   foreach ($result as $record) {
     $rb_path = 'admin/reportback/' . $record->rbid;
-    // @TODO: temporarily disable pre-approved to unblock campaigns team from reviewing reportbacks.
-    // Refs #4797
-    // $status = 'approved';
+    $status = 'approved';
```
weerd, line 172: If we re-approve everything, we come back to the issue of user's submitting an RB and it getting auto-approved into the reviewed listing.

mikefantini:
@angaither just a note on how to test, should we follow these steps more specifically?
- load rb inbox that has 25+ fids in the inbox
- have a user load up a campaign
- have that user submit a rb
- wait until rb is submitted
- review rbs that are in the inbox and click save

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
mikefantini, line 172: @angaither @weerd does it make sense to hold off on adding back approve as default selected until we verify that we're no longer getting the illegal choice detected warning?

angaither, line 172: yeah, pre-approved wasn't the thing that was causing the illegal choice error. 
it causes things to be approved that aren't actually being reviewed.

weerd, line 172: The illegal choice error was that we previously were **requiring** a status to be selected, and since we turned off selecting "approved" it was throwing the error for the recently submitted RB that it was bringing into the query of items to consider.

angaither:
@mikefantini those are the correct steps @mikefantini

weerd:
@angaither since you have the PR ready, update the language to represent the change from _most recent_ to _oldest_

https://github.com/DoSomething/phoenix/blob/1f1a7b12dd57385d13f28ea72d36e4df0ebf3262/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc#L158

Maybe something like this, so it'll update depending on `order` specified:

``` php
// Display reportback forms in asc order.
$params['order'] = 'asc';

...

$page_size_copy = '';

if ($total > $page_size) {
  $label = ($params['order'] === 'asc') : 'Displaying the oldest ' . $page_size ? 'Displaying the most recent ' . $page_size;

  $page_size_copy = '<p>' . t($label) . '</p>';
}
```

Haven't actually tried the above code, just a suggestion :)

angaither:
Removing the required status field, and not having things preselected will allow people to submit the form and save a reportback will no status.

weerd:
In the above case, I believe the RBs stay in the inbox as pending...

angaither:
nope -_- 
once they are in the inbox, and that field isn't required it looks like it's been reviewed, but it has no status

weerd:
Bummer. Two steps forward, one step back :\

weerd:
We could maybe add some logic where while it doesn't validate them, it also takes those out from being processed and just leaves them untouched?

angaither:
yeah, we could talk about adding some custom validation perhaps

weerd:
Sounds good.

mikefantini:
to summarize the current status, because of drupal's default querying for information again upon form submit, this is the current reality given our options:
- approve not pre-selected, status required -> illegal choice error messages will occur when inbox is less than 25 and a user submits a reportback while an editor is reviewing the inbox and goes to submit the form
- approve not pre-selected, status not required -> if an editor accidentally doesn't select a status for a reportback, the reportback will go into reviewed with no status
- approve pre-selected, status not required -> in the case where there are less than 25 reportbacks in the inbox and a user submits a reportback while an editor is reviewing, those new reportbacks will skip the inbox and be marked as approved. we can bring this trade-off to the campaigns team and see if they are willing to double check the reviewed tab after they have submitted the form to verify that nothing sneaked through while they were reviewing.

and yes let's discuss other options tomorrow after standup

weerd:
Did we decide how we're moving forward on with this one?

mikefantini:
ok so just to recap, chatted w/ some campaigns folks about this and we're going to stick with what we've got:
- approve not pre-selected
- status required

we'll continue to monitor with them and readdress if this is slowing them down too much moving forward

weerd:
Cool cool. Thanks for the recap :)

PR #4861, opened 2015-08-18
Moving images to sites/default directory

### LINKED ISSUE #4837: Tmblr Problem Share Functionality Lagging
image

After trying to share a problem statement to TUmblr, a popup appears and appears to continue to try and load for ages without ever displaying content. Weirdly, shares work from the permalink.

sbsmith86:
Updating the problem shares image magic script to place the generated images in `/sites/default/files/images` directory. This seems to be a more standard path used across all instances and it is shared across all web heads. 

Should fix #4837 

@angaither @sheydari-ds


angaither:
:+1:

PR #4862, opened 2015-08-18
Hide problem shares when win module is enabled

sbsmith86:
When the win module is active, this should hid problem shares. Also fixed and issue where we were checking if the `$hot_module_enabled` variable was true or false, but that variable will only be set if the hot module is active so the check needs to be for `isset()`.

Fixes #4862 

@DoSomething/front-end


DFurnes:
Cool cool cool. :+1:

PR #4863, opened 2015-08-18
This value was set to approved above, but since that's commented out,…

angaither:
… this shouldn't be set.

Refs #4797 

for now, let's leave things required, and not pre-selected as approved.


weerd:
:+1:

PR #4864, opened 2015-08-19
Give '' as default status.

angaither:
This way after a reportback has been reviewed the status that was selected will show.
Refs #4797


weerd:
:+1:

PR #4868, opened 2015-08-19
Hot module refactor

sbsmith86:
Moving the hot module mark up to it's own template, similar to what we did with the win module to clean up the campaign tpl a little bit.


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -251,18 +250,18 @@ function dosomething_campaign_preprocess_hot_module(&$vars, &$wrapper) {
...
   }
   else {
-    $vars['time_left'] = t($hours_left . ' hours left');
+    $time_left = t($hours_left . ' hours left');
   }
 
   //Calculate progress goal & copy
   $goal = dosomething_helpers_get_variable('node', $vars['nid'], 'goal');
-  $vars['goal'] = $goal;
+  // $vars['goal'] = $goal;
```
angaither: should this actually be commented out? or just remove it if it's not needed anymore

angaither:
quick comment but :+1:

Issue #4829, opened 2015-08-11
Flagged RBs Show NULL and 0 Statuses


namimody:
```
select rb.rbid, rbf.fid, rb.flagged, rb.promoted, rbf.status, from_unixtime(rb.created), from_unixtime(rbf.reviewed) as reviewed
from dosomething_reportback rb
inner join dosomething_reportback_file rbf on rb.rbid = rbf.rbid
and rbf.status = 'flagged'
and (rb.flagged is null or rb.flagged = 0)
order by reviewed desc;
```

![image](https://cloud.githubusercontent.com/assets/12396076/9212464/046cb4c0-4057-11e5-9e9b-be07113cb397.png)

92113 - reviewed 10:55:54 on 8/11 as flagged, status shows up as 0 instead of 1
92140 - reviewed 10:55:54 on 8/11 as flagged, status shows up as NULL instead of 1

Code deployed for this at 16:33 on 8/3:
![image](https://cloud.githubusercontent.com/assets/12396076/9212520/944e34ba-4057-11e5-8cc1-96e8b520a864.png)

(98019 and 98022 reflect a separate known issue.)

namimody:
Another instance -- pulled 8.12.15 at 7:20 PM on Visionary Volunteer. See 92139

![image](https://cloud.githubusercontent.com/assets/12396076/9239169/2bcbdfd4-4127-11e5-86a1-0d5861525f98.png)

namimody:
Another instance - Get Lucky pulled 8.17 around 10 AM. See RB 99004 and 99282. note this is from the slave database.
![image](https://cloud.githubusercontent.com/assets/12396076/9307728/1e6cf77e-44ce-11e5-94cb-e098c9e2437d.png)

namimody:
Get Lucky, pulled @ 4:50 pm 8.17

![image](https://cloud.githubusercontent.com/assets/12396076/9316022/a782163e-4500-11e5-8606-b006e1e412e5.png)

namimody:
Storm Drain Street Art, pulled 4:57 PM, 8.17. RB 99147 flagged and excluded but marked as NULL.
![image](https://cloud.githubusercontent.com/assets/12396076/9316203/a82d55ca-4501-11e5-9aca-a903e16949a3.png)

angaither:
@namimody @mikefantini have you been able to replicate this or #4829 on stage?

weerd:
[...]

PR #4872, opened 2015-08-20
Allow extra roles to review/edit reportbacks.

### LINKED ISSUE #4871: when editors and/or member support roles review reportbacks, flagged status is set to null
when admins review reportbacks flagged status is set properly (0 for flagged, 1 for everything else)

we recreated this several times on staging and production with the same user and just changing their permissions

every single reportback that Lizzy reviewed has flagged status of null on nid 5566 has flagged status null. her roles are editor and member support.

when diego reviewed reportbacks on nid 5566 they had the correct flagged status. he is an admin. when we changed him to editor role, the reportbacks he reviewed had incorrect flagged status

also saw this with nid 2900 and sam's accounts

angaither:
Fixes #4871, Refs #4829


weerd:
:+1: 😭

PR #4878, opened 2015-08-20
Additional reportback cleanup from role issues

angaither:
Update reportbacks that the file has the correct status, but the rb itself doesn't.

Refs #4872


weerd:
:dancer: :+1:

PR #4882, opened 2015-08-21
Google analytics refactor

### LINKED ISSUE #4793: Google Analytics: Fender Refactor
I think we should refactor how we implement google analytics on the front end. Documenting it here for discussion. 

My idea is that we come up with some standard class names and data attributes to use and then create a function that sends listens to click events on that class and sends different info to GA based on what data attributes it has?

For example, if tracking clicks we could simply do: 

`<a href="http://www.yerp.com" class="ga-click" data-event="share" data-type="facebook" data-module="problem statement" />`

Thoughts? (naming is just for example, im not sure what better names should be) @DoSomething/front-end

sbsmith86:
Resolves #4793 

Updated the analytics script to be a bit more DRY. Now, all we have to do to track custom click events is add a `ga-click` class on the element we are tracking with some data attributes that pass in the custom info that we want to track. 

@DoSomething/front-end

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -219,7 +219,7 @@ function dosomething_campaign_preprocess_win_module(&$vars, $wrapper) {
     )
   );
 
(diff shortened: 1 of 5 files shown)


DFurnes:
:+1:!

Issue #4867, opened 2015-08-19
Stuck with old cached assets after deploy


sergiitk:
```
Dave Furnes: hmm, some font issues on production post-deploy

[7:30 PM] Dave Furnes uploaded an image: cursive font 
Add Comment

[7:30 PM] Dave Furnes uploaded an image: social icons 
Add Comment

[7:31 PM] Dave Furnes: We did switch up where some fonts are hosted (served via URL on our CDN now, rather than included in the NPM package), but the icon fonts are still being packaged the same way

[7:31 PM] Dave Furnes: So not sure why that’s borked too... Anyone have any ideas if anything might have changed with CORS settings perhaps?

[7:33 PM] Sergii Tkachenko: Looking into it

[7:33 PM] Mike Fantini: @dave nothing looks weird to me on production on chrome logged in and logged out

[7:33 PM] Dave Furnes: Kk, thanks Sergii! Seems to be failing in different ways in different browsers.

[7:33 PM] Dave Furnes: Yeah, Chrome is a-okay. @fantini

[7:33 PM] Sergii Tkachenko: it’s good here, right? http://staging.beta.dosomething.org/campaigns/closed-campaign?source=node/1165
Closed Campaign
This is a test closed campaign. (56KB)


[7:34 PM] Sergii Tkachenko uploaded an image: chrome / mac / prod 
Add Comment

[7:34 PM] Sergii Tkachenko uploaded an image: chrome / mac / staging 
Add Comment

[7:34 PM] Dave Furnes: Yep. That second one is correct.

[7:35 PM] Dave Furnes: It looks like it might be caching an old version of the CSS file sometimes? I’m getting errors of it trying to load the included icon fonts from the wrong path.

[7:35 PM] Sergii Tkachenko: so looks like it fail in chrome

[7:35 PM] Sergii Tkachenko: ok

[7:35 PM] Sergii Tkachenko uploaded an image: Pasted image at 2015-08-19, 7:35 PM 
Add Comment

[7:36 PM] Dave Furnes: yep. it should be looking in `/assets/icons` rather than `/assets/fonts/icomoon`

[7:36 PM] Sergii Tkachenko: staging Request URL:https://www.dosomething.org/sites/default/files/fonts/covered-grace/CoveredGrace-Regular.woff

[7:37 PM] Sergii Tkachenko: hmhm

[7:37 PM] Dave Furnes: that’s correct

[7:40 PM] Sergii Tkachenko: so where should it be?

[...]

PR #4888, opened 2015-08-24
Add a query string to bust cache on deploys.

DFurnes:
# Changes

Adds a query string to CSS files (e.g. `app.css?v1.0.32`) so that they properly invalidate cache on deploys. References #4867.

For review: @sergii-tkachenko


sergiitk:
:+1:

PR #4898, opened 2015-08-25
Adds new confirmation field & styling to form

### LINKED ISSUE #4838: Odd Activity on DONATE Button Back End (Stripe)
Seems like something similar to automated tests are running on the site's donate button, resulting in hundreds of $1.00 failed donations from gibberish gmail addresses.

This is a problem because:
1) it makes it very difficult for Kenia to find real donations
2) sometime donations go through and then are cancelled
2a) Kenia needs to record all donation activity
2b) we are charged for fraudulent activity
3) our partnership with Alexander Wang launches in Sept and will rely on this page working well

More info to come on when this began.

itsjoekent:
Fixes #4838 

Adds a new email confirmation field to the form and updates the form submit logic to include this new field. Also inserts styling to correctly display the new field. 

Has been verified to work on local development and on the remote service by @namimody 

@angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_payment/dosomething_payment.module
+++ b/lib/modules/dosomething/dosomething_payment/dosomething_payment.module
@@ -245,6 +245,12 @@ function dosomething_payment_form_submit($form, &$form_state) {
       );
     }
   }
+  else {
+    // Send stathat info.
+    if (module_exists('stathat')) {
+      stathat_send_ez_count('drupal - Donate - honeypot', 1);
+    }
+  }
 }
 
 /**
--- a/lib/modules/dosomething/dosomething_payment/dosomething_payment.module
+++ b/lib/modules/dosomething/dosomething_payment/dosomething_payment.module
@@ -226,20 +226,24 @@ function dosomething_payment_form_submit($form, &$form_state) {
   $dollars = $values['amount'];
   // Expected amount to charge to Stripe should be in cents.
   $values['amount'] = $values['amount'] * 100;
-  // Send Payment request to Stripe API.
-  if (dosomething_payment_post_payment_stripe($values) && !isset($values['email'])) {
-    // Send stathat info.
-    if (module_exists('stathat')) {
-      stathat_send_ez_count('drupal - Donate - count', 1);
-      stathat_send_ez_value('drupal - Donate - value', $dollars);
+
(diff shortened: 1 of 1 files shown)


DFurnes:
This looks _sweet_.

lib/modules/dosomething/dosomething_payment/dosomething_payment.module
```diff
@@ -214,7 +227,7 @@ function dosomething_payment_form_submit($form, &$form_state) {
   // Expected amount to charge to Stripe should be in cents.
   $values['amount'] = $values['amount'] * 100;
   // Send Payment request to Stripe API.
-  if (dosomething_payment_post_payment_stripe($values)) {
+  if (dosomething_payment_post_payment_stripe($values) && !isset($values['email'])) {
```
DFurnes: Has this been tested? I think this will still send the payment information to Stripe even if the `$values['email']` is set. If you move the `isset()` check first, it should short-circuit.

itsjoekent: Oh right, because we're calling that function and checking its return value. Yeah Ill move it up. I tested it on my local and didn't get the thanks for donating page and I figured that meant it worked. but now I see why it didnt

lib/modules/dosomething/dosomething_payment/dosomething_payment.module
```diff
@@ -213,20 +226,24 @@ function dosomething_payment_form_submit($form, &$form_state) {
...
   $values['amount'] = $values['amount'] * 100;
-  // Send Payment request to Stripe API.
-  if (dosomething_payment_post_payment_stripe($values)) {
-    // Send stathat info.
-    if (module_exists('stathat')) {
-      stathat_send_ez_count('drupal - Donate - count', 1);
-      stathat_send_ez_value('drupal - Donate - value', $dollars);
+
+  // Check the honey pot for honey
+  if (!isset($values['email'])) {
```
angaither, line 231: let's add a an else here to track to stathat if the value is set

DFurnes, line 231: +1. That'll be useful to keep an eye on.

angaither:
:+1:

PR #4900, opened 2015-08-25
Fix style bug on problem shares

### LINKED ISSUE #4899: problem share text spacing is off
screen shot 2015-08-25 at 10 21 49 am

sbsmith86:
Fixes #4899 

I updated the share bar function to not take in an analytics class parameter anymore, but I was still passing it when generating the problem share bar. The function was then not applying share bar classes I passed to it, because the ordering was off.


angaither:
:+1: 

the js-class gets added elsewhere now?

sbsmith86:
@angaither we use a generic `ga-click` tracking class that the js binds to, and then we use data attributes to pass in the custom stuff we want to track.

PR #4903, opened 2015-08-25
Quick form submit logic fix

itsjoekent:
@angaither  or @DFurnes 

$values['email'] was an empty string, and thus isset couldnt do the job.


angaither:
:+1:

Issue #4908, opened 2015-08-25
Last reportback item overrides all the items


itsjoekent:
Let's say theres 3 reportback items, you flag the first two and promote the third. If you check the database for that reportback, FLAGGED = 0 and FLAGGED_REASON = NULL

Example: http://staging.beta.dosomething.org/admin/reportback/3069
- Flagged second RB item
- Promoted third RB item

What the DB shows

![screen shot 2015-08-25 at 4 57 27 pm](https://cloud.githubusercontent.com/assets/897368/9479216/677f7450-4b4a-11e5-8117-97a7da8d29c0.png)

I'm self assigning because on my local I've already fixed this

angaither:
duplicate of #4847

Issue #4880, opened 2015-08-20
Saving from RB review page sets all prior flagged_reason & promoted_reason values to NULL


weerd:
**Context:** When viewing a RB with numerous items in the admin interface, clicking save without changing any values will update the database and remove any of the `flagged_reason` and `promoted_reason` values that were previously set and change them to `NULL`.

This is because we are not loading the form with the reason indicated, just the `flagged` or `promoted` radio button selected, so when the form goes through and collects the form entries, it doesn't see anything for the reason and set it as `NULL` when it goes to update the form data in the db.

@mikefantini @namimody @angaither

PR #4909, opened 2015-08-25
Updates review reportback logic to support multiple files

itsjoekent:
Rewrote a good chunk of the setFlaggedPromoted function, as a result this PR should fix the following issues, #4908 #4880 

Currently testing it, progress is being logged in a Google Sheet here: https://docs.google.com/a/dosomething.org/spreadsheets/d/1YA5QSNf5Jhs9myY-CUcY3Ca3arkDWgr4MUR7AiphniQ/edit?usp=sharing

Logic is much more clear and documented now, accounts for each reportback item before determining what the values should be. Clearly labels & separates the business logic. After this is merged I can another round of thorough testing with @namimody 

@angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
+++ b/lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
@@ -333,6 +333,33 @@ class Reportback extends Entity {
    *   Array storing values for flagged/promoted reasons.
    */
   public function setFlaggedPromoted($status = NULL, $values = NULL) {
+    $items = ReportbackItem::get($this->getFids());
+
+    // Seperate multiple file logic from single file logic
+    if (count($items) > 1) {
+      // Determine all of the reportback statuses
+      $flaggedReportbacks = false;
+      $promotedReportbacks = false;
+
+      foreach ($items as $item) {
+        if ($item->status === "flagged") {
+          $flaggedReportbacks = TRUE;
+        }
+        else if($item->status === "promoted") {
+          $promotedReportbacks = TRUE;
+        }
+      }
+
+      // The optional conditional verifies that flagged always overrides promoted
+      // regardless if its a new or existing reportback status
+      if ($flaggedReportbacks || $status === "flagged") {
+        $status = "flagged";
+      }
+      else if($promotedReportbacks) {
+        $status = "promoted";
+      }
+    }
+
     if ($status === 'flagged') {
       $this->flagged = 1;
       $this->flagged_reason = $values['flagged_reason'] ?: NULL;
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
```diff
@@ -333,27 +333,72 @@ public function deleteFiles() {
    *   Array storing values for flagged/promoted reasons.
    */
   public function setFlaggedPromoted($status = NULL, $values = NULL) {
+    $items = ReportbackItem::get($this->getFids());
+
+    // Determine all of the reportback statuses
+    $flaggedReportbacks = false;
+    $promotedReportbacks = false;
```
angaither: php vars should be `snake_case`

lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
```diff
@@ -333,27 +333,72 @@ public function deleteFiles() {
...
+
+    // Based on the logic decided above, modify the review values
     if ($status === 'flagged') {
       $this->flagged = 1;
-      $this->flagged_reason = $values['flagged_reason'] ?: NULL;
       $this->promoted = 0;
+
+      // Make sure the reason is correct
+      if (is_string($values['flagged_reason'])) {
+        if (empty($values['flagged_reason'])) {
```
angaither, line 371: why is it necessary to check both if it's a string and if it's empty?

lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
itsjoekent: sorry been writing a lot of JS lately haha

lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
itsjoekent, line 371: is_string verifies that the value isnt an array. If you don't select any option for an item the value in values['flagged_reason'] will be an array containing each option and 0. the second part verifies that the string being sent isnt empty (happens when you just save the page without clicking anything new)

itsjoekent:
@angaither ready for another look

lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
```diff
@@ -333,27 +333,78 @@ public function deleteFiles() {
    *   Array storing values for flagged/promoted reasons.
    */
   public function setFlaggedPromoted($status = NULL, $values = NULL) {
+    $items = ReportbackItem::get($this->getFids());
+
+    // Determine all of the reportback statuses
+    $flagged_reportbacks = false;
+    $promoted_reportbacks = false;
```
angaither: false should be CAPS!

angaither:
other than that :+1:

PR #4917, opened 2015-08-26
Fixes logic flow that allowed fatal error during parner image render.

### LINKED ISSUE #4916: using an image w/o a landscape crop in the sponsor image field on /node/add/campaign results in error page
went to create a new campaign on staging (http://staging.beta.dosomething.org/campaigns/please-dont-break-anything) and got the following error page

image

sergiitk:
Fixes #4916.
Plus, a helper function created for rendering no-image-found.png tag.

cc @mikefantini

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc
@@ -107,6 +107,10 @@ function dosomething_helpers_preprocess_partners_vars(&$vars) {
       if (!empty($image_node->field_image_landscape)) {
         $img = dosomething_image_get_themed_image($image_node->nid, 'landscape', '550x300');
         $vars['partner_info'][$delta]['image'] = $img;
+      } else {
+        // Must be overridden: partner image is unrendered object by default,
+        // but rendered string is expected in the template.
+        $vars['partner_info'][$delta]['image'] = dosomething_image_get_themed_no_image();
       }
     }
   }
--- a/lib/modules/dosomething/dosomething_image/dosomething_image.module
+++ b/lib/modules/dosomething/dosomething_image/dosomething_image.module
@@ -167,6 +167,8 @@ function dosomething_image_get_themed_image_by_fid($fid, $style, $alt = '') {
  *   Img tag of no-image-found.png
  */
 function dosomething_image_get_themed_no_image() {
+  // The function output is always the same, so static caching used.
+  // See drupal_static() documentation.
   $img = &drupal_static(__FUNCTION__);
   if (!isset($img)) {
     $args = [
--- a/lib/modules/dosomething/dosomething_image/dosomething_image.module
+++ b/lib/modules/dosomething/dosomething_image/dosomething_image.module
@@ -142,13 +142,7 @@ function dosomething_image_get_themed_image_by_fid($fid, $style, $alt = '') {
 
   $image = image_load($file->uri);
   if (!$image) {
-    $args = array(
-      'path' => DS_ASSET_PATH .'/dist/images/no-image-found.png',
-      'alt' => 'No image found placeholder',
-      'title' => 'No Image Found',
-      'attributes' => NULL,
-    );
-    return theme_image($args);
+    return dosomething_image_get_themed_no_image();
   }
   else {
     $content = array(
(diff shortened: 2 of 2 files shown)


lib/modules/dosomething/dosomething_image/dosomething_image.module
```diff
@@ -165,6 +159,28 @@ function dosomething_image_get_themed_image_by_fid($fid, $style, $alt = '') {
...
 /**
+ * Returns no image tag.
+ *
+ * Uses static cache to speed up the output.
+ *
+ * @return string
+ *   Img tag of no-image-found.png
+ */
+function dosomething_image_get_themed_no_image() {
+  $img = &drupal_static(__FUNCTION__);
```
DFurnes, line 172: Can you add some comments for what's going on here? It's very opaque to a lay person.

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc
```diff
@@ -107,6 +107,10 @@ function dosomething_helpers_preprocess_partners_vars(&$vars) {
       if (!empty($image_node->field_image_landscape)) {
         $img = dosomething_image_get_themed_image($image_node->nid, 'landscape', '550x300');
         $vars['partner_info'][$delta]['image'] = $img;
+      } else {
```
angaither, line 110: do we need this else? won't `$img` be set to the `no image` from `dosomething_image_get_themed_image_by_fid()`?

if we do need it, else should be on it's own line

lib/modules/dosomething/dosomething_image/dosomething_image.module
sergiitk, line 172: Actually, this is very common cache practice in drupal: https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/drupal_static/7
Plus, I mentioned this in the function docblock.

But you're right, I'll add the comment anyways.

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc
sergiitk, line 110: No, without this `else` statement the image stays an object, see https://github.com/DoSomething/phoenix/blob/2b6cf2c850c5e2b0fa91de0f92e97f92d33f4307/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc#L108-L109.

This caused fatal error in #4916.

sergiitk, line 110: Also not `dosomething_image_get_themed_image_by_fid()` used, but `dosomething_image_get_themed_image()`. It doesn't look an image up by fid and doesn't return the `no image` placeholder.

angaither, line 110: shouldn't `dosomething_image_get_themed_image()` then return the no image placeholder?
it shouldn't be necessary to call `dosomething_image_get_themed_no_image()` each time an image is requested, that should happen within the image module.

sergiitk, line 110: I don't think it should. This function is only called when `field_image_landscape` exists:

``` php
if (!empty($image_node->field_image_landscape)) {
  $img = dosomething_image_get_themed_image($image_node->nid, 'landscape', '550x300');
  $vars['partner_info'][$delta]['image'] = $img;
}
```

> it shouldn't be necessary to call `dosomething_image_get_themed_no_image()`

It uses static cache, it's very fast.

angaither, line 110: I understand that the call to the function is fast. I'm saying I don't think every image call should check for the field and then have to call to get the placeholder image.  the theme shouldn't be responsible for checking that
this should be all that's needed in the theme. 

```
$img = dosomething_image_get_themed_image($image_node->nid, 'landscape', '550x300');
$vars['partner_info'][$delta]['image'] = $img;
```

`dosomething_image_get_themed_image()` should return the image in the crop that's requested or the placeholder image if it doesn't have anything.

sergiitk, line 110: OK, I'll check other function calls and will refactor `dosomething_image_get_themed_image()` to provide default image as a fallback.

sergiitk, line 110: @angaither It looks like the initial logic is inherited in many other places, like here. https://github.com/DoSomething/phoenix/blob/2b6cf2c850c5e2b0fa91de0f92e97f92d33f4307/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc#L402

I'm not sure if default pic is needed for all other cases. Should we leave it like this for now and create a ticket to refactor `dosomething_image_get_themed_image()` to provide `no image` fallback?

angaither, line 110: I think since the quick fix isn't needed, (we can just add a landscape crop to fix the error) we should probably take the time to fix this in a more sustainable way

Issue #4922, opened 2015-08-26
Translatable fields are missing from API exports


sergiitk:
Reported by @aaronschachter:
http://dev.dosomething.org:8888/api/v1/campaigns.json?mobile_app=true&term_ids=669,667,668,670

Related #4834.

cc @mikefantini @sheydari-ds

aaronschachter:
I believe Message Broker still uses the older `content` endpoint, which calls a different callback function than the `campaigns` endpoint does. Something to double check as well.

sergiitk:
So the thing is that `LANGUAGE_NONE` language is used by default for extracting data:

Call_for_action/tagline loaded here: https://github.com/DoSomething/phoenix/blob/76c82fb77b1accdf015642588b75916e88c13791/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php#L528

Function [`dosomething_helpers_extract_field_data()` ](https://github.com/DoSomething/phoenix/blob/2b6cf2c850c5e2b0fa91de0f92e97f92d33f4307/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module#L105) is used to extract actual data. It has a language param, but it falls back to `LANGUAGE_NONE` by default.

There are a number of translatable fields, for which, the language param needs to be provided from now.

sergiitk:
@aaronschachter I believe we need to extend API with mandatory `$language` param, like
http://dev.dosomething.org:8888/api/v1/campaigns.json?mobile_app=true&term_ids=669,667,668,670&langauge=en

sergiitk:
@aaronschachter and I just had a chat and decided the following:
1. For now quickest and easies fix would be using `en` language as a new default for `dosomething_helpers_extract_field_data()` function. If the english value isn't there, fall back to `LANGUAGE_NONE`
2. Long term solution is API extension that makes it possible to render output in different languages, using `?language` URL parameter

cc @weerd

angaither:
I'm going to close this issue out, we only support `en` in the api, and this is intended for now.
We can re-visit if we prioritize other languages in external services

PR #4923, opened 2015-08-26
Provides temporary solution for API output of translatable fields.

sergiitk:
Solution 1 for #4922:
Temporary fix for translatable fields.
If no specific language requested, attempt return English value first.

image


aaronschachter:
Sweet :+1:

PR #4936, opened 2015-08-27
Fingerprint stylesheet assets with Webpack

DFurnes:
# Changes

Process stylesheets and assets using Webpack. This allows us to "fingerprint" assets in the `dist` folder so that they break cache only if they change. It also allows Webpack to handle whether to inline assets as data URLs based on their file size: for now, it's configured to inline any files that are less than 4kb.

screen shot 2015-08-27 at 12 21 52 pm

References #4867.

For review/discussion: @DoSomething/front-end


DFurnes:
Closing this, since it ended up getting merged with the shared config introduced in #4867!

PR #4945, opened 2015-08-27
Fixes staff pick label on homepage.

### LINKED ISSUE #4944: staff pick label missing from homepage
Yellow staff pick label on homepage also seems to have disappeared. happening on staging and prod.

working properly on both staging and prod on /volunteer/

also working properly when i filter campaigns on the homepage

image

sergiitk:
Fixes #4944.
`field_staff_pick` is not translatable.

cc @mikefantini


DFurnes:
:+1:

PR #4950, opened 2015-08-28
Adds custom error copy

### LINKED ISSUE #4938: add more copy to error message if cc doesn't go through on /donate
can't find existing error message copy but please add to it "Having problems with donating? Please try again or email donate@dosomething.org."

@deadlybutter please post what full copy of error message would be here just to make sure it all makes sense together.

related to #4937

itsjoekent:
Fixes #4938 

@angaither


angaither:
:+1:

PR #4952, opened 2015-08-28
Win module background image updates. Fixes #4941

### LINKED ISSUE #4941: adding win module image borks the win module
http://staging.beta.dosomething.org/campaigns/hunt-day-1-superstressface?source=node/1165

image node is http://staging.beta.dosomething.org/node/1696

image

sbsmith86:
Updates the win module background image so that it works at all break points. This background image is a little funky because the pattern lies around centered text, so we needed to play around with breakpoints and font sizes a bit to get it to work at all sizes. 

@DoSomething/front-end 

Notes: 
- I will make another issue to move this image from an image node to an svg.
- This also includes a small lint error I was getting in `Analytics.js`

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_win-module.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_win-module.scss
@@ -19,12 +19,18 @@
       }
     }
 
+    // Font sizes here are hard coded because they are
+    // outlier to our regular font system. Specific to the win module.
     h2 {
       color: $black;
       font-size: 76px;
 
+      @include media($medium) {
+        font-size: 120px;
+      }
+
       @include media($large) {
-        font-size: 124px;
+        font-size: 200px;
       }
     }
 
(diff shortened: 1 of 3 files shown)


lib/themes/dosomething/paraneue_dosomething/scss/content/_win-module.scss
```diff
@@ -34,9 +40,13 @@
       font-weight: $weight-normal;
       text-transform: uppercase;
 
-      @include media($large) {
+      @include media($medium) {
         font-size: $font-larger;
       }
+
+      @include media($large) {
+        font-size: $font-medium * 2;
```
DFurnes: Would `$font-hero` make sense here?

DFurnes: If not, it's probably best to just hardcode `48px` like you do for the others.

DFurnes:
One suggestion/comment, otherwise :ship:!

PR #4954, opened 2015-08-28
Remove crons that update flagged/promoted values.

### LINKED ISSUE #4905: remove reportback crons
we currently have crons looking for inconstancies in reportback items and the reportback flagged state
refs #4658
it might be time to remove those

angaither:
This shouldn't be needed anymore, best not to run updates when we don't need them!
Fixes #4905


weerd:
Cooool. :+1:

PR #4960, opened 2015-08-28
Fix hot module share bar

### LINKED ISSUE #4918: Problem Shares on Hot Module Campaigns Misaligned
image

sbsmith86:
Fixes #4918 

I made an update the caused the author information to float left on both the win and hot module, but we only needed that for the win module so that the share bar can sit next to it. 

@DoSomething/front-end


DFurnes:
:ship:

PR #4967, opened 2015-08-31
Adds override copy to campaign settings

### LINKED ISSUE #4638: progress copy override

itsjoekent:
Fixes #4638 
- Adds progress copy override field to campaign custom settings /node/1283/custom-settings
- Removes the old field from the campaign settings /admin/config/dosomething/dosomething_campaign
- Updates the preprocess logic to use the new field

@angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -278,7 +278,7 @@ function dosomething_campaign_preprocess_hot_module(&$vars, &$wrapper) {
     $rb_percent = 0;
   }
 
-  $override_copy = variable_get('dosomething_campaign_progress_copy_override', NULL);
+  $override_copy = dosomething_helpers_get_variable('node', $vars['nid'], 'progress_copy');
   if ($override_copy) {
     $progress_copy = $override_copy;
   }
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -278,7 +278,7 @@ function dosomething_campaign_preprocess_hot_module(&$vars, &$wrapper) {
     $rb_percent = 0;
   }
 
-  $override_copy = variable_get('dosomething_campaign_progress_copy_override', NULL);
+  $override_copy = dosomething_helpers_get_variable('node', $vars['nid'], 'progress_copy');
```
angaither: shouldn't this be `progress_copy_override`?

itsjoekent: yeah, woops

I added _override in a diff. commit and forgot to change the preprocess

itsjoekent: OR apparently I did change and I just forgot to save the file

screen shot 2015-08-31 at 11 28 01 am

angaither:
:+1:

PR #4970, opened 2015-08-31
Removing language links from search index node view

### LINKED ISSUE #4969: Remove language from search snippet
When apachesolr indexes the nodes it uses the 'search_index' node view to which language links are added. This causes the wrong language to be added to the beginning of the search snippet. No language should be added to the node content/snippet.

blisteringherb:
- The translation module adds links to the node which need to be removed when
  the node is indexed so that snippets are built properly.

Resolves #4969


angaither:
:+1:

blisteringherb:
I put this against the wrong base branch, so I'm going to wait until we've merged the apachesolr_multilingual code into dev. This code isn't a blocker or critical.

PR #4971, opened 2015-08-31
Moves entity save to before review call

### LINKED ISSUE #4879: Updating flagged RBItem status does not update the flagged state
**Context:** viewing a Reportback with a series of reviewed items in the CMS and one of them has a status of _flagged_, all of the items including the flagged one will have a `1` in the `flagged` field.

If you update the flagged item and change the status to promoted or approved, the `status` field updates properly, but the `flagged` field remains as `1` for all the items in the RB.

@mikefantini @namimody @angaither

itsjoekent:
The reason the bug was occuring and why this change fixes it is outlined in detail here: https://github.com/DoSomething/phoenix/issues/4879#issuecomment-134745880

Fixes #4879 

@angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/includes/ReportbackItem.php
+++ b/lib/modules/dosomething/dosomething_reportback/includes/ReportbackItem.php
@@ -182,6 +182,10 @@ class ReportbackItem extends Entity {
     if (isset($values['promoted_reason'])) {
       $promoted_reason = $values['promoted_reason'];
     }
+
+    // Must be saved before calling setFlaggedPromoted
+    $updatedReportbackItem = entity_save('reportback_item', $this);
+
     $reportback->setFlaggedPromoted($this->status, $values);
 
     if (!empty($values['delete'])) {
@@ -189,7 +193,7 @@ class ReportbackItem extends Entity {
     }
 
     // Save the reviewed properties.
-    return entity_save('reportback_item', $this);
+    return $updatedReportbackItem;
   }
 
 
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_reportback/includes/ReportbackItem.php
```diff
@@ -182,6 +182,10 @@ public function review($values) {
     if (isset($values['promoted_reason'])) {
       $promoted_reason = $values['promoted_reason'];
     }
+
+    // Must be saved before calling setFlaggedPromoted
+    $updatedReportbackItem = entity_save('reportback_item', $this);
```
angaither: javascript is invading your life! `snake_case`

itsjoekent: oh my god it really is, i think i need to see a doctor

angaither:
@deadlybutter I'm curious _why_ calling save on the item after the `setFlaggedPromoted` function doesn't allow this to work...

itsjoekent:
sure! the problem is we're calling setFlaggedPromoted, which looks at all of the reportback items to determine if it should be flagged, promoted, etc. Because this item wasn't saved yet, that function still thinks it's flagged. Because this also happens to be the reportback item with the reason, the logic ignores it thinking we're supposed to flag.

Does that make sense...?

angaither:
ahhh, got it thanks @deadlybutter

itsjoekent:
@angaither good to go?

angaither:
:+1:

PR #4972, opened 2015-08-31
Installing apachesolr_configen and apachesolr_multilingual

### LINKED ISSUE #4884: Install and configure Solr modules to support translation
The Drupal Solr installation doesn't have support for translation out of the box, and the Apachesolr Multilingual module adds some add'l support for non-english languages.

https://www.drupal.org/project/apachesolr_multilingual

It depends on ApacheSolr Configen https://www.drupal.org/project/apachesolr_confgen, so that must be installed also

blisteringherb:
- These modules add support for indexing fields in solr in all langs
  and for displaying search results appropriately based on the user's
  language

@angaither Can you take a look?
Resolves #4884


angaither:
:+1:

PR #4974, opened 2015-08-31
Hot module share

### LINKED ISSUE #4643: enhanced FB/Tumblr share for hot module
enhancement over #4447 this issue is to include the below image as the photo in FB/Tumblr shares

image

this image would need to be updated as cron runs so that it shows updated numbers

DFurnes:
# Changes

Adds exclusive new share images for the _hottest_ of modules. Closes #4643.

hot-1426

Creates a new `application_path` helper function for getting filesystem paths relative to the application root (i.e. above the `html/` directory). Also extracts "x days/hours remaining" logic into a helper function so it can be shared between preprocess function and the share cron job.

For review: @angaither @sbsmith86


lib/modules/dosomething/dosomething_campaign/dosomething_campaign_hot_shares/dosomething_campaign_hot_shares.cron.inc
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * @file
+ * Provides cron job command for dosomething_campaign_problem_shares.module.
+ */
+
+/**
+ * Implements hook_cron().
+ */
+function dosomething_campaign_hot_shares_cron() {
+  $results = db_query("SELECT nid FROM node WHERE type ='campaign'");
```
angaither: we should only have to run this on published campaigns

lib/modules/dosomething/dosomething_campaign/dosomething_campaign_hot_shares/dosomething_campaign_hot_shares.cron.inc
```diff
@@ -0,0 +1,36 @@
...
+
+  // Get the relevant paths
+  $path = dosomething_helpers_application_path('/lib/modules/dosomething/dosomething_campaign/dosomething_campaign_hot_shares');
+  $font_path = dosomething_helpers_application_path('/lib/modules/dosomething/dosomething_campaign/fonts');
+  $image_location = dosomething_helpers_application_path('/html/sites/default/files');
+
+  // Each active campaign, call shell script to create problem share images.
+  foreach($results as $result) {
+    $campaign = Campaign::get($result->nid)[0];
+    $node = node_load($result->nid);
```
angaither, line 27: I'm curious here, since we are essentially get the campaign twice here, (one is a `node_load` and the other is with the class), which takes longer, and if we could squish this into only one way of getting the node

we could either attach `time_left` onto the campaign class, or get the reportback `noun & verb` from the sql query

my worry is that we have a lot going on in this cron and I don't want it to be super slow, but maybe that's an unfounded worry

sbsmith86:
Looks great to me :100:

lib/modules/dosomething/dosomething_campaign/dosomething_campaign_hot_shares/dosomething_campaign_hot_shares.cron.inc
DFurnes: Updated!

lib/modules/dosomething/dosomething_campaign/dosomething_campaign_hot_shares/dosomething_campaign_hot_shares.cron.inc
DFurnes, line 27: Did a little benchmarking to see. When running on my local development machine (so not necessarily a scientific test), the total time spent on making hot share images was `22.258s` and the total time making the problem shares was `25.073s`, so they're in the same ballpark.

From (again, unscientific) benchmarking, it looks like loading an individual campaign with `Campaign::get` takes about `0.005s` and loading a campaign node with `node_load` takes about `0.000036s`. From what I can tell, the overwhelming bulk of the time is spent running the Imagemagick shell script.

angaither, line 27: so, unfounded then :+1:

DFurnes, line 27: :grimacing:

PR #4979, opened 2015-09-01
Enable the hot module share module.

DFurnes:
# Changes

Enables the hot module share module _for real_.
For review: @angaither 

enable hot shares module


angaither:
:droplet: :ship:

PR #4983, opened 2015-09-01
Fixes hot&win module timing issue

### LINKED ISSUE #4926: hot module not showing despite being in range
[This campaign](http://staging.beta.dosomething.org/node/1656/edit) created by [uid 8](http://staging.beta.dosomething.org/user/8) has a high season that ends on August 27th 2015. However, on August 26th 2015 at 8:24pm EST the hot module is not showing up for me as [this new user - uid 1705230](http://staging.beta.dosomething.org/user/1705230) I just created.

For [uid 8](http://staging.beta.dosomething.org/user/8) who created the campaign, it is showing up as ending in 3 hours (as of August 26th 8:24pm EST).

uid 8 has time zone set as America/New York. uid 1705230 has time zone set as UTC.

How do we make it so that:
1. The hot module expires at 11:59pm EST the night of the end of the high season
2. That all users see this as being the end time (i.e. at 10:59pm EST users in New York, Chicago, LA, Sydney, Rome, etc. all see 1 hour left)

image

UID 1705230:
image

UID 8:
image

itsjoekent:
Fixes #4926

Updates all logic to be "New York" time based. All user times and campaign times are converted into this.

Updated relevant helper functions to reflect this along with all the instances where we're creating start and end dates. Also made code DRY in several places.

@angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -738,8 +738,9 @@ function dosomething_campaign_add_shipment_form_vars($node) {
  */
 function dosomething_campaign_is_hot_module($vars) {
   $today = dosomething_campaign_set_est_timezone(new DateTime());
-  $start_date = dosomething_campaign_set_est_timezone(new DateTime($vars['field_high_season'][0]['value']), TRUE);
-  $end_date = dosomething_campaign_set_est_timezone(new DateTime($vars['field_high_season'][0]['value2']), TRUE, TRUE);
+  $start_date = new DateTime($vars['field_high_season'][0]['value']);
+  $end_date = new DateTime($vars['field_high_season'][0]['value2']);
+  $end_date->setTime(23, 59, 59);
 
   // Hot module is only enabled when variable is set, high season dates exist and there is still days left.
   if (!empty($vars['field_high_season']) && (($start_date < $today)  && ($today < $end_date))) {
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -843,14 +839,22 @@ function dosomething_campaign_is_active($node) {
...
    * @param boolean $end_of_day;
    *  Optional - set the date to end of day
    *
    * @return DateTime
    *  Updated time
    */
-function dosomething_campaign_set_est_timezone($date) {
+function dosomething_campaign_set_est_timezone($date, $add_four = FALSE, $end_of_day = FALSE) {
+  if ($add_four) {
+    $date->add(new DateInterval('PT4H'));
```
angaither, line 855: why are we adding four? for UTC? not all users are in UTC so won't this break for people who have timezones set?

itsjoekent, line 855: right, forgot we had users without UTC. Ill figure out some magic and fix this

itsjoekent, line 855: looks like I should be able to use this function,

http://php.net/manual/en/datetimezone.getoffset.php

something along the lines of,

```
if ($add_four) {
  global $user;
  $user_timezone = $user->timezone;
  $difference = (new DateTime().getOffsest($user_timezone) / 60) / 60;
  $date->add(new DateInterval('PT' . $difference . 'H'));
}
```

angaither, line 855: shouldn't we just be able to calculate time left in the hot module and set the timezone to EST there?

itsjoekent, line 855: no, if you dont add the four hours what happens is it thinks its 23:59:59 of the day beforehand because when you convert you loose four hours (its really confusing I know, damn time).

I got the code working 

```
  global $user;
    $user_timezone = new DateTimeZone($user->timezone);
    $user_date = new DateTime("now", $user_timezone);
    $new_york_timezone = new DateTimeZone('America/New_York');
    // Get the timezone difference, divide by total seconds in hour, convert to int
    $difference = (int) ($new_york_timezone->getOffset($user_date) / 3660);
```

but then I realized this is pointless! the add_four is only ever used by 'end/start date' variables because of how we modify them. this if block should actually be ignoring the user timezone.

this code might still be useful to some degree. I want to test as a user in another time zone and see if the new DateTime reflects that timezone or it defaults to UTC as well. If it defaults to UTC then the code above is sorta needed. 

If this doesnt make sense I can try explaining it again

angaither:
:+1:

PR #4984, opened 2015-09-01
Don't preprocess hot/win vars for sms campaigns.

### LINKED ISSUE #4982: sms campaigns attempt to display hot module
in `dosomething_campaign_preprocess_node` it’s checking if it’s hot before it’s checking if it’s an sms campaign. the checks for `is_hot` and `is_win` should probably be at the end of the preprocess (along with `preprocess_common_vars`

angaither:
moved the call to test if campaigns are in a hot/win state after the check for sms campaigns. 
Fixes #4982


DFurnes:
:ok_hand:!

PR #4985, opened 2015-09-01
Updates promoted image processing

### LINKED ISSUE #4976: win module reportback image
the win module is showing a random reportback image, but it isn't from that campaign.

it should be a random reportback image from the campaign that was hot

itsjoekent:
Fixes #4976 
- Adds try/catch around the ReportbackItem Find to solve the broken page error
- Adds a fallback to 'approved' status if no promoted items are found
- Adds a null check to the template in the event no approved items are also found

@angaither


angaither:
:+1:

PR #4991, opened 2015-09-02
Prevent SMS games from being configured with both SP & MP vars.

### LINKED ISSUE #4990: Clarify Fields on SMS Game Backend
image

The field labels on the opt-in paths under configuration > dosomething config > third party opt ins > sms games are unclear, and have resulted in the wrong fields being filled out for multiplayer v. single player games.

DFurnes:
# Changes

Fixes #4990. :100: :phone: :mobile_phone_off: :baggage_claim: 

For review: @angaither @tongxiang


angaither:
code looks :ok: 
but just to clarify, are you hiding form fields from the admin, or is this the form that the user fills out?

DFurnes:
This just throws an error on the SMS game page if a campaign is configured incorrectly, rather than silently discarding fields & trying to make the best of it.

DFurnes:
(So if all four fields are filled out, it now shows the same error message that we display if none of the fields are filled out.)

angaither:
:+1:

PR #5003, opened 2015-09-02
Win Module Share Bar Style Update

### LINKED ISSUE #4987: Adjust Sharing Button Alignment on Win Module
image

image

image

Seems that because of the varying length of the job titles, sharing buttons are being pushed down to look a bit out of place.

sbsmith86:
Fixes #4987

Set a width on the campaign author info and the share bar so that longer campaign author titles will wrap and they will always stay next to each other at larger breakpoints.

@DoSomething/front-end


DFurnes:
:art: :ship:

PR #5004, opened 2015-09-02
Hides non translatable fields from Mexico and Brazil admins

### LINKED ISSUE #4934: br admin and mx admin should only have access to br and mx fields
As a brazil admin or mexico admin, I should only have access to Brazilian Portuguese fields or Mexican Spanish fields respectively (i.e. I shouldn't be able to create a campaign in the English fields nor should I be able to edit any content in the English fields).

Currently, when I go to create a campaign as a brazil admin I:
1. default am presented with English fields
2. have access to non translated fields (like Primary Cause, Cause, Primary Action Type, etc.)

Also when switching from English to Brazilian Portuguese I still have the non translated fields presented to me.

What should happen is that when trying to create a campaign as a brazil admin:
1. Brazilian Portuguese should be default selected as the language on /node/add/campaign
2. I shouldn't be able to change this language
3. Only Brazilian Portuguese fields should show to me - so if the field was not made translatable, then I should not be able to enter content in that field.

cc: @julielorch @sergii-tkachenko @blisteringherb @sheydari-ds

itsjoekent:
Fixes #4934 

_Side note, first function in dosomething global module_ :tada: :boom: :tada: 

Adds a node_form_alter and removes access from non translatable fields for brazil and mexico admins 

Do I send these to @blisteringherb or @angaither ?

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -10,17 +10,32 @@ include_once 'dosomething_global.features.inc';
  * Implements hook_form_alter().
  */
 function dosomething_global_form_campaign_node_form_alter(&$form, &$form_state, $form_id) {
-  global $user;
   // Check if the user is a mexico or brazil admin
-  if (in_array('brazil admin', $user->roles) || in_array('mexico admin', $user->roles)) {
+  if (dosomething_global_is_regional_admin()) {
     // Get translatable fields.
     $fields = $form_state['field'];
     foreach ($fields as $field) {
-      $translatable = $field['en']['field']['translatable'];
-      // If the field is not translatable disable access 
-      if (!$translatable) {
-        $form[$field[LANGUAGE_NONE]['field']['field_name']]['#access'] = FALSE;
+      $field_name = $field[LANGUAGE_NONE]['field']['field_name'];
+      $field_info = field_info_field($field_name);
+      // If the field is not translatable disable access
+      if (!$field_info['translatable']) {
+        $form[$field_name]['#access'] = FALSE;
       }
     }
   }
+}
+
+/**
+ * Returns if the user is a regional admin
+ *
+ * @param Object $user
+ *   Optional - A specified user to check
+ *
+ * @return SelectQuery object
+ */
+function dosomething_global_is_regional_admin($user = NULL) {
+  if (!isset($user)) {
+    global $user;
+  }
+  return (in_array('brazil admin', $user->roles) || in_array('mexico admin', $user->roles));
 }
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -10,13 +10,17 @@ include_once 'dosomething_global.features.inc';
  * Implements hook_form_alter().
  */
 function dosomething_global_form_campaign_node_form_alter(&$form, &$form_state, $form_id) {
-  // Get translatable fields.
-  $fields = $form_state['field'];
-  $tfields = array();
-  foreach ($fields as $field) {
-    $translatable = $field['en']['field']['translatable'];
-    if ($translatable) {
-      $form[$field['en']['field']['field_name']]['#access'] = FALSE;
+  global $user;
+  // Check if the user is a mexico or brazil admin
+  if (in_array('brazil admin', $user->roles) || in_array('mexico admin', $user->roles)) {
+    // Get translatable fields.
+    $fields = $form_state['field'];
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -5,3 +5,22 @@
...
 
 include_once 'dosomething_global.features.inc';
+
+/**
+ * Implements hook_form_alter().
+ */
+function dosomething_global_form_campaign_node_form_alter(&$form, &$form_state, $form_id) {
+  global $user;
+  // Check if the user is a mexico or brazil admin
+  if (in_array('brazil admin', $user->roles) || in_array('mexico admin', $user->roles)) {
```
angaither: these should probably be defined as a var at the top as a constant so it's easier to add new roles to this and compare the arrays

angaither:
@deadlybutter when I pulled this down on my local it removed every field on the campaign node creation page
screen shot 2015-09-02 at 1 55 27 pm

lib/modules/dosomething/dosomething_global/dosomething_global.module
itsjoekent: this + make a function that checks all the permissions (so we don't need to replicate this every time)

itsjoekent:
@angaither do you have the latest staging DB? thats what happens when its out of date because no fields are marked as translatable

angaither:
@deadlybutter yeah, just pulled it down this morning

itsjoekent:
@angaither okay then thats weird. when you get out of that meeting can we look at your computer and see whats going on?

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -5,3 +5,22 @@
...
+    // Get translatable fields.
+    $fields = $form_state['field'];
+    foreach ($fields as $field) {
+      $translatable = $field['en']['field']['translatable'];
+      // If the field is not translatable disable access
```
angaither: instead of checking the lang @sergii-tkachenko suggested testing with `field_info`
https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_global/dosomething_global.drush.inc#L145-L152

itsjoekent: Yeah I couldn't get field info to work. Drupal docs & that commit weren't much help cause everything looked :ok: 

edit: to add, I figured there will always be an english lang available to read from and it was :ok:

itsjoekent:
I think were good now @angaither

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -5,3 +5,37 @@
  */
 
 include_once 'dosomething_global.features.inc';
+define(GLOBAL_ADMIN_ROLES, ['mexio admin, brazil admin]');
```
angaither: `define(GLOBAL_ADMIN_ROLES, ['mexio admin, brazil admin']);`

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -5,3 +5,37 @@
...
+ * @param Object $user
+ *   Optional - A specified user to check
+ *
+ * @return SelectQuery object
+ */
+function dosomething_global_is_regional_admin($user = NULL) {
+  if (!isset($user)) {
+    global $user;
+  }
+  return in_array('brazil admin', GLOBAL_ADMIN_ROLES);
```
angaither: only testing brazil admin here

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -5,3 +5,38 @@
...
+ *   Optional - A specified user to check
+ *
+ * @return SelectQuery object
+ */
+function dosomething_global_is_regional_admin($user = NULL) {
+  if (!isset($user)) {
+    global $user;
+  }
+  $roles = explode(",", GLOBAL_ADMIN_ROLES);
+  return in_array('brazil admin', $roles);
```
angaither: this will always return true if you are just testing if brazil admin is in the `global_roles` array

itsjoekent:
@angaither fixed!

angaither:
@deadlybutter this isn't working for mexico admins

itsjoekent:
Typo in mexico -- apparently my test account had mexico & brazil admin checked so I didn't notice

angaither:
:+1:

itsjoekent:
:boom: :tada: :confetti_ball: :boom:

PR #5007, opened 2015-09-02
Put hardcoded share language through the t() function

### LINKED ISSUE #4998: problem share copy not in t()
[this copy](https://github.com/DoSomething/phoenix/blob/4934d70f3d69508146c4fb7cb9abdf77b25b4594/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc#L142) should be wrapped should be in the t function 

@mikefantini found this by searching from the [sample list of things to translate](https://github.com/DoSomething/phoenix/issues/4988#issuecomment-137147053)

sbsmith86:
Fixes #4998 

Wrapped all the hardcoded share language in the t() function so it can be translated.


angaither:
:+1: :earth_asia:

PR #5009, opened 2015-09-02
Update to fix bug with URI in usage of RB when not instance of class.

### LINKED ISSUE #4977: URI property in Reportback response object broke and only outputting ".json"
Figure out how it broke and fix it!

weerd:
### Fixes #4977

This PR fixes a bug where Reportback object in ReportbackItem response is not an instance of Reportback class, so it doesn't contain the `uri` property. This update addresses adding `uri` property if it doesn't exist.

---

@angaither


angaither:
:+1:

PR #5014, opened 2015-09-03
Update to pass correct data type to entity func to avoid php warnings.

### LINKED ISSUE #5013: PHP Warning: array_flip bug in API
Keep getting a warning in PHP related to an incorrect parameter being passed to `array_flip()` function in entity controller.

weerd:
### Fixes #5013

Within the `transformReportbackItem()` method, when no **kudos** on an item, `null` was being passed to the `array_flip()` function as the data got passed to the `EntityAPIController`, since it only checks for if the variable is `!empty`. This fixes it by setting the variable to an empty array if item has no kudos data.

---

@angaither


angaither:
:+1:

PR #5017, opened 2015-09-03
Fix broken revealer galleries.

### LINKED ISSUE #4939: taxonomy collections displaying too many capaigns
/volunteer/animals should show 9 campaigns and then with view more load in 2 rows of campaigns at a time (so 8 more at a time)

instead right now all campaign tiles are displaying

DFurnes:
# Changes

Fixes #4939, by properly filtering the DOM query (rather than filtering it's children). 

For review: @weerd


weerd:
hokay! :+1:

PR #5018, opened 2015-09-03
Adding the dosomething_auto_translate module

blisteringherb:
- This module will collect all nodes with type campaign and add
  google translations for pt-br and es-mx.
- To be run with drush php-eval
- Should only be enabled by hand
- Requires the api key to be added manually to the environment where it's run

@angaither


lib/modules/dosomething/dosomething_auto_translate/dosomething_auto_translate.module
```diff
@@ -0,0 +1,74 @@
+<?php
+
+/**
+ * @file dosomething_auto_translate.module
+ * This module uses the Google Translate API to translate campaign nodes.
+ */
+
+/**
+ * @file dosomething_auto_translate.module
+ * TODO: This doesn't scale.
```
angaither, line 10: maybe just set a limit on how many nodes it gets?

angaither:
:+1: 
how do you run this?

blisteringherb:
I'll run it with `drush php-eval "dosomething_auto_translate_campaigns();"` In the future I can probably create a drush command with arguments and such, but I just wanted a quick and dirty way to run this for testing on Thor.

PR #5025, opened 2015-09-04
Removes metatag and revision access from regional admins

### LINKED ISSUE #4934: br admin and mx admin should only have access to br and mx fields

itsjoekent:
#### What's this PR do?

Removes access from editing meta tags and content revisions from regional admins

Create a new user and give them the a regional admin role (Mexico or Brazil admin) and check if this box 
image

is at the bottom of this page
/node/add/campaign

The content revision permissions did not remove the revision form from node/add, had to manually remove access in the form alter instead. 

Fixes #4934 
https://github.com/DoSomething/phoenix/issues/4934#issuecomment-137541461

screen shot 2015-09-04 at 10 52 12 am

@angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -22,6 +22,7 @@ function dosomething_global_form_campaign_node_form_alter(&$form, &$form_state, 
         $form[$field_name]['#access'] = FALSE;
       }
     }
+    $form['revision_information']['#access'] = FALSE;
   }
 }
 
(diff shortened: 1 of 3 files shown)


angaither:
:+1:

Issue #5024, opened 2015-09-04
Langauge index errors on dosomething_campaign


angaither:
`Notice: Undefined index: und in dosomething_campaign_get_hot_time_left() (line 622 of /var/www/dev.dosomething.org/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc).`

angaither:
also seeing
`Notice: Undefined index: 0 in dosomething_campaign_is_hot_module()`

angaither:
`Notice: Undefined index: und in dosomething_campaign_run_preprocess_winners() (line 133 of /var/www/dev.dosomething.org/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module).`

angaither:
`Notice: Undefined index: und in dosomething_campaign_run_preprocess_klout_gallery() (line 153 of /var/www/dev.dosomething.org/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module).`

itsjoekent:
I can look at the hot module stuff, might be related to recent changes I made

itsjoekent:
I'm not seeing this occurring right now, but there are other important dblog errors that have to be resolved. I'll make another issue, @angaither should we close this until we see these errors popup again?

edit; new issue: https://github.com/DoSomething/phoenix/issues/5866

angaither:
@deadlybutter seeing the klout gallery and winner errors on thor
along with plenty undefined notices
https://thor.dosomething.org/us/campaigns/dont-be-sucker
https://thor.dosomething.org/us/campaigns/comeback-clothes

itsjoekent:
importing tor content on my local to see if i can reproduce

PR #5029, opened 2015-09-04
Check to see if values are set in high season

angaither:
don't test high/win season dates if dates are not set on a campaign node
Refs #5024

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -62,19 +62,16 @@ function dosomething_campaign_preprocess_node(&$vars) {
     return;
   }
 
-
-  // If there are dates set, check hot/win states.
-  if (isset($vars['field_high_season'][0])) {
-    // If the hot module is enabled and the high season is set on a campaign.
-    if (dosomething_campaign_is_hot_module($vars)) {
-      dosomething_campaign_preprocess_hot_module($vars, $wrapper);
-    }
-
-    // If the win module is enabled
-    if (dosomething_campaign_is_win_module($vars)) {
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -62,14 +62,18 @@ function dosomething_campaign_preprocess_node(&$vars) {
...
-  // If the hot module is enabled and the high season is set on a campaign.
-  if (dosomething_campaign_is_hot_module($vars)) {
-    dosomething_campaign_preprocess_hot_module($vars, $wrapper);
-  }
 
-  // If the win module is enabled
-  if (dosomething_campaign_is_win_module($vars)) {
-    dosomething_campaign_preprocess_win_module($vars, $wrapper);
+  // If there are dates set, check hot/win states.
+  if (isset($vars['field_high_season'][0])) {
```
DFurnes: Would it make sense to put this in `dosomething_campaign_is_hot_module` and `dosomething_campaign_is_win_module`, so those methods always return false if they don't have the proper variables to check?

angaither: The only reason I didn't go that route was because it was going to check the same thing in both functions. I'm down for either
`¯\_(ツ)_/¯`

DFurnes: I think it should be negligible performance-wise, and it makes things a bit more readable/defensive against future misuse.

angaither: :+1:

angaither:
@DFurnes updated

DFurnes:
:boat: :runner:

PR #5030, opened 2015-09-04
One point graph

### LINKED ISSUE #4925: Hot Module: Handle graph when there is only one data point
The hot module graph is a line graph. Line graphs require at least two points to draw a line between. We are seeing a case arise when a new campaign is launched in the hot phase and there is only one day worth of reportback data. The graph doesn't plot anything because there are not two points to draw a line between.

The proposed solution is that, when there is only one point of data, adding in a dummy point of data for the day before with 0 reportbacks. 

Example. 

When the new campaign launches in the hot phase, the data arrays look like: 

```
$dates = ['08/01/15'];
$quantity = [2];
```

Then we add points to the arrays so that they now look like:

```
$dates = ['07/31/15', '08/01/15'];
$quantity = [0, 2];
```

This should allow the graph to properly plot the data. 

cc @mikefantini

sbsmith86:
#### What's this PR do?

In this PR I check to see if we only have one point of data, and if so explicitly draws that point on the graph. I also added some sanity checks on the backend side of things to see make sure we even have data to pass to the frontend. I was seeing errors in the drupal log when we were trying to send variables in the json that weren't defined because the there was no data for that campaign.

I suggest starting in `HotModule.js`. The `hasOnePoint()` function does the testing and then on line 300, we do the actual drawing of the point.

Create new campaign and launch it in the hot phase, then generate some reportbacks on it and run cron. Then check to see if the graph shows just the one progress point. Compare that to another campaign with more data to make sure we see the pretty line graph.

This PR solves an issue we were seeing with the hot module graph when there was only one point of data. This would pop up, for example, when a new campaign is launch in the hot phase and starts getting reportbacks that day. It seems that issue is that the graph is a line graph, and it expects at least two points on the graph to draw a line between. 

Fixes #4925 

ONE POINT!

screen shot 2015-09-04 at 12 52 50 pm
#### Questions:
- Is there a blog post? Nope 
- Does the knowledge base need an update? NOOOOPPE
- Does this add new (Python) dependencies which need to be added to chef? Python? What's that?

@DoSomething/front-end

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -322,8 +322,8 @@ function dosomething_campaign_preprocess_hot_module(&$vars, &$wrapper) {
     array('dosomethingCampaign' =>
       array(
         'goals' => array(
-          'dates' => (isset($dates)) ? json_encode($dates) : null,
-          'quantities' => (isset($quantities)) ? json_encode($quantities) : null,
+          'dates' => (isset($dates)) ? $dates : null,
+          'quantities' => (isset($quantities)) ? $quantities : null,
           'highSeason' => array(
             'start' => $start_date->format($date_format),
             'end' => $end_date->format($date_format)
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -311,6 +311,7 @@ function dosomething_campaign_preprocess_hot_module(&$vars, &$wrapper) {
                                FROM dosomething_reportback_progress_log
                                WHERE nid = :nid
                                GROUP BY date(from_unixtime(timestamp));", array(':nid' => $vars['nid']))->fetchAll();
+
   foreach($quantity_by_date as $date) {
     $maxdate = new DateTime($date->maxdate);
     $dates[] = $maxdate->format($date_format);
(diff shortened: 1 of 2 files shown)


lib/themes/dosomething/paraneue_dosomething/js/campaign/HotModule.js
```diff
@@ -103,6 +103,13 @@ const createQuantityArray = function(dates, dateLabels, quantities) {
 };
 
 /*
+ * Checks to see if we only have one point of data.
+ */
+const hasOnePoint = function(data) {
+  return (JSON.parse(data.quantities) && JSON.parse(data.quantities).length === 1) ? true : false;
```
DFurnes: I'd recommend running `JSON.parse()` on the `progressData` back when you load it from the settings object, so you don't have to re-parse it each time you want to do something with it.

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -321,8 +322,8 @@ function dosomething_campaign_preprocess_hot_module(&$vars, &$wrapper) {
     array('dosomethingCampaign' =>
       array(
         'goals' => array(
-          'dates' => json_encode($dates),
-          'quantities' => json_encode($quantities),
+          'dates' => (isset($dates)) ? json_encode($dates) : null,
```
DFurnes: Actually, is the `json_encode` necessary here? I _think_ it might be safe to remove and then you wouldn't have to use `JSON.parse()` on the client-side.

DFurnes:
This is a powerful pull-request. :muscle:

angaither:
notable PR. WHAT IS PYTHON

DFurnes:
:snake: :interrobang:

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
sbsmith86: @DFurnes my bad. I didn't see this comment! Yea, I don't think it is necessary. Going to remove it and rework the client side stuff. Thanks!

Issue #5020, opened 2015-09-03
BR/MX admin need to to be able to edit content not created by them


mikefantini:
We'll need to add these permissions to BR/MX admin roles:
- [ ] Campaign: Edit any content
- [ ] Static content: Edit any content

keep in mind though, that Brazil admins should only have access to Brazilian Portuguese fields and Mexico admins should only have access to Mexican Spanish fields from @blisteringherb's [comment](https://github.com/DoSomething/phoenix/issues/4934#issuecomment-136467741) on #4934:

"I'm thinking that putting in a hook_menu_alter for node/%/edit that triggers the add translation chain if the allowed translation hasn't yet been created for that node and if it already exists it will send the user to node/%/edit/es-mx so they never actually get to the edit screen for the non-allowed translations.

For instance if a user with the Brazil admin role tries to visit `node/36/edit`, they'll be redirected to `node/36/edit/pt-br` if the translation already exists, and if not they'll go to `node/57/edit/add/en/pt-br`.  The other languages in the upper right corner of the node edit screen will be hidden except for those that they're allowed to see, and they won't have access to the translate tab."

note: whether or not they need access to `Home: Edit any content` is dependent on the decision made in #4968

itsjoekent:
Just to clarify/update after talking w/ Fantini 

The problem this issue needs to solve is how the URL redirects when you click 'Edit' as a regional admin. Fields are already hidden on the node edit page due to this PR: https://github.com/DoSomething/phoenix/pull/5004

Also, the sub header fields need to display properly. EG: I'm currently a MX admin and I can click the PT/EN translations (should not be able too)-
<img width="347" alt="screen shot 2015-09-09 at 11 19 58 am" src="https://cloud.githubusercontent.com/assets/897368/9765977/c4a05744-56e4-11e5-92ae-4ce5fa1f53ae.png">

itsjoekent:
~~Mapping the URL's will be based on the variables soon to be created in this PR:~~

~~https://github.com/DoSomething/phoenix/pull/5050#discussion-diff-39051771~~

Never mind, thats not happening yet. Just going to hardcode the values for now.

https://github.com/DoSomething/phoenix/pull/5050#discussion_r39073166

PR #5050, opened 2015-09-09
Fix campaign saves dying

### LINKED ISSUE #5042: save on a new campaign dies!

angaither:
#### What's this PR do?

Sets the default language of campagin nodes to the site's default language (english)
Checks campaign nodes on save, and set's the language to the correct role of the user.

Create a campaign node as a brazil/mexico admin and check the lang after save.
Also create a campaign node ad a regular editor and check the language is english (and the page doesn't die!) 

Campaign node saves were dying in a fire. 

Fixes #5042

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -6,6 +6,19 @@
 
 include_once 'dosomething_global.features.inc';
 define('GLOBAL_ADMIN_ROLES', 'mexico admin,brazil admin');
+define('GLOBAL_ADMIN_LANGUANGES', serialize(['es-mx' => 'mexico admin', 'pt-br' => 'brazil admin']));
+
+/**
+ * Implements hook_node_presave().
+ */
+function dosomething_global_node_presave($node) {
+  $user = user_load($node->uid);
+  // If we are saving a campaign and the user is a global admin.
+  if ($node->type == 'campaign' && dosomething_global_is_regional_admin($user)) {
+    // Set the node language to the author's lang.
+    $node->language = dosomething_global_get_user_language($user);
+  }
+}
 
 /**
  * Implements hook_form_alter().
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -6,6 +6,19 @@
 
 include_once 'dosomething_global.features.inc';
 define('GLOBAL_ADMIN_ROLES', 'mexico admin,brazil admin');
+define('GLOBAL_ADMIN_LANGUANGES', serialize(['es-mx' => 'mexico admin', 'pt-br' => 'brazil admin']));
```
blisteringherb, line 9: I've been thinking that since we're eventually going to have to point to language definitions in a bunch of places, maybe we should have a strongarm variable with something like this:

```
  $strongarm = new stdClass();
  $strongarm->disabled = FALSE; /* Edit this to true to make a default strongarm disabled initially */
  $strongarm->api_version = 1;
  $strongarm->name = 'dosomething_global_language_map';
  $strongarm->value = array(
    'es-mx' => array(
      'display_name' => 'Spanish, Mexico',
      'default_roles' => array('mexico admin'),
      'country' => 'MX',
    ),
    'pt-br' => array(
      'display_name' => 'Portuguese, Brazil',
      'default_roles' => array('brazil admin'),
      'country' => 'BR',
    ),
    'en' => array(
      'display_name' => 'English',
      'default_roles' => array(),
      'country' => 'US'
    ),
    'en-global' => array(
      'display_name' => 'English, Global'
      'default_roles' => array(),
    ),
  );
  $export['dosomething_global_language_map'] = $strongarm;

  return $export;
```

And then retrieve the values with variable_get() and maybe a helper function or two for transformation purposes like these:

```
  $lang_map = variable_get('dosomething_global_language_map');

  function dosomething_global_get_country_mappping() {
    $countries = array();
      foreach ($dosomething_global_language_map as $lang => $data) {
        $countries[$data[$country]] =  $dosomething_global_language_map[$lang];
      }
    return $countries
  }
```

So then when we can use these language definitions as a sort of one-stop rule set which we can use here and also when we get the user's country from the GeoIP data or the URL.

Any thoughts?

angaither, line 9: @blisteringherb that sounds :ok_hand:

blisteringherb, line 9: @angaither Adding that in can be part of refactoring or at least be added in when we do the session/language switching code in conjunction with Fastly.

angaither, line 9: Yeah, that's what I was thinking, ok to merge this as is and create an issue for the strongarm lang switching?

blisteringherb:
:+1:

lib/modules/dosomething/dosomething_global/dosomething_global.module
itsjoekent, line 9: I'm sort of blocked on this being implemented in (for re-directs) https://github.com/DoSomething/phoenix/issues/5020

I can hardcode the values for now and make it dynamic later when we add in the strongarm

PR #5056, opened 2015-09-09
Adding capacity to post to mobile only campaigns.

### LINKED ISSUE #4986: Allow for mobile only campaigns
Some campaigns are only going to be displayed in the mobile app and not in the web app at all.
The easiest solution is to leave campaigns unpublished, and return unpublished campaigns with the `on_mobile_app` field checked via the api.

Things to check for unpublished campaigns
- [x] users can signup 
- [x] users can reportback 
- [x] kudos work

weerd:
### Fixes #4986

Updating logic in `dosomething_campaign_is_active($node)`  to return `true` when a campaign is on mobile app and may or may not be publsihed.

---

@angaither


angaither:
:+1:

Issue #4886, opened 2015-08-24
Display finder content to users who visit a country's home page in the appropriate order


blisteringherb:
Users should see:
1. native campaigns in language
2. global campaigns in language
3. other global campaigns in global english (publish global campaigns initially)

This has to be changed in the finder code, and there may be changes that need to be made in the view.

blisteringherb:
@julielorch Is the logic for #2 necessary for this iteration since we'll only have global campaigns in English for the launch?

mikefantini:
@blisteringherb just chatted with @julielorch and let's just focus on:
1. campaigns that exist in user language (regardless of source language)
2. campaigns that exist in Global English

since it seems harder to add an extra layer of prioritization (campaigns that have source language as MX SP prioritized over campaigns with source language US EN that are translated into MX SP) let's just leave that out for now.

so the only prioritization would be show campaigns that exist in MX SP before showing campaigns that exist in GL EN.

anything else you need on that @blisteringherb?

blisteringherb:
Based on the solution I put into #5123, adding the other global languages shouldn't be too heavy of a lift, but we will need to add "pt-global" and "es-global" as languages. The testing will probably be the hardest part because we'll have to have significant numbers of campaigns that are only translated into certain languages. I can expand on the dosomething_auto_translate() module to facilitate that if everyone thinks it's a good use of time.

joolaay:
@mikefantini @blisteringherb ^ "pt-global" would be a global campaign, in portuguese, that hasn't yet been localized?

blisteringherb:
@julielorch Yes, that sounds right to me.

mikefantini:
as we talked about IRL let's hold on adding anything for pt-global or es-global. that will be work for future iterations when we decide we want to serve portuguese/spanish speakers globally

PR #5062, opened 2015-09-10
Global finder

blisteringherb:
First pass at language-aware finder
- This code will show the translated nodes in the finder
- Adding a limit and some fixes to the auto_translate module
- Fixing broken images on home page tiles
- Adding filter for language based on user's default language
- Adding Strongarm variable with rules for each language
  - Can be used by any language switching logic

This PR is an intermediary step towards #4886. The finder is still missing infill from global-en campaigns.


lib/modules/dosomething/dosomething_auto_translate/dosomething_auto_translate.module
```diff
@@ -0,0 +1,88 @@
...
+ * @file dosomething_auto_translate.module
+ *
+ */
+function dosomething_auto_translate_campaigns($start = NULL, $limit = NULL) {
+  $query = new EntityFieldQuery();
+
+  $query->entityCondition('entity_type', 'node')
+    ->entityCondition('bundle', 'campaign')
+    ->propertyCondition('status', NODE_PUBLISHED)
+    ->range($start, $limit);
```
angaither, line 18: will this break if no values are passed? maybe set them to numbers instead of `NULL` by default?

lib/modules/dosomething/dosomething_auto_translate/tests/dosomething_auto_translate.test
```diff
@@ -0,0 +1,41 @@
+<?php
```
angaither, line 1: :+1: :memo: testing!!! :clap:

angaither:
@blisteringherb looks good other than the comment
also, do you mean to merge this into `dev` or `global`?

lib/modules/dosomething/dosomething_auto_translate/dosomething_auto_translate.module
blisteringherb, line 18: If range() gets NULL, it'll remove the range query and execute the query with no limit. I could add a default number to $limit in the case that someone passes in a $start but no $limit.

blisteringherb:
Whoa! Thanks for the reminder. I meant to merge into global. I think I might have to close this PR and open a new one.  Boo. :-1:

lib/modules/dosomething/dosomething_auto_translate/dosomething_auto_translate.module
angaither, line 18: :+1:

angaither:
ruffffff

Issue #5044, opened 2015-09-08
Campaign Finder Intermittently Only Showing 1/2 of Causes and 1/2 of Types


namimody:
The campaign finder dropdown should look like this:

![image](https://cloud.githubusercontent.com/assets/12396076/9743935/febc9784-5636-11e5-864b-8e73da7255ca.png)

But it looks like this for some users:
![image](https://cloud.githubusercontent.com/assets/12396076/9743993/4850af8e-5637-11e5-8e57-5c403ece4fd9.png)

As a note, these screenshots were pulled from different comps within minutes of each other, so it seems like it's working for some users. As a note, the missing content occurred in both logged in and logged out views.

DFurnes:
Do you have any details on the computers where it is and isn't working (i.e. browser, browser version, operating system)?

DFurnes:
@namimody Are you communicating in morse-code via issue assignments?

namimody:
@DFurnes hahahha just trying to resolve a confusion...by creating confusion.

namimody:
@DFurnes not working on: Chrome Version 45.0.2454.85 (64-bit) / Yosemite

Working on: 0S X version 10.9.5 and chrome Version 44.0.2403.157 (64-bit)

mikefantini:
worked for me as both an admin and anonymous user in the "working" versions above

mikefantini:
now that i updated my chrome to 45.0.2454.85 (64-bit) seeing this as well both logged in and logged out

![image](https://cloud.githubusercontent.com/assets/6330971/9768458/d83b63ae-56f1-11e5-90e5-cb2c3279defd.png)

namimody:
Unsure of how helpful this, but seems like it things appear correctly on
Safari 8.0.6

On Wed, Sep 9, 2015 at 12:54 PM, mikefantini notifications@github.com
wrote:

> now that i updated my chrome to 45.0.2454.85 (64-bit) seeing this as well
> both logged in and logged out
> 
> [image: image]
> https://cloud.githubusercontent.com/assets/6330971/9768458/d83b63ae-56f1-11e5-90e5-cb2c3279defd.png
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/DoSomething/phoenix/issues/5044#issuecomment-138973391
> .

## 

Namita (Nami) Mody

Junior Product Manager
@dosomething http://twitter.com/dosomething @imanami13
https://twitter.com/imanami13
[...]

PR #5065, opened 2015-09-10
Fix issue with columns disappearing in Chrome 45.

DFurnes:
See #5044. Chrome 45 includes a new CSS column implementation, which seems to break with block elements. Since overflow only affects block and inline-block elements, I also had to switch how we were hiding checkboxes.

Tested in Chrome 45, Chrome Canary, Firefox latest, Safari latest, IE8, IE9, IE10, IE11.


sbsmith86:
:tada:

PR #5068, opened 2015-09-10
First pass at language-aware finder

blisteringherb:
- This code will show the translated nodes in the finder
- Adding a limit and some fixes to the auto_translate module
- Fixing broken images on home page tiles
- Adding filter for language based on user's default language
- Adding Strongarm variable with rules for each language
  - Can be used by any language switching logic


angaither:
:+1:

PR #5074, opened 2015-09-10
Add correct lang attribute to HTML element.

DFurnes:
Discovered during Global Tech meeting... we're currently hardcoding `lang` and `dir` attributes on the HTML element. That's not #global. :globe_with_meridians: 

For review: @angaither


angaither:
lovely :+1:

PR #5079, opened 2015-09-10
Redirects global lovely

### LINKED ISSUE #5020: BR/MX admin need to to be able to edit content not created by them

itsjoekent:
#### What's this PR do?
- Exports new regional admin permissions to edit any campaign content and static content
- Redirects users to the proper translation page if the node source language does not match the user language 

dosomething_campaign.module -- campaign_init()
- Create a new campaign 
- Under a new account with a different language, edit the campaign. 
  - It should redirect you to a url containing /add/<node language>/<user language>
- Save the campaign
- Hit 'Edit' again
  - It should redirect you to a url ending with your user language 

Multiple hooks were tried and discussed in #global but in the end we decided to go with hook_init() 
#### Relevant tickets

Fixes #5020 

@angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -26,6 +26,37 @@ function dosomething_campaign_form_campaign_node_form_alter(&$form, &$form_state
   $form['#after_build'][] = 'dosomething_campaign_form_campaign_node_form_after_build';
 }
 
+function dosomething_campaign_init() {
+  // Get the URL and split it up
+  $current_path = current_path();
+  $path_items = explode('/', $current_path);
+
+  // Verify we're dealing with a node edit with no translation specification in the URL
+  if ($path_items[0] == "node" && is_numeric($path_items[1]) && count($path_items) == 3) {
+    // Load the page node and user
+    $nid = $path_items[1];
+    $node = node_load($nid);
+    global $user;
+
+    // Load the languages being used
+    $src_language = $node->language;
+    $user_language = $user->language;
+
+    // If we're not working on a translation, no need to re-direct.
+    if ($src_language != $user_language) {
+      // Determines if the url needs the /add/source/ component
+      $new_translation_prefix = isset($node->translations->data[$user_language]) ? '' : 'add/' . $src_language . '/';
+
+      // Append the user language
+      $lang_append = $new_translation_prefix . $user_language;
+
+      // Build complete URL and redirect
+      $new_url = str_replace(' ', '', '/node/' . $nid . '/edit/' . $lang_append);
+      drupal_goto($new_url);
+    }
+  }
+}
+
 /**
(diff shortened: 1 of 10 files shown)


DFurnes:
Solid branch name. :trophy:

itsjoekent:
credit goes to @angaither

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -26,6 +26,37 @@ function dosomething_campaign_form_campaign_node_form_alter(&$form, &$form_state
   $form['#after_build'][] = 'dosomething_campaign_form_campaign_node_form_after_build';
 }
 
+function dosomething_campaign_init() {
+  // Get the URL and split it up
+  $current_path = current_path();
```
angaither: @deadlybutter this looks like the older version of the code

itsjoekent: yeah it is :rage:

itsjoekent:
@angaither I think its ready

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.strongarm.inc
```diff
@@ -15,7 +15,7 @@ function dosomething_campaign_strongarm() {
   $strongarm->api_version = 1;
   $strongarm->name = 'entity_translation_settings_node__campaign';
   $strongarm->value = array(
-    'default_language' => 'xx-et-default',
+    'default_language' => 'xx-et-author',
```
angaither: @deadlybutter we want to keep this as `default` (it's a setting in that entity translate tab) 
you can just make the change to the file manually, and amend the commit, or export the setting again
http://superuser.com/questions/229290/how-to-amend-the-last-commit-to-un-add-a-file

angaither:
:+1:

PR #5080, opened 2015-09-10
Display Partner logos on Static Content Types

### LINKED ISSUE #5037: static content partner logo
for the yahoo partnership we need the ability to add partner logo on static content
this field is already there, but needs to be output on the template

sbsmith86:
#### What's this PR do?
- Displays sponsor logos if they exists in the global header, which is what is used on static content pages. 
- Added some null checks for variables being sent to fact pages that were throwing error if no sponsor was set. 
- Adds the info bar to fact pages. 

`preprocess.inc` is where variables are preprocessed and sent to the header theme function.

Add a sponsor to a static content page, and you should see the sponsor logo displayed on the page. 

Fixes #5037

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/fact/node--fact_page.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/fact/node--fact_page.tpl.php
@@ -22,7 +22,7 @@
         <h2 class="header__subtitle"><?php print $subtitle; ?></h2>
       <?php endif; ?>
 
-      <?php print $promotions; ?>
+      <?php if (isset($promotions)) { print $promotions; } ?>
     </div>
   </header>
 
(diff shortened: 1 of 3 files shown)


lib/themes/dosomething/paraneue_dosomething/templates/fact/node--fact_page.tpl.php
```diff
@@ -100,4 +100,8 @@
       </div>
     </div>
   <?php endif; ?>
+
+  <?php if ($info_bar): ?>
```
angaither, line 104: what's an `$info_bar`?

angaither:
:+1:

lib/themes/dosomething/paraneue_dosomething/templates/fact/node--fact_page.tpl.php
sbsmith86, line 104: @angaither it's the little bar above the footer that says "In partnership with..."

angaither, line 104: enjoyable info!

Issue #4989, opened 2015-09-02
Make DS Campaign Config translatable


mikefantini:
need to make all copy that's in DoSomething Configuration -> DoSomething Campaign translatable

related to #4988

angaither:
we might need to turn on `variable_translate` and hook into that for dosomething helpers config variables. 

we also might want to create custom forms per language we are translating to (depending who is doing the translation)

PR #5093, opened 2015-09-11
Updating t() function to not pass variable without sanitation.

weerd:
#### What's this PR do?

This PR wraps updates how the `t()` is used in the DoSomething Config -> Campaign config, to pass the variable value indirectly so the function can properly sanitize the value.

In the main **dosomething_campaign.admin.inc** file and just review that the updated code looks correct.

Refs #4989

---

@angaither


angaither:
:ship:

PR #5097, opened 2015-09-11
Update roles lang to use the map

### LINKED ISSUE #5054: create a strongarm langage/role switcher
I've been thinking that since we're eventually going to have to point to language definitions in a bunch of places, maybe we should have a strongarm variable with something like this:

```
  $strongarm = new stdClass();
  $strongarm->disabled = FALSE; /* Edit this to true to make a default strongarm disabled initially */
  $strongarm->api_version = 1;
  $strongarm->name = 'dosomething_global_language_map';
  $strongarm->value = array(
    'es-mx' => array(
      'display_name' => 'Spanish, Mexico',
      'default_roles' => array('mexico admin'),
      'country' => 'MX',
    ),
    'pt-br' => array(
      'display_name' => 'Portuguese, Brazil',
      'default_roles' => array('brazil admin'),
      'country' => 'BR',
    ),
    'en' => array(
      'display_name' => 'English',
      'default_roles' => array(),
      'country' => 'US'
    ),
    'en-global' => array(
      'display_name' => 'English, Global'
      'default_roles' => array(),
    ),
  );
  $export['dosomething_global_language_map'] = $strongarm;

  return $export;
```

And then retrieve the values with variable_get() and maybe a helper function or two for transformation purposes like these:

```
  $lang_map = variable_get('dosomething_global_language_map');

  function dosomething_global_get_country_mappping() {
    $countries = array();
      foreach ($dosomething_global_language_map as $lang => $data) {
        $countries[$data[$country]] =  $dosomething_global_language_map[$lang];
      }
    return $countries
  }
```

So then when we can use these language definitions as a sort of one-stop rule set which we can use here and also when we get the user's country from the GeoIP data or the URL.

angaither:
Refactor the language and user role mapping to use the language map strongarm
Fixes #5054

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -90,6 +90,12 @@ function dosomething_global_is_regional_admin($user = NULL) {
   return FALSE;
 }
 
+/**
+ * Get the roles that are defined as regional admins.
+ *
+ * @return - array
+ *  The roles that are regional admins.
+ */
 function _dosomething_global_get_regional_roles() {
   $languages = variable_get('dosomething_global_language_map', '');
   foreach ($languages as $language) {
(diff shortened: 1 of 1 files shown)


blisteringherb:
:+1:

PR #5098, opened 2015-09-11
Fix reportback items

chloealee:
Ref https://github.com/DoSomething/api/issues/8
#### What's this PR do?

Updates to the correct endpoint to get northstar user information (we have the drupal id so uses drupal id to look up user where before we were trying to use the drupal id with the `GET /users/_id/<user_id>` endpoint).


chloealee:
@weerd just added to Reportback.php as well and updated to return with `drupal_id` and `id` whenever you get a chance to peek!

weerd:
Lovely! :dancer: :+1:

PR #5105, opened 2015-09-14
Exporting langugage detection settings to strongarm

### LINKED ISSUE #5000: Set language detection to honor user's language preference
At admin/config/regional/language/configure, you can enable which type of language detection is enabled and change the weight for which detection gets priority. The user's language preference in their settings should take precedence for both interface and content translation. The variables to export are `locale_language_providers_weight_language` and `locale_language_providers_weight_language`.

blisteringherb:
- Enabling language detection based on user's preferences
- For the interface the hierarchy is:
  - User
  - URL
  - Session
  - Browser
  - Default
- For the content the hierarchy is:
  - URL (URL is not enabled in this commit)
  - User
  - Session
  - Browser
  - Interface
  - Default

Resolves #5000


angaither:
:+1:

PR #5106, opened 2015-09-14
Using the default language image for search results

### LINKED ISSUE #5059: searching "doe" on thor as BR gives broken results
I'm not sure exactly what the issue is here (it may be multiple issues), but one thing that appears to have happened is that in the translation:
1. not all fields were translated over
2. the Google Translate script attempted to translate images as well (meaning drupal would be looking to reference images that don't exist and so there would be no image associated with the translation)

I believe this what is causing the following to happen when I search "doe" as a user with language set to Brazilian Portuguese:

image

blisteringherb:
- Translated nodes might not always have a unique cover image,
  so this code will use the node's default language image if
  the current langugage does not have a cover image.
- This construct probably appears in multiple places, and we could
  eventually refactor this with a helper like dosomething_helpers_get_cover_image()

Resolves #5059


angaither:
:+1: looks good, but can you make a new issue to refactor this later?

PR #5111, opened 2015-09-14
Fixes node_save() error

### LINKED ISSUE #5048: Errors on campaign node_save() as admin
`Warning: array_key_exists(): The first argument should be either a string or an integer in drupal_array_get_nested_value() (line 6774 of /var/www/dev.dosomething.org/html/includes/common.inc).`

when you save a campaign node as a brazil/mexico admin getting tons of these errors on the global branch.
this seems to be the offending code

```
    foreach ($fields as $field_name => $field) {
      $field_info = field_info_field($field_name);
      // If the field is not translatable disable access
      if (!$field_info['translatable']) {
        $form[$field_name]['#access'] = FALSE;
      }
    }
```

### LINKED ISSUE #5061: Thor not saving content in BR Portuguese
I just created a translation for Brazilian Portuguese of node 5613 with Title "Pretend this is Portuguese" and CTA "Pretend this is Portuguese" and saved it.

However, when I go back to edit it, I see this:

image

itsjoekent:
#### What's this PR do?

Fixes the node save error regional admins experience in #5048 and consequently also fixes #5061  

Edit a campaign as a regional admin and save it. Verify 
1. This error does not popup,  `
   Warning: array_key_exists(): The first argument should be either a string or an integer in drupal_array_get_nested_value() (line 6774 of /var/www/dev.dosomething.org/html/includes/common.inc).`
2. The content is actually saved 

The code used to hide non translatable fields was also targeting #parent and this caused the error. It was discovered earlier, and further tested after making this change that the failures occurring here were the reason translations didn't save in #5061 

Fixes #5048 
Fixes #5061

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -59,7 +59,7 @@ function dosomething_global_form_campaign_node_form_alter(&$form, &$form_state, 
     foreach ($fields as $field_name => $field) {
       $field_info = field_info_field($field_name);
       // If the field is not translatable disable access
-      if (!$field_info['translatable']) {
+      if (!$field_info['translatable'] && strpos($field_name, '#') === FALSE) {
         $form[$field_name]['#access'] = FALSE;
       }
     }


blisteringherb:
:+1:

PR #5100, opened 2015-09-14
Kudos fix

chloealee:
#### What's this PR do?

Adds more information about the user to Kudos
#### Where should the reviewer start?

`Kudos.php` lines 104-115
#### How should this be manually tested?

via Postman/Paw - hit the `/kudos` endpoint to view all user info. 
#### What are the relevant tickets?

Fixes https://github.com/DoSomething/api/issues/8

PR #5118, opened 2015-09-14
Transformer

chloealee:
#### What's the PR do?

Adds the drupal_id to the `Transformer.php`

`Transformer.php` line 461

Hit `/reportback-items.json` endpoint to view the exposed `drupal_id` in the user object.  

Meant to put in this fix #5100


Vagrantfile
```diff
@@ -37,16 +37,16 @@ Vagrant.configure("2") do |config|
...
   # With Varnish
-  config.vm.network :forwarded_port, guest: 6081, host: 9999
+  config.vm.network :forwarded_port, guest: 6081, host: 3333
 
   # Solr.
-  config.vm.network :forwarded_port, guest: 8983, host: 8983
+  config.vm.network :forwarded_port, guest: 8983, host: 4444
 
   config.vm.provision :shell, :inline => 'more /vagrant/scripts/install_complete.txt'
 end
```
weerd, line 52: Should this Vagrant file be committed and updated?

lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -458,6 +458,7 @@ protected function transformReportbackItem($data) {
   protected function transformUser($data) {
     return array(
       'id' => $data->id,
+      'drupal_id' => $data->drupal_id,
```
weerd, line 461: So I don't think we actually want to return the Drupal ID back in the API response. Drupal ID shouldn't mean anything to anyone using the api... is there a specific reason why we need this returned?

If we **do** need it, we could call it something agnostic to the framework, like `legacy_id` or something. But ideally we shouldn't be outputting this back in the JSON API Response...

chloealee, line 461: Oooh okay, my mistake! I thought we wanted this but totally makes sense! Should I just close out this PR?

weerd:
If we don't need the Vagrant update either, then yeah just close it on up :)

PR #5123, opened 2015-09-15
Adding "en-global" campaigns to end of finder results

### LINKED ISSUE #4886: Display finder content to users who visit a country's home page in the appropriate order

blisteringherb:
#### What's this PR do?
- Surfaces en-global content along with the user's language content in the finder
- Adds "en-global" with "OR" to finder query
- Adds a boost query to the user's langugage to prioritize those results over en-global results
- Prerequisites:
  - Translate a significant amount of campaigns, enough to populate finder results for a given set of filters
    - Note: To test this properly you must have a significant amount of campaigns that **do not** have the user's language translation but **do** have the "en-global" translation.
  - Create dummy en-global campaigns
  - Login and set your language to something other than 'en'
- Use the finder as usual.

Resolves #4886

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/finder/SolrAdapter.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/finder/SolrAdapter.js
@@ -48,7 +48,7 @@ define(function(require) {
 
     // This would be awesome with an object key:value, but Solr allows multiple of the same key
     defaultQuery: [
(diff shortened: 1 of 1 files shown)


angaither:
:+1:

PR #5131, opened 2015-09-15
Remove language switcher for non-staff roles.

### LINKED ISSUE #5117: hide language settings from all users except for admins/global admins
authenticated users can change their language currently. we don't want this to be the case right now.

currently being set on https://thor.dosomething.org/user/2987552/edit

image

this should block merging the global branch to dev

admins and global admins should be able to see this on other users' profiles

cc: @angaither

angaither:
Fixes #5117


blisteringherb:
:+1:

PR #5134, opened 2015-09-16
Card font color

### LINKED ISSUE #5132: no info getting pulled into permalink page as owner or non-owner
items missing from owner:
- caption
- author name
- why participated

items missing from non-owner:
- problem
- solution (fact OR statement)
- caption
- author name

owner on staging:

image

owner on prod:

image

non-owner on staging:

image

non-owner on prod:

image

DFurnes:
Fixes #5132. The new `.container.-dark` pattern was setting white font color (because that's a sane default for that pattern), but the `.card` pattern was inheriting that and making text appear invisible.
#### Before

screen shot 2015-09-16 at 1 42 59 pm
#### After

screen shot 2015-09-16 at 1 42 33 pm

---

@DoSomething/front-end


weerd:
:+1:

PR #5135, opened 2015-09-16
Global finder

blisteringherb:
There was one final pull request against the global branch that got approved after we did the merge. This will merge that in.

@angaither


angaither:
:+1:

Issue #5121, opened 2015-09-15
make Global, BR, and MX mailchimp fields required in third-party opt-ins


mikefantini:
related to #5072 

blocked by having values for these

PR #5143, opened 2015-09-17
Adds global mailchimp settings

### LINKED ISSUE #5072: Global MailChimp Settings
Currently a MailChimp list ID (`mailchimp_list_id`: f2fab1dfd4) is set in the Drupal Third Part Opt-In -> General form: https://www.dosomething.org/admin/config/dosomething/opt_in. 
image

This value is included in user_registration transactional message requests sent to Message Broker as a part of the `activity": "user_register` request payload:

```
"mailchimp_list_id": "f2fab1dfd4"
```

The form needs to be expanded to support `mailchimp_list_id` values for:
- Global
- US (United States)
- BR (Brazil)
- MX (Mexico)
- CA (Canada)
- UK (United Kingdom)

Perhaps support for the remaining affiliates should also be added:
- ID (Indonesia)
- etc.

**Related**:
- https://github.com/DoSomething/phoenix/issues/4955

itsjoekent:
#### What's this PR do?

This PR adds mailchimp settings for each country registered. It also maintains the existing mobile opt-in settings for the US but it's clearly labeled how to make it not US specific (For #5121). 

Also adds new helper functions to dosomething_global for retrieving a list of countries and getting a countries language. 

dosomething_signup.admin is where a bulk of the work takes place

Visit /admin/config/dosomething/opt_in

This logic is _not_ fastly based because the user language should have already been setup before this step based on the fastly headers. See #global where I made sure this was :ok: 

Fixes #5072 

This is what the admin UI looks like,
screen shot 2015-09-15 at 2 12 03 pm

This is logging the mailchimp list field name for a US/EN user after registering (to prove the logic is able to build it correctly in dosomething_signup) 
en register
#### Don't merge yet!

Still need to write an install hook which imports the old vars into the new one

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -134,3 +134,24 @@ function dosomething_global_get_user_language($user = NULL) {
   // If we didn't find a user role return english default.
   return 'en';
 }
+
+/*
+ * Returns a list of all the countries defined in the strongarm
+ *
+ * @return array
+ *  All of the supported countries
+ */
+ function dosomething_global_get_countries() {
+   $lang_map = variable_get('dosomething_global_language_map', '');
+   $countries = array();
+   foreach ($lang_map as $lang) {
+     $country = $lang['country'];
+
+     // Prevents global english from slipping through
+     if ($country != NULL) {
+       array_push($countries, $country);
+     }
+   }
+   array_push($countries, "global");
+   return $countries;
+ }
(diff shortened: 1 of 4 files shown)


deezone:
@deadlybutter What's the policy for supporting Global (catch-all for countries we don't support / non-affiliates)? There needs to be a MailChimp setting to support a different MailChimp list for Global transactional message requests.

itsjoekent:
It's cut off in the screenshot but there's a global option at the bottom!

itsjoekent:
@DeeZone All the work here is done, can you verify this is :ok: (the message broker parts)

@angaither ready for review

angaither:
:+1:

Issue #4956, opened 2015-08-28
Global Campaign Signup Transactional Email


mikefantini:
When a user signs up for a campaign, s/he should receive a campaign signup email in the same language as the translation of the campaign s/he signed up for. i.e. if I sign up for the Brazilian Portuguese translation of node 123, I should receive a Brazilian Portuguese campaign signup email
- [x] Brazilian Portuguese campaign signup email
- [x] Mexican Spanish campaign signup email

Also note that we need to make sure we're pulling in the proper fields for this since technically the Brazilian Portuguese and Mexican Spanish content are in different fields than the US English fields.

@sergii-tkachenko @sheydari-ds if you have any thoughts to advise @DeeZone on this please include here.

marahml:
For separate Global transactionals, we currently don't have any new work slotted for v1. Only BR & MX. 

So a total of 8 new templates (4 per country) that are translated and match to new fields is our current focus.

If steps will be different for Global vs. US campaigns - is that something we can solve for on the content side? @mikefantini @DeeZone

mikefantini:
created a new issue for Global English in case we can't get all templates up for launch: #5091

deezone:
@marahml What is the difference between the expected email message sent to users who defined as "GLOBAL" vs "US"?

Is the only difference the MailChimp account / list assignment?

marahml:
I think the main difference would be the merge tags we pull in -- but just basing that off of the issue #5091.

This is lowest priority of all templates - so focus is still on Registration & PW reset currently, then Signup and RB.

mikefantini:
yup lowest priority. the key difference is we'll just need to literally pull from different fields because the content will be stored in different fields for users from US/BR/MX/everywhere else

blisteringherb:
@mikefantini @marahml @DeeZone 

Can I get a list of what fields are are different? I've got the translated version of "field_pre_step_header" and "field_post_step_header" being pushed into the payload. Are those fields not used for br and mx?

mikefantini:
just talked with @blisteringherb IRL. the translated copy is being pulled in properly so we're all good

sheyd:
@blisteringherb Morgan, can we close this out?

blisteringherb:
Yes, I think so. I pushed those changes yesterday and they look good to me today.

Issue #5130, opened 2015-09-15
Display country-specific content at URLs with country prefix


blisteringherb:
#### Dosomething.org
- Friendly to Global users
- Available in Global English, Spanish, Portuguese translations 
- Surfaces when there isn’t a country specific URL
#### DoSomething.org/mx
- Country specific redirects with native content + global content

Basic Fastly flow here - https://drive.google.com/a/dosomething.org/file/d/0BwVjZK9CrAH_VGMyeXphU0V3YWc/view

In most cases Fastly will direct the user to the appropriate URL, but content that's visited directly with a country prefix should not be redirected.

mshmsh5000:
Every variation of this needs to be represented in the doc that @namimody is putting together.

mshmsh5000:
These tests (in red) are meant to relate to this issue. We'll need to agree on these as a group, how they get implemented, and in what layers (app vs. Fastly)

![redirect-tests](https://cloud.githubusercontent.com/assets/149811/9919949/33eaf0fa-5c9e-11e5-8b57-df6111e9c313.png)

angaither:
things we talked about 
- [ ] url alias per language
- [ ] translation redirect module

mikefantini:
URL alias per language: http://thor.dosomething.org/admin/config/search/path/patterns

seems like we'll have to add in https://www.drupal.org/project/pathauto_i18n to apply alias to each translation

cc: @blisteringherb

mshmsh5000:
Current bad behavior: #5163

angaither:
hm
I also just found this setting

`https://thor.dosomething.org/admin/config/regional/language/edit/en`
![screen shot 2015-09-18 at 3 56 08 pm](https://cloud.githubusercontent.com/assets/645205/9969660/d986bbae-5e1d-11e5-960b-35471ad25664.png)

mshmsh5000:
BLARGH -- good find @angaither -- I don't think we've configured this at all correctly yet:

![detection](https://cloud.githubusercontent.com/assets/149811/9970102/0a773632-5e21-11e5-8fb3-3c75c518522b.png)

mshmsh5000:
This is set up correctly, i.e., to obey path prefix and not domain:

https://thor.dosomething.org/admin/config/regional/language/configure/url

I changed [the detections settings](https://thor.dosomething.org/admin/config/regional/language/configure) to simplify:

[...]

PR #5153, opened 2015-09-17
Using user's account language to set campaign params

### LINKED ISSUE #5091: Global English campaign signup transactional email
When a user signs up for a campaign that is translated in Global English, s/he should receive a campaign signup transactional email in Global English. This is needed because content could potentially be different in |STEP_ONE| and |STEP_THREE|

Also note that we need to make sure we're pulling in the proper fields for this since technically the Global English content is in different fields than the US English fields.

related to #4956 

**Code of interest**:
https://github.com/DoSomething/phoenix/blob/ffb2ee968fc45dc5854ccb03c3dbf993eb38fe05/lib/modules/dosomething/dosomething_signup/dosomething_signup.module#L585-L601

**Called from**:
- **dosomething_signup_get_mbp_params**: https://github.com/DoSomething/phoenix/blob/ffb2ee968fc45dc5854ccb03c3dbf993eb38fe05/lib/modules/dosomething/dosomething_signup/dosomething_signup.module#L578

then:
- **dosomething_signup_mbp_request**: https://github.com/DoSomething/phoenix/blob/ffb2ee968fc45dc5854ccb03c3dbf993eb38fe05/lib/modules/dosomething/dosomething_signup/dosomething_signup.module#L534

then:
- **dosomething_signup_get_mbp_params**: https://github.com/DoSomething/phoenix/blob/ffb2ee968fc45dc5854ccb03c3dbf993eb38fe05/lib/modules/dosomething/dosomething_signup/dosomething_signup.module#L534-L537

blisteringherb:
- When user's sign up for a campaign, they will now receive their signup
  emails with copy in their language of choice.
- The user's language should always be set appropriately either when they create an account or when they change their own language settings.

Fixes #5091 and #4956

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -221,6 +221,7 @@ function dosomething_global_convert_country_to_language($country) {
     }
   }
   return 'en';
+}
 
 /*
  * Get the appropriate language for the application.
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -232,7 +232,7 @@ function dosomething_global_convert_country_to_language($country) {
  *
  */
 function dosomething_global_get_language($account, stdClass $node = NULL) {
-  if ($node = $options['node']) {
+  if ($node) {
     // If there is a published translation for the node in the user's language, use it.
     if ($node->translations->data['$account->language']['status']) {
       $language = $account->language;
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -221,4 +221,37 @@ function dosomething_global_convert_country_to_language($country) {
     }
   }
   return 'en';
+
+/*
+ * Get the appropriate language for the application.
+ *
+ * @param $account
+ *   The user's account
+ * @param  $node
+ *   Optional node.
+ *
+ */
+function dosomething_global_get_language($account, stdClass $node = NULL) {
+  if ($node = $options['node']) {
+    // If there is a published translation for the node in the user's language, use it.
+    if ($node->translations->data['$account->language']['status']) {
+      $language = $account->language;
+    }
+    // If there's no published translation in their language try 'en-global'.
+    elseif ($node->translations->data['en-global']['status']) {
+      $language = 'en-global';
+    }
+    // Default to node language if there's no translation in their language and no 'en-global'.
+    else {
+      $language = $node->language;
+    }
+  }
+  elseif ($language = $account->language) {
+    return $language;
+  }
+  else {
+    $language = dosomething_global_get_user_language($account);
+  }
+
+  return $language;
 }
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -12,7 +12,7 @@ include_once 'includes/Campaign.php';
 include_once 'includes/CampaignController.php';
 include_once 'includes/CampaignTransformer.php';
 include_once 'dosomething_campaign.theme.inc';
-define('DOSOMETHING_CAMPAIGN_PIC_STEP_HEADER', 'Snap a Pic');
+define('DOSOMETHING_CAMPAIGN_PIC_STEP_HEADER', t('Snap a Pic'));
 
 /**
  * Implements hook_form_alter().
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -592,3 +592,38 @@ function dosomething_helpers_get_current_language_code() {
 function dosomething_helpers_application_path($path = '') {
   return realpath(DRUPAL_ROOT . '/../' . $path);
 }
+
+/**
+ * Get the appropriate language for the application.
+ *
+ * @param $account
+ *   The user's account
+ * @param  $node
+ *   Optional node.
+ *
+ */
+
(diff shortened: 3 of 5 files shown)


angaither:
code looks :ok: 
but I don't think we are allowing users to change their own lang? what if this isn't set?

blisteringherb:
The language by default when a user creates an account will be 'en'. We'll need a hook_user_insert to make sure the language gets set properly based on the Fastly headers, so the user's language should always be set properly. If the hook_user_insert malfunctions for some reason then the user will get content in 'en'. We may want to log when the user's language setting is different than what we get from Fastly.

itsjoekent:
Yeah this was the point I raised yesterday @angaither 

https://dosomething.slack.com/archives/global/p1442416380000776

We decided it's safe to assume when a user registers they've always have the correct language set

angaither:
...right but what about users we have already w/o a lang?

itsjoekent:
what users dont have a language set? dont they have to have at least english set?

blisteringherb:
@angaither @deadlybutter Hm, it looks like in the db the language column is empty or und for all but about 6 users. Should we run a script now before we turn on global to set all of those users' language to 'en'?  Is there any advantage to keeping the language empty or 'und' for a user?

itsjoekent:
I don't see any advantage, I think we should put everyone on the same system. @angaither ?

angaither:
@blisteringherb rather than writing a script to handle this, I have sort of done it here
https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_global/dosomething_global.module#L116

blisteringherb:
@angaither That function accounts for users with global admin accounts but not others. We could expand that function or write some other helpers to return the language based on logic, but I think it would be more performant if we make sure the user's language is set properly in hook_user_insert and then generally use the user's language whenever we need to show them something.

angaither:
@blisteringherb that sounds good to me, all I meant was we should have a default of english if the user doesn't have a language set on insert (which I think is the case for most of our users now)

itsjoekent:
according to morgans comment above, only 6 users don't have a language set. can we set it for them & assume from thereon out all users will have a language set?

angaither:
@deadlybutter, he said all but 6 have a lang set. 
I just ran the query ~20 k have something set
the other 3 million have nothing or `und` set

itsjoekent:
OH. well that changes things. 

EDIT: so yeah, we should probably make a helper function which we call to grab a users language that can provide a default fallback. (should this be global-en, or en?)

mikefantini:
since the majority of our current users are :us: and the ratio of new users will most likely stay majority :us: for a while I would say default should be en.

mshmsh5000:
:+1:. Because of #5130, the system's default language may become Global English. Makes sense for users still to default to US English, though.

deezone:
@blisteringherb @mikefantini There might need to be clarification on this issue as my understanding is a campaign sign up should trigger a transactional email in the **language of the campaign** NOT in the language of the user (as the issue description is suggesting).

Looking at the code changes in this PR it looks like you are using the language of the campaign. Just wanted to clarify as the bulk of the comments on this issue are discussing the user's language preference at the time of signup.

mikefantini:
@blisteringherb @DeeZone the languages of the emails should be as follows:

| Type of Email | Language |
| --- | --- |
| User Registration | Language of user |
| PW Reset | Language of user |
| Campaign signup | Language of campaign |
| Campaign reportback | Language of campaign |

Let me know if any further clarification is needed.

blisteringherb:
@DeeZone @mikefantini I'm unsure what you mean by "language of campaign". The user's language preference or the language from the URL are the only languages that we can use. A campaign will have multiple languages.

mikefantini:
@blisteringherb take the following case:
- Brazilian user
- signs up for node 123
- node 123 has a US English translation published and a Global English translation published but has no BR-PT translation published
- in this case even though the user's language setting would be PT-BR, because she signed up for a Global English translation of the campaign, her transactional email should be in Global English, not BR-PT
- user then signs up for node 124
- node 124 has a US English translation published, a Global English translation published, and a BR-PT translation published
- in this case, the user would be seeing the BR-PT translation of the campaign and would then get a transactional email for node 124 in BR-PT

does that make sense/seem doable @blisteringherb?

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -221,4 +221,37 @@ function dosomething_global_convert_country_to_language($country) {
...
+ * Get the appropriate language for the application.
+ *
+ * @param $account
+ *   The user's account
+ * @param  $node
+ *   Optional node.
+ *
+ */
+function dosomething_global_get_language($account, stdClass $node = NULL) {
+  if ($node = $options['node']) {
```
angaither: `$options['node']` isn't getting passed in?

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -221,4 +221,37 @@ function dosomething_global_convert_country_to_language($country) {
...
+    elseif ($node->translations->data['en-global']['status']) {
+      $language = 'en-global';
+    }
+    // Default to node language if there's no translation in their language and no 'en-global'.
+    else {
+      $language = $node->language;
+    }
+  }
+  elseif ($language = $account->language) {
+    return $language;
```
angaither, line 251: shouldn't this say so you aren't returning in the middle?

``` php
elseif (isset($account->language)) {
 $language = $account->language
}
```

blisteringherb, line 251: Is there a reason to avoid returning here?

mshmsh5000, line 251: This is an assignment, not a comparison. `if ($language = $account->language)` will always be `true`.

If the intended logic is "if the current language is equal to the account language, then just return the language here and skip the last few lines," that makes sense to me.

angaither, line 251: :+1: ya agree, I think I just misunderstood that

PR #5164, opened 2015-09-21
Content not saving on new nodes

### LINKED ISSUE #5126: nothing but title saves first time i create a campaign as regional admin
http://staging.beta.dosomething.org/node/1701/edit

image

cc: @angaither

angaither:
Changing the node language in `hook_node_presave()` was too late and only the title was saving.
Since this only needs to run on newly created nodes, we can change the language on `hook_node_insert()`

Fixes #5126


blisteringherb:
:+1:

PR #5175, opened 2015-09-21
Update to include unpublished mobile app campaigns in index collection.

weerd:
Refs #5174 
#### What's this PR do?

This PR updates the query used to collect an index listing of campaigns for the `/campaigns` endpoint to temporarily allow the retrieval of unpublished campaigns that are also marked as being on the `mobile_app`.

Just confirming that the code looks correct and consider if there's any potential conflicts.

Prior PR #4986 only allowed retrieval of unpublished single campaigns, however the mobile team also needs to be able to access unpublished collections of campaigns.
#5174

---

@angaither @jonuy @chloealee


angaither:
:+1: 
i don't even think this needs a `@todo` since that's the solution we decided to go with :wink:

weerd:
LOL. I'm just worried about it. :concerned_parent:

PR #5177, opened 2015-09-21
Mobile app dates

### LINKED ISSUE #5041: Campaign API query parameter to return mobile campaigns valid for a given date
Campaigns can be set with an `mobile_app` flag along with mobile app start and end dates. Currently through the API we're only able to query against `mobile_app=true` which gives all campaigns that have the flag set regardless of dates they should be available.

We'd like to also query campaigns that are available for a given date.

For example, if I have a campaign with a mobile app start of 3/1/2015 and end of 4/1/2015, this campaign should be returned if I query with a date like 3/20/2015. But if I query with 4/20/2015, then it shouldn't.

I have no strong preferences for parameter name. Something like... `available_in_mobile_app_date`..? :grimacing:

chloealee:
#### What's this PR do?

Functionality to query campaigns that are available for a given date. 

`campaign_resource.inc` lines 41 - 48, then line 289
`CampaignTransformer.php` line 25
`dosomething_campaign.module` lines 457-462

example query: `/campaigns.json?mobile_app_date=2015-09-02` should return all campaigns with query date in between mobile app start and end dates

Fixes #5041


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -453,6 +454,13 @@ function dosomething_campaign_build_campaigns_query($params = array()) {
     $query->condition('m.field_on_mobile_app_value', 1);
   }
 
+  if (isset($params['mobile_app_date'])) {
+    $query->join('field_data_field_mobile_app_date', 'm', 'm.entity_id = n.nid');
+    $query
+      ->condition('m.field_mobile_app_date_value', dosomething_helpers_convert_date($params['mobile_app_date']), '<=')
+      ->condition('m.field_mobile_app_date_value2', dosomething_helpers_convert_date($params['mobile_app_date']), '>=');
```
weerd, line 461: I might just be confused, but wouldn't this just grab everything? Less than or equal to the specified date, and greater than or equal to the specified date?

chloealee, line 461: @weerd this is treated as less than or equal to && great than or equal to -- both must be true! So this checks that the date is in between the start and end dates.

weerd, line 461: Ah gotcha. Cool cool :)

weerd:
:+1:

PR #5185, opened 2015-09-22
Path-driven MX, BR, US, global English translation behavior

mshmsh5000:
#### What's this PR do?

Updates the `dosomething_global` feature with configuration-based behaviors to provide MX, BR, Global English, or US English translations of content based on the path structure.

Using the language preferences settings, this feature update maps these URL prefixes:
- **/mx/** to Mexican Spanish (es-mx)
- **/br/** to Brazilian Portugues (pt-br)
- **/us/** to U.S. English (en-us)
- **[no prefix]** to Global English

This PR also **sets Global English as the default language**.

You can update this feature locally, or test on Thor, where the same configuration is in place (not yet managed by this Feature update).

The changes to the `strongarm.inc` and `language.inc` files are most of the URL behavior config. The changes to `dosomething_global.module` are commenting and readability, and a bit of cleanup to make my IDE happy.

Basic test:
1. Edit a campaign.
2. Using the **Translations** tab, add the translations for es-mx, pt-br, Global English.
3. Edit each translation, changing enough obvious content (e.g., the title) that you can easily see the differences among translations.
4. After saving, review the resulting URL and verify the translated content.
5. At the end, you should have four unique URLs loading four translated versions of the same campaign.

If this works, we need to implement the same changes for other URL-accessible content we want to translate.

**Gotcha:** If you get a redirect loop on the Global English version of a campaign, go to the Edit view. You'll see a redundant redirect in the Redirects tab at the bottom of the edit form. Delete that.

If we see this as regular behavior, then of course we'll need to address as a bug. I saw it once.

**Deployment:** This should not change any existing paths in Drupal. We do want to make _super sure_ there are no side effects in Thor from changing the default language -- for instance, when creating new content as admins in different regions.
#5130

See all of the screenshots and notes in these comments: [comment 1](https://github.com/DoSomething/phoenix/issues/5130#issuecomment-142116905) and [comment 2](https://github.com/DoSomething/phoenix/issues/5130#issuecomment-142306499)
#### Questions:
- Is there a blog post? _No._
- Does the knowledge base need an update? _Yes -- will update the Global section of the Phoenix wiki once this functionality is stable.
- Does this add new Drupal dependencies? _No._


lib/modules/dosomething/dosomething_global/dosomething_global.features.language.inc
```diff
@@ -33,7 +33,7 @@ function dosomething_global_locale_default_languages() {
     'plurals' => 0,
     'formula' => '',
     'domain' => '',
-    'prefix' => 'es-mx',
+    'prefix' => 'mx',
```
angaither, line 36: if we are changing the prefixes here, should we also change them in the language map strongarm variable?

mshmsh5000, line 36: So we're talking about [this in Strongarm](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_global/dosomething_global.strongarm.inc#L17). My understanding is that this mapping is based on the language's "Language code", not the "Path prefix". The language code isn't editable (a good thing). See here:

lang-code-prefix

PR #5188, opened 2015-09-22
Translate home page content

### LINKED ISSUE #4968: need BR, MX, and Global versions of homepage
we decided we're going to be translating the home content type just like we did with campaign content types.

---

Would include the 9 campaigns we want featured to BR, MX, and global users + majority of copy needs to be in appropriate language.

I think it makes most sense to give Brazil Admin, Mexico Admin, and Global Admin access to create/edit a home node and then just serve up the appropriate homepage based on where the user is coming from.

The other option would be to make all of the pages on the home node translatable, including which of the 9 campaigns we're featuring on the homepage.

thoughts @blisteringherb @sheydari-ds @angaither

angaither:
#### what does this pr do?
- adds permission for regional admins to edit home page nodes
- allows `campaigns` field reference field to be translated
- allows `subtitle` field to be translated
- run node_alter hook to hide non-translatable fields only on translatable content types
#### background context

I am not granting regional admins the ability to create new home page nodes, only translate the current one. 
#### relevant tickets

fixes #4968

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.features.field_base.inc
+++ b/lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.features.field_base.inc
@@ -33,6 +33,7 @@ function dosomething_campaign_group_field_default_field_bases() {
     'locked' => 0,
     'module' => 'entityreference',
     'settings' => array(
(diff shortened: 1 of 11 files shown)


blisteringherb:
:+1:

PR #5190, opened 2015-09-22
Global Date Validation

### LINKED ISSUE #5084: need dd/mm/yyyy birthday field for non-US users
non-us users (so BR, MX, and global users) should get presented with a different birthday field that accepts dates as mm/dd/yyyy. All birthday fields regardless of how they are input should obviously be stored the same way in the same column in the db.

@mnelson338 does anyone else use the mm/dd/yyyy format besides :us:?

related to #5046

sbsmith86:
#### What's this PR do?

Updates the birthday field in the registration modal to ask for and validate against the appropriate date format based on the users country code. BR, MX, and Global users should get `DD/MM/YYYY`. US users should get `MM/DD/YYYY` 

`dosomething_user.module` passes the the date format we want to use for validation to the JS.
`form.inc` just updates the placeholder text appropriately. 

Vist the site using different country code headers to verify you are being asked for your birthday in the correct format. 

Also, try bad dates (bad month, years in the future, etc) and make sure the field is being validated correctly. 

Similar functionality to how we do this in the `dosomething_uk` module. 

Fixes #5084

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -19,11 +19,14 @@ function dosomething_user_preprocess_page(&$vars) {
     dosomething_helpers_add_analytics_event("login", $_SESSION['dosomething_user_log_login']);
     unset($_SESSION['dosomething_user_log_login']);
   }
+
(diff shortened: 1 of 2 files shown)


angaither:
:+1:
not needed for merge, but I imagine we are going to need this in a bunch of places, maybe we should eventually functionify date format?

sbsmith86:
@angaither I had that thought as well. I can work it into another issue probably.

PR #5208, opened 2015-09-23
Hide the translate tab for people who don't have access to translate …

### LINKED ISSUE #5191: only administrators and global admins should have access to Translate Tab
image

angaither:
…any entity.

Only staff members with the role [editor, admin, creative team, global admin] will have access to the translate tab

Fixes #5191


blisteringherb:
:+1:

Issue #5205, opened 2015-09-23
add other affliates to lang map


angaither:
for auth flow needs to hide/show certain fields based on country codes. 

UK and CA should be added to the lang map variable
should they be global english as language? 

refs #5052 #5085
cc: @blisteringherb @mshmsh5000

mshmsh5000:
I'm not clear on why this is a Global 1.0 task, since UK and CA are out of scope and still have their own sites.
- en-gb
- en-ca
- fr-ca

Especially given Canada's two languages, I don't think we want to touch this right now.

angaither:
@mshmsh5000 If we want to hide/show specific fields in the auth flow, I think it will be easier to have this stuff in the strongarm so @sbsmith86 can call one function to check countries, [instead of creating an array of country codes](https://github.com/sbsmith86/dosomething/commit/75aed93bde496350645ec49a01e38afde482c227#diff-b4b14fec58a66c7539097ce097603450R82)
I don't think we need to define roles, or permissions or the language, just the country if we want the auth flow to show different things.

mshmsh5000:
Got it, got it. :+1:

PR #5210, opened 2015-09-23
Hide/Show cell phone field based on country code

### LINKED ISSUE #5085: hide cell phone field on auth flow for global users
- If US, MX, BR, Ca, UK then all auth fields show
- If global, then suppress SMS 
- If we anticipate building an SMS program in a specific country, then will will show SMS field 6 mos. ahead of time

related to #5046

sbsmith86:
#### What's this PR do?

This PR hides the cell phone field if the users country code is not in our country language map, otherwise for US, MX, BR, CA and UK we show and collect that info.

The magic happens in `register.inc`

Change your country code or use VPN to pretend you are in another country and when you try to create an account you should only see the cell phone field if you are in US, MA, BR, CA, or UK. 

One thing, I sorta lied. Right now, CA and UK are not in our language map so the cell phone field will be hidden for those counties. However, #5205 will add those countries to keep me honest. 

Fixes #5085

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/auth/register.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/auth/register.inc
@@ -82,7 +82,8 @@ function paraneue_dosomething_form_alter_register(&$form, &$form_state, $form_id
 
     if (!in_array(dosomething_settings_get_geo_country_code(), dosomething_global_get_countries())) {
       $form['field_mobile']['#access'] = FALSE;
-    } else {
+    }
+    else {
       $form['account']['field_mobile'] = $form['field_mobile'];
       $form['account']['field_mobile']['#weight'] = 20;
       $form['account']['field_mobile'][LANGUAGE_NONE][0]['value']['#title'] = t('Cell Number') . ' <span class="field-label-optional">' . t("(optional)") . '</span>';
--- a/lib/themes/dosomething/paraneue_dosomething/includes/auth/register.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/auth/register.inc
@@ -79,14 +79,20 @@ function paraneue_dosomething_form_alter_register(&$form, &$form_state, $form_id
     $form['account']['mail']['#attributes']['data-validate-required'] = '';
     unset($form['account']['mail']['#description']);
 
-    $form['account']['field_mobile'] = $form['field_mobile'];
-    $form['field_mobile']['#access'] = FALSE;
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/auth/register.inc
```diff
@@ -79,14 +79,20 @@ function paraneue_dosomething_form_alter_register(&$form, &$form_state, $form_id
...
-
-    $form['account']['field_mobile']['#weight'] = 20;
-    $form['account']['field_mobile'][LANGUAGE_NONE][0]['value']['#title'] = t('Cell Number') . ' <span class="field-label-optional">' . t("(optional)") . '</span>';
-    $form['account']['field_mobile'][LANGUAGE_NONE][0]['value']['#attributes']['placeholder'] = t('(555) 555-5555');
-    $form['account']['field_mobile'][LANGUAGE_NONE][0]['value']['#attributes']['class'] = array('js-validate');
-    $form['account']['field_mobile'][LANGUAGE_NONE][0]['value']['#attributes']['data-validate'] = 'phone';
+
+    if (!in_array(dosomething_settings_get_geo_country_code(), dosomething_global_get_countries())) {
+      $form['field_mobile']['#access'] = FALSE;
+    } else {
```
angaither: else on a new line!

lib/themes/dosomething/paraneue_dosomething/includes/auth/register.inc
```diff
@@ -79,14 +79,20 @@ function paraneue_dosomething_form_alter_register(&$form, &$form_state, $form_id
...
+      $form['account']['field_mobile'] = $form['field_mobile'];
+      $form['field_mobile']['#access'] = FALSE;
```
angaither: you have `$form['field_mobile']['#access'] = FALSE` in both the if and the else, is that real?

sbsmith86: @angaither haha it is real, I copied that over from how the form altering was working before. It seems to hide the original field? I'm not sure. Let me see if removing it breaks anything.

lib/themes/dosomething/paraneue_dosomething/includes/auth/register.inc
sbsmith86: @angaither whoops, i'm in JS land.

sbsmith86:
@angaither updated all the things.

angaither:
:+1:

PR #5211, opened 2015-09-23
Updating to remove admin Views related strings.

weerd:
Updates #5200
#### What's this PR do?

This PR updates the extracted POT file for the dosomething_user module to remove strings that only apply to the Views admin interface.

Double check the removed code.
#5200

---

@angaither


angaither:
:+1:

PR #5214, opened 2015-09-23
POT files and the lovely t function

angaither:
@weerd has been merging into global branch still. whoops! 

This is all of his most recent changes
- wrapping `t()` around custom dosomething settings
- adding POT files into version control


blisteringherb:
:+1:

PR #5220, opened 2015-09-24
Exporting language prefix settings and url alias settings

### LINKED ISSUE #5130: Display country-specific content at URLs with country prefix

blisteringherb:
- Includes Pathauto i18n settings for campaigns in en, en-global, es-mx,
  and pt-br.
- Includes updates to the path prefix for es-mx (mx) and pt-br (br)

Fixes #5130


angaither:
:+1:

PR #5223, opened 2015-09-24
Merge latest dev changes into Global

blisteringherb:
There are upcoming changes to the URL structure that have been and will be made against the Global branch and deployed to Thor, so this is making sure global is up to date with the latest dev changes so everything can be tested together.


angaither:
:+1: :facepalm:

weerd:
Oh crap. Ok, should I be merging into Global then? Cause I was and then we had decided not to anymore, so I went back to merging into Dev...

Issue #5193, opened 2015-09-22
need a global solution for address collection


mikefantini:
related to #5052 

need to:
1. swap out US-centric fields for Global ones when users are non-US (i.e. no state, zip)
2. figure out a solution for address validation

mshmsh5000:
_Re: Validation:_ I'm into the validation approach in the [Phone module](https://www.drupal.org/project/phone) -- patterns broken out by countries.

_Re: Address collection:_ The more recent [Address module](https://www.drupal.org/project/address) seeks to address this.

PR #5227, opened 2015-09-24
Hide address field for non-us users

### LINKED ISSUE #5052: hide address field for non-US users
currently shows up on the profile and any campaigns where we collect address.

right now we have US-centric fields (state, zip code). obviously we've resolved this issue on individual international sites, but need a solution for the one site to rule them all **TM**
#### solution

one fine solution is punting for now on this by hiding these fields on the user profile for non-US users and not enabling address collection on campaign nodes

sbsmith86:
#### What's this PR do?

Hides address fields for non-us users on the update profile form.

Only one place to go `form.inc`

Log in as an authenticated non-admin user and edit your profile. You should see address fields if your country code is `US`, you should **not** see address fields if it is not. 

This is a temp solution for the global 1.0 release. In global 1.1 we will be deciding on a global solution for address collection (see #5193 )

Fixes #5052


angaither:
:+1:

PR #5231, opened 2015-09-24
#5130: BR, MX path patterns; default lang

mshmsh5000:
#### What's this PR do?

Updates the `dosomething_global` feature with path patterns for campaigns.

This PR also **sets Global English as the default language**.

Strongarm var changes.

Basic test:
1. Edit a campaign.
2. Using the **Translations** tab, add the translations for es-mx, pt-br, Global English.
3. Edit each translation, changing enough obvious content (e.g., the title) that you can easily see the differences among translations.
4. After saving, review the resulting URL and verify the translated content.
5. At the end, you should have four unique URLs loading four translated versions of the same campaign.

If this works, we need to implement the same changes for other URL-accessible content we want to translate.

**Gotcha:** If you get a redirect loop on the Global English version of a campaign, go to the Edit view. You'll see a redundant redirect in the Redirects tab at the bottom of the edit form. Delete that.

If we see this as regular behavior, then of course we'll need to address as a bug. I saw it once.

**Deployment:** This should not change any existing paths in Drupal. We do want to make _super sure_ there are no side effects in Thor from changing the default language -- for instance, when creating new content as admins in different regions.
#5130

See all of the screenshots and notes in these comments: [comment 1](https://github.com/DoSomething/phoenix/issues/5130#issuecomment-142116905) and [comment 2](https://github.com/DoSomething/phoenix/issues/5130#issuecomment-142306499)
#### Questions:
- Is there a blog post? _No._
- Does the knowledge base need an update? _Yes_ -- will update the Global section of the Phoenix wiki once this functionality is stable.
- Does this add new Drupal dependencies? _No._


mshmsh5000:
This also removes session/request variables as a language detection mechanism. @angaither @blisteringherb it seems simpler to have the smallest possible number of mechanisms.

blisteringherb:
:+1:

PR #5247, opened 2015-09-25
Use the node's default lang to get cover image.

### LINKED ISSUE #5246: images missing from homepage on thor
as a US user, or an anon user, images are missing from the homepage on thor. 

screen shot 2015-09-25 at 9 15 30 am

``` php
 $image_nid = $node->field_image_campaign_cover[$lang_code][0]['target_id'] ?: $node->field_image_campaign_cover[language_default()->language][0]['target_id'];
```

on thor for an anon user that evaluates to 
get image nid in global english, or image id in global english.

it works on stage, because the default language is still english. 

refs #5185

angaither:
### what's this pr do?

Updates call to get node cover images
### context

when we changed the default language on thor, the images went POOF
the code was trying to get the images in global english, but the nodes were in english. 
### other things

I _think_ the code should be like this, we try to get the image in the user's language first, then get the image in the node's original language.

Fixes #5246


blisteringherb:
:+1:

PR #5248, opened 2015-09-25
Reportback emails 4957

### LINKED ISSUE #4957: Global Report Back Transactional Email
When a user reports back on a campaign, s/he should receive a report back transactional email in the same language as the translation of the campaign s/he reported back on. i.e. if I report back on the Brazilian Portuguese translation of node 123, I should receive a Brazilian Portuguese report back transactional email
- [ ] Brazilian Portuguese report back transactional email
- [ ] Mexican Spanish report back transactional email

Also note that we need to make sure we're pulling in the proper fields for this since technically the Brazilian Portuguese and Mexican Spanish content are in different fiends than the US English fields.

And same as my comment on #4955 we need to make /campaigns will take the user to the correct version of that URL depending on their location, but I think this should be fine.

@sergii-tkachenko @sheydari-ds if you have any thoughts to advise @DeeZone on this please include here.

blisteringherb:
Updating reportback queries to be language-aware so that reportback emails display the appropriate language to users.

Fixes #4957

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -1026,6 +1026,8 @@ function dosomething_reportback_flush_caches() {
  * @return SelectQuery object
  */
 function dosomething_reportback_build_reportbacks_query($params = array()) {
+  global $language;
+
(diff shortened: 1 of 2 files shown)


angaither:
:+1: what issue is this for?

PR #5257, opened 2015-09-25
Use node's lang for CTA

### LINKED ISSUE #5249: CTAs not appearing on the homepage for anonymous users
related to #5246 and #5217 as an anonymous user visiting Thor from the :us: I'm not seeing some CTAs on the homepage

I don't think that the solution we went with in #5246 should perfectly apply here because CTA is a required field, so there _shouldn't be_ any option for a translation to be published without having a CTA. And if we did go the route of #5246 we'd then be pulling in the CTA in the original language (most likely US-English) which wouldn't make sense for a Mexican user to see.

If the campaign doesn't exist at all in a translation that is appropriate for the user (i.e. doesn't exist in Mexican Spanish or Global English) then it shouldn't even be appearing in the finder in the first place.

But maybe I'm missing something.

image

cc: @angaither @blisteringherb

angaither:
Fixes #5249

### DIFF
--- a/lib/modules/dosomething/dosomething_home/dosomething_home.module
+++ b/lib/modules/dosomething/dosomething_home/dosomething_home.module
@@ -36,15 +36,15 @@ function dosomething_home_preprocess_node(&$vars) {
       }
       $node = node_load($nid);
 
-      // Grab all tiles attributes.
       global $language;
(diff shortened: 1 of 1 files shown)


blisteringherb:
:+1:

PR #5259, opened 2015-09-25
Verifies lang exists before assigning it

### LINKED ISSUE #5256: Check if we actually have the language before giving it to a user
Users from GB or CA are currently being assigned languages we don't support when they register (Thor), drupal gives them English instead of Global english. 

We should have checks to prevent this from happening and give them global english instead.

itsjoekent:
#### What's this PR do?

Verifies the system language is actually installed before giving it to a user. Default to global en if it doesnt exist.

Create a new user and spoof the country code as GB or CA.

Fixes #5256


angaither:
:+1:

Issue #5263, opened 2015-09-25
Thor: Registration doesn't pass country code to Northstar


mshmsh5000:
For a Sena's test user (See @sheyd for details), which was a BR user, he received the PT-BR confirmation transactional. But the user data recorded in Northstar has no country at all.

May be incomplete data passed to NS, or NS not recording data we passed. Calling it a Phoenix issue until we determine otherwise.

angaither:
NS attempts to [get the country from the address field](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module#L164)
we've disabled that field for non-us users. we need to save that in the `dosomething_user.module`

we currently have a [conditional that fires on a new user to set the user's language.](https://github.com/DoSomething/phoenix/blob/a6944b7444172f0e5119e8a9754e87a0a96406cb/lib/modules/dosomething/dosomething_user/dosomething_user.module#L467-L476)
this country save should be done in a similar way, and perhaps these two should be grouped together in their own function. `dosomething_user_set_new_user_global_attributes`

itsjoekent:
@chloealee wanna hold off on this one real quick while I fix-up #5168 and make that function?

itsjoekent:
@chloealee this is unblocked :D see the PR above for the function where we gotta add northstar support

chloealee:
great, thanks @deadlybutter ! Super helpful to get started on this :)

angaither:
reopening to deal with sending the entire address, along with the country code

PR #5266, opened 2015-09-28
Updates global 

### LINKED ISSUE #5168: Set user language based on fastly header when registering
When a user signs up we should read the Fastly headers and set there language appropriately. Functions in dosomething_settings can be found to read the data embedded in the headers.

cc: @DeeZone @mikefantini 

**Related**:
- https://github.com/DoSomething/phoenix/issues/4955 - User Registration transactional email
- https://github.com/DoSomething/phoenix/issues/4958 - Password reset transactional email

itsjoekent:
#### What's this PR do?

Updates the logic used to seta users language upon registering. See here for summary of what was discussed in #global https://github.com/DoSomething/phoenix/issues/5168#issuecomment-143766220

All of this was also moved into a new function per @angaither 's suggestion here https://github.com/DoSomething/phoenix/issues/5263#issuecomment-143743816

Also updates dosomething_global_convert_country_to_language to return null rather than 'en'

dosomething_user_set_global_attributes();

Create new users and spoof there country codes, check there languages to make sure they match. Also try making a user without a country code header and make sure it goes to English. 

Slack discussion from here and down https://dosomething.slack.com/archives/global/p1443451428000871

Fixes #5168 
Unblocks #5263

Also @DeeZone this is going to break your logic here for users coming from countries outside US/MX/BR https://github.com/DoSomething/phoenix/blob/cf11f1b791a9668ae73f403b8121dc055156b447/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module#L235

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -205,7 +205,7 @@ function dosomething_global_convert_language_to_country($language) {
 }
 
 /**
- * Converts the given country into its assigned language
+ * Converts the given country into its assigned language.
  *
  * @param string country
  *  The country Code
@@ -220,6 +220,7 @@ function dosomething_global_convert_country_to_language($country) {
       return $lang_key;
     }
   }
+  // It is not the job of this function to determine a default language
   return NULL;
 }
 
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -211,7 +211,7 @@ function dosomething_global_convert_language_to_country($language) {
  *  The country Code
  *
  * @return
- *  The language mapped to this country
+ *  The language mapped to this country or NULL
  */
 function dosomething_global_convert_country_to_language($country) {
   $lang_map = variable_get('dosomething_global_language_map', '');
@@ -220,7 +220,7 @@ function dosomething_global_convert_country_to_language($country) {
       return $lang_key;
     }
   }
-  return 'en';
+  return NULL;
 }
 
 /*
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -465,19 +465,7 @@ function dosomething_user_new_user($form, &$form_state) {
   $account = $user;
 
   if (module_exists('dosomething_global')) {
-    // Adjust language based on location
-    $user_country_code = dosomething_settings_get_geo_country_code();
-    $coverted_language = dosomething_global_convert_country_to_language($user_country_code);
-    $system_languages = language_list();
-    $language = 'en-global';
-    
-    foreach ($system_languages as $lang_key => $system_lang) {
-      if ($coverted_language == $lang_key) {
-        $language = $coverted_language;
-        break;
-      }
-    }
-    user_save($account, ['language' => $language]);
+    dosomething_user_set_global_attributes();
   }
 
   // Should we sign this kid up for messages?
@@ -526,6 +514,44 @@ function dosomething_user_new_user($form, &$form_state) {
    }
 
   dosomething_mbp_request('user_register', $params);
+}
+
+/**
+ * Sets the user global attributes and sends them to Northstar.
+ *
+ * @param object $account
+ *   The account to return value for. If NULL, uses global $user.
+ * @param string $country_code
+ *   The country code where the user currently resides. If NULL, uses
+ *   the headers in current request.
+ * @param string $language
+ *   The new language to set on the user. If NULL, gets the matching
+ *   language or defaults to global-en. 
+ */
+function dosomething_user_set_global_attributes($user = NULL, $country_code = NULL, $language = NULL) {
+  if ($user == NULL) {
+    global $user;
+  }
+  if ($country_code == NULL) {
+    $country_code = dosomething_settings_get_geo_country_code();
+  }
+  if ($language == NULL) {
+    $coverted_language = dosomething_global_convert_country_to_language($country_code);
+    $language = 'en-global';
+    $system_languages = language_list();
+
+    // Verify the language is installed on the system,
+    // if not default to global en
+    foreach ($system_languages as $lang_key => $system_lang) {
+      if ($coverted_language == $lang_key) {
+        $language = $coverted_language;
+        break;
+      }
+    }
+  }
+
+  user_save($account, ['language' => $language]);
+  //TODO: (#5263) Send $language and $country_code to NorthStar
 }
 
 /**
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -250,6 +250,9 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
       if ($form_id == 'user_register_form' && $_SERVER['REQUEST_URI'] != '/admin/people/create') {
         $form['#action'] = '/user/register';
         $form['#submit'][] = 'dosomething_user_new_user';
+        if (module_exists('dosomething_global')) {
+          $form['#submit'][] = 'dosomething_user_new_user_attributes';
+        }
         // Unsets relevant register form fields based on configuration variables.
         _dosomething_user_register_display_fields($form);
         // Add campaign data, if needed.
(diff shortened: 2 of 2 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -524,6 +517,51 @@ function dosomething_user_new_user($form, &$form_state) {
...
+ */
+function dosomething_user_set_global_attributes($user = NULL, $country_code = NULL, $language = NULL) {
+  if ($user == NULL) {
+    global $user;
+  }
+  if ($country_code == NULL) {
+    $country_code = dosomething_settings_get_geo_country_code();
+    // If no Fastly headers set, use US EN
+    if ($country_code == NULL) {
+      $language = 'en';
```
deezone, line 543: `$language = 'en';` and `$language = 'en-global';` on ln 544. Are they two different languages?

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -220,7 +220,7 @@ function dosomething_global_convert_country_to_language($country) {
       return $lang_key;
     }
   }
-  return 'en';
+  return NULL;
```
angaither, line 224: can you add comments for why we are returning null here?

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -465,14 +465,7 @@ function dosomething_user_new_user($form, &$form_state) {
...
   if (module_exists('dosomething_global')) {
-    // Adjust language based on location
-    $user_country_code = dosomething_settings_get_geo_country_code();
-    $language = dosomething_global_convert_country_to_language($user_country_code);
-    $system_languages = language_list();
-    if (!in_array($langauge, $system_languages)) {
-      $language = 'en-global';
-    }
-    user_save($account, ['language' => $language]);
+    dosomething_user_set_global_attributes();
```
angaither: instead of calling this from here, let's call it in the form_alter, like `dosomething_user_new_user()` is called

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -524,6 +517,51 @@ function dosomething_user_new_user($form, &$form_state) {
...
+        if ($converted_language == $lang_key) {
+          $language = $converted_language;
+          break;
+        }
+      }
+    }
+  }
+
+  user_save($user, ['language' => $language]);
+  //TODO: (#5263) Send $language and $country_code to NorthStar
```
angaither, line 565: we will also want to save user country to the user

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -524,6 +517,51 @@ function dosomething_user_new_user($form, &$form_state) {
...
+    global $user;
+  }
+  if ($country_code == NULL) {
+    $country_code = dosomething_settings_get_geo_country_code();
+    // If no Fastly headers set, use US EN
+    if ($country_code == NULL) {
+      $language = 'en';
+    }
+  }
+  if ($language == NULL) {
```
angaither, line 547: can you add a comment block above explaining the logic that happens below?

lib/modules/dosomething/dosomething_user/dosomething_user.module
itsjoekent, line 543: Yes they are two different languages. en is just for US, en global is for all countries outside of the US we don't explicitly support

lib/modules/dosomething/dosomething_global/dosomething_global.module
itsjoekent, line 224: yep- just to add it here first, 

this function shouldn't be deciding what the correct default value should be, as the name implies its only purpose is to find the matching language for that country. In some cases, that country doesn't have a matching language and it's useful to have it return a default lang, in others, it might be more preferable to have it return NULL.

lib/modules/dosomething/dosomething_user/dosomething_user.module
itsjoekent: sure thing

lib/modules/dosomething/dosomething_user/dosomething_user.module
itsjoekent, line 565: I don't see this field anywhere, has it been created yet?

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -524,6 +517,52 @@ function dosomething_user_new_user($form, &$form_state) {
...
+      foreach ($system_languages as $lang_key => $system_lang) {
+        if ($converted_language == $lang_key) {
+          $language = $converted_language;
+          break;
+        }
+      }
+    }
+  }
+
+  user_save($user, ['language' => $language]);
```
deezone: Is the `$country_code` value suppose to be included in the `user_save($user, ['language' => $language]);` ? 

Perhaps looking up the user country should always be done based on the user language setting? This seems problematic long run if we have more than one country with the same language but need know the users actual country.

itsjoekent: country code doesnt exist yet but yes it will be added.

and ideally the country code is always specified by the fastly headers and/or this variable we're creating. the language conversion hacks have already been made because we didn't have a country field before.

lib/modules/dosomething/dosomething_user/dosomething_user.module
angaither, line 565: it's a part of field_address

itsjoekent, line 565: ohh gotcha

lib/modules/dosomething/dosomething_user/dosomething_user.module
itsjoekent: ACTUALLY the country code exists already and was unintentionally fixed when I updated the get affiliate code logic 

https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_user/dosomething_user.module#L1167

lib/modules/dosomething/dosomething_user/dosomething_user.module
itsjoekent, line 565: see https://github.com/DoSomething/phoenix/pull/5266#discussion_r40583279

itsjoekent:
@angaither I think its ready for re-review. I commented above re: countries, the logic is already in place

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -250,6 +250,9 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
       if ($form_id == 'user_register_form' && $_SERVER['REQUEST_URI'] != '/admin/people/create') {
         $form['#action'] = '/user/register';
         $form['#submit'][] = 'dosomething_user_new_user';
+        if (module_exists('dosomething_global')) {
+          $form['#submit'][] = 'dosomething_user_new_user_attributes';
```
angaither, line 254: new_user_global_attributes?

angaither:
@deadlybutter what do you mean country code was unintentionally fixed? @mshmsh5000 and @sheyd  said country code isn't getting passed correctly on friday. looks like that [logic was added a while ago](https://github.com/DoSomething/phoenix/blame/dev/lib/modules/dosomething/dosomething_user/dosomething_user.module#L1167)

itsjoekent:
I just created two new users with diff country codes in the headers, printed them out and the country field in address was correct. I can do it again and screenshot it? and the function being called was updated, not that line, thats why it was unintentionally fixed 

https://github.com/DoSomething/phoenix/blob/458433df8cf46467efd178763f222bf51d7bd873/lib/modules/dosomething/dosomething_settings/dosomething_settings.module#L73

lib/modules/dosomething/dosomething_user/dosomething_user.module
itsjoekent, line 254: new function https://github.com/DoSomething/phoenix/pull/5266/files#diff-ad30a5d569ca2ae6d2e04223aa501104R518

angaither:
@deadlybutter address was set correctly in the user profile?

itsjoekent:
user profile doesnt display country 
screen shot 2015-09-28 at 2 14 24 pm

angaither:
so I _think_ the issue #5263 for it passing into NS still stands, but we are saving it correctly to the user profile, correct @deadlybutter?

itsjoekent:
yes and yes. just verified country is in the DB. just needs to be passed into northstar but thats #5263 

id rather get this fixed rather than wait for northstar integration as its blocking our testing later

angaither:
yip :+1:

PR #5269, opened 2015-09-28
Fixes #4973: title translation overwrites original title.

### LINKED ISSUE #4973: Adding a translated title overwrites original language title
When adding a new translation and title or when modifying the title of an existing translation, the title of the original translation is overwritten with the translation when the node/translation is published. The original language title can be fixed, but it will be overwritten again any time a new translation and title are added or when an existing translation title is modified.

This seems to only affect the title field as far as I can tell.

sergiitk:
#### What's this PR do?
- Fixes a bug with original title being overwritten with new translation
- Add a translation for a campaign, ex.: http://dev.dosomething.org:8888/node/1727/edit
- Change the title
- Make sure the title isn't changed on original (English) translation: http://dev.dosomething.org:8888/node/1727/edit/en?language=en
- Make sure the title is changed on the translation of your choice

See https://www.drupal.org/node/1991988.

Fixes #4973

cc @mikefantini @deadlybutter

### DIFF
--- a/lib/profiles/dosomething/drupal-org.make
+++ b/lib/profiles/dosomething/drupal-org.make
@@ -202,6 +202,9 @@ projects[taxonomy_access_fix][subdir] = "contrib"
 ; Fixes a bug with table layout on content type manage fields page.
 projects[title][version] = "1.x-dev"
 projects[title][subdir] = "contrib"
+; Fixes #4973: title translation overwrites original title.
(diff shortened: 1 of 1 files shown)


blisteringherb:
:+1:

itsjoekent:
@sergii-tkachenko good catch, thanks so much for figuring this one out

Issue #5046, opened 2015-09-08
Registration form client-side validation


DFurnes:
As we enter the wider global world, we should make sure that our validation isn't too US-centric! For example, we currently offer suggestions for common misspellings on email addresses. Do we need to offer additional suggestions for common mail providers in other countries?

(Making this issue as follow-up to an offline conversation with @mikefantini.)

DFurnes:
International phone numbers currently validate correctly on the client-side, and we can use translation to offer region-specific placeholders:

![screen shot 2015-09-08 at 4 57 41 pm](https://cloud.githubusercontent.com/assets/583202/9747265/bb580802-564a-11e5-99d5-f3cce43411f6.png)

Regional email addresses also seem to validate correctly:

![screen shot 2015-09-08 at 4 58 11 pm](https://cloud.githubusercontent.com/assets/583202/9747277/cccbb80e-564a-11e5-910c-dbd0104f14f5.png)

DFurnes:
Here's the list of domains we currently use for "suggestions":

> "yahoo.com", "google.com", "hotmail.com", "gmail.com", "me.com", "aol.com", "mac.com",
> "live.com", "comcast.net", "googlemail.com", "msn.com", "hotmail.co.uk", "yahoo.co.uk",
> "facebook.com", "verizon.net", "sbcglobal.net", "att.net", "gmx.com", "mail.com", "outlook.com",
> "aim.com", "ymail.com", "rocketmail.com", "bellsouth.net", "cox.net", "charter.net", "me.com",
> "earthlink.net", "optonline.net", "dosomething.org"

It looks like we already have some common international domains, but perhaps it'd be good to look into any that we're missing?

DFurnes:
Also pinging @julielorch, because I heard you've been giving this some thought! :smile:

angaither:
waiting on recs from @julielorch

joolaay:
@DFurnes @angaither reviewing auth flow ideas at this week's tech sync, this is part of it. Here's some more info: https://docs.google.com/document/d/1BvhHRUy7KkiZ2AgtfhtvHNtUGjubCnbGYuPtZGRXevU/edit#

joolaay:
@angaither sorry, which recs?

angaither:
@julielorch based on the doc you linked, but it looks like, other than content changes, there's no dev work to be done on the auth flow

joolaay:
[...]

PR #5273, opened 2015-09-28
Validate additional international domains with Mailcheck.

DFurnes:
References #5046. Adds `sina.com` to list of "suggestion" domains.

Also switches to extending [Mailcheck's defaults](https://github.com/mailcheck/mailcheck/blob/4ad175791da491f39ebb99c33e8ace9736372c56/src/mailcheck.js#L16-L25) rather than completely overriding with a custom list... this allows us to take advantage of upstream updates while still maintining our own custom additions. (The ability to extend these defaults rather than outright replacing them was added in Mailcheck v1.1.1).

Side benefit? We also get some more common domestic & international domains (that've been added since we started using Mailcheck) for free! The added domains are listed below:

> telus.net, optusnet.com.au, qq.com, sky.com, icloud.com, sympatico.ca, xtra.co.nz, web.de, rogers.com, verizon.net, btinternet.com, shaw.ca

:white_check_mark: 

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/validators/auth.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/validators/auth.js
@@ -12,8 +12,16 @@ define(function(require) {
   "use strict";
 
   var Validation= require("dosomething-validation");
-  var mailcheck = require("mailcheck");
   var setting = require('../utilities/Setting');
+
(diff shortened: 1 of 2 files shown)


angaither:
such fance!

sbsmith86:
coooool :+1:

weerd:
Awesomeness :cocktail: :+1:

Issue #5136, opened 2015-09-16
SEO: Ensure proper hreflang and canonical URL markup


mshmsh5000:
Relates to #4910 and this article:

https://support.google.com/webmasters/answer/189077?hl=en

Two principles I think we can agree on:
1. If a page has variations in other languages, they should be noted in the header as `<link rel="alternate" hreflang="es-mx" href="https://www.dosomething.org/mx/my/url" />`
2. If a page has no variations in other languages, but is accessible by multiple URLs, the header should specify the [canonical URL](https://support.google.com/webmasters/answer/139066?hl=en) as `<link rel="canonical" href="https://www.dosomething.org/my/url" />`

weerd:
@mshmsh5000 @angaither any suggestions on how to best determine if the page _has_ variations in other languages? Some pages (like Facts pages) might have some stuff translated which has been included in POEditor, but not the rest of the page content:

![image](https://cloud.githubusercontent.com/assets/105849/10110036/a47f5be8-6398-11e5-9075-da9d708e774a.png)

mikefantini:
@weerd what's the url for that fact page?

weerd:
@joe just showed me that if a node has translations the `translations` object should be populated with the available translations (which I didn't have before since my campaigns on local are not translated, but I just added one).

![image](https://cloud.githubusercontent.com/assets/105849/10110214/d39cce14-6399-11e5-8cda-9a4f1f92e244.png)

However, stuff with partial content translated will still pose an issue.

weerd:
@mikefantini it's on my local since my local has the .PO file imported with some translations from POEditor.

itsjoekent:
are fact pages actually supposed to have two different languages? that doesnt seem right

weerd:
@deadlybutter it's not that they have two different languages, it's that there's `node` content and content from DB getting output on the same page. The DB content is translated via `t()` and the **.po** file, while currently we aren't translating Fact page `node` content.

itsjoekent:
gotcha... yeah this is a tricky one

weerd:
@mshmsh5000 for number 2 in the issue, by "accessible by multiple URLs", do you mean content that may have redirects from other urls? Is there an example on the DS site I can refer to?

weerd:
[...]

PR #5275, opened 2015-09-28
Output head links for available translated content nodes.

weerd:
Addresses part of #5136
#### What's this PR do?

This PR adds the ability to output `<link>` tags in the `<head>` that provide references to translated content if it's available, so it can help search engines show a user appropriate links to site content pages depending on user's language/country origin. So if the user is in Mexico, they will see search results that link them to the `/mx/` page if it's available. 

So, for the below content page:

```
http://dev.dosomething.org:8888/campaigns/patient-playbooks
```

If there is a mexican spanish translation available, there will be the following link in the head:

```
<link rel="alternate" hreflang="es-mx" href="http://dev.dosomething.org:8888/mx/campaigns/playbooks-para-pacientes">
```

There will be `<link>` tags added for all available content, and it will **not** add tags if url in question is already a specific language url; no `<link>` tag for `es-mx` if url in question is: `http://dev.dosomething.org:8888/mx/campaigns/playbooks-para-pacientes`.

Also, no `<link>` tags added for global english.

Start at the `paraneue_dosomething_preprocess_node()` function, specifically at the `$regional_url_data` variable.

If you so desire, pull down the branch load up a node page that has translations and confirm `<link>` tags for the languages is output.
#5136

---

@angaither 
cc: @mshmsh5000

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
@@ -359,7 +359,11 @@ function paraneue_dosomething_get_regional_url_data($node) {
     return NULL;
   }
 
-  $supported_languages = variable_get('dosomething_global_language_map');
+  $supported_languages = language_list();
+  foreach ($supported_languages as $language) {
+    $language->country = dosomething_global_convert_language_to_country($language->language);
+  }
+
   $regional_url_data = [];
 
   foreach ($available_translations as $language_key => $translation) {
@@ -369,7 +373,7 @@ function paraneue_dosomething_get_regional_url_data($node) {
     }
 
     if (array_key_exists($language_key, $supported_languages) && $language_key !== $language_url->language) {
-      $country = strtolower($supported_languages[$language_key]['country']);
+      $country = strtolower($supported_languages[$language_key]->country);
       $regional_url_data[$language_key] = $base_url . '/' . $country . '/' . drupal_get_path_alias('node/' . $node->nid, $language_key);
     }
   }
--- a/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
@@ -344,6 +344,40 @@ function paraneue_dosomething_get_themed_placeholder_reportbacks($count = 6) {
 }
 
 /**
+ * Get regional url data for specified node if it has translated content available.
+ *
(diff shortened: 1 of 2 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -344,6 +344,40 @@ function paraneue_dosomething_get_themed_placeholder_reportbacks($count = 6) {
...
+  global $language_url;
+  global $base_url;
+
+  $available_translations = dosomething_helpers_isset($node->translations->data);
+
+  if (!$available_translations) {
+    return NULL;
+  }
+
+  $supported_languages = variable_get('dosomething_global_language_map');
```
angaither: we should use [language_list()](https://api.drupal.org/api/function/language_list/7) here
so it's the languages actually installed/enabled on our site

weerd: Oh, so `dosomething_global_language_map` isn't that? Cause it outputs the list of languages we're supporting?

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -344,6 +344,40 @@ function paraneue_dosomething_get_themed_placeholder_reportbacks($count = 6) {
...
+  $regional_url_data = [];
+
+  foreach ($available_translations as $language_key => $translation) {
+    // Skip creating regional url for global english.
+    if ($language_key === 'en-global') {
+      continue;
+    }
+
+    if (array_key_exists($language_key, $supported_languages) && $language_key !== $language_url->language) {
+      $country = strtolower($supported_languages[$language_key]['country']);
```
angaither: country won't be mapped if we use language_list, so there's a convert to language function in `dosomething_global`

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
angaither: yes, but there's some additional countries in there for auth modal form, so it's safer to use the installed language list from droops

angaither: you might have to do a `drush fra -y` to see the additional ones

weerd: ah, ok. good to know! I'll update :smiley_cat:

angaither:
:+1:

PR #5296, opened 2015-09-28
remove ._dosomething.northstar.install and move .dosomething_northsta…

### LINKED ISSUE #5267: Remove excess dosomething_northstar files from docroot
In the project root:

```
$ ls -lA | grep northstar
-rw-r--r--   1 mholford  staff    4096 Sep 18 13:52 ._dosomething_northstar.install
-rw-r--r--   1 mholford  staff     327 Sep 18 13:52 dosomething_northstar.install
```

There's no `.install` file in the Northstar module itself, so first we need to determine whether this is supposed to be there. Then clean up the files.

chloealee:
…r.install file to correct location
#### What's this PR do?

removes ._dosomething.northstar.install and moves .dosomething_northstar.install file to correct location (lib/modules/dosomething/dosomething_northstar/dosomething_northstar.install)

Fixes #5267


angaither:
:+1:

PR #5306, opened 2015-09-29
hide secondary tabs from regional admins

### LINKED ISSUE #5305: secondary tabs showing on node/edit/add
screen shot 2015-09-29 at 11 17 44 am

these should be hidden for regional admins

angaither:
Fixes #5305

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -91,7 +91,7 @@ function dosomething_global_menu_alter(&$items) {
  */
 function dosomething_global_menu_local_tasks_alter(&$data, $route_name) {
   // Verify we're operating on a node edit page
-  if ($route_name['path'] != 'node/%/edit') {
+  if ($route_name['path'] != 'node/%/edit' || $route_name != 'node/%/edit/add/%/%') {
       return;
   }
   // Only perform this for regional admins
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -91,13 +91,12 @@ function dosomething_global_menu_alter(&$items) {
...
 function dosomething_global_menu_local_tasks_alter(&$data, $route_name) {
   // Verify we're operating on a node edit page
-  if ($route_name['path'] != 'node/%/edit') {
-      return;
-  }
-  // Only perform this for regional admins
-  if (dosomething_global_is_regional_admin()) {
-    // Remove tabs from the edit page
-    unset($data['tabs'][1]['output']);
+  if ($route_name['path'] === 'node/%/edit' || $route_name['path'] === 'node/%/edit/add/%/%') {
```
sergiitk, line 94: In order to avoid PHP notices, check if `$route_name['path']` is set.

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -91,13 +91,12 @@ function dosomething_global_menu_alter(&$items) {
...
+    // Only perform this for regional admins
+    if (dosomething_global_is_regional_admin()) {
+      // Remove tabs from the edit page
+      unset($data['tabs'][1]['output']);
```
sergiitk: In order to avoid PHP notices, check if `$data['tabs']`, `$data['tabs'][1]` and `$data['tabs'][1]['output']` are set.

lib/modules/dosomething/dosomething_global/dosomething_global.module
angaither, line 94: since this is in menu_local_task alter those arguments will not be null

blisteringherb:
:+1:

PR #5307, opened 2015-09-29
Dr00ps style overrides cleanup

DFurnes:
# Changes

Just tidying up the overrides stylesheet. :leaves: Nothing terribly exciting going on, but I recommend going commit-by-commit when reviewing since there's a couple of different changes.


sbsmith86:
cool beans :+1:

PR #5308, opened 2015-09-29
Update logic for the publish transltion checkbox.

### LINKED ISSUE #5260: Translation appears to be unpublished but has published status on /translate
Edit of the version is unpublished:

image

But appears as published on /translate:

image

angaither:
### what does this pr do

Set the 'publish translation checkbox' to match the status of the translation
I set the default value to false, in #5180 but once a translation was published, the checkbox was still unchecked. This is some gross work around, because it appears there isn't a setting to not autopublish translations.
### additional context

there's no setting for auto publish translation content?
### apologies

this code seems wildly unnecessary    
### tickets

Fixes #5260
[see comment](https://github.com/DoSomething/phoenix/issues/5260#issuecomment-144094623)


blisteringherb:
:+1:

PR #5311, opened 2015-09-29
Fixes logout URL

### LINKED ISSUE #5283: Incorrect logout link for BR user
When logged in as a Brazilian user, this is the logout link

https://files.slack.com/files-pri/T024GV2BW-F0BESA4J1/screen_shot_2015-09-28_at_4.11.38_pm.png

/bruser/logout

itsjoekent:
#### What's this PR do?

Fixes the logout URL (on path prefixes it fails)

Go to any page, including path prefixed ones, and try logging out

Issue was the user/logout was being appended to the path prefix, so you'd get a url like ususer/logout. I tried initially just adding a slash to /user/logout, which fixes it on path prefixes. However on other pages this URL breaks. The solution was to just create the URL from scratch and guarantee it's the same URL for everyone, every time on every page.

Fixes #5283

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/system/partials/navigation.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/system/partials/navigation.tpl.php
@@ -35,11 +35,11 @@
           <a href="#" class="navigation__dropdown-toggle"><?php print $user_identifier ?></a>
           <ul>
             <li><?php print l("My Account", 'user/'. $user->uid); ?></li>
-            <li><a id="link--logout" href="<?php print $GLOBALS['base_url'] ?>/user/logout"><?php print t('Log Out'); ?></a></li>
+            <li><?php print l(t('Log Out'), 'user/logout', array('attributes' => array('class' => array('secondary-nav-item')))) ?></li>
           </ul>
         </li>
       <?php else: ?>
-        <li class="login"><a id="link--login" class="secondary-nav-item" href="<?php print $front_page; ?>user/login" data-modal-href="#modal--login"><?php print t('Log In'); ?></a></li>
+        <li class="login"><?php print l(t('Log In'), 'user/login', array('attributes' => array('class' => array('secondary-nav-item'), 'data-modal-href' => '#modal--login'))) ?></li>
       <?php endif; ?>
     </ul>
   </div>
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/templates/system/partials/navigation.tpl.php
```diff
@@ -35,7 +35,7 @@
           <a href="#" class="navigation__dropdown-toggle"><?php print $user_identifier ?></a>
           <ul>
             <li><?php print l("My Account", 'user/'. $user->uid); ?></li>
-            <li><a id="link--logout" href="<?php print $front_page; ?>user/logout"><?php print t('Log Out'); ?></a></li>
+            <li><a id="link--logout" href="<?php print $GLOBALS['base_url'] ?>/user/logout"><?php print t('Log Out'); ?></a></li>
```
blisteringherb: I think we should be using l() here instead of printing out each part of the url.

itsjoekent: oh I never heard of that function, looks like what we need

sergiitk:
Yes, `l()` function needed. Also applies for login link.

itsjoekent:
@sergii-tkachenko done and done

lib/themes/dosomething/paraneue_dosomething/templates/system/partials/navigation.tpl.php
```diff
@@ -35,11 +35,11 @@
           <a href="#" class="navigation__dropdown-toggle"><?php print $user_identifier ?></a>
           <ul>
             <li><?php print l("My Account", 'user/'. $user->uid); ?></li>
-            <li><a id="link--logout" href="<?php print $front_page; ?>user/logout"><?php print t('Log Out'); ?></a></li>
+            <li><?php print l(t('Log Out'), 'user/logout', array('attributes' => array('class' => array('secondary-nav-item')))) ?></li>
```
sergiitk: Just to make sure, will it generate `id="link--logout"`, as it used to be?

itsjoekent: Oh forgot about that one. It doesnt apply any styles, what do we use it for?

sergiitk: Ask @DFurnes

DFurnes: It's used as a hook for CasperJS tests.

sergiitk: I can see it here: https://github.com/DoSomething/phoenix/blob/dev/tests/integration/user/login.js#L28

sergiitk:
Let's bring back the ids and we're good to go :+1:

PR #5313, opened 2015-09-29
Fixes #5183: Redirect requests based on user/GeoIP info

### LINKED ISSUE #5183: Redirect a global URL request to translated version, if available
### User story (primary): anonymous -> 301
- If I am an anonymous visitor,
- and I am making a request from a country where we provide translated content (e.g., Mexico),
- and I request a bare URL (e.g., `/campaigns/sample`),
- and there is a translation available (e.g., `/voluntario/ejemplo`),
- then issue a `301 redirect` to the translation's URL.
### User story (secondary): authenticated -> 302
- If I am an authenticated visitor,
- and my profile includes a language preference,
- and I request a bare URL (e.g., `/campaigns/sample`),
- and there is a translation available (e.g., `/voluntario/ejemplo`),
- then issue a `302 redirect` to the translation's URL.
### Clues
- [Custom logic here](https://www.drupal.org/node/1280468#comment-6229920), based on the stock Translation Redirect behavior
- [https://www.drupal.org/project/multilink](MultiLink Redirect), "A module which might save a _lot_ of work...if you have content in two or more languages."

mshmsh5000:
#### What's this PR do?

Addresses the two use cases in #5183, specifically when and how to redirect anonymous and authenticated user requests to node URLs.

All of the work is in `dosomething_global.module`. Start at this comment in `dosomething_global_init()`:

```
  // Handle translation redirects for authenticated and anonymous users when  
  // they request bare (Global English) URLs.
```

The code below calls a different function depending on whether the user is authenticated: `dosomething_global_redirect_node_auth()` or `dosomething_global_redirect_node_anon()`. The rest of the logic is in those two functions.

Both functions end with calling the helper function `dosomething_global_redirect_node_language()`, which is also part of this PR.

_As an anonymous user:_ Use Chrome's [ModHeaders](https://chrome.google.com/webstore/detail/modheader/idgpnmonknjnojddfkpgkljpfnnfcklj?hl=en) extension, or something similar, to fake requests from different locations to node content. The header is

`X-Fastly-Country-Code: US | BR | MX | etc.`
1. Request a bare URL (e.g., `/campaigns/test`) from a supported country (e.g., `BR`) where the node has a translated, published version. You should get redirected.
2. Request a bare URL (e.g., `/campaigns/test`) from a supported country (e.g., `BR`) where the node has a translated, unpublished version. You should not get redirected.
3. Request a bare URL (e.g., `/campaigns/test`) from an unsupported country (e.g., `IN`). You should not get redirected.

_As an authenticated user:_ Update the user profile to test with different language settings.

_N.b. If you are a privileged user, you will not get redirected._

We base "privileged" on whether you can access the admin area (permission: `access administration menu`).
1. Request a bare URL where the node has a published version in your language. You should get redirected.
2. Request a bare URL where the node has an unpublished version in your language. You should not get redirected.

This gets hella bejesus confusing, so the goal in the code is to make the logic as simple as possible.
#5183

N/A
#### Questions:
- Does the knowledge base need an update? _Probably_

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -7,6 +7,9 @@
 include_once 'dosomething_global.features.inc';
 define("TRANSLATABLE_TYPES", 'static_content,campaign,home');
 
+// Default language: Global English
+define('DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE', 'en-global');
+
 /**
  * Implements hook_init().
  */
@@ -256,8 +259,8 @@ function dosomething_global_get_language($account, stdClass $node = NULL) {
       $language = $account->language;
     }
     // If there's no published translation in their language try 'en-global'.
-    elseif ($node->translations->data['en-global']['status']) {
-      $language = 'en-global';
+    elseif ($node->translations->data[DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE]['status']) {
+      $language = DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE;
     }
     // Default to node language if there's no translation in their language and no 'en-global'.
     else {
@@ -293,7 +296,7 @@ function dosomething_global_is_translation_request() {
 
         // If the prefix matches a language prefix besides that of Global
         // English:
-        if ($matches[1] == $sys_lang->prefix && $sys_lang->language != 'en-global') {
+        if ($matches[1] == $sys_lang->prefix && $sys_lang->language != DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE) {
 
           return TRUE;
         }
@@ -327,7 +330,7 @@ function dosomething_global_redirect_node_auth() {
   // All languages, as an array of objects.
   $languages = language_list();
 
-  $global_lang_code = 'en-global';
+  $global_lang_code = DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE;
 
   if (empty($languages[$global_lang_code])) {
     watchdog('dosomething_global', "Can't load %lang language",
@@ -362,7 +365,7 @@ function dosomething_global_redirect_node_anon() {
 
   // Map GeoIP country to a system language. Default to 'en-global'.
   $user_country_code = dosomething_settings_get_geo_country_code();
-  $target_lang_code = 'en-global';
+  $target_lang_code = DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE;
 
   $lang_map = variable_get('dosomething_global_language_map', '');
   foreach ($lang_map as $lang_key => $lang) {
@@ -373,7 +376,7 @@ function dosomething_global_redirect_node_anon() {
   }
 
   // If we're actually supposed to show the en-global version:
-  if ('en-global' == $target_lang_code) {
+  if ($target_lang_code == DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE) {
     return;
   }
 
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -11,12 +11,13 @@ define("TRANSLATABLE_TYPES", 'static_content,campaign,home');
  * Implements hook_init().
  */
 function dosomething_global_init() {
+  global $user;
+
   // Verify we're dealing with a node edit with no translation specification in the URL
   if (arg(0) == "node" && is_numeric(arg(1)) && arg(2) == "edit" && null == arg(3)) {
     // Load the page node and user
     $nid = arg(1);
     $node = node_load($nid);
-    global $user;
 
     // Load the languages being used
     $src_language = $node->language;
@@ -35,6 +36,22 @@ function dosomething_global_init() {
       drupal_goto($new_url);
     }
   }
+
+  // Handle translation redirects for authenticated and anonymous users when
+  // they request bare (Global English) URLs.
+  //
+  // Passing 'noredirect=1' will defeat the redirect.
+  if (arg(0) == 'node' && is_numeric(arg(1)) && '' == arg(2) &&
+      !dosomething_global_is_translation_request() &&
+      empty($_GET['noredirect'])) {
+
+    if ($user->uid) {
+      dosomething_global_redirect_node_auth();
+    } else {
+      dosomething_global_redirect_node_anon();
+    }
+  }
+
 }
 
 /**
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -35,6 +36,22 @@ function dosomething_global_init() {
       drupal_goto($new_url);
     }
   }
+
+  // Handle translation redirects for authenticated and anonymous users when
+  // they request bare (Global English) URLs.
+  //
+  // Passing 'noredirect=1' will defeat the redirect.
+  if (arg(0) == 'node' && is_numeric(arg(1)) && '' == arg(2) &&
```
sergiitk: Maybe `empty(arg(2))` would read better than `'' == arg(2)`?

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -256,3 +273,143 @@ function dosomething_global_get_language($account, stdClass $node = NULL) {
...
+
+  $languages = language_list();
+  $matches = array();
+
+  if (preg_match('/([^\/]+)\//', request_path(), $matches)) {
+    foreach ($languages as $sys_lang) {
+
+      // If the prefix matches a language prefix besides that of Global
+      // English:
+      if ($matches[1] == $sys_lang->prefix && $sys_lang->language != 'en-global') {
```
sergiitk: Wr should check if `$matches[1]` is set (better `!empty()`) to avoid PHP notices.

lib/modules/dosomething/dosomething_global/dosomething_global.module
itsjoekent: not really a major deal, but instead of `' ' == arg(2)` couldnt this just be `empty(arg(2))` ?

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -256,3 +273,143 @@ function dosomething_global_get_language($account, stdClass $node = NULL) {
...
+    return;
+  }
+
+  // User language.
+  $user_lang_code = $user->language ?: 'en';
+
+  // All languages, as an array of objects.
+  $languages = language_list();
+
+  $global_lang_code = 'en-global';
```
sergiitk: Let's make a php constant for `en-global`.

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -35,6 +36,22 @@ function dosomething_global_init() {
...
+
+  // Handle translation redirects for authenticated and anonymous users when
+  // they request bare (Global English) URLs.
+  //
+  // Passing 'noredirect=1' will defeat the redirect.
+  if (arg(0) == 'node' && is_numeric(arg(1)) && '' == arg(2) &&
+      !dosomething_global_is_translation_request() &&
+      empty($_GET['noredirect'])) {
+
+    if ($user->uid) {
```
itsjoekent, line 51: does this throw errors in the log? should we wrap this in an isset() ?

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -256,3 +273,143 @@ function dosomething_global_get_language($account, stdClass $node = NULL) {
...
+
+  if (empty($languages[$global_lang_code])) {
+    watchdog('dosomething_global', "Can't load %lang language",
+      array('%lang' => $global_lang_code), WATCHDOG_ERROR);
+    return;
+  }
+  $language_global = $languages[$global_lang_code];
+
+  // Don't redirect for Global English speakers.
+  if ($user_lang_code == $language_global->language) {
```
itsjoekent: Can this be simplified as $user_lang_code == $global_lang_code ? I don't see anywhere else we use $language_global

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -256,3 +273,143 @@ function dosomething_global_get_language($account, stdClass $node = NULL) {
...
+ * @param $node
+ * @param $language_code
+ * @param int $redirect_status
+ */
+function dosomething_global_redirect_node_language($node, $language_code, $redirect_status = 302) {
+
+  $languages = language_list();
+
+  // Can only redirect to an published, translated version.
+  if (!empty($node->translations->data[$language_code]) && (bool)$node->translations->data[$language_code]['status']) {
```
itsjoekent, line 408: hmm, will accessing ['status'] throw log errors if the translation doesn't exist?

lib/modules/dosomething/dosomething_global/dosomething_global.module
mshmsh5000, line 51: Bit of Drupal-ese: `$user->uid` is always set -- anonymous users have `$user->uid = 0`.

mshmsh5000, line 51: https://api.drupal.org/api/drupal/modules!user!user.module/function/user_is_logged_in/7

lib/modules/dosomething/dosomething_global/dosomething_global.module
mshmsh5000: Yes! Thanks. This was leftover from part of a previous implementation.

lib/modules/dosomething/dosomething_global/dosomething_global.module
mshmsh5000, line 408: If the translation doesn't exist, the code after `!empty($node->translations->data[$language_code])` won't execute. If `$node->translations->data[$language_code]` isn't empty, then `['status']` will be either `1` or `0`.

lib/modules/dosomething/dosomething_global/dosomething_global.module
itsjoekent, line 51: ooo thats useful to know

lib/modules/dosomething/dosomething_global/dosomething_global.module
itsjoekent, line 408: gotcha

lib/modules/dosomething/dosomething_global/dosomething_global.module
mshmsh5000: Good idea.

mshmsh5000:
Example of simple local tests (assuming a node with published translations) for the anonymous use case:

```
$ curl -I -H "X-Fastly-Country-Code: BR" http://dev.dosomething.org:8888/campaigns/boss-test-campaign
HTTP/1.1 301 Moved Permanently
[...]
Location: http://dev.dosomething.org:8888/br/voluntario/boss-test-campaign

$ curl -I -H "X-Fastly-Country-Code: MX" http://dev.dosomething.org:8888/campaigns/boss-test-campaign
HTTP/1.1 301 Moved Permanently
[...]
Location: http://dev.dosomething.org:8888/mx/campa%25C3%25B1as/boss-test-campaign
```

sergiitk:
Looks good!

PR #5317, opened 2015-09-29
Changes campaign translation settings.

### LINKED ISSUE #5198: Campaign Content added by DS Mexico only appearing in Active view - not in Admin/Edit
The content for Do Something Mexico campaign: Noche De Monopoly (https://www.dosomething.org/node/5757 ) in the active view of the campaign screen shot 2015-09-23 at 10 48 34 am
but is not reflected in the Admin/Edit view of the campaign 
screen shot 2015-09-23 at 10 46 44 am

sergiitk:
#### What's this PR do?
- Changes campaign translation settings:
  image
  image

Should fix #5198.


angaither:
this is fine, but we did this before and came across this error #5042 because none of our local admins have a language set

sergiitk:
I just tried to save the article as an admin and it worked. I think it may be related to the language selector, which I also made visible.

sergiitk:
@mikefantini let's test this carefully.

Issue #5285, opened 2015-09-28
MX & BR: SIGNUP EMAIL: Merge tags are pulling in GL content


marahml:
Merge tags for MX Signup Email should pull in merge tags for MX content. 

![image](https://cloud.githubusercontent.com/assets/6784628/10147942/d9c0bbc2-65fe-11e5-8ede-c446c92ee186.png)

mikefantini:
let's hold on this until #5282 is completed because fixing this may resolve the issue

PR #5323, opened 2015-09-30
Fixing syntax error and ensuring user's country is set for emails

### LINKED ISSUE #5286: GL: SIGNUP EMAIL: Incorrect template showing for GL signup email
Template shown below is incorrect (note that this is NOT the GL signup email, this is just an example of the template that is showing: 
image

Correct template looks like: 

image

blisteringherb:
#### What's this PR do?

When checking for the appropriate language to use when getting a signup, the code was trying to point to the wrong name of the $node object property.

Also making sure the user's country is set so that the template is properly selected.

If you don't have the messagebroker setup working locally, this will have to be tested locally by signing up for a campaign with your BR or MX proxy on and looking at the emails you receive. 

Fixes #5286 and #5285

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -247,7 +247,7 @@ function dosomething_global_convert_country_to_language($country) {
 function dosomething_global_get_language($account, stdClass $node = NULL) {
   if ($node) {
     // If there is a published translation for the node in the user's language, use it.
-    if ($node->translations->data['$account->language']['status']) {
+    if ($node->translations->data{$account->language}['status']) {
       $language = $account->language;
     }
     // If there's no published translation in their language try 'en-global'.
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -247,7 +247,7 @@ function dosomething_global_convert_country_to_language($country) {
 function dosomething_global_get_language($account, stdClass $node = NULL) {
   if ($node) {
     // If there is a published translation for the node in the user's language, use it.
-    if ($node->translations->data['$account->language']['status']) {
+    if ($node->translations->data{$account->language}['status']) {
```
sergiitk, line 250: Why curly braces? Did I miss a change in array access element syntax?

blisteringherb, line 250: It's not an array, it's an object, so the old syntax was always evaluating to false.

sergiitk, line 250: I found this on manual, curly and square braces are interchangeable when accessing array element:
http://php.net/manual/en/language.types.array.php#language.types.array.syntax.accessing
I still won't recommend using this notation as i ist very uncommon.

> It's not an array, it's an object

But you're accessing it here as an array. Did you mean 

``` php
if ($node->translations->data->{$account->language}['status']) {
```

Note the `->` after data.

blisteringherb, line 250: Can we leave it as is for now and circle back to it later for refactoring at some point so we can move forward with testing in Thor?

sergiitk, line 250: Yes, but create please create an issue for this.

sergiitk:
:+1:, but needs syntax improvements.

blisteringherb:
http://stackoverflow.com/questions/3235632/dynamic-access-to-a-php-array

PR #5324, opened 2015-09-30
Adding grouping to the finder query to prevent dupes in results

blisteringherb:
- Adding the group parameters changes the format of the output and
  returns a set of results grouped in our case by entity id. Only the
  top result from each group is returned, and because the result matching
  the user's language is boosted, it should be returned first in all cases.


sergiitk:
Looks great. Do you want me to test this on local env first?

blisteringherb:
@sergii-tkachenko If you'd like to. You'll have to run `ds grunt` to rebuild the js.

sergiitk:
:+1:

PR #5329, opened 2015-10-01
Using url() to build links for the home page campaign tiles

### LINKED ISSUE #5282: on thor: mx campaigns 404-ing
### quick form issue

home page links are missing country code
### long form issue

from the homepage node
as a user viewing `https://thor.dosomething.org/mx`
I click on a campaign in the homepage and get directed to `https://thor.dosomething.org/voluntario/gtfo-mx?source=node/1141`

this should be `https://thor.dosomething.org/mx/voluntario/gtfo-mx?source=node/1141`

blisteringherb:
#### What does this PR do?
- The url() function permits language and prefix to be passed in as
  options to accomodate our url structure.

Look at the URLs on the home page for the campaign tiles to make sure they have the appropriate prefix

Fixes #5282

### DIFF
--- a/lib/modules/dosomething/dosomething_home/dosomething_home.module
+++ b/lib/modules/dosomething/dosomething_home/dosomething_home.module
@@ -18,9 +18,6 @@ function dosomething_home_preprocess_node(&$vars) {
   // Adds $subtitle variable.
   $vars['subtitle'] = $vars['field_subtitle'][0]['safe_value'];
 
-  // Store this node as the signup source.
-  $source = '?source=node/' . $vars['nid'];
-
   // Prepare campaign thumbnails for homepage
   $campaigns = array();
   if(isset($vars['field_campaigns'])) {
@@ -44,7 +41,7 @@ function dosomething_home_preprocess_node(&$vars) {
       $tiles['nid'] = $nid;
       $tiles['is_staff_pick'] = $node->field_staff_pick[LANGUAGE_NONE][0]['value'];
       $tiles['title'] = $node->title;
-      $tiles['tagline'] = $node->field_call_to_action[$node->language][0]['safe_value'];
+      $tiles['tagline'] = $node->field_call_to_action[$lang_code][0]['safe_value'];
 
       // Get the campaign cover image. If there is no image for the user's language,
       // use the image from the node's language.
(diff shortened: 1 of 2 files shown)


DFurnes:
The tile pattern is also used on the "campaign group" node (e.g. `/volunteer/animals`) so it's worth checking that as well. (Not that it looks like it'll break anything, just as a heads up!)

blisteringherb:
@DFurnes Thanks for the heads up. There's also a similar construct in the campaign-block.tpl.php, but I'm not sure where that's used. Would you happen to know?

DFurnes:
Ack, I think that's just old code. :grimacing: Should be totally safe to delete.

blisteringherb:
@DFurnes Glad you brought up /volunteer/animals, b/c it's not working properly now on Thor with the 'br' or 'mx' preview. I think it's a separate ticket though b/c the URLs work, but the images are what's broken.

alt text

lib/themes/dosomething/paraneue_dosomething/templates/system/patterns/tile.tpl.php
```diff
@@ -20,7 +20,7 @@
 <article class="tile <?php print $classes; ?>">
 
   <?php if ($content['status']): ?>
-  <a class="wrapper" href="/<?php print $content['link']; ?>">
```
angaither, line 23: just making sure we wanna kill this slash

lib/modules/dosomething/dosomething_home/dosomething_home.module
```diff
@@ -82,7 +79,17 @@ function dosomething_home_preprocess_node(&$vars) {
...
+      $url_options = array(
+        'language' => $lang_code,
+        'query' => array(
+          'source' => 'node/' . $vars['nid'],
+        ),
+      );
+      if ($country = strtolower(dosomething_global_convert_language_to_country($lang_code))) {
+        $url_options['prefix'] = $country . '/';
+      }
+      $tiles['link'] = url('node/' . $nid, $url_options);
```
angaither, line 92: does this still print `/campaigns/[title]` don't we want `drupal_get_path_alias($nid, $lang_code)`?

lib/themes/dosomething/paraneue_dosomething/templates/system/patterns/tile.tpl.php
blisteringherb, line 23: Leaving in that slash breaks the URL. The url() function builds a url that can be put directly into the href.

angaither, line 23: :+1:

lib/modules/dosomething/dosomething_home/dosomething_home.module
blisteringherb, line 92: The url() function automatically will return an alias, and giving it the 'language' param will give you the appropriately translated version of the url, like http://dev.dosomething.org:8888/br/campanhas/gtfo-pt-br?source=node/1141

angaither:
kewl :+1:

PR #5333, opened 2015-10-01

blisteringherb:
Since we haven't yet brought the global changes back into dev, this will merge the dev changes back into global so we can continue deploying the global branch to Thor.


angaither:
lovely :+1:

Issue #5147, opened 2015-09-17
Update/add code to use t() on config field items


weerd:
Need to scour the code where ever the config setting field values are used and make sure there's a preprocess function that handles them first and we wrap `t()` around the items before they're passed through to the template.
- [x] campaign config: standard campaigns
- [x] campaign config: sms games
- [x] campaign config: closed campaign
- [x] campaign config: permalink
- [x] campaign config: goals and hot module
- [x] campaign config: win module
- [x] fact page: intro copy
- [x] notfound: projects copy
- [x] notfound: grants copy
- [x] notfound: search copy
- [x] payment: donate
- [x] search: no results found copy
- [x] settings: sitewide copy
- [x] user: login (works wonky)
- [x] user: profile
- [x] user: school
- [x] zendesk: form copy

weerd:
Leaving some notes of things for me to look into that aren't currently working:
- [ ]  permalink page share copy translation in share popups
- [ ] hot module shares
- [ ] win module shares
- [ ] problem shares
- [x] need way to better test non-owner permalink copy is properly translated
- [ ] need to confirm value for **Closed Campaign:Presignup Form Callout Copy** is not escaped in the input otherwise translation fails (update to use typed apostrophe)
- [ ] consider how to handle **NotFound: Search filter words** config field
- [ ] add "No results found" to POT which is a default set on `variable_get()` for search
- [x] Config Variables files has been updated, make sure to import the new one!
- [ ] Consider how to handle hardcoded strings with markup in the string... (e.g. dosomething_signup.signup_data_form.inc l:382)

*NOTE: the above items that weren't checked off have been moved off into their own respective issues.

mshmsh5000:
@weerd take a nap.

angaither:
@weerd we are de-prioritizing shares for the launch

weerd:
cool cool. Imma break the problematic ones into separate issues that can be de-prioritized or moved into 1.1 milestone :)

@mshmsh5000 :sleeping: soooon

weerd:
@angaither I reset the milestone for this issue to Global 1.0 and moved the share stuff, etc to their own issues marked as Global 1.1 on the respective issues :)

PR #5336, opened 2015-10-01
Updating School translateable config variable.

weerd:
Refs #5147
#### What's this PR do?

This PR updates the configuration variables for School section of variables within **DoSomething Config -> DoSomething User**.

Just confirming that the code looks correct.

The fields are translated both on the main site page, user-facing output and within the input fields in the CMS. They have all been manually tested to confirm that if a translation is available, the translation is used.
#5147

---

@angaither


angaither:
:+1: not sure if this is actually needed since we probably only have US schools in here, but can't hurt to add all the t()

weerd:
Yep yep. Figured it wouldn't hurt since I was already in there and on a roll ;)

PR #5337, opened 2015-10-01
Fixes #5162: Copy obeys user's lang over URL

### LINKED ISSUE #5162: copy is showing in user's language regardless of URL structure
if you go to the direct path https://thor.dosomething.org/br/voluntario/notas-de-shawn-pt the content should be in BR PT regardless of the user's language settings.

However, what's happening currently is drupal is determining what content to display to the user based on the user's language settings.

Language settings should really only determine what URL to take the user to, not what copy to display once on that URL.

i.e. if my language is BR PT, and I go to thor.dosomething.org/campaigns/notes-shawn, then I should instead be taken to thor.dosomething.org/br/voluntario/notas-de-shawn-pt and the content should be in BR PT.

but if I go straight to thor.dosomething.org/us/campaigns/notes-shawn, then the content should be in US EN even though my language is BR PT.

once we have the language picker, we'll allow users to specifically indicate that they want the global version, but until then I don't think we'll be allowing users to specifically access the GL EN version if a translation exists in their language.

since this is all such an edge case experience, I think this should be low priority.

see screenshots in comments below.

---

previous description:

created a Global English translation on https://www.dosomething.org/node/32/translate

a new Problem fact was referenced and different copy was written in the Supporting Statement on the Solution.

However, when you go to view https://www.dosomething.org/campaigns/13-gallon-challenge?language=en-global (which you can get to by clicking on the title in the Global English row of http://staging.beta.dosomething.org/node/1656/translate) it is not displaying the new Problem fact or the copy that is in the Supporting Statement on the Solution.

(previously the Supporting Statement on the Solution was showing the proper copy, but the Problem fact was wrong, now both are incorrect).

Everything is being stored properly in the database.

mshmsh5000:
#### What's this PR do?

Addresses inconsistent behavior where a logged-in user's profile language could override the language indicated by the URL prefix.

This is just a config update to the `dosomething_global` feature, specifically for the language detection settings (`admin/config/regional/language/configure`).
1. Log in.
2. Load a campaign in multiple languages, each in its own tab.
3. You should see the appropriate language for all copy (given that the translation exists).
4. Change your user's preferred language.
5. Reload the various campaign tabs.
6. You should still see the appropriate language in each tab, based on the language indicated by the URL.

N/A
#5162

N/A
#### Questions:
- Does the knowledge base need an update? _No_


angaither:
:+1:

PR #5339, opened 2015-10-01
Only validate address on edit profile page, when in the US.

### LINKED ISSUE #5314: User profile tries to validate address fields even if not present.
When viewing authenicated user profile and editing it, if the address fields don't show up for non-us users, the validation on the backend still tries to validate as if the fields were present.

angaither:
Since we only display the address field in the US, let's not try to validate the fields when we don't show them. 
Fixes #5314
Refs #5227


blisteringherb:
:+1:

PR #5346, opened 2015-10-01
Server Side Date Validation for different date formats

### LINKED ISSUE #5262: Date formats are not passing server side date validation
When registering from a country that is not US, users should enter their birthdays in the `DD/MM/YYYY`. This is passing the client side validation, but is not passing server side validation when the registration form is submitted. 

screen shot 2015-09-25 at 5 03 56 pm

sbsmith86:
#### What's this PR do?

This PR uses an `#after_build` function to change the date format setting for the birthdate field widget so that when users enter a date in the form of `DD/MM/YYYY` the date will pass server side date validation. 

Also, in this PR I update the format of default value for the birthdate field in the user profile edit form to switch based on country code.

Start at `dosomething_user.module`

You have try registering a new account as a global user and a US user and it should work with the different formats.

Fixes #5262

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -292,8 +292,9 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
         $form['#after_build'][] = 'dosomething_user_add_validation_attributes_to_address_fields';
       }
 
+      // Change date format for non-us users.
       if (dosomething_settings_get_geo_country_code() !== 'US') {
-        $form['#after_build'][] = 'dosomething_user_birthdate_settings';
+        $form['#after_build'][] = 'dosomething_user_birthdate_format';
       }
 
       // Custom validation & submission handlers.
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -305,6 +311,15 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
 }
 
 /**
+ * After build to change date format of birthdate field for non-us users
+ * so it passes date validation
+ */
+function dosomething_user_birthdate_format($form, &$form_state) {
+  $form_state['field']['field_birthdate']['und']['instance']['widget']['settings']['input_format'] = 'd/m/Y - H:i:s';
```
angaither: use constant `[LANGUAGE_NONE]` instead of `['und']` here

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -291,6 +291,12 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
       if ($is_validate_address_set) {
         $form['#after_build'][] = 'dosomething_user_add_validation_attributes_to_address_fields';
       }
+
+      // Change date format for non-us users.
+      if (dosomething_settings_get_geo_country_code() !== 'US') {
+        $form['#after_build'][] = 'dosomething_user_birthdate_format';
```
angaither: maybe to be more clear `dosomething_user_non_us_birthday_format`

sbsmith86:
@angaither updated!

angaither:
:+1:

PR #5360, opened 2015-10-02
Contact Emails for BR and MX users

### LINKED ISSUE #5078: BR/MX users should see email address instead of contact us modal
Instead of "Questions? Contact Us" which then links to a modal, users in Brazil and Mexico should be directed to email us.
- [x] BR users - "Questions? Email ajuda@dosomething.org"
- [x] MX users - "Questions? Email ayuda@dosomething.org"

and we decided global users should still have the same Zendesk flow as US users currently have.

---

@mnelson338 can you please edit this comment to have the email addresses you want these questions to go to?

also @mnelson338 how do you want global users to be treated? should they submit zendesk tickets like US users do? Personally this makes the most sense to me, but your call.

sbsmith86:
#### What's this PR do?

For BR and MX users, we display a contact email in the info bar instead of the zendesk form modal.

`preprocess.inc` sets all the variables we need for the info bar depending on if it's a country that has an email contact or uses the zendesk form.

Fixes #5078

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
@@ -305,7 +305,7 @@ function paraneue_dosomething_preprocess_node(&$vars) {
  * @see paraneue_dosomething_preprocess_node().
  */
 function paraneue_dosomething_add_info_bar(&$vars) {
-  // Initialize info_bar_vars with avariable that will always be set:
+  // Initialize info_bar_vars with a variable that will always be set:
   if (isset($vars['formatted_partners'])) {
     $info_bar_vars = array(
       'formatted_partners' => $vars['formatted_partners'],
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
@@ -305,7 +305,7 @@ function paraneue_dosomething_preprocess_node(&$vars) {
  * @see paraneue_dosomething_preprocess_node().
  */
 function paraneue_dosomething_add_info_bar(&$vars) {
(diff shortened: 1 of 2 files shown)


mikefantini:
@sbsmith86 think we can assign over to @weerd to review since @angaither will be out all week?

sbsmith86:
@mikefantini for sure

lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
```diff
@@ -305,23 +305,30 @@ function paraneue_dosomething_preprocess_node(&$vars) {
...
-  // Initialize info_bar_vars with variable that will always be set:
-  $zendesk_form_header = t(variable_get('dosomething_zendesk_form_header'));
-  $info_bar_vars = array(
-    'zendesk_form_header' => $zendesk_form_header,
-  );
-  // List $vars variable names to pass to info bar:
-  $info_var_names = array(
-    'formatted_partners',
-    'zendesk_form',
+  // Initialize info_bar_vars with a variable that will always be set:
```
weerd: Maybe remove this comment cause it's a little confusing since it doesn't apply anymore to the code immediately below it... if the `formatted_partners` isn't set then the `info_bar_vars` _doesn't_ get a vlaue w/ a var that will always be set...

weerd:
One quick comment on a comment, but otherwise splendid! :+1:

sbsmith86:
@weerd thanks for catching that! Was actually a bigger problem then just the comment because I needed `$info_bar_vars` to be initialized regardless if `formatted_partners` was set because we need to show the info bar even if it isn't sponsored. Phew. Power of peer review!

weerd:
haha, inadvertent catch! :fishing_pole_and_fish:

PR #5363, opened 2015-10-02
Global country code

### LINKED ISSUE #5263: Thor: Registration doesn't pass country code to Northstar

chloealee:
#### What's this PR do?

Saves country code in phoenix and correctly passes it to ns. 

lib/modules/dosomething/dosomething_user/dosomething_user.module lines 527 and 570
lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module lines 188-189

create new user on ds.org. Hit ns endpoint `GET /users/email/<email>` and make sure country code is returned. https://github.com/DoSomething/northstar/wiki/Spec#retrieving-a-user

Fixes #5263

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -68,7 +68,6 @@ function dosomething_northstar_register_user($form_state) {
     'method' => 'POST',
     'data' => json_encode($ns_user),
     ));
-
   if ($response->code === '200' && module_exists('stathat')) {
     stathat_send_ez_count('drupal - Northstar - user migrated - count', 1);
   }
@@ -170,7 +169,6 @@ function dosomething_northstar_build_ns_user($user, $form_state) {
     'addr_zip' => 'postal_code',
   ];
 
-
   $ns_user = [
     'email'         => $user->mail,
     'drupal_id'     => $user->uid,
@@ -193,10 +191,10 @@ function dosomething_northstar_build_ns_user($user, $form_state) {
   // Don't send blank passwords from the user update screen.
   if (!empty($form_state['values']['pass'])) {
     $ns_user['password'] = $form_state['values']['pass'];
+  }
 
   // Return fully built northstar user object.
   return $ns_user;
-  }
 }
 
 
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -68,6 +68,7 @@ function dosomething_northstar_register_user($form_state) {
     'method' => 'POST',
     'data' => json_encode($ns_user),
     ));
+
   if ($response->code === '200' && module_exists('stathat')) {
     stathat_send_ez_count('drupal - Northstar - user migrated - count', 1);
   }
@@ -191,10 +192,10 @@ function dosomething_northstar_build_ns_user($user, $form_state) {
   // Don't send blank passwords from the user update screen.
   if (!empty($form_state['values']['pass'])) {
     $ns_user['password'] = $form_state['values']['pass'];
-  }
 
   // Return fully built northstar user object.
   return $ns_user;
+  }
 }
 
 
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -183,8 +183,8 @@ function dosomething_northstar_build_ns_user($user, $form_state) {
   }
   foreach ($address as $ns_key => $drupal_key) {
     $field = $user->field_address[LANGUAGE_NONE][0];
-    if (isset($field[$drupal_key]['value'])) {
-      $ns_user[$ns_key] = $field[$drupal_key]['value'];
+    if (isset($field[$drupal_key])) {
+      $ns_user[$ns_key] = $field[$drupal_key];
```
angaither, line 187: won't this break sending the address as a whole?

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -522,6 +522,7 @@ function dosomething_user_new_user($form, &$form_state) {
 
 function dosomething_user_new_user_attributes($form, &$form_state) {
   dosomething_user_set_global_attributes();
+  dosomething_northstar_update_user($form_state);
```
angaither: since this is calling code in another module put this around it
`if (module_exists('dosomething_northstar')` so it doesn't fail if the module isn't there

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
chloealee, line 187: I think this was incorrect because when I `dpm(($field[$drupal_key]['value']));` it returns `B`. When I `dpm($field[$drupal_key]);` it returns `BR` (the country code I used in the header). Also, when I check if it reaches inside the `if (isset($field[$drupal_key]['value']))` block but `dpm('hi');`, it doesn't return `hi` like expected. 

Might have been assumed that it was the same format as the `foreach ($optional as $ns_key => $drupal_key)` above - checked that and correctly returns the information!

lib/modules/dosomething/dosomething_user/dosomething_user.module
chloealee: ahh makes sense - updated!

mikefantini:
assigning to @DFurnes because he has some experience with the :stars:

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
DFurnes, line 187: The reason we use the `'value'` index of the array is for Drupal fields, which expose a plain-text value but then also a `'safe_value'` index that has an escaped version of that string. From checking on my local, it seems like that isn't the case with these address fields so I _think_ your change should be good.

~~The remaining question is if there are some cases where a field on this `$address` array could have the value and safe_value indexes rather than just a plain string value. Do you know if this function is ever called outside of the new user registration flow?~~ Nevermind, I think this is safe! Just tried editing my user address from the admin panel and they're all plain string values.

DFurnes:
Looks good to me! :ship:

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
angaither, line 187: @chloealee right, sorry my question was a bit unclear
if you save a user profile with an address set, does the entire address with all values still send, I think those were stored in the `value` field

```
19 W. 21 St
NY, NY, USA
```

DFurnes, line 187: From testing it on my local, it seemed like that wasn't the case but I could've been looking at the wrong thing.

chloealee, line 187: ahh I see what you mean now, thanks for clarifying, @angaither ! I was only testing against creating a new user, where they don't ask for the address fields. Just tested on updating the profile and adding in the address and you're right - it doesn't send over the address fields since I deleted `value`. I'll work on updating this now!

PR #5373, opened 2015-10-05
Remove asset path

DFurnes:
# Changes

International sites are still configured to point to Akamai (`dosomething-a.akamaihd.net`) through the "Asset Path" theme setting. This is kinda lame, since we have Fastly now! :tada: These commits remove the theme setting and some of the hairier conditional logic happening in the template.

I also switched to including `app.css` via `drupal_add_css()`, since it automatically adds a cache-busting query string to the outputted style tag. Unless we get collisions, this may be a better solution than using the version as a query string since a malicious user couldn't just ping future asset versions and cache them prematurely.

For review: @DoSomething/front-end @sergii-tkachenko

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -479,9 +479,6 @@ function pull_db {
       echo "Setting Stage File Proxy to staging URL..."
       drush vset stage_file_proxy_origin "http://staging.beta.dosomething.org"
 
-      echo "Unsetting Asset Path theme setting..."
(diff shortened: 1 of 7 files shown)


sergiitk:
Looks :+1:! Do you want me to test it locally against international installation?

DFurnes:
I don't think anything should be appreciably different, but it couldn't hurt if you have the time!

sergiitk:
Didn't test it, but :shipit:

DFurnes:
:racehorse:

PR #5379, opened 2015-10-05
allows for mobile_app and mobile_app_date querys to be used at same time

### LINKED ISSUE #5328: Error on campaigns API query with mobile_app and mobile_app_date params
Separately, the params seem to be fine. Combined they throw an error.

**Example 1 - this works**
https://staging.dosomething.org/api/v1/campaigns.json?mobile_app=1

**Example 2 - and this works**
https://staging.dosomething.org/api/v1/campaigns.json?mobile_app_date=2015-10-01

**Example 3 - but this doesn't**
https://staging.dosomething.org/api/v1/campaigns.json?mobile_app=1&mobile_app_date=2015-10-01

it returns this...

```
{
  error: {
    message: "SQLSTATE[42S22]: Column not found: 1054 Unknown column 'm.field_mobile_app_date_value' in 'where clause'"
  }
}
```

More info from Andrea:

> I think because if the first and second are set, it’s trying to do the join statement twice

chloealee:
#### What's this PR do?

Allows for both mobile_app and mobile_app_date queries to be used at the same time.

http://dev.dosomething.org:8888/api/v1/campaigns.json?mobile_app=1&mobile_app_date=2015-10-01 should return campaigns that satisfy both `mobile_app` and `mobile_app_date` queries. 

Fixes #5328

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -458,10 +458,10 @@ function dosomething_campaign_build_campaigns_query($params = array()) {
   }
 
   if (isset($params['mobile_app_date'])) {
-    $query->join('field_data_field_mobile_app_date', 'm', 'm.entity_id = n.nid');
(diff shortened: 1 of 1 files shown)


jonuy:
:+1:

PR #5384, opened 2015-10-06
Update to fix PHP errors with regional tabs

### LINKED ISSUE #5383: Bug in Regional Admin tabs if no primary tabs present
If the array for `$vars['tabs']` is has an empty '#primary' array item, then PHP will throw a bunch of warnings.

weerd:
Fixes #5383
#### What's this PR do?

When passing `$vars[tabs]` to the paraneue_dosomething_edit_regional_admin_tabs()`function, if the`#primary` item in the array is empty, then the following code in the function shouldn't execute. It previously did and PHP would throw warnings. So now we check to make sure it's not an empty array; if it is, then we return back with the untouched .
#5383

---

@sbsmith86 @dfurnes


DFurnes:
:+1:!

PR #5385, opened 2015-10-06
Only add a prefix if the user's language has a corresponding country

### LINKED ISSUE #5366: Sign Up Transactional Email - Campaign Link Bug on Production
"Prove It" button links to https://thor.dosomething.org//node/?source=node/2805#prove for GL EN user.

image

cc: @marahml

blisteringherb:
- Prefixes should only be added if the user is coming from a supported
  country and not for global users

Fixes #5366

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -562,11 +562,14 @@ function dosomething_signup_get_mbp_params($account, $node, $opt_in) {
   $url_options = array(
     'language' => $language,
     'absolute' => TRUE,
-    'prefix' => strtolower(dosomething_global_convert_language_to_country($language)) . '/',
     'query' => array(
       'source' => 'node/' . $node->nid,
     ),
   );
+
+  if ($country = strtolower(dosomething_global_convert_language_to_country($language))) {
+    $url_options['prefix'] = $country . '/';
+  }
 
   $params = array(
     'email' => $account->mail,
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -562,12 +562,15 @@ function dosomething_signup_get_mbp_params($account, $node, $opt_in) {
   $url_options = array(
     'language' => $language,
     'absolute' => TRUE,
-    'prefix' => strtolower(dosomething_global_convert_language_to_country($language)) . '/',
     'query' => array(
       'source' => 'node/' . $node->nid,
     ),
   );
 
+  if ($country = strtolower(dosomething_global_convert_language_to_country($language))) {
```
sergiitk: I don't like this construct. This if always will be TRUE. Here's better version:

``` php
$country = dosomething_global_convert_language_to_country();
if (!empty($country) {
  $url_options['prefix'] = strtolower($country) . '/';
}
```

blisteringherb: It will not always be true. If that function returns nothing, it will evaluate to false.

sergiitk: Actually you're right. Still I prefer my version better, it's easier to read and more explicit.

mshmsh5000: Interesting that both @sergii-tkachenko and I misread this. @blisteringherb that's a sign to me that it's idiomatic PHP that we should remove in favor of something more explicit. If you really want to keep the construct, please add a comment explaining when and how it evaluates to FALSE, for Future Generations.

(Also a sign that @sergii-tkachenko and I need to read more carefully.)

PR #5386, opened 2015-10-06
Asset path wonkiness

### LINKED ISSUE #5374: Modernizr 404 due to country code prefix
It looks like `modernizr.js` is being loaded with the country code prefix... this path doesn't exist, so...

screen shot 2015-10-05 at 3 52 20 pm

Note the added `/us/content` on the beginning of the URL. Strangely, only happens to this one script.

DFurnes:
Fixes #5374.

Use `url()` to ensure we're always making root-relative URLs. I accidentally'd the prefixed `/` on the Modernizr script in #5373 so it was requiring relative and (duh) not existing. This fixes that and also runs asset paths through `url()` so we don't make that mistake again in the future!

Also renames `PARANEUE_DS_PATH` to `PARANEUE_PATH` since it reads easier.

For review: @weerd

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
@@ -5,7 +5,7 @@
  */
 function paraneue_dosomething_preprocess_html(&$vars) {
   // Add theme stylesheet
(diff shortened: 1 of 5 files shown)


weerd:
¡OLÉ! :dancer:

PR #5394, opened 2015-10-06
Kidding, let's _not_ pass asset URLs through `url()`.

DFurnes:
Once more, with feeling! :musical_score: 

It turns out that `url()` adds the required `/` to indicate a root-relative URL, but on international sites it _also_ seems to add the country code. Lame. So we'll just hardcode the path, but ensure that it always gets the prepended slash.

For more information, see #5373 and #5386.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/system/html.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/system/html.tpl.php
@@ -22,17 +22,17 @@
   <?php print $styles; ?>
 
   <!--[if lte IE 8]>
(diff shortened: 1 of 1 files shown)


DFurnes:
I'm 99% sure this should fix it, but can't test with absolute certainty since it's only occurring on environments with Fastly enabled. :cactus:

weerd:
:sheep: it!

sbsmith86:
seem's fine to me! I've been using a plugin to fake the fastly headers for more robust testing on my local.

DFurnes:
Oh, tricky! :computer:

itsjoekent:
@DFurnes https://github.com/DoSomething/phoenix/wiki/Fastly-Global-Header-Setup#quick-setup-for-local-testing

PR #5405, opened 2015-10-07
Use node's language for CTA.

### LINKED ISSUE #5404: remove global features from staging DB
when the global stuff got deployed to Staging it didn’t get removed properly from the DB

DFurnes:
Fixes #5404. This fix was originally applied in 1aa08b4, and got accidentally overwritten in 51b8254.

For review: @blisteringherb


mshmsh5000:
Yes :+1: -- We want to restore this code.

PR #5407, opened 2015-10-07
Fix broken references to DS_ASSET_PATH.

DFurnes:
`DS_ASSET_PATH` was removed in #5373, but I forgot to update these three calls to use the remaining PARANEUE_PATH constant (which is what `DS_ASSET_PATH` was getting set from when not set to a CDN path override). The `/` is added to force root-relative URLs (without getting country prefix, which seems to be automatically added if using the `url()` function).


weerd:
:+1:

Issue #5342, opened 2015-10-01
Enable the retrieve endpoint for the taxonomy_term resource


jonuy:
![screen shot 2015-10-01 at 5 08 19 pm](https://cloud.githubusercontent.com/assets/696595/10233933/7c063a62-685f-11e5-907a-2b0e1885b857.png)

I'm planning to use it for the mobile app to retrieve tag names that'll be displayed for campaign groups. So I'll be making calls like...

https://staging.dosomething.org/api/v1/taxonomy_term/670.json

It's sounding like it gets automatically disabled every time there's a deploy, so until then I'll just be manually reenabling it.

chloealee:
@angaither I'm having trouble finding the module name where this is located. I tried to implement a `hook_form_alter` in `dosomething_api.module` and a `hook_form_FORM_ID_alter` in `dosomething_helpers.module` using the form `id="services_edit_form_endpoint_resources"` but neither worked. Am I adding in the incorrect module?

angaither:
ahh, @chloealee services is a bit different, this is a setting that you have to change in the drupal settings and then export with features (kind of like what we did when adding new fields) 
check out [this page](http://dev.dosomething.org:8888/admin/structure/services/list/drupalapi/resources) locally

chloealee:
oh gotcha, thanks @angaither ! So, I'd click on it through that link, save it, and then export with the pre-checked feature? Export like this: https://www.drupal.org/node/2294809

angaither:
yes, except export with `drush fe` which I can walk you through if you want :smile:

chloealee:
oh awesome - that'd be great! I just found this as an example: 

```
features-export
explanation: Export a feature from your site into a module. If the feature exists, new components are added into that feature.
aliases: fe
Syntax: drush features-export name_of_feature component1 [component2...]
Example: drush fe my_new_blog_feature views_view:view_name
```

in my vagrant box, would I just do `drush fe enable_retrieve_endpoint resources`?

angaither:
to see what hasn't been exported you can run
`drush fc` 
it will show you a list of options and services will be one of them
from there, you can see what has/hasn't been exported and you can get the name of the feature component

then you will export to `dosomething_api`

chloealee:
oh great, thanks so much Andrea!! I just did `drush fe dosomething_api views_view:taxonomy` and it seems it created the module. How do I know if it worked correctly to submit for review?

angaither:
[...]

Issue #4571, opened 2015-06-05
Forward campaign activity to northstar


angaither:
When a user signs up and reportsback on a campaign, forward that data into northstar

angaither:
@chloealee

chloealee:
just chatted with @uy and thought of a possible solution - since Northstar already has a signup/reportback endpoint that forwards information onto Phoenix, make Phoenix web submissions go through Northstar instead of straight into it's own database. @angaither @weerd would love your insight if this is the best way to proceed!

angaither:
I'm not sure if we are quite at the point that we can remove drupal as a backend for these requests

@DFurnes might have more input on this when he returns...
but another possible solution is I have a [message broker campaign activity repo](https://github.com/DoSomething/mbc-logging) that takes campaign signups and reportbacks and stores them in mongo. since we already queue everything with mb, that might be a good thing to hook into and pass things into northstar

chloealee:
@angaither awesome, thanks! Should I get started via that direction or wait for Dave's input?

DFurnes:
Chloe & I just talked this over IRL. We're thinking the best course of action would be to try to mirror how we deal with users and see if we can create a synchronized copy of all this data in Northstar. The mobile app will only have a subset of campaigns so this hopefully shouldn't be _too_ heavy of a dataset, since I think we're still uncertain whether campaign activity should live in the user documents...

chloealee:
I just noticed that there is already a check to see if a campaign signup already exists [here](https://github.com/DoSomething/northstar/blob/dev/app/Http/Controllers/CampaignController.php#L119) so I don't think we'll actually run into the infinite loop between ns and phoenix.

Another question I forgot about that we ran into when first discussing is creating a new function that forwards the signup from phoenix to drupal - to hit the `POST user/campaigns/:campaign_id/signup` endpoint, I'll need the session token in the header. To get this, will I have to sign in the user via the `POST /login` endpoint first? That seems a little messy though.

chloealee:
@angaither @DFurnes whenever you get a chance for the above!

chloealee:
@angaither @DFurnes the only other thing I can think of is to maybe update the `signup` function in the Campaign Controller to first check if there is a session token and if so, validate the request body, else sign user in?

DFurnes:
[...]

PR #5408, opened 2015-10-07
Reportback

### LINKED ISSUE #5415: NS user load will never load from cache
https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module#L131

`if (!empty($northstar_user_info)) {` below will always evaluate to `FALSE`, because `$northstar_user_info` is not declared.

```
/**
 *Get user data from northstar.
 */
function dosomething_northstar_get_northstar_user($id) {
  if (!empty($northstar_user_info)) {
    $northstar_user_info = cache_get('northstar_user_info_' . $id, 'cache_northstar_user_info');
  } else {
```

chloealee:
#### What does this PR do?

Allows the `$full` param to be passed so we can set to `false` when bulk-processing (will not make call to ns). 

lib/modules/dosomething/dosomething_reportback/includes/Reportback.php line 142, then 162-175, then 127

lib/modules/dosomething/dosomething_reportback/includes/ReportbackItem.php line 95, then 120-133, then 80

When approve/promote/flag etc. reportback page is saved, northstar call is not run. Test with New Relic?

Fixes #5415


chloealee:
Hey @mshmsh5000 ! I think this is what you meant with the shallow load param but let me know if I'm totally off! 

Also, @weerd had a great suggestion to help with load time by putting all `drupal_id`s in an array and sending to ns all in one call rather than one at a time like I did here https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module#L128

mshmsh5000:
The code looks like what I imagined!

Can you please remove `._dosomething_northstar.install` from this PR? I'm guessing it's some temp file created by your editor.

[Here](https://github.com/DoSomething/phoenix/pull/5408/files#diff-14a05db5e6c9435171679a1e613d23dfR140) and [here](https://github.com/DoSomething/phoenix/pull/5408/files#diff-14a05db5e6c9435171679a1e613d23dfR140), I'd specify that this param controls whether to load full _user_ data, since the rest of the Reportback/Item data load in both cases.

Re: The great suggestion from @weerd, depends on whether NS has an endpoint to accept lists of IDs -- I just don't know. On the Drupal side, to maintain backwards compatibility, you could implement like 

```
/**
 * Get user data from Northstar.
 * 
 * @param mixed $ids Single user ID, or array of user IDs
 * @return 
 */
function dosomething_northstar_get_northstar_user($id) {
    // Normalize to array
    $id = (!is_array($id)) ? array($id) : $id;

    // ... Loop through the jamz
}
```

Also note that because of #5415 we're never loading from cache. We'll have to fix that, too, as a core performance bug.

chloealee:
@mshmsh5000 thanks for the comments! I don't know why those files got added and moved again but deleted `._dosomething_northstar.install` and moved `dosomething_northstar.install` to the correct place. Also specified user data in the comment blocks.

Re: suggestion from @weerd, it looks like you can only retrieve either one user or all users right now. Might be worth creating a separate issue for this to tackle?

Re: #5415 - added a fix and updated the PR description :)

mshmsh5000:
Looks good. Re: #5415, here's my suggested version:

https://gist.github.com/mshmsh5000/1f011e2073c05ba00652

I've added exception logging for error responses from Northstar. We should try to catch and log those, so we know when Phoenix and Northstar aren't playing well together.

This will add Watchdog log entries like this:

```
Exception: Error fetching Northstar user data, uid=1704977: 'Not Found' in dosomething_northstar_get_northstar_user() (line 153 of/var/www/dev.dosomething.org/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module).
```

Let me know what you think!

chloealee:
Thanks for this, Matt! Makes total sense and very smart. Added your suggestion :)

weerd:
Howdy! Yep yep, NS doesn't support that yet. It was a suggestion that would require a bit of extra lift to add that capability to NS to help further efficiency in those requests. Uy created an issue for it today: https://github.com/DoSomething/northstar/issues/212

So one thing to note w/ the full/shallow load approach. This will affect all `find()` related calls at the moment. So any calls to the API that aim to find RBs or RB Items based on filters passed via params in the URL request to API will be affected. Just something to note. Otherwise need to add some sort of functionality (another URL param? something in class transformer?) to allow deciding _when_ to do a `shallow` load instead of the full load. Noting here in case it affects current API calls already retrieving user info that was bringing back full details from NS.

@aaronschachter @jonuy ^^

chloealee:
thanks, @weerd ! @aaronschachter @jonuy this should be fine though since right now, it's just getting one reportback item at a time, correct?

jonuy:
We actually do query for a bunch of reportback-items at a time. On the Android app I think it's set to 25 at a time in order to populate the gallery of photos on the main feed.

@chloealee Do you think you'd want to tackle DoSomething/northstar#212 as an additional task on top of this issue? Not sure what your workload is looking like right now. Otherwise, I could try to help by doing that one.

chloealee:
oooh gotcha, thanks @jonuy! Right now, I'm working on sending address from phoenix->ns since I messed that up when I tried to send over global country code stuff, [Enabling the retrieve endpoint for the taxonomy_term resource](https://github.com/DoSomething/phoenix/issues/5342), and [forwarding campaign activity to northstar](https://github.com/DoSomething/phoenix/issues/4571). I'm not sure if I can finish all by Friday but will try my best to! Priority wise, what should be tackle this first?

mshmsh5000:
Sounds as though we need to resolve the find() problem that @weerd [mentions above](https://github.com/DoSomething/phoenix/pull/5408#issuecomment-146585864) before merging this. Or at least affirm that it won't screw up calls the native apps are making now.

Correct @aaronschachter @jonuy @chloealee?

chloealee:
From what I understand that makes sense! I'll get started on that. @aaronschachter @jonuy right now, we only want full load for the gallery of reportback items? Perhaps, per @weerd 's suggestion, I can add another URL param $gallery, and if that exists, change pass `true` instead of `false`?

weerd:
Cool! @chloe one suggestion would be to name the parameter something other than `$gallery`. That seems very specific to the case of loading the items for the gallery on mobile. But what if down the line we want to load the full data for some other purpose... passing a `?gallery=true` or something like that would seem odd and not make sense given a different context. I'd say name the parameter something that implies what it actually does which is indicate _retrieval of full user data_, so maybe a short word or two that could describe that? Not sure if I explained this clearly :grimacing:

chloealee:
@weerd super smart and totally makes sense - thanks for flagging! I'll name the parameter `$full_user` :)

weerd:
:+1:

aaronschachter:
So what should the updated `reportback-items` query look like in the mobile apps? `reportback-items?full_data=true` ?
`full_data` seems a little vague. Possibly something like `load_user_data`? `full_user_data` ?

jonuy:
I'll throw my vote for `?load_user_data=true` into the hat. :tophat:

chloealee:
@aaronschachter @jonuy you got it! `$load_user_data` it is.

mshmsh5000:
This follows the "do the most obvious thing" rule, which is always +1.

CTO
@mshmsh5000 http://twitter.com/mshmsh5000 @dosomething
http://twitter.com/dosomething

chloealee:
@weerd @mshmsh5000 added the `$load_user_data` param and confirms that it is working correctly! I was having trouble testing to make sure if it returned the ns user data because I think the cache already set to null but commented out the cache lines and worked fine.

lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
```diff
@@ -121,12 +121,22 @@ public static function find(array $filters = []) {
       throw new Exception('No reportback data found.');
     }
 
-    foreach($results as $item) {
-      // @TODO: remove need for passing variable for constructor check.
-      $reportback = new static(['ignore' => TRUE]);
-      $reportback->build($item);
+    if (isset($filters['load_user_data'])) {
```
weerd: I don't think you need the `load_user_data` check here, which forces you to repeat the code in both the `if` and then the `else`. I would recommend putting it in the `build()` function. Then it's only a single check there when building out the full object... helps keep it DRY :smile: Instead of passing a hardcoded `true/false` to the `build()` function, you can now just pass the value of the `$filters['load_user_data']` which itself is now a boolean :)

``` php
<?php
$load_user_data = dosomething_helpers_isset($filters['load_user_data'], NULL, FALSE);
$reportback->build($item, $load_user_data);
```

It'll pass `true` or `false` to the build function without need for the `if/else` and duplicate code :dancer:

weerd: the `dosomething_helpers_isset()` function is just a helper function to help deal w/ the whole `isset()` crap :)

lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
```diff
@@ -137,8 +147,9 @@ public static function find(array $filters = []) {
    * Build out the instantiated Reportback class object with supplied data.
    *
    * @param object $data
+   * @param bool  $full Boolean to decide whether to fetch full user data.
    */
-  private function build($data) {
+  private function build($data, $full = true) {
```
weerd: We should probs default this to `false` now that we have a URL parameter to control it, and since the resource file also defaults it to false in the services specification array.

lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
chloealee: oh this is so smart, thanks Diego! I was trying to figure out how to extract the code so I wouldn't repeat lines - so helpful!

lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
chloealee: changed!

weerd:
Coool! Looking good :dancer: :+1:

chloealee:
sorry - just found another bug with the cache - it's not setting quite right anymore. I'll submit another comment for final approval when it's working again!

chloealee:
@mshmsh5000 @weerd this should be the last one! Everything works as it should now :) Who do I need approval from to merge?

chloealee:
Oh right - I forgot about the undeclared variable part! That makes sense why it's working when I changed it back because it still makes the call to ns every time. 

I changed it back because right now, it returns the user info. the first time when it makes the call but when I refresh to test that `cache_get` is working properly, it returns `null` for all the user's data.  Found that if I delete the optional `$bin` parameter in `cache_set`, it now seems to work. 

One thing though - when I visit the site, I'm getting the error 

```
Warning: json_decode() expects parameter 1 to be string, object given in Reportback->build() (line 165 of /var/www/dev.dosomething.org/lib/modules/dosomething/dosomething_reportback/includes/Reportback.php).
```

but I'm pretty sure it is already passing it a string, especially because it worked the first time too. Is this a concern?

weerd:
@chloealee something must have changed that it's now passing an object. If it's popping up as an error message it's definitely a concern. We'd want to fix that! :wrench:

chloealee:
@mshmsh5000 @weerd thanks for your patience and help with this - finally figured out the issue and confirms that cache is working properly and no errors appear on the site anymore! 

@weerd I tried to change `load_user_data` to just `load_user` but for some reason, that broke everything. Looks like @aaronschachter also added the query string already - should we keep for now or still try to make the change?

weerd:
Might have been a cache clear thing again if it broke everything... but if it's too much lift if Aaron already added the query string on a released version of the app, then maybe hold off?

aaronschachter:
O Hai, the query string isn't in a release yet, so it's easy to change, just let me know what to do

weerd:
Cool! Then maybe give it a whirl @chloealee and just make sure to clear cache when you do, cause Droooops will be remembering the old infoz :wink:

chloealee:
@weerd always right! `drush cc all`: 2, Chloe: 0. Double fail for this week :-1: 

Just updated the param to `load_user` :) @aaronschachter

weerd:
@chloealee I had the same stats at first. Eventually you get used to the Drooops idiosyncrasies :wink:

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -125,20 +125,32 @@ function dosomething_northstar_update_user($form_state) {
...
+       'method' => 'GET',
+       ));
+
+    $northstar_user_info = $northstar_user_info->data;
+
+      if (!empty($northstar_user_info->error)) {
+        $error = sprintf("Error fetching Northstar user data, uid=%d: '%s'", $id, $northstar_user_info->error);
+        watchdog_exception('northstar', new Exception($error));
+      } else {
+        // die($northstar_user_info);
```
weerd: remove debug stuffz :8ball:

weerd:
Just one debug item to remove, other than that looks good! :+1:

chloealee:
gah - thanks for catching that! Removed :)

weerd:
:+1:

mshmsh5000:
:+1:

PR #5414, opened 2015-10-07
Updating home page URLs 

blisteringherb:
#### What does this PR do?

Currently in the dev branch even when the prefix is removed in the UI, the home page URLs still append a prefix with the country code. Additionally, the URLs for the home page are using `global $language`, when they should instead be going through the dosomething_get_language() function which respects the following precedence:
1. URL prefix
2. User's account language if published
3. System default language if published
4. Node's default language

The home page URLs are now using the prefix as defined in the language object returned by `language_list()`.

I've also updated the title and the CTA to use the same precedence so that URLs are in sync with the title and CTA. I did some debugging, and it doesn't look like the $node->language changes to the appropriate language on node_load(). It works fine for now in dev, but it doesn't seem to show the correct CTA when there are multiple translations.

Visit the home page and observe the URLs, titles and CTAs for the tiles. You can also change the prefix manually to make sure that the translations and URLs change properly with the addition of a prefix.

@mshmsh5000 @DFurnes

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -244,7 +244,7 @@ function dosomething_global_convert_country_to_language($country) {
   return NULL;
 }
 
-/*
+/**
  * Get the appropriate language for the application.
  *
  * @param $account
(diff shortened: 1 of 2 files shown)


DFurnes:
I think this seems okay! :ship:

PR #5428, opened 2015-10-08
Using the dosomething_global_url helper function to build signup link

blisteringherb:
- The campaign signup link should go through the same chain of hierarchy as
  the home page URLs to determine which language and then apply the appropriate
  prefix to the campaign signup URL


sergiitk:
:+1:

PR #5437, opened 2015-10-09
Enabling pathauto i18n node module

blisteringherb:
After translating around 200 nodes, rebuilding the URL alias table, and reindexing solr, it seems that not all the correct aliases are being created. I believe we need to enable pathauto_i18n_node in addition to the base pathauto_i18n module.

@mshmsh5000 @deadlybutter


itsjoekent:
I dunno morgan, I think this needs more comments 

I'll let it slide though, +1

mshmsh5000:
I'm with @deadlybutter -- :+1: despite your PR formatting rebellion.

PR #5450, opened 2015-10-13
Updates variable name to validate address.

### LINKED ISSUE #5435: Unset variable in user module
https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_user/dosomething_user.module#L291

This doesn't exist. It used to, but got renamed. Should be `$validate_address` as defined [here](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_user/dosomething_user.module#L228).

angaither:
This variable name was updated, but not everywhere, and was causing undefined errors, and address forms not to be validated. 
Fixes #5435


blisteringherb:
:+1:

PR #5454, opened 2015-10-13
Global footer

### LINKED ISSUE #5166: global footer needed
not sure exactly how we're doing this work, but want to make sure we have a place to discuss how we'll be doing this work.

DFurnes:
### Changes

Closes #5166. Adds theme settings to set different footers depending on which global country the current session is coming from (4c2978d). The footer (and theme settings admin panel) [will both fallback to the old general footer theme settings](https://github.com/DoSomething/phoenix/blob/417d532382d704251830bf46f7ed3b1148a7161f/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc#L192-L197) if no country-specific value has been set. 

I also removed some old theme settings that were no longer needed after we re-worked the footer markup in ~~Neue~~ Forge 6.0 (b125728).

I added a new helper function which returns the session's country code, but only if it's in our array of "whitelisted" countries. (5caab39) From what I could tell, this didn't already exist but please let me know if I missed something!
### Screenshots

Example "global" footer, filled with the specified links from @julielorch's [Google Doc](https://docs.google.com/document/d/1_nK9VZQ6DQIZ46A6XCxQqeK1TUcO1ssMyJUNSbyPDr4):
screen shot 2015-10-13 at 12 32 28 pm

---

For review: @angaither
/cc @weerd @sbsmith86

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/theme-settings.php
+++ b/lib/themes/dosomething/paraneue_dosomething/theme-settings.php
@@ -307,7 +307,7 @@ function _paraneue_dosomething_theme_settings_user(&$form, $form_state) {
   );
   $form_user = &$form['user'];
 
-  // Validaions.
+  // Validations.
   $form_user['validations'] = array(
     '#type'        => 'fieldset',
     '#title'       => t('JS Validations'),
--- a/lib/themes/dosomething/paraneue_dosomething/theme-settings.php
+++ b/lib/themes/dosomething/paraneue_dosomething/theme-settings.php
@@ -297,6 +297,8 @@ function _paraneue_dosomething_theme_settings_footer(&$form, $form_state) {
 
       $link_column = &$links[$country. '_links'][$prefix];
 
+      // Get new country code setting if it exists, if not fallback to old setting.
+      // @TODO: Remove this after we've updated theme settings everywhere.
       $heading_default = theme_get_setting('footer_links_' . $country . '_' . $column . '_column_heading');
       if(!$heading_default === NULL) {
         $heading_default = theme_get_setting('footer_links_' . $column . '_column_heading');
@@ -308,6 +310,8 @@ function _paraneue_dosomething_theme_settings_footer(&$form, $form_state) {
         '#default_value' => $heading_default,
       ];
 
+      // Get new country code setting if it exists, if not fallback to old setting.
+      // @TODO: Remove this after we've updated theme settings everywhere.
       $links_default = theme_get_setting('footer_links_' . $country . '_' . $column . '_column_links');
       if(!$links_default === NULL) {
         $links_default = theme_get_setting('footer_links_' . $column . '_column_links');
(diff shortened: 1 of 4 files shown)


angaither:
we might want to make this against the `global` branch, not the `dev` branch.
though, if no fastly headers are being sent in prod, it should be fine. not sure if we are doing fastly headers on production. 

to see this on Thor then, we can either
1. merge dev -> global
2. re-open this pull request against the global branch. 

cc @blisteringherb @sheyd

DFurnes:
Yeah, good catch! I totally forgot `global` branch was a thing. :cactus: This should definitely be safe to deploy to production (works happily on dev where headers are not being set), but I'm not sure if it's confusing to do so?

sheyd:
@DFurnes @angaither Fastly Headers are enabled in Prod at this point.

lib/themes/dosomething/paraneue_dosomething/theme-settings.php
```diff
@@ -272,46 +272,59 @@ function _paraneue_dosomething_theme_settings_footer(&$form, $form_state) {
...
+        '#collapsed' => TRUE,
+      ];
+
+      $link_column = &$links[$country. '_links'][$prefix];
+
+      // Get new country code setting if it exists, if not fallback to old setting.
+      // @TODO: Remove this after we've updated theme settings everywhere.
+      $heading_default = theme_get_setting('footer_links_' . $country . '_' . $column . '_column_heading');
+      if(!$heading_default === NULL) {
+        $heading_default = theme_get_setting('footer_links_' . $column . '_column_heading');
```
angaither, line 304: since these old theme variables are going to be defunct, it might be cleaner to have an update hook move the settings to use the `us` prefix country code.

DFurnes:
Awesome! So it should just use the "US" footer settings in that case, and we'll hold off on customizing other countries or "global" settings until Global launches? (Or if that's confusing, happy to merge into `global` instead. Just wary of the two branches diverging too much.)

lib/themes/dosomething/paraneue_dosomething/theme-settings.php
DFurnes, line 304: Oh, that's a good idea. I haven't made one of those before.

angaither:
As long it defaults to the US settings if nothing is set and we are :ok: with having that `add global footer links` step to a `GLOBAL DEPLOY CHECKLIST`™ (@fantini) then we should be ok to leave this in the `dev` branch.
but we should verify on stage (are fastly headers being sent on stage _...she wondered aloud directed at @sheyd_)

mikefantini:
@angaither I'm fine with adding global footer links being on this checklist.

and I believe that staging.dosomething.org has fastly headers set up but staging.beta.dosomething.org does not

angaither:
:8ball: let's leave this as is then and merge `dev` -> `global` today

DFurnes:
Cool, merging this in after talking things over with @angaither. I'll add an install hook to `dosomething_settings` to update the theme settings in a later PR.

PR #5458, opened 2015-10-13
Redirects users on homepage to proper path prefix

### LINKED ISSUE #5348: Geo-based redirect for /campaigns
Because this isn't a node, `/campaigns` won't get redirected with the same logic as #5183. Need to handle separately (with similar logic).

### LINKED ISSUE #5349: Home page is not redirecting anonymous users to correct country
- If I'm in the US,
- And I request the home page (e.g., `https://thor.dosomething.org/`),
- Then I should get redirected to the US home page.

itsjoekent:
#### What's this PR do?

Redirects users on the homepage to the proper path prefix (eg: dosomething.org/us, /mx, etc) if they're in one of the countries we support

dosomething_global_init() changes

Access the homepage as an MX user, verify you goto MX homepage (and rest of our supported languages)
Access the homepage as an French user, verify you goto global homepage (no redirect)

I think the comments covered everything that needs context. 

Fixes #5349 
Fixes #5348

This code effects _all_ users, not just anonymous ones like the title of the issue states. Raised this [here](https://dosomething.slack.com/archives/global/p1444758523000759) in Slack for further discussion

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -307,12 +307,8 @@ function dosomething_global_convert_country_to_language($country) {
  *  The URL prefix mapped to this language
  */
 function dosomething_global_get_prefix_for_language($language) {
-  $result = db_select('languages', 'l')
-    ->fields('l')
-    ->condition('language', $language, '=')
-    ->execute()
-    ->fetchAssoc();
-  return $result['prefix'];
+  $languages = language_list();
+  return $languages[$language]->prefix;
 }
 
 /**
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -16,16 +16,17 @@ define('DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE', 'en-global');
 function dosomething_global_init() {
   global $user;
 
-  // Redirect people hitting the homepage to there local version
-  if (drupal_is_front_page() && is_numeric(arg(1))) {
-
+  // Redirect people hitting the homepage or explore campaigns to there local version
+  $path = request_path();
+  if ((drupal_is_front_page() && is_numeric(arg(1))) || $path == 'campaigns') {
     // Cannot use arg() as it doesn't return path prefixes
-    $path = request_path();
     $args = explode('/', $path);
 
-    // Only apply logic to Global homepage.
-    if (!empty($path) && $args[0] != 'node') {
-      return;
+    // Only apply logic to Global homepage
+    if (drupal_is_front_page()) {
+      if (!empty($path) && $args[0] != 'node') {
+        return;
+      }
     }
 
     $country_code = dosomething_settings_get_geo_country_code();
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -16,7 +16,7 @@ define('DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE', 'en-global');
 function dosomething_global_init() {
   global $user;
 
-  // Redirect people hitting the homepage or explore campaigns to there local version
+  // Redirect people hitting the homepage or explore campaigns to their local version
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -269,6 +298,24 @@ function dosomething_global_convert_country_to_language($country) {
...
 /**
+ * Converts the given language into its assigned URL prefix.
+ *
+ * @param string language
+ *  The language to get the prefix for (eg: 'en', 'mx')
+ *
+ * @return
+ *  The URL prefix mapped to this language
+ */
+function dosomething_global_get_prefix_for_language($language) {
```
blisteringherb, line 310: Drupal has a function language_list() that returns all of the languages with their information. This query works, but I'm thinking that using the built-in function is more drupally, and it's a little less code for us to maintain. This is what language_list() returns by default, but it can return a list keyed by the field you pass in.

```
Array
(
    [en] => stdClass Object
        (
            [language] => en
            [name] => English
            [native] => English
            [direction] => 0
            [enabled] => 1
            [plurals] => 0
            [formula] => 
            [domain] => 
            [prefix] => us
            [weight] => 0
            [javascript] => 
        )

    [en-global] => stdClass Object
        (
            [language] => en-global
            [name] => English, Global
            [native] => English, Global
            [direction] => 0
            [enabled] => 1
            [plurals] => 0
            [formula] => 
            [domain] => 
            [prefix] => en-global
            [weight] => 0
            [javascript] => 
        )

    [pt-br] => stdClass Object
        (
            [language] => pt-br
            [name] => Portuguese, Brazil
            [native] => Português Brasileiro
            [direction] => 0
            [enabled] => 1
            [plurals] => 2
            [formula] => ($n!=1)
            [domain] => 
            [prefix] => br
            [weight] => 0
            [javascript] => 
        )

    [es-mx] => stdClass Object
        (
            [language] => es-mx
            [name] => Spanish, Mexico
            [native] => Español Mexicano
            [direction] => 0
            [enabled] => 1
            [plurals] => 0
            [formula] => 
            [domain] => 
            [prefix] => mx
            [weight] => 0
            [javascript] => 
        )

)
```

itsjoekent, line 310: oh thats better, for some reason i thought language list only returned a list of the languages, not all of the details. Ill switch this out!

itsjoekent:
@blisteringherb fixed, good to merge?

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -16,6 +16,35 @@ define('DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE', 'en-global');
 function dosomething_global_init() {
   global $user;
 
+  // Redirect people hitting the homepage to there local version
```
angaither: their\* local version (sp)

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -16,6 +16,35 @@ define('DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE', 'en-global');
...
+  if (drupal_is_front_page() && is_numeric(arg(1))) {
+
+    // Cannot use arg() as it doesn't return path prefixes
+    $path = request_path();
+    $args = explode('/', $path);
+
+    // Only apply logic to Global homepage.
+    if (!empty($path) && $args[0] != 'node') {
+      return;
+    }
```
angaither, line 30: this seems like some wild code to determine the home page, whyyyy? what's going on?

itsjoekent, line 30: unfortunately it had to be wild...

all of these are supporting edge cases / weird things I found. (eg: even if the URL is just / the arg 0 and 1 params still return the homepage node). The homepage is a weird thing, the comments point out which part handles what.

itsjoekent:
@blisteringherb I added support for /campaigns to fix #5348 as well, can you re-look it over real quick?

angaither:
:+1: for now 
but I think we should try to refactor some of this to make it a little more readable/maintainable

itsjoekent:
yes for sure

PR #5464, opened 2015-10-13
Removes regional admin access from misc. menu's 

### LINKED ISSUE #4901: br and mx admin roles shouldn't have access to configuration or users admin menu items
remove permissions from regional admin roles 
- [x] view user profiles
- [x] translate interface texts

itsjoekent:
#### What's this PR do?

Removes regional admin access from the following places https://github.com/DoSomething/phoenix/issues/4901#issuecomment-138592399

Test those configurations aren't accessible by regional admins but can still be accessed by normal admins

Fixes #4901

### DIFF
--- a/lib/modules/dosomething/dosomething_shipment/dosomething_shipment.module
+++ b/lib/modules/dosomething/dosomething_shipment/dosomething_shipment.module
@@ -40,7 +40,7 @@ function dosomething_shipment_menu() {
     'page callback' => 'drupal_get_form',
     'page arguments' => array('dosomething_shipment_admin_config_form'),
     'access callback' => 'user_access',
-    'access arguments' => array('administer modules'),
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_shipment/dosomething_shipment.module
```diff
@@ -40,7 +40,7 @@ function dosomething_shipment_menu() {
     'page callback' => 'drupal_get_form',
     'page arguments' => array('dosomething_shipment_admin_config_form'),
     'access callback' => 'user_access',
-    'access arguments' => array('administer modules'),
+    'access arguments' => array('edit campaign overrides'),
```
angaither: since this seems like a kind of arbitrary permission, let's add a comment to say why we are using this one.

PR #5465, opened 2015-10-13
fixes user country code, address, and update function

### LINKED ISSUE #5263: Thor: Registration doesn't pass country code to Northstar

chloealee:
#### What's this PR do?

Fixes to send both country code and address fields from phoenix -> ns.

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module lines 195-202 grabs correct country value and includes address values. 
lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module `dosomething_northstar_update_user` function - updates to hit `/users/:user_id` endpoint. For some reason, `/users` endpoint wasn't saving address properly.

Create new user on ds.org. Hit ns endpoint GET /users/email/<email> and make sure country code is returned. https://github.com/DoSomething/northstar/wiki/Spec#retrieving-a-user

Update user profile and add address. Hit the same endpoint above and make sure address is returned.

Fixes #5263

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -107,43 +107,17 @@ function _dosomething_northstar_build_http_client() {
 function dosomething_northstar_update_user($form_state) {
   $user = $form_state['user'];
   $id = $user->uid;
-
   $ns_user = dosomething_northstar_build_ns_user($user, $form_state);
 
   $northstar = new Northstar();
   $client = _dosomething_northstar_build_http_client();
 
-  // $northstar_user_info = dosomething_northstar_get_northstar_user($id);
-  // $northstar_user_info = json_decode($northstar_user_info->data, TRUE);
-  // $northstar_user_info = (object) $northstar_user_info['data'][0];
-  // $ns_id = $northstar_user_info->_id;
-
-  $response = drupal_http_request($client['base_url'] . '/users/drupal_id' . $id, array(
+  $response = drupal_http_request($client['base_url'] . '/users/drupal_id/' . $id, array(
     'headers' => $client['headers'],
     'method' => 'PUT',
     'data' => json_encode($ns_user),
     ));
 
-  if ($response->code === '200' && module_exists('stathat')) {
-    stathat_send_ez_count('drupal - Northstar - user updated - count', 1);
-  }
-  elseif ($response->code !== '200') {
-    watchdog('dosomething_northstar', 'User not updated : ' . $response->code, NULL, WATCHDOG_ERROR);
-  }
-
-}
-
-function dosomething_northstar_update_user($form_state) {
-  $user = $form_state['user'];
-  $ns_user = dosomething_northstar_build_ns_user($user, $form_state);
-
-  $northstar = new Northstar();
-  $client = _dosomething_northstar_build_http_client();
-  $response = drupal_http_request($client['base_url'] . '/users', array(
-    'headers' => $client['headers'],
-    'method' => 'POST',
-    'data' => json_encode($ns_user),
-    ));
   if ($response->code === '200' && module_exists('stathat')) {
     stathat_send_ez_count('drupal - Northstar - user updated - count', 1);
   }
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -106,15 +106,24 @@ function _dosomething_northstar_build_http_client() {
  */
 function dosomething_northstar_update_user($form_state) {
   $user = $form_state['user'];
+  $id = $user->uid;
+
   $ns_user = dosomething_northstar_build_ns_user($user, $form_state);
 
   $northstar = new Northstar();
   $client = _dosomething_northstar_build_http_client();
-  $response = drupal_http_request($client['base_url'] . '/users', array(
+
+  $northstar_user_info = dosomething_northstar_get_northstar_user($id);
+  $northstar_user_info = json_decode($northstar_user_info->data, true);
+  $northstar_user_info = (object) $northstar_user_info['data'][0];
+  $ns_id = $northstar_user_info->_id;
+
+  $response = drupal_http_request($client['base_url'] . '/users/' . $ns_id, array(
     'headers' => $client['headers'],
-    'method' => 'POST',
+    'method' => 'PUT',
     'data' => json_encode($ns_user),
     ));
+
   if ($response->code === '200' && module_exists('stathat')) {
     stathat_send_ez_count('drupal - Northstar - user updated - count', 1);
   }
@@ -183,8 +192,13 @@ function dosomething_northstar_build_ns_user($user, $form_state) {
   }
   foreach ($address as $ns_key => $drupal_key) {
     $field = $user->field_address[LANGUAGE_NONE][0];
-    if (isset($field[$drupal_key])) {
+
+    if ($drupal_key == 'country') {
       $ns_user[$ns_key] = $field[$drupal_key];
+    }
+
+    if (isset($field[$drupal_key]['value'])) {
+      $ns_user[$ns_key] = $field[$drupal_key]['value'];
     }
   }
 
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -104,34 +104,34 @@ function _dosomething_northstar_build_http_client() {
 /**
  * Send user profile updates to northstar.
  */
-// function dosomething_northstar_update_user($form_state) {
-//   $user = $form_state['user'];
-//   $id = $user->uid;
+function dosomething_northstar_update_user($form_state) {
+  $user = $form_state['user'];
+  $id = $user->uid;
 
-//   $ns_user = dosomething_northstar_build_ns_user($user, $form_state);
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -106,15 +106,24 @@ function _dosomething_northstar_build_http_client() {
...
   $user = $form_state['user'];
+  $id = $user->uid;
+
   $ns_user = dosomething_northstar_build_ns_user($user, $form_state);
 
   $northstar = new Northstar();
   $client = _dosomething_northstar_build_http_client();
-  $response = drupal_http_request($client['base_url'] . '/users', array(
+
+  $northstar_user_info = dosomething_northstar_get_northstar_user($id);
```
angaither: you're passing `$user` in above, not `$id` does that still work?

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -106,15 +106,24 @@ function _dosomething_northstar_build_http_client() {
...
+  $northstar_user_info = json_decode($northstar_user_info->data, true);
```
angaither: `TRUE` should be in caps

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -106,15 +106,24 @@ function _dosomething_northstar_build_http_client() {
...
+  $northstar_user_info = (object) $northstar_user_info['data'][0];
+  $ns_id = $northstar_user_info->_id;
+
+  $response = drupal_http_request($client['base_url'] . '/users/' . $ns_id, array(
     'headers' => $client['headers'],
-    'method' => 'POST',
+    'method' => 'PUT',
     'data' => json_encode($ns_user),
     ));
+
```
angaither, line 120: I can see this function being reallly slow in prod since we are making two requests in a row.. can we do this with one request? I thought the `user` POST endpoint would search to see if the user exists before creating a new one?

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
chloealee: yup! Double checked and tested again and it's working :)

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
chloealee: changed!

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
chloealee, line 120: Yes, totally agree. When testing using just `/users` POST, it didn't update the address. I'll take a look at the ns files to see if there is a bug there.

chloealee, line 120: @angaither I think [this](https://github.com/DoSomething/northstar/blob/dev/app/Http/Controllers/UserController.php#L54) might be preventing us from being able to use the `/users` POST endpoint. Would it be correct to add this to the block: 

```
        if ($found_user && password_verify($input['password'], $found_user->password)) {
            $user = $found_user;
        } else {
        // check to make sure user is logged in
        }
```

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -104,6 +104,35 @@ function _dosomething_northstar_build_http_client() {
...
+//     'headers' => $client['headers'],
+//     'method' => 'PUT',
+//     'data' => json_encode($ns_user),
+//     ));
+
+//   if ($response->code === '200' && module_exists('stathat')) {
+//     stathat_send_ez_count('drupal - Northstar - user updated - count', 1);
+//   }
+//   elseif ($response->code !== '200') {
+//     watchdog('dosomething_northstar', 'User not updated : ' . $response->code, NULL, WATCHDOG_ERROR);
```
angaither: hmm, I don't think we want to comment out this whole block?

chloealee: sorry - not finished working on this! I had a question above but it looks like it got hidden. Here it is below.

@angaither I think [this](https://github.com/DoSomething/northstar/blob/dev/app/Http/Controllers/UserController.php#L54)  might be preventing us from being able to use the /users POST endpoint. Would it be correct to add this to the block:

```
        if ($found_user && password_verify($input['password'], $found_user->password)) {
            $user = $found_user;
        } else {
        // check to make sure user is logged in
        }
```

angaither: oh, because we are logging the user in after the request now. I see

angaither: argh. hm, perhaps it is best to call the PUT with the user id then. 
...i wonder if there's a way we can call that without knowing the mongo id. that might be a better way to go, but @jonuy might have more feelings on this.

jonuy: agh. I still feel iffy about that login we have in northstar's `UserController::store`. Probably not something we can remove though since the apps have been built with the expectation of session info being there.

My vote I think is to support a PUT request with a drupal ID in addition to the mongo ID it already supports. Are we guaranteed that these users we're updating are already on Northstar?

chloealee: @angaither @jonuy These users aren't necessarily on northstar but I think it gets created simultaneously because to test, I created a new user on ds.org and then updated the address by editing my user profile.

jonuy: Another potential option is on the POST request we add a query param like `login_on_success=true` or something. And the login code will only run if that's set. It does mean changes on the app, but minor compared to other solutions.

angaither: I like the query param option!

chloealee: @jonuy by login code, do you mean this line? 
`$found_user && password_verify($input['password'], $found_user->password)`

jonuy: @chloealee Nah, I meant [this code](https://github.com/DoSomething/northstar/blob/dev/app/Http/Controllers/UserController.php#L100)

```
// Log the user in & attach their session token to response.
$token = $user->login();
$user->session_token = $token->key;
```

chloealee: oohh gotcha! I'm still a little confused though because it still requires the password to validate the user [here](https://github.com/DoSomething/northstar/blob/dev/app/Http/Controllers/UserController.php#L54).
When updating the user profile (where it gets the address from and sends to ns) the current password is not autofilled like email or first name are. Where does it get the password information?

jonuy: @chloealee Oh hm, if you don't have the password available, then I don't think the `POST` will work.

In which case you'll then need to send the update as a `PUT`. Thing is, we only support the `_id` param for `PUT` right now which it sounds like isn't available to you. We'd have to modify the endpoint so that you could use something like the drupal_id. Then we could give you something like...

`PUT /users/drupal_id/:drupal_id`
`PUT /users/_id/:mongo_id`

Does this make sense to do?

chloealee: thanks @jonuy ! Ah I see - so instead of making two calls to NS (1. to get the ns id and then 2. send the update via the `PUT` endpoint), just modify the endpoint to accept the drupal id so we only have to make one call?

jonuy: @chloealee Yea. Do you want to make that change in Northstar? Or at least create the issue for it? Then depending on workload, mabey either one of us could to take it on.

chloealee: Yes I will do both!

chloealee:
@angaither @jonuy updated the northstar endpoint so we only have to make one request to ns to update the user. Changes are now reflected and everything seems to work whenever you both get a chance!

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -183,9 +186,14 @@ function dosomething_northstar_build_ns_user($user, $form_state) {
   }
   foreach ($address as $ns_key => $drupal_key) {
     $field = $user->field_address[LANGUAGE_NONE][0];
-    if (isset($field[$drupal_key])) {
+
+    if ($drupal_key == 'country') {
       $ns_user[$ns_key] = $field[$drupal_key];
     }
+
+    if (isset($field[$drupal_key]['value'])) {
```
angaither: `else if`

angaither:
minor note, but otherwise :ok: :+1:

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
chloealee: updated!

PR #5468, opened 2015-10-13
Fix bug with nesting t().

### LINKED ISSUE #5455: Win module copy not replacing tokens
We are seeing this on the win module. The entity tokens are not being replaced correctly. 

screen shot 2015-10-13 at 1 37 31 pm

Possibly related to the translation work being done here and we need to figure out the correct way of doing the translation for that token replacement to go through.
https://github.com/DoSomething/phoenix/blob/c35f51e6e990773826a8ea83169136d4800b70e7/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc#L172

weerd:
Fixes #5455
#### What's this PR do?

If the t() function is nested inside the token_replace() function then it all kinda breaks and the tokens can't be replaced properly. This PR fixes that up; solution is to separate them out.

There is only one file... Choose wisely.

Meh. Let's just agree to test it on staging once it's up. 
#5455

---

@sbsmith86 

cc: @mikefantini


sbsmith86:
:dancer: :100:

PR #5478, opened 2015-10-14
Fix login and registration events

### LINKED ISSUE #5119: Turn Off Login Event Fire on Page Load
We released a code change around 21 July that has been automatically firing a “login” event on every page load, resulting in incorrect bounce rate stats.

image

It is likely due to: 

https://github.com/DoSomething/phoenix/commit/83d934114f5b9328637dae87e1d836a20b68a71b
and 
https://github.com/DoSomething/phoenix/pull/4705

and more specifically:
https://github.com/deadlybutter/dosomething/blob/3ffd7710fb56df776ed59a7853bda70ea1ff5e2b/lib/modules/dosomething/dosomething_user/dosomething_user.module#L235

The login form renders on every page, so this form_alter runs on every page. That’s why the event always fires.

DFurnes:
#### Changes

Fixes #5119. Since the `dosomething_user_form_alter` and `dosomething_user_preprocess_page` hooks are called on every page load (as we prepare that form on every page load), this event was constantly being triggered. This PR moves that logic in to the appropriate hooks for the user login and registration events.

I tested on my local box by adding a `dpm` call to [`dosomething_helpers_add_analytics_event`](https://github.com/DFurnes/phoenix/blob/d0627f7e9a2aebf64f877138987c59bab4998ee9/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module#L521-531). Everything seemed to function correctly when logging in or registering from either the header or the campaign pitch page.
#### Questions

Because `hook_user_login` is also called after registering a new user, registration will trigger both an `Authentication - Register` and an `Authentication - Login` event on the subsequent page load. I think this makes sense and keeps the implementation from getting too hairy... does anyone see any problems?

---

For review: @angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -471,7 +471,24 @@ function dosomething_user_update_user($form, &$form_state) {
  * Does actions after a new user has registered via the web.
  * Sign user up for emails/texts.
  */
-function dosomething_user_new_user($form, &$form_state) {
+function dosomething_user_new_user($form, &$form_state)
+{
+  // Trigger an analytics event on registration
+  // @NOTE: The 'Authentication - Login' event will *also* be triggered here,
+  // since registered users are subsequently logged in to their new accounts.
+  dosomething_helpers_add_analytics_event('Authentication', 'Register');
+
+  // Sign the user up for transactional messaging
+  _dosomething_user_send_to_message_broker();
+}
+
+/**
+ * Sign the authenticated user up for transactional messaging through Message Broker
+ * if the required modules are enabled and the user meets relevant criteria.
+ *
+ * @see dosomething_user_new_user
+ */
+function _dosomething_user_send_to_message_broker() {
   if (!module_exists('dosomething_mbp') || !module_exists('dosomething_signup')) {
     return;
   }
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -698,6 +699,11 @@ function dosomething_user_user_insert(&$edit, &$account, $category = NULL) {
     $edit['name'] = $account->uid;
     $account->name = $account->uid;
   }
+
+  // Trigger an analytics event on registration
+  // @NOTE: The 'Authentication - Login' event will *also* be triggered here,
+  // since registered users are subsequently logged in to their new accounts.
+  dosomething_helpers_add_analytics_event('Authentication', 'Register');
```
angaither: instead of doing this in `hook_insert()` maybe we can do it in the `new_user` submit handler it was in before. it's not a huge deal, but since it's not changing the user object, and it doesn't need to be in the insert hook. just a thought.

angaither: that one _should_ only fire from the `user/register` event

DFurnes: Hmm, I looked at the `new_user` method but it seemed to be mostly concerned with transactional logic. What's the benefit of doing things in that method over the `user_insert` hook?

DFurnes: Ah, it looks like `user_insert` is intended more for database changes? If so, that makes sense to move it. :+1:

DFurnes: Updated in bc092e8. :hammer:

angaither:
:+1: lovely, nice new function too!

Issue #5469, opened 2015-10-13
unpublishing a translation of a campaign doesn't actually unpublish it


mikefantini:
https://thor.dosomething.org/mx/node/36/translate shows that the MX SP version of the campaign is Not Published, however if you go to https://thor.dosomething.org/mx/voluntario/bailando-con-las-personas-mayores as an anonymous user you're able to see the campaign.

Possibly related to #5260? not sure, I _thought_ I saw this working recently.

angaither:
@deadlybutter can you take a look at this today?

itsjoekent:
sure

itsjoekent:
So the status in the translations object is actually being set to 1 when you publish a translation (and its 0 when not). That works fine.

I think the problem is our logic ignores whether a translation is published or not, and just works off whether it exists or not. I'll update the redirect logic to also make sure a translation not only exists, but it's been published.

mikefantini:
as we discussed IRL you're also able to directly hit the URL even if it's unpublished as an anonymous user which should not be the case.

i.e. if I'm anonymous and /mx/voluntario/whatever is unpublished, I should not be able to see the content when I go directly to that URL

PR #5479, opened 2015-10-14
Returns 404 when user access non published translation

itsjoekent:
#### What's this PR do?

Prevents users from accessing unpublished campaign translations. 

line 77 in the diff

Go to an unpublished campaign translation and make sure you get 404'd
Make sure translation redirects work for published campaigns

Yes.

I had to remove the call to `dosomething_global_is_translation_request()`, as it checks if the prefixes are present in the URL (And it works fine). The problem is it doesn't check if node actually has a published translation for that language. This meant a URL like dosomething.org:8888/mx/voluntario/closed-campaign was never redirected, even though in this case the spanish translation for that campaign isn't published. 

We could have duplicated the logic in is_translation_request from the redirect functions so we can fetch this data needed to make these determinations. But duplicated code is bad... :/

The temp solution for now was to just remove that function call and assume every URL needs to be double checked for both redirect AND translation publish status. If it fails the publish status, we return a 404 (confirmed that should be the functionality with Fantini) 

In the global 1.3.14159265359 we can re-write this logic to be less hacky (hence why I didnt fully remove the function from the code, it still works and is useful!)
#5469

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -529,12 +529,7 @@ function dosomething_global_redirect_node_language($node, $language_code, $redir
   $languages = language_list();
 
   // Can only redirect to an published, translated version.
-  if (!empty($node->translations->data[$language_code])) {
-    if (!$node->translations->data[$language_code]['status']) {
-      drupal_not_found();
-      drupal_exit();
-      return;
-    }
+  if (!empty($node->translations->data[$language_code]) && $node->translations->data[$language_code]['status']) {
     $target_node = $node->translations->data[$language_code];
 
     // Might be overly formal: We could pull 'node/nid' from the current
@@ -543,6 +538,9 @@ function dosomething_global_redirect_node_language($node, $language_code, $redir
     $target_path = sprintf('node/%s', $target_node['entity_id']);
 
     drupal_goto(url($target_path, array('language' => $languages[$language_code])), array(), $redirect_status);
+  } else {
+    drupal_not_found();
+    drupal_exit();
   }
 
 }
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -530,12 +526,15 @@ function dosomething_global_redirect_node_anon() {
  * @param int $redirect_status
  */
 function dosomething_global_redirect_node_language($node, $language_code, $redirect_status = 302) {
-
   $languages = language_list();
 
   // Can only redirect to an published, translated version.
-  if (!empty($node->translations->data[$language_code]) && (bool)$node->translations->data[$language_code]['status']) {
```
angaither, line 537: couldn't we also return a drupal not found from the bottom of this if statement? it was previously checking status

itsjoekent, line 537: hmm. I don't see any logical reason you couldn't, was basically just trying to breakup a long if statement.

angaither, line 537: It's easier to understand the exit point, and why a 404 is happening when it follow a path. 

```
if (has translation && it's published) {
 // go to that published translation
} 
else {
 // 404 the user
}
```

vs

```
if (has translation) {
 if (not published) {
 // 404 
}  
// go to translation
```

in the second case it's less clear why we are going to the translation

itsjoekent:
@angaither fixed the IF, good to merge?

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -544,7 +538,11 @@ function dosomething_global_redirect_node_language($node, $language_code, $redir
     $target_path = sprintf('node/%s', $target_node['entity_id']);
 
     drupal_goto(url($target_path, array('language' => $languages[$language_code])), array(), $redirect_status);
+  } else {
```
angaither: else on a new line!

PR #5480, opened 2015-10-14
Removing code that adds a prefix to signup email link

blisteringherb:
#### What does this PR do?

In production because we're not using prefixes, adding the prefix to the URL for signup emails is breaks things. When merging dev back in to global, which will remove the code necessary to make signup email links work, this code needs to be readded.

In an environment that's set up like production, e.g. where prefixes are disabled, and the default language is 'en', sign up for a campaign, and check the "get started" link to make sure it goes to the appropriate place.

Currently in production the link looks like `https://www.dosomething.org/us/node/?source=node/5840`. It should look like `https://www.dosomething.org/node/5840?source=node/5840`, or have the URL alias of the node.

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -562,7 +562,6 @@ function dosomething_signup_get_mbp_params($account, $node, $opt_in) {
   $url_options = array(
     'language' => $language,
     'absolute' => TRUE,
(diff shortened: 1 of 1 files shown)


angaither:
@blisteringherb can you make an issue to revert this back once it's not needed?
:+1:

blisteringherb:
@angaither Sure. As soon as dev is merged back into global, I'll make PR to merge the reverse changes into global.

Issue #5476, opened 2015-10-14
remove !necesario from fields in MX SP


mikefantini:
MX log in:

![image](https://cloud.githubusercontent.com/assets/6330971/10488208/9ca85a1a-7265-11e5-8f07-7e30a6f3df38.png)

MX account creation:

![image](https://cloud.githubusercontent.com/assets/6330971/10488226/b52546fc-7265-11e5-96bf-c3279526c12f.png)

also bizarre that birthday doesn't have it, because birthday is also required

also this doesn't appear to be happening for BR PT

weerd:
Created a new PR #5483 to remove the problematic items causing above issue from the POT file. But also need to clean up POEditor to remove from there as well. Thus the following checklist:
- [x] Remove offending string from POT file.
- [x] Remove offending string from POEditor.

weerd:
I updated Thor too. Once we push POT stuff to Prod it should be fine since the string has been removed from POEditor :)

PR #5483, opened 2015-10-14
Removing item from pot file that should not be translated.

weerd:
Refs #5476
#### What's this PR do?

This PR removes some unnecessary strings from a POT file. Putting these Drupal variables into pot files and having them translated was affecting the replacement of them in code for Droooopal.

There is only one file to rule them all. Take a guess.
#5476

---

@DFurnes 

cc: @mikefantini


DFurnes:
:+1: :+1:

PR #5491, opened 2015-10-15
dev -> global

angaither:
fixed the merge commits locally, and merged dev into global

@blisteringherb @deadlybutter can you do a spot check that [`dosomething_global.module`](https://github.com/angaither/dosomething/blob/global/lib/modules/dosomething/dosomething_global/dosomething_global.module) looks correct


itsjoekent:
I think its good to go

blisteringherb:
:+1:

PR #5492, opened 2015-10-15
Readding code to add prefixes to URLs in signups

blisteringherb:
This code was removed because adding prefixes in production breaks things at the moment. This PR readds it.


angaither:
:+1:

PR #5493, opened 2015-10-15
Add country to login/reg form action url

### LINKED ISSUE #5462: Login/reg form needs to prepend country in action URL
If I am an anonymous user on a non-Global English URL:

`/us/campaigns/out-closet-cards`

the login (and probably reg) form has `/campaigns/out-closet-cards` as the `action` param. This responds with a `301`, which then loses the `POST` info.

We could handle this via #5445 (i.e., don't redirect for any `POST`), but in this case we want to preserve the `/us` prefix so the user lands on the right location after submitting the form.

See _science diagram_ below.

login-form

sbsmith86:
#### What's this PR do?

Adds the country code to the url used as the form action for login and registration forms on campaign so when a user is on a campaign page and signs up, and logs in or creates an account, they will be redirected back to the correct campaign page.

Only one place to go `dosomething_user.module`

Fixes #5462

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -725,12 +725,18 @@ function _dosomething_user_add_signup_data(&$form) {
         '#access' => FALSE,
       );
     }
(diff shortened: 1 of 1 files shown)


angaither:
:+1:

PR #5514, opened 2015-10-16
Show country to staff

### LINKED ISSUE #5500: where the f is the country code?
admins should be able to see the country the user is in from the profile page

currently that field doesn't seem to be exposed

angaither:
expose the country field on the user/edit form to staff members
allow there to be more than the US selected, because #global 

fixes #5500


blisteringherb:
:+1:

PR #5515, opened 2015-10-16
Disable translatable from field_created_copy

### LINKED ISSUE #5420: Make field_created_copy not translatable on the campaign template
@julielorch can you please verify that none of the campaigns in the Ready to be Translated list on https://trello.com/b/HWVvUsxW/translation-pipeline were created by any members?

also @mnelson338 this would mean we couldn't translate any campaigns that were created by members. if this is going to be the case for the next few months then that's fine, we can always change our minds later. but i just don't want to go back and forth.

angaither:
Fixes #5420


blisteringherb:
:+1:

PR #5520, opened 2015-10-16
BR/MX contact us emails instead of zendesk form 

### LINKED ISSUE #5078: BR/MX users should see email address instead of contact us modal

sbsmith86:
#### What's this PR do?

When I first implemented this functionality, I had based the logic on the users country code. This PR updates that logic to use the country prefix in the url. This way, for example, if a use is in the US but visits a `/mx` url they will get the contact email for mx users. 

I made a helper function in `dosomething_global.module` that finds the current country prefix in the URL and then just updated `preprocess.inc` to use this new function instead of country code, as well as some sanity checks.

Fixes #5078

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
@@ -351,7 +351,8 @@ function paraneue_dosomething_add_info_bar(&$vars) {
       if (array_key_exists($country_prefix, $country_contact_emails)) {
         $info_bar_vars['contact_us_email'] = '<a href="mailto:' . $country_contact_emails[$country_prefix] .'">'. $country_contact_emails[$country_prefix] . '</a>';
       // All other countries get the zendesk form.
-      } else {
+      }
+      else {
         $zendesk_form_header = t(variable_get('dosomething_zendesk_form_header'));
         $info_bar_vars['zendesk_form_header'] = $zendesk_form_header;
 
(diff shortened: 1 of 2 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
```diff
@@ -338,23 +338,27 @@ function paraneue_dosomething_add_info_bar(&$vars) {
...
+        'mx' => 'ayuda@dosomething.org',
+      );
 
-    if (isset($vars['zendesk_form'])) {
-      $info_bar_vars['zendesk_form'] = $vars['zendesk_form'];
+      // For the countries in $country_contact_emails, use a contact email address instead of zendesk form.
+      if (array_key_exists($country_prefix, $country_contact_emails)) {
+        $info_bar_vars['contact_us_email'] = '<a href="mailto:' . $country_contact_emails[$country_prefix] .'">'. $country_contact_emails[$country_prefix] . '</a>';
+      // All other countries get the zendesk form.
+      } else {
```
angaither: else on a new line

angaither:
otherwise :+1:

lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
sbsmith86: ack...

PR #5521, opened 2015-10-16
Refactors ALL THE GLOBAL THINGS.

### LINKED ISSUE #5497: non-admins can't access any campaign pages because of double-country code in URL
As a user coming from Mexico, if I filter home page campaigns by "Environment" I see results which include the following:

image

When selecting the "Fish Bowl Baggies" campaign, I get a 404 page and the URL has duplicate country prefixes which is likely the cause for the 404:

`https://thor.dosomething.org/mx/mx/voluntario/fish-bowl-baggies`

image

### LINKED ISSUE #5506: /user links are pre-appended with /us
when I login from 
`https://thor.dosomething.org/mx` I am logged in and then get redirected to `https://thor.dosomething.org/us/user/3118670`
it should be
`https://thor.dosomething.org/mx/user/3118670`

@mshmsh5000 also noticed password redirect links are getting sent to `us/user/password-reset`

itsjoekent:
ALL OF THEM.
#### What's this PR do?
- Refactors the dosomething_global_init() to be closer to english than brainfuck
  - Gathers all variables needed at the start using helper functions
  - Breaks conditional logic out into new function and cuts duplicated, confusing or un-needed code
- Adds dosomething_global_is_path_prefix to check if a string is a path prefix
- Condenses the node redirect logic into one small function `dosomething_global_redirect_node` (thanks to the refactoring from above)

And yes, there are lots of comments.

init()

Where was that doc with all of the test cases again? We'll be needed that...

See my or andrea's brain.

TLDR: The old global init() was a mess from all of us adding in new URL edge cases and wasn't maintainable, causing multiple bugs.

Fixes #5506
Fixes #5497

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -52,7 +52,7 @@ function dosomething_global_init() {
 
   // Redirect people hitting the homepage
   // or explore campaigns to their local version.
-  if ($url_variables['is_homepage'] || $path == 'campaigns') {
+  if ($url_variables['is_homepage'] || $url_variables['path'] == 'campaigns') {
     // Redirect user to /country_code page.
     drupal_goto($url_variables['url_path_prefix'] . '/' . current_path());
     return;
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -14,77 +14,61 @@ define('DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE', 'en-global');
  * Implements hook_init().
  */
 function dosomething_global_init() {
-  global $user;
-
   // We don't need to redirect cli request, such as drush.
   if (drupal_is_cli()) {
     return;
   }
 
-  // Redirect people hitting the homepage or explore campaigns to their local version
-  $path = request_path();
-  if ((drupal_is_front_page() && is_numeric(arg(1))) || $path == 'campaigns') {
-    // Cannot use arg() as it doesn't return path prefixes
-    $args = explode('/', $path);
-
-    // Only apply logic to Global homepage
-    if (drupal_is_front_page()) {
-      if (!empty($path) && $args[0] != 'node') {
-        return;
-      }
-    }
-
-    $country_code = dosomething_settings_get_geo_country_code();
-    $language = dosomething_global_convert_country_to_language($country_code);
-    // No matching language, serve the global page
-    if ($language == NULL) {
-      return;
-    }
-
-    // Get users local path prefix and verify we support it
-    $path_prefix = dosomething_global_get_prefix_for_language($language);
-    if ($path_prefix == NULL) {
-      return;
-    }
-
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -93,36 +76,91 @@ function dosomething_global_init() {
...
-  }
+/**
+ * Extract relevant user variables from the current request
+ * to be used in dosomething_global_init().
+ *
+ * @return array
+ *      header_country_code - Country code specified in the headers
+ *      user_language - Matching language, or NULL for no matching installed language
+ *      user_path_prefix - The prefix which matches the users language
+ *      is_user_page - Is this the user profile page?
```
angaither: it would be helpful to know types of these vars

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -413,146 +461,76 @@ function dosomething_global_get_language($account, stdClass $node = NULL) {
...
-  $target_lang_code = DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE;
-
-  $lang_map = variable_get('dosomething_global_language_map', '');
-  foreach ($lang_map as $lang_key => $lang) {
-    if ($lang['country'] == $user_country_code) {
-      $target_lang_code = $lang_key;
-      break;
+  // Don't do redirects for privileged users.
+  if ($redirect_status == 302) {
+    if (user_access('access administration menu')) {
```
angaither, line 506: I don't see this in the previous version, this means anyone on our staff (admin, editor, site admin, etc) and regional admins won't get redirected here.. is that intended?

itsjoekent, line 506: https://github.com/DoSomething/phoenix/blob/global/lib/modules/dosomething/dosomething_global/dosomething_global.module#L457

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -14,77 +14,60 @@ define('DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE', 'en-global');
...
-
-      // Append the user language
-      $lang_append = $new_translation_prefix . $user_language;
-
-      // Build complete URL and redirect
-      $new_url = str_replace(' ', '', '/node/' . $nid . '/edit/' . $lang_append);
-      drupal_goto($new_url);
-    }
+  // Configure the correct path prefix
+  $url_variables['url_path_prefix'] = dosomething_global_get_prefix_for_language($user_variables['user_language']);
```
angaither, line 50: can this prefix stuff be refactored a bit to use the function @sbsmith86 just wrote? `dosomething_global_get_current_prefix()`

lib/modules/dosomething/dosomething_global/dosomething_global.module
angaither, line 506: ahhh, hidden way above :+1:

angaither:
:scream: let's try this :+1:

itsjoekent:
:boom:

PR #5529, opened 2015-10-16
Fixes redirect loop

### LINKED ISSUE #5528: Redirect loop when trying to go to thor homepage from Japan
image

itsjoekent:
# WHAT

Fixes a redirect loop that global users experience 
# HOW

Verifying there is an actual path prefix before performing the magic 
# FIXES #5528

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -46,8 +46,13 @@ function dosomething_global_init() {
     return;
   }
 
+  $correct_perfix = dosomething_global_get_prefix_for_language($user_variables['user_language']);
+  if ($correct_perfix == NULL) {
+    return;
+  }
+
   // Configure the correct path prefix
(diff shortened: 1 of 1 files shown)


itsjoekent:
DONT MERGE. :x: 

still have to fix the other comment

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -46,8 +46,13 @@ function dosomething_global_init() {
     return;
   }
 
+  $correct_perfix = dosomething_global_get_prefix_for_language($user_variables['user_language']);
+  if ($correct_perfix == NULL) {
```
sergiitk, line 50: I'd say either `if (!empty($correct_perfix))` or `if ($correct_perfix)`. `== NULL` smells like Jave too much :smile:

sergiitk:
:+1:, minor note.

angaither:
fixed with #5542

Issue #5504, opened 2015-10-15
Privacy Policy in login/register popup needs to be translated


weerd:
![image](https://cloud.githubusercontent.com/assets/105849/10525758/8eec6700-7354-11e5-9e53-5d4bb98cab0a.png)

Need to make sure "Privacy Policy" in the small greyed out text is translated.

---

cc: @mikefantini

mikefantini:
Hmmm @weerd I'm not seeing this in POEditor.

mikefantini:
Ah it appears as if it's because POEditor is pulling in "!terms". See https://poeditor.com/projects/po_edit?order=&filter=&id=39074&id_language=196&search=Creating+an+account.

weerd:
So just to clarify on @mikefantini point, Drupal uses `!terms` and `!privacy` to pull in and insert the links to the the respective pages so those tokens are needed. This all happens in `t()` so it should be translating. There was something up with the code which should be fixed now and will be confirming on Thor it works.

weerd:
Confirmed on Thor; both "Terms of Service" and "Privacy Policy" are translated.

PR #5535, opened 2015-10-19
Updating legal copy in create account.

weerd:
Refs #5504
#### What's this PR do?

While working to translate the term "Privacy Policy" a bug was noticed that the legal copy was missing the Terms of Service and the variables in the code were mis-labled. They are now corrected and the copy updated to read properly. The copy in POEditor will be updated shortly to match.

With the change in **register.inc**.
#5504

---

@angaither 

cc: @mikefantini


angaither:
:exclamation: :+1:

PR #5536, opened 2015-10-19
Clean up user login/registration hooks a bit.

DFurnes:
#### Changes

I was thinking about this after making changes in #5478. Login and registration forms both _shared_ the `dosomething_user_login_submit` hook, which was confusing at first glance. This shuffles things around to have a shared `authentication` submit handler, and a unique `login` handler for the login form.

Side effect: we can now cleanly have login-form specific actions, such as a `Authentication - Login` event that only fires when the user specifically uses the login form. Sweet. :candy: 

I tested to see that expected behavior still works (`dpm` every step of the way~), but would appreciate an extra pair of :eyes: just to make sure I didn't do something stupid. :grimacing: 

For review: @angaither


angaither:
brilliant :+1:

PR #5537, opened 2015-10-19
Fix undefined variable error.

DFurnes:
#### Changes

Fixes an undefined variable that was causing a `TypeError: undefined is not an object (evaluating 'this.$image.data')` error on window resize, by y'know using a _defined_ variable. Whoops.
#### How should I test this?

Go to the homepage and resize the window. You'll see the error in your Web Console. Then pull down this code and resize the window. You won't see the error. :bug: :hammer: 

For review: @weerd


weerd:
hokay! :+1:

Issue #5527, opened 2015-10-16
In POEditor but translation not appearing on thor


mikefantini:
- [x] First Name
- [ ] User account
- [ ] "(And you've entered to win the scholarship!)" - on alpha permalink when scholarship attached.

weerd:
Where specifically? Is this from the modal? Profile page?

weerd:
Found it. On the modal ;)

weerd:
Referencing these in following doc: https://docs.google.com/spreadsheets/d/1dFO6KGS5ryqI-x2EjPNDupwHqS6p9fK32oeHfTx8t0M/edit#gid=0

angaither:
closing in favor of #5517

Issue #5517, opened 2015-10-16
Need to ensure translation of a variety of miscellaneous strings.


weerd:
Strings to add to POEditor via POT files:
- [x] "Registration successful. You are now logged in." (seen when creating a new account)
- [x] "User account" (seen when viewing own user profile)
- [x] Months of the year
- [x] "by"
- [x] "Privacy Policy"
- [x] "My Account"
- [x] "Sign Up"
- [x] "Create new account"
- [x] "MM/DD/YYYY"
- [x] "Edit Status"

Refs #5513 
Refs #5352 
Refs: #5504

---

@mikefantini

weerd:
Updated on POEditor and Thor as well.

mikefantini:
@weerd what about MM/DD/YYYY (which should be DD/MM/AAAA)? It's in POEditor but not displaying https://poeditor.com/projects/po_edit?order=&filter=&id=39074&id_language=196&search=MM%2FDD%2FYYYY

![image](https://cloud.githubusercontent.com/assets/6330971/10551277/ff8765b2-7419-11e5-8a75-5ffa5061336c.png)

cc: @sbsmith86

sbsmith86:
Yea, i didn't run that placeholder text through `t()` :flushed: 

https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_user/dosomething_user.module#L18

weerd:
Ah, righty-o. I can update that!

weerd:
actually, @sbsmith86 it seems like you're doing some logic there with the ternary to decide on how to display date format, so `t()` probs _shouldn't_ be used right? I just need to update the `YYYY` to be `AAAA` 

año === year (in spanish, ¡olé!)

mikefantini:
@weerd can you please make sure it's DD/MM/AAAA for Brazilian users as well? Thanks!

sbsmith86:
@weerd I'm not all clear on all the details of how we are using `t()`... but I thought that all hard coded strings need to be ran through it. If the ternary needs to be removed for it to work, then feel free to rework that logic. We just need to make sure to show `MM/DD/YYY` placeholder text for US users only.

weerd:
[...]

PR #5543, opened 2015-10-19
Updating account creation module to translating missing strings.

weerd:
Refs #5527
Refs #5517
#### What's this PR do?

The First Name (and Last Name) fields did not have their labels translated, so that has now been added. Also the format for the date in the placeholder needed to be updated to be wrapped in  as well. 

De files boss?! De FILES!

I will be testing on Thor once this is merged to make sure it is working correctly.
#5527
#5517

---

@angaither 

cc: @sbsmith86 @mikefantini

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -15,7 +15,7 @@ include_once 'dosomething_user.theme.inc';
  */
 function dosomething_user_preprocess_page(&$vars) {
   $school_api_endpoint = variable_get('dosomething_user_school_api_endpoint', 'http://lofischools.herokuapp.com/search');
-  $date_format = t('MM/DD/YYYY');
+  $date_format = (dosomething_settings_get_geo_country_code() === 'US') ? 'MM/DD/YYYY' : 'DD/MM/YYYY';
   drupal_add_js(
     array('dosomethingUser' =>
       array(
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -15,7 +15,7 @@ include_once 'dosomething_user.theme.inc';
  */
 function dosomething_user_preprocess_page(&$vars) {
   $school_api_endpoint = variable_get('dosomething_user_school_api_endpoint', 'http://lofischools.herokuapp.com/search');
-  $date_format = (dosomething_settings_get_geo_country_code() === 'US') ? 'MM/DD/YYYY' : 'DD/MM/YYYY';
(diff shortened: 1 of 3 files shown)


angaither:
@sbsmith86 might want to take a look at this too since she was doing date switching validation

but first & last name look :+1:

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -15,7 +15,7 @@ include_once 'dosomething_user.theme.inc';
  */
 function dosomething_user_preprocess_page(&$vars) {
   $school_api_endpoint = variable_get('dosomething_user_school_api_endpoint', 'http://lofischools.herokuapp.com/search');
-  $date_format = (dosomething_settings_get_geo_country_code() === 'US') ? 'MM/DD/YYYY' : 'DD/MM/YYYY';
```
sbsmith86, line 18: Hmm why do we not need this logic anymore? will the POEditor know to translate `MM/DD/YYYY` to the correct string for all users except US?

DFurnes, line 18: Yeah, this is used to toggle [how client-side validation validates](https://github.com/DoSomething/phoenix/blob/1798036e74112e05af06cf67f71568e8e2707fde/lib/themes/dosomething/paraneue_dosomething/js/validators/auth.js#L87). Should probably keep as is!

weerd, line 18: oops. commented on wrong thing.

sbsmith86, line 18: The client side validation is happening [here](https://github.com/DoSomething/phoenix/blob/ecfc90e864f949bdb4ecce8bc9dc281428073e27/lib/modules/dosomething/dosomething_user/dosomething_user.module#L24), that is where we set the `dateFormat` setting for the JS to use. 

This code is just dealing with the placeholder text for the field, but I think I'm just confused because I was expecting it to look more like:

`$date_format = (dosomething_settings_get_geo_country_code() === 'US') ? 'MM/DD/YYYY' : t('DD/MM/YYYY')` 

Unless, what you are saying is that POEditor will see `MM/DD/YYYY` and change it accordingly?

weerd, line 18: Yeah, I think I was confused since you pointed me to that file location @sbsmith86 and thought I needed to change it there. So that actually deals with setting how the data validates, but not how the placeholder, etc is presented in the field. Gonna change it back, cause that's not what we want for that location. Thanks for the catch @sbsmith86 & @DFurnes

sbsmith86:
:+1: :dancer: :100:

PR #5544, opened 2015-10-19
Using the dosomething_global_url() function to build URLs on profile

### LINKED ISSUE #5423: Profile links missing country code prefix in URLs
link in You're Doing section of profile is missing country code prefix (see url that's loading at the bottom of the image when I hover over Prove It image):

image

blisteringherb:
#### What does this PR do?

This PR uses the dosomething_global_url() function to build the URLs for the "Prove It" button and the links to previously completed campaigns. Using dosomething_global_url adds the prefixes where appropriate.

Visit the profile page at `/user`, and make sure all links are appropriately prefixed including the "Prove It" link, and the links to previously completed campaigns.

@DFurnes There were comments in the code that the full URL is necessary and that's why the $base_url was included, so I wanted your input to make sure this looks OK. I've tested it locally with TorGuard and the links seem to be correct.

Fixes #5423

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
@@ -655,7 +655,7 @@ function paraneue_dosomething_preprocess_user_profile(&$vars) {
         'title' => $value['link'],
         'image' => $image,
         'description' => $value['call_to_action'],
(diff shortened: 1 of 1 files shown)


DFurnes:
@blisteringherb It should be fine! I think the problem was that we were getting URLs like `campaigns/teens-for-jeans` rather than `/campaigns/teens-for-jeans` but using `url()` should handle that properly.

DFurnes:
:+1: from me!

PR #5545, opened 2015-10-19
Fix issue with column overflow in Chrome 45.

### LINKED ISSUE #5409: highlight on finder straying from where it should be located when selecting Education
highlight on Education indicating that cause was selected in finder is straying.

I've noticed this on thor and now seeing it on staging.dosomething.org

using Chrome Version 45.0.2454.101 (64-bit)

image

DFurnes:
#### Changes

Fixes #5409. Resolves issue with columns overflowing in Chrome 45. (Get it together, Google!)
#### How should I test this?

Check it out in all sorts of browsers and screen sizes. Tested across a few screen sizes in Chrome 45, Safari 9, Firefox 41 and thing seem to be working well.
#### Screenshots

Before (in Chrome 45):
before

After (in Chrome 45):
after
## 

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_finder.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_finder.scss
@@ -29,7 +29,9 @@
     }
 
     li {
-      display: inline;
+      float: left;
+      clear: both;
+      display: inline-block;
       position: relative;
       margin: 0;
       break-inside: avoid;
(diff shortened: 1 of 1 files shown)


sbsmith86:
looks good to me :) what's the deal? Google no like `display:inline` anymore?

DFurnes:
We had some prior issues [with CSS columns in the Chrome 45](https://github.com/DoSomething/phoenix/issues/5044) that were fixed by switching things from `display: block` to `inline`. It turns out that causes some other problems (again, only in Chrome though... :weary:), but `inline-block` seems like the happy medium.

sbsmith86:
cool!

PR #5547, opened 2015-10-19
Remove default lang search from content search

### LINKED ISSUE #5539: remove default current user lang in content search
in search content view that is set to a default, remove it

angaither:
fixes #5539


weerd:
:+1:

PR #5548, opened 2015-10-19
Update to add additional translation capabilities.

weerd:
Refs #5517
#### What's this PR do?

Updating buttons and other copy to either wrap in  or add the terms to their respective POT files for inclusion into POEditor.

Just make sure code looks right; start at which ever file you fancy.

It will be tested once code is up on Thor.

Refs #5517

---

@angaither 

cc: @mikefantini


angaither:
👍

PR #5571, opened 2015-10-21
Node access hook for translated content

### LINKED ISSUE #5519: Anonymous users are able to access unpublished translations of nodes
The Brazilian Portuguese translation of School Insider (https://thor.dosomething.org/br/voluntario/escola-insider) is unpublished. However, as an anonymous user I'm able to go directly to that link.

Also, when I go directly to that link, the source language Title is pulled in.

See how on https://thor.dosomething.org/br/node/5356/translate, the BR PT translation is unpublished, and yet I'm able to view it as an anonymous user, and the US English title is pulled in (i.e. School Insider vs. Escola Insider.

image

image

This is why it appeared as if the homepage for various translations was pulling in the GL EN title, it was because those translations were unpublished.

Brazil:

image

US:

image

angaither:
deny users access to unpublished translated content

fixes #5519

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -85,6 +85,9 @@ function dosomething_global_init() {
   }
 }
 
+/**
+ * Implements hook_node_access.
+ */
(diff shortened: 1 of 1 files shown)


blisteringherb:
:+1:

Issue #5566, opened 2015-10-21
Can't create an account from /us/admin/people/create


mikefantini:
Need to go to global URL (/admin/people/create).

mikefantini:
To be clear, you don't get redirected if you go to a base admin url (i.e. I can navigate to www.dosomething.org/admin/people/create) just fine from the US. The weird behavior happens when I go to www.dosomething.org, I get taken to www.dosomething.org/us and then when I try to navigate to /admin/people/create the /us/ persists, which seems to be causing the problems

angaither:
let's remove the country prefix from admin urls, I think it was just caused by our init hook logic and it's not there for any real reason

angaither:
@sheyd and @blisteringherb `mod_rewrite` vs `outbound_alter` we should chat. 
also refs #5482

angaither:
@sheyd is going to set up fastly/varnish bypass rule for any url with `admin` in it

mikefantini:
@angaither @sheyd does this mean we can apply similar logic to resolve #5482?

sheyd:
@angaither Rules setup on thor and you were right. Re-tracing the steps, and we still get `/us/admin` path entry. There is a SUPER hacky way to fix this in Fastly/Varnish, but I think it's better to do this in app.

blisteringherb:
@mikefantini Do you remember what errors you were running into or any additional details about the admin user or what failed? I just tried navigating to `https://thor.dosomething.org/us/admin/people/create`, and was able to create an account. I suspect that if this is only showing up in production, introducing all of the language, path and redirect settings into production should fix this. Also the dosomething_global_init() function seems to have been greatly simplified so that may have also fixed the issue.

itsjoekent:
Yeah when I rewrote the init function it should have fixed all of the problems like this. I'll try on my local as well just for the hell of it, but I think this is :ok: if it worked on Thor

itsjoekent:
Nope not working on my local. Both on /admin and /us/admin

Nope just worked on http://dev.dosomething.org:8888/admin/people/create wtf

blisteringherb:
@deadlybutter It should work fine at `/admin/people/create`, but Fantini was saying it doesn't work at `/us/admin/people/create`. When you tried on your local the second time, was it just at `/admin/people/create` or was it also at `/us/admin/people/create`?

itsjoekent:
Does not work from /us/admin

itsjoekent:
[...]

PR #5573, opened 2015-10-21
Blocking redirect for non-translatable nodes

### LINKED ISSUE #5505: 11 Facts page are 404ing
Happens as anonymous user coming from both US or MX, but loads if signed in as a US admin:

https://thor.dosomething.org/facts/11-facts-about-human-trafficking

blisteringherb:
#### What does this PR do?
- Fact pages shouldn't be redirected to a prefixed URL. Because they're
  non-translatable, we can check for that and return before doing any
  redirection logic.

Visit a fact page as an MX, US or BR user and make sure that you can view the node at the bare URL without the prefix.

Fixes #5505

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -60,6 +60,11 @@ function dosomething_global_init() {
     return;
   }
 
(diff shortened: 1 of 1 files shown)


angaither:
:+1: 

I think we should probably do the same for admin pages (see #5566)

PR #5586, opened 2015-10-22
Updating code with t() and adding strings to POT files.

weerd:
Refs #5517
#### What's this PR do?

There were a few strings related to the image cropping tool that needed to be wrapped in t() to be translated, along with adding those strings and a few others to POT files for inclusion into POEditor.

Meh. I'll be testing on Thor once POEditor is updated.
#5517

---

@angaither 

cc: @namimody


angaither:
:+1:

PR #5598, opened 2015-10-22
Updating view more button text to allow translation.

weerd:
Refs #5517 
#### What's this PR do?

The 'View more' button in the Campaign gallery needs to be translated, but since it is generated via JavaScript, needed to use the custom Drupal JS object and Drupal.t() to translate it.

Meh. Will be testable once POEditor has term included and imported PO file into CMS.
#### What are the relevant tickets
#5517

---

@angaither 

CC: @namimody


angaither:
:+1:

PR #5561, opened 2015-10-20
Gallery heading

DFurnes:
#### Changes

Fall cleaning! :maple_leaf: This cleans up padding between gallery titles and their items, using the new [Gallery heading](http://forge.dosomething.org/#gallery) optional element. While we're at it, why not [use those handy theme functions instead of unique markup and render array](https://github.com/DoSomething/phoenix/commit/e60ff927434cf6be3c38b6524ef1a85c472d3b60). So clean! :sparkles:
#### How should I review?

Take a peek at static content. Does it look slightly better? Good. Does it look completely broken? Bad.
#### Show me some screenshots!

Before (ewww):
staging

After (yayyyy):
dev

---

For review: @DoSomething/front-end

PR #5605, opened 2015-10-23
Fix undefined function.

### LINKED ISSUE #5604: Fatal error, function `dosomething_static_content_preprocess_gallery_vars()` not found
See [`dosomething_action_guide_preprocess_node()`](https://github.com/DoSomething/phoenix/blob/5896b2823f770c6aa3a1e9cf46d474ef4ae60564/lib/modules/dosomething/dosomething_action_guide/dosomething_action_guide.module#L33).

pasted_image_at_2015_10_23_10_46_am_720

The bug introduced in #5561.

---

cc @namimody

DFurnes:
#### Changes

Fixes #5604 by updating function call that was renamed in #5561.

Ideally, this preprocess function should be moved to the theme since it's a bit weird for it to be living in the static content module but being referenced elsewhere. Will fix in a separate PR so we can unblock deploy.

Does it explode on an action guide? :boom: 

---

For review: @sergii-tkachenko 
/cc @mikefantini


sergiitk:
You missed `dosomething_campaign_group.module:107`

sergiitk:
:+1:

PR #5606, opened 2015-10-23
Fixes issue where a caption may not break to a new line.

### LINKED ISSUE #5589: Long RB Image Captions Do Not Wrap on iPhone
Note that 60 character limit _seems_ to be applying on the field, but text is then not wrapping when a user goes back to action page. Happening on at least global EN, US, and MX.

image

Ew.

DFurnes:
#### Changes

Fixes issue where a caption may not break to a new line if the caption includes a very long word, because the browser will not create a line break within a word unless `break-word` is specified.

Closes #5589.

Use [a word that is at least 40 characters long](https://en.wikipedia.org/wiki/Pneumonoultramicroscopicsilicovolcanoconiosis) as your caption. It should correctly break to two lines.

---

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_reportback.scss
@@ -209,6 +209,7 @@
   font-size: $font-smaller;
   line-height: 1.1;
   padding-top: ($base-spacing / 2);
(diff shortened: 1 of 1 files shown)


weerd:
Some of the strongest code I've ever seen. Solid :+1:

DFurnes:
:bow:

PR #5613, opened 2015-10-23
Tidy up global

DFurnes:
#### Changes

Tidy up some things in the `dosomething_global` module. :leaves: 

Also fixes three warnings: ensures that the index `country` exists before comparing it (f4c5282), explicitly define `$roles` before assigning to an index of the array (f262a56), and ensure that `$languages` is defined since we're using that object in scenarios where `$request_path[0]` is empty.

Ehhh. Most changes are within docblocks. Skimming the code should be enough. :)

---

For review: @angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -411,7 +411,7 @@ function dosomething_global_convert_language_to_country($language) {
 function dosomething_global_convert_country_to_language($country) {
   $lang_map = variable_get('dosomething_global_language_map', '');
   foreach ($lang_map as $lang_key => $lang) {
-    if ($lang['country'] == $country) {
+    if (isset($lang['country']) && $lang['country'] == $country) {
       return $lang_key;
     }
   }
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -203,7 +203,7 @@ function dosomething_global_node_insert($node) {
 /**
  * Implements hook_form_alter().
  */
-function dosomething_global_form_node_form_alter(&$form, &$form_state, $form_id) {
+function dosomething_global_form_node_form_alter(&$form, &$form_state) {
```
angaither: i think form id is needed https://api.drupal.org/api/drupal/modules%21system%21system.api.php/function/hook_form_alter/7

DFurnes: Hmm, I don't think it should make a difference (the variable is neither used nor returned), but happy to add it back in.

DFurnes: Looks like we still reference it in all our other `hook_form_alter`s, so I removed that commit. Strange, couldn't remember seeing that one before.

angaither:
:+1:

PR #5616, opened 2015-10-23
Adds !empty check before json_decode

chloealee:
#### What does this PR do?

Adds a check to see if ns user is null and if so, do not try to json_decode to prevent error:

```
if (!empty($northstar_user_info->error)) {
        $error = sprintf("Error fetching Northstar user data, uid=%d: '%s'", $id, $northstar_user_info->error);
        watchdog_exception('northstar', new Exception($error));
      } else {
        cache_set('northstar_user_info_' . $id, $northstar_user_info, 'cache_northstar_user_info', REQUEST_TIME + 60*60*24);
}
```

endpoint: `/api/v1/reportbacks.json?load_user=true` should return user info. and not break. Works on local - tested by hard coding `$northstar_user_info == NULL` before new check and it by passed this correctly, return null for all user info in the api response. How do we test if this works with staging without merging?

Right now, the request to ns to get user information remains the same as what's in this code, so we know that it's successfully getting user info. on staging.
#5408


lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
```diff
@@ -162,8 +162,11 @@ private function build($data, $full = false) {
...
       $northstar_user = dosomething_northstar_get_northstar_user($data->uid);
-      $northstar_user = json_decode($northstar_user, true);
-      $northstar_user = (object) $northstar_user['data'][0];
+
+      if (!empty($northstar_user)) {
+        $northstar_user = json_decode($northstar_user, true);
+        $northstar_user = (object) $northstar_user['data'][0];
+      }
 
       $this->user = [
```
weerd: Does this need some sort of logic too? So if `$northstar_user` ends up being `null`, how is the `id`, `first_name`, `last_name` etc being assigned that depends on getting an object from NS and not getting `null`?

weerd: Maybe use something like this:

```
$this->user['drupal_id'] = $data->uid;  // we by default pass the id to the new user property in the object

if(!empty($northstar_user)) {
  $this->user['id'] = $northstar_user->_id,
  $this->user['first_name'] = $northstar_user->first_name,
  $this->user['last_name'] = $northstar_user->last_name,
  $this->user['photo'] = $northstar_user->photo,
  $this->user['country'] = $northstar_user->country,
}
```

I think that could work...

weerd: Actually, scratch my suggestion above... it would make the returned data inconsistent and remove properties if `null` so wouldn't be as reliable... 

instead, maybe just wrap the calls when assigning the properties on the user array to use the `dosomething_helpers_isset()`:

```
...
  'first_name' => dosomething_helpers_isset($northstar_user->first_name),
...
```

chloealee:
@weerd I'm getting this error:

```
Notice: Undefined index: data in Reportback->build() (line 168 of /var/www/dev.dosomething.org/lib/modules/dosomething/dosomething_reportback/includes/Reportback.php).
```

which refers to this line: 

`$northstar_user = (object) @$northstar_user['data'][0];`

I'm not sure why it's saying that it's an undefined index and I looked it up on google - if I add an `isset()` in front of it or `null !==` like many things suggest, it no longer returns the correct data. I know it's probably not great practice to use the [`@ operator`](http://php.net/manual/en/language.operators.errorcontrol.php) but this is the only way I could get the notices to disappear in the logs. Is this the best solution to use?

chloealee:
wait - just checked to test again and looks like it isn't work now because of the rebase - hold off on checking this please and I'll keep you updated! Sorry! @weerd

chloealee:
@weerd okay now back in business! All should be working fine now. Just unsure about the comment above with the @ operator. Also, for some reason, `dosomething_helpers_isset` wasn't working correctly so changed it to a ternary operator. Let me know your thoughts!

chloealee:
@weerd for some reason, `dosomething_helpers_isset();` was causing the notices in the logs. I swapped it out for just a regular `isset();` check with a ternary operator and seemed to get rid of the notices! Now the logs are clear and correct data is returned!

weerd:
I looked into it, and it's how the `dosomething_helpers_isset()` is used. If it's an object you are passing through to it, you need to also specify the `$key` for it :)

chloealee:
@weerd tested `reportback-items` endpoint as well and everything is working with no errors in the logs!

weerd:
yaaaas!

lib/modules/dosomething/dosomething_reportback/includes/ReportbackItem.php
```diff
@@ -76,8 +76,9 @@ public static function find(array $filters = []) {
     }
 
     foreach($results as $item) {
-      $reportbackItem = new static;
-      $reportbackItem->build($item);
+      $load_user = dosomething_helpers_isset($filters['load_user'], NULL, FALSE);
+      $reportbackItem = new static(['ignore' => true]);
```
weerd: Didn't notice this before. No need for passing the `ignore` to the `new static()`. That's only needed for the `Reportback` class.

lib/modules/dosomething/dosomething_reportback/includes/ReportbackItem.php
```diff
@@ -90,8 +91,9 @@ public static function find(array $filters = []) {
    * Build out the instantiated Reportback Item class object with supplied data.
    *
    * @param object $data
+   * @param bool  $full Boolean to decide whether to fetch full user data.
    */
-  private function build($data) {
+  private function build($data, $full = false) {
```
weerd, line 96: don't forget to add

```
$northstar_user = (object) [];
```

chloealee:
@weerd thank you! Made those two changes and tested again - still works! :)

weerd:
Awesome-sauce! :dancer: 

Merge it! :+1:

chloealee:
:tada: :tada: :tada: BEST DAY OF MY LIFE!

Issue #5584, opened 2015-10-22
Clicking on Photo or "PROVE IT" on USer Profile Brings User to /campaigns


namimody:
As a non-admin user coming from Japan: if I click the photo or "PROVE IT buttons on these campaigns on my user profile: 

![image](https://cloud.githubusercontent.com/assets/12396076/10672689/a82d6654-78bf-11e5-92af-9ebce061003d.png)

I get brought to /campaigns when I should be brought to the action page or prove it section of the campaign
![image](https://cloud.githubusercontent.com/assets/12396076/10672695/b7ebe930-78bf-11e5-9f16-ae3b67888526.png)

angaither:
looks like that url is missing the country prefix

angaither:
@weerd do you mind taking a look at this?

weerd:
Seems like the `%23` is messing with the URL, which technically shouldn't since it's just the url encoded `#`. If I switch it to a `#` it goes to the right page.

PR #5618, opened 2015-10-23
Fixing url generated by Droops

weerd:
Refs #5584
#### What's this PR do?

Instead of adding the #prove to the path passed to the url() function, need to pass it within the options array as a 'fragment', so Droops can construct the url properly without encoding the # character.

I'll give you a sticker if you guess which file ;)
#5584

---

@angaither 

cc: @DFurnes


angaither:
PREPROCESS.INC

angaither:
CAN I HAVE A STICKER?

angaither:
:+1:

weerd:
You get a :star: sticker!

PR #5631, opened 2015-10-27
Action guide content kills signups.

### LINKED ISSUE #5585: Signing Up Via Permalink Page Experience Breaks
Happening from most browsers (Chrome, IE, Android, Safari/iOS) and especially via Japan (but problems reported from Brazil and US).

Users who try to sign up for a campaign from a permalink or pitch page are 404'd, brought to their "account information page", or page shows up in barebones HTML:
image

angaither:
This is caused because `LANGUAGE_NONE` is added to the array for non-admins.
The signup appears to fail because of a metadata wrapper error.
Fixes #5585

this [comment](https://github.com/DoSomething/phoenix/issues/5585#issuecomment-151245196) provides the most context


weerd:
:+1:

PR #5632, opened 2015-10-27
Country-specific "Who We Are" node links

### LINKED ISSUE #5629: Hacky header translation business!
Due to an issue with translation and nested field collections, we can't translate static content nodes. (Come on, Dries!) Instead, we're going to make a unique node for each country and link to the relevant node IDs for each country's header & footer settings.

Via @mshmsh5000 in Slack:

> Matt Holford [4:09 PM]
> Soooo if the header and footer are both manually manageable per country, then we should be able to create new static content nodes for any static content nodes we’d like to translate that have galleries.
> 
> Matt Holford [4:09 PM]
> We’ll lose any Drupal-side redirect business: Specifically, if I hit `/about/who-we-are` from Mexico, and we’ve created a separate node for the Mexican version of that page, I won’t get automatically redirected to the correct page.
> 
> Matt Holford [4:10 PM]
> For the cost of that, we’ll have the ability to enter all translated static content without having to fix the field collection translation bug that’s currently broken ​_for the entire world,_​ not just for us.

DFurnes:
#### Changes

Fixes #5629 by adding theme settings to set a unique node to link "Who We Are" to for each country. I've also added an update hook to automatically migrate the old settings to the new country-specific ones.
#### How should I manually test this?

Pull this branch and run the update hook. Everything should still be good. If you go to the admin panel, you should see fields to set a link for each country, pre-filled with the old universal link. If you change the link for a given country, you should see that node linked on pages with that country's prefix in the URL.
#### Screenshots

screen shot 2015-10-27 at 10 45 57 am

### DIFF
--- a/lib/modules/dosomething/dosomething_settings/dosomething_settings.install
+++ b/lib/modules/dosomething/dosomething_settings/dosomething_settings.install
@@ -37,3 +37,17 @@ function dosomething_settings_update_7001() {
   variable_set('dosomething_settings_copy_scholarships', $copy);
 
 }
+
+/**
+ * Update theme settings for Global 1.0.
+ */
+function dosomething_settings_update_7002() {
+  $paraneue_settings = variable_get('theme_paraneue_dosomething_settings', []);
+
+  $countries = dosomething_global_get_countries();
+  foreach($countries as $country) {
+    $paraneue_settings['header_who_we_are_link_' . $country] = $paraneue_settings['header_who_we_are_link'];
+  }
+
+  variable_set('theme_paraneue_dosomething_settings', $paraneue_settings);
+}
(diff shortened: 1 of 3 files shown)


DFurnes:
For the most part, this seems good. The autocomplete on these theme setting fields is broken though, which makes the admin experience _frustrating_. I checked the same field on staging and it doesn't seem to autocomplete there either, so perhaps this is already a known issue. The field type is `entity_autocomplete`.

I've also experienced intermittent issues with Drupal discarding all these settings, though I can't seem to reliably reproduce it yet. Definitely not ready to merge, but would be happy to get any :eyes:s on this early to catch any obvious issues!

angaither:
I was testing campaign group stuff and noticed autocomplete fields were borked locally as well.

mikefantini:
@DFurnes since we're able to get the field to reference the proper node by putting in the full copy (i.e. It's Your World [id:1674]) I think we can ship with this as is. Obviously not an ideal long-term solution, but it works. 

Do you mind opening (or finding) an issue for autocomplete on this field so that we fix that later?

DFurnes:
Alright, this seems safe to merge now! Rebased, and fixed a wee issue that was causing Drupal to drop all the settings (2b8dc59). Will make another issue about the entity autocomplete problem.

angaither:
:+1:

PR #5633, opened 2015-10-27
Merging latest dev changes into global 10.27

blisteringherb:
@angaither I did a merge instead of a rebase of dev into global, and here's the result. I confirmed with @sergii-tkachenko that the latest dev changes were the correct changes for /bin/ds-global which had most of the conflicts.


angaither:
:+1:

PR #5634, opened 2015-10-27
Refactors campaign template to not use entities 

### LINKED ISSUE #5556: Blank translated fields pull in from other translations
GTFO MX's (https://thor.dosomething.org/mx/voluntario/gtfo) problem fact, FAQs, and DO IT first step fields are blank. Logic seems to be that when non-English fields are blank, content from English translation (https://thor.dosomething.org/us/node/5459/edit/en) gets pulled in. Seems to occur for any field not filled out on Spanish or Portuguese translation.

See erroneous problem fact here:

image

itsjoekent:
#### What's this PR do?

Refactors out all ->value() use in the campaign preprocessing. 

Verify that MX fields are only being displayed on MX translation
Verify that if an MX translation does not have a field filled out, but EN does, the MX translation does not display anything

Also make sure the campaign template looks :ok: 

The value() function in entities doesn't care if a translation of a field does / doesn't exist. This was causing the bug in #5556 

Tried replacing the entity contructors to include lang codes but that didnt do anything. 

Instead I just replaced almost all uses of entity with a field getter function @weerd made a while back that accepts language parameter

Fixes #5556 

CC: @weerd

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -84,7 +84,7 @@
  *   - vips (string). HTML.
  *
  */
-function dosomething_campaign_load($node, $public = FALSE) {
+function dosomething_campaign_load($node, $public = FALSE, $language = DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE) {
   if ($node->type != 'campaign') { return FALSE; }
 
   // Set the image src ratio and styles.
@@ -99,7 +99,7 @@ function dosomething_campaign_load($node, $public = FALSE) {
   $campaign->title = $node->title;
   $campaign->status = dosomething_campaign_get_campaign_status($node);
   $campaign->type = dosomething_campaign_get_campaign_type($node);
-  $campaign->scholarship = $wrapper->field_scholarship_amount->value();
+  $campaign->scholarship = dosomething_helpers_extract_field_data($node->field_scholarship_amount, $language);
   if ($campaign->scholarship) {
     $campaign->scholarship = (int) $campaign->scholarship;
   }
@@ -112,11 +112,10 @@ function dosomething_campaign_load($node, $public = FALSE) {
 
   // Plain text properties.
   $plain_text = dosomething_campaign_get_property_names_plain_text();
-
   foreach ($plain_text as $property) {
     $field_name = 'field_' . $property;
     if (isset($wrapper->{$field_name})) {
-      $campaign->{$property} = $wrapper->{$field_name}->value();
+      $campaign->{$property} = dosomething_helpers_extract_field_data($node->{$field_name}, $language);
     }
   }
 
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
```diff
@@ -99,7 +99,7 @@ function dosomething_campaign_load($node, $public = FALSE) {
   $campaign->title = $node->title;
```
angaither, line 99: can we get rid of this line then `$wrapper = dosomething_helpers_node_metadata_wrapper($node);`

angaither:
@deadlybutter a lot of stuff uses this function, including the api, can you work with @weerd to ensure that isn't broken?

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
```diff
@@ -84,7 +84,7 @@
  *   - vips (string). HTML.
  *
  */
-function dosomething_campaign_load($node, $public = FALSE) {
+function dosomething_campaign_load($node, $public = FALSE, $language = DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE) {
```
angaither, line 87: does this default value actually work? it's not being included as a constant at the top of the file

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -120,6 +120,15 @@ function dosomething_helpers_extract_field_data($field, $language = LANGUAGE_NON
 
     $data = $field[$language];
 
+    if (!isset($data)) {
+      if (isset($field[LANGUAGE_NONE])) {
+        $data = $field[LANGUAGE_NONE];
+      }
+      else {
+        return;
```
angaither: do we really want to return with no values found, shouldn't this be an error?

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
itsjoekent, line 87: yep, adding this `dpm(DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE);` in dosomething_campaign.helpers outputs correctly

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
itsjoekent, line 99: the wrapper was still being used in a few places and I didn't wanna make more work than what was necessary for now

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
itsjoekent: I guess it could return an error, but the point of returning nothing was there is literally nothing to return.

if you specify a language and there is no content for it OR LANGUAGE_NONE, than there isnt anything to return thats relevant to what you requested.

actually, can we just make it return NULL?

weerd: definitely `return null;`... so at least you have something to test against in the function that called this.

angaither:
@deadlybutter can we close this pr, but leave the branch and re-visit this after #global

Issue #5636, opened 2015-10-27
Update date format translations in POEditor


sbsmith86:
This issue is related to #5576. In order to get the experience working correctly we need to update the POEditor translation to just translate `YYYY` to `AAAA`. We will handle switching out date formats on the code side based on country code.

weerd:
Should be all set in POEditor :) 

#5639 is just to add it to version control, so not dependent o having it merged to close this, huzzah!

PR #5637, opened 2015-10-27
Date field placeholder format based on country code

### LINKED ISSUE #5576: Birthday Date client-side validation not working for MX users on Thor
Before I fill out the form:
image

After I filled out the form:
image

cc: @weerd

sbsmith86:
#### What's this PR do?

We had moved translation of the placeholder text for the birthday field into POEditor, but the rules around that conflicted with the format being used for validation. So after much meeting...here comes the fix (i hope) in 3 lines of code. 

POEditor will just be responsible for translating the string `YYYY` to `AAAA` when it needs to. 
This code just updates the placeholder logic to ask for the correct date format based on the users country code which matches up with the validation logic.

The real sanity check for this will be to visit the site as a US user, but then hit a `/mx/` url and then try to register a new account. The format the form asks for would be `MM/DD/AAAA` and it should pass validation.

Fixes #5576 
Waiting on #5636 for the translation of `YYYY` to `AAAA` to be added. But it is not a blocker.

/cc @DFurnes @angaither

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/form.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/form.inc
@@ -298,10 +298,13 @@ function paraneue_dosomething_form_user_profile_after_build($form, &$form_state)
   $form['#weight'] = '-18';
   $form['#attributes']['class'] = array('auth-twocol');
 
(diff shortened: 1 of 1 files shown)


angaither:
:+1:

PR #5639, opened 2015-10-27
Udpate to date format translation

weerd:
Refs #5636
#### What's this PR do?

We have decided to change how the date format is translated so it only handles the portion of the date that actually needs to be translated which is the year format (YYYY).
#5636

---

@angaither 

cc: @sbsmith86


sbsmith86:
:100: :100:

angaither:
:8ball:

PR #5640, opened 2015-10-27
hide our secrets

### LINKED ISSUE #5638: revoke node edit access
image

This is not happening on production.

It might be because of this function `dosomething_global_node_access` that was added. 
refs #5571

angaither:
Don't return node access allow from the end of this function, yikes.
Fixes #5638


blisteringherb:
Ha. :+1:

PR #5641, opened 2015-10-27
Add action guide and notfound to staff only types.

### LINKED ISSUE #5627: Authenticated users should not be able to access action guides
Currently as an authenticated user I'm able to access an action guide by going directly to the node. 

Access to action guides should behavior the same way Images or Campaign Runs do, users can see them on the campaign, but not if they go directly to the image node.

image

angaither:
fixes #5627


blisteringherb:
:+1:

PR #5642, opened 2015-10-27
Fixes fatal error on campaign collections signup flow.

### LINKED ISSUE #5433: Child Campaign/Group Campaign Signup Error Message
Getting a "there was an error processing your request" error when signing up for a child campaign where on the campaign group we configure it to do an automatic parent signup

example staging.dosomething.org/node/1418/edit

image

image

@angaither we'll have to fix this before GGW goes live November 1st/ideally before we send it over to AARP October 17th.

sergiitk:
#### What's this PR do?
- Fixes fatal error on campaign collections signup flow.

Read #5433
1. Open http://dev.dosomething.org:8888/volunteer/grandparents-gone-wired
2. Signup for it
3. Make sure this error doesn't appear:
   image

Context: `dosomething_signup_get_mbp_params()` works both for translatable campaigns and not translatable campaign collections. The changes in #5153 ([this line](https://github.com/DoSomething/phoenix/pull/5153/files#diff-480d0c18d1c3b8dc40bff0004b3e6d12R569)) assumed that normal node `title` is always replaced by translatable `field_title`, which is just not the case for campaign collections.
So access to unknown `title_field` in wrapper caused this:
image

```
EntityMetadataWrapperException: Unknown data property title_field. in EntityStructureWrapper->getPropertyInfo() (line 335 of /var/www/dev.dosomething.org/html/profiles/dosomething/modules/contrib/entity/includes/entity.wrapper.inc).
```

Fixes #5433.

---

cc @blisteringherb @DeeZone

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -567,13 +567,21 @@ function dosomething_signup_get_mbp_params($account, $node, $opt_in) {
     ),
   );
 
+  // Get node title, normal for collections, translatable for campaigns.
+  $title = '';
+  if (isset($node->title_field)) {
+    $title = $wrapper->language($language)->title_field->value();
+  } else {
+    $title = $node->title;
+  }
+
   $params = array(
     'email' => $account->mail,
     'uid' => $account->uid,
     'first_name' => dosomething_user_get_field('field_first_name', $account),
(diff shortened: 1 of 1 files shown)


angaither:
:+1:  nice find
is this something we will want to abstract into a helper function eventually?

sergiitk:
@angaither Not sure. I think eventually we will have all content types translatable. I'll create an issue to think about it though.

PR #5643, opened 2015-10-27
Campaign collection links

### LINKED ISSUE #5474: incorrect URLs in campaign collections
child campaigns in campaign collections have the following URL structure https://staging.dosomething.org/volunteer/campaigns/hunt-day-1?source=node/1172 but they should have https://staging.dosomething.org/campaigns/hunt-day-1?source=node/1172 as an example (i.e. lose the /volunteer/ for the child campaigns).

example: https://staging.dosomething.org/volunteer/aaron-test-grouped-campaign

image

I believe this was introduced with the script that was written to fix #5341 

@angaither we resolved this issue on production by using optimizely to fix URLs on https://www.dosomething.org/volunteer/boosomethingorg. We'll be reviewing content for GGW with a sponsor next week so would be great to get this resolved by then.

DFurnes:
#### Changes

Fixes #5474. Campaign collections were missing the ol' `/` on their links, and so were trying to load campaigns relatively from the collection page. This was 404ing and making people sad. This PR runs those path aliases through the ol' `url()` so that they can be properly formatted. Hurrah!

First, check out [this commit](https://github.com/DoSomething/phoenix/commit/20ea0c7f4eef9e6b6822dc67e10a6dafe65bc847) where I add some PHP 5.4 hotness. Then wander over to [this commit](https://github.com/DoSomething/phoenix/commit/551d8fa5ccc73d25c23f1d81a4ff3efc51435c21) which fixes the actual issue at hand. :computer: :boom: :hammer: 

Go to a campaign collection. Click on a tile. Does it 404? No? Great.

---

For review: @angaither

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
@@ -224,10 +224,11 @@ function paraneue_dosomething_get_node_gallery_item($nid, $source = NULL) {
   $classes_array = [];
 
   $node = node_load($nid);
-  $link = drupal_get_path_alias('node/' . $nid);
+  $url_options = [];
   if ($source) {
-    $link .= '?source=' . $source;
+    $options['query']['source'] = $source;
   }
+  $link = url(drupal_get_path_alias('node/' . $nid), $url_options);
 
   $title = $node->title;
   if ($node->status == 0) {
--- a/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
@@ -221,7 +221,7 @@ function paraneue_dosomething_get_search_gallery($results) {
  *   The source query string to append to the link.
  */
 function paraneue_dosomething_get_node_gallery_item($nid, $source = NULL) {
-  $classes_array = array();
+  $classes_array = [];
 
   $node = node_load($nid);
   $link = drupal_get_path_alias('node/' . $nid);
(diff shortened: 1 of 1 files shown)


angaither:
VALUE ADDED :+1:

sergiitk:
I think we have a logic error here.

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -221,13 +221,14 @@ function paraneue_dosomething_get_search_gallery($results) {
...
 function paraneue_dosomething_get_node_gallery_item($nid, $source = NULL) {
-  $classes_array = array();
+  $classes_array = [];
 
   $node = node_load($nid);
-  $link = drupal_get_path_alias('node/' . $nid);
+  $url_options = [];
   if ($source) {
-    $link .= '?source=' . $source;
+    $options['query']['source'] = $source;
```
sergiitk, line 229: Please note that `$url_options` is not the same as `$options`.  `$options` is created on spot an never used. `$url_options`, on the other hand, stays empty.

sergiitk:
@angaither @DFurnes Please check my comment on the diff. I think we need to fix this.

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
DFurnes, line 229: Oh, good eyes! :eyes: Will fix.

DFurnes:
@sergii-tkachenko Good catch! Added a fix in #5649.

PR #5644, opened 2015-10-27
Add Portuguese and Spanish Static Content pathauto templates.

### LINKED ISSUE #5531: Need similar BR/MX path setups for /about/ as we have for /campaigns/
Once static content is translated, we need a similar solution for /about/ as we have for /campaigns/ (i.e. /voluntario is the alias translation).

@angaither I assume this can be 1.1 but not sure if it will mess things up for launch if we change it later.

sergiitk:
#### What's this PR do?
- Sets up Portuguese and Spanish language pathauto patterns for Static Content, as requested in #5531
  
  image

Fixes #5531.


angaither:
:+1:

PR #5646, opened 2015-10-27
Share links should not have a lang pre-appended

### LINKED ISSUE #5482: Facebook Problem Shares Not Creating Image
On Thor US: after clicking Facebook icon on a problem share, brought to this popup:
image

and after sharing the pop up modal shows a 404 and the following is posted to FB:
image

@sbsmith86 do you think this has to do with Facebook's caching logic? It's working for me on phoenix but not on Thor and TSG reported as a bug. Thanks!

angaither:
to get share links without a `us` prefix, the lang should be `en-global` with no prefix. 

Fixes #5482


sbsmith86:
:100: :dancer:

PR #5650, opened 2015-10-28
Adds language code if not present in URL

### LINKED ISSUE #5581: Translations get unpublished after each save from the edit tab
Translation is published:

image

I go to edit it as an editor and it's marked as "Not Published". Also the field to publish it when I first save it is disabled:

image

And so when I save it as an Editor, the US EN translation which _was_ published is now no longer published.

image

Then when I go to edit it again as an Editor, I am able to publish it:

image

And I can confirm that it is published:

image

But then as an Editor when I go to edit the content again, the translation is once again set to be unpublished and again the field to publish it is disabled.

image

And so it goes back to being unpublished:

image

And then the cycle repeats itself.

This explains why we are seeing so many campaigns on production have unpublished US EN translations cc: @namimody

itsjoekent:
#### What's this PR do?

Adds the source nodes language to the form alter logic when there is no lang specified in the URL.

Step 1. Take in Fantini's late night debugging issue https://github.com/DoSomething/phoenix/issues/5581
Step 3. Test if its not happening anymore. 
Step 4. Don't ask why I skipped step 2.

Fixes #5581

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -209,7 +209,8 @@ function dosomething_global_form_node_form_alter(&$form, &$form_state, $form_id)
     return;
   }
 
-  // Get the last item in the url, to get the edit/add node language.
+  // Get the last item in the url, and verify its the node edit language.
+  // If not set it to the source language
   $node_lang = array_pop(explode('/', (parse_url(current_path(), PHP_URL_PATH))));
   if ($node_lang == 'edit') {
     $node_lang = $form['#node']->language;
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -209,7 +209,6 @@ function dosomething_global_form_node_form_alter(&$form, &$form_state, $form_id)
     return;
   }
 
-
   // Get the last item in the url, to get the edit/add node language.
   $node_lang = array_pop(explode('/', (parse_url(current_path(), PHP_URL_PATH))));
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -208,8 +208,12 @@ function dosomething_global_form_node_form_alter(&$form, &$form_state, $form_id)
   if (!(in_array($form['type']['#value'], $translatable_types))) {
     return;
   }
+
   // Get the last item in the url, to get the edit/add node language.
   $node_lang = array_pop(explode('/', (parse_url(current_path(), PHP_URL_PATH))));
+  if ($node_lang == 'edit') {
```
angaither, line 215: the most pedantic comment, but since it isn't always going to be a language, can we call the var something else? 
`$url_path`?  
then set `$node_lang` to the node's language or the path?

itsjoekent, line 215: well we could but then we'd have to add an else statement under this new IF I created, feels like it's just making more code than we really need

is it confusing the way it is?

angaither, line 215: ternary FTW. it's not terribly confusing the way it is, it's up to you. 
will future joe (et al.) be confused by this code?

itsjoekent, line 215: I think if it's just this one edge case we'll be fine, I can see now why it might be confusing though. should i redo the commenting?

itsjoekent:
@angaither I rewrote the comment, let me know if it makes more sense now

angaither:
:+1:

PR #5660, opened 2015-10-28
Translated facts bug fix olé

### LINKED ISSUE #5572: Problem and Solution Facts Don't display correctly on frontend
As a an admin or a Brazilian admin, you can't change the problem or solution fact for the Spanish or Portugese translations of the campaign.

The backend will shows: 
image

Front end shows: 
image

Occurs even if I use TorGuard to come from Brazil OR try to use a fact that exists in prod fact database.

weerd:
Fixes #5572
#### What's this PR do?

To minimize the potential for breaking lots of campaign related code, took a more subtle approach to fixing the translation bug related to Facts. The `EntityDrupalWrapper` class was causing problems with grabbing the correct facts related to regional content (translated, but as separate unique nodes and not nodes with translations as part of the node object). 

Instead, created a separate function to handle grabbing that data from the node, without interrupting how the rest of the campaign node data is loaded when constructing the campaign object. It allows for grabbing content if there are translations and ones specific to the language defined by the url path (`/us/` or `/mx/` etc) or just grabbing the available content if no language specific options available, but there is definitely content in the field.

Probably in the **dosomething_campaign.helpers.inc** file at line 183 with the `@TODO` explanation and then follow the bread crumb trail of code... :bread: 

We'll test on Thor which has a bunch of language related content to make testing easier.

To help with mitigating bugs and not break how the `dosomething_campaign_load()` function loads and creates the `campaign` standard object, the aim with the custom function was to set it up in a way that it would return data back to the `dosomething_campaign_load()` function in the same format/structure that the `EntityDrupalWrapper` would have, because that's the structure the rest of the code is relying on. Deviating from that structure could end up causing more bugs that we'd need thorough testing to actually find and then iron out. So went with the path of least resistance; but also why some of the code may see a little odd in places, in the effort to accommodate the above intentions! :earth_americas: 
#5572

US specific content at `/us/node/1283`:
image

MX specific content at `/mx/node/1283`:
image

MX specific content, but none entered in the admin `/mx/node/1321`:
image

---

@angaither 

cc: @mikefantini @namimody

### DIFF
--- a/lib/modules/dosomething/dosomething_fact/dosomething_fact.module
+++ b/lib/modules/dosomething/dosomething_fact/dosomething_fact.module
@@ -109,15 +109,19 @@ function dosomething_fact_get_facts_data($node, $field_names) {
       $facts[$field_name]['nid'] = $entity_data['nid'];
       $fact_sources = dosomething_helpers_extract_field_data($entity->field_source_copy);
       $fact_sources = dosomething_fact_get_formatted_sources($fact_sources);
-      $facts[$field_name]['footnotes'] = '';
-      foreach ($fact_sources as $index => $source) {
-        $count = count($sources);
-        $facts[$field_name]['footnotes'][] = $count + 1;
-        $facts[$field_name]['sources'][$count] = $source;
-        $sources[] = $source;
-      }
 
-      $facts[$field_name]['footnotes'] = implode(' ', $facts[$field_name]['footnotes']);
+      if ($fact_sources) {
+        $facts[$field_name]['footnotes'] = '';
+
+        foreach ($fact_sources as $index => $source) {
+          $count = count($sources);
+          $facts[$field_name]['footnotes'][] = $count + 1;
+          $facts[$field_name]['sources'][$count] = $source;
+          $sources[] = $source;
+        }
+
+        $facts[$field_name]['footnotes'] = implode(' ', $facts[$field_name]['footnotes']);
+      }
     }
   }
 
@@ -134,10 +138,16 @@ function dosomething_fact_get_facts_data($node, $field_names) {
 }
 
 /**
+ * Get specific data from the fact entity.
+ *
  * @param $entity
  * @return array
  */
 function dosomething_fact_get_entity_data($entity) {
+  if (!$entity && !is_object($entity)) {
+    return NULL;
+  }
+
   $data = [];
 
   $data['title'] = $entity->title;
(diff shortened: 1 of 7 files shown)


lib/modules/dosomething/dosomething_fact/dosomething_fact.module
```diff
@@ -77,6 +77,110 @@ function dosomething_fact_get_mutiple_fact_field_vars($node, $field_names) {
...
+ *
+ * @param object $node The loaded node to collect field data from.
+ * @param array $field_names Array of field machine names to collect data from.
+ * @return array|null
+ */
+function dosomething_fact_get_facts_data($node, $field_names) {
+  $entity = NULL;
+  $facts = [];
+  $sources = [];
+  $language = dosomething_global_convert_country_to_language(dosomething_global_get_current_prefix());
```
angaither, line 90: does `global $language_content` work?
https://api.drupal.org/api/drupal/developer%21globals.php/global/language_content/7

angaither, line 90: oh, also check out the helper function that @sergii-tkachenko just wrote in #5661

weerd, line 90: Is `global $language_content` based on just the path prefix or does it take into account user preference, headers, etc?

I can take a look and see what it outputs in various url path scenarios...

lib/modules/dosomething/dosomething_fact/dosomething_fact.module
```diff
@@ -77,6 +77,110 @@ function dosomething_fact_get_mutiple_fact_field_vars($node, $field_names) {
...
+        $entity = $field_data[0]['entity'];
+      }
+    }
+
+    if ($entity) {
+      $entity_data = dosomething_fact_get_entity_data($entity);
+
+      $facts[$field_name]['fact'] = $entity_data['title'];
+      $facts[$field_name]['nid'] = $entity_data['nid'];
+      $fact_sources = dosomething_helpers_extract_field_data($entity->field_source_copy);
```
angaither, line 110: also...this function can take a lang argument...does that simplify things?

angaither:
just some minor comments, but not needed for merge :+1: 
olé 

:tada: :dancer:

lib/modules/dosomething/dosomething_fact/dosomething_fact.module
weerd, line 110: Yeah, so at the point where this is called `dosomething_helpers_extract_field_data()` in that segment of code, the lang assigned to the data is `und` so doesn't really play a role here. This is for the fact sources which are assigned to the Fact entity and would potentially be translated, but wouldn't end up having a language assigned to them... just the parent Fact entity.

PR #5661, opened 2015-10-28
Improves translatable fields language detection.

### LINKED ISSUE #5647: Issues with Campaign Collections on Staging
- Edited https://staging.dosomething.org/volunteer/grandparents-gone-wired
- Cleared cache
- Received this error

image

I did the above steps because I was trying to figure out why so little of the content for the collection is showing up:

image

Even though way more fields are filled out:

image

image

Assigned to @angaither for now, but @DFurnes @sergii-tkachenko not sure if either of you know what's going on?

### LINKED ISSUE #5657: Turn off 'User' content language detection strategy on dev branch?
Do we need `User` Content language detection on US now?
image

We have 1330 users (including 20 from the US) that have other than US English language set in their profile (most of those are set to `en-global`). Some pages that are not supposed to be translatable look wrong for them because Drupal tries to display content fields in language of their preference, but it does not exists for this content, see #5647.
I don't know if we're using `$user->language` for non-admin users at all.

If we really don't need it, I'd suggest to turn it off. Thor lunch will do this anyway:
image

Before launching Thor it makes total sense to fallback all content to English.

---

@mshmsh5000 @angaither @blisteringherb @mikefantini

sergiitk:
#### What's this PR do?
- Turns off `User` and `Interface` content language detection methods:
  image
- Prefers content language over site language in translatable field language detection
- Change your user's language to en-global
- Open http://dev.dosomething.org:8888/volunteer/grandparents-gone-wired
- Note that campaigns aren't there
- Switch to this branch's code
- `drush fra`
- Open http://dev.dosomething.org:8888/volunteer/grandparents-gone-wired
- Make sure the campaigns are back
- Open http://dev.dosomething.org:8888/admin/config/regional/language/configure
- Make sure **Content language detection** is the same as on the picture above

Reasoning provided in #5647.

Closes #5647, #5657.

---

:warning: This PR will cause merge conflict of `dosomething_global.strongarm.inc` with global branch.
:warning: Please solve it in favor of global's code.

---

cc @blisteringherb @mshmsh5000

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -576,8 +576,8 @@ function dosomething_helpers_node_metadata_wrapper($node, $ignore_language = FAL
  *   ISO-code of current language
  */
 function dosomething_helpers_get_current_language_code() {
-  global $language;
-  return $language->language;
+  global $language_content;
+  return $language_content->language;
 }
 
 /**
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -576,6 +576,17 @@ function dosomething_helpers_node_metadata_wrapper($node, $ignore_language = FAL
  *   ISO-code of current language
  */
 function dosomething_helpers_get_current_language_code() {
+  global $language;
+  return $language->language;
+}
+
+/**
+ * Returns current language_content code.
+ *
+ * @return string
+ *   ISO-code of current language
+ */
+function dosomething_helpers_get_current_language_content_code() {
   global $language_content;
   return $language_content->language;
 }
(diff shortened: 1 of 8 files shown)


lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -576,8 +576,8 @@ function dosomething_helpers_node_metadata_wrapper($node, $ignore_language = FAL
  *   ISO-code of current language
  */
 function dosomething_helpers_get_current_language_code() {
-  global $language;
-  return $language->language;
+  global $language_content;
```
angaither, line 590: does this have unexpected side effects if we always rely on the language content?

angaither:
just the one :question: but otherwise looks :ok:

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
sergiitk, line 590: This should be right since we're detecting language for field content. On the other hand, somebody could use this helper for other purposes. So you are right. I'll move this to its own function.

angaither, line 590: it doesn't look like it's used for anything but content, but maybe rename to `get_current_node_lang_code`

sergiitk:
@angaither Please check out the updated code.

angaither:
:+1:

Issue #5445, opened 2015-10-09
Skip all country-based redirect logic for POSTs


mshmsh5000:
@sergii-tkachenko suggested in analyzing my fix for #5441 that we shouldn't be redirecting any POST requests at all. Which makes sense, and will simplify the logic in `dosomething_global_init()`.

PR #5662, opened 2015-10-29
Prevents POSTS from being redirected in global init

itsjoekent:
## WHAT

Adds check at top of global init() to make sure this ain't a POST request
Also removes the duplicate logic from one of the edge cases
## WHY

Cause POST requests don't need to be redirected 
## ARE YOU SURE?

YEAH! I tested a few post requests and they worked fine. 
## OK. WHAT DOES IT FIX?

NUMBA #5445

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -31,6 +31,11 @@ function dosomething_global_init() {
   $user_variables = dosomething_global_get_user_variables();
   $node_variables = dosomething_global_get_node_variables();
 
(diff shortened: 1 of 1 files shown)


angaither:
if you're **so so so** sure this doesn't break anything :+1: and merge

itsjoekent:
oh you have no idea how so so so sure I am (everything :boom: )

itsjoekent:
craperdoodles i didnt merge this before the freeze. the phone has failed me twice today dammit. 

is this gonna be a problem now that we're ditching global?

angaither:
@deadlybutter yeah, this can wait till after, it'll have to be merged into dev

PR #5678, opened 2015-11-02
I feel dumb.

### LINKED ISSUE #5668: field_fact_problem displays twice if field_fact_solution doesn't exist on a translation
Possibly related to #5572 

Currently if `field_fact_solution` doesn't exist on a translation, `field_fact_problem` is being output underneath The Solution instead of `field_solution_copy`

image

image

https://thor.dosomething.org/mx/voluntario/limpieza-de-calles-por-la-tortuga-marina

The logic should be:
- Display `field_fact_problem` underneath The Problem
- Display either `field_fact_solution` _or_ `field_solution_copy` and then `field_solution_support` underneath The Solution

weerd:
Fixes #5668
#### What's this PR do?

The  needed to be set inside the foreach loop so that it would be reset each time it looped for each fact (both problem and solution facts). With it being set outside the loop, if the fact problem existed the  would be set for that fact, and then on the second run, if the solution fact was missing, it would keep the  set as the one for the problem fact. Shiza.

One file. Compare the red highlight to the green one ;)
#5668

---

@angaither 

cc: @mikefantini

### DIFF
--- a/lib/modules/dosomething/dosomething_fact/dosomething_fact.module
+++ b/lib/modules/dosomething/dosomething_fact/dosomething_fact.module
@@ -84,12 +84,13 @@ function dosomething_fact_get_mutiple_fact_field_vars($node, $field_names) {
  * @return array|null
  */
 function dosomething_fact_get_facts_data($node, $field_names) {
-  $entity = NULL;
   $facts = [];
   $sources = [];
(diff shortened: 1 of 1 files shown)


angaither:
:+1:

PR #5686, opened 2015-11-02
Update the redirect logic for translated nodes.

### LINKED ISSUE #5677: US users are not redirected to US URL if there is no GL EN translation
Level Playing Field:
- node is published
- US EN translation is published
- GL EN translation does not exist

As an anonymous user coming from the US, if I go to https://thor.dosomething.org/campaigns/level-playing-field, I'm _not_ being redirected to the US experience. The desired behavior is that I would be getting redirected.

Some tangible side effects to this are that
1. Problem fact is missing on the pitch page
2. I'm seeing the GL EN footer

image

image

Since we have a lot of marketing out there pointing people directly to the base URL (i.e. www.dosomething.org/campaigns/level-playing-field) we need to resolve this before we launch.

angaither:
Fixes #5677
This was failing because the node had one translation in the array, the source lang node.
There was no redirect logic being applied, even though it should have been.
This now only exits the code if the node has no translations.
### How to test
- find a campaign with only one translation (local example [thumb wars](http://dev.dosomething.org:8888/campaigns/thumb-wars))
- go directly with no /us in the url
- does it redirect to /us? 
### Additional context

should also not perform redirect logic on fact pages or things without translations still

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -78,8 +78,8 @@ function dosomething_global_init() {
     return;
   }
 
-  // If the node isn't translatable, don't do any redirection.
-  if (count($node_variables['node']->translations->data) <= 1) {
+  // If the node doesn't have translations, don't do any redirection.
+  if (empty($node_variables['node']->translations->data)) {
     return;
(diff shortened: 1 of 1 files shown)


blisteringherb:
:+1:

PR #5688, opened 2015-11-02
404 users if they try to access and unpublished translation.

### LINKED ISSUE #5675: Anonymous users are able to access unpublished translations of nodes
Related to #5519: 

The Brazilian translation of node 5475 is unpublished. However as an anonymous user I'm still able to access https://thor.dosomething.org/br/voluntario/citação-de-um-campeão?source=node/1141

image

@angaither I think this is a blocker with the new strategy of launching with unpublished translations.

angaither:
Fixes #5675


blisteringherb:
:+1:

PR #5689, opened 2015-11-02
Quick fix for login/logout logic

### LINKED ISSUE #5683: Logging in on a non-campaign page from the US takes user to /user/ not /us/user/
As a user located in the US, if I log in on any non-campaign pages (i.e. the homepage), I get taken from www.dosomething.org/us to www.dosomething.org/user/[uid].

Because it doesn't have /us/ in the URL, it's not pulling in content properly (i.e. campaign title, image, CTA for campaigns I'm signed up for).

I believe we want the user go to /us/user/[uid]

image

itsjoekent:
#### What's this PR do?

Fixes login/logout logic / general init() logic stuff

Can you login / logout? 

THIS BREAKS THE PURE /LOGIN URL (AGAIN, no path prefix). 

Unfortunately there is no easy way to path prefix that URL because its the return of a POST request and it behaves funky. It's better (I think) to merge what I have here and fix log in / log out and user profiles and not block deployment over this. I can keep working at it if you want though

Fixes #5683 

CC: @mikefantini thoughts on the context ^?

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -32,7 +32,7 @@ function dosomething_global_init() {
   $node_variables = dosomething_global_get_node_variables();
 
   // Never redirect POST requests
-  if ($url_variables['request_method'] == 'post') {
+  if ($url_variables['request_method'] == 'post' && !$user_variables['is_login_page']) {
     return;
   }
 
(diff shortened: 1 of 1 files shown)


mikefantini:
@deadlybutter @angaither personally I don't think this issue is a blocker. I added it to 1.0 to prioritize it over some of the 1.1 issues, but this issue itself should not block deployment IMO.

itsjoekent:
I'll leave it open then, I'll prob need to redo part of it anyway so it doesn't break the /login

angaither:
@deadlybutter can we close this pr?

PR #5696, opened 2015-11-03
Fact problem share bug olé

### LINKED ISSUE #5684: Problem Statement Not Appearing in Permalink View
image

Should appear like this (although this screenshot is missing a CTA above sign up)
image

(Problem statement appears on pitch page for anon users)

weerd:
Fixes #5684
#### What's this PR do?

As it turns out not all stages in Drupal request/page data load cycle are created equal. During the loading of the node for the reportback share, when the node is loaded for the campaign, not all the nested nodes (ie: facts) are loaded as well; instead those nodes are loaded with just a target_id indicator. So sometimes we need to roll up our sleeves and load the node ourselves!

We can test it on Thor to make sure content is showing up, and relevant based on regional prefixes (and if content is available).
#5684

---

@angaither 

CC: @mikefantini


angaither:
:+1:

PR #5701, opened 2015-11-03
Fixes static content images.

sergiitk:
So when anonymous user accesses the US static content page it can't see the images on it. Admin can.
Hardcode debugging shows that [here](https://github.com/DoSomething/phoenix/blob/1a0e3efc77ba7c43fdebc6904ad0324593340868/lib/modules/dosomething/dosomething_static_content/dosomething_static_content.module#L143) `field_gallery_image` just doesn't exist for normal users and anonymous. Have NO IDEA why.

Luckily, the duplicate of that info was in `#entity` object. I just took it from there.

image
## 

cc @mikefantini @sheyd @mshmsh5000


weerd:
:+1:

sergiitk:
Fixes gallery for #5702.

mshmsh5000:
Fucking champion.

PR #5713, opened 2015-11-05
Fixes hotmodule time left

### LINKED ISSUE #5582: Hot Module Shows Date 1 Day Before Deadline
Backend shows:
image

Front end shows:
image

Note that the header photo and the hot module show different dates when the "show end date on pitch page and action page" box is checked. The header photo shows the correct date, and the hot module shows the day before. Therefore, this is not a problem on campaigns without the hot module.

Solve until this is fixed: set deadline on the backend to one day later than the actual deadline, don't show date in header photo.

itsjoekent:
#### What's this PR do?

Fixes the "days left" in hotmodule

Does the hotmodule display correct days left

It has something to do with user languages, still can't find the exact cause _why_ that matters. For now though this fixes the problem.

Fixes #5582 

screen shot 2015-11-05 at 10 23 28 am

Note on Git history: "Quick fix for login/logout" was an accidental commit I made on my dev branch which I forgot about, which meant when I check'd out new branches it was there. Because of this I didn't think cherry pick would work and I decided to just revert the commit. Sorry :/

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -627,12 +627,13 @@ function dosomething_campaign_get_scholarship($nid) {
 function dosomething_campaign_get_hot_time_left($node) {
   $today = dosomething_campaign_set_est_timezone(new DateTime());
   $end_date = dosomething_campaign_set_est_timezone(new DateTime($node->field_high_season[LANGUAGE_NONE][0]['value2']), TRUE, TRUE);
+  $end_date_no_formatting = new DateTime($node->field_high_season[LANGUAGE_NONE][0]['value2']);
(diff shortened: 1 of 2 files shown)


angaither:
@deadlybutter can you update the PR on why you have to remove the converting to EST if there's less than 8 days left in the campaign? seems like this logic is getting tough to follow

itsjoekent:
code says if days is greater than 8, not less!

As for _why_ this works, when you call the set est timezone function, it actually sets the end time to be the day before it ends, at 23 hours and 59 seconds (this was intentional, theres a pr or issue somewhere in the past describing this, basically it had to deal with calculating time left). 

Using the raw value in the node solves our problem though if we're just getting days left.

Make sense?

angaither:
...okay, seems a little wild though
:+1:

itsjoekent:
It is wild, but i dunno how else to do it cause manipulating time is weird. blame Einstein and dries

Issue #5705, opened 2015-11-04
DBLog error for user language in API request


weerd:
We are attempting to set the `$language` [here](https://github.com/DoSomething/phoenix/blob/859331aaf2bb15dfe7a85860e8dc39e0261539ad/lib/modules/dosomething/dosomething_reportback/includes/Reportback.php#L151) but if user is anonymous (ie: anon GET request) we get errors.

weerd:
@blisteringherb I believe this is an update you made to include the `$language`. Would you mind filling me in or pointing me to the PR/Issue related to the update? I'm curious cause this isn't necessarily the user's language that created the RB, but the language for the requestor. Just trying to get more context on why it was included and where it ends up getting used once included in the object (since there's also an error related to it, if it's an anon user).

blisteringherb:
@weerd It gets used for the reportback email to select the correct template for sending the reportback email to the user.

weerd:
Gotcha @blisteringherb ... so this would send an email to the user regarding their specific reportback? Cause from the API side, it's always going to be the same if it's the API user... so it's not the specific language for the user that reported back.

blisteringherb:
Yup, that's right. Will you be sending reportbacks via the API?

weerd:
Yar.

blisteringherb:
Hm, how will we be getting the user's language via the API?

weerd:
@blisteringherb I can update since I have to fix the anon user issue w/ the language anyhoo. It would also make more sense to pop it into the User object nested in the Reportback object, that contains the rest of the user info too. Might just need to update the email part that uses that variable.

weerd:
@blisteringherb I suspect via Northstar, which is where we get the rest of the user data.

Could you point me to where in the code that `$language` property was otherwise then being used? Or is that a Message Broker thing?

weerd:
Also, something to keep in mind is that this Reportback class is currently ONLY really used for the API.

weerd:
@blisteringherb I chatted with @DeeZone and doesn't seem like this is being used within email things. Could it have been an update specifically for mobile app stuff? @jonuy mind chiming in?

jonuy:
Can we set a default language in the Reportback code to like, english or global english or whatever the default is? Northstar I think does make some of the GET /reportback calls, but the mobile app itself also makes its own directly to Phoenix. With Phoenix as the final destination for both of the requests, having a fallback set there I think makes sense.

[...]

PR #5717, opened 2015-11-05
Api undefined property bug fix

### LINKED ISSUE #5706: DBLog error with why_participated and flagged in API request
Getting errors when assigning the `why_participated` and `flagged` properties on a Reportback. The database no longer seems to be adding that data in with the returned query response.

weerd:
Fixes #5706 
Refs #5705 
#### What's this PR do?

If you are an anonymous requester retrieving data from the api, some properies such as 'flagged' and 'why_participated' are not exposed, but were not being checked properly to provide a sensible default. Using the dosomething_helpers_isset() function allows us to check the data for the property and if not set assign a default in one fell swoop!

Just take a gander at the changes; they're pretty straightforward ;)
#5706

---

@angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
+++ b/lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
@@ -89,9 +89,6 @@ class Reportback extends Entity {
     $reportbacks = [];
     $results = dosomething_reportback_get_reportbacks_query_result(['rbid' => $ids]);
 
-//    print_r($results);
-//    die();
-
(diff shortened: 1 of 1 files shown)


angaither:
:+1: 
tbd: I feel we might want to tweak the lang default? but not needed for merge

weerd:
Yep. I did that for now. I'm pretty sure it's going to get entirely removed from the RB object and the nested Campaign and User objects will have `language` properties assoc with them instead? Not entirely sure; still having some convos about what makes the most sense.

PR #5720, opened 2015-11-05
Prevents regional admins from being redirected on node edit page.

### LINKED ISSUE #5716: Incorrect redirect of global admins on node edit page
As an admin accessing the backend from outside the US I'm trying to edit nodes that are not translatable, like this:
https://thor.dosomething.org/node/5780/edit
But it redirects me to:
https://thor.dosomething.org/node/5780/edit/add/en/en-global

Hence I can't see or edit all untranslatable nodes.

---

cc @mikefantini

sergiitk:
Fixes #5716.

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -38,7 +38,9 @@ function dosomething_global_init() {
 
   // Verify we're dealing with a node edit with no translation specification in the URL
   if ($node_variables['is_node'] && $node_variables['is_node_edit'] && $node_variables['node_edit_language'] == NULL) {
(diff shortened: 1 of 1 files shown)


sergiitk:
@mikefantini asks:

>  is this only for nodes that are not translatable? We ​_do_​ want this functionality on nodes that are translatable

itsjoekent:
Nono this prevents non regional admins from being redirected (so 'us') but it now needs to be adapted to work (https://github.com/DoSomething/phoenix/issues/5716#issuecomment-154124890)

do you got it or want me to hop back on it?

edit: sorry this makes no sense, clarified in slack

sergiitk:
This is wrong.

PR #5721, opened 2015-11-05
Moves user redirect above node redirect

### LINKED ISSUE #5683: Logging in on a non-campaign page from the US takes user to /user/ not /us/user/

itsjoekent:
# WHAT

Fixes the issue where user pages weren't being prefixed
# HOW

There is a node translation check occurring before the user redirect, which obviously fails because this is a user page. I just moved this before that check.
# PROOF

Does login / out still work?
Are user profiles redirected correctly?

Fixes #5683

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -78,6 +78,17 @@ function dosomething_global_init() {
     return;
   }
 
+  // Handle translation redirects for authenticated users when they request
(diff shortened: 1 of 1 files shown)


sergiitk:
Looks :+1:

PR #5723, opened 2015-11-05
Adds check to make sure its a translatable content type

### LINKED ISSUE #5716: Incorrect redirect of global admins on node edit page

itsjoekent:
#### What's this PR do?

Makes sure the node is translatable before performing a node edit redirect 

Can you correctly access a campaign group?

This was the only reliable method I could find for determining if content is translatable or not

Fixes #5716

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -38,7 +38,10 @@ function dosomething_global_init() {
 
   // Verify we're dealing with a node edit with no translation specification in the URL
   if ($node_variables['is_node'] && $node_variables['is_node_edit'] && $node_variables['node_edit_language'] == NULL) {
-    dosomething_global_redirect_node_edit($node_variables['node'], $user_variables['user_language']);
+    // Make sure this is a translatable content type
+    if ($node_variables['node']->translations->original != NULL) {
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -38,7 +38,10 @@ function dosomething_global_init() {
 
   // Verify we're dealing with a node edit with no translation specification in the URL
   if ($node_variables['is_node'] && $node_variables['is_node_edit'] && $node_variables['node_edit_language'] == NULL) {
-    dosomething_global_redirect_node_edit($node_variables['node'], $user_variables['user_language']);
+    // Make sure this is a translatable content type
+    if ($node_variables['node']->translations->original != NULL) {
```
sergiitk: This could generate php notices or even warnings (depending on how the class structured).

``` php
if (!isset($node_variables['node']->translations->original)) {
```

Only in very very rare cases you want to compare something with NULL. 99% is either `isset()` or `empty()`.

itsjoekent:
@sergii-tkachenko all good to go here?

sergiitk:
:+1:

PR #5727, opened 2015-11-05
Adds langauge to page preprocessing

### LINKED ISSUE #5666: Subtitle not displaying on static content
https://staging.dosomething.org/node/538/edit

image

image

This is occurring on prod, staging, and thor.

itsjoekent:
#### What's this PR do?

Adds the language when extracting fields in the page preprocess. Thus fixing the subtitle not displaying.

Do subtitles display correctly?

We switched to global and that LANGUAGE_NONE is obsolete here!

Fixes #5666

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
@@ -151,12 +151,16 @@ function paraneue_dosomething_preprocess_page(&$vars) {
       'title' => drupal_get_title(),
     );
     // If there's a subtitle field on this node, send it to the header
-    if (isset($vars['node']->field_subtitle[LANGUAGE_NONE][0]['safe_value'])) {
-      $header_vars['subtitle'] = $vars['node']->field_subtitle[LANGUAGE_NONE][0]['safe_value'];
+    global $user;
+    $language = $user->language;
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
```diff
@@ -151,12 +151,16 @@ function paraneue_dosomething_preprocess_page(&$vars) {
       'title' => drupal_get_title(),
     );
     // If there's a subtitle field on this node, send it to the header
-    if (isset($vars['node']->field_subtitle[LANGUAGE_NONE][0]['safe_value'])) {
-      $header_vars['subtitle'] = $vars['node']->field_subtitle[LANGUAGE_NONE][0]['safe_value'];
+    global $user;
+    $language = $user->language;
```
weerd: Not sure off hand, but does this run when it's an anonymous user? Cause this will result in `undefined` errors in the **dblog** if this gets loaded at whatever point by user that's not logged in...

You can do something™ similar to what I did here to help avoid the error for anonymous user:
https://github.com/DoSomething/phoenix/blob/29373fa2ba9e67f33e4c7e5aa7f26fdb0aa4bd2c/lib/modules/dosomething/dosomething_reportback/includes/Reportback.php#L155

Just not sure if the default should be `en-global`, etc...

weerd:
Coooooool :+1:

PR #5728, opened 2015-11-05
pass TRUE to load individual reportbacks and reportback-items

### LINKED ISSUE #5707: Add load_user parameter to the retrieve operation for RB and RB Items endpoints
We allow specifying to `load_user=true` for the **index** collection API operation, but not for when showing individual **retrieve/show** for Reportbacks and ReportbackItems.

chloealee:
#### What does this PR do?

Passes `TRUE` when retrieving individual reportbacks or reportback-items to always load full user data. 

Both `GET http://dev.dosomething.org:8888/api/v1/reportback-items/:id.json` and `GET https://www.dosomething.org/api/v1/reportbacks/:id.json` endpoints will return full user data.

Fixes #5707


weerd:
:+1:

PR #5731, opened 2015-11-05
Ensures that the desired language is available.

### LINKED ISSUE #5730: PHP warning on static pages due to data not being available in specified language
Just need to ensure that the language is available before attempting to extract the content from the field data.

weerd:
Fixes #5730
Refs #5727
#### What's this PR do?

When extracting field datat if a specific language is provided, then we need to make sure that if the data is not available in the language specified, then return NULL and exit out of the function otherwise we will get PHP warnings.

With the new conditional added to **dosomething_helpers.module**

We were getting a bunch of PHP Warnings because a language was being passed, but the data was not available in that language and we didn't have a check for that.
#5730

---

@deadlybutter 

cc: @angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -118,6 +118,10 @@ function dosomething_helpers_extract_field_data($field, $language = LANGUAGE_NON
     }
     // ---
 
+    if (!array_key_exists($language, $field)) {
+      return NULL;
+    }
+
     $data = $field[$language];
 
     $values = array();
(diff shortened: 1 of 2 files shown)


itsjoekent:
I'm gonna have to modify the preprocess to allow for a NULL possibility but this PR is :ok:

weerd:
@deadlybutter I don't think you will need to, since you already have checks in place:

https://github.com/DoSomething/phoenix/blob/d97bf592f59d917db51277f4543255628dd0d2d4/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc#L168

That will result in `FALSE` and thus not run; which is exactly what happened when I tested :) Should be ok!

Issue #5664, opened 2015-10-29
Potentially refactor how language code is acquired.


weerd:
Where we are retrieving and setting the language code here:
https://github.com/DoSomething/phoenix/blob/a37f06325cd4dd13495be1f5e13c952676d2ead2/lib/modules/dosomething/dosomething_fact/dosomething_fact.module#L90

Might be easier to user this helper function instead:
https://github.com/DoSomething/phoenix/blob/a37f06325cd4dd13495be1f5e13c952676d2ead2/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module#L578

Test to make sure that what is returned as the language code is actually based on the url path, because that is what we require for how we decided what regional Fact content to show.

cc: @angaither

itsjoekent:
I think this requires a bigger discussion planning, as I think a few parts of the code will need to be refactored. The big problem comes down to when a single country has two languages, as currently all of our functions assume its 1 lang per country 
_(TLDR: We need to decouple the country and language logic)_

Did some research and I think this is our best option. We'll create a new function to read the "Accept-Language" header all browsers should be sending. Whats great about this header is that it sends the sub code as well if its one of those cases where its not the primary language of the country. From the W3 website,

> Mainstream browsers allow these language preferences to be modified by the user. The value itself is a defined by BCP 47, typically as a two or three letter language code (eg. fr for French), followed by optional subcodes representing such things as country (eg. fr-CA represents French as spoken in Canada).

Debugging will be more annoying cause I'm not sure how we'll be able to fake this / override the browser (we can edit our browser language I guess), however the alternative is storing a giant list of cities and there associated language, and comparing that on every request. Not ideal.

itsjoekent:
I'll prototype around with that ^ and report my findings

---

This is the rule I'll follow _(running on the assumption we're adding a language selector)_

Anonymous user: 
- Get Country => Fastly
- Get Language => Accept User Lang header

Auth user:
- Get country => Fastly 
- lang when signing up => Accept User Lang header
- lang everywhere else => User lang

The user lang header technically comes with the matching country, however
1) According to W3 it's not always sent
2) Our country logic should stay the way it is so we know when to call them a global user or not

itsjoekent:
Started implementing ^^, so far everything works :ok: 

all the refactoring I've done so far --> 
https://github.com/DoSomething/phoenix/compare/DoSomething:dev...deadlybutter:5664

itsjoekent:
[...]

PR #5733, opened 2015-11-06
Filing divorce papers for languages and countries.

itsjoekent:
### EXPLOSIVE CHANGES

:boom: :boom: :boom: :boom: :boom: :boom: :boom: 

This PR is removing the connection between languages and countries in the code. For further global expansion this is a must. Example:

If a Canadian user visits the site, do we serve them french or english?

Our current logic wouldn't know, and just return the first lang listed under the country. Many distinctions and assumptions are made throughout the code that one language always relates to one country (except global). For 1.0 this is ok, but going beyond that we need to update this.

To solve it, I'm relying on a header value called "Accept-Language" that's basically guaranteed to be sent by the browser. You can see the reasoning I chose to use this in the issue for this PR here (and more details on how it works) https://github.com/DoSomething/phoenix/issues/5664#issuecomment-153436293

In the event a Accept-Language header is not sent by the browser, there is a fallback. First, for authenticated users it will try your account language. If that is not set, or you're an anonymous user, it will return global english.

Here is where things get complicated, **for testing this you now need to change your browser language** in addition to using Torgaurd. Yes, this makes testing more annoying, but it's also the most accurate simulation we can make next to actually being in the country. 

After this PR is merged I'll be able to begin the large global documentation readme & diagramming that explains which functions should be used in what cases, in the event parts of this are still unclear.
### EXPLOSIVE CHANGES

:boom: :boom: :boom: :boom: :boom: :boom: :boom: 
**Do not merge yet, still testing and developing**

### DIFF
--- a/lib/modules/dosomething/dosomething_fact/dosomething_fact.module
+++ b/lib/modules/dosomething/dosomething_fact/dosomething_fact.module
@@ -86,7 +86,7 @@ function dosomething_fact_get_mutiple_fact_field_vars($node, $field_names) {
 function dosomething_fact_get_facts_data($node, $field_names) {
   $facts = [];
   $sources = [];
-  $language = dosomething_global_convert_country_to_language(dosomething_global_get_current_prefix());
+  $language = dosomething_settings_format_browser_language();
 
   foreach($field_names as $field_name) {
     $entity = NULL;
--- a/lib/modules/dosomething/dosomething_fact/dosomething_fact.module
+++ b/lib/modules/dosomething/dosomething_fact/dosomething_fact.module
@@ -86,7 +86,7 @@ function dosomething_fact_get_mutiple_fact_field_vars($node, $field_names) {
 function dosomething_fact_get_facts_data($node, $field_names) {
   $facts = [];
   $sources = [];
-  $language = dosomething_settings_format_browser_language();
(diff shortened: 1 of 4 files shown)


sheyd:
@deadlybutter Thor deployments only work from the Phoenix repo.

itsjoekent:
gotcha, when its ready for testing I'll make another branch here than real quick

itsjoekent:
Basic testing on Thor really quick seems to be fine. Will do more later

itsjoekent:
sena im assigning you to this so we remember to investigate this later, gonna hit pause on this for now and get some of the bugs out of the way

cc @mikefantini @angaither

deezone:
@deadlybutter Fantastic that you're getting to this issue. It warms my Canadian heart.

If I understand the description correctly this makes the assumption that the browser language setting takes precedence over the DoSomething user account (session) language preference? This seems problematic if the user logs into different computers with different language preferences at the browser level (a possibility for a typical Canadian student in a shared school computer lab). 

Wouldn't the user specific language preference first, then browser setting second not be a more accurate choice?

itsjoekent:
No with this change the user language still _should_ always take precedence. This mostly impacts anonymous users.

Also not sure when / if this will be merged, had complications and then we had higher priority things to take care of.

PR #5737, opened 2015-11-06
Adding missing term to POT file and fixing text wrapped in t().

weerd:
Refs #5517
#### What's this PR do?

This PR adds a term from the campaign theme to the **paraneue_dosomething.pot** file and also updates the code with how the term is wrapped in `t()` function.

Take a quick :eyes: 

Will be tested on Thor.

So when importing PO files that have terms with html tags inside of them into Drupal, Drupal gets angry and doesn't accept them. So I'm trying an alternative and using the `\r\n` characters instead of explicitly using the `<br>` tag to see if this works better.
#5517

---

@angaither


angaither:
:+1:

PR #5748, opened 2015-11-09
Missing translations 04 olé

weerd:
Refs #5517
#### What's this PR do?

This PR adds a missing term to the **paraneue_dosomething.pot** file and also ensures the associated term is wrapped in the `t()` function wherever it is output.

Take a quick :eyes: 

Will be tested on Thor.
#5517

---

@angaither


angaither:
:+1:

PR #5749, opened 2015-11-09
Revising a term and updating styles with how it is rendered.

weerd:
Refs #5517
#### What's this PR do?

This PR revises a term in the **paraneue_dosomething.pot** file and associated term in the `t()` function wherever it is output. It also adjusts some styles related to how the term is output in the Reportback gallery. There is an issue with include html tags (or even `\r\n` characters) in a term imported via PO file into Drupal, so the best solution was to remove the extra phrasing (which was planning on being shortened anyway).

Take a quick :eyes: 

Will be tested on Thor.
#5517

---

@angaither 

cc: @DFurnes


DFurnes:
A lovely new coat of paint! :art:

angaither:
what happened to the :cool: kids? 
:+1:

weerd:
All the cool kids realized we're all in this together! :globe_with_meridians:

weerd:
#global

PR #5757, opened 2015-11-10
removes last_name in returned user

### LINKED ISSUE #5756: Do not return last_name field in user object when using `load_user=true`
Since we are displaying `first_name` and photos, and a anon user can get this information, we do not want to expose the user's `last_name`

chloealee:
#### What does this PR do?

Removes `last_name` from returned user object when retrieving a reportback or reportback-item. Since we are displaying both `first_name` and `photo` in the user object (with the param `load_user=true`) we do not want to expose the user's `last_name`.

This gets deleted in Transformer.php but is also taken out of the user object in Reportback.php and ReportbackItem.php

http://dev.dosomething.org:8888/api/v1/reportback-items.json?load_user=true and/or
http://dev.dosomething.org:8888/api/v1/reportbacks.json?load_user=true should not have `last_name` in the user object.

Fixes #5756 
cc @jonuy @aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -459,7 +459,6 @@ abstract class Transformer {
     return array(
       'id' => $data->id,
       'first_name' => $data->first_name,
-      'last_name' => $data->last_name,
       'photo' => $data->photo,
       'country' => $data->country,
     );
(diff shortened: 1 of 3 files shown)


weerd:
I think we should leave the last name in the `Reportback` and `ReportbackItem` classes, but keep it out of the Transformer like you are doing. The classes are internal to the app, while the Transformer is the only thing that takes that data and exposes it to the api. We internally may need the last name for the user down the line.

chloealee:
got it, thanks @weerd! Makes a lot of sense to have it there if we need access to it but just not expose it. Added it back in!

weerd:
:+1:

PR #5760, opened 2015-11-11
Change additional text type from text, to formatted text.

### LINKED ISSUE #5302: Translated content is shown as cached arrays
to recreate go to https://thor.dosomething.org/mx/node/2070 as a user with language set as MX SP.

you'll see something like this:

image

even though:
- "Bzzzinga! Esto es lo que necesita saber para crear sus bandas abejorros." should be the campaign starter statement - nothing showing
- "Averiguar donde plantar las flores. Usted puede comenzar un jardín de abejas cualquier lugar donde haya tierra - en su patio trasero, en la escuela, o puede Bee-ify un jardín ya existente." should be the Time and Place - but "Array" is showing

angaither:
Since this is a long_text field type, we need to grab the value(), otherwise Array is printed.
Fixes #5302


sergiitk:
:+1:

PR #5766, opened 2015-11-12
Removes entity usage in static content long text preprocess

### LINKED ISSUE #5747: /br/sobre/faqs and /mx/sobre/contacto throw error messages
https://thor.dosomething.org/br/sobre/faqs

https://thor.dosomething.org/mx/sobre/contacto

image

"The website encountered an unexpected error. Please try again later."

All other static content pages seem to work.

itsjoekent:
#### what

Removes entity usage in the static content pre processing for the long text fields. 
#### why does it work?

Cause these entity wrappers are being :hankey: lately with global stuff. good ol dosomething helpers extract field data saves the day again!
#### test it?

All static content should load without error now, regardless if it has intro filled out or not 

fixes #5747

### DIFF
--- a/lib/modules/dosomething/dosomething_static_content/dosomething_static_content.module
+++ b/lib/modules/dosomething/dosomething_static_content/dosomething_static_content.module
@@ -50,8 +50,7 @@ function dosomething_static_content_preprocess_node(&$vars) {
             break;
 
           case 'text_formatted':
-            global $user;
-            $vars[$label] = dosomething_helpers_extract_field_data($vars['node']->{$field_name}, $user->language);
+            $vars[$label] = dosomething_helpers_extract_field_data($vars['node']->{$field_name}, $vars['node']->language);
             break;
 
           case 'image':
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_static_content/dosomething_static_content.module
```diff
@@ -50,7 +50,8 @@ function dosomething_static_content_preprocess_node(&$vars) {
             break;
 
           case 'text_formatted':
-            $vars[$label] = $field->value->value();
+            global $user;
+            $vars[$label] = dosomething_helpers_extract_field_data($vars['node']->{$field_name}, $user->language);
```
angaither: i think this language thing might cause some wonkyness since we didn't translate static content.
there might not be content in the user's language

itsjoekent: so use the nodes language?

angaither: yeah, I think so

itsjoekent:
@angaither :boom: done!

angaither:
:+1:

PR #5768, opened 2015-11-12
Fixes Array showing on static content

itsjoekent:
quick fix for static content, works on my local now: 

screen shot 2015-11-12 at 4 16 09 pm

Accidentally set the fields array to the variable instead of the formatted field.


itsjoekent:
about page

screen shot 2015-11-12 at 4 20 07 pm

angaither:
:+1:

PR #5769, opened 2015-11-12
Update grunt-sass version.

DFurnes:
The older version relied on node-sass 2.x, which had a [known error with compiling Libsass](http://dfurn.es/1PqJnsU). This has been tested manually and works now on Thor.

:computer: :fire: :fire_engine: 

For review: @sheyd


sheyd:
:8ball:

PR #5772, opened 2015-11-13
Builds a correct global url when getting campaign block vars

### LINKED ISSUE #5673: Campaigns on SMS Game Confirmation Page go to /campaigns
After I complete an SMS game, I get taken to a page confirming my submission went through: https://thor.dosomething.org/us/campaigns/5685/confirmation

The campaigns that are highlighted should link to the campaign page. Currently, the user gets taken to /campaigns for each.

This is working on production, but not on thor.

image

itsjoekent:
#### What's this PR do?

Uses the global way of building node URL's when getting the campaign block vars

Doe the campaign gallery thing-er-mo-bobs work? eg: the one here: http://dev.dosomething.org:8888/us/campaigns/1269/confirmation

The old url's being generated weren't very global friendly. These are!

Fixes #5673

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -542,7 +542,12 @@ function dosomething_campaign_get_campaigns_doing($uid = NULL) {
 function dosomething_campaign_get_campaign_block_vars($nid, $image_size = '740x480', $source = NULL) {
   $node = node_load($nid);
   $clc = dosomething_helpers_get_current_language_content_code();
-  $path_alias = drupal_get_path_alias('node/' . $nid);
+
+  global $user;
+  $language = dosomething_global_get_language($user, $node);
+  // Build a global URL and remove the first slash returned
+  $path_alias = preg_replace('/\//', '', dosomething_global_url('node/' . $nid, array('language' => $language)), 1);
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -542,7 +542,12 @@ function dosomething_campaign_get_campaigns_doing($uid = NULL) {
 function dosomething_campaign_get_campaign_block_vars($nid, $image_size = '740x480', $source = NULL) {
   $node = node_load($nid);
   $clc = dosomething_helpers_get_current_language_content_code();
-  $path_alias = drupal_get_path_alias('node/' . $nid);
+
+  global $user;
+  $language = dosomething_global_get_language($user, $node);
```
angaither, line 547: shouldn't this be the node lang? what if the campaign doesn't exist in the user's lang?

itsjoekent, line 547: thats the point of the function, it first checks if a translation exists in a users lang and then defers to the default language (global) and if nothing there then the source language 

https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_global/dosomething_global.module#L494

angaither:
:+1:

PR #5776, opened 2015-11-13
Before calling theme on video id, check there's a value in the field.

### LINKED ISSUE #5773: A black rectangle lingers after removing a video on the campaign template
https://staging.dosomething.org/us/campaigns/teens-jeans

image

@angaither we need a solution for this for Monday's Teens for Jeans soft launch.

angaither:
Fixes #5773


blisteringherb:
:+1:

PR #5777, opened 2015-11-16
Api cleanup

weerd:
#### What's this PR do?

This PR does some general cleanup of the API code and makes a couple fixes.

With the changed files and just double-check everything looks alright.

---

@angaither 

cc: @jonuy

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -32,7 +32,7 @@ function _campaign_resource_definition() {
           ],
           [
             'name' => 'mobile_app',
-            'description' => 'Flag to indicate request was made from the mobile app.',
+            'description' => 'Flag to select campaigns available on the mobile app.',
             'optional' => TRUE,
             'type' => 'boolean',
             'source' => ['param' => 'mobile_app'],
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -241,7 +241,7 @@ function _campaign_resource_access_alt($op = 'view', $args = array()) {
 /**
  * Determine whether the current user can access campaign resource.
  *
- * @param string $op
+ * @param  string  $op
  * @return bool
  */
 function _campaign_resource_access($op) {
@@ -260,7 +260,7 @@ function _campaign_resource_access($op) {
 /**
  * Retrieve and show response for campaign request.
  *
- * @param string $nid
+ * @param  string  $nid
  * @return object
  */
 function _campaign_resource_retrieve($nid) {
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -271,14 +271,14 @@ function _campaign_resource_retrieve($nid) {
...
  *
- * @param string $ids
- * @param bool $staff_pick
- * @param bool $mobile_app
- * @param string $mobile_app_date
- * @param string $term_ids
- * @param int $count
- * @param bool $random
- * @param int $page
+ * @param  string  $ids
```
jonuy, line 274: Lining up the type and vars like this, is this a style we should be doing in all our PHP code now? Pretty sure we haven't been doing it in Northstar.

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -58,7 +58,7 @@ function _campaign_resource_definition() {
             'name' => 'count',
             'description' => 'The number of campaigns to retrieve.',
             'optional' => TRUE,
-            'type' => 'string',
+            'type' => 'int',
```
jonuy, line 61: This seems fine. Just curious what it was before that broke when it was a  `string` and if there's potential for anything new to break with this change.

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
weerd, line 274: Yeah, not sure. I'm trying it. It's something I pulled from how Laravel does docblocks (which goes to extra lengths to make sure the code and comments look appealing and are easier to digest). Kinda curious if others find it useful/appealing.

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
weerd, line 61: Nothing really broke. I believe in the each resources respective class Transformer we cast the value to an integer anyway, to make sure it's an int; but figured I'd update this to be consistent and so that Services will know ahead of time if an `int` is not passed.

angaither:
:+1:

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
jonuy, line 274: Just my two cents, I tend to not do it in our Java code because you'll have wildly different name sizes for your types sometimes which I think can be weird. So you could see something like...

```
* @param bool                is_diego
* @param CampaignTransformer transformer
* @param int                 some_num
* @param Campaign            campaign
```

... which, eh I guess isn't terrible.

DFurnes, line 274: One other concern I'd have is this messes with diffs (i.e. you may have to change all lines for indentation when you only _really_ changed one parameter). I don't have particularly strong feelings either way, but something to consider! :)

weerd, line 274: @uy yeah, I was concerned about that too, but curious to see what you think of how Taylor handles that same issue in Laravel. See [here](https://github.com/laravel/framework/blob/1e55c79fd1d3e778d867a43f5495957fcb52a655/src/Illuminate/Cache/DatabaseStore.php#L45) for an example. Seems to use hierarchy in that scenario.

weerd, line 274: While I do find it more appealing, I would rather figure out something that everyone feels comfortable doing so we're all on the same page. Not sure what the proper venue for discussion is, but definitely something to tackle during documentation days...

weerd, line 274: @DFurnes that's part of the beauty of Laravel's approach... if a specific class is referenced with the path, etc that aligns on its own depending on length of path, as long as it's 2 spaces between the class and the variable name. Only really the parameters that are specifically data types do the whole alignment thing with varying spaces... The amount of "space" adding would be minimal.

```
Example:
/**
 * Create a new database store.
 *
 * @param  \Illuminate\Database\ConnectionInterface  $connection
 * @param  \Illuminate\Contracts\Encryption\Encrypter  $encrypter
 * @param  string  $table
 * @param  string  $prefix
 * @return void
 */

Example:
/**
 * Increment or decrement an item in the cache.
 *
 * @param  string  $key
 * @param  mixed   $value
 * @param  \Closure  $callback
 * @return void
 */
```

PR #5778, opened 2015-11-16
Api caching

### LINKED ISSUE #5763: Drupal caching on campaign API endpoints
Both Runscope tests and anecdotal mobile app testing have shown occasionally slow response times or 503 error codes on GETs to the campaigns API endpoint. I'm not sure how much of a gain we'd get here compared to #5762 and I'm not even particular certain on what objects we'd cache, but it's worth investigating.

In particular, the mobile app makes calls like:
- GET www.dosomething.org/api/v1/campaigns.json?term_ids=1300,1301,1302,1303&mobile_app=true&mobile_app_date={yyyy-mm-dd}
- GET www.dosomething.org/api/v1/campaigns.json?term_ids=1300&mobile_app=true&mobile_app_date={yyyy-mm-dd}
- GET www.dosomething.org/api/v1/campaigns.json?term_ids=1301&mobile_app=true&mobile_app_date={yyyy-mm-dd}
- GET www.dosomething.org/api/v1/campaigns.json?term_ids=1302&mobile_app=true&mobile_app_date={yyyy-mm-dd}
- GET www.dosomething.org/api/v1/campaigns.json?term_ids=1303&mobile_app=true&mobile_app_date={yyyy-mm-dd}

**Bonus though should totally maybe be a separate issue**
These calls also get made from the Android app and Runscope's recently been showing ~10000 ms response times on them from staging at least:
- [staging] GET www.dosomething.org/api/v1/taxonomy_term/667.json
- [staging] GET www.dosomething.org/api/v1/taxonomy_term/668.json
- [staging] GET www.dosomething.org/api/v1/taxonomy_term/669.json
- [staging] GET www.dosomething.org/api/v1/taxonomy_term/670.json
- [prod] GET www.dosomething.org/api/v1/taxonomy_term/1300.json
- [prod] GET www.dosomething.org/api/v1/taxonomy_term/1301.json
- [prod] GET www.dosomething.org/api/v1/taxonomy_term/1302.json
- [prod] GET www.dosomething.org/api/v1/taxonomy_term/1303.json

cc: @marahml @angaither

weerd:
Fixes #5763
#### What's this PR do?

This PR creates a new API caching class, `ApiCache` within the **dosomething_api** module that we can use to cache any of the API requests to Phoenix. It creates a unique id for requests based on the set of URL parameters (if any) passed along with the endpoint request and caches it for 24 hours.

There is also a new parameter added that can be included in the request URL (`/campaigns.json?cache=false`) which indicates that the request should not be cached, and if it was already cached in a prior request, to go ahead and clear the cache. This will help with development purposes.

Start at the **campaign_resource.inc** file and follow the breadcrumbs from there!

If you feel the desire to pull down a copy of this branch, run `drush updb` to get the new `cache_dosomething_api` table set up and then run a couple API queries on the `/campaigns.json` endpoint and confirm it gets cached in the database.

Currently this caching only works for the `/campaigns` index collection endpoint, but can be expanded to the rest.

One thing I noticed was that the string booleans passed in the URL parameters are not getting properly casted to actual booleans in the resource transformers, because it casts a string... so there's a new helper function `dosomething_helpers_convert_string_to_boolean()` created to help properly change these string booleans to actual booleans. It will be applied to other URL param items in upcoming PR.
#### Benchmarks

These benchmarks are from caching to the database; expect it to be even faster when caching to Redis. Including for future reference:

```
Request to /campaigns.json?ids=1663
- Before caching: 0.026s
- After caching:  0.002s

Request to /campaigns.json
- Before caching: 0.362s
- After caching:  0.006s
```
#5763

---

@uy @angaither 

cc: @DFurnes

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/ApiCache.php
+++ b/lib/modules/dosomething/dosomething_api/includes/ApiCache.php
@@ -2,18 +2,33 @@
 
 class ApiCache {
 
+  /**
+   * @param string  $endpoint
+   * @param array   $parameters
+   * @return mixed
+   */
   public function get($endpoint, $parameters) {
     $id = $this->generate_id($endpoint, $parameters);
 
-    return cache_get($id, 'cache_dosomething_api');
+    $cache = cache_get($id, 'cache_dosomething_api');
+
+    if ($cache && $cache->expire < REQUEST_TIME) {
+      cache_clear_all($id, 'cache_dosomething_api');
+
+      return FALSE;
+    }
+
+    return $cache;
   }
 
 
   /**
-   * @param  $endpoint
-   * @param  $parameters
-   * @param  $data
+   * @param  string  $endpoint
+   * @param  array   $parameters
+   * @param  mixed   $data
    * @return bool
+   *
+   * @todo: not ideal that Drupal doesn't return anything concrete from the cache_set() function.
    */
   public function set($endpoint, $parameters, $data) {
     $id = $this->generate_id($endpoint, $parameters);
--- a/lib/modules/dosomething/dosomething_api/includes/ApiCache.php
+++ b/lib/modules/dosomething/dosomething_api/includes/ApiCache.php
@@ -3,6 +3,30 @@
 class ApiCache {
 
   /**
+   * Clear a specified item from cache.
+   *
+   * @param  mixed  $id
+   * @param  mixed  $endpoint
+   * @param  mixed  $parameters
+   * @return bool
+   */
+  function clear($id = NULL, $endpoint = NULL, $parameters = NULL) {
+    if (!$id) {
+      if ($endpoint && $parameters) {
+        $id = $this->generate_id($endpoint, $parameters);
+      }
+      else {
+        return FALSE;
+      }
+    }
+
+    cache_clear_all($id, 'cache_dosomething_api');
+
+    return TRUE;
+  }
+
+
+  /**
    * Get data from cache based on id from endpoint and URL parameters.
    *
    * @param  string  $endpoint  Type of resource based on endpoint.
@@ -12,10 +36,16 @@ class ApiCache {
   public function get($endpoint, $parameters) {
     $id = $this->generate_id($endpoint, $parameters);
 
+    if (!dosomething_helpers_convert_string_to_boolean($parameters['cache'])) {
+      $this->clear($id);
+
+      return FALSE;
+    }
+
     $cache = cache_get($id, 'cache_dosomething_api');
 
     if ($cache && $cache->expire < REQUEST_TIME) {
-      cache_clear_all($id, 'cache_dosomething_api');
+      $this->clear($id);
 
       return FALSE;
     }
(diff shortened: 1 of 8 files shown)


angaither:
SUCH OO!

angaither:
VERY OBJECT

angaither:
WOW

weerd:
:dog2:

weerd:
so... cool to merge? lol

lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -56,14 +56,14 @@ protected function getNextPageUri($data, $parameters, $endpoint) {
...
-   * @param array $data
+   * @param  array $data
    *   An associative array of pagination parameters:
    *   - total: (int) Complete total number of available data items found.
    *   - per_page: (int) Total number of data items to show per page.
    *   - current_page: (int) Current page number.
    *   - total_pages: (int) Total number of pages available.
-   * @param array $parameters An associative array of current location parameters.
-   * @param string $endpoint Endpoint for building URI.
+   * @param  array $parameters An associative array of current location parameters.
```
jonuy: Not doing the whole lining em up thing here?

```
* @param  array  $parameters An associative array of current location parameters.
* @param  string $endpoint Endpoint for building URI.
```

weerd: No lining up of the description, just the data type and the variable name. The variable names can range in size significantly, so the start of the description text would vary a lot and that'll just be a nightmare :scream:

lib/modules/dosomething/dosomething_api/includes/ApiCache.php
```diff
@@ -0,0 +1,110 @@
...
+   * @return bool
+   */
+  public function set($endpoint, $parameters, $data) {
+    $id = $this->generate_id($endpoint, $parameters);
+
+    if (!dosomething_helpers_convert_string_to_boolean($parameters['cache'])) {
+      return FALSE;
+    }
+
+    cache_set($id, $data, 'cache_dosomething_api', REQUEST_TIME + (60 * 60 * 24));
```
jonuy: I wouldn't be opposed to moving this `(60 * 60 * 24)` up as a class property? As a property it could make it easier to spot that the cache time is 24 hours.

lib/modules/dosomething/dosomething_api/includes/Transformer.php
jonuy: oh, yea, I thought I saw the variable names weren't lining up

jonuy: Same deal on lines 38, 39. Unless my eyes are failing me and they are lined up (which could totally be the case)

jonuy:
Only minor comments. Otherwise, :dog2:

lib/modules/dosomething/dosomething_api/includes/ApiCache.php
weerd: @uy yeah, I tried that but it wouldn't let me have an equation as the property. I could set the property to be the hours to cache, and that just gets passed in:

```
(60 * 60 * $this->expires_hours)
```

or something and with a better name for the property...

weerd:
@uy you were right and I hadn't finished cleaning up the comments on the Transformer class. Should be updated now :)

jonuy:
:shiba inu: :shiba inu: :shiba inu:

shiba inu

weerd:
yaaas :)

Issue #5742, opened 2015-11-06
Disable X-Frame-Options header


sergiitk:
See #5392, https://github.com/DoSomething/phoenix/issues/5392#issuecomment-154554172.

Before enabling CSP, we need to save into features and disable `X-Frame-Options` header.
Currently it's out of sync on prod and stage.

---

CC @sheyd

Issue #5392, opened 2015-10-06
Enable CSP (Content-Security-Policy) Within Phoenix


sheyd:
X-Frame-Options is deprecated per various sources. Good round-up here:
https://appsec-labs.com/portal/anti-clickjacking/

Cobalt report highlighted this. We need to figure out exact policy, and enable as features within Drupal SecKit:

https://www.dosomething.org/admin/config/system/seckit

@sergii-tkachenko I'll focus on making sure the security policy part is correct, and then I'll assign to you to put into features.

sheyd:
The following fields need to be set under `Content Security Policy`:

`default-src: 'self' *.dosomething.org`
`frame-src: 'self' *.dosomething.org optimizely.com`

Remaining tasks before pushing to Prod:
1. Deploy to Staging. 
2. Create an Optimizely test in for staging that includes use of a frame element.
3. Test Optimizely A/B test, and confirm all functionality.

@sergii-tkachenko assigning this to you to implement as part of Global 1.1. One you're done, assign it back to me, and I'll coordinate testing with @namimody.

sergiitk:
Reference: https://cobalt.io/dosomething/reports/59

sergiitk:
To record: staging and prod are desynchronized in `X-Frame-Options` settings.
`X-Frame-Options` returned `SameOrigin` on stage and `Allow-From: http://optimizely.com` on prod.
Hence provided clickjacking html example doesn't work on stage and works on prod.

sergiitk:
Note: `frame-src` direcrtive has been replaced with `child-src`.
http://www.w3.org/TR/CSP/#frame_src

sergiitk:
Chat log of @sheyd and myself:

​[11:32 PM] I did enable and tuned up csp
​[11:32 PM]  it took more rules to make it work
​[11:33 PM]  but it turned out that `frame-src` is a directive that regulates frames embedded into our page
​[11:33 PM]  not pages our page is embedded into
​[11:34 PM]  withregard X-Frame-Options
​[11:34 PM]  https://developer.mozilla.org/en-US/docs/Web/HTTP/X-Frame-Options
​[11:35 PM]  it does work as intened
​[11:35 PM]  but, allow-from isn’t supported in chrome and safari
​[11:36 PM]  CSP proposes the feature to solve it
​[11:36 PM] called `frame-ancestors`
​[11:36 PM] but only in v2
[...]

PR #5779, opened 2015-11-16
Enables X-Frame-Options header.

sergiitk:
We turned clickjacking protection because we thought it doesn't work, but it turned out that it's only Chrome and Safari that doesn't support `Allow-from` value of this header:
image
See https://developer.mozilla.org/en-US/docs/Web/HTTP/X-Frame-Options

Related #5392, #5742.


sheyd:
:+1:

PR #5785, opened 2015-11-16
Fixes regional users from not being able to see global content

### LINKED ISSUE #5620: Users should be able to access GL EN content if translation is unpublished or does not exist in their region
If a user is coming from Mexico and a campaign has a published translation in GL EN but MX SP translation either doesn't exist or is unpublished, then the Mexican user should 1. have the GL EN content surface via SOLR (finder/search/ /campaigns) 2. be able to access that content (i.e. not get 404'd).

MX user:
- GL EN = published
- MX SP = published
- go to www.dosomething.org/campaigns/sea-turtle-street-sweep, get redirected to https://www.dosomething.org/mx/voluntario/limpieza-de-calles-por-la-tortuga-marina and see MX SP content

MX user:
- GL EN = published
- MX SP = unpublished or does not exist
- go to www.dosomething.org/campaigns/sea-turtle-street-sweep, I get GL EN content

---

Coming from US as an anon user.

Re-poppin Bottles and Quote of a Champion have published translations in GL EN but not US EN. The GL EN versions show up in solr, but if hover over them they show a URL like, "thor.dosomething.org/us/campaigns/quote-champion"

image

and you get 404'd when you click:
image

itsjoekent:
#### What's this PR do?

Fixes global redirects for when the local translation doesn't exist
Fixes the links on the homepage to reflect the correct URL for the user.
Adds a session variable `dosomething_global_redirected` to keep track of when a user was redirected already (prevents redirect loop)

Do the cases outlined above ^ now work?

Mostly explained in the Whats this PR do

Fixes #5620

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -19,6 +19,11 @@ function dosomething_global_init() {
     return;
   }
 
+  if ($_SESSION['dosomething_global_redirected']) {
+    $_SESSION['dosomething_global_redirected'] = FALSE;
+    return;
+  }
+
   $languages = language_list();
   if (empty($languages[DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE])) {
     watchdog('dosomething_global', "Can't load %lang language",
@@ -603,7 +608,7 @@ function dosomething_global_redirect_node($user_lang, $node, $redirect_status) {
     drupal_exit();
     return;
   }
-
+  $_SESSION['dosomething_global_redirected'] = TRUE;
   // Might be overly formal: We could pull 'node/nid' from the current
   // request. This is in case of the unlikely (impossible?) event that
   // a translated version of this node has a different entity ID.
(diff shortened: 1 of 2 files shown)


sergiitk:
The code looks good.
Please reflect changes made to the `dosomething_global_redirect_node()` function in its docblock.

It would be perfect to also provide in comments real-world example it works like it works. You can use @mikefantini's description:

> - For on-site discovery purposes, we’ll be exposing all country specific content and GL EN content. Preference would be given to country specific content and obviously if a campaign exists in both MX SP and GL EN, we would show users the MX SP translation and not the GL EN translation.
> - When going directly to a regional URL, a user should be able to access any country specific translation regardless of where they are located (i.e. if I’m in BR, I should be able to see US EN translation by going directly to www.dosomething.org/us/campaigns/campaign-title).
> - When going to the base URL of a campaign, we should redirect the user to the published translation that is most appropriate for them.

itsjoekent:
When we do this upcoming documentation sprint I planned on putting longer forms of documentation like that in the module readme, I can quickly change the docblock though.

sergiitk:
:+1:

PR #5786, opened 2015-11-16
Preventing unpublished nodes being indexed in solr

### LINKED ISSUE #5570: unpublished translations showing up in solr
related #5519

image

image

blisteringherb:
#### What does this PR do?
- Fetches a patch submitted to D.O. to prevent the apachesolr_multilingual module from indexing unpublished translations.

Preventing unpublished translations from being indexed should fix several problems. There isn't currently a mechanism to filter out unpublished translations from Solr results in the Finder or in search, so this is the best option to ensure that all code that relies on Solr functions properly.

Rebuild phoenix, then delete and reindex Solr to make sure no unpublished translations show up in the index.

Fixes #5570


angaither:
:+1:

PR #5790, opened 2015-11-17
Forces a slash to be inserted into the URL

itsjoekent:
URL's we're _sometimes_ missing a slash after the prefix due to a change in here: https://github.com/DoSomething/phoenix/pull/5785

I've tested this change on 10 campaigns and they all worked :ok:


sergiitk:
:+1:

PR #5795, opened 2015-11-17
user data clean up

### LINKED ISSUE #5698: user data clean up
once we remove the temp fastly headers we will need to update the language and country for users with these countries
- [ ] xx
- [ ] xr

katiecrane:
fixes #5698 

db updates to change brazil and mexico users to the proper language and country code

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.install
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.install
@@ -30,43 +30,36 @@ function dosomething_global_update_7002(&$sandbox) {
 }
 
  /**
-  * change US users' language from en-global to english
+  * change US users' language from en-global to english,
+  * change MX users' language from global english to mexican spanish, country from XX to MX, 
+  * change BR users' language from global english to pt-br, country from XR to BR 
   */
 function dosomething_global_update_7003(&$sandbox) {
+  //US
   $results = db_query("SELECT u.uid, u.mail, a.field_address_country, u.language
                         FROM users u
                         INNER JOIN field_data_field_address a on a.entity_id = u.uid
                         WHERE u.language = 'en-global'
                         AND a.field_address_country = 'US';");
 
-  foreach ($results as $result)
-  {
+  foreach ($results as $result) {
     $user = user_load($result->uid);
-    $user->language = 'english';
+    $user->language = 'en';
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_global/dosomething_global.install
```diff
@@ -28,3 +28,45 @@
...
+  */
+function dosomething_global_update_7003(&$sandbox) {
+  $results = db_query("SELECT u.uid, u.mail, a.field_address_country, u.language
+                        FROM users u
+                        INNER JOIN field_data_field_address a on a.entity_id = u.uid
+                        WHERE u.language = 'en-global'
+                        AND a.field_address_country = 'US';");
+
+  foreach ($results as $result)
+  {
```
angaither: in drupal brackets should be on the line above `foreach (blah as blah) {`

lib/modules/dosomething/dosomething_global/dosomething_global.install
```diff
@@ -28,3 +28,45 @@
...
+    $user = user_load($result->uid);
+    $user->language = 'english';
```
angaither: does this work? shouldn't it be the code `en`?

lib/modules/dosomething/dosomething_global/dosomething_global.install
```diff
@@ -28,3 +28,45 @@
...
+/**
+ * change BR users' language from global english to pt-br, country from XR to BR 
+ */
+function dosomething_global_update_7005(&$sandbox) {
+  $results = db_query("UPDATE users as u
+                        INNER JOIN field_data_field_address a on a.entity_id = u.uid
+                        SET u.language = 'pt-br', a.field_address_country = 'BR', a.language = 'pt-br'
+                        WHERE u.language = 'en-global'
+                        AND a.field_address_country = 'XR';");
+
```
angaither: you can do these all in the same update hook, they don't need to be 3 different ones

lib/modules/dosomething/dosomething_global/dosomething_global.install
katiecrane: it doesn't break but yeah the code should be `en`

angaither:
:+1:

PR #5798, opened 2015-11-18
Add drush disable command to ds pull stage

### LINKED ISSUE #1507: Disable optimizely in ds pull

deezone:
Fixes #1507 

The Optimizely module needs to be disabled in staging builds to prevent it from running Javascript based tests. 
- `ds pull stage` triggers `pull_db if [[ ${alias[0]} == "stage" ]`

Using `drush`, disable the Optimizely module.

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -485,6 +485,9 @@ function pull_db {
       echo "Reindexing Solr..."
       drush solr-mark-all && drush solr-index
 
+      echo "Disabling Optimizely module on staging."
+      drush dis optimizely
+
     else
       echo "Pulling down the db from ${alias[1]} staging"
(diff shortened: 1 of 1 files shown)


bin/ds
```diff
@@ -485,6 +485,9 @@ function pull_db {
       echo "Reindexing Solr..."
       drush solr-mark-all && drush solr-index
 
+      echo "Disabling Optimizely module on staging."
+      drush dis optimizely
```
angaither: this should maybe be done earlier in the process, perhaps before cache is cleared?

deezone:
@angaither Good point, moved as suggested.

angaither:
:+1:

Issue #5796, opened 2015-11-17
malformed urls on campaigns with special chars


angaither:
try with `/node/2259` on thor/prod

clicked view from the node edit page
![image](https://cloud.githubusercontent.com/assets/645205/11221675/5e45d768-8d34-11e5-97c1-52d40c8830cd.png)

clicked on the campaign from the homepage
![image](https://cloud.githubusercontent.com/assets/645205/11221686/6ff62a12-8d34-11e5-9707-655fd8df4e87.png)

hit the url dosomething.org/node/2259
also got a 404

mshmsh5000:
Clicking through from the search results [works fine for me](http://cl.ly/3A1U2n3b0Z2G). This version of the URL works fine when pasted directly into Chrome, for instance:

https://www.dosomething.org/br/voluntario/bandeira-de-espa%C3%A7o-seguro

Chrome substituted the appropriate character, and then successfully makes the request: http://cl.ly/1X2x39351n04

In fact, I'm unable to reproduce this on any browser @mikefantini @angaither. I wonder if it's (a) a transient problem or (b) browser/browser-version-dependent.

angaither:
hm, also unable to reproduce coming from br
though when I clicked `view` from `/node/2259/edit/pt-br` I get taken to the english url with a redirect loop
![image](https://cloud.githubusercontent.com/assets/645205/11747864/8d5e4608-9ff3-11e5-848b-3a73a078e32d.png)

itsjoekent:
unable to reproduce redirect loop above ^ ?? should my torgaurd be enabled to test that?

angaither:
@deadlybutter yeah, was testing on thor, with torguard

angaither:
as a br editor, with my lang set ;)

itsjoekent:
connected to brazil via tor gaurd, set user lang to br, clicked edit, went to the /edit/pt-br, clicked view, went to https://www.dosomething.org/br/voluntario/bandeira-de-espa%C3%A7o-seguro (worked on, no redirect loop error).

da fuq is happening

angaither:
malformed is no longer an issue, opening a new one for the redirect loop

itsjoekent:
:repeat:

PR #5808, opened 2015-11-19
Fixes user profile campaign links

### LINKED ISSUE #5805: Clicking to Campaign from User Profile 404s
Clicking on a campaign photo, campaign title, or PROVE IT button on a user profile should bring you to the campaign action page (and clicking on the photo or PROVE IT button should bring you directly to the PROVE IT section of the campaign). 

Currently, clicking on an image or clicking on the PROVE IT button brings the user to a 404. The URL after clicking the image or the PROVE IT button is something like: https://www.dosomething.org/us/us/campaigns/staring-stars#prove. It is likely that the /us/us is what is causing it to 404.

image

cc @mikefantini

itsjoekent:
#### What's this PR do?

Fixes the campaign links on the user profile page

Do the links work now?

Something is funky with dosomething_global_url() and drupal path prefixes, this method will always work 

Fixes #5805

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -651,6 +651,24 @@ function dosomething_global_url($path, $options = NULL) {
 }
 
 /**
+ * Build a global node URL that adds the prefix and a fragment
+ *
+ * @param string $nid
+ *    The nid of the node
+ * @param string $language
+ *    The language of the URL
+ * @param string $fragment
+ *    Optional fragment to append
+ */
+function dosomething_global_node_url($nid, $language, $fragment = NULL) {
+  $raw_url = 'node/' . $nid;
+  if (isset($fragment)) {
+    $raw_url .= '#' . $fragment;
+  }
+  return $GLOBALS['base_url'] . '/' . dosomething_global_get_prefix_for_language($language) . '/' . drupal_get_path_alias($raw_url, $language);
+}
+
+/**
  * Return the current country prefix being used.
  *
  * @return string
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
@@ -658,6 +658,9 @@ function paraneue_dosomething_preprocess_taxonomy_term(&$vars) {
  */
 function paraneue_dosomething_preprocess_user_profile(&$vars) {
   global $base_url;
+  global $user;
+  $language = $user->language;
+  
   if ($doing = $vars['doing']) {
     foreach ($doing as $delta => $value) {
       $image = dosomething_image_get_themed_image($value['nid_image'], 'landscape', '400x400');
@@ -665,7 +668,7 @@ function paraneue_dosomething_preprocess_user_profile(&$vars) {
         'title' => $value['link'],
         'image' => $image,
         'description' => $value['call_to_action'],
-        'url' => dosomething_global_url($value['path_alias'], ['fragment' => 'prove']),
+        'url' => $GLOBALS['base_url'] . '/' . dosomething_global_get_prefix_for_language($language) . '/' . drupal_get_path_alias('node/' . $value['nid'], $language),
         'button_text' => t("Prove It"),
       );
       $doing_items[$delta] = paraneue_dosomething_get_gallery_item($item, 'figure');
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
@@ -660,7 +660,7 @@ function paraneue_dosomething_preprocess_user_profile(&$vars) {
   global $base_url;
   global $user;
   $language = $user->language;
-  
+
   if ($doing = $vars['doing']) {
     foreach ($doing as $delta => $value) {
       $image = dosomething_image_get_themed_image($value['nid_image'], 'landscape', '400x400');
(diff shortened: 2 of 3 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
```diff
@@ -665,7 +668,7 @@ function paraneue_dosomething_preprocess_user_profile(&$vars) {
         'title' => $value['link'],
         'image' => $image,
         'description' => $value['call_to_action'],
-        'url' => dosomething_global_url($value['path_alias'], ['fragment' => 'prove']),
```
angaither, line 668: we wanna keep the `#prove` fragment here

angaither:
@deadlybutter if something is wrong with `dosomething_global_url` perhaps we should update that so we have a reliable function to call when creating global urls

itsjoekent:
yeah I was just worried about breaking stuff if I messed with that function, standby I have an idea

lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
```diff
@@ -658,6 +658,9 @@ function paraneue_dosomething_preprocess_taxonomy_term(&$vars) {
  */
 function paraneue_dosomething_preprocess_user_profile(&$vars) {
   global $base_url;
+  global $user;
+  $language = $user->language;
```
sergiitk, line 662: You can use `dosomething_helpers_get_current_language_code()` instead if you want to.

lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
```diff
@@ -665,7 +668,7 @@ function paraneue_dosomething_preprocess_user_profile(&$vars) {
         'title' => $value['link'],
         'image' => $image,
         'description' => $value['call_to_action'],
-        'url' => dosomething_global_url($value['path_alias'], ['fragment' => 'prove']),
+        'url' => $GLOBALS['base_url'] . '/' . dosomething_global_get_prefix_for_language($language) . '/' . drupal_get_path_alias('node/' . $value['nid'], $language),
```
sergiitk: Not sure if using `$GLOBALS['base_url']` is correct. It's better to use `url()` function.

lib/themes/dosomething/paraneue_dosomething/includes/preprocess.inc
itsjoekent, line 668: (fixed)

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -651,6 +651,24 @@ function dosomething_global_url($path, $options = NULL) {
 }
 
 /**
+ * Build a global node URL that adds the prefix and a fragment
+ *
+ * @param string $nid
+ *    The nid of the node
```
sergiitk, line 657: Please end documentation sections with a period: https://www.drupal.org/node/1354#order.

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -651,6 +651,24 @@ function dosomething_global_url($path, $options = NULL) {
...
+ * @param string $language
+ *    The language of the URL
+ * @param string $fragment
+ *    Optional fragment to append
+ */
+function dosomething_global_node_url($nid, $language, $fragment = NULL) {
+  $raw_url = 'node/' . $nid;
+  if (isset($fragment)) {
+    $raw_url .= '#' . $fragment;
```
angaither, line 666: i know we've seen issue with manually concating the url like this before, it's best to pass this into url() with fragment as an option.
that will also allow for other types of urls to use this. (with a query string etc)

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -651,6 +651,24 @@ function dosomething_global_url($path, $options = NULL) {
...
+  }
+  return $GLOBALS['base_url'] . '/' . dosomething_global_get_prefix_for_language($language) . '/' . drupal_get_path_alias($raw_url, $language);
```
sergiitk, line 668: Same here, why not `url()` function?

lib/modules/dosomething/dosomething_global/dosomething_global.module
itsjoekent, line 666: I'm doing this because the url function has been the cause of every problem we've been seeing regarding bad URL's, it doesn't play well with languages and prefixes

this is the method I've been using for quite a while now in a few places and it's been working great

lib/modules/dosomething/dosomething_global/dosomething_global.module
itsjoekent, line 657: I think we have a lot of cleaning up to do then, a lot of these comment clocks have no periods haha

lib/modules/dosomething/dosomething_global/dosomething_global.module
itsjoekent, line 668: see here: https://github.com/DoSomething/phoenix/pull/5808#discussion_r45373947

lib/modules/dosomething/dosomething_global/dosomething_global.module
sergiitk, line 666: What's wrong exactly with `url()`?

itsjoekent, line 666: - I've caught it not properly apply the path alias on nodes (on global / translated titles)
- It often makes duplicate prefixes (like what the referenced bug here is describing)
- I've also had problems with it applying the correct language to a URL

This is why for many days now I've been manually building the URL with path_alias and get prefix, because it _always_ works fine without problem. Whereas url() consistently gives me problems and does not behave well with global functionality.

EDIT: and if you go through most of the current global code, from the node redirect logic to the new stuff I'm adding here, you'll notice none of it is using url() for this reason

sergiitk, line 666: Well, we're using `l()` function a lot. And `l()` uses `url()` under the hood. That's why I think for we need to make `url()` reliable as oppose to generate links by our own.
What do you think @angaither?

angaither, line 666: I agree, we've seen a lot of issues with manually making them
#5796 #5474 #5423

sergiitk:
:+1:

itsjoekent:
merging this now per sergi's thumbs up, going to take a deeper look at url later

PR #5809, opened 2015-11-19
Updating form output for school finder modal.

### LINKED ISSUE #5788: Extra Space in Between Copy and State Selector Dropdown Menu
The extra space appears on https://staging.dosomething.org/us/campaigns/teens-jeans (or any campaign with a school finder) like so:

image

weerd:
Fixes #5788
#### What's this PR do?

There was an additional form wrapper for elements that did not contian any markup or content but was still being output as an empty div with styled padding which was causing the extra spacing.

Quick :eyes: on the one file changed.

@DFurnes do you happen to know if something gets otherwise populated into the `elements` container in some other version of the school finder modal (if there is one) that I might not be accounting for?
#5788

---

@DFurnes 

cc: @angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
@@ -314,14 +314,6 @@ function dosomething_signup_user_signup_data_form($form, &$form_state, $signup) 
     '#suffix' => '</p></div>',
   );
 
-  // Container for form elements
-  $form['elements'] = array(
-    '#type' => 'container',
-    '#attributes' => array(
-      'class' => array('modal__block'),
-    ),
-  );
-
   // If we are collecting User address:
   if ($config['collect_user_address']) {
     dosomething_user_address_form_element($form['elements'], $form_state);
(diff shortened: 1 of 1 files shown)


DFurnes:
Hmmm, it looks like the sections on the modal [were intended to be placed in there](https://github.com/DoSomething/phoenix/commit/19915b0c9302d941e7a191a76875665ad0f4d249). I don't really remember why that was the way I set things up though. :sweat_smile: 

This seems reasonable, though it might be good to remove the following lines (like [this one](https://github.com/DoSomething/phoenix/blob/510b845faef5c57eb1751622d72bac090c46fa6b/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc#L319)) that reference `$form['elements']` since it won't exist anymore. It also looks like the t-shirt form used that container, but that's RIP :skull: so probably no worries there.

weerd:
Hmmm, good call. I was trying to see if there was a way to check if stuff was getting placed in there and if empty then unset it....?

DFurnes:
Yeah, ideally we only create the container if we need it. But it's strange that it's showing up as empty since it looks like we're putting things in there a few lines below. :confused:

weerd:
So the issue is that since it's building a Droooooooopal render array thing, it populates with some preliminary stuff like the `#attributes` and `#type` so there will technically _always_ be something in the array when it checks whether to output or not.

DFurnes:
:grimacing: :ship: :+1:

PR #5812, opened 2015-11-19
Turn of xml enpdoints in the api.

### LINKED ISSUE #4462: Consider disabling XML Response from Services Module
Not sure if the default XML response is used anywhere, but we could consider disabling it within the Services module if it is not used and we can designate the default to be JSON.

image

angaither:
You can't sit with us.

Fixes #4462


angaither:
lookatme! living-a-forkless-life!

weerd:
:tada: :+1:

PR #5813, opened 2015-11-19
Turn off optimizely on build.

### LINKED ISSUE #5811: disable optimizely halts ds build
@DeeZone just noticed this when I rebuilt

the command should have a `-y` with it 
`drush dis optimizely -y`

https://github.com/DoSomething/phoenix/pull/5798/files#diff-38f75dde0b5d5fcc10326bb75fab555dR473

angaither:
Yes, I really do mean it.
Fixes #5811

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -470,8 +470,8 @@ function pull_db {
       echo "Reverting features..."
       drush -y fra
 
-      echo "Disabling Optimizely module on staging."
-      drush dis optimizely -y
+      echo "Disabling Optimizely module locally."
+      drush -y dis optimizely
 
       echo "Clearing cache..."
       drush cc all
--- a/bin/ds
+++ b/bin/ds
@@ -471,7 +471,7 @@ function pull_db {
       drush -y fra
 
       echo "Disabling Optimizely module on staging."
-      drush dis optimizely
+      drush dis optimizely -y
 
       echo "Clearing cache..."
       drush cc all


sergiitk:
@angaither But this code is being executed on `pull stage`.
Did you intend it to?

angaither:
@sergii-tkachenko yes, just so we can turn off it off locally #1507

sergiitk:
@angaither Can you also `drush pmu` (`pm-uninstall`) it?

angaither:
@sergii-tkachenko sure, but why would we also need to uninstall?

sergiitk:
@angaither Because that's the correct way to turn modules off.
It will drop its schema and remove its variables. See `optimizely_uninstall()`.

bin/ds
```diff
@@ -471,7 +471,7 @@ function pull_db {
       drush -y fra
 
       echo "Disabling Optimizely module on staging."
```
sergiitk: Also, not "on staging", locally. This is the line that confused me initially, I thought it's intended to be disabled on staging.

angaither:
@sergii-tkachenko I don't think we need to uninstall it, it's just removing tables, all we want is the JS to not be executed locally

bin/ds
```diff
@@ -471,7 +471,7 @@ function pull_db {
...
-      drush dis optimizely
+      drush dis optimizely -y
```
sergiitk: It will be more consistent to pass `-y` as the first argument: this is a core-global-option. Not all drush submodules accept global options passed passed as the first argument.

sergiitk:
:+1:

PR #5815, opened 2015-11-19
Adding class to fix list styling in school finder help text.

### LINKED ISSUE #5789: Markdown Not Appearing For Tips In School Finder
Help text appears like this in the front end:
image

But should be numbered, as the content is numbered in the Do Something config.

Not sure if this is helpful/expected behavior, but markdown is also not possible in the "form copy" field. **Bolding** words does not work.
image

weerd:
Fixes #5789
#### What's this PR do?

Adds a quick class to the modal on the container that houses the select dropdown and help text for the school finder so that the formatting of the list items is numbered and styled correctly.

Just double-check (:heavy_check_mark: :heavy_check_mark:) the codez.

While this fixes the formatting, there is still an issue with the escaped apostrophe's and quotes. This is unrelated to markdown and more with how Drupal escapes those entities when the text is saved via the CMS. Might need to try entering those entities differently...
#5789
#### Screenshots

image

---

@DFurnes

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
@@ -278,6 +278,7 @@ function dosomething_signup_node_signup_data_form_submit(&$form, &$form_state) {
 function dosomething_signup_user_signup_data_form($form, &$form_state, $signup) {
   // Load signup data form config to determine what form elements we need.
   $config = dosomething_signup_get_signup_data_form_info($signup->nid);
+
   $form['sid'] = array(
     '#type' => 'hidden',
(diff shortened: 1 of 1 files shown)


DFurnes:
:rocket:!

DFurnes:
For the escaping issue, it looks like Drupal is escaping the content twice. For example, from the second list item the `&amp;quot;` should just be `&quot;`...strangeness.

``` html
Try abbreviating or unabbreviating. If &amp;quot;St. Mary&amp;#039;s&amp;quot; doesn&amp;#039;t work, try &amp;quot;Saint Mary&amp;#039;s&amp;quot; or vice versa. Same with JFK vs. John F. Kennedy.
```

weerd:
Yeah, good call! I'm taking a look at that next :)

PR #5820, opened 2015-11-19
Updating install file for dosomething_northstar module.

weerd:
Refs #5778
#### What's this PR do?

This PR updates the **dosomething_northstar.install** file to modify how the function to create the caching table is created. The `drupal_get_schema_unprocessed()` function relies on being passed the name of an existing table to grab the schema from it and copy to create a new table with the same schema. The previous code referred to a table that does not exist so the function was likely silently failing.

Just check the codez changes and make sure it looks correct :)

While this updates and successfully creates a new `cache_dosomething_northstar` table in the database, it might not really matter since we're using Redis for our caching. However, the redis module overriding caching to put all caching in Redis seems to be broken on **local** development so it is difficult to confirm. @sena and I have chatted a bit about this and need to confirm it is still properly working on our test and prod versions and need to troubleshoot why local is broken.
#5778 reference to caching related things

---

@angaither 
cc: @chloealee @sheyd

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.install
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.install
@@ -5,10 +5,22 @@
  */
 
 /**
+ * Create a new database table "cache_dosomething_northstar" caching.
  * Implements hook_schema().
  */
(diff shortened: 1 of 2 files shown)


angaither:
:+1:

PR #5822, opened 2015-11-20
Fixes node redirect after signup

### LINKED ISSUE #5792: When any non-US user creates an account and simultaneously signs up for a GL EN the URL changes to /us/
Recreating this consistently. When a Brazilian user is already logged in and signs up for a campaign, they remain on the GL EN URL.

http://screencast.com/t/QrrJw6wt

itsjoekent:
#### What's this PR do?

Fixes the form['redirect'] when a user signs up / logs in on a node.

As a global user, go to campaign, sign up. Are you still on the global node or did you get redirected to /us/ ?

Previous system for constructing the URL neglected language.

Fixes #5792 

cc: @namimody this is that bug the GB users are experiencing

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -691,7 +691,8 @@ function dosomething_user_authentication_submit($form, &$form_state) {
   }
   elseif (isset($nid)) {
     // Redirect to campaign node on reportback permalink pages.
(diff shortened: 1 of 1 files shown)


angaither:
one minor thing, this function calls other things that need the user object `dosomething_signup_user_signup` might be worth putting `global $user` above and passing it into functions that need it
:+1:

angaither:
:+1:

PR #5823, opened 2015-11-20
Adds session var to user redirect

itsjoekent:
Quick fix for login/logout going into a redirect loop for global users.

This is happening for the same reason campaign global nodes go into redirect loops (well used too) which is why the session fix also works for this.


itsjoekent:
@angaither can you +1 ?

sergiitk:
:+1:

PR #5828, opened 2015-11-20
Adds globe update functionality

itsjoekent:
#### What's this PR do?

Adds some code in dosomething settings that updates the lobby dashboard everytime someone hits the site. Why? FOR UPDATING THIS NIFTY GLOBE OF COURSE!

screen shot 2015-11-20 at 4 26 06 pm

It also adds in a new config variable to the dosomething_settings admin page. If this is blank the functionality will be disabled. Otherwise it contains the URL to ping for the lobby dashboard.

Deploy this (globe) branch to Thor. Set the setting in dosomething settings to http://lobby.dosomething.org/services/drupal

Go to http://lobby.dosomething.org/module/static and watch bubbles appear!

I think this is fabulous.

It's currently live on Thor at the time I'm writing this if you'd like to see it work. Otherwise follow the instructions above ^

### DIFF
--- a/lib/modules/dosomething/dosomething_settings/dosomething_settings.module
+++ b/lib/modules/dosomething/dosomething_settings/dosomething_settings.module
@@ -23,6 +23,25 @@ function dosomething_settings_menu() {
   return $items;
 }
 
+function dosomething_settings_init() {
+  // We don't need to redirect cli request, such as drush.
+  if (drupal_is_cli()) {
+    return;
+  }
+
+  $update_url = variable_get('dosomething_settings_realtimefeed_update_url');
+  if (!empty($update_url)) {
+    $data = ['code' => dosomething_settings_get_geo_country_code()];
+    $response = drupal_http_request($update_url, array(
+       'method' => 'POST',
+       'data' => json_encode($data),
+       'max_redirects' => 0,
+       'headers' => array('Content-Type' => 'application/json', 'Accept' => 'application/json'))
+     );
+  }
+
+}
+
 /**
  * Determines whether current site is an international affiliate.
  *
--- a/lib/modules/dosomething/dosomething_settings/dosomething_settings.module
+++ b/lib/modules/dosomething/dosomething_settings/dosomething_settings.module
@@ -24,7 +24,7 @@ function dosomething_settings_menu() {
 }
 
 function dosomething_settings_init() {
-  // We don't need to redirect cli request, such as drush.
+  // We don't need to track cli request, such as drush.
   if (drupal_is_cli()) {
     return;
   }
--- a/lib/modules/dosomething/dosomething_settings/dosomething_settings.module
+++ b/lib/modules/dosomething/dosomething_settings/dosomething_settings.module
@@ -23,25 +23,6 @@ function dosomething_settings_menu() {
   return $items;
 }
 
-function dosomething_settings_init() {
-  // We don't need to track cli request, such as drush.
-  if (drupal_is_cli()) {
-    return;
-  }
-
-  $update_url = variable_get('dosomething_settings_realtimefeed_update_url');
-  if (!empty($update_url)) {
-    $data = ['code' => dosomething_settings_get_geo_country_code()];
-    $response = drupal_http_request($update_url, array(
-       'method' => 'POST',
-       'data' => json_encode($data),
-       'max_redirects' => 0,
-       'headers' => array('Content-Type' => 'application/json', 'Accept' => 'application/json'))
-     );
-  }
-
-}
-
 /**
  * Determines whether current site is an international affiliate.
  *
--- a/lib/themes/dosomething/paraneue_dosomething/js/app.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/app.js
@@ -45,13 +45,14 @@ $(document).ready(function() {
    * it exists on the page, and short-circuiting if not.
    */
 
+  // Ping the dashboard with my location
+  Globe.init();
+
   // Initialize analytics.
   Analytics.init();
 
   // Initialize the donation form on the donate page.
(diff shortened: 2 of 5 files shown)


lib/modules/dosomething/dosomething_settings/dosomething_settings.module
```diff
@@ -23,6 +23,25 @@ function dosomething_settings_menu() {
   return $items;
 }
 
+function dosomething_settings_init() {
+  // We don't need to redirect cli request, such as drush.
```
sergiitk: Not to redirect in this case, `to track`.

itsjoekent:
can confirm that the globe also works with Ukraine :D

lib/modules/dosomething/dosomething_settings/dosomething_settings.module
itsjoekent: right copied this from global module, updating!

lib/modules/dosomething/dosomething_settings/dosomething_settings.module
```diff
@@ -23,6 +23,25 @@ function dosomething_settings_menu() {
...
+  if (drupal_is_cli()) {
+    return;
+  }
+
+  $update_url = variable_get('dosomething_settings_realtimefeed_update_url');
+  if (!empty($update_url)) {
+    $data = ['code' => dosomething_settings_get_geo_country_code()];
+    $response = drupal_http_request($update_url, array(
```
DFurnes: Will this cause a HTTP request to be made for every page load? If so, there might be performance concerns here... I think `drupal_http_request` is blocking.

itsjoekent: Oh it is? hmm. Is there anyway to async it?

weerd: :scream:

sergiitk:
This is all great, but I fear this could make Drupal REALY REALY slow. Also, if the remote server is down, each web request would be blocked for [30 sec](https://api.drupal.org/api/drupal/includes%21common.inc/function/drupal_http_request/7). This would kill all webheads.
Plus, this won't work behind varnish and Fastly.

I recommend to implement non-blocking JavaScript solution for this.

What do you think @angaither?
cc @mshmsh5000 @sheyd

itsjoekent:
non blocking JS could work, Ill try that

mshmsh5000:
Yes @sergii-tkachenko +1. The only way this can work is from the browser, so non-blocking JS is the way to go.

lib/themes/dosomething/paraneue_dosomething/js/app.js
```diff
@@ -50,6 +51,8 @@ $(document).ready(function() {
   // Initialize the donation form on the donate page.
   Donate.init();
 
+  Globe.init();
```
sergiitk: I think it'd be better to run this at the very beginning of the `$(document).ready()`, so the request has better change to pull through if user hits refresh or leaves the page.

DFurnes: Please add a comment explaining what this module does (see the other ones for examples).

lib/themes/dosomething/paraneue_dosomething/js/utilities/Globe.js
```diff
@@ -0,0 +1,12 @@
+import $ from 'jquery';
+import setting from './Setting';
+
+function init() {
+  const globeUpdateUrl = setting('dosomethingSetting.globeUrl');
```
sergiitk, line 5: Are you sure you want to use `const`? [IE ≥ 11](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Statements/const).

lib/themes/dosomething/paraneue_dosomething/js/utilities/Globe.js
```diff
@@ -0,0 +1,12 @@
...
+  if (globeUpdateUrl !== undefined) {
```
DFurnes, line 6: I think you also probably want to check that it's not an empty string here.

lib/themes/dosomething/paraneue_dosomething/js/utilities/Globe.js
DFurnes, line 5: @sergii-tkachenko It's fine, we're running all our scripts through [Babel](https://babeljs.io) to transpile to ES5.

sergiitk, line 5: Wow, awesome!

lib/themes/dosomething/paraneue_dosomething/js/utilities/Globe.js
sergiitk, line 6: @DFurnes Maybe like that?

``` js
const globeUpdateUrl = setting('dosomethingSetting.globeUrl', '');
if (globeUpdateUrl !== '') {
  ...
}
```

itsjoekent, line 6: I can add an empty string check for saftey but I actually need an undefined check because the preprocess logic only runs when the dosomething_setting variable is set. 

```
  $update_url = variable_get('dosomething_settings_realtimefeed_update_url');
  if (!empty($update_url)) {
    drupal_add_js(
```

So if I don't set the variable, refresh the page and prinout the globalUpdateUrl, it's actually undefined

screen shot 2015-11-23 at 9 38 03 am

DFurnes, line 6: Ah, gotcha. That should be safe then.

DFurnes:
Seems good to me. :+1:

PR #5829, opened 2015-11-20
Removing old code to request reportbacks query.

### LINKED ISSUE #4701: Replace dosomething_reportback_get_reportbacks_query_result()
Replace `dosomething_reportback_get_reportbacks_query_result()` with `dosomething_reportback_get_reportbacks_query()` to reduce number of parameters and keep them with single `$params` array. This will help with how the function is called in `Reportback` class instances and avoid the need for passing arbitrary params that make it less dynamic.

weerd:
Fixes #4701
#### What's this PR do?

This updates just wraps up the removal of the reportbacks query result function that was not as useful for the API because it required passing too many parameters, and the new funciton allows passing an array of parameters.

Just take a gander at the code and make sure it looks right.

If you so desire you can pull down the code and do a couple hits to the `/reportback.json` endpoint and make sure you're still getting data.

This code was set up ahead of time during the development of the API to allow removal of the old function, so shouldn't cause problems with this update.
#4701

---

@angaither 
cc: @jonuy

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -1098,41 +1098,29 @@ function dosomething_reportback_build_reportbacks_query($params = array()) {
 /**
  * Returns Reportbacks query result to loop through.
  *
- * @param array $params
+ * @param  array  $params
  *   An associative array of conditions to filter by.
- *   @see dosomething_reportback_build_reportbacks_query()
- * @param int $count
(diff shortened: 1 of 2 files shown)


angaither:
seeing one more call to [this function](https://github.com/DoSomething/phoenix/blob/d97bf592f59d917db51277f4543255628dd0d2d4/lib/modules/dosomething/dosomething_reportback/includes/Reportback.php#L90)

weerd:
Fixing! Good catch :)

angaither:
:+1:

PR #5831, opened 2015-11-23
Makes sure we dont add a slash if the prefix ix blank

itsjoekent:
Also cleans up the node redirect logic to the same standard as everything else. I don't think we have an issue for this and it's blocking deploy.

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -614,7 +614,7 @@ function dosomething_global_redirect_node($user_lang, $node, $redirect_status) {
     drupal_exit();
     return;
   }
-  
+
   $_SESSION['dosomething_global_redirected'] = TRUE;
   drupal_goto(dosomething_global_node_url($node->nid, $user_lang));
 }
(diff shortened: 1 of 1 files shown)


itsjoekent:
please note i have no idea why style ci is failing for 23 files not touched by this pull request, they should probably fix that

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -666,7 +663,11 @@ function dosomething_global_node_url($nid, $language, $fragment = NULL) {
   if (isset($fragment)) {
     $raw_url .= '#' . $fragment;
   }
-  return $GLOBALS['base_url'] . '/' . dosomething_global_get_prefix_for_language($language) . '/' . drupal_get_path_alias($raw_url, $language);
+  $alias = dosomething_global_get_prefix_for_language($language);
```
sergiitk: Please rename this variable, `$alias` is incorrect name and very confusing. This should be named `$prefix`.
Actual alias is returned by `drupal_get_path_alias()`.

sergiitk:
Besides one code style note, excellent! Please fix the var name and it's good to go :+1:

itsjoekent:
yeah im a doof, i was was playing with all of them as variables and accidentally left that one. WOOPS, fixing now

sergiitk:
:+1:

PR #5834, opened 2015-11-23
Reportback module cleanup

weerd:
#### What's this PR do?

This PR updates the function name for reportback-item query builder, and also separates out the **reportback** and **reportback_item** query functions out from the **dosomething_reportback.module** into their own separate files that get included at the top of the module.

Wherever they fancy, but probably starting at the **dosomething_reportback.module** file would be good.

Feel free to run some local API requests to `/reportbacks` and `/reportback-items` endpoints and make sure you get data.

The **dosomething_reportback.module** file is ginormous and getting a bit out of hand to work with. It would help to break it up a bit into logical files that can be included into the module and this is an attempt at starting that process.

---

@angaither 
cc: @DFurnes


lib/modules/dosomething/dosomething_reportback/dosomething_reportback_item.query.inc
```diff
@@ -0,0 +1,134 @@
...
+  $query->orderRandom();
+  }
+  elseif (isset($params['order'])) {
+  $query->orderBy('rb.updated', $params['order']);
+  }
+
+  else {
+  $query->orderBy('rb.updated', 'DESC');
+  }
+
```
angaither, line 99: looks like spacing is off in this func

weerd:
@angaither yeah, spacing was off on most of that file. I pasted the code in before adding `<?php` at the top of the file and it threw the rest off :(

angaither:
:+1:

PR #5835, opened 2015-11-23
Fixes infobar for global users

### LINKED ISSUE #5832: zendesk form not appearing on GL campaigns
action page on gl english campaigns are not displaying the zendesk form
https://www.dosomething.org/campaigns/veggie-surprise?noredirect=true

image

itsjoekent:
#### What's this PR do?

Switches out prefix usage for country code in the info bar preprocess 

Do global users now have the zendesk help?

The logic was not executing for global users because a global user has no prefix, and thats what this logic was looking for.

Fixes #5832

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/system/partials/info-bar.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/system/partials/info-bar.tpl.php
@@ -33,4 +33,3 @@
   </div>
 </footer>
 <?php endif; ?>
-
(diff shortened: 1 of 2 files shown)


lib/themes/dosomething/paraneue_dosomething/templates/system/partials/info-bar.tpl.php
```diff
@@ -33,4 +33,3 @@
   </div>
 </footer>
 <?php endif; ?>
-
```
itsjoekent, line 36: not sure why this occurred. do i need to put it back?

sergiitk:
:+1:

angaither:
@deadlybutter i don't think this will work, we aren't talking about users who don't have a prefix, we want them to see the zendesk form, not a specific country email address

angaither:
maybe the ticket was unclear but this was for people with the base url, coming from non-us, non br, or mx

itsjoekent:
so this fix works fine, but theres some extra logic which doesnt have to be there, making another PR

PR #5841, opened 2015-11-24
tracking reportback errors

### LINKED ISSUE #4027: Monitoring coverage for reportback error states
@mikefantini brought up the great point that it'd be awesome to be able to monitor how often users are hitting error states, to see if we can improve the UX.

Could you figure out how to monitor the following things using StatHat please @weerd, and add the coverage to the campaigns StatHat dashboard that's visible to everyone? It'd be great to be able to see graphs of how many users are hitting these errors states each day.
- [ ] User uploads file that's too small in image dimensions
- [ ] User uploads file that's too big in file size
- [ ] User enters a caption that's too long
- [ ] please upload an image error (#4427)

katiecrane:
fixes #4027 

image file size too big + image dimensions too small:
-these are in ImageUpload.js because that is where those checks actually occur (aka before the form is submitted)
-the calls to ga are definitely in the right places, but I'm not 100% sure the parameters are what we want them to be

no image uploaded:
-this is in dosomething_reportback.forms.inc because we only check for that when the user clicks submit, so there is no way to check for this in the js file
-less certain about the php syntax for ga, but this seems to follow what I could find online
-I am not sure if drupal gets mad at this or not, as I can't get it to throw this error on my machine (when I click submit without an image, it just scrolls up?)

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -372,7 +372,9 @@ function dosomething_reportback_form_validate_file($form, &$form_state) {
 
   // If this is a new reportback form, file is mandatory.
   elseif ($form_state['values']['rbid'] == 0) {
-    ga('send', 'event', 'Reportback', 'no image uploaded');
+    $ga = new Elements_Tools_Serversideanalytics();
+    $ga->setEvent('send', 'event', 'Reportback', 'no image uploaded');
+    $ga->createEvent();
     form_set_error('reportback_item', t('Please upload an image.'));
   }
 }
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -370,9 +370,7 @@ function dosomething_reportback_form_validate_file($form, &$form_state) {
   }
   // If this is a new reportback form, file is mandatory.
   elseif ($form_state['values']['rbid'] == 0) {
-    $ga = new Elements_Tools_Serversideanalytics();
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -370,6 +370,9 @@ function dosomething_reportback_form_validate_file($form, &$form_state) {
   }
   // If this is a new reportback form, file is mandatory.
   elseif ($form_state['values']['rbid'] == 0) {
+    $ga = new Elements_Tools_Serversideanalytics();
```
DFurnes: I'm not sure this class exists without installing some extra packages... we have a [custom helper function](https://github.com/DoSomething/phoenix/blob/44e8094e68ef24dcfe9c4e8875c68960ab52578f/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module#L527-L545) that you can use to inject a analytics event on the following page load.

lib/themes/dosomething/paraneue_dosomething/js/images/ImageUpload.js
```diff
@@ -75,6 +75,7 @@ define(function(require) {
       }
 
       if (!_this.validFileSize(files)) {
+        ga('send', 'event', 'Reportback', 'image file too large');
```
DFurnes: Just to be safe, I'd recommend wrapping this in a quick check that the `ga` object exists:

``` js
if (typeof ga !== 'undefined' && ga !== null) {
  // your analytics event
}
```

angaither:
@katiecrane see [`dosomething_helpers_add_analytics_event`](https://github.com/DoSomething/phoenix/blob/44e8094e68ef24dcfe9c4e8875c68960ab52578f/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module#L535) for adding events in php

PR #5842, opened 2015-11-25
Applied fixes from StyleCI

angaither:
This pull request applies code style fixes from an analysis carried out by [StyleCI](https://styleci.io).

---

For more information, click [here](https://styleci.io/analyses/87eK6q).


angaither:
current config

```
preset: psr1

enabled:
  - whitespacy_lines
  - trailing_spaces
  - ternary_spaces
  - spaces_cast
  - single_quote
  - elseif
  - extra_empty_lines
  - concat_with_spaces
  - operators_spaces
  - php_closing_tag
  - standardize_not_equal
  - align_double_arrow
  - eof_ending
  - function_declaration
  - function_call_space
```

config/default.settings.local.php
```diff
@@ -13,18 +13,18 @@
  */
 
 // Database settings.
-$databases = array (
+$databases = array(
   'default' =>
-    array (
+    array(
```
DFurnes, line 18: Should we add the `short_array_syntax` fixer?

angaither, line 18: I'm conflicted on that one. _Technically_ it [isn't a drupal standard](https://www.drupal.org/coding-standards#array) but we have been using it...

angaither, line 18: it is a lot prettier...

config/default.settings.local.php
```diff
@@ -13,18 +13,18 @@
...
-        array (
+        array(
           'database' => 'dosomething',
           'username' => 'root',
           'password' => '',
-          'host' => 'localhost',
-          'port' => '',
-          'driver' => 'mysql',
-          'prefix' => '',
+          'host'     => 'localhost',
```
DFurnes, line 24: I'm not super enthusiastic about the `align_double_arrows` fixer, since it means potentially mangling diffs by having to re-align other array items if you add a key that "pushes" the other arrows over.

lib/themes/dosomething/paraneue_dosomething/templates/fact/node--fact_page.tpl.php
```diff
@@ -81,7 +81,7 @@
             <h4 class="js-footnote-toggle"><?php print t('Sources'); ?></h4>
             <ul class="js-footnote-hidden">
               <?php foreach ($sources as $key => $source): ?>
-                <li><sup><?php print ($key + 1); ?></sup> <?php print $source; ?></li>
+                <li><sup><?php print($key + 1); ?></sup> <?php print $source; ?></li>
```
DFurnes, line 84: Hmm, this isn't really a comment on the fixer, but it could be worthwhile to standardize on whether we want to use parentheses for language constructs like `print` or `echo`. It's a little weird seeing it both ways on one line.

config/default.settings.local.php
itsjoekent, line 24: yeah this could get annoying

config/default.settings.local.php
DFurnes, line 18: I think it's worth adding it since it improves readability by a ton. A lot of our more recent code has been using short-array syntax as well so it's kinda all over the place right now.

angaither, line 18: okay!

Issue #5554, opened 2015-10-20
Bump Vagrant box version to 1.0.6


mshmsh5000:
https://atlas.hashicorp.com/dosomething/boxes/phoenix

Need to load newest Vagrant box.

angaither:
closed with #5555

sergiitk:
No, it hasn't been merged yet.

angaither:
surely this has been merged by meow.

mshmsh5000:
Yeah @angaither our Vagrant box is version like 6000 now. SIX THOUSAND

PR #5845, opened 2015-11-30
Updating vagrant box version to use latest released

weerd:
#### What's this PR do?

This PR updates the version of the `dosomething/phoenix` vagrant box to use the latest released version.

Just take a :eyes: on quick change.

I was not sure if there was a specific reason why this wasn't updated. I had tried to `vagrant up` with `v1.0.5` and was running into some weird errors when trying to download the box.

---

@sergii-tkachenko 
cc: @mshmsh5000 @angaither

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -2,7 +2,7 @@ Vagrant.configure("2") do |config|
 
   ## Choose your base box
   config.vm.box = "dosomething/phoenix"
-  config.vm.box_version = "1.0.5"
+  config.vm.box_version = "1.0.6"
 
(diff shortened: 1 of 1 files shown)


weerd:
I'll wait to make the StyleCI changes if it turns out that we do want to bump the version to `1.0.6` cause those are a lot of style updates :\

mshmsh5000:
This duplicates #5554, eh?

weerd:
Oh, haha. :) I will close this then!

sergiitk:
@weerd I'll retest Matt's PR today.

weerd:
Sounds good!

PR #5851, opened 2015-11-30
Updating translation file directories.

weerd:
#### What's this PR do?

Moving pots into a translations directory and adding in the POEditor .po files latest exports for backup.

Just take a :eyes: at the general changes.

---

@angaither


angaither:
:+1:

PR #5850, opened 2015-11-30
Applies bounce rate fix to international branch

sergiitk:
Closes #5846.

PR #5156, opened 2015-09-18
Adds new functions to read fastly geo headers 

itsjoekent:
#### What's this PR do?
- Adds new functions to dosomething settings (outlined in more detail [here](https://github.com/DoSomething/phoenix/issues/4959#issuecomment-140189850) ) to read the Fastly headers 
- Updates the affiliate logic to prefer these headers over the old logic per Dee's request https://github.com/DoSomething/phoenix/issues/4959#issuecomment-135904328
#### Where should the reviewer start?

dosomething_settings
#### How should this be manually tested?

Tricky one. Follow this guide to get setup https://github.com/DoSomething/phoenix/wiki/Testing-global-functionality-on-your-local-environment

Then to get an output of all the functions, you can add this to the top of dosomething_settings temporarily to test,

```
function dosomething_settings_init() {
  dpm($_SERVER);
  dpm(dosomething_settings_get_geo_country_code());
  dpm(dosomething_settings_get_geo_city());
  dpm(dosomething_settings_get_geo_country_name());
  dpm(dosomething_settings_get_geo_latlong());
  dpm(dosomething_settings_get_geo_region());
  dpm(dosomething_settings_get_affiliate_country_code());
}
```
#### Any background context you want to provide?

drupal_get_http_header was giving me issues, $_SERVER worked fine so I went with that
#### What are the relevant tickets?

Fixes #4959 

@angaither

PR #5852, opened 2015-11-30
Introduces geo headers functions to international sites

sergiitk:
We need functions from #5156 for #5850 patch to work.

cc @DFurnes


angaither:
does international stuff run different? do we need to set up fastly things?

sergiitk:
I just closed this because @DFurnes, @sheyd and I just decided to go the other way.

sergiitk:
The other way = #5853.

PR #5853, opened 2015-11-30
Use old international implementation of country codes.

DFurnes:
#### Changes

Rather than using the new Fastly country code function, which doesn't exist on international (and wouldn't make sense there), we'll just use the old "multisite" version that uses the country setting.

For review: @sergii-tkachenko 
/cc @sheyd


sergiitk:
:+1:

sheyd:
:panda_face:

sergiitk:
cc @angaither

angaither:
ahh, makes more sense. I was wondering what _the other way_ was

sergiitk:
@angaither Yes, sorry, we were discussing in Hangout this change at the moment I posted https://github.com/DoSomething/phoenix/pull/5852#issuecomment-160712701.

PR #5857, opened 2015-12-01
Adds hotmodule toggle to custom settings

### LINKED ISSUE #5827: Add toggle to enable/disable Hot Module per campaign
Currently, the hot module is switched on when the current date is between the high season dates set on the back end of a campaign. 

While we will be using the hot module more ubiquitously going forward, there are instances where we would like to set a hot season to get the perks of inputting a high season (like appearing higher in the finder) but without showing the module (because the campaign is a more sensitive rapid response, for example).

Just wanted to create an issue as I know we briefly discussed this possibility and @angaither said it was doable :)

itsjoekent:
#### What's this PR do?

Adds a custom toggle switch to the custom settings of a node under "goals", that is by default off.

If this switch is enabled & the high season is set the hot module & win module will display. 

Variable created in dosomething_helpers.variables.inc -- after that its used in the dosomething_campaign module.

Set the high season of a campaign to end sometime in the coming days. Verify the hotmodule does not display.
Then go to the campaign custom settings and enable the hotmodule.
Does the hotmodule display now?

Next set the end season date to a few days ago, does the win module display?
Disable the hot module in custom settings, does the win module still display?

Everything is covered in the issue comments

Fixes #5827

screen shot 2015-12-01 at 2 19 14 pm

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -759,7 +759,8 @@ function dosomething_campaign_add_shipment_form_vars($node) {
  * @return bool
  */
 function dosomething_campaign_is_hot_module($vars) {
-  if (isset($vars['field_high_season'])) {
+  $hot_module_enabled = dosomething_helpers_get_variable('node', $vars['nid'], 'hot_module_enabled');
+  if (isset($vars['field_high_season']) && $hot_module_enabled) {
     $today = dosomething_campaign_set_est_timezone(new DateTime());
     $start_date = dosomething_campaign_set_est_timezone(new DateTime($vars['field_high_season'][0]['value']), TRUE);
     $end_date = dosomething_campaign_set_est_timezone(new DateTime($vars['field_high_season'][0]['value2']), TRUE, TRUE);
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -759,7 +759,8 @@ function dosomething_campaign_add_shipment_form_vars($node) {
  * @return bool
  */
 function dosomething_campaign_is_hot_module($vars) {
-  if (isset($vars['field_high_season'])) {
+  $hot_module_enabled = dosomething_helpers_get_variable('node', $vars['nid'], 'hot_module_enabled');
```
angaither, line 762: should fail out of this function `if (!$hot_module_enabled)`

itsjoekent, line 762: the IF block is under that line

itsjoekent, line 762: https://github.com/DoSomething/phoenix/pull/5857/files#diff-ae398bb8864fc3c26317f7ccaef04c0aR866

angaither, line 762:
```
if (!$hot_module_enabled) { return false; } 
if (isset($vars['field_high_season....
```

itsjoekent, line 762: I'm not sure I see the difference? is using && not allowed anymore with our code style?

angaither, line 762: it's allowed, but it doesn't make sense to go through all the date checks if hot module isn't enabled on that campaign

itsjoekent, line 762: it's not going through all of the date checks? its just checking if both fields exist 

```
if (isset($vars['field_high_season']) && $hot_module_enabled) {
```

weerd, line 762: I think that seems right, unless I'm missing something... although the encompassing function name is weird to me, haha.

itsjoekent, line 762: from slack w/ @angaither 

> reverse the order of the if and :ok_hand: 

https://dosomething.slack.com/archives/D03T8S9QY/p1449003349000023

itsjoekent, line 762: reversed, everything seems good

Issue #4904, opened 2015-08-25
Make 54 Language-neutral campaigns English


sergiitk:
This will fix #4897 and bring more consistency to the Thor.
We can do it programmatically (will take about 3 hours) or assign somebody to go through it manually.

![image](https://cloud.githubusercontent.com/assets/672669/9473358/02ef1ad6-4b63-11e5-954b-fe379dc37c38.png)

cc @sheydari-ds @mikefantini @julielorch @namimody

sergiitk:
[7:55 PM] Sergii Tkachenko: as i understand, the field display depends on users’s language
[7:56 PM] Sergii Tkachenko: and when the language is explicitly set to English, it looks for a English “translation” of the image field
[7:56 PM] Sergii Tkachenko: i don’t think this campaigns are suppose to be lanugage-neutral in the first place
[7:57 PM] Sergii Tkachenko: can we clarify that?
[7:57 PM] Sena Heydari: Yeah, let’s ask that during stand-up. It does raise the question of if we should have a “Fallback” image though.

mikefantini:
@angaither @aaronschachter i assume we're good to just move everything set to English instead of having some in English some in Language neutral but just wanted to see if you could think of any gotcha's in doing this?

angaither:
I think we are ok to set things to english. my only concern would be all the things we have set to check `[und]` (undefined language) in the code. some classes and checking for variables _might_ break

https://github.com/DoSomething/phoenix/search?utf8=%E2%9C%93&q=und

sergiitk:
@angaither Good catch. However, this is inevitable since most of campaign are English, not neutral already.

sergiitk:
![image](https://cloud.githubusercontent.com/assets/672669/9477015/ac8871c2-4b78-11e5-942e-a03f79224f46.png)

sergiitk:
CC @DFurnes: see last 3 comments.

angaither:
@sergii-tkachenko what will `[und]` be when the language is set? is there a way to scope it for all languages?

sergiitk:
It depends. If that's field's class and field is not translatable, it'll stay `[und]`. If translatable: I suppose it will be set to current language.

sergiitk:
We need to work out a way to use different language-independent CSS class names.

DFurnes:
[...]

PR #5861, opened 2015-12-01
Extends `campaign_signup` and `campaign_reportback` payloads with campaign language

### LINKED ISSUE #5685: campaign_language in Campaign Signup and Reportback Transactions
When a user signs up (`campaign_signup`) for a campaign or reports back (`campaign_reportback`) a request for a transactional message is sent to the Message Broker system. The payload of the message request needs to include the `campaign_language`. The value comes from the language setting for the related campaign node.

**NOTE**: This is different than the `user_language` and `user_country` which are currently in the request payload but does not address this issue.

**Example payload**:

```
(
    [activity] => campaign_signup
    [email] => mfantini+pttranslator@dosomething.org
    [uid] => 3202400
    [merge_vars] => Array
        (
            [MEMBER_COUNT] => 4.3 million
            [FNAME] => Brazil
            [CAMPAIGN_TITLE] => Sea Turtle Street Sweep
            [CAMPAIGN_LINK] => https://thor.dosomething.org/us/campaigns/sea-turtle-street-sweep?source=node/3624
            [CALL_TO_ACTION] => Save sea turtles by picking up trash before it reaches the ocean.
            [STEP_ONE] => Clean Up Trash!
            [STEP_TWO] => Snap a Pic
            [STEP_THREE] => Haul It Away
        )

    [user_country] => US
    [user_language] => en
    [subscribed] => 1
    [event_id] => 3624
    [email_tags] => Array
        (
            [0] => 3624
            [1] => drupal_campaign_signup
        )

    [activity_timestamp] => 1446224947
    [application_id] => US
)
```

**Related Issues**:
- https://github.com/DoSomething/phoenix/issues/5601
- https://github.com/DoSomething/phoenix/issues/5804

sergiitk:
#### What's this PR do?
- Extends `campaign_signup` MB payload with campaign language.
- Extends `campaign_reportback` MB payload with campaign language.
#### Results

I made Thumb Wars English Global for testing. Here's the results:
1. Signup
   screen shot 2015-12-02 at 12 06 51 am 1
2. Reportback
   screen shot 2015-12-02 at 12 15 30 am

Closes #5685.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -857,6 +857,10 @@ function dosomething_reportback_mbp_request($entity) {
       'impact_noun' => $entity->noun,
       'image_markup' => $image_markup,
     );
+    $node = node_load($entity->nid);
+    if ($node) {
+      $params['campaign_language'] = $node->language;
+    }
     if (module_exists('dosomething_mbp')) {
       dosomething_mbp_request('campaign_reportback', $params);
     }
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -585,6 +585,7 @@ function dosomething_signup_get_mbp_params($account, $node, $opt_in) {
     'campaign_title' => $title,
     'campaign_link' => dosomething_global_url('node/' . $node->nid, $url_options),
     'user_language' => $language,
+    'campaign_language' => $node->language,
   );
 
   // Don't subscribe 26+ yo users for Mobile Commons.


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -857,6 +857,10 @@ function dosomething_reportback_mbp_request($entity) {
       'impact_noun' => $entity->noun,
       'image_markup' => $image_markup,
     );
+    $node = node_load($entity->nid);
```
deezone, line 860: I've been told to avoid `node_load()` due to the weight of using the function. Would it be less load on the system if the Reportback entity had the campaign language property that could be accessed in the transactional request?

sergiitk, line 860: You've been told wrong. Drupal will load it from node cache, which makes this operation lightning fast.

deezone:
If the language is not defined for the campaign does the default site language get sent in the Message Broker request? Does there need to be a default value?

sergiitk:
The language is one of node's primary attributes and it's always there. However, there's a "language-neutral" language, index `und`. We made sure there's no campaigns with language-neutral language, see #4904.

deezone:
:+1:

PR #5864, opened 2015-12-02
Reportback item status params update

### LINKED ISSUE #5653: Reportback Items API: status property is null for all Reportback Items
screen shot 2015-10-28 at 9 14 35 am

weerd:
Fixes #5653
#### What's this PR do?

This PR provides a level of data security based on user permissions. Prior to this PR, all (or some) status types for Reportback Items could be requested by passing a URL parameter `?status=pending,flagged,excluded,promoted,approved` by an anonymous `GET` request. 

This PR updates the API so that if an anonymous user is requested restricted statuses of `pending`, `flagged` or `excluded`, those types get filtered out and the request responds with no items and an error response or just providing `promoted` and/or `approved` items (depending if those were also requested). Now only admin users have access to request those status types.

This PR also updates the default `status` value to be both `promoted` and `approved`.

The main logic happens in **dosomething_reportback_item.query.inc**

If you want, pull down the branch and then hit the `/reportback_items` endpoint and pass some restricted statuses in query URL parameters while in incognito browser mode and see if any results show up that shouldn't.

---

@aaronschachter 

cc: @angaither @jonuy

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
@@ -49,7 +49,7 @@ function _reportback_item_resource_definition() {
             'optional' => TRUE,
             'type' => 'string',
             'source' => array('param' => 'status'),
-            'default value' => 'approved',
+            'default value' => 'promoted,approved',
           ),
           array(
             'name' => 'count',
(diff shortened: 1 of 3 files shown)


aaronschachter:
Looks ok to me :elephant: :+1: :iphone:

PR #5867, opened 2015-12-02
Turn off language fallback

### LINKED ISSUE #5556: Blank translated fields pull in from other translations

angaither:
This option was hidden in the settings, it seems to work.
image

Fixes #5556


sergiitk:
Good to know! :+1:

PR #5869, opened 2015-12-02
Make additional fields translatable on the campaign node

### LINKED ISSUE #5849: Make additional fields translatable on the campaign template
We've run into a few instances in which we wanted to create translations for campaigns (GL EN or otherwise) but couldn't because certain fields weren't translatable. It seems like this was mainly caused by:
1. campaign had a scholarship
2. campaign had a file in the Plan It Section in English
3. campaign was in high season (and that content is not ready for POEditor)

So because of this, it seems like we need to make the following fields translatable:
- [ ] field_scholarship_amount
- [ ] field_official_rules
- [ ] field_downloads
- [ ] field_high_season
- [ ] field_display_end_date

cc: @namimody

angaither:
Fields changed to translatable
- [x] field_scholarship_amount
- [x] field_official_rules
- [x] field_downloads
- [x] field_high_season
- [x] field_display_end_date

Fixes #5849

@sergii-tkachenko will need to run the drush command to fix this on thor/prod ;)


sergiitk:
:+1:

sergiitk:
@angaither did you check if the fields are displayed properly?

PR #5875, opened 2015-12-02
Removes install.php after ds build.

sergiitk:
#### What's this PR do?
- Removes install.php after ds build

Recent Drupal core Critical PSA recommends removing install.php after installation: https://www.drupal.org/PSA-2015-001


sheyd:
:+1:

PR #5882, opened 2015-12-03
Updates Apache Solr Search to 1.8.

sergiitk:
#### What's this PR do?
- Updates Apache Solr Search to 1.8.

Advisory ID: DRUPAL-SA-CONTRIB-2015-170
https://www.drupal.org/node/2627580


sergiitk:
I tested updated Solr module with Drupal, seems to work well. We need to figure out thought if the content needs to be reindexed.

sheyd:
:100: Yeah, we should after we deploy this, just to be sure. Doesn't take too terribly long anyways.

PR #5892, opened 2015-12-03
Creates a helper function to make more campaign fields translatable.

sergiitk:
#### What's this PR do?
- Creates a helper function to make more campaign fields introduced in #5869 translatable
#### Usage
- Make sure to run as `dosomething` user
- `cd /var/www/www.dosomething.org/current/`
- `ds-global make-more-campaign-fields-translatable`


angaither:
:+1:

Issue #4133, opened 2015-03-06
Update reportback permissions lingo


angaither:
`view any reportback` is used for admin view any reportback, it's confusing, rename it to `review any reportback`

since we now have this permalink business that sort of acts as a reportback.

PR #5895, opened 2015-12-04
Api reportback flagged parameter

### LINKED ISSUE #5879: Do not return flagged Reportbacks to anonymous user in API response
Similar to the filtering down in PR #5864 to stop RB items with statuses of `pending`, `flagged`, `excluded` from being return with an anonymous request to the API response, we want to black Reportbacks with a `flagged` property of `true`. Only admins should be able to request those.

cc @aaronschachter

weerd:
Fixes #5879
#### What's this PR do?

This PR updates the `/reportbacks` endpoint so that flagged reportbacks are not returned in any anonymous `GET` requests to the API. It also adds a new `flagged` URL parameter to the `/reportbacks` endpoint so that if an admin wants to retrieve both flagged and not flagged reportbacks, they can pass `flagged=true` as a URL parameter in the request (can also set it to `false` but that behaves as the default of not passing the parameter).

Start at **reportback_resource.inc** where the new URL parameter is added to Services module.

Pull down the branch, make sure to `drush cc all` to clear the cache and have Services register the new `flagged` URL parameter. Then hit a few `/reportback` endpoints as both an admin/anonymous user, passing `?flagged=true` or `?flagged=false` or omitting it entirely and confirm expected results!
#5879

---

@aaronschachter @DFurnes 

cc: @jonuy @angaither @chloealee

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
@@ -55,6 +55,15 @@ function dosomething_reportback_build_reportbacks_query($params = array()) {
     }
   }
 
+  if (!isset($params['flagged'])) {
+    $query->condition('rb.flagged', 0);
+  }
+  else {
+    if (!$params['flagged']) {
+      $query->condition('rb.flagged', 0);
+    }
+  }
+
   // Public API properties to expose:
   $rb_fields = array('rbid', 'created', 'updated', 'quantity', 'uid');
 
@@ -71,6 +80,7 @@ function dosomething_reportback_build_reportbacks_query($params = array()) {
 
   $query->groupBy('rb.rbid');
 
+  // @TODO: After campaign gallery changes, update this condition to: isset($params['random']) && $params['random']
   if (isset($params['random'])) {
     $query->orderRandom();
   }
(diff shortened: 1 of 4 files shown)


weerd:
@DFurnes huzzah!

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
```diff
@@ -111,3 +124,23 @@ function dosomething_reportback_get_reportbacks_query($params) {
...
 
+
+/**
+ * Filter out requested flagged Reportbacks based on user permissions.
+ *
+ * @param  array  $params
+ * @return mixed
+ */
+function dosomething_reportback_filter_flagged_reportbacks($params) {
+  if (user_access('view any reportback')) {
```
angaither, line 135: bah, this one is so confusing without #4133

angaither:
:+1:

DFurnes:
:1234: :computer: :cactus:

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
weerd, line 135: Yeah, I know. That one needs to happen. It kinda makes sense, but not entirely :\

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
```diff
@@ -55,6 +55,15 @@ function dosomething_reportback_build_reportbacks_query($params = array()) {
     }
   }
 
+  if (!isset($params['flagged'])) {
```
jonuy, line 58: This block looks like it could be simplified to

```
if (!isset($params['flagged']) || !params['flagged']) {
  $query->condition('rb.flagged', 0);
}
```

weerd, line 58: Yeah, I considered that too but in practice it doesn't work :(

weerd, line 58: The issue is that the `flagged` param is sometimes passed and sometimes not (like in the case for viewing a specific RB where you just pass an id; so the `rbid` is the only param in the array).

The issue is that as an admin, you can pass `true` OR `false`... in that scenario `isset()` always evaluates to true, since there is a value set... and the condition runs (even if `false` is passed).

PR #5896, opened 2015-12-04
Fixing a few DBLog related notices.

weerd:
#### What's this PR do?

@joe saw a bunch of PHP notices being thrown and alerted me to issues with the query builder functions for RB and RB items. This PR fixes a few issues with how I was using my own `dosomething_helpers_isset()` function @_@ Since passing an array to the function, need to separate the key we're looking for as the second argument passed to the function.

image

Take a close :mag_right: look at it to make sure I didn't make any more dumb mistakes...

---

@DFurnes @deadlybutter

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
@@ -112,8 +112,8 @@ function dosomething_reportback_get_reportbacks_query($params) {
   $params = dosomething_reportback_filter_flagged_reportbacks($params);
 
   $query = dosomething_reportback_build_reportbacks_query($params);
-  $offset = dosomething_helpers_isset($params['offset'], NULL, 0);
-  $count = dosomething_helpers_isset($params['count'], NULL, 25);
+  $offset = dosomething_helpers_isset($params, 'offset', 0);
+  $count = dosomething_helpers_isset($params, 'count', 25);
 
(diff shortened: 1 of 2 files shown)


DFurnes:
:+1: from me!

DFurnes:
@weerd One follow-up thought/suggestion...  It might help to rename this method to `dosomething_helpers_array_get` or something similar (inspired by Laravel's [`array_get`](http://laravel.com/docs/5.1/helpers#method-array-get) helper). I feel like the `isset` naming might tend to "trick" you into using it like the built-in language construct.

weerd:
Potentially, but it's meant for both single values, arrays or objects... It's really just meant as an `isset()` replacement and to remove the need for crazy ternaries each time you want to get a specific variable... could maybe separate out into 2 functions, but that might be overkill? Not sure.

DFurnes:
Hmm, I still worry that having the same name but a different usage than `isset` could be confusing, but it's not something I feel super strongly about. Let's just keep an eye on if we keep tripping up with it.

weerd:
It's a good point. I'd be down for renaming it to something that makes more sense, it's just the `_array_get()` wasn't really telling the full story ;)

DFurnes:
Yeah, I totally missed that part. Perhaps `dosomething_helpers_get_value`?

weerd:
Oh! I could be down w/ that name :) I'll make a issue for myself.

PR #5902, opened 2015-12-07
Fixes galleries on campaign collections.

### LINKED ISSUE #5901: Gallery isn't showing up on campaign collection nodes
https://www.dosomething.org/node/4965/edit -> https://www.dosomething.org/volunteer/sports

I believe this is what's causing the random white space at the bottom of the page as well:

image

sergiitk:
#### What's this PR do?
- Fixes galleries on campaign collections.

Since #5605 galleries on campaign collections are pre-rendered with `dosomething_static_content_preprocess_galleries($vars);` function that returns markup, but `node--campaign_group.tpl.php` still tries to render it by itself as an array of field elements.
I replaced template with [this code](https://github.com/DoSomething/phoenix/blob/9066d50c878dac21db311f392012ac9429814cd5/lib/modules/dosomething/dosomething_static_content/node--static_content.tpl.php#L30-L36).
#### Results

image

Fixes #5901.


DFurnes:
:+1: :ship:

PR #5904, opened 2015-12-07
Adding conditional logic for empty NS data.

### LINKED ISSUE #5903: Kudos API request is missing fallback logic for failed Northstar request
Just need to update it with the same logic used in RB and RB items requests.

If the request to get user info via Northstar fails, the logic still continues assuming something was found, so we get a bunch of undefined variable errors cause the object returned is an error object and not an object with expected user data.

weerd:
Fixes #5903
#### What's this PR do?

We updated the Kudos call to grab Northstar user information, but did not include the same logic we did for the Reportbacks and ReportbackItems in cases where the request to NS fails. This updates to include the same logic.

Only one file, so take a wild guess!

Pull down branch, `drush cc all` and give the `/kudos` endpoint a few hits. Check the DBLog errors in Drupal and see if a long list of notices shows up for undefined items if no NS user is found.
#5903

---

@angaither @jonuy @chloealee

### DIFF
--- a/lib/modules/dosomething/dosomething_kudos/includes/Kudos.php
+++ b/lib/modules/dosomething/dosomething_kudos/includes/Kudos.php
@@ -50,7 +50,7 @@ class Kudos extends Entity {
 
     foreach($results as $item) {
       $kudos = new static;
-      $kudos->build($item);
+      $kudos->build($item, TRUE);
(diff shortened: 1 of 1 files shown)


angaither:
:+1:

PR #5906, opened 2015-12-07
Translated Problem Share Images

### LINKED ISSUE #5751: MX/BR problem share images need to be generated using the SP/BR fact
image

Also note for these we are not actually translating the fact, we're just referencing a different fact in the MX/BR translation of the campaign.

sbsmith86:
#### What's this PR do?

This PR updates the cron job that creates problem share images so that it runs through each translation of the problem fact field and creates an image with that version of the fact statement. The images that it creates are appended with the country code associated with the language and defaults to 'global' if nothing else. 

Then we find which image to use in the share bar based on what URL prefix is currently being used to get the campaign page.

The cron job gets updated in `dosomething_campaign_problem_shares.cron.inc`

Going to different versions of a campaign page and making sure you get the correctly translated version of the problem statement image.

One thing to note is that I didn't translate the "Make a difference" header in the problem share images. So they end up looking like: 

screen shot 2015-12-07 at 2 04 20 pm

I can revisit this in a separate issue.

Fixes #5751

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -127,7 +127,7 @@ function dosomething_campaign_preprocess_common_vars(&$vars, &$wrapper) {
 
 
     // Create the image path to the problem fact statement image generated for the node.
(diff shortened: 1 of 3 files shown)


angaither:
:+1:

PR #5909, opened 2015-12-07
Fixes access control to reportback POST.

sergiitk:
See https://cobalt.io/dosomething/reports/70.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -309,7 +309,7 @@ function dosomething_reportback_form_validate($form, &$form_state) {
   }
 
   // If user isn't signed up for the campaign.
-  if (!dosomething_signup_exists($campaign_id)) {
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -293,14 +293,27 @@ function dosomething_reportback_validate_quantity($element, &$form_state) {
...
     // Tell them that.
     drupal_set_message(t("You are no longer logged in. Please log in."), 'error');
     // Redirect to the original node.
     // Without this, the user is redirected to the confirmation page.
-    drupal_goto('node/' . $form_state['values']['nid']);
+    drupal_goto('node/' . $campaign_id);
   }
+
+  // If user isn't signed up for the campaign.
+  if (!dosomething_signup_exists($campaign_id)) {
```
angaither: can we add an additional check here to allow staff to reportback without signing up?

sergiitk: Good point!

angaither:
:+1:

Issue #5907, opened 2015-12-07
Hot module time left way off


mikefantini:
Everything says 9 hours left regardless of amount of time left:

https://thor.dosomething.org/us/campaigns/grandparents-gone-wired-0

https://thor.dosomething.org/us/campaigns/superstressface

sbsmith86:
It works on my Dev, so something is off between environments.

mikefantini:
Works for:
- https://www.dosomething.org/us/campaigns/grandparents-gone-wired-0
- https://www.dosomething.org/us/campaigns/50-cans

Doesn't work for:
- https://www.dosomething.org/us/campaigns/reelwomen

mikefantini:
Actually no campaigns are working. They just appeared to be working because of optimizely tests that are running

mikefantini:
Note for myself to set these as hot again once this is pushed to prod: https://docs.google.com/spreadsheets/d/1uOMlmIle58u5h0XkKWQhSAsJA9Ehqwtjs3wZVexqLpc/edit#gid=0&vpid=A1

angaither:
fixed with #5910

PR #5910, opened 2015-12-07
Update hot module timing logic.

angaither:
Instead of calculating the time a bunch, just set it to vars array on the node.

refs #5907

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -624,16 +624,16 @@ function dosomething_campaign_get_scholarship($nid) {
  * @return string
  *  Formatted time (e.g. "4 days left")
  */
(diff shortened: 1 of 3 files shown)


sbsmith86:
:+1: :watch:

aaronschachter:
This seems to have broken things on staging:

screen shot 2015-12-07 at 5 05 00 pm

angaither:
hmm @aaronschachter strange that you aren't getting taken to `/us`

I think it's just an issue with not having translations on staging, but I'll look into this

aaronschachter:
FYI - `staging.dosomething.org` directs me to /`us`, but `staging.beta.dosomething.org` does not. Will purge browser history to stop visiting it. We've already made the changes in the LDT iOS app to point to `staging.dosomething.org/api`

mikefantini:
@aaronschachter staging.dosomething.org is behind Fastly whereas staging.beta.dosomething.org is not. This is why staging.dosomething.org directs you to /us but staging.beta does not.

PR #5916, opened 2015-12-08
Fix fatal error on stage.

angaither:
Refs #5910

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -624,8 +624,8 @@ function dosomething_campaign_get_scholarship($nid) {
  * @return string
  *  Formatted time (e.g. "4 days left")
  */
-function dosomething_campaign_get_hot_time_left(&$vars) {
-  $today = $vars['hm_today'];
+function dosomething_campaign_get_hot_time_left($node, &$vars = NULL) {
+  $today = isset($vars['hm_today']) ?: dosomething_campaign_set_est_timezone(new DateTime());
   $end_date = $vars['hm_end_date'];
   $difference = $today->diff($end_date, true);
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
```diff
@@ -624,8 +624,8 @@ function dosomething_campaign_get_scholarship($nid) {
  * @return string
  *  Formatted time (e.g. "4 days left")
  */
-function dosomething_campaign_get_hot_time_left(&$vars) {
-  $today = $vars['hm_today'];
+function dosomething_campaign_get_hot_time_left($node, &$vars = NULL) {
```
sbsmith86, line 627: Do you end up using `$node` anywhere?

angaither, line 627: `$node` is used [here](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_campaign/dosomething_campaign_hot_shares/dosomething_campaign_hot_shares.cron.inc#L30)

sbsmith86:
Looks good! Except, I still don't see the `$node` parameter being used anywhere in the `dosomething_campaign_get_hot_time_left()` function. Does the cron get called somewhere in this functions?

angaither:
@sbsmith86 yeah it's passed in in the cron, we might need to update that later today

angaither:
yeah cron errors, I'll make an issue for that :wink:

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
sbsmith86, line 627: @angaither right, but the actual function doesn't ever use it. It uses `$vars`.

Issue #5924, opened 2015-12-08
Fatal error upon log in


mikefantini:
#### Current Behavior

When I log as an existing user on a campaign pitch page I see a fatal error.
#### Desired Behavior

I should be logged into my account and taken to the action page.
#### Steps to Replicate
- Go to thor.dosomething.org
- Click on Explore Campaigns
- Click Just Because
- Taken to Pitch Page
- Go to log in as an existing test user and received the following error "Error. Error message." (see screenshot below)
#### Why This Matters

Existing users can't log in.
#### Relevant Screenshots

![image](https://cloud.githubusercontent.com/assets/6330971/11666346/0a41b3ee-9dba-11e5-8fb1-8096d86a3a1c.png)

mikefantini:
Possibly related to https://github.com/DoSomething/phoenix/commit/c8cdad599a5938381b44600fcc905e23688ab036?

@sergii-tkachenko mind taking a look?

sergiitk:
Yes, I can do that.

sergiitk:
So attempted fix #5925 didn't help.
@mikefantini Could you please find the same flow on staging?

sergiitk:
Note, I'm able to reproduce the bug directly without going through Explore Campaigns. It also doesn't matter if I have a signup for this campaign.

mikefantini:
@sergii-tkachenko I don't believe it's just happening on `Just Because` based on @namimody's comment in slack re: GGW https://dosomething.slack.com/archives/phoenix/p1449605593002216

sergiitk:
Good Find. So we need to found out what's common between `Just Because` and `Grandparents Gone Wired`.

mikefantini:
@sergii-tkachenko I just recreated this on prod with my admin log in on https://www.dosomething.org/us/campaigns/reelwomen?source:

![image](https://cloud.githubusercontent.com/assets/6330971/11668912/b75f524a-9dc7-11e5-92ae-4091ee615067.png)

sergiitk:
I can't reproduce locally even with thor DB...

sergiitk:
Just signed in to prod with no errors:
![image](https://cloud.githubusercontent.com/assets/672669/11669450/6804423a-9e05-11e5-8232-1f1eea1b1f6f.png)

mikefantini:
@sergii-tkachenko is there any way to know how often this is occurring on prod?

sergiitk:
[...]

PR #5925, opened 2015-12-08
Changes a ways session is initialized in `dosomething_global_init()`

sergiitk:
#### What's this PR do?
- Changes a ways session is initialized in `dosomething_global_init()`

See #5924

I wasn't able to reproduce it locally, but I think this might fix it.
`drupal_session_initialize()` looks more safe than `drupal_session_start()`.

Proposes a fix for #5924.


itsjoekent:
yeah lets try this :+1:

PR #5928, opened 2015-12-08
Removes session initialization.

sergiitk:
Testing if the session initialization in dosomething_global_init causes #5924.


itsjoekent:
:+1:

sergiitk:
Reopened #5620.

PR #5931, opened 2015-12-09
Use global english URL for image URLs so there is no country prefix

### LINKED ISSUE #5752: Image is not being pulled in for hot shares
image

sbsmith86:
#### What's this PR do?

Before a country prefix was being appended to the base url used to find the hot share image. This resulted in an incorrect url and the image was not able to be found. ex:

`http://dev.dosomething.org:8888/ussites/default/files/images/hot-1283.png` 

This PR uses the global english url for the hot share image URLs so that a country prefix doesn't get appended.

`http://dev.dosomething.org:8888/sites/default/files/images/hot-1283.png`

Fixes #5752


weerd:
:+1: :globe_with_meridians:

DFurnes:
:globe_with_meridians: :boat:

PR #5932, opened 2015-12-09
Experimenting with adding composer for PHPUnit use and future improvements.

weerd:
## Do NOT Merge (yet)!

As we're working towards updating the API and expanding on it, the need for tests covering methods used by the different classes is becoming more apparent. Adding unit tests using [PHPUnit](https://phpunit.de/) would help mitigate a lot of the issues that have been cropping up with bugs as we modify the code. 

However, Drupal's architecture does not allow us to use PHPUnit properly because Drupal requires its bootstrap process to run for all its files/classes/functions/etc to be loaded properly. PHPUnit runs by testing small "units" of code, and calling methods on classes directly, instatiating those classes and injecting dependencies as needed _without_ doing a full load of a framework's bootstrap. 

In an effort to modernize our code a bit, and modify the architecture a bit to help PHPUnit run tests efficiently, this PR experiments with implementing the use of composer to install our dependent libraries as well as providing a means to autoload all our modern class code as well as letting us namespace our classes.

So far we have tested this updated architecture with the API related classes (`Transformer`, `Campaign`, `Reportback`, etc) and using namespacing and autoloading is working great.


DFurnes:
excited corgi

weerd:
image

:dancer:

DFurnes:
For now, we're inserting the autoloader into the core `index.php`, but considering other options!

``` diff
<?php

/**
 * @file
 * The PHP page that serves all page requests on a Drupal installation.
 *
 * The routines here dispatch control to the appropriate handler, which then
 * prints the appropriate page.
 *
 * All Drupal code is released under the GNU General Public License.
 * See COPYRIGHT.txt and LICENSE.txt.
 */

/**
 * Root directory of Drupal installation.
 */
define('DRUPAL_ROOT', getcwd());

+ require_once DRUPAL_ROOT . '/../lib/autoload.php';

require_once DRUPAL_ROOT . '/includes/bootstrap.inc';
drupal_bootstrap(DRUPAL_BOOTSTRAP_FULL);
menu_execute_active_handler();
```

DFurnes:
Hmm, well we ran into an issue with that approach. It turns out `index.php` is only run on web requests, so Drush was exploding since it never got the autoloader. Moving it into `includes/bootstrap.inc` seems to keep everything working, although sadly still means hacking core.

``` diff
<?php

/**
 * @file
 * Functions that need to be loaded on every Drupal request.
 */

+/**
+ * Require Composer's autoloader.
+ */
+require_once __DIR__ . '/../../lib/autoload.php';

/**
 * The current system version.
 */
define('VERSION', '7.39');
```

DFurnes:
We've also looked into [`hook_boot`](https://api.drupal.org/api/drupal/modules%21system%21system.api.php/function/hook_boot/7) and [`hook_init`](https://api.drupal.org/api/drupal/modules%21system%21system.api.php/function/hook_init/7) but we're afraid that those aren't going to be guaranteed to run early enough in the bootstrapping process (e.g. if we want to `use` one of our PSR-4 app classes in a `*.module` file, how are we guaranteed the other module's `hook_init` has already run before that module's file is parsed?)

We also tried adding it to [`config/settings.php`](https://github.com/DoSomething/phoenix/blob/f84305ebf52515e25147f6eb530a9048553ec4ca/config/settings.php) but it seems like that runs too late.

sergiitk:
Why settings.php are too late?

sergiitk:
It’s included at the very start: [`_drupal_bootstrap_configuration()`](https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/drupal_settings_initialize/7) -> [`drupal_settings_initialize()`](https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/drupal_settings_initialize/7).

DFurnes:
@sergii-tkachenko Not sure, but adding it to `config/settings.php` like so…

``` diff
<?php

+/**
+ * Require Composer's autoloader.
+ */
+require_once __DIR__ . '/../lib/autoload.php';

/**
 * Lets do something about globals
 */
define('DS_PROFILE_PATH', 'profiles/dosomething');
define('DS_MODULES_PATH', DS_PROFILE_PATH . '/modules');
define('DS_THEMES_PATH', DS_PROFILE_PATH . 'themes');
define('DS_LIBRARIES_PATH', DS_PROFILE_PATH . '/libraries');

```

…still results in an error:
screen shot 2015-12-10 at 12 06 27 pm

DFurnes:
Ooooof, trying the exact same thing on @weerd's machine seems to work. I think my `config/settings.php` symlink might have gotten broken somehow.

weerd:
Yep, confirmed! DS site, API and drush all work when placing the `require` statement in the `settings.php` file on my local.

sergiitk:
Good! I was debugging `drupal_bootstrap()` already to find how it could include modules earlier than settings :smile:

DFurnes:
Another nice thing we've been able to do with this is use the single global `composer.json` for loading third-party dependencies (like the Zendesk API). This means we don't have to manually `cd` into each of those directories and separately install their dependencies, and it also means we can lock to a specific semver range rather than relying on a git branch. :smile:

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -215,45 +202,3 @@ function dosomething_northstar_build_ns_user($user, $form_state) {
   // Return fully built northstar user object.
   return $ns_user;
 }
-
-
-class Northstar {
```
angaither, line 220: if we can autoload we can switch to guzzle and update to the code that was commented out :)

tests/phpunit/ExampleTest.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+class ExampleTest extends TestCase
+{
+
+  /**
+   * Test all the things!
+   */
+  public function testItAll()
```
angaither, line 9: ha

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
weerd, line 220: Indeed!

angaither, line 220: don't remove it!

DFurnes, line 220: I'd like to make an actual PHP library that we could share between projects, so let's clear this up for now (since it's currently just squatting on the global `Northstar` namespace).

angaither, line 220: otay. I thought about that when building out aurora and this functionality, but then I felt like I was doing the most

DFurnes:
Tested a fresh `bin/ds build` on my local and everything seems happy. :tada:

weerd:
Yay! :dancers: :dancer: :dancers:

sergiitk:
Do we need to install dependencies for messagebroker-phplib?

sergiitk:
Also, what's with guzzle @angaither?

angaither:
@sergii-tkachenko used that to make easier http requests to NS

sergiitk:
@angaither Yes, I remeber that. So if we still need it, should we add it [here](https://github.com/DoSomething/phoenix/pull/5932/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R21)?

DFurnes:
> Do we need to install dependencies for messagebroker-phplib?

@sergii-tkachenko Ah, good call... that was a mistake. The module is hard-coded to expect to be a library, so we'll need to do some additional work if we want to add it to the global `package.json`.

> @angaither Yes, I remeber that. So if we still need it, should we add it here?

As far as I know, we're not actually using Guzzle anywhere in the codebase... is that wrong?

weerd:
@DFurnes I believe something was up with how we were trying to use Guzzle and it wasn't working, so instead we defaulted to the Drupal HTTP request function, but ideally I think we want to remove that code and reuse Guzzle again :)

angaither:
@weerd exactly 

> something was up with how we were trying to use Guzzle and it wasn't working, so instead we defaulted to the Drupal HTTP request function, but ideally I think we want to remove that code and reuse Guzzle again :)

DFurnes:
This has the _most_ merge conflicts. Closing.

Issue #5920, opened 2015-12-08
Global redirects are behaving erratically


mikefantini:
#### Current Behavior

When I go to www.dososomething.org/stress I'm correctly redirected to www.dosomething.org/us/campaigns/less-stress-text and www.grandparentsgonewired.com I'm correctly redirected to www.dosomething.org/us/campaigns/grandparents-gone-wired-0. 

However, I have seen Adam have an inconsistent experience where sometimes he is redirected to the /us/ version correctly but sometimes it stays on www.dosomething.org/campaigns/less-stress-text or www.dosomething.org/campaigns/grandparents-gone-wired-0.

This results in the error "Global translation unavailable for."
#### Desired Behavior

The desired behavior is that users in the US are redirected to the US translation of these campaigns.
#### Steps to Replicate

Go to www.dosomething.org/stress or www.grandparentsgonewired.com and see the below error.
#### Why This Matters

All of our external marketing for these two campaigns are to these vanity URLs.
#### OS/Browser

OS X El Capitan, 10.11.1 (15B42)
Chrome: Version 47.0.2526.73 (64-bit)
Safari: Version 9.0.1 (11601.2.7.2)
#### Relevant Screenshots

![image](https://cloud.githubusercontent.com/assets/6330971/11663679/94de893c-9dab-11e5-88fb-c7e862e463a2.png)

sergiitk:
Hi! I think it's related to the origin country of the user accessing this short urls. I'm not getting redirected either.

mikefantini:
He is also not being redirected when he goes to www.dosomething.org/campaigns/grandparents-gone-wired-0 to the /us/ translation.

Also he is being sent to the GL EN homepage in Safari and Chrome when he goes to www.dosomething.org (instead of being taken to /us)

sergiitk:
cc @deadlybutter. Can you give it a look?

itsjoekent:
ok thats really weird

itsjoekent:
maybe dumb question -- should I check adams computer for anything funky? maybe some privacy setting or incorrect language is enabled?

mikefantini:
@deadlybutter give it a shot

itsjoekent:
found this on adams computer in the request headers

![image](https://cloud.githubusercontent.com/assets/897368/11664113/dc44f250-9dad-11e5-9c04-978c22348408.png)

shouldnt that be X-Fastly-Country-Code (not -Client) ??

[...]

PR #5933, opened 2015-12-09
Refactors out session logic in Global redirects, also allows regional users to see global content

### LINKED ISSUE #5620: Users should be able to access GL EN content if translation is unpublished or does not exist in their region

itsjoekent:
#### What's this PR do?

Refactors redirect logic to no longer use session based state tracking

Currently deployed on thor, otherwise you can deploy this branch to it

Sessions are potentially the cause of #5920 and generally not the long term solution we want. I refactored the redirect logic to be DRY and subsequently made support to prevent looping before it starts (what sessions were fixing).

Fixes #5620 

cc: @mikefantini @blisteringherb @mshmsh5000 @mikefantini

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -102,12 +102,13 @@ function dosomething_global_init() {
   //
   // Passing 'noredirect=1' will defeat the redirect.
   if ($node_variables['is_node'] && !$node_variables['is_node_edit'] && !$node_variables['node_no_redirect']) {
-    if ($user_variables['user']->uid) {
-      dosomething_global_redirect_node($user_variables['user_language'], $node_variables['node'], 302);
+    $appropriate_language = dosomething_global_get_language($user, $node_variables['node'], FALSE);
+
+    if ($user->uid) {
+      dosomething_global_redirect_node($appropriate_language, $node_variables['node'], 302);
     } else {
-      dosomething_global_redirect_node($user_variables['user_language'], $node_variables['node'], 301);
+      dosomething_global_redirect_node($appropriate_language, $node_variables['node'], 301);
     }
-    return;
   }
 
 }
@@ -510,7 +511,6 @@ function dosomething_global_get_language($account, stdClass $node = NULL, $respe
   // If the user is anonymous, try to convert there country into a language
   if (!isset($account->language)) {
     $account->language = dosomething_global_convert_country_to_language(dosomething_settings_get_geo_country_code());
-
     // If that failed, set it to global en
     if (!isset($account->language)) {
       $account->language = DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE;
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -601,33 +595,14 @@ function dosomething_global_redirect_node($user_lang, $node, $redirect_status) {
...
-  else if (!empty($node->translations->data[DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE]) && $node->translations->data[DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE]['status']) {
-    $target_node = $node->translations->data[DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE];
-    $language = DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE;
-  }
-  // If neither of these exist, 404 the user
-  else {
-    drupal_not_found();
-    drupal_exit();
+  // Make sure we're not already on the correct page
+  if (dosomething_global_get_prefix_for_language($language) == dosomething_global_get_current_prefix()) {
```
sergiitk: Make it `===`. Just in case.

itsjoekent: :+1:

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -592,7 +586,7 @@ function dosomething_global_redirect_node_edit($node, $user_language) {
  * Else
  *  - 404 the user
  */
-function dosomething_global_redirect_node($user_lang, $node, $redirect_status) {
+function dosomething_global_redirect_node($language, $node, $redirect_status) {
```
sergiitk, line 589: Also please change `@param $user_lang` in the docblock.

itsjoekent, line 589: (done but line comment didnt reflect it)

sergiitk:
YES! Now we're talking. This is much better solution. :shipit: :+1: :ship:

itsjoekent:
been a while since someone used :shipit: haha

PR #5934, opened 2015-12-10
Fixes get_hot_time_left to unblock cron

### LINKED ISSUE #5917: running cron to errors out
Due to #5916 and #5910 the hot module cron shares error out

refactorrrr

itsjoekent:
#### What's this PR do?

The $end_date in get_home_time_left() had no default option if hm_end_date was NULL, unline $today which did. Just applied the same style of logic.

Does cron run?

Fixes #5917

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -629,7 +629,7 @@ function dosomething_campaign_get_scholarship($nid) {
  */
 function dosomething_campaign_get_hot_time_left($node, &$vars = NULL) {
   $today = isset($vars['hm_today']) ? $vars['hm_today'] : dosomething_campaign_set_est_timezone(new DateTime());
-  $end_date = $vars['hm_end_date'];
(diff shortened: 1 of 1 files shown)


itsjoekent:
@angaither can you review?

angaither:
@deadlybutter can you breakout getting the time into a function so we can use it [here](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module#L764-L766) or in case it isn't set

also you could remove the `$node` arg from here and the call in cron since it isn't being used

itsjoekent:
Can I do this in the current hot module branch im working on? Ill merge this and rebase what im workin on.

angaither:
that works, I'll close this one

itsjoekent:
so do the cron fix in that one too?

angaither:
@deadlybutter yeah that works

itsjoekent:
i just logged it with print R, when cron runs the node argument is filled out and vars is empty

angaither:
yes, so you can remove `$node` from `function dosomething_campaign_get_hot_time_left($node` and leave `Svars = NULL`

Issue #5936, opened 2015-12-10
Run URLs through Facebook Debug endpoint


sbsmith86:
We have a lot of issues with Facebook not crawling new or updated content fast enough. For example, if a campaign is just published, then we have to wait for the page to be shared at least once before Facebook will pull in the data. Another example is permalink pages where each page is a new url custom to the user, and when the user goes to share it, the data isn't ready for Facebook.

Good news is, I found out that [Facebook has an endpoint](https://developers.facebook.com/docs/sharing/opengraph/using-objects#update) you can hit that runs URL's through the debugger so we can pre-cache them whenever we want. 

I suggest we, 
- [ ] Create a function that accepts a URL and makes a call to that endpoint so that it can get scraped.
- [ ] When something gets published, run it through this function
- [ ] Before taking a user to their permalink page, run it through this function. 

Thoughts, @angaither ?

angaither:
@sbsmith86 I'm down for this but maybe the permalink call should be a non-blocking js solution? I don't want to slow requests for the user before showing them the permalink page

on publishing a campaign we need to call it for problem shares/hot module, right?
we could run that as a post publish hook, or as a part of the cron that creates the shares

sbsmith86:
@angaither Hot shares and problem shares are a little different because they aren't dependent on the meta tags on the page, instead we explicitly pass the data as a query string in the link used to create the facebook share dialog, so we should be fine there. This is specific for when you like, copy a url, and paste it into facebook to share it. 

Issue #5096 is for changing the permalink Facebook share functionality to behave more like hot/problem shares, maybe we can go that route instead so we don't have to worry about blocking the page.

sbsmith86:
#5096 was done, so I think we can close this.

Issue #5439, opened 2015-10-09
Campaign Title and Cover photo are not being pulled in on social shares


lizzydivine:
When someone posts "Just Don't Wear Blackface" on Facebook, the name of the campaign and the picture don't appear

![](https://slack-files.com/files-tmb/T024GV2BW-F0C6W1WSH-baeffda165/screenshot_2015-10-09_14.59.40_360.png)

mikefantini:
@sbsmith86 do you have thoughts on this one?

i'm still seeing this not working still. however it's working for me on https://www.dosomething.org/campaigns/social-media-makeover?source=campaigns when I just post both links as my status on FB.

do we need to put _all campaigns_ through the FB linter after we create them?

sbsmith86:
@mikefantini something very weird is going on. i'm not seeing any og meta tags on the page, so there actually is no image set for facebook to crawl. It doesn't look like running it through fb linter helps fix it either. See the screenshot below of what I got back when I ran Just Don't Wear Black Face through the linter.

<img width="938" alt="screen shot 2015-10-14 at 1 37 31 pm" src="https://cloud.githubusercontent.com/assets/1700409/10492257/0f8bef8e-7279-11e5-9c6d-3624e0bea289.png">

mikefantini:
hmmm yeah bummer so meta data on the node https://www.dosomething.org/node/5840/edit:

![image](https://cloud.githubusercontent.com/assets/6330971/10492463/46be212e-727a-11e5-8dc2-847da2224383.png)

compared to an old campaign https://www.dosomething.org/node/2544/edit:

![image](https://cloud.githubusercontent.com/assets/6330971/10492478/5f6f9be4-727a-11e5-9233-0d70d8180695.png)

@sbsmith86 @angaither @sergii-tkachenko probably something with translation?

mikefantini:
hmmm although this is showing up on a new campaign I created as an editor on staging:

https://staging.dosomething.org/node/1735/edit?destination=admin/content

sbsmith86:
Also seeing the issue why the `og:image` tag isn't being set. In [this line](https://github.com/DoSomething/phoenix/blob/5896b2823f770c6aa3a1e9cf46d474ef4ae60564/lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module#L80) Is returning nothing so `$image_nid` is not set and we can create the image from it. Seems to be tied to using `LANGUAGE_NONE` key to find the image nid

[...]

Issue #5645, opened 2015-10-27
Create a helper function to return node title?


sergiitk:
### Refactoring discussion

Since we have 2 different ways to retrieve node title for translatable and not translatable content types, it might make sense to create a helper to return title in one unified way.
See https://github.com/DoSomething/phoenix/pull/5642#issuecomment-151628801

---

cc @angaither

PR #5937, opened 2015-12-10
Fix facebook sharing

sbsmith86:
#### What's this PR do?

Facebook was throwing some warnings: 

screen shot 2015-12-09 at 5 12 04 pm

They are suggesting that we explicitly add all of the og tags, listed above, on the page. This PR adds them all for campaign pages. Also extracts this logic into a function since we also do it on permalink pages. We should probably add them on all pages, but I'm not sure what all the settings should be for everytype of page. Campaign pages seem to be priority for now. 

[Start here](https://github.com/DoSomething/phoenix/compare/dev...sbsmith86:fix-facebook-sharing?expand=1#diff-34bae740579fece653f7f7b32b22baa9R54). I set all the Facebook properties and call the new function that adds the tags to the page.

This probably won't fix anything, but we are seeing issues with Facebook pulling inconsistent data from the page, so I want to clean up these warnings so we can continue and debug the issue. 

However, [I think this issue](https://github.com/DoSomething/phoenix/issues/5936), will fix most of our facebook sharing woes.

Addresses #5439

### DIFF
--- a/lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
+++ b/lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
@@ -8,6 +8,7 @@
  * Implements hook_preprocess_page().
  */
 function dosomething_metatag_preprocess_page(&$vars) {
+  global $user;
 
   if (!empty($_GET['pass-reset-token'])) {
     // Do not cache pages with reset token set.
@@ -52,11 +53,13 @@ function dosomething_metatag_preprocess_page(&$vars) {
 
   // Add explicit facebook og meta tags to campaign pages.
   if (isset($node) && $node->type == 'campaign') {
+    $langauge_code = dosomething_global_get_language($user, $node);
+
     $facebook_data = array(
       'type' => 'dosomething_facebook:campaign',
       'url' => url(current_path(), array('absolute'=> TRUE)),
       'title' => $node->title,
-      'description' => "Volunteer and take action! " . $node->field_call_to_action[$node->language][0]['value'],
+      'description' => "Volunteer and take action! " . $node->field_call_to_action[$langauge_code][0]['value'],
     );
     dosomething_metatag_add_facebook_tags($facebook_data);
   }
@@ -103,6 +106,7 @@ function dosomething_metatag_get_metatag_image($node = NULL) {
  */
 function dosomething_metatag_add_metatag_image($node = NULL) {
   $image = dosomething_metatag_get_metatag_image($node);
+
   if ($image) {
     // Add social sharing thumbnail meta tag,
     // <link rel="image_src" href="[image]"> into the HEAD.
@@ -128,6 +132,8 @@ function dosomething_metatag_add_metatag_image($node = NULL) {
     );
     drupal_add_html_head($element, 'og:image');
 
+    // Facebook best practices say to add width & height tags
+    // 740x480 is the crop we are using on the images.
     $element = array(
       '#tag' => 'meta',
       '#attributes' => array(
@@ -227,6 +233,7 @@ function dosomething_metatag_get_permalink_vars() {
     'title' => $node->title,
     'description' => $share_copy,
     'image' => $image,
+    // 480x480 is the crop we are using on the images.
     'width' => 480,
     'height' => 480,
   );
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
```diff
@@ -50,10 +50,20 @@ function dosomething_metatag_preprocess_page(&$vars) {
   // Add relevant image metatags.
   dosomething_metatag_add_metatag_image($node);
 
+  // Add explicit facebook og meta tags to campaign pages.
+  if (isset($node) && $node->type == 'campaign') {
+    $facebook_data = array(
+      'type' => 'dosomething_facebook:campaign',
+      'url' => url(current_path(), array('absolute'=> TRUE)),
+      'title' => $node->title,
+      'description' => "Volunteer and take action! " . $node->field_call_to_action[$node->language][0]['value'],
```
angaither: does this work on different translations? 
check [this](https://github.com/DoSomething/phoenix/blob/872486f8f1528224cab1db71d93db732f0a6e5cb/lib/modules/dosomething/dosomething_global/dosomething_global.module#L496)

lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
```diff
@@ -116,7 +126,26 @@ function dosomething_metatag_add_metatag_image($node = NULL) {
...
+    drupal_add_html_head($element, 'og:image:width');
+
+    $element = array(
+      '#tag' => 'meta',
+      '#attributes' => array(
+        "property" => "og:image:height",
+        "content" =>  480,
+      ),
+    );
+    drupal_add_html_head($element, 'og:image:height');
```
angaither, line 153: what's the deal with `740x480` does it have to be hardcoded? can it be pulled from the image, or is this always the size we want?

lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
```diff
@@ -190,6 +220,27 @@ function dosomething_metatag_get_permalink_vars() {
   );
   drupal_add_html_head($element, 'twitter:url');
 
+  // Facebook.
+  $facebook_data = array(
+    'type' => 'dosomething_facebook:campaign',
+    'url' => $share_url,
+    'title' => $node->title,
```
angaither, line 233: same here,  with translations... this will pull the original node title 
cc @sergii-tkachenko #5645

lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
sbsmith86, line 153: @angaither Facebook suggests that we include these dimensions so the image doesn't have to load asynchronously as a part of it's [best practices](https://developers.facebook.com/docs/sharing/best-practices#precaching). 740x480 is the crop we use to build the image URL. I can either hard code here, or try to extract it from the string we pass the function that builds the image URL.

lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
sbsmith86, line 233: @angaither hmm this seems to be pulling the right translation. Maybe because I'm not pulling it from `title_field`? It seems the issue that is referenced is for translatable campaign collections?

screen shot 2015-12-10 at 2 38 27 pm
screen shot 2015-12-10 at 2 38 18 pm
screen shot 2015-12-10 at 2 38 03 pm

lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
sbsmith86: ah i will update

lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
angaither, line 153: maybe just add a comment about where the size is coming from?

lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
angaither, line 233: kewl

sbsmith86:
@angaither addressed your comments! Can you take another look?

angaither:
:+1:

PR #5940, opened 2015-12-10
Adds specific deadline to hotmodule stat card

### LINKED ISSUE #5917: running cron to errors out

### LINKED ISSUE #5922: Display only days left
on the hot module, instead of saying date it ends, just display number of days
- [x] Display days left
- [ ] add help text (TBD) to say it ends x day in EST

cc @mikefantini @lkpttn

itsjoekent:
#### What's this PR do?

Adds a specific deadline to the hot module stat card stating the end date time pm/am and timezone 
Also fixes hm_readable_end_date

Also covers this PR https://github.com/DoSomething/phoenix/pull/5934

Currently on Thor, heres an example https://thor.dosomething.org/us/campaigns/smiles-soldiers

otherwise test if the hot module is outputting correctly 

For some reason the PHP decided to do formatting in GMT, this caused confusion because of all our settings are defaulted to UTC. The solution was date_default_timeszone_set("EST") as I saw on PHP.net http://php.net/manual/en/function.date.php#refsect1-function.date-examples

Fixes #5922 
Fixes #5917  

screen shot 2015-12-10 at 1 20 58 pm
screen shot 2015-12-10 at 1 22 20 pm

This is what it looks like as an Australian user normally 1 day ahead of the US/East (Date in backend is the 15th) 
screen shot 2015-12-10 at 2 24 09 pm
# DONT MERGE YET

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -628,7 +628,10 @@ function dosomething_campaign_get_scholarship($nid) {
  *  Formatted time (e.g. "4 days left")
  */
 function dosomething_campaign_get_hot_time_left($node, &$vars = NULL) {
-  $timing = dosomething_campaign_get_goals_times($vars);
+  if (!isset($node) && !isset($vars)) {
+    return;
+  }
+  $timing = dosomething_campaign_get_goals_times($vars ?: $node);
   $today = $timing['today'];
   $end_date = $timing['end_date'];
   $difference = $today->diff($end_date, true);
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -628,12 +628,16 @@ function dosomething_campaign_get_scholarship($nid) {
  *  Formatted time (e.g. "4 days left")
  */
 function dosomething_campaign_get_hot_time_left($node, &$vars = NULL) {
-  if (!isset($node) && !isset($vars)) {
-    return;
+  if (!isset($vars) && !isset($vars['hm_today'])) {
+    $timing = dosomething_campaign_get_goals_times($node);
+
+    if (!isset($timing)) {
+      return;
+    }
   }
-  $timing = dosomething_campaign_get_goals_times($vars ?: $node);
-  $today = $timing['today'];
-  $end_date = $timing['end_date'];
+
+  $today = $vars['hm_today'] ?: $timing['today'];
+  $end_date = $vars['hm_end_date'] ?: $timing['end_date'];
   $difference = $today->diff($end_date, true);
   $days_left = $difference->d;
   $hours_left = $difference->h + ($days_left * 24);
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -628,8 +628,9 @@ function dosomething_campaign_get_scholarship($nid) {
  *  Formatted time (e.g. "4 days left")
  */
 function dosomething_campaign_get_hot_time_left($node, &$vars = NULL) {
-  $today = isset($vars['hm_today']) ? $vars['hm_today'] : dosomething_campaign_set_est_timezone(new DateTime());
-  $end_date = $vars['hm_end_date'];
+  $timing = dosomething_campaign_get_goals_times($vars);
+  $today = $timing['today'];
+  $end_date = $timing['end_date'];
   $difference = $today->diff($end_date, true);
   $days_left = $difference->d;
   $hours_left = $difference->h + ($days_left * 24);
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -763,12 +763,14 @@ function dosomething_campaign_is_hot_module(&$vars) {
   if ($hot_module_enabled && isset($vars['field_high_season'])) {
     $today = dosomething_campaign_set_est_timezone(new DateTime());
     $start_date = dosomething_campaign_set_est_timezone(new DateTime($vars['field_high_season'][0]['value']), TRUE);
+    $unmodified_end_date = new DateTime($vars['field_high_season'][0]['value2']);
     $end_date = dosomething_campaign_set_est_timezone(new DateTime($vars['field_high_season'][0]['value2']), TRUE, TRUE);
 
     // Set the values to the vars array, so we don't need to do the math again.
     $vars['hm_today'] = $today;
     $vars['hm_start_date'] = $start_date;
     $vars['hm_end_date'] = $end_date;
+    $vars['hm_readable_end_date'] = $unmodified_end_date->format('F jS h') . ':' . $unmodified_end_date->format('i T');
 
     // Hot module is only enabled when variable is set, high season dates exist and there is still days left.
     if (!empty($vars['field_high_season']) && (($start_date < $today)  && ($today < $end_date))) {
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -762,19 +762,16 @@ function dosomething_campaign_is_hot_module(&$vars) {
   $hot_module_enabled = dosomething_helpers_get_variable('node', $vars['nid'], 'hot_module_enabled');
   if ($hot_module_enabled && isset($vars['field_high_season'])) {
     date_default_timezone_set('EST');
-    $today = dosomething_campaign_set_est_timezone(new DateTime());
-    $start_date = dosomething_campaign_set_est_timezone(new DateTime($vars['field_high_season'][0]['value']), TRUE);
-    $unmodified_end_date = new DateTime($vars['field_high_season'][0]['value2']);
-    $end_date = dosomething_campaign_set_est_timezone(new DateTime($vars['field_high_season'][0]['value2']), TRUE, TRUE);
+    $timing = dosomething_campaign_get_goals_times($vars);
 
     // Set the values to the vars array, so we don't need to do the math again.
-    $vars['hm_today'] = $today;
-    $vars['hm_start_date'] = $start_date;
-    $vars['hm_end_date'] = $end_date;
-    $vars['hm_readable_end_date'] = $unmodified_end_date->format('F jS h') . ':' . $unmodified_end_date->format('ia T');
+    $vars['hm_today'] = $timing['today'];
+    $vars['hm_start_date'] = $timing['start_date'];
+    $vars['hm_end_date'] = $timing['end_date'];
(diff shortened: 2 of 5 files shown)


itsjoekent:
StyleCI... glitched? Am I crazy?

This is the line its complaining about

```
<p><?php print t("Deadline for reaching to goal is") ?></p>
```

https://github.com/DoSomething/phoenix/pull/5940/files#diff-bbf3eaf4992babe1e8035368de5f7ec7R19

But StyleCI claims its a single quote

```
<p><?php print t('Deadline for reaching to goal is') ?></p>
```

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -761,14 +761,17 @@ function dosomething_campaign_add_shipment_form_vars($node) {
 function dosomething_campaign_is_hot_module(&$vars) {
   $hot_module_enabled = dosomething_helpers_get_variable('node', $vars['nid'], 'hot_module_enabled');
   if ($hot_module_enabled && isset($vars['field_high_season'])) {
+    date_default_timezone_set('EST');
```
angaither: if we are setting the timezone here, do we also need `dosomething_campaign_set_est_timezone`?

DFurnes:
> But StyleCI claims its a single quote

@deadlybutter StyleCI is saying you _should_ use single-quotes, not that you have already. It's displaying it as a diff of the changes that you need to make to your PR to satisfy the linter rules.

itsjoekent:
@DFurnes OH, ok. thanks!

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
itsjoekent: No, we've had it semi wrong, see the context section of the PR

angaither: so can we remove that code?

itsjoekent: good question - ill try and see what happens

itsjoekent:
gonna need some @mikefantini input here

Since I found a much better and cleaner way of converting the time to EST, we dropped the old function I was using. In doing so, this happened:

screen shot 2015-12-10 at 3 28 49 pm

(It used to say 4 days left)

While it's technically correct (3 days and however many hours), I just want to make sure its not confusing and :ok:

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
```diff
@@ -628,8 +628,12 @@ function dosomething_campaign_get_scholarship($nid) {
  *  Formatted time (e.g. "4 days left")
  */
 function dosomething_campaign_get_hot_time_left($node, &$vars = NULL) {
```
angaither, line 630: shouldn't need node in here anymore

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
```diff
@@ -628,8 +628,12 @@ function dosomething_campaign_get_scholarship($nid) {
...
-  $today = isset($vars['hm_today']) ? $vars['hm_today'] : dosomething_campaign_set_est_timezone(new DateTime());
-  $end_date = $vars['hm_end_date'];
+  if (!isset($node) && !isset($vars)) {
+    return;
+  }
+  $timing = dosomething_campaign_get_goals_times($vars ?: $node);
```
angaither: this should already be set in the `vars` array (except in the instance of cron, so this can be run only if you need a fallback

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -761,17 +761,17 @@ function dosomething_campaign_add_shipment_form_vars($node) {
 function dosomething_campaign_is_hot_module(&$vars) {
   $hot_module_enabled = dosomething_helpers_get_variable('node', $vars['nid'], 'hot_module_enabled');
   if ($hot_module_enabled && isset($vars['field_high_season'])) {
-    $today = dosomething_campaign_set_est_timezone(new DateTime());
-    $start_date = dosomething_campaign_set_est_timezone(new DateTime($vars['field_high_season'][0]['value']), TRUE);
-    $end_date = dosomething_campaign_set_est_timezone(new DateTime($vars['field_high_season'][0]['value2']), TRUE, TRUE);
+    date_default_timezone_set('EST');
```
angaither: you're setting this date default here and in `dosomething_campaign_get_goals_times`

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -908,6 +912,25 @@ function dosomething_campaign_set_est_timezone($date, $add_four = FALSE, $end_of
 }
 
 /**
+ * Returns array of date times ofor the given campaign node
+ *
+ */
+function dosomething_campaign_get_goals_times($node) {
```
angaither, line 924: i think this is `$vars`, not `node`

angaither, line 924: and if you pass by reference, you can set this as an array on `$vars`

mikefantini:
@deadlybutter I think this is fine. I've been looking at Kickstarter and can't figure out the specific pattern but they would do 4 days when it's like 3 days and more than 6 hours left and then switch to 3 days if it's 3 days and less than 6 hours left (6 hours is my approximation looking at a ton of projects).

Obviously just keeping it as you have it here is easiest and I don't think it's overly confusing personally. Especially because we go to hours once it gets within 48.

itsjoekent:
@mikefantini :+1:

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
itsjoekent, line 630: resolved https://dosomething.slack.com/archives/phoenix/p1449848952002812

itsjoekent:
@angaither good to go?

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -908,6 +911,34 @@ function dosomething_campaign_set_est_timezone($date, $add_four = FALSE, $end_of
...
+ *    today - The current day
+ *    start_date - The day the campaign starts
+ *    end_date - The day the campaign ends
+ */
+function dosomething_campaign_get_goals_times($node) {
+  if (!isset($node)) {
+    return NULL;
+  }
+  date_default_timezone_set('EST');
+  $high_season = is_array($node) ? $node['field_high_season'] : $node->field_high_season[LANGUAGE_NONE];
```
angaither: this field is translatable now, you have to get the correct lang for this

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -247,7 +247,7 @@ function dosomething_campaign_preprocess_hot_module(&$vars, &$wrapper) {
   $start_date = $vars['hm_start_date'];
   $end_date = $vars['hm_end_date'];
   $readable_end_date = $vars['hm_readable_end_date'];
-  $time_left = dosomething_campaign_get_hot_time_left($node, $vars);
+  $time_left = dosomething_campaign_get_hot_time_left($vars);
```
angaither: doesn't this break if the first arg in the function is still $node?

itsjoekent: nope, multiple safety checks are in place

https://github.com/DoSomething/phoenix/pull/5940/files#diff-f3333e41a97facbf8cfed1f49cdee095R634

https://github.com/DoSomething/phoenix/pull/5940/files#diff-ae398bb8864fc3c26317f7ccaef04c0aR925

angaither: the function definition is `function dosomething_campaign_get_hot_time_left($node, &$vars = NULL)`

and you are calling it here without a NULL `$node` value, this will treat the vars array as the node object
dosomething_campaign_get_hot_time_left($vars);

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
itsjoekent: are you sure? I logged this and it was either LANGUAGE_NONE or just 0 (depends how you access it, node vs vars), and this is how it always was, see link below 

https://github.com/DoSomething/phoenix/blob/16bde548324a55db89de482a57ea85ba3ea3b04c/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module#L765

Was this field recently made translatable?

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
itsjoekent: Ohhhhhhh yeah

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
angaither: yes https://github.com/DoSomething/phoenix/pull/5869

itsjoekent: Is the hot module still working on Thor/Prod? Something isnt right then because the existing logic doesn't specify language

angaither: right, because it's using the `$vars` array

itsjoekent: Oh and on the node you need to specific it, ok that makes sense.

Errr... how do we get language in this context though? No URL's or users to base it off of... Source language?

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -867,16 +866,20 @@ function dosomething_campaign_is_active($node) {
...
-     }
+  $win_module_enabled = variable_get('dosomething_campaign_enable_win_module', NULL);
+  if ($win_module_enabled == 1 && isset($vars['field_high_season']) && $hot_module_enabled) {
+    $timing = dosomething_campaign_get_goals_times($vars);
+
+    // Set the values to the vars array, so we don't need to do the math again.
+    $vars['hm_today'] = $timing['today'];
+    $vars['hm_end_date'] = $timing['end_date'];
+
+    $win_module_end = dosomething_campaign_set_est_timezone(date_modify(clone $timing['end_date'], '+30 days'), TRUE, TRUE);
```
angaither: i thought the goal was to remove this function (`dosomething_campaign_set_est_timezone`) since it caused wonkyness with the timezone?

itsjoekent: good catch! ill fix that

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
angaither: the node should have a language attached to it `$node->language`

itsjoekent:
@angaither I think we're good now!

angaither:
:+1: kewl give it a merge and push to thor
cc @namimody!

itsjoekent:
deployed

PR #5945, opened 2015-12-11
making new cache for campaign objects

### LINKED ISSUE #2771: Campaign object caching
Implement a custom cache table to store the Campaign object returned by `dosomething_campaign_load`.
- [ ] Insert/update cache record upon node insert/update
- [ ] Return cache record within `dosomething_campaign_load`
- [ ] Add extra param to `dosomething_campaign_load` to bypass cache

katiecrane:
fixes #2771 

making new cache for campaign objects and storing them when `dosomething_campaign_load` is called

@weerd can you check on the `if ($public)` part of this chunk? does that need to be there if the campaign object is being pulled from the cache?

```
if ($cached_node = cache_get('ds_campaign_' . $node->nid, 'cache_dosomething_campaign') && $cache) { 
    // If this is accessed via API: 
    if ($public) { 
      // Add stats property. 
      dosomething_campaign_load_stats($campaign); 
      // Unset properties which shouldn't be public. 
      unset($campaign->variables); 
    } 
    return $campaign = $cached_node->data; 
  }
```

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -278,10 +278,10 @@ function dosomething_campaign_load($node, $public = FALSE, $cache = TRUE) {
 
     if($cache)
     {
-      cache_set('ds_campaign_' . $campaign->nid, 'cache_dosomething_campaign');
+      cache_set('ds_campaign_' . $campaign->nid, $campaign, 'cache_dosomething_campaign');
     }
   }
-  
+
   return $campaign;
 }
 
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -275,7 +275,7 @@ function dosomething_campaign_load($node, $public = FALSE, $cache = TRUE) {
     // Load Creator property.
     dosomething_campaign_load_creator($campaign, $wrapper);
 
-    cache_set('ds_campaign_' . $campaign->nid, $campaign, 'cache_dosomething_campaign');
+    cache_set('ds_campaign_' . $campaign->nid, $campaign, 'cache_dosomething_campaign', REQUEST_TIME + 60*60*24);
 
     // If this is accessed via API:
     if ($public) {
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -86,7 +86,7 @@
  *   - vips (string). HTML.
  *
  */
-function dosomething_campaign_load($node, $public = FALSE, $cache) {
+function dosomething_campaign_load($node, $public = FALSE, $cache = TRUE) {
   if ($node->type != 'campaign') { return FALSE; }
 
   if ($cached_node = cache_get('ds_campaign_' . $node->nid, 'cache_dosomething_campaign')) {
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -88,8 +88,11 @@
  */
 function dosomething_campaign_load($node, $public = FALSE, $cache = TRUE) {
   if ($node->type != 'campaign') { return FALSE; }
+  
+  // Get the appropriate lang
+  $current_lang = dosomething_global_convert_country_to_language(dosomething_global_get_current_prefix());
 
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
```diff
@@ -84,190 +86,205 @@
...
-    $campaign->scholarship = (int) $campaign->scholarship;
+  if ($cached_node = cache_get('ds_campaign_' . $node->nid, 'cache_dosomething_campaign') && $cache) {
+    // If this is accessed via API:
+    if ($public) {
+      // Add stats property.
+      dosomething_campaign_load_stats($campaign);
+      // Unset properties which shouldn't be public.
+      unset($campaign->variables);
+    }
+    return $campaign = $cached_node->data;
```
angaither, line 103: i think you would want to remove the variables from the `$cached_node->data`

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -1171,3 +1171,11 @@ function dosomething_campaign_get_campaign_runs($node) {
 
   return $campaign_runs;
 }
+
+/**
+ * Implements hook_flush_caches()
+ *
+ */
+function dosomething_campaign_flush_caches() {
```
angaither, line 1179: you may want to add a flush_cache function for a specific campaign and update it on node updates

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
katiecrane, line 103: I'm not sure what you mean?

angaither, line 103: warning, fake code below

```
$cached_node = cache_get('whatever whatever)->data

if (public)
unset $cached_node->variables

return $cached_node
```

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
katiecrane, line 1179: where do node updates happen?

katiecrane:
@angaither I think this is done now? just pending approval from @weerd

weerd:
Hey! Sorry @katiecrane was wrapped up in a few things end of last week; will check first thing on monday!

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
weerd, line 1179: There should be some stuff in the `dosomething_campaign.module` file that has update related things.

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
```diff
@@ -84,190 +86,205 @@
...
+      // Same as the image_cover.
+      $campaign->image_header = $campaign->image_cover;
+    }
+
+    // Load Creator property.
+    dosomething_campaign_load_creator($campaign, $wrapper);
 
-  // Load Creator property.
-  dosomething_campaign_load_creator($campaign, $wrapper);
+    cache_set('ds_campaign_' . $campaign->nid, $campaign, 'cache_dosomething_campaign');
```
weerd: Might be good to not cache it forever, but add a time for how long to cache the node for.

angaither: @weerd there's a cache clear when the node gets updated, forever should be fine

weerd:
@katiecrane also regarding the `$public` part, I don't think you'd need it if it's being pulled from cache.... Unless there's the possibility that when the first call for the campaign load happens and asks for public data, caches that and then on a load request of a cached version, passing `FALSE` for `$public`, could still load the cached version w/ public data exposed?

That all being said, this function is entirely _not_ used for the API except in the gallery (which I'm working ons witching out), so it's mostly just used internally ;)

angaither:
> Unless there's the possibility that when the first call for the campaign load happens and asks for public data, caches that and then on a load request of a cached version, passing FALSE for $public, could still load the cached version w/ public data exposed?

@weerd yes, that's why I had her add that part. didn't want non-public data to be exposed in the cache. another way to go about this is to only cache it after public data has been removed.

weerd:
@angaither Yeah, unless for some reason you need a version with the public data and the cached version has none?

But again, this is mostly used for internal stuff and no longer really being used for public API stuff other than gallery which is getting deprecated anyhoo...

angaither:
@weerd got it. I don't think there's any harm in leaving it like it is though, right?

weerd:
Yeah, I don't think so. Should be fine. I might add an expiration time to the `cache_set()` just in case for the time being, but other than that, should be fine (famous last words).

katiecrane:
@angaither I added the timestamp back in and nothing broke. I think it's good now?

angaither:
:+1:

PR #5946, opened 2015-12-11
Switch permalink to use facebook feed dialog

### LINKED ISSUE #5096: Change Permalink Facebook Share Functionality from Meta Data to Explicit Call
@sbsmith86 poor attempt at an issue...

During staging tests, noticed that the RB photo is only intermittently shared from a reportback permalink to Facebook. This is because the photo being pulled in is referenced through metadata, not explicitly. Pulling it in explicitly should eliminate the problem.

image

sbsmith86:
#### What's this PR do?

Switch the type of share we use for the Facebook share functionality on permalink pages. The feed dialog allows us to customize the data that is shared manually, without depending on meta tags. This helps us get around Facebook scraping and caching issues on these pages that get created on the fly. 

Was working on manually forcing facebook to crawl these pages using their API, but that has the potential to block the page from loading. This approach allows us to have more custom functionality for the facebook share, but only when the user actually clicks the facebook share button.

Fixes #5096


angaither:
:+1:

PR #5948, opened 2015-12-14
Removes comments

itsjoekent:


weerd:
:no_entry: comments!

:+1:

itsjoekent:
they were notes to self i left on friday before end of day, i wouldnt even call them that haha

PR #5949, opened 2015-12-14
Adds 23 hours and 59 minutes to end date

itsjoekent:
Hot module now reads

Deadline for reaching the goal is
December 31st 11:59pm EST

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -933,7 +933,7 @@ function dosomething_campaign_get_goals_times($node) {
   $timing = [
     "today" => new DateTime(),
     "start_date" => new DateTime($high_season[0]['value']),
-    "end_date" => new DateTime($high_season[0]['value2']),
+    "end_date" => date_modify(new DateTime($high_season[0]['value2']), '+23 hours +59 minutes + 59 seconds'),
     "win_module_end" => date_modify(new DateTime($high_season[0]['value2']), '+30 days'),
   ];
   return $timing;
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -933,7 +933,7 @@ function dosomething_campaign_get_goals_times($node) {
   $timing = [
     "today" => new DateTime(),
     "start_date" => new DateTime($high_season[0]['value']),
-    "end_date" => new DateTime($high_season[0]['value2']),
+    "end_date" => date_modify(new DateTime($high_season[0]['value2']), '+23 hours +59 minutes + 59 seconds'),
```
itsjoekent: ahhhh wait there shouldnt be a space in seconds even though we're not using it

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -933,7 +933,7 @@ function dosomething_campaign_get_goals_times($node) {
   $timing = [
     "today" => new DateTime(),
     "start_date" => new DateTime($high_season[0]['value']),
-    "end_date" => new DateTime($high_season[0]['value2']),
+    "end_date" => date_modify(new DateTime($high_season[0]['value2']), '+23 hours +59 minutes +59 seconds'),
```
angaither, line 936: curious why you switched from `setTime` to `datemodify`
https://github.com/DoSomething/phoenix/blob/beb2cbb520264f8b3abfbe8be6fc6ba6b13367ea/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module#L899

itsjoekent, line 936: Just seemed cleaner and more reliable. Also looking at it should hopefully make more sense "we are adding 23 hours 59 minutes and seconds to this date object".

itsjoekent, line 936: This is used by the win module right below it, and it hasn't really had any problems.

angaither, line 936: okay

angaither:
:+1:

PR #5950, opened 2015-12-14
There's an additional `days` attribute on the diff.

### LINKED ISSUE #5856: Hot Module Ignores Month When Counting Down
For example, https://www.dosomething.org/us/campaigns/treat-yo-friends ends on Dec 31. However, the page shows: 

image

with the correct date in the header photo, the correct x-axis on the graph, but the dark blue banner (showing 30 hours left) is counting down from the date only -- as if tomorrow is the 31st and the deadline is November 31st, seemingly ignoring the end month.

Currently hacked in Optimizely here: 
Treat Yo Friends: https://app.optimizely.com/edit?experiment_id=4029790077
50 Cans: https://app.optimizely.com/edit?experiment_id=4029960026
GGW: https://app.optimizely.com/edit?experiment_id=4027640328

angaither:
Using that will not ignore the month the campaign ends.
Fixes #5856

instead of calculating how many days in that month and adding it all together, the `days` does that for you
image


itsjoekent:
oh that works

:+1:

PR #5951, opened 2015-12-14
Remove unused entity ref hooks

### LINKED ISSUE #5944: remove node reference logic/views
no one uses the image fact node view that pulls all references of where the image/fact is viewed
- [ ] [helper code](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module#L274-L375)
- [ ] [image view code](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_image/dosomething_image.module#L106-L122)
- [ ] [fact view code](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_fact/dosomething_fact.module#L22-L39)

angaither:
These entity reference things were not being used, so :boom: it
fixes #5944


blisteringherb:
:+1:

PR #5953, opened 2015-12-15
Reportback status security refactor

### LINKED ISSUE #5874: Calculation issue with total property in pagination object for API response
image

`retrieved_count` is 25 but the `total` is 0? :confused:

weerd:
Fixes #5874
#### What's this PR do?

The logic to remove restricted statuses has been switched around from being in the **dosomething_reportback_item.query.inc** file to being a private method of the `ReportbackItemTransformer` class. The security measures in place are really only being used by the API, and this also allows for securing the filters earlier on in the transformation process, rather than later when it hits the query builder, allowing the need to process the filters only once instead of multiple times. 

These changes were also required for how the totals are calculated and relying on accurate use of the RB item statuses along with pagination for the gallery.

Most of the magic is happening in the **ReportbackItemTransformer.php**.
#5874

---

@DFurnes @angaither 

cc: @jonuy @aaronschachter


DFurnes:
Cool cool. :ship:

Issue #5929, opened 2015-12-09
Revise API Reportback Item accessibility based on user permissions + status


weerd:
For the mobile app, want to allow users to see their RBs (including pending, flagged, excluded) and other user's public profiles (which includes excluded). Thus the API needs to allow more fine-grained permissions access to reportback items.

Part of the issue seems to be that Northstar uses a single authenticated admin user to hit the Phoenix API, which complicates the capacity to determine proper permissions... although not 100% sure about this.

@aaronschachter @DFurnes

DFurnes:
Can we only authorize against the currently logged-in user, or is it possible to provide a UID with the request?

weerd:
For the currently logged-in user, you'd need to have Northstar authenticate with Phoenix using that user for such calls, but there are other calls that I suspect NS would need to make as an authenticated admin... so it would have to be able to flip-flop between calls to Phoenix as user or admin.

DFurnes:
So the only way to authorize a request is against the current session token? Bummer, that makes things a bit more complicated.

aaronschachter:
Any update on this? If anything, returning all statuses would at least unblock any Profile development in LDT. cc @marahml @jonuy

weerd:
I believe @jonuy had mentioned the possibility of, for the time being, on the app to just have both an NS auth user and then the anon user (possibly managing two tokens) and using the NS auth user to get all the RB items regardless of status... but not sure if that was going to be an attempted experiment first?

jonuy:
To get to our ideal solution for this, it sounds like we'll need to revamp our user authentication so that user's can auth against Northstar and Phoenix in a similar manner.

I think the work involved for that is [here](https://github.com/DoSomething/northstar/issues/202) @DFurnes? And it could take a while until we get to that point.

In the meantime, from the mobile app, we really might just have to separately authenticate with the Phoenix API and manage multiple sessions (Northstar and Phoenix).

DFurnes:
@jonuy Yup, that's probably the best course of action for now.

aaronschachter:
The big blocker right now in mobile app is the self can't see their own reportbacks regardless of status. Isn't this just a bug since the Northstar endpoint the mobile app is using should be authenticated as an admin?

I'm wondering if a compromise would be to just keep the reportbacks public regardless of status for now as far as the public profile, instead of each mobile app maintaining two separate authenticated sessions...

jonuy:
[...]

PR #5954, opened 2015-12-15
Updating show method for RB Item results with some security.

weerd:
Refs #5929
#### What's this PR do?

While the index collection of RB Items from the API is now doing a better job of securing content that should not be accessible to anonymous API requests, the show view for a single RB Item was not. This update adds a method to check whether an RB Item returned via requests to `/reportback-items` should be accessible or not. If the anonymous user does not have access to flagged, excluded, and pending items that are specifically requested by their ID, then an exception is thrown and caught via the normal process within the API request cycle.

In the one file changed, *\* ReportbackItemTransformer.php**.

Not 100% sure on the naming of method in this PR or in #5953, but good enough for now! :sheep: 
#5929

---

@DFurnes @angaither 
cc: @jonuy @aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php
+++ b/lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php
@@ -40,6 +40,7 @@ class ReportbackItemTransformer extends ReportbackTransformer {
   public function show($id) {
     try {
       $reportbackItem = ReportbackItem::get($id);
+      $reportbackItem = $this->removeUnauthorizedResults($reportbackItem);
       $reportbackItem = services_resource_build_index_list($reportbackItem, 'reportback-items', 'id');
     }
     catch (Exception $error) {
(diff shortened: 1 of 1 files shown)


DFurnes:
:+1: :ship:

lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php
```diff
@@ -128,4 +129,29 @@ private function removeUnauthorizedStatuses($statuses) {
...
   }
 
+  /**
+   * Remove Reportback Items with unauthorized statuses based on user permission.
+   *
+   * @param  array $data
+   * @return mixed
+   * @throws Exception
+   */
+  private function removeUnauthorizedResults($data) {
```
angaither, line 139: this looks very similar to the pr you just merged... can it not be the same func?

weerd, line 139: It's different in where the data gets checked. The other one checks the statuses in the `$filters` passed to the query... this one is removing unauthorized items after the query has already run. It depends on whether it's an `index` or `retrieve` method which is why it has to be two different methods.

angaither, line 139: shouldn't the unauthorized items be removed before the query is run?

weerd, line 139: Ideally yes, which is what is happening in the index collection call. But in a retrieve where you hit something like `/reportback-items/1234` you won't know ahead of time if that item is flagged or not since you're requesting a specific one.

angaither, line 139: aye

Issue #5859, opened 2015-12-01
User activity sync script 


chloealee:
I am working on an update to sync user activity data between pheonix and northstar so every time a user signs up or submits a reportback on the web, the mobile device will also be aware of this and all user activity will be up-to-date (https://github.com/DoSomething/phoenix/issues/4571).  

However, this will only sync activity after the code is merged so we'll need to run a script that transfers/syncs campaign activity that happened pre-merge of https://github.com/DoSomething/phoenix/issues/4571.

angaither:
before any of this happens, we need a way to mass migrate this info, without logging in each user.
we will also need to override the `created_at` attributes and submit the timestamps from the signups and reportbacks

### migrate signups
1. will need a way to force send a timestamp
2. timestamp should be converted to mongo date format

```
select u.uid, u.mail, s.sid, s.nid, s.timestamp
from users u
inner join dosomething_signup s on s.uid = u.uid;
```

### migrate reportbacks

```
select u.uid, u.mail, r.rbid, r.nid, r.created, r.quantity, r.why_participated, r.num_participants, rf.fid, f.filename, concat("/sites/default/files/reportbacks/", r.nid , "/", f.filename) as link
from users u
inner join dosomething_reportback r on r.uid = u.uid
left join dosomething_reportback_file rf on r.rbid = rf.rbid
inner join file_managed f on f.fid = rf.fid
```

this above query will NOT work for campaigns that have been closed and the reportback is now associated with a different nid

angaither:
Instead of syncing data to northstar, which will probably get out of sync, we can call to get this info from NS so we are going #6110 route instead

PR #5955, opened 2015-12-15
Reportback api multiple ids

weerd:
Refs #5859
#### What's this PR do?

This update allows for passing specified IDs as a comma separated list to the `/reportbacks` endpoint via the `url` parameter and thus collect a curated list of multiple Reportbacks with all their Reportback Items in tow. Also did a bit of file cleanup to update array syntax :tada: 

At **reportback_resource.inc** and follow the breadcrumb trail from there... :bread: 

This came about from a question related to #5859, regarding getting all reportbacks for a user, along with the respective list of reportback items. If we have a list of Reportback IDs for a user, we can then use this update to request them all in one call.

For example:

```
api/v1/reportbacks?ids=3419,3465,3052
```
#5859

---

@DFurnes @angaither 
cc: @jonuy @aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/reportback_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/reportback_resource.inc
@@ -1,104 +1,104 @@
 <?php
 
 function _reportback_resource_definition() {
-  $reportback_resource = array();
-  $reportback_resource['reportbacks'] = array(
-    'operations' => array(
+  $reportback_resource = [];
+  $reportback_resource['reportbacks'] = [
+    'operations' => [
 
-      'retrieve' => array(
+      'retrieve' => [
         'help' => 'Retrieve a specified reportback.',
-        'file' => array(
+        'file' => [
           'type' => 'inc',
           'module' => 'dosomething_api',
           'name' => 'resources/reportback_resource'
-        ),
+        ],
         'callback' => '_reportback_resource_retrieve',
-        'args' => array(
-          array(
+        'args' => [
+          [
             'name' => 'rbid',
             'description' => 'The rbid of the reportback to retrieve.',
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_api/resources/reportback_resource.inc
```diff
@@ -133,14 +154,12 @@ function _reportback_resource_index($campaigns, $status, $count, $random, $page,
     'page' => $page,
     'load_user' => $load_user,
     'flagged' => $flagged,
-  );
+  ];
 
-  $reportbacks = new ReportbackTransformer;
-  return $reportbacks->index($parameters);
+  return (new ReportbackTransformer)->index($parameters);
```
DFurnes, line 159: Hmm, I feel like putting this as a one-liner reads a bit strangely.

weerd, line 159: I changed it cause I have it like that in the campaigns resource and else where too... if it's not as readable I can change the resources globally. Don't really have a preference other than making all match :)

DFurnes, line 159: Ah, yeah. That's fine, not a merge blocker.

Longer term, this is the kind of thing we'd probably want to pull from an service container anyways if we're greenlit to keep working on the prototype started in #5932. Something like..

``` php
return app('\Phoenix\Reportbacks\ReportbackTransformer')->index($parameters);
```

weerd, line 159: oooooooo :drooling:

DFurnes:
Oh, totally meant to :+1: this. Ship it!

weerd:
LOL

PR #5957, opened 2015-12-15
Exclude reportback items parameter

### LINKED ISSUE #4049: Campaign gallery API endpoint improvements

weerd:
Fixes #4049
#### What's this PR do?

This PR provides the much needed capacity to _exclude_ certain RB items that are specified, from the result set when doing an index collection request to `/reportback-items`.

Start at the **reportback_item_resource.inc** file and  go from there.

Try hitting the `/reportback-items` endpoint and pass the new url parameter `?exclude=1234,5678` and make sure those items don't show up in the results.
#4049

---

@DFurnes @angaither 
cc: @jonuy @aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php
+++ b/lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php
@@ -85,6 +85,7 @@ class ReportbackItemTransformer extends ReportbackTransformer {
   private function setFilters($parameters) {
     $filters = [
       'nid' => dosomething_helpers_format_data($parameters['campaigns']),
+      'exclude' => dosomething_helpers_format_data($parameters['exclude']),
       'status' => $this->removeUnauthorizedStatuses(dosomething_helpers_format_data($parameters['status'])),
       'count' => (int) $parameters['count'] ?: 25,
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php
```diff
@@ -94,6 +95,10 @@ private function setFilters($parameters) {
 
     $filters['offset'] = $this->setOffset($filters['page'], $filters['count']);
 
+    if ($filters['exclude'] && !is_array($filters['exclude'])) {
```
weerd, line 98: Not the cleanest, but the item here _needs_ to always be an array for Drupal to perform the `NOT IN` query condition.

DFurnes:
:boat:

PR #5967, opened 2015-12-21
Remove destination from node edit links.

### LINKED ISSUE #5806: Some Edit Links on Content Search View Don't Work
In the content search view (https://www.dosomething.org/admin/content/search?title=&type=All&field_cause_tid=All&status=All&language=All&page=1), some of the edit links simply refresh the view instead of loading the /edit tab of a campaign. 

For example, experience is borked for "Senseless Attacks" and "Happy Scrappy Holidays" and "Let's DO This" but works for "Paris peace sign" and the wrapping paper waste facts. I can't seem to find a clear rule about what works and what doesn't, but if that rule is key to fixing the bug, I can do some more digging.

image

This is related to https://github.com/DoSomething/phoenix/issues/5114

angaither:
This was blocking/messing with a lot of the global redirect logic.
Fixes #5806

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -569,22 +569,11 @@ function dosomething_helpers_application_path($path = '') {
 }
 
 /**
- * Implements hook_form_node_admin_content_alter().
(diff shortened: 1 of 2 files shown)


weerd:
:+1:

PR #5971, opened 2016-01-04
Fixes undefined property notices on the campaings page.

sergiitk:
Just fixes the undefined property notices.

screen shot 2016-01-04 at 7 50 51 pm

```
Notice: Undefined property: stdClass::$sm_field_call_to_action in dosomething_search_views_pre_render() (line 290 of /var/www/dev.dosomething.org/lib/modules/dosomething/dosomething_search/dosomething_search.module).
```


angaither:
:+1:

PR #5972, opened 2016-01-04
Fixes partners load on campaign pages.

### LINKED ISSUE #5968: Removing Sponsor Logo Results in Error Page
#### Current Behavior

When a campaign's sponsor is removed from the "name" field of a campaign (under edit > sponsors and partners), the "view" tab shows an error like the one shown below, for both logged in admins and unauthorized users.
#### Desired Behavior

Removing content from any previously populated field should not produce an error.
#### Steps to Replicate

Go to a campaign with a sponsor in the sponsors and partners field. Remove by deleting the text in the "name" field. Save the campaign.
#### Why This Matters

Most pressingly, a suite of campaigns is no longer sponsored, and the sponsor's logo needs to be removed.
#### Relevant Screenshots

Field:
image

After content is removed from field:
image

sergiitk:
#### What's this PR do?
- Sets up additional safe checks on partners loading

See #5968

Fixes #5968.

### DIFF
--- a/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
+++ b/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
@@ -150,11 +150,15 @@ function dosomething_taxonomy_get_partners_data($field_partners, $logo_style = '
     if (!isset($item['value'])) {
       continue;
     }
+
     $fc_item = entity_metadata_wrapper('field_collection_item', $item['value']);
+    if (!$fc_item->field_partner->value()) {
+      continue;
+    }
 
     // Gather partner term data.
     $term = taxonomy_term_load($fc_item->field_partner->value()->tid);
-    $term_wrapper = entity_metadata_wrapper('taxonomy_term', $term);
+    $term_wrapper = dosomething_helpers_term_metadata_wrapper($term);
     $partners[$delta]['tid'] = $term_wrapper->getIdentifier();
     $partners[$delta]['name'] = $term_wrapper->label();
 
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
```diff
@@ -150,11 +150,15 @@ function dosomething_taxonomy_get_partners_data($field_partners, $logo_style = '
     if (!isset($item['value'])) {
       continue;
     }
+
     $fc_item = entity_metadata_wrapper('field_collection_item', $item['value']);
+    if (!$fc_item->field_partner->value()) {
+      continue;
```
angaither, line 156: the `continue`s seem a bit strange

angaither:
@sergii-tkachenko based on your [comment](https://github.com/DoSomething/phoenix/issues/5968#issuecomment-168765219), just want to be clear
when an editor wants to remove a partner, instead of removing content in the field itself, the entire field group needs to be "removed"? 

likewise for any other content in a field group?

lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
sergiitk, line 156: Why? Take a look at the lines 150-151.

sergiitk:
@angaither No, this is not correct. Editor should use Remove button in order to remove a lineitem from the collection. However, if an editor just removes name field, it should not fall into a Fatal Error.

sergiitk:
As I understand, @namimody planned to remove entire group, but removed partner's name instead, right, @namimody?

lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
angaither, line 156: I wasn't under the impression we used continue, `¯\_(ツ)_/¯`

angaither:
:+1:

angaither:
@sergii-tkachenko I think the plan was just to remove the name. we had a similar issue with this on the Wednesday before break. I just want to make sure we communicate the correct way to remove sponsors/other field data to campaigns team

PR #5976, opened 2016-01-04
Fix for missing CTAs on Fact pages.

### LINKED ISSUE #5941: 11 Facts CTAs not Appearing on Front End
#### Current Behavior

While CTA content exists on the back end for 11 Facts pages, CTAs do not appear on the front end between facts 5 and 6 or at the bottom of the page. 
#### Desired Behavior

If CTA content exists on the back end of an 11 Facts page, it should appear on the front end of the facts page. If it helps: on Thor v2.0.48, CTAs were working correctly. On prod v2.0.46, they were not working.
#### Steps to Replicate

Look at the back end of an 11 facts page. If it has content in the "Call to Action" field, but the CTA is not showing on the front end, you replicated the bug.
#### Why This Matters

11 Facts get a lot of traffic, and the CTAs are our only conversion points on those pages.
#### Relevant Screenshots

https://www.dosomething.org/facts/11-facts-about-hurricanes
image

image

weerd:
Fixes #5941 
#### What's this PR do?

At some point the expected ['content'] array data changed and no longer houses the CTA information (along with Fact page images), so the Fact page template was breaking on both CTA content and image content. The solution involved adding a bit of new logic to account for  information as well as adjusting where the data was being retrieved from. It now refers to the data in the main  array.#### What's this PR do?

The fix mostly happens in the **dosomething_fact_page.module** so that's probs a good place to :eyes: 

We can do a push to Thor and check to make sure all is hunky dory again!

Tried a few different approaches but this seemed to be the most feasible. I had originally intended to use `dosomething_helpers_extract_field_data()` which worked for most of the fields but threw a wrench with getting the CTA link data, and also when the arrays were structured differently depending on whether it was a US or Global url :rage2: 
#5941

---

@angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_fact_page/dosomething_fact_page.module
+++ b/lib/modules/dosomething/dosomething_fact_page/dosomething_fact_page.module
@@ -73,6 +73,8 @@ function dosomething_fact_page_theme($existing, $type, $theme, $path) {
  * Implements hook_preprocess_node().
  */
 function dosomething_fact_page_preprocess_node(&$vars) {
+  $supported_languages = dosomething_global_get_language_prefix_list();
+
   if ($vars['type'] == 'fact_page') {
     $template_vars = array(
       'text' => array(
@@ -92,9 +94,12 @@ function dosomething_fact_page_preprocess_node(&$vars) {
         $field = $vars["field_{$label}"];
 
         if (isset($field) && !empty($field)) {
-          if (array_key_exists('en', $field)) {
-            $field = $field['en'];
+          $prefix = array_keys($field)[0];
+
+          if (!is_int($prefix) && in_array($prefix, $supported_languages)) {
+            $field = $field[$prefix];
           }
+
           $field = $field[0];
 
           switch ($key) {
@@ -110,6 +115,7 @@ function dosomething_fact_page_preprocess_node(&$vars) {
             default:
               break;
           }
+
         }
 
       }
--- a/lib/modules/dosomething/dosomething_fact_page/dosomething_fact_page.module
+++ b/lib/modules/dosomething/dosomething_fact_page/dosomething_fact_page.module
@@ -96,6 +96,8 @@ function dosomething_fact_page_preprocess_node(&$vars) {
         if (isset($field) && !empty($field)) {
           $prefix = array_keys($field)[0];
 
+          // Depending on language availability, the data we need to grab
+          // may be nested in an associated array using a language prefix.
           if (!is_int($prefix) && in_array($prefix, $supported_languages)) {
             $field = $field[$prefix];
           }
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_fact_page/dosomething_fact_page.module
```diff
@@ -90,22 +89,29 @@ function dosomething_fact_page_preprocess_node(&$vars) {
 
     foreach ($template_vars as $key => $labels) {
       foreach ($labels as $label) {
-        $field = "field_{$label}";
-          if (isset($content[$field])) {
+        $field = $vars["field_{$label}"];
+
+        if (isset($field) && !empty($field)) {
+          if (array_key_exists('en', $field)) {
+            $field = $field['en'];
```
angaither: this could be another language though, correct?

weerd: I thought we aren't translating these? Is it more of a, not right now... but down the line?

angaither: hmm, well you are correct they are not translatable, so all fact pages are currently in `en` but any new ones created would be `en-global` since that's the site default.

angaither: to protect against the #future could we foreach around the [`language_list`](https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/language_list/7)

weerd: Okie doke. So I just need to account for those 2? I can also make it to account for an array of possible languages available. I believe I remember doing something similar before the break... gonna find it in closed PRs

weerd: Yep yep, I did something similar in this PR with what you're suggesting :) https://github.com/DoSomething/phoenix/commit/8ec376c56643d1d4196b1920587e36477a6bb5a3

angaither: :+1:  that's the spirit

lib/modules/dosomething/dosomething_fact_page/dosomething_fact_page.module
```diff
@@ -90,22 +91,33 @@ function dosomething_fact_page_preprocess_node(&$vars) {
...
+        $field = $vars["field_{$label}"];
+
+        if (isset($field) && !empty($field)) {
+          $prefix = array_keys($field)[0];
+
+          if (!is_int($prefix) && in_array($prefix, $supported_languages)) {
+            $field = $field[$prefix];
+          }
+
+          $field = $field[0];
```
weerd, line 105: The above lines (96-103) of logic is not super pretty but gets the job done. I could extract it to a function later if it ends up we have to do similar stuff in other locations.

angaither, line 105: sounds good, maybe some comments would be good here!

angaither:
:+1:

lib/modules/dosomething/dosomething_fact_page/dosomething_fact_page.module
weerd, line 105: Done. :page_facing_up:

PR #5978, opened 2016-01-05
Makes sure that all node edit links are prefixed 

### LINKED ISSUE #5952: clicking view on campaign edit without prefix, causes redirect loop
if an editor navigates from the backend directly to a node/edit link, when clicking `view` on the node they are taken to the global english url and in a redirect loop.
this happens when you have a translation, are using torguard, and you are and editor of that county with user language set to match

when I clicked `view` from `/node/2259/edit/pt-br` I get taken to the english url with a redirect loop
image

itsjoekent:
#### What's this PR do?

Makes sure that the correct prefix is applied to node edit pages

Can you click/goto global edit links, and then view the page without error? 
What about normal edit links?

The node edit logic never cared about prefixes as they aren't required for the correct node edit page. The implications of this however are the cause of #5952 

Fixes #5952

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -97,7 +97,7 @@ function dosomething_global_init() {
     return;
   }
 
-  // Handle translation redirects for authenticated and anonymous users whe n
+  // Handle translation redirects for authenticated and anonymous users when
   // they request bare (Global English) URLs to nodes.
   //
   // Passing 'noredirect=1' will defeat the redirect.
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -566,7 +566,9 @@ function dosomething_global_redirect_node_edit($node, $user_language) {
     $new_url = str_replace(' ', '', '/node/' . $node->nid . '/edit/' . $lang_append);
     drupal_goto($new_url);
   }
-  else if (empty(dosomething_global_get_current_prefix()) && !empty($prefix)) {
+  // If the page has no prefix, add one
+  else if (dosomething_global_get_current_prefix() === '' && !empty($prefix)) {
+    // Add prefix to the current path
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -565,6 +566,9 @@ function dosomething_global_redirect_node_edit($node, $user_language) {
     $new_url = str_replace(' ', '', '/node/' . $node->nid . '/edit/' . $lang_append);
     drupal_goto($new_url);
   }
+  else if (dosomething_global_get_current_prefix() == '' && $prefix != '') {
```
sergiitk: How `dosomething_global_get_current_prefix()` can return an empty string?

itsjoekent: global language

sergiitk: Then better use `===` operand and explain the logic in comment.
Also, `!empty($prefix)`.

itsjoekent: ah, yeah Ill use the empty function

itsjoekent:
@sergii-tkachenko fixed!

sergiitk:
I think it's better to use `=== ''` for the first statement, as I explained in the https://github.com/DoSomething/phoenix/pull/5978#discussion_r48862200.

sergiitk:
And explain in the comment after if that the function returns an empty string for Global language.

itsjoekent:
whats the difference between === ' ' and empty?

sergiitk:
Empty = FALSE = '' = 0 = NULL = "0" = 0.0 = array(), etc...
`=== ''` just equals empty string.

sergiitk:
http://php.net/manual/en/function.empty.php

itsjoekent:
@sergii-tkachenko updated, I think I got what you were suggesting now

sergiitk:
Thanks @deadlybutter! :+1:

Issue #5877, opened 2015-12-02
Associating Multiple Runs with Campaign, Prioritizing One Run to Show


namimody:
**Problem/Requirement:**
Campaigns open and close multiple times. Each time a campaign closes, a run is created, and signups/reportbacks are archived to the run. When the campaign reopens and runs again, we should be able to associate multiple runs with the same campaign. Then, we should be able to choose which one should display when the campaign closes again.

**Feature Summary:**
Currently, multiple runs can be associated with the same campaign, but one cannot choose which run is the most recent for display purposes.

**Current Workaround w/o Feature (if applicable):**
Currently, we have to associate old runs with other random (open) campaigns and associate the correct run with the correct campaign, close the campaign, and then publish the run. This has resulted in many runs randomly and incorrectly associated throughout the site. It is also a confusing experience for editors.

angaither:
this should be a node reference field (perhaps a dropdown) `field_current_run` on the campaign node

the field should display only campaign run types
in the campaign module form alter, we can scope this to display only runs that have the current campaign nid in the field.

angaither:
this field should be required!

sbsmith86:
Should this field (should all fields?) be translatable?

![screen shot 2016-01-05 at 9 56 39 am](https://cloud.githubusercontent.com/assets/1700409/12118109/b9753a00-b392-11e5-86d0-52b4fd53ac32.png)

@angaither

angaither:
@sbsmith86 great question. currently campaign runs are not translatable. So I suppose yes. so we can have different runs for different translations of the node. 

cc @namimody @mikefantini @mnelson338

sbsmith86:
Closed with #5979

PR #5979, opened 2016-01-05
Current Run Field

sbsmith86:
#### What's this PR do?

Creates a new field on campaign nodes that allows the user to select the current run associated with the campaign. Only shows campaign runs that are tied to the campaign nid as options, otherwise it shows all of them. 

A lot of feature export stuff, and then in `dosomething_campaign.module` I handle the form altering to filter the options.

Edit or create a new campaign, and you should see a field "Current Run" under "basic info" that allows you to select a campaign run. You should only see campaign runs that are tied that campaign nid, unless no campaign runs are currently tied to it, in which case you should see all campaign runs as options.

See the ticket.

addresses #5877 -> @angaither i'm not sure just adding this field really fixes the issue. I'm guessing we also need to now use the new field to determine which campaign run to show when the campaign closes. Is there an issue for that?


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -26,6 +26,26 @@ function dosomething_campaign_form_campaign_node_form_alter(&$form, &$form_state
   dosomething_campaign_run_add_campaign_run_info($form, $form_state);
   // Use #after_build to add JS even on validation errors: https://drupal.org/node/1253520#comment-4881588
   $form['#after_build'][] = 'dosomething_campaign_form_campaign_node_form_after_build';
+
+  $nid = $form['nid']['#value'];
```
angaither, line 30: this might not be set on the `node/add/campaign` page, but this hook still runs. might cause some trouble below

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -26,6 +26,26 @@ function dosomething_campaign_form_campaign_node_form_alter(&$form, &$form_state
...
+
+  // Find all the campaign runs associated with this campaign.
+  $campaign_runs = db_query("SELECT entity_id
+                             FROM field_data_field_campaigns
+                             WHERE field_campaigns_target_id = :nid
+                             AND bundle = 'campaign_run';", array(':nid' => $nid))->fetchAll();
```
angaither: this should probably be abstracted to a `function dosomething_campaign_get_runs($nid)`

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
sbsmith86, line 30: oooph good point

angaither:
as far as getting this to be the active run, I think you would be fixing up [this function](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module#L227)

angaither:
:ship:

Issue #5783, opened 2015-11-16
Check CSP (Content-Security-Policy) reports, adjust its settings and turn on


sergiitk:
Now Content-Security-Policy is working in Report-Only mode. After a while, we need to parse its reports to see if we missed anything that needs to be unblocked. Then adjust rules according to that and turn off the report mode.

cc @sheyd

sergiitk:
Remaining tasks before pushing to Prod:
- [ ] Deploy to Staging
- [ ] Create an Optimizely test in for staging that includes use of a frame element
- [ ] Test Optimizely A/B test, and confirm all functionality
- [ ] Coordinate testing with @namimody

DFurnes:
Via Slack, some warnings that are occurring on prod right now:

![screen shot 2015-11-30 at 2 48 33 pm](https://cloud.githubusercontent.com/assets/583202/11482647/fc11780c-9771-11e5-8c34-d29ac2d9d323.png)

sergiitk:
So here's drupal log parse results:

```
dosomething@ip-10-100-35-13:~$ sudo grep "|seckit|" /var/log/syslog | grep -Po "Blocked URI: [^ ]+\. " | cut -d " " -f3 | sed "s/\(.*\).\{1\}/\1/" | cut -d "/" -f3 | sort | uniq -c | sort -nr
  16964 cdn.optimizely.com
  13978 dsrealtimefeed.herokuapp.com
   4939 bam.nr-data.net
   2496 js-agent.newrelic.com
   1155 n131adserv.com
    447 localhost
    403 pixel.yabidos.com
    299 data
    291 www.youtube.com
    209 lofischools.herokuapp.com
    209 cdn3.org
    206 p361.atemda.com
    194 s.atemda.com
    144 partners.cmptch.com
    123 static.cmptch.com
     99 cdncache-a.akamaihd.net
     40 pstatic.bestpriceninja.com
     33 storage.googleapis.com
     29 secure.adnxs.com
     28 service.securesrv12.com
     25 ajax.googleapis.com
     24 pstatic.eshopcomp.com
     24 gateway.zscalerone.net
     24 cdnstats-a.akamaihd.net
     21 jsl.infostatsvc.com
     20 www.tr553.com
     20 istatic.eshopcomp.com
     20 app.bestpriceninja.com
     19 fonts.gstatic.com
     19 2090121062.log.optimizely.com
     18 self
     17 fonts.googleapis.com
[...]

PR #5774, opened 2015-11-13
Enables CSP (Content-Security-Policy) in Report-only mode

sergiitk:
#### What's this PR do?
- Exports default CSP settings to a feature
- Sets up CSP directives
- Enables CSP in Report-only mode
#### Any background context you want to provide?

`default-src: 'self' *.dosomething.org`
`script-src: 'self' 'unsafe-inline' 'unsafe-eval' *.dosomething.org *.google-analytics.com`
`style-src: 'self' 'unsafe-inline' *.dosomething.org`
`img-src: 'self' 'unsafe-inline' *.dosomething.org *.google-analytics.com data:`

The module will log the report data into the watchdog.
#### What are the relevant tickets?

Fixes #5392.

PR #5982, opened 2016-01-05
Extends CSP rules: first batch

### LINKED ISSUE #5977: Seckit blacklisting good URIs
Seckit settings ( https://www.dosomething.org/admin/config/system/seckit ) currently blocking
- https://cdn.optimizely.com
- https://js-agent.newrelic.com
- https://bam.nr-data.net
- https://dsrealtimefeed.herokuapp.com

Super blowing up our logs.

(a) Can we please fix this?
(b) Do we want to enable the dsrealtimefeed connection from prod?

/cc @deadlybutter @sergii-tkachenko @angaither

sergiitk:
#### What's this PR do?
- This PR extends CSP policies with report results discovered in #5783.

Previous set of rules: #5774.
I extended specific rules domain by domain by parsing logs for specific CSP violations, see https://github.com/DoSomething/phoenix/issues/5783#issuecomment-169132423.

Note: This PR is much easier to review commit-by-commit.

Closes #5977.


mshmsh5000:
How did you decide which domains to assign to which `*-src` settings?

sergiitk:
By parsing logs carefully. For example, googleapis.com.
First, I'm checking which non-default rules have been violated:

```
dosomething@ip-10-100-35-13:~$ sudo grep "|seckit|" /var/log/syslog | grep -P "Blocked URI: http(s)?://.*googleapis.com\." | grep -Po "CSP: Directive .* violated\." | cut -d" " -f3 | sort -u
default-src
img-src
script-src
style-src
```

Then, if `default-src` has been violated, I'm checking which _effective_ rules were defaulted to it:

```
dosomething@ip-10-100-35-13:~$ sudo grep "|seckit|" /var/log/syslog | grep -P "Blocked URI: http(s)?://.*gstatic.com\." | grep -P "CSP: Directive default-src" | grep -Po "\[effective-directive\] => [^#]+" | cut -d" " -f3 | sort -u
font-src
media-src
```

So, I allow `*.googleapis.com` in:
- `img-src`
- `script-src`
- `style-src`
- `font-src`
- `media-src`

Asterisk wildcard is used because various subdomains have been requested for googleapis.com, checking here: https://github.com/DoSomething/phoenix/issues/5783#issuecomment-169109043.

mshmsh5000:
Thanks -- this all makes sense.

lib/modules/dosomething/dosomething_settings/dosomething_settings.strongarm.inc
```diff
@@ -52,14 +52,14 @@ function dosomething_settings_strongarm() {
       'checkbox' => 1,
       'report-only' => 1,
```
mshmsh5000, line 53: Is there a reason that this PR keeps Seckit in report-only mode?

sergiitk, line 53: I'd like to give the report-only mode another one shot and see how it changes compared to https://github.com/DoSomething/phoenix/issues/5783#issuecomment-169110035.
Just in case we need to allow one of whitelist domains in another one rule or we simply forgot of something important.

mshmsh5000, line 53: Super, thanks.

mshmsh5000:
:+1:

Issue #4068, opened 2015-02-25
Display reportbacks from past runs in Campaign Gallery


aaronschachter:
Related to #4049

The gallery API will need to grab the relevant promoted+approved images from any Campaign Run nodes which belong to the parent Campaign node, and display those last, after the Campaign nid no longer has any more gallery images to display.

aaronschachter:
Perhaps this should be a parameter you can pass into the gallery endpoint, e.g. `campaigns/[nid]/gallery?all-runs=TRUE`, for any instances where maybe we'd want only reportbacks from the current run...

namimody:
When this is fixed, we should also check that published closed states of campaigns with archived activity continue to show member RBs.

angaither:
we can close this because of #6056

PR #5983, opened 2016-01-05
Campaign Action Page Gallery Refactoring

### LINKED ISSUE #5170: RB Gallery Not Showing Photos Using Correct Logic
The first 6 photos of the gallery ​_should_​ pull from first promoted and then if less than 6 promoted, approved photos.

Additionally, when no photos were promoted on the back end, but were approved, the gallery should pull the approved ones in rather than show placeholder images.

Currently, on Suspended for What, 6 photos are approved and only 5 are showing:
image

### LINKED ISSUE #5980: Remove old Showcase Header Gallery code
The code to add the Showcase Gallery in the header on both campaign pitch and action pages still exists and should be removed; currently the gallery is getting added, but hidden from CSS :scream:

weerd:
Fixes #5980
Fixes #5170
Refs #5874 (partially fixes `total` for RB items; need to double check for campaigns)
#### What's this PR do?

This PR performs a much needed update to the main Campaign Action page gallery, as well as tackling a few bugs/issues that cropped up with the API and main DS app code (that also affect our products using the API) as the updates/refactoring of the gallery were taking place.

Since there was a lot that needed to happen in this PR to switch over the gallery, I attempted to break the process down into [logical commits](https://github.com/DoSomething/phoenix/pull/5983/commits), so it might be best to go by each individual commit instead of checking all the files at once! Baby steps :baby: 

Pull down the branch, `drush cc all` and load up a gallery in a campaign action page. You can also do a few hits to the `/reportback-items` endpoint and make sure you get no errors.
#### Next steps
- More ideal if when caching the total tally for items based on query, we could specify the expiration time for the cached data longer than the default of 24 hours; would need to update ApiCache class to allow passing an expiration time in hours via `set()` method. I refrained for now because there is no way to clear that cache currently... which brings me to...
- On reportback save, update & delete, need to add a call to clear cache for RB item tally with specific campaign id (something like `cache_clear_all('reportback-items-tally_api_request|campaigns=:id', 'cache_dosomething_api', TRUE)`
- Remove **all** the old `gallery` API endpoint code for the old gallery!

---

@angaither @DFurnes 

CC: @mikefantini


lib/modules/dosomething/dosomething_reportback/dosomething_reportback_item.query.inc
```diff
@@ -120,20 +120,31 @@ function dosomething_reportback_build_reportback_items_query($params = array())
 * @return object
 *    An executed database query object to iterate through.
 */
-function dosomething_reportback_get_reportback_items_query($params) {
+function dosomething_reportback_get_reportback_items_query($params, $tally = FALSE) {
```
DFurnes, line 124: Add `$tally` to the docblock? :)

weerd, line 124: :mag_right:

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
```diff
@@ -199,70 +193,66 @@ function paraneue_dosomething_preprocess_node_campaign(&$vars) {
 /**
  * Expands $reportbacks_gallery array with themed Reportback items and data for given node preprocess $vars.
  */
-function paraneue_dosomething_preprocess_reportback_gallery(&$vars) {
-  $reportbacks_gallery = &$vars['reportbacks_gallery'];
-
-  if (!isset($reportbacks_gallery['desired_count'])) {
-    $reportbacks_gallery['desired_count'] = 6;
+function paraneue_dosomething_preprocess_reportback_gallery(&$vars, $count = 6) {
+  $gallery = [];
```
DFurnes: Would this read better to initialize all as one statement, like `$filters` below?

``` php
$gallery = [
  'admin_access' => user_access('view any reportback') ? 'true' : 'false',
  'desired_count' => $count;
  // and so on...
];
```

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
```diff
@@ -89,43 +87,30 @@ define(function(require) {
...
 
-      $images = $cluster.find("img");
+      $images = $cluster.find('img');
 
       $images.unveil(100, function () {
         var $this = $(this);
 
         $this.load(function () {
-          $this.parent(".photo").addClass("-unveiled");
+          $this.parent('.photo').addClass('-unveiled');
```
DFurnes: Should we rename this class to `.is-unveiled` since it's a state class? (Sorry, I know this didn't really change.)

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
weerd: Yeah, good call :+1:

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
weerd: LOL. Sure thing. I can update it!

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
```diff
@@ -147,7 +134,7 @@ define(function(require) {
      * @return {string}  API base url.
      */
     buildApiUrl: function () {
-      return this.apiUrl + "/" + this.nid + "/gallery.json";
+      return this.apiUrl + '?campaigns=' + this.nid + '&status=promoted,approved' + '&exclude=' + this.itemIds + '&count=' + this.requestSize;
```
DFurnes: This would be a great case for a template string, imo.

``` js
return `${this.apiUrl}?campaigns=${this.nid}&status=promoted,approved&exclude=${this.itemIds}&count${this.requestSize}`;
```

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
```diff
@@ -208,34 +169,37 @@ define(function(require) {
...
-          if ((this.currentIndex += this.bundleSize) >= this.itemsTotal) {
-            this.currentIndex = this.itemsTotal;
-          }
-        }
-        else {
-          this.appendEntries(this.currentIndex);
-          this.currentIndex += this.bundleSize;
-        }
+      this.$viewMoreContainer = $('<div class=\'reportback__view-more\'></div>');
+      this.$viewMoreButton = $('<button id=\'view-more-button\' class=\'button button--view-more -tertiary inline-alt-text-color\'>' + buttonText + '</button>');
```
DFurnes: Another good case for a template string, since you can avoid having to escape the `'`s!

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
DFurnes: :japanese_ogre:

DFurnes:
Looks good! And yay for deleting code! One other thing that _could_ be really nice would be to remove the `/* eslint-disable */` from the script touched in this file and fix any linting errors, but that's definitely not merge-blocking.

weerd:
@DFurnes regarding the ES6 update, I sorta started and then decided best to hold off for another PR because I already had a bunch on my plate with this update, and need a refresher on the ES6 changes ;) But I'll create an issue for it for the near future. Now that there's less code in the file, should help do that faster!

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -73,6 +73,55 @@ function dosomething_helpers_convert_date($date, $format = 'ISO-8601') {
...
+            $parameters[$key] = implode(',', $value);
+            break;
+
+      case 'load_user':
+            $parameters[$key] = 'true';
+            break;
+
+      case 'random':
+            $parameters[$key] = 'true';
+            break;
```
angaither, line 110: can you combine the ones that have the same code below?

```
case 'load_user':
case `random`:
    $parameters[$key] = 'true';
     break;
```

weerd, line 110: Is it possible to check the case with:

``` php
case 'load_user' || 'random':
```

?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback_item.query.inc
```diff
@@ -117,23 +117,35 @@ function dosomething_reportback_build_reportback_items_query($params = array())
...
+
+  if ($tally) {
+    $count = 'all';
+  }
+  else {
+    $count = dosomething_helpers_isset($params, 'count', 25);
+  }
 
   if ($count && $count !== 'all') {
     $query->range($offset, $count);
```
angaither, line 140: my brain is bleeding. 
but, I _think_ you could move this `$query->range($offset, $count);` into the above else to avoid checking `if ($tally)` twice
and then the logic would read

```
if ($tally) 
  $count = $all;
  $query->execute;
  return $result ->rowCount()

else {
  the rest with the fetch all
```

weerd, line 140: So I think I did it this way because you could pass `all` as the count without the intention to tally.

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
weerd, line 110: Oh duh, I didn't realize you're doing the pass through in the switch in the example! cooooool! very smart. I'll give it a whirl!

lib/modules/dosomething/dosomething_reportback/dosomething_reportback_item.query.inc
angaither, line 140: kewl, that works!

angaither:
looks like a :ship:

mikefantini:
@weerd this addresses #4068 as well correct?

weerd:
@mikefantini Not currently. Held off on run stuff since that's all getting updated. Will likely happen when the API is updated to account for runs as well.

PR #5999, opened 2016-01-07
Current Run JS toggle

### LINKED ISSUE #5994: JS to require campaign run id
Add JavaScript to force a campaign run relationship before allowing a campaign node to be published 

the field should not be required in the db

sbsmith86:
#### What's this PR do?

Disable the publish option on campaign nodes if no campaign run is selected. Now that we handle the requirement of the current run field with JS, this PR also reverts required option on that field from the database.

**Campaign run selected:**
screen shot 2016-01-07 at 10 06 49 am

Gravy.

**Campaign run not selected:**

screen shot 2016-01-07 at 10 07 36 am

Do not pass go.

On a campaign node edit form, if `- None -` is selected on the current run field, then when you go to publish or change the published state of a campaign, you should see an error message. But, if you select a campaign run, you should be able to do what you need to do.

Fixes #5994 

@DoSomething/front-end @angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -43,6 +43,10 @@ function dosomething_campaign_form_campaign_node_form_alter(&$form, &$form_state
       $filtered_options[$run->entity_id] = $all_options[$run->entity_id];
     }
 
+    // Add empty option to the options array. Helps to ensure that the user
+    // explicity chooses a campaign instead of defaulting to the first option
+    // in the list.
+    $filtered_options = array("_none" => "- None -") + $filtered_options;
     $form['field_current_run'][$language]['#options'] = $filtered_options;
   }
 }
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -43,6 +43,10 @@ function dosomething_campaign_form_campaign_node_form_alter(&$form, &$form_state
       $filtered_options[$run->entity_id] = $all_options[$run->entity_id];
     }
 
+    // Add empty option to the options array. Helps to ensure that the user
+    // explicity chooses a campaign instead of defaulting to the first option
+    // in the list.
+    $filtered_options = array("_none" => "- None -") + $filtered_options;
```
DFurnes, line 49: This might read more clearly with [`array_unshift`](http://php.net/manual/en/function.array-unshift.php)?

sbsmith86, line 49: @DFurnes `array_unshift` reindexes the array keys, and we need those since they are the nids of the runs. When you add arrays, the array keys are preserved.

DFurnes, line 49: oh gotcha, that makes sense. :grimacing: never mind!

DFurnes:
Awwww yissss jquery :muscle: Looks good from a fender perspective!

mshmsh5000:
This is purrrrrty hot.

angaither:
:chompy: :shipit:

PR #6011, opened 2016-01-08
Adds campaign run nid to signups and reportbacks

### LINKED ISSUE #6008: Associate signups/reportbacks to runs
currently when a user signsup or reportsback the campaign nid is stored, the current run nid needs to be stored to this row as well.

itsjoekent:
#### What's this PR do?

Adds campaign run nid to signups and reportbacks
(Did not touch the API)

Is the campaign run NID filled out in signups/reportbacks when you do those respective things? 

Fixes #6008

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -326,8 +326,8 @@ function dosomething_reportback_form_submit($form, &$form_state) {
 
   $campaign = node_load($values['nid']);
   global $user;
-  $language = dosomething_global_get_language($user, $campaign);
-  $values['run_nid'] = $campaign->field_current_run[$language][0]['target_id'];
+  $language_code = dosomething_global_get_language($user, $campaign);
+  $values['run_nid'] = $campaign->field_current_run[$language_code][0]['target_id'];
 
   // If username field is set, this is an admin form.
   $is_admin_form = isset($form_state['values']['username']);
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -324,6 +324,11 @@ function dosomething_reportback_form_validate($form, &$form_state) {
 function dosomething_reportback_form_submit($form, &$form_state) {
   $values = $form_state['values'];
 
+  $campaign = node_load($values['nid']);
+  global $user;
+  $language = dosomething_global_get_language($user, $campaign);
+  $values['run_nid'] = $campaign->field_current_run[$language][0]['target_id'];
+
   // If username field is set, this is an admin form.
   $is_admin_form = isset($form_state['values']['username']);
   // Load uid by selected username.
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -325,8 +325,7 @@ function dosomething_reportback_form_submit($form, &$form_state) {
   $values = $form_state['values'];
 
   $campaign = node_load($values['nid']);
-  global $user;
-  $language_code = dosomething_global_get_language($user, $campaign);
+  $language_code = $language_code = $campaign->language;
   $values['run_nid'] = $campaign->field_current_run[$language_code][0]['target_id'];
 
   // If username field is set, this is an admin form.
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -217,8 +217,8 @@ function dosomething_signup_create($nid, $uid = NULL, $source = NULL, $timestamp
   // @todo: If campaign is unpublished and non-staff $uid, return error.
 
   $campaign = node_load($nid);
-  $language = dosomething_global_get_language($user, $campaign);
-  $run_nid = $campaign->field_current_run[$language][0]['target_id'];
+  $language_code = dosomething_global_get_language($user, $campaign);
+  $run_nid = $campaign->field_current_run[$language_code][0]['target_id'];
 
   $values = array(
     'nid' => $nid,
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -206,8 +206,8 @@ function dosomething_signup_get_query_source() {
  *   The sid of the newly inserted signup, or FALSE if error.
  */
 function dosomething_signup_create($nid, $uid = NULL, $source = NULL, $timestamp = NULL) {
+  global $user;
   if ($uid == NULL) {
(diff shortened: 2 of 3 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -324,6 +324,11 @@ function dosomething_reportback_form_validate($form, &$form_state) {
 function dosomething_reportback_form_submit($form, &$form_state) {
   $values = $form_state['values'];
 
+  $campaign = node_load($values['nid']);
+  global $user;
+  $language = dosomething_global_get_language($user, $campaign);
```
sergiitk: It's better to name this `$language_code` to avoid confusion, as `$language` is global variable.

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -216,9 +216,14 @@ function dosomething_signup_create($nid, $uid = NULL, $source = NULL, $timestamp
   // @todo: If the campaign is closed, return error.
   // @todo: If campaign is unpublished and non-staff $uid, return error.
 
+  $campaign = node_load($nid);
+  $language = dosomething_global_get_language($user, $campaign);
```
sergiitk: Same here

itsjoekent:
@sergii-tkachenko fixed!

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -324,6 +324,11 @@ function dosomething_reportback_form_validate($form, &$form_state) {
 function dosomething_reportback_form_submit($form, &$form_state) {
   $values = $form_state['values'];
 
+  $campaign = node_load($values['nid']);
+  global $user;
+  $language_code = dosomething_global_get_language($user, $campaign);
+  $values['run_nid'] = $campaign->field_current_run[$language_code][0]['target_id'];
```
sergiitk, line 329: What will happen if the field hasn't been translated to the language?

itsjoekent, line 329: oh wait, its the campaign run that gets translated for each, not the field in the campaign? got confused, sorryz

so this should be the nodes source lang, not based on the user lang?

sergiitk, line 329: Hm. Depends on what do you want to achieve here.

itsjoekent, line 329: I'm getting the campaigns current NID.

now that I've rethought about it, I think the procedure is this

campaign blah blah
source lang -> campaign run

campaign run
 en --> end date, title, etc
 global --> etc etc
 etc

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -206,8 +206,8 @@ function dosomething_signup_get_query_source() {
  *   The sid of the newly inserted signup, or FALSE if error.
  */
 function dosomething_signup_create($nid, $uid = NULL, $source = NULL, $timestamp = NULL) {
+  global $user;
   if ($uid == NULL) {
-    global $user;
```
angaither, line 210: i think we will want to keep this the way it was

sergiitk, line 210: Why? I agree with @deadlybutter here. It's a common practice to define globals at the beginning.

angaither, line 210: because you could be passing in a different user id from the func call

itsjoekent, line 210: oh yeah andrea's right here, this is conflicting with the param usage. ill fix fix

sergiitk, line 210: Oh, yes. Good find @angaither, thank you!

itsjoekent:
@sergii-tkachenko ok, i think its good now. changed $language_code to just be $campaign->language

sergiitk:
Are you sure this behaves as expected? Could you please create a campaign in 2 translations and check if `$campaign->language` changes as we think it is. Also, what if campaign run isn't translated to requested language?

itsjoekent:
blocked on this until a decision is made regarding the biz logic

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -210,15 +210,21 @@ function dosomething_signup_create($nid, $uid = NULL, $source = NULL, $timestamp
...
+
   // If a signup already exists, exit.
   if (dosomething_signup_exists($nid, $uid)) { return FALSE; }
 
   // @todo: If the campaign is closed, return error.
   // @todo: If campaign is unpublished and non-staff $uid, return error.
 
+  $campaign = node_load($nid);
+  $language_code = $campaign->language;
+  $run_nid = $campaign->field_current_run[$language_code][0]['target_id'];
```
weerd, line 222: You could potentially use this the field extractor function?

``` php
dosomething_helpers_extract_field_data($campaign->field_current_run, $campaign->language);
```

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -324,6 +324,10 @@ function dosomething_reportback_form_validate($form, &$form_state) {
 function dosomething_reportback_form_submit($form, &$form_state) {
   $values = $form_state['values'];
 
+  $campaign = node_load($values['nid']);
+  $language_code = $language_code = $campaign->language;
+  $values['run_nid'] = $campaign->field_current_run[$language_code][0]['target_id'];
```
weerd, line 329: Same as below comment to this one:

``` php
dosomething_helpers_extract_field_data($campaign->field_current_run, $campaign->language);
```

weerd:
@deadlybutter @angaither I pulled down this branch and merged it with the work I'm doing. However, I'm already making a bunch of changes which include the code in this PR. I might recommend closing this one and waiting for the upcoming PR to avoid a slew of merge conflicts? It's not a huge deal if there are conflicts, because we can fix them, but figured I'd mention it :)

angaither:
@weerd I would rather add this in and change it later. I really want to keep the pull requests as small as possible. What did you change here? 

I'm also ok with merging this in now if that makes things easier

weerd:
Had to move the variables around cause they're needed earlier in the function, and a couple other things. Yeah, can we merge this now and I can just rebase the shiz and fix any conflicts :)

Just wasn't sure how long we'd need to hold off on this being merged. But if we can merge now that'd be awesome.

weerd:
@deadlybutter :+1:

itsjoekent:
:boom:

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
sergiitk, line 329: There will be a notice here if `$campaign->field_current_run` doesn't have a `$language_code` key. Please handle this case properly.

Issue #6016, opened 2016-01-12
Translate title field on campain run node


angaither:
Use the title module to translate the title of campaign run nodes

refs #5987

sergiitk:
Technical notes:
- Replace title builtin with field provided by title module
- Save it to the features
- Create a helper to replace title values. See `ds-global`

itsjoekent:
field exported, going to make the helper now

PR #6026, opened 2016-01-12
campaign status field TAKE 2

### LINKED ISSUE #5989: update campaign status field
- disable the field in the ui
- auto calculate the field on campaign run node save
- create a cron job to auto calculate the field 
- make the field translatable

`field_campaign_status` would be `open` when the run has a start date and no end date
`field_campaign_status` would be `closed` when the run end date is before `now`

sbsmith86:
#### What's this PR do?
- Disables the campaign status field on campaign nodes so editors can not use it.
- Makes the campaign status field on campaign nodes translatable.
- Adds a presave hook to campaign runs that will update campaign status when a user saves a campaign run node.
- Creates a helper function that takes a start and end date and determines if a campaign is active or not
- Slightly outside of the scope of this issue, but I was getting some preprocess errors on pages without subtitle and partners fields that I fixed up.
- Creates a cron job that updates the campaign status field based on run dates.

The main business is in these here files, there rest is feature export stuff.

`dosomething_campaign_run.module`
`dosomething_helpers.module`
`dosomething_campaign_run.cron.inc`

Create or edit a campaign run, and run dates, hit save. Go to the campaign you tied the run to, the campaign status field should be disabled, but you should be able to see the current status of the campaign per translation. 

Status would be `active` when the run has a start date and no end date
Status would be `closed` when the run end date is before `now`

The cron job is a bit trickier to test, it really only comes in handy when there is a campaign run dates set in the future or past. The cron job should update the status of the campaign node to closed or active based on the dates without a use having to manually save the campaign run again.

Campaign runs...amirite?

Fixes #5989

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module
+++ b/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module
@@ -5,6 +5,7 @@
  */
 
 include_once 'dosomething_campaign_run.features.inc';
+include_once 'dosomething_campaign_run.cron.inc';
 
 /**
  * Adds loaded closed campaign_run node to the given parent campaign's $vars.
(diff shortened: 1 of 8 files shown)


sbsmith86:
@angaither sorry I opened a new PR cause I got deep in a git hell hole on that other one.

sbsmith86:
@angaither This PR has been updated to use a new query in the cron job that uses the `entity_translation` table. 

I also had to make a slight change to the update hook that @chloealee wrote. The date format for the generated nodes was causing issues in the presave hook (to update campaign status) that gets called when campaign runs get saved. The update hook erred out and no runs were generated. 

I updated it to use the same date format drupal would store the date as if you filled out those fields through the CMS (using `-` instead of `/`). Apparently the `DateTime` class can't create dates from strings in the format `y/m/d` because if there are slashes it assumes it is in the American format `d/m/y'.

lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module
```diff
@@ -296,3 +297,32 @@ function dosomething_campaign_run_form_campaign_run_node_form_alter(&$form, &$fo
...
+    // Determine the active and closed state for each translation of the run_date field
+    foreach ($run_dates as $lang => $date) {
+      $start = new DateTime($date[0]['value']);
+      $end = new DateTime($date[0]['value2']);
+      $status = dosomething_helpers_get_campaign_status($start, $end);
+      $run_campaigns = $node->field_campaigns[$lang];
+
+      // Update the campaign status field on each campaign node tied to this run.
+      // There could be multiple campaigns per campaign run. Probably not, but we do allow for it.
+      foreach ($run_campaigns as $campaign) {
```
angaither, line 318: since this little chunk is the same here and in the cron, it could be a helper function `_dosomething_campaign_set_campaign_status`

angaither:
one comment, not needed for merge :+1: 
but let's hold off on merging until the fix for #6016 is in

sbsmith86:
@angaither we can merge this now, si?

sbsmith86:
@angaither I made an update to this that only updates the status if the current run is being saved. see [dbd09f9](https://github.com/sbsmith86/dosomething/commit/dbd09f9bc9cf0e923e13278210507d4a32369068)

Let me know when this is good to merge

angaither:
:+1:

PR #6018, opened 2016-01-12
Campaign run cleanup

chloealee:
#### What does this PR do?

Creates campaign run nodes for all active campaigns (published or unpublished).  
#### How should this be manually tested?

When ssh'd in your vagrant environment, run `drush updb -y` in `dosomething@dev:/var/www/dev.dosomething.org/html$`

Note: this has not yet been tested in staging/Thor
#### What are the relevant tickets?

Fixes #5961

PR #6034, opened 2016-01-13
Revert "Campaign run cleanup"

angaither:
Reverts DoSomething/phoenix#6018

until the fix for #6016 is in place


chloealee:
:+1:

PR #6038, opened 2016-01-13
Updating database schema for ds_signups & ds_reportbacks

### LINKED ISSUE #5993: un-unique uid/nid combo in signup and reportback tables
Currently the uid/nid combo in both the `dosomething_signup` and `dosomething_reportback` tables need to be unique

we need to update this combo to be `uid`/`nid`/`run_id`

weerd:
Refs #5993
#### What's this PR do?

This updates the run_nid on both dosomething_signup and dosomething_reportback tables to act as an index and a combination unique key.

In the respective _**.install_\* files for each DB schema.

Take a gander and confirm this all looks alright and makes sense. You can test it by pulling it down and running `drush updb` to have the database changes take place. Then take a look at the DB and make sure the indexes and keys are updated for the **dosomething_signup** and **dosomething_reportback** table structures.
#5993

---

@angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
@@ -92,9 +92,13 @@ function dosomething_reportback_schema() {
       ),
     ),
     'primary key' => array('rbid'),
+    'unique keys' => array(
+      'rbid-uid-nid-run_nid' => array('rbid', 'uid', 'nid', 'run_nid'),
+    ),
     'indexes' => array(
       'uid' => array('uid'),
       'nid' => array('nid'),
+      'run_nid' => array('run_nid'),
     ),
   );
   $schema['dosomething_reportback_log'] = array(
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
```diff
@@ -92,9 +92,13 @@ function dosomething_reportback_schema() {
       ),
     ),
     'primary key' => array('rbid'),
+    'unique keys' => array(
+      'rbid-uid-nid-run_nid' => array('rbid', 'uid', 'nid', 'run_nid'),
```
angaither: I'm pretty sure you don't need to add `rbid` here since it's serial, and the primary key..it's already unique

weerd: ah! good point.

weerd: I think it may require it:

```
SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '1702838-1227-0' for key 'uid-nid-run_nid'
```

since all the `run_nid`'s are currently `0`... it should have worked though, so not sure :confounded:

angaither: from slack

> ```
> select rb.rbid, rb.uid, rb.nid, count(*)
> from dosomething_reportback rb
> group by rb.nid, rb.uid
> having count(*) > 1;
> ```

this proves there are multiple reportbacks with the combo of same nid/uid in prod/thor currently

so we can't have these be unique (at least not yet)

angaither:
:+1: merge away

PR #6037, opened 2016-01-13
Sets author to campaign run nodes

chloealee:
#### What does this PR do?

Sets author to campaign run nodes. 
#### What are the relevant tickets?

Nodes created in https://github.com/DoSomething/phoenix/pull/6018

fixes #5961 with the help of #6016

PR #6042, opened 2016-01-14
Revert "Sets author to campaign run nodes"

angaither:
Reverts DoSomething/phoenix#6037


angaither:
title field per language was not set correctly.

Issue #5961, opened 2015-12-16
Create campaign runs in a script


angaither:
Create campaign run nodes for all open campaigns (published or unpublished)

angaither:
@mikefantini we will need to know what we should set the open date to, as well as what the title of the nodes should be

angaither:
@mikefantini When we automatically create runs for each campaign, we should create one run per campaign, with multiple translations, correct?
as many translations as the original content?

mikefantini:
Re: translations yes one run per campaign with multiple translations.

This may get tricky though re: title because the title should be in the actual language of the translation. A bit behind at the moment but I'll get you answers re: title and date first thing tomorrow.

mikefantini:
Ok so let's do this:

Title:
- Make the title of the campaign run "campaign_title run 1". Let's do the translated title for each translation of run (so "Limpieza de calles por la tortuga marina run 1" for the Mexican Spanish run of node 3624). This will require #6016 to be done first.
- Since many campaigns we're writing this script for will be on their second (or third) run, I will manually update these nodes.
- Let's update the user profile to pull in the campaign title rather than the run title so that run titles are no longer user facing.

Open Date:
- Let's go with date node was created.
- For campaigns that will be on their second (or third) run I'll update the date manually.

PR #6043, opened 2016-01-14
adds title to also be translated in campaign run nodes 

chloealee:
#### What does this PR do?

Adds title field to be also translated in campaign runs (was forgotten about in #6037)

In your vagrant, run `drush updb -y` in `dosomething@dev:/var/www/dev.dosomething.org/html$`
#6037 and #5961


angaither:
:+1:

PR #6045, opened 2016-01-14
Fixes broken endpoint when no parameters passed.

### LINKED ISSUE #6044: Campaign endpoint failing when not cached
This is due to the location of where a variable is set.

weerd:
Fixes #6044
#### What's this PR do?

Turns out I had moved the  variable generation inside a scoped logic block; if the logic didn't pass then the variable wasn't set and it's being used a little farther down in the code so it needs to be set regardless. Moved it up and out of the logic block. Works fine now.

Just take a look at where I moved the variable definition to :)
#6044

---

@DFurnes @aaronschachter @angaither 

cc: @jonuy


DFurnes:
:shipit:

PR #6053, opened 2016-01-15
Adds helper function to get campaign runs

itsjoekent:
#### What's this PR do?

Adds a helper function to get campaign runs

I currently have a test line of code in the dosomething_global.module file so you can play around with it real quick @weerd if you need it

This follows the business logic I agreed upon a few days ago with Fantini/Andrea/Etc

Technically none?
#### Questions:

@weerd do you want me to just merge in the helper function, or should I go back and update the reportback & signup form stuff to use this as well (Or are you doing that?)

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -36,6 +36,7 @@ function dosomething_global_init() {
   $url_variables = dosomething_global_get_url_variables();
   $user_variables = dosomething_global_get_user_variables();
   $node_variables = dosomething_global_get_node_variables();
+  dosomething_helpers_getcampaign_run($node_variables['nid']);
 
   // Never redirect POST requests
   if ($url_variables['request_method'] == 'post') {
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -36,7 +36,6 @@ function dosomething_global_init() {
   $url_variables = dosomething_global_get_url_variables();
   $user_variables = dosomething_global_get_user_variables();
   $node_variables = dosomething_global_get_node_variables();
-  dosomething_helpers_get_campaign_run($node_variables['nid']);
 
   // Never redirect POST requests
   if ($url_variables['request_method'] == 'post') {
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -36,7 +36,7 @@ function dosomething_global_init() {
   $url_variables = dosomething_global_get_url_variables();
   $user_variables = dosomething_global_get_user_variables();
   $node_variables = dosomething_global_get_node_variables();
-  dosomething_helpers_getcampaign_run($node_variables['nid']);
+  dosomething_helpers_get_campaign_run($node_variables['nid']);
 
   // Never redirect POST requests
   if ($url_variables['request_method'] == 'post') {
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -501,8 +501,8 @@ function dosomething_helpers_floor_decimal($value, $decimals = 1) {
  *
  * @param String category The category of the event
  * @param String action The action being performed
- * @param String label The event label (Optional)
- * @param String value The value to be sent (Optional)
+ * @param (optional) String label The event label
+ * @param (optional) String value The value to be sent
  */
 function dosomething_helpers_add_analytics_event($category, $action, $label = NULL, $value = NULL) {
   $event_string = '"' . $category . '"' . ', ' . '"' . $action . '"';
@@ -636,7 +636,7 @@ function dosomething_helpers_get_current_language_content_code() {
  * Get the filesystem path relative to the application root.
  *
  * @param $path
- *  Optional subdirectory to get the path of.
+ *  (optional) subdirectory to get the path of.
  *
  * @return string
  *  Filesystem path
@@ -702,10 +702,10 @@ function dosomething_helpers_get_campaign_status($start_date, $end_date) {
  *
  * @param int $nid
  * @param object $user
- *  Optional. Will default to global $user if not supplied
+ *  (optional) Will default to global $user if not supplied
  *
- * @return node $run
- *  Returns the campaigns node in the correct language.
+ * @return
+ *  Returns the campaigns node in the correct language or false.
 */
 function dosomething_helpers_get_campaign_run($nid, $user = NULL) {
   // Setup variables
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -696,3 +696,41 @@ function dosomething_helpers_get_campaign_status($start_date, $end_date) {
     return 'closed';
   }
 }
+
+/**
+ * Gets the current run for a given campaign nid and user.
+ *
+ * @param int $nid
+ * @param object $user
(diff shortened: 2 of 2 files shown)


weerd:
@deadlybutter yeah, I'm handling that :) Just merge when this gets approved and I'll use it's power!

lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -36,6 +36,7 @@ function dosomething_global_init() {
   $url_variables = dosomething_global_get_url_variables();
   $user_variables = dosomething_global_get_user_variables();
   $node_variables = dosomething_global_get_node_variables();
+  dosomething_helpers_getcampaign_run($node_variables['nid']);
```
weerd: can we add an underscore between `get` and `campaign`? :grimacing:

itsjoekent:
ok, ill remove the test line then and cc:ing @sergii-tkachenko to review as well

lib/modules/dosomething/dosomething_global/dosomething_global.module
weerd: actually, you have an underscore in the function def. So how is the above one working?

itsjoekent: i did add a space here but i guess it didnt commit

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -696,3 +696,41 @@ function dosomething_helpers_get_campaign_status($start_date, $end_date) {
...
+ * Gets the current run for a given campaign nid and user.
+ *
+ * @param int $nid
+ * @param object $user
+ *  Optional. Will default to global $user if not supplied
+ *
+ * @return node $run
+ *  Returns the campaigns node in the correct language.
+*/
+function dosomething_helpers_get_campaign_run($nid, $user = NULL) {
```
angaither, line 710: i think this is similar to a function @sbsmith86 wrote

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -696,3 +696,41 @@ function dosomething_helpers_get_campaign_status($start_date, $end_date) {
...
 }
+
+/**
+ * Gets the current run for a given campaign nid and user.
+ *
+ * @param int $nid
+ * @param object $user
+ *  Optional. Will default to global $user if not supplied
+ *
+ * @return node $run
```
sergiitk: That's not the right way to declare return.
See example here https://github.com/drupal/drupal/blob/7.x/modules/node/node.module#L960-L961 and here https://www.drupal.org/node/1354#return.

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
sbsmith86, line 710: I updated this function to use the new `field_current_run` but it looks like @deadlybutter's function takes into account business logic differently. Also it returns the whole run node, not just the nid

https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module#L224

I think we should just pick one to use, and make updates where we need to use the correct function to pull current run.

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -696,3 +696,41 @@ function dosomething_helpers_get_campaign_status($start_date, $end_date) {
     return 'closed';
   }
 }
+
+/**
+ * Gets the current run for a given campaign nid and user.
+ *
+ * @param int $nid
+ * @param object $user
+ *  Optional. Will default to global $user if not supplied
```
sergiitk: Optional — should be lowercase and in parentheses, see https://www.drupal.org/node/1354#param

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
weerd, line 710: :scream:

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -696,3 +696,41 @@ function dosomething_helpers_get_campaign_status($start_date, $end_date) {
...
+  $run_nid = dosomething_helpers_extract_field_data($node->field_current_run, $user_language);
+
+  // If not try the nodes source language
+  if (!isset($run)) {
+    $node_language = $node->language;
+    $run_nid = dosomething_helpers_extract_field_data($node->field_current_run, $node_language);
+
+    // Reutrn NULL if the campaign doesnt have a run nid
+    if (!isset($run_nid)) {
+      return NULL;
```
sergiitk: It's better to return `FALSE` to be consistent with `node_load()` (as you return node_load result below).

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
itsjoekent, line 710: I dont believe `function dosomething_campaign_run_get_closed_run_nid($nid) {` will work under some of the edge cases Fantini has described, which is why I did a lot of seemingly redundant checks in this.

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
itsjoekent: ah, didnt know that was convention. will do!

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
itsjoekent: I fixed the other uses of optional within this file as well

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
sbsmith86, line 710: @deadlybutter cool we can go with this and update stuff that is using `dosomething_campaign_run_get_closed_run_nid($nid)` to use this function instead.

itsjoekent, line 710: @weerd should I do this? I dont think it interferes with your stuffs

itsjoekent:
@sergii-tkachenko good to merge?

sergiitk:
Looks good! :+1:

itsjoekent:
hooray!

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -696,3 +696,41 @@ function dosomething_helpers_get_campaign_status($start_date, $end_date) {
...
+  if (!isset($user)) {
+    global $user;
+  }
+
+  // First check if we have a run in the user lang
+  $user_language = dosomething_global_get_language($user, $node);
+  $run_nid = dosomething_helpers_extract_field_data($node->field_current_run, $user_language);
+
+  // If not try the nodes source language
+  if (!isset($run)) {
```
weerd, line 722: How is this checking if `$run` is set, if it doesn't get assign beforehand? It will always fail won't it?

sergiitk, line 722: Good point!

weerd, line 722: turns out it's supposed to be `$run_nid`... just chatted w/ @deadlybutter

sergiitk, line 722: Did you mean to check `$run_nid` there @deadlybutter?

weerd, line 722: @sergii-tkachenko beat you to the post by like a millisecond ;)

sergiitk, line 722: Haha, yes, we sent our comments literally at the same second.

Issue #6054, opened 2016-01-15
Update any create/edit/delete functions for SignUps & Reportbacks to account for run_nid


weerd:
Refs #5993

Need to go through the signup and reportback processes (DS App & API) and update each function to now also account for the new `run_nid` and the unique combinations with `uid` and `nid` when creating entitites.

angaither:
refs #6004

Issue #5991, opened 2016-01-06
add run id to api responses 


angaither:
The `field_current_run` id needs to be exposed in the api in the following responses 
- [ ] ~~get signups~~
- [x] get reportbacks
- [x] get campaigns

weerd:
@angaither Would the `current_run` for a reportback actually be the current run or the run it's actually associated with?

angaither:
whaaa

weerd:
If you're doing a GET request for `/reportbacks` what does "current_run" mean in that context? Wouldn't it just be "campaign_run" that the reportback corresponds to? I don't think it has anything to do with what the current run is... the only place that matters with a reportback is when submitting one and needing to apply the current run id to that uploaded file.

angaither:
you are correct good sir!

angaither:
this can be closed now, right?

weerd:
Not yet. I need to update the old GET signups endpoint

weerd:
working on it...

weerd:
@aaronschachter is the GET `/signups` endpoint used anywhere? I can't find any documentation on it in the API wiki, so not sure what parameters it's implying I can pass, but all I get is:

``` json
[
  "Missing required argument parameters"
]
```

when I hit `/api/v1/signups`

aaronschachter:
Kill it with fire. It's code I added before we were using Northstar for the mobile app for Signup Groups, which we've since abandoned:  https://github.com/DoSomething/phoenix/pull/4243

weerd:
@angaither I think we can close this one now. The only signup related stuff w/ `run_nid` happens in `/campaigns/:id/signup` and the `run_nid` stuff accounted for in the function called `dosomething_signup_create()` from within the endpoint callback :)

PR #6060, opened 2016-01-19
Campaign class updates

weerd:
Refs #5991
Refs #6054
#### What's this PR do?

Foundation work that will benefit resolving the above issues. This PR expands on the `Campaign` class to add properties for `language`, `translation`, and `campaign_runs`. This information is also now available via the API.

Start with the **Campaign.php** file and then move onto the **Transformer.php**.

Pull down branch, clear cache and hit the API asking for a few specific campaigns and see if the new properties show up.

@joe I updated the name of `dosomething_helpers_get_campaign_run()` to `dosomething_helpers_get_current_campaign_run_for_user()` to more accurately reflect the purpose of the function. Based on work I did for this PR, it seems like there will be other such functions that are similar but don't rely on need for user's language.
#5991
#6054

---

@DFurnes @angaither 

cc: @deadlybutter @jonuy

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -190,6 +190,7 @@ abstract class Transformer {
     $output = [
       'id' => isset($data->id) ? $data->id : $data->nid,
       'title' => $data->title,
+      'campaign_runs' => $data->campaign_runs,
     ];
 
     // If an instance of Campaign class, then there is much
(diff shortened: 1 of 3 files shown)


itsjoekent:
re: dosomething_helpers_get_current_campaign_run_for_user

yeah, if the need arises for more of these the core logic of it can be extracted into something else, and the "run for user" will just automatically supply the language of the user based on the node like it does now.

weerd:
@deadlybutter yeah, that sounds good ;)

DFurnes:
:ship: :boat: :rocket:

PR #6061, opened 2016-01-19
Fix campaign view based on status

### LINKED ISSUE #6057: Campaign status and actual view of campaign not matching up
#### Current Behavior

If a campaign has a run associated with it that has an end date that has already passed, then the campaign status will be set to closed and the view of the campaign will be in the closed state. If I change the end date on the run so that it is a date in the future, the campaign status will be set to active. However, the view of the campaign remains in the closed state.
#### Desired Behavior

If campaign status = active, then the campaign view should be in the active state.
#### Steps to Replicate
1. Go to a campaign that is closed
2. Edit the end date of the current run to a date in the future
3. Verify that the status of the campaign is now set to active
4. Verify that despite step 3, the view of the campaign is still in the closed state.
#### Why This Matters

This prevents campaigns from changing state from active to closed which has a big impact especially on sponsored campaigns.

Comeback Clothes on thor: https://thor.dosomething.org/us/node/362/edit

Campaign status = active
image

View = closed state still
image

Comeback Clothes run 2 on thor: https://thor.dosomething.org/us/node/5439/edit

image

sbsmith86:
#### What's this PR do?

Fixes an issue where the campaign closed/active state wasn't being reflected on the node view. So closed campaigns were still showing the active view instead of the closed state. The issue arose because we made the campaign status field translatable, so now we have to grab the value based on the current language instead of the default undefined drupal language. 

`dosomething_campaign.module`

Set a campaign translation to closed by updating the run dates on it's current run (in the same language) to be in the past, then go to view the campaign. You should see the closed state. Set the dates to be current and the campaign status should switch to active and when you view the campaign, you should see it in it's active state.

Fixes #6057

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -856,8 +856,12 @@ function dosomething_campaign_is_pitch_page($node) {
  * @return bool
  */
 function dosomething_campaign_is_closed($node) {
-  if (isset($node->field_campaign_status[LANGUAGE_NONE][0]['value'])) {
-    return $node->field_campaign_status[LANGUAGE_NONE][0]['value'] == 'closed';
+  // Get the language.
+  global $user;
+  $language = dosomething_global_get_language($user, $node);
+
+  if (isset($node->field_campaign_status[$language][0]['value'])) {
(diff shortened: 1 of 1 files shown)


angaither:
:+1:

angaither:
also should fix #6068

PR #6063, opened 2016-01-19
Update to fix up Campaign::get() to return logical data.

weerd:
Refs #6054
#### What's this PR do?

This is a quick update to the returned data from Campaign::get() static method. It allows passing single string or multiple ids as an array, however if passing a single string it makes it a bit cumbersome to remember to pluck the first item out as the actual data you want to use. This update fixes that so if you pass a single string id, it'll return the Campaign object, if you pass an array of ids it'll return a collection of campaign objects. Related to work for campaign runs, since the **signup** and **reportback** functions are calling `Campaign::get()` in a few areas.

Review both the **Campaign.php** and **CampaignTransformer.php** files to make sure code looks hunky dory.

I updated the **dosomething_campaign_hot_shares.cron.inc** file where the call to `Campaign::get()` for a single item is being used.
#6054

---

@DFurnes 

cc: @jonuy @aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/includes/CampaignTransformer.php
+++ b/lib/modules/dosomething/dosomething_campaign/includes/CampaignTransformer.php
@@ -62,7 +62,7 @@ class CampaignTransformer extends Transformer {
   public function show($id) {
     try {
       $campaign = Campaign::get($id, 'full');
-      $campaign = services_resource_build_index_list($campaign, 'campaigns', 'id');
+      $campaign = array_pop(services_resource_build_index_list([$campaign], 'campaigns', 'id'));
     }
     catch (Exception $error) {
       return [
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_campaign/includes/CampaignTransformer.php
```diff
@@ -62,7 +62,7 @@ public function index($parameters) {
   public function show($id) {
     try {
       $campaign = Campaign::get($id, 'full');
-      $campaign = services_resource_build_index_list($campaign, 'campaigns', 'id');
+      $campaign = array_pop(services_resource_build_index_list([$campaign], 'campaigns', 'id'));
```
weerd, line 65: _note_: the `services_resource_build_index_list()` seems to require passing object within an array for it to function as before...

DFurnes:
Seems good to me! :ship:

PR #6066, opened 2016-01-21
Signup function updates

weerd:
Refs #6054
#### What's this PR do?

This PR updates all the signup related functions `dosomething_signup_create()`, `dosomething_signup_delete_signup()`, `dosomething_signup_exists()` and whatever other signup functions call these functions to make sure we now take account of the Campaign Run ID, and pass/accept the right arguments within these functions.

Probably best to go by [commit order](https://github.com/DoSomething/phoenix/pull/6066/commits) to see the updates in a logical order.

Try signing up for a campaign with an account that isn't signed up for said Campaign. Then check the `dosomething_signup` table in the database, sort by **timestamp** and see if that latest signup has an actual `run_nid` filled in!

I also updated the `dosomething_helpers_get_current_campaign_run_for_user()` function to use the `Campaign::get(:id)` method when loading a campaign/node data. Trying to move towards using the static method to load the data so that there it complete parity between loading Campaigns in the DS App backend or for the API (same source of data!).
#6054

---

@angaither @DFurnes 

CC: @deadlybutter

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -700,37 +700,40 @@ function dosomething_helpers_get_campaign_status($start_date, $end_date) {
 /**
  * Gets the current run for a given campaign nid and user.
  *
- * @param int $nid
- * @param object $user
- *  (optional) Will default to global $user if not supplied
- *
- * @return
- *  Returns the campaigns node in the correct language or false.
+ * @param  int     $nid
+ * @param  object  $user
+ * @param  object  $campaign
+ * @return bool|mixed  Returns the campaigns node in the correct language or false.
+ * @throws Exception
 */
-function dosomething_helpers_get_current_campaign_run_for_user($nid, $user = NULL) {
-  // Setup variables
-  $node = node_load($nid);
+function dosomething_helpers_get_current_campaign_run_for_user($nid, $user = NULL, $campaign = NULL) {
+  if (!isset($campaign)) {
+    $campaign = Campaign::get($nid);
+  }
+
   if (!isset($user)) {
     global $user;
   }
 
-  // First check if we have a run in the user lang
-  $user_language = dosomething_global_get_language($user, $node);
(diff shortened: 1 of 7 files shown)


weerd:
Huh?

itsjoekent:
I must be on some crappy Github A/B test so I can't reply to the above comment chain, but Diego is right as I spoke with him before about `global_lang_code` being removed

DFurnes:
Seems reasonable to my eyes. :boat: :eyes:

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -700,37 +700,40 @@ function dosomething_helpers_get_campaign_status($start_date, $end_date) {
...
   if (!isset($user)) {
     global $user;
   }
 
+  $node = $campaign->getNodeData();
+
   // First check if we have a run in the user lang
-  $user_language = dosomething_global_get_language($user, $node);
-  $run_nid = dosomething_helpers_extract_field_data($node->field_current_run, $user_language);
+  $userLanguage = dosomething_global_get_language($user, $node);
```
angaither, line 721: `$user_langauge`

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -700,37 +700,40 @@ function dosomething_helpers_get_campaign_status($start_date, $end_date) {
...
+  $run = dosomething_helpers_isset($campaign->campaign_runs['current'], $userLanguage);
 
-  // If not try the nodes source language
-  if (!isset($run_nid)) {
-    $node_language = $node->language;
-    $run_nid = dosomething_helpers_extract_field_data($node->field_current_run, $node_language);
+  // If not try the Campaign's source language
+  if (!isset($run)) {
+    $campaignLanguage = $campaign->language['language_code'];
```
angaither, line 726: no :camel:s  plz

weerd, line 726: Boo! I like :camel:s. Aight, I'll update for next PR

PR #6076, opened 2016-01-22
Vagrant: add access to RabbitMQ admin.

sergiitk:
See http://dev.dosomething.org:15672:
image

cc @DeeZone


angaither:
woah

itsjoekent:
:mindblown:

sergiitk:
It already was sitting there for a while :)

deezone:
`15672` is the RabbitMQ Management Plugin port. The UI as well as API calls go through the same port. Pointing to `/admin/config/services/message-broker-producer` should also work as it gets Rabbit health stats via this port.

Nice :+1:

PR #6080, opened 2016-01-25
Reportback caption

### LINKED ISSUE #5807: RB Photo Caption Cannot Be Edited
Sometimes, campaign leads need to edit campaign reportbacks. They do so using this view (https://www.dosomething.org/reportback/87623/edit). While the impact # and "response to "why is this campaign important to you?" can be edited and saved, the caption can be edited, but the changes do not save.

chloealee:
#### What does this PR do?

Allows for admins to edit campaign reportback captions and have them save correctly. 

`dosomething_reportback.forms.inc` - `dosomething_reportback_form_submit` function, lines 337-229 -> `dosomething_reportback.module` - line 693 -> line 699 

In your own account, sign in as an admin. Reportback for a campaign. Edit reportback submission via `/reportback/:rbid/edit` and change "caption", "total #", and "why is this important fields." Press "Updated Submission" button and make sure all changes have been properly made.  

Fixes #5807

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -99,7 +99,6 @@ function dosomething_reportback_form($form, &$form_state, $entity = NULL) {
       $reportback_submissions_data[$index] = reportback_file_load($fid);
       $reportback_submissions_data[$index]->image = dosomething_image_get_themed_image_by_fid($fid, '400x400');
     }
-
     $reportback_prior_submissions = array_reverse($reportback_submissions_data);
     $reportback_prior_submissions = array_slice($reportback_prior_submissions, 0, $reportback__submissions_count);
     $reportback_latest_submission = array_shift($reportback_prior_submissions);
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -353,7 +353,7 @@ function dosomething_reportback_form_submit($form, &$form_state) {
 
   // Save uploaded file.
   if ($file = dosomething_reportback_form_save_file($form_state)) {
-    //  Store new file's fid into values.
+    // Store new file's fid into values.
     $values['fid'] = $file->fid;
   }
 
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -322,6 +322,12 @@ function dosomething_reportback_form_validate($form, &$form_state) {
  */
 function dosomething_reportback_form_submit($form, &$form_state) {
   $values = $form_state['values'];
+  $campaign = node_load($values['nid']);
+  $language_code = $language_code = $campaign->language;
+  $values['run_nid'] = $campaign->field_current_run[$language_code][0]['target_id'];
+  $reportback = reportback_load($values['rbid']);
+  $photo_fids = $reportback->fids;
+  $most_recently_added_photo_fid = end($photo_fids);
 
   // If username field is set, this is an admin form.
   $is_admin_form = isset($form_state['values']['username']);
@@ -342,6 +348,8 @@ function dosomething_reportback_form_submit($form, &$form_state) {
   if ($file = dosomething_reportback_form_save_file($form_state)) {
     // Store new file's fid into values.
     $values['fid'] = $file->fid;
+  } else {
+    $values['fid'] =  $most_recently_added_photo_fid;
   }
 
   // Save values into reportback.
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -99,6 +99,7 @@ function dosomething_reportback_form($form, &$form_state, $entity = NULL) {
       $reportback_submissions_data[$index] = reportback_file_load($fid);
       $reportback_submissions_data[$index]->image = dosomething_image_get_themed_image_by_fid($fid, '400x400');
     }
+
     $reportback_prior_submissions = array_reverse($reportback_submissions_data);
     $reportback_prior_submissions = array_slice($reportback_prior_submissions, 0, $reportback__submissions_count);
     $reportback_latest_submission = array_shift($reportback_prior_submissions);
@@ -350,10 +351,10 @@ function dosomething_reportback_form_submit($form, &$form_state) {
     $values['uid'] = $user->uid;
   }
 
+  // Save uploaded file.
   if ($file = dosomething_reportback_form_save_file($form_state)) {
-    // Save uploaded file.
     $file = dosomething_reportback_form_save_file($form_state);
-      //  Store new file's fid into values.
+    //  Store new file's fid into values.
     $values['fid'] = $file->fid;
   }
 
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -690,14 +690,16 @@ function dosomething_reportback_save($values) {
...
     $entity = entity_load_single('reportback', $rbid);
+    $reportback_item = entity_load_single('reportback_item', $values['fid']);
   }
   try {
     // Set entity properties.
     dosomething_reportback_set_properties($entity, $values);
     // Set entity files.
-    dosomething_reportback_set_files($entity, $values);
-    // Save the entity.
+    dosomething_reportback_set_files($reportback_item, $values);
```
angaither: hmm, does this save reportbacks properly still? I think we will still want to set the files on the reportback entity, we just also want to save the reportback_item entity info

chloealee: @angaither when it was saved to the reportback entity, I wasn't able to access `$entity` after `$entity->save();` and it seemed to break the code, never reaching the next `if` statement and then redirecting incorrectly since it didn't have the `$entity->rbid` to return back to lines 359-366 in `dosomething_preortback_form_submit` ( `dosomething_reportback.forms.inc` file). 

I'm not sure what you mean by set the files on the reportback entity and then also save the reportback_item entity info?

angaither: this function inside the `try` is for creating new reportbacks. the `set_files` portion attached each file that the user uploaded to the reportback. 
I'm afraid if you call set files on the `reportback_item` it won't save the reportback_items with the correct rbid

chloealee: Ohh still might be a little confused - so we want to update the caption for the file in the `else` block right before the `try`?

angaither: well I _think_ (keep in mind i have not tested this at all) you'll want to leave that entity->save
and have a new one, that saves the files as well.

chloealee:
@angaither got it to work now with different scenarios! I'm saving the reportback file and caption to the `$reportback` if it's the first time the user is creating the reportback. If it is just to edit, as Nami suggests in the issue, I save the files to the `$reportback_item` that should already have an association to the reportback entity and already have a `rbid` to save to.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -690,12 +690,19 @@ function dosomething_reportback_save($values) {
...
     // Set entity properties.
     dosomething_reportback_set_properties($entity, $values);
-    // Set entity files.
-    dosomething_reportback_set_files($entity, $values);
+
+    if ($reportback_item) {
+      // Set entity files and save reportback item.
+      dosomething_reportback_set_files($reportback_item, $values);
+      $reportback_item->save();
+    } else {
```
angaither: `else` on a new line.

angaither:
@chloealee because of issues (namely #4529) I'm suppppper tentative to change the logic for this stuff. we will need to quadruple confirm things still work when this is merged in the DB and the counts match

cc @mikefantini @namimody

chloealee:
@angaither got it! Should I find some time to test with someone to make sure I'm understanding how everything is supposed to work together and to have more eyes on this?

angaither:
@chloealee that would be great!

chloealee:
sounds good! @namimody @mikefantini would either of you have time to go through this together and test?

namimody:
@chloealee I can! I know this would push merging/deploying this to next sprint, but could we do Friday morning sometime? Lots going on this sprint :see_no_evil: If yes, I'll put 1/2 hour on for us. :)

chloealee:
awesome, thanks @namimody ! Yes of course - Friday works for me! This issue was on the backlog that I picked up so I don't think it's that urgent :)

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -331,6 +334,15 @@ function dosomething_reportback_form_submit($form, &$form_state) {
   if ($is_admin_form) {
     $user = user_load_by_name($form_state['values']['username']);
     $values['uid'] = $user->uid;
+
```
weerd, line 338: PHP is throwing a fatal error here! Missing closing `}` on the `if`

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -711,18 +711,23 @@ function dosomething_reportback_save($values, $user = NULL) {
...
   }
 
   try {
     // Set entity properties.
     dosomething_reportback_set_properties($entity, $values);
 
-    // Set entity files.
-    dosomething_reportback_set_files($entity, $values);
-
+    if ($reportback_item) {
```
weerd: Since the assignment of `$reportback_item` happens in an `if/else` statement above it may mean that it doesn't get created and assigned, so when you check for it here there'll be an error if it doesn't exist. Best to wrap this check in `isset($reportback_item)`!

chloealee: ah makes sense, thanks Diego!

weerd:
Aight! Pulled the branch down and tested both my previous updates and updates in this PR and all seems like it's working well :)

:+1:

chloealee:
Great to hear, thanks @weerd ! @angaither should I still go through and test with @namimody on Friday?

weerd:
WAIT!

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -711,18 +711,23 @@ function dosomething_reportback_save($values, $user = NULL) {
...
     // Set entity properties.
     dosomething_reportback_set_properties($entity, $values);
 
-    // Set entity files.
-    dosomething_reportback_set_files($entity, $values);
-
+    if (isset($reportback_item)) {
+      // Set entity files and save reportback item.
+      dosomething_reportback_set_files($reportback_item, $values);
+      $reportback_item->save();
```
weerd, line 728: This isn't a method that exists! Apparently... not sure what's up, but it's yelling when it hits this.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -332,6 +335,16 @@ function dosomething_reportback_form_submit($form, &$form_state) {
...
   }
+
+  // Load uid by selected username and give access to file.
+  if ($is_admin_form) {
+    $account = user_load_by_name($form_state['values']['username']);
+    $values['uid'] = $account->uid;
+
+    $reportback = reportback_load($values['rbid']);
+    $most_recently_added_photo_fid = end($reportback->fids);
+    $values['fid'] =  $most_recently_added_photo_fid;
```
chloealee: @weerd would it be too hacky if I added `$values['user'] = 'admin';` so that in `dosomething_reportback.module` on line 721 I could check `if ($values['user'] == 'admin')`?

chloealee: @weerd actually that doesn't even work!

weerd: hmmm, plus as an admin I can still update Reportbacks on my own account when testing, so I feel the error would still happen.

weerd:
Tested and checks out in the various scenarios!

image

:+1:

chloealee:
yay, thanks Diego! So I guess now back to above question @angaither should @namimody and I still test on Friday?

weerd:
FYI, I tested the branch with code rebased from `dev` and things checked out ;)

PR #6081, opened 2016-01-25
Update to have Reportback class fully load a Campaign class instance.

weerd:
Refs #5991
#### What's this PR do?

This PR updates the Reportback class so it uses the `Campaign::get()` static method to completely load the necessary Campaign data along with the new `campaign_runs`, `language`, and `translations` data.

Just 1 file! **Reportback.php**

I was running into some errors since the old method of not fully loading the Campaign was missing vital data that, with the latest updates, the transformation process was now expecting. This should fix that issue moving forward.

---

@angaither


angaither:
:+1:

PR #6082, opened 2016-01-25
Allow admins to un-publish nodes without runs.

### LINKED ISSUE #6073: Allow unpublish without a run
[some](https://www.dosomething.org/us/campaigns/50-cans) [campaigns](https://thor.dosomething.org/us/node/3478) are already published without a run and they can't be unpublished
we should only disallow that box from being checked, not unchecked
Refs #5994

sbsmith86:
#### What's this PR do?

Before disabling the publish check box on nodes, checks to see if the box not checked. If the box is checked, then we allow the user to unpublish the node. Otherwise, if the box is not checked, and a run is not selected, we disable the box so they can't publish without a run. 

We used JS to disable the publish check box on campaign nodes that didn't have a run selected to ensure the relationship between campaigns and campaign runs. But we needed to allow already published campaigns that didn't have a run tied to them to be unpublished.

Fixes #6073

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/js/campaign_node_form.js
+++ b/lib/modules/dosomething/dosomething_campaign/js/campaign_node_form.js
@@ -56,7 +56,7 @@
 
     function togglePublish() {
       $currentRun = $( "select[name*='field_current_run']" ).val();
-      if ($currentRun === '_none') {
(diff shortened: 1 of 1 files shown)


angaither:
:+1:

Issue #6007, opened 2016-01-07
Add run filtering to admin views


angaither:
signup and reportback (reviewed only) needs to be able to filter by run

the default should be to display the current run
the run_nid should also be exposed in those tables

itsjoekent:
The problem I'm running into on both views is that I can't grab the current run, these are my options

![screen shot 2016-01-21 at 2 34 25 pm](https://cloud.githubusercontent.com/assets/897368/12492032/3485cdd4-c04c-11e5-8024-1c218eb1252f.png)

A textbox that has an option to input the NID, or a dropdown that checks if we have a run for that language.

Is there a way to "force" the current NID into textbox?

angaither:
@deadlybutter you might have to views_alter is
check out what @sbsmith86 did [here](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module#L36) with the current run field on the campaign node

itsjoekent:
Not having much luck with any of the view alters here https://api.drupal.org/api/views/views.api.php/group/views_hooks/7

I'm able to change the exposed inputs value, but it never reflects in the view itself (regardless of which hook I use in the execution order). :js: solution maybe?

PR #6086, opened 2016-01-25
Add run id as a join to signup table view.

angaither:
Expose that field to allow different things to be input.

cc @deadlybutter 
view setup 

what it looks like
screen shot 2016-01-25 at 4 05 27 pm

since stage doesn't have language/multiple runs I don't know what this will do if there's more than one run for a node. 

Refs #6007


itsjoekent:
:ship: :it: 

`:ship: :it:`

_get it.._

PR #6087, opened 2016-01-25
Fixes reportback cron

### LINKED ISSUE #6004: update goals counters
custom settings page on nodes, have a goals section where it runs totals for impact and reportbacks, these need to we re-written to only include the impact of the current run
it also needs to be updated to deal with multiple translations of the node. see #6065

itsjoekent:
#### What's this PR do?

Fixes the reportback cron so

1) The quantity is no longer r \* t (r= reportback total, t= # of translations)
2) The quantity is specific to the run

Use this query on a DB with :hankey: data

```
SELECT n.nid, SUM(rb.quantity) as total, run.field_current_run_target_id as run_nid
FROM node n
INNER JOIN field_data_field_campaign_status s on n.nid = s.entity_id AND n.language = s.language
INNER JOIN dosomething_reportback rb on n.nid = rb.nid
INNER JOIN field_data_field_current_run run on n.nid = run.entity_id
WHERE n.type = 'campaign'
AND s.field_campaign_status_value = 'active'
AND rb.flagged = 0
AND rb.run_nid = run.field_current_run_target_id
GROUP BY n.nid
```

Some of the values should be inflated based on the # of translations for that node. 

Run the cron in this PR and re-run that ^ query. Are the values back to normal? Check the dosomething_reportback_progress_log table...

See the comments in the issue.

Fixes #6004 
#### Screenshots

The reportback log progress table ordered by timestamp, after running cron. Notice the top one is correct "118".
screen shot 2016-01-25 at 4 12 30 pm

This was the query beforehand saying it should be 118.
screen shot 2016-01-25 at 4 12 38 pm

Proof it should be 118. (100 + 17 + 1, one of them is flagged)
screen shot 2016-01-25 at 4 16 00 pm

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.cron.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.cron.inc
@@ -34,4 +34,18 @@ function dosomething_reportback_cron() {
         ))
       ->execute();
   }
+
+  $results = db_query("SELECT n.nid, SUM(rb.quantity) as total
+                       FROM node n
+                       INNER JOIN field_data_field_campaign_status s on n.nid = s.entity_id AND n.language = s.language
+                       INNER JOIN dosomething_reportback rb on n.nid = rb.nid
+                       WHERE n.type = 'campaign'
+                       AND s.field_campaign_status_value = 'active'
+                       AND rb.flagged = 0
+                       GROUP BY n.nid;");
+
+  // Updated reportback counts accordingly.
+  foreach ($results as $result) {
+    dosomething_helpers_set_variable('node', $result->nid, 'sum_rb_all_quantity', (int) $result->total);
+  }  
 }
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.cron.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.cron.inc
@@ -9,7 +9,7 @@
  * Implements hook_cron().
  */
 function dosomething_reportback_cron() {
-  // Get all reportbacks for the current run updated in the past hour.
+  // Query for the total quantity for all active campaigns, for the current run
   $results = db_query("SELECT n.nid, SUM(rb.quantity) as total, run.field_current_run_target_id as run_nid
                        FROM node n
                        INNER JOIN field_data_field_campaign_status s on n.nid = s.entity_id AND n.language = s.language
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.cron.inc
```diff
@@ -10,13 +10,15 @@
  */
 function dosomething_reportback_cron() {
   // Get all reportbacks updated in the past hour.
-  $results = db_query("SELECT n.nid, SUM(rb.quantity) as total
+  $results = db_query("SELECT n.nid, SUM(rb.quantity) as total, run.field_current_run_target_id as run_nid
                        FROM node n
-                       INNER JOIN field_data_field_campaign_status s on n.nid = s.entity_id
+                       INNER JOIN field_data_field_campaign_status s on n.nid = s.entity_id AND n.language = s.language
                        INNER JOIN dosomething_reportback rb on n.nid = rb.nid
+                       INNER JOIN field_data_field_current_run run on n.nid = run.entity_id
```
angaither, line 17: since this field is translatable, you may have to join on both id and language here as well

itsjoekent, line 17: worked fine without it, see the screenshots above

angaither, line 17: does that node only have one translation though?

itsjoekent, line 17: has all 4

angaither:
:+1:

angaither:
before you merge can you create the new variable based on the total over all runs?

itsjoekent:
yus one sec!

itsjoekent:
hmmm, there doesnt seem to be much precedent for storing variables globally (everything right now is per node), and the function itself requires an entity ID, which seems contradicting considering this variable isn't specific to one entity.

angaither:
@deadlybutter this would still be a per node value, just an aggregate over multiple runs

itsjoekent:
OH sorry misunderstood, :+1:

itsjoekent:
@angaither ready for re-review!

angaither:
@deadlybutter can you add/update the documentation above both queries

itsjoekent:
@angaither done

angaither:
:+1:

PR #6089, opened 2016-01-26
returns all missing information

### LINKED ISSUE #6077: Missing fields from /api/v1/campaigns/:nid response
**BLOCKED**
Currently with both the `/api/v1/content/:nid` to `/api/v1/campaigns/:nid` endpoints are not providing the fields listed below. It's currently not possible to generate digest messages.

**Reason**:
`/api/v1/content/:nid` is set to be depreciated based on the documentation (https://github.com/DoSomething/phoenix/wiki/API#retrieve-a-campaign-will-be-deprecated) and currently has bugs in it's response. 

**Goal**:
Replacing the call in generation of the weekly digest email message by the Message Broker system from `/api/v1/content/:nid` to `/api/v1/campaigns/:nid`.

**Needs**:
The digest requires the following missing fields returned as a part of the `campaigns/:nid` response:
- DO IT: During Tip Header - `step_pre[0]->header`
- DO IT: During Tip Body - `step_pre[0]->copy`

image
- `latest_news`

image

cc @sergii-tkachenko 

**Related**:
- Message Broker Trello card: https://trello.com/c/QVBFbjTl

chloealee:
#### What does this PR do?

Returns missing information needed for weekly digest email message by Message Broker for `/api/v1/campaigns/:nid` endpoint. 

`Campaign.php` then to `Transformer.php`

Make sure `/api/v1/campaigns/:nid` endpoint has returns information for `pre_step_header`, `promoting_tips`, and `latest_news`. Use http://dev.dosomething.org:8888/api/v1/campaigns/362 to check Comeback Clothes. 
#### Notes

Not sure if the labels `pre_step_header` and `promoting_tips` are the most intuitive. Would it make more sense to change these to `during_tips_header` and `during_tips_copy` or even separate into 

```
-during_tips: {
  - header: "Run Your Drive!", 
  - copy: {
      raw: "Make flyers with the nearest drop-off box location. Put them on trash cans so people recycle     their clothes instead of throwing them away.",
formatted: "<p>Make flyers with the nearest drop-off box location. Put them on trash cans so people recycle their clothes instead of throwing them away.</p> "
}
```

Also, sometimes `latest_news` will return `null`. Is this okay for when using to populate content the email digest?
cc: @DeeZone 

Fixes #6077

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -229,8 +229,7 @@ abstract class Transformer {
 
         $output['solutions'] = $data->solutions;
 
-        $output['pre_step_header'] = $data->pre_step_header;
-        $output['promoting_tips'] = $data->promoting_tips;
+        $output['pre_step'] = $data->pre_step;
 
         $output['latest_news'] = $data->latest_news;
 
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -201,8 +201,6 @@ abstract class Transformer {
 
       // Show all properties for "full" display.
       if ($data->display === 'full') {
-        // print_r($data);
-        // die();
         $output['tagline'] = $data->tagline;
 
         $output['created_at'] = $data->created_at;
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -201,6 +201,8 @@ abstract class Transformer {
 
       // Show all properties for "full" display.
       if ($data->display === 'full') {
+        // print_r($data);
+        // die();
         $output['tagline'] = $data->tagline;
 
         $output['created_at'] = $data->created_at;
@@ -228,6 +230,11 @@ abstract class Transformer {
         $output['facts']['sources'] = $data->facts['sources'] ? $data->facts['sources'] : NULL;
 
         $output['solutions'] = $data->solutions;
+
+        $output['pre_step_header'] = $data->pre_step_header;
+        $output['promoting_tips'] = $data->promoting_tips;
+
+        $output['latest_news'] = $data->latest_news;
 
         $output['causes'] = $data->causes;
 
--- a/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
+++ b/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
@@ -22,7 +22,7 @@ class Campaign {
   public $staff_pick;
   public $facts;
   public $solutions;
-  public $pre_step_header;
+  public $pre_step;
   public $promoting_tips;
   public $latest_news;
   public $causes;
@@ -173,8 +173,7 @@ class Campaign {
         $solution_data = $this->getSolutionData();
         $this->solutions = $solution_data;
 
-        $this->pre_step_header = $this->getPreStepHeader();
-        $this->promoting_tips = $this->getPromotingTips();
+        $this->pre_step = $this->getPreStepInfo();
         $this->latest_news = $this->getLatestNews();
 
         $this->causes = $this->getCauses();
@@ -470,25 +469,16 @@ class Campaign {
   }
 
   /**
-   * Get campaign's pre-step header.
+   * Get campaign's pre-step header and caption.
    *
    * @return array
    */
-  protected function getPreStepHeader() {
-    return [
-      'pre_step_header' => dosomething_helpers_extract_field_data($this->node->field_pre_step_header),
-    ];
-  }
+  protected function getPreStepInfo() {
+    $pre_step = [];
+    $pre_step['header'] = (dosomething_helpers_extract_field_data($this->node->field_pre_step_header));
+    $pre_step['copy'] = (dosomething_helpers_extract_field_data($this->node->field_pre_step_copy));
 
-  /**
-   * Get campaign's promoting tips.
-   *
-   * @return array
-   */
-  protected function getPromotingTips() {
-    return [
-      'promoting_tips' => dosomething_helpers_extract_field_data($this->node->field_promoting_tips),
-    ];
+    return $pre_step;
   }
 
   /**
--- a/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
+++ b/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
@@ -22,6 +22,9 @@ class Campaign {
   public $staff_pick;
   public $facts;
   public $solutions;
+  public $pre_step_header;
+  public $promoting_tips;
+  public $latest_news;
   public $causes;
   public $action_types;
   public $issue;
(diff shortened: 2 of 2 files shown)


deezone:
@chloealee 

For reference, the digest logic that processes the results from this endpoint: https://github.com/DoSomething/mbc-digest-email/blob/master/src/MBC_DigestEmail_Campaign.php#L134-L205 Fields that are required vs optional are listed.

**In response to your questions**:
- `latest_news` is an optional field, it's expected to be NULL sometimes. The digest app deals with empty values for this field.
- currently, `during_tip_header` and `during_tip_copy` are working as replacements for the once working / now empty `step_pre[0]->header` and `step_pre[0]->copy`. Will these values remain as duplicates or will you remove them in this PR?
- From my perspective, feel free to change the return field names but ideally a change like this would result in a new version number: `/api/v2/campaigns/:nid`The goal is the endpoint returns consistent / non-breaking results as other systems are depending on it that are often beyond the control of the source application (DS DrupalAPI).

The Digest runs every Thursday. Currently it's working in production with the recently appeared  `during_tip_header` and `during_tip_copy` fields. The timing of this PR will need to happen a day before or after Thursday so tests can be run to ensure the Digest has everything it needs.

chloealee:
Thanks for the detailed response and reference to the digest logic, @DeeZone ! 

I see what you are saying now! Perhaps it is easiest if I leave them as `step_pre[0]->header` and `step_pre[0]->copy`? 

I'm also a bit confused by what you're referring to as the current `during_tip_header` and `during_tip_copy` - is this something that already exists? I'm not seeing it in the code.

lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -201,6 +201,8 @@ protected function transformCampaign($data) {
 
       // Show all properties for "full" display.
       if ($data->display === 'full') {
+        // print_r($data);
+        // die();
```
weerd: debug comment info! remove!

deezone:
> I'm also a bit confused by what you're referring to as the current during_tip_header and 
> during_tip_copy - is this something that already exists? I'm not seeing it in the code.

My bad, it's `pre_step_header` and `pre_step_copy` rather than`during_tip_header` and `during_tip_copy`.

```
curl https://www.dosomething.org/api/v1/content/362 -H "Accept: application/json"

...
    [pre_step_header] => Run Your Drive!
    [pre_step_copy] => <p>If you’re lucky, you'll collect a boatload of clothes and feel like a superhero. Because you're LITERALLY saving the world!</p>
   [latest_news_copy] => 
...
```

I've tested the `latest_news_copy` field on production and it's returning the expected value in the cURL request.

The `pre_step_header` and `pre_step_copy` fields seem to have replaced `step_pre[0]` values in the endpoint response.

So at the end of the day the endpoint is responding with "workable" values. It still remains a point of concern that existing fields in the endpoint response have changed. If this happens again another digest run of 600,000+ email messages will have missing fields again.

I'm good with this PR and the related issue being closed as resolved.

lib/modules/dosomething/dosomething_api/includes/Transformer.php
chloealee: ah - thank you for catching that!

weerd:
@chloealee @DeeZone My vote is to rename the property like you mentioned above. The aim with the `Campaign` class is to return data in a more logical and organized way, rather than the mess of an array with everything at the same level that Drupal normally returns.

@DeeZone with this update you'll already be switching to the new endpoint and need to make updates regardless. V1 of the `/campaigns` endpoint is to organize things logically :dancer:

chloealee:
thanks @DeeZone and @weerd for explaining! 

To double check, we should rename the properties to `pre_step_header`, `pre_step_copy`, and `latest_news_copy`?

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
```diff
@@ -22,6 +22,9 @@ class Campaign {
   public $staff_pick;
   public $facts;
   public $solutions;
+  public $pre_step_header;
```
weerd: The `pre_step` could probs be it's own item, and have a nested properties like you indicated in the description:

```
"pre_step": {
  "header": "Some header",
  "copy": {
    "raw": "lorem ipsum",
    "formatted": "<p>lorem ipsum</p>"
  }
}
```

something like that?

weerd: Regardless of whether you need to output the "copy" content now or later, at least it starts to set up the structure properly.

chloealee: @weerd gotcha! Will update now!

weerd:
@chloealee what's the difference between the "promoting tips" and the "pre_step copy"?

chloealee:
@weerd nada - they are the same!

deezone:
@weerd I like your thinking in terms of organization. But to be clear and I think you'll agree, once an API endpoint is published its response values should never disappear or be renamed.

weerd:
@DeeZone I do agree! Which is why we should add these items correctly and follow the same structure we've been using for the new endpoint. The **old** `/content` endpoint had the items differently than the new `/campaigns` one :)

These additions are items to the new endpoint that never existed before, so there's nothing that has been published with using this new endpoint yet!

deezone:
@weerd Ahhh, I was confused. I thought this PR was to adjust / fix the `/content` response? The related Issue (https://github.com/DoSomething/phoenix/issues/6077) it's trying to address is definitely about a breaking change in the `/content` response.

I'm fully in support of keeping the response of `/campaigns` well structured and not polluted with the legacy of `/content`.

chloealee:
@weerd thank you for catching me on that! I was grabbing the wrong field. Just updated it with `field_pre_step_copy` instead of `field_promoting_tips`

weerd:
@DeeZone Yep yep! So I thought you wanted us to expose those two items in new `/campaigns` endpoint so you could do an ole switcheroo and use the new `/campaigns` _instead_ of the old `/content`? Was that not the game plan?

deezone:
@weerd :bulb:  OK, now I'm understanding... yes, structure `/campaigns` as you see fit with the three new fields to support the Message Broker needs. Name as you see fit. The "ole switcheroo" is the dream.

Sorry for the confusion on my end.

weerd:
No worries! Sounds good @DeeZone! Just glad we've cleared it up :dancers:

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
```diff
@@ -408,6 +413,17 @@ protected function getLanguage() {
   }
 
   /**
+   * Get campaign's latest news.
+   *
+   * @return array
+   */
+  protected function getLatestNews() {
+    return [
+      'latest_news' => dosomething_helpers_extract_field_data($this->node->field_latest_news_copy),
```
weerd, line 422: Is there a way to test what the would return if it has sample data in it? Cause if it returns an array of stuff, those may need to be further processed... I could envision the `latest_news` property being an array of latest news objects with their own data...

angaither, line 422: @weerd it's just a text area field, nothing special inside it

weerd, line 422: Ah ok! I wasn't sure :)

weerd:
Looks like awesome-sauce! :+1:

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
weerd, line 422: It returns as an empty array within the `node` so was suspecting maybe it could be more items?

chloealee:
@weerd ready for final review per previous edits whenever you have a second!

PR #6091, opened 2016-01-26
add link to FAQs in contact us modal

### LINKED ISSUE #6062: Add campaign FAQs to Contact Us modal on action page
#### User Story

As a user, when I go to submit a question through the Contact Us modal, I want to see the FAQs before submitting my question to see if it can be answered without waiting for a response from a campaign lead.

In my mind the approach is to:
1. Edit the copy here https://www.dosomething.org/admin/config/dosomething/dosomething_zendesk so that we acknowledge the FAQs
2. Add the FAQs between ^ intro text and the submission form.

However this might look wonky so I'm open to other ideas and would be happy to talk through options.

Also this shouldn't matter, but FAQs are updated throughout the campaign so these would need to update accordingly as well.
#### Why This Matters

It would allow the user to get her question answered more quickly than having to wait for a campaign lead to respond.

Also, currently about 10% of campaign leads' time is spent answering Zendesk questions. About 80% of these questions are covered in the FAQs.

katiecrane:
Fixes #6062

Zendesk Contact Us modal at the bottom of campaign pages now includes a link to the FAQs for that campaign if they exist. 

If there are FAQs:
image

If there aren't FAQs:
image

The link opens the FAQ modal which is unchanged.

cc: @mikefantini

### DIFF
--- a/lib/modules/dosomething/dosomething_zendesk/dosomething_zendesk.module
+++ b/lib/modules/dosomething/dosomething_zendesk/dosomething_zendesk.module
@@ -147,9 +147,10 @@ function dosomething_zendesk_form($form, &$form_state, $entity = NULL)  {
     '#required' => TRUE,
     '#title' => t(variable_get('dosomething_zendesk_form_body_label')),
   );
-  //check if FAQs exist
-  //if so, array_push '#description' => 'Check out our <a href="#" data-modal-href="#modal-faq">FAQs</a> first', onto $form['body']
-
+  // Add link to FAQs if they exist
+  if(isset(dosomething_campaign_load($entity)->faq)) {
+    $form['body']['#description'] = 'Check out our <a href="#" data-modal-href="#modal-faq">FAQs</a> first';
+  }
 
   $form['submit'] = array(
     '#type' => 'submit',
--- a/lib/modules/dosomething/dosomething_zendesk/dosomething_zendesk.module
+++ b/lib/modules/dosomething/dosomething_zendesk/dosomething_zendesk.module
@@ -147,6 +147,10 @@ function dosomething_zendesk_form($form, &$form_state, $entity = NULL)  {
     '#required' => TRUE,
     '#title' => t(variable_get('dosomething_zendesk_form_body_label')),
   );
+  //check if FAQs exist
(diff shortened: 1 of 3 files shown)


mikefantini:
@katiecrane is it possible to be a bit more heavy handed with the FAQs suggestion?

Can we:
- have the text read "Or get your question answered right away by first checking our FAQs which are updated regularly." (linking "checking our FAQs")
- have the text be in our primary font color vs. secondary font color
- put that copy between the "Enter your question below" copy and the form field

katiecrane:
@mikefantini We can definitely do option 1 and I think option 3, not sure about option 2. I'll show you what 1 and 3 look like when I'm in on Friday

katiecrane:
@mikefantini:
Option 1:
image
This checks to see if the campaign has FAQs and only displays that text if it does.

Option 2:
image
caveat: I put this in by changing the form copy here https://www.dosomething.org/admin/config/dosomething/dosomething_zendesk
I haven't been able to figure out a way to first check if campaigns have FAQs and only display that part if they do. Do all campaigns have FAQs?

mikefantini:
@katiecrane visually Option 2 looks great. Exactly what I was looking for. But yeah would only want this if FAQs exist on the campaign, and I don't believe all campaigns have FAQs.

I was talking about it with @DFurnes and he may have some thoughts on how this could be implemented... Stay tuned!

katiecrane:
@mikefantini Yeah I agree that that one looks better. I haven't been able to figure out how to edit that copy from inside the code. Not sure if it's not a thing or just I can't find it. Staying tuned!

DFurnes:
Yes!

In `preprocess-node.inc`, we send along `$info_bar_vars['zendesk_form_header']` to the template to render that top paragraph (which is set from that admin panel variable right now).

Based on whether the node in `$vars` has anything set in the FAQs field, you could either add a new option to that array of theme variables (like a `$info_bar_vars['link_to_faq_modal']` boolean), or conditionally append another paragraph to that header variable.

DFurnes:
Feel free to give me a shout if that doesn't seem to work & we can try to figure it out together.

DFurnes:
:100: :shipit: from me!

katiecrane:
@mikefantini all your dreams have come true!

mikefantini:
Amazing! Thanks @katiecrane

PR #6093, opened 2016-01-27
Refactor campaign status.

### LINKED ISSUE #6092: High season end date appears in header photo on campaigns in the closed state
#### Current Behavior

High season end date is showing on header photo when a campaign is in the closed state if "Show End Date on Pitch & Action Page" is enabled.
#### Desired Behavior

Since the campaign is closed, this end date should no longer appear.
#### Steps to Replicate
- View a campaign that is closed
- Make sure "Show End Date on Pitch & Action Page" is enabled
- Verify that end date appears in the header photo below the call to action
#### Why This Matters

It's confusing for the user to tell them a campaign is going to end when it already has.

For an example, check out https://www.dosomething.org/us/campaigns/bake-sales-babies and see that Ends February 22nd appears on the header photo.

image

angaither:
Call a function we already use to determine the status of the campaign.

Fixes #6092

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -113,7 +113,7 @@ function dosomething_campaign_load($node, $public = FALSE, $cache = TRUE) {
     $campaign = new stdClass();
     $campaign->nid = $node->nid;
     $campaign->title = $node->title;
-    $campaign->status = dosomething_campaign_get_campaign_status($node);
+    $campaign->status = dosomething_campaign_is_closed($node) ? 'closed' : 'active';
     $campaign->type = dosomething_campaign_get_campaign_type($node);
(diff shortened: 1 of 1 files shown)


weerd:
:+1: 

_side note_: is this fine? or does it need to be somewhat updated as well? https://github.com/DoSomething/phoenix/blob/cc027e9c8ac0261c6ddd02dc7b2f089dd12ccac5/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php#L567

angaither:
@weerd I dunno, does it take lang into account?

weerd:
It does if you pass it a language, otherwise defaults to `en`. https://github.com/DoSomething/phoenix/blob/cc027e9c8ac0261c6ddd02dc7b2f089dd12ccac5/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module#L172

I built the extractor function with the intent of being able to account for languages down the line when we started using them. What we'd have to do is update the `Campaign` class to grab a user's language and then it can be passed. But not a high priority thing at the moment...

PR #6020, opened 2016-01-12
Adds run nid to reportbacks and signups

itsjoekent:
#### What's this PR do?

Adds the current run nid to the reportback and signup 
#### How should this be manually tested?

For reportbacks, run this query after the update hook runs. There shouldn't be any run nid's with 0 left.

```
SELECT dosomething_reportback.run_nid, field_data_field_current_run.field_current_run_target_id
                    FROM dosomething_reportback
                    INNER JOIN field_data_field_current_run
                    ON dosomething_reportback.nid = field_data_field_current_run.entity_id
```

Same thing but for signups.

```
SELECT dosomething_signup.run_nid, field_data_field_current_run.field_current_run_target_id
                    FROM dosomething_signup
                    INNER JOIN field_data_field_current_run
                    ON dosomething_signup.nid = field_data_field_current_run.entity_id
```
#### What are the relevant tickets?

Fixes #6014  
#### Screenshots

Reportbacks 
screen shot 2016-01-12 at 2 17 18 pm
screen shot 2016-01-12 at 2 21 59 pm

Signups
screen shot 2016-01-12 at 3 13 28 pm

PR #6094, opened 2016-01-27
Revert "Adds run nid to reportbacks and signups"

angaither:
Reverts DoSomething/phoenix#6020


angaither:
this doesn't work for _reasons_ 
`Integrity constraint violation: 1062 Duplicate entry[error] '1643687-2178-6293' for key 'uid-nid-run_nid'`

so we are getting duplicate rows when updating all signups to the current run nid

Issue #6095, opened 2016-01-27
Cleanup duplicate signups


itsjoekent:
- [ ] update `dosomething_signup_exists` to check for a signup without the run nid
- [ ] reconcile all duplicate signups
- [ ] reconcile all signups attached to the wrong run id

To find duplicate signups use this query

```
select s.nid, s.uid, s.run_nid, s.sid, from_unixtime(s.timestamp), r.field_current_run_target_id
from dosomething_signup s
inner join field_data_field_current_run r on r.entity_id = s.nid
where  1 < (select count(*) from dosomething_signup s2 where s2.uid = s.uid and s2.nid = s.nid);
```

This is blocking https://github.com/DoSomething/phoenix/issues/6014 as duplicates need to be removed before we can associate run nids on signups.

deadlybutter:
questions @angaither 

1. after this is done, are we re-merging https://github.com/DoSomething/phoenix/issues/6020 ?
2. I'm assuming we're making an update hook for this, how are generating the new ID's to make them unique? Is there a drupal function for this?

angaither:
closed with #6108

PR #6098, opened 2016-01-27
These runs sure are hot. 

angaither:
Update the dosomething_signup_exists function.
This will fix the problem we are having where it appears users are not signed up for a campaign
because the old signup might not have had the correct run attached to it.
This is not a pretty fix, but sometimes that is ok.
Refs #6095

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -320,7 +320,7 @@ function dosomething_signup_exists($nid, $run_nid = NULL, $uid = NULL, $presignu
   // If not see if they are signed up for the campaign without the run.
   // This section is added to test for duplicate signup problems.
   // Refs #6095.
-  if (!is_numeric($sid)) {
+  if (!is_numeric($sid) && !$presignup) {
      $result = db_select('dosomething_signup', 's')
       ->condition('uid', $uid)
       ->condition('nid', $nid)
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -308,22 +307,42 @@ function dosomething_signup_exists($nid, $runId = NULL, $uid = NULL, $presignup
...
+  // If not see if they are signed up for the campaign without the run.
+  // This section is added to test for duplicate signup problems.
+  // Refs #6095.
+  if (!is_numeric($sid)) {
+     $result = db_select('dosomething_signup', 's')
+      ->condition('uid', $uid)
+      ->condition('nid', $nid)
+      ->fields('s', array('sid'))
+      ->execute();
+      $running_free = TRUE;
```
mshmsh5000, line 329: OMG "running free"

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -308,22 +307,42 @@ function dosomething_signup_exists($nid, $runId = NULL, $uid = NULL, $presignup
...
       ->execute();
   }
+  $sid = $result->fetchField();
 
-  if (!isset($result)) {
-    return FALSE;
+  // If not see if they are signed up for the campaign without the run.
+  // This section is added to test for duplicate signup problems.
+  // Refs #6095.
+  if (!is_numeric($sid)) {
```
mshmsh5000: This will also trigger if we're running in "presignup" mode, and that query returned FALSE-y. There may be no side effects (not sure I have all the context), but you could eliminate entirely by checking for `&& !$presignup` here.

angaither:
@mshmsh5000 goooood catch, fixed

mshmsh5000:
Great! :+1: :shipit:

Issue #6096, opened 2016-01-27
Remove old code for deprecated /signups endpoint


weerd:
- [x] Turn off the endpoint in services
- [x] Remove the code for the resource.

Issue #6033, opened 2016-01-13
Remove all old code for the /campaigns/gallery.json endpoint


weerd:
With the updates in #5983, the old `/campaigns/gallery.json` relationship endpoint in Services is no longer needed, so we should remove all that old code.
- [x] Turn off the endpoint in services
- [x] Remove the code for the resource.

angaither:
bold.

> Services is no longer needed

weerd:
Go big or go home :dancer:

PR #6099, opened 2016-01-27
Updating Services to disable API endpoints no longer in use.

weerd:
Refs #6033
Refs #6096
#### What's this PR do?

This PR begins the first steps in the above 2 issues, to remove the deprecated endpoints from being accessible via Services. They are no longer in use, or otherwise were never used so best to kill them! :hocho: 

Take a look at the Feature and make sure it looks right.

Pull down and run a Feature revert. Check the Services endpoints in the CMS and see if the respective items no longer have little checkmarks next to them!

The `/campaigns/gallery.json` endpoint was only used in the Campaign Action page gallery, but that has since been changed to use the new `/reportback-items` endpoint.

The `/signups` endpoint was used for some testing and then scrapped for moving signup stuff for mobile app to the Northstar API.
#6033
#6096

---

@angaither 

CC: @DFurnes @aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_api/dosomething_api.info
+++ b/lib/modules/dosomething/dosomething_api/dosomething_api.info
@@ -10,4 +10,4 @@ features[ctools][] = services:services:3
 features[ctools][] = views:views_default:3.0
 features[features_api][] = api:2
 features[services_endpoint][] = drupalapi
-mtime = 1447955638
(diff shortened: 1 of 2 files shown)


DFurnes:
Seems good to me! Northstar only uses the `signup` action on the campaign resource.

weerd:
K, cool :)

PR #6100, opened 2016-01-27
Only fetch field inside the if.

angaither:
When reloading the page, the pitch page appeared again.
refs #6095


mshmsh5000:
Oh whoops, right -- I missed it too. :+1:

PR #6102, opened 2016-01-27
Kill old campaign gallery endpoint

### LINKED ISSUE #6033: Remove all old code for the /campaigns/gallery.json endpoint

weerd:
Fixes #6033 
#### What's this PR do?

This PR removes all the code related to the `/campaigns/gallery.json` endpoint since it is no longer in use and helps cleanup the resource file.

Take a look at the **campaign_resource.inc** file and make sure nothing extra was deleted!
#6033

---

@DFurnes 

CC: @angaither @aaronschachter


DFurnes:
:+1:

PR #6104, opened 2016-01-28
Removing all old code related to the unused /signups endpoint.

### LINKED ISSUE #6096: Remove old code for deprecated /signups endpoint

weerd:
Fixes #6096
#### What's this PR do?

This PR removes all the code related to the `/signups` endpoint since it is no longer in use and removes the resource file related to it.

Take a look at the campaign_resource.inc file and make sure nothing extra was deleted!
#6096

---

@DFurnes 

cc: @aaronschachter @angaither


DFurnes:
Fab! :rocket:

PR #6107, opened 2016-01-28
Fix the campaign run cron job

### LINKED ISSUE #6105: Fix campaign run cron
Currently, there is a campaign run cron job that gets all campaign runs and updates the campaign status. 

But this isn't right. 

Getting all the campaign runs and updating the status of the campaign their tied to doesn't work because not all campaign runs are the current run. So the status is being updated based on old runs sometimes. 

Not sure what I was thinking with that one. 

Anyway....I need to update this cron job to get all campaigns, then get the current run for the campaign, and then update the status based on that run.

sbsmith86:
#### What's this PR do?

Updates the cron job that sets campaign status based on campaign run dates so that we do it based on the campaign's current run. Before, we were setting the status for every run in the system. But not all runs are current. 

Fixes #6105 (background info in that ticket)


angaither:
:+1:

PR #6108, opened 2016-01-28
Add a script to remove zero and null duplicate signup rows.

angaither:
First pass at removing duplicates. 
This removes signups where the user has two records for the same node, and one of them is zero or null. it removes that row. 

Refs #6095

I tested this with a copy of the prod data and it removes the duplicate rows down to ~250

### DIFF
--- a/scripts/remove-null-zero-duplicate-signups.php
+++ b/scripts/remove-null-zero-duplicate-signups.php
@@ -21,6 +21,7 @@ foreach ($duplicates as $duplicate) {
       // Remove that row.
       db_delete('dosomething_signup')->condition('sid', $sid)->execute();
       echo 'removed row: ' . $sid . "\n";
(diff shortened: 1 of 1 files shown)


scripts/remove-null-zero-duplicate-signups.php
```diff
@@ -0,0 +1,26 @@
...
+                        ORDER BY s.nid;");
+
+foreach ($duplicates as $duplicate) {
+  $sids = explode(',', $duplicate->sids);
+  $run_nids = explode(',', $duplicate->run_nids);
+  foreach ($sids as $key => $sid) {
+    if ($run_nids[$key] == 0) {
+      // Remove that row.
+      db_delete('dosomething_signup')->condition('sid', $sid)->execute();
+      echo "removed row: " . $sid . "\n";
```
itsjoekent: do we still need this?

blisteringherb:
:+1:

PR #6109, opened 2016-01-28
Allowing custom expiration when setting cache.

### LINKED ISSUE #6031: Update ApiCache class to allow passing a custom expiration time
Allow passing a custom expiration time in hours to the `set()` method of the ApiCache class.

Once this is complete, update the reportback items tally caching to have a much longer caching period.

weerd:
Fixes #6031 
#### What's this PR do?

This updates the `set()` method in the `ApiCache` class so you can pass a custom expiration in hours, `false` to permanently cache, or let it default to a sensible value of temporarily caching until a general cache clear is performed on the app.

Most of the action is happening in the **ApiCache.php** file.

You can push down the branch, clear cache and load a few `/campaigns` endpoints. Check Redis for the data. Alternatively, I used a series of `watchdog()` calls to help let me know what was going on with the cache, but took them out after otherwise it was going to be too much noise in the DBLog reports. I may add them back later, along with a feature flag to show or hide the logging for test purposes.

Previously, Campaigns retrieved via the API were getting cached for 24 hours regardless. You could pass `?cache=false` to stop caching the response and also clear the cache for that query. Now that still works, but you can also clear the cache when Drush performs a cache clear or when the cache is manually cleared via the Admin interface.
#6031

---

@DFurnes @angaither 

CC: @aaronschachter @uy this should help in the future w/ the mobile app as campaign changes are made and you need to retrieve latest data and not needing to send an arbitrary request to the endpoint with `?cache=false` :tada:

### DIFF
--- a/lib/modules/dosomething/dosomething_api/dosomething_api.module
+++ b/lib/modules/dosomething/dosomething_api/dosomething_api.module
@@ -45,6 +45,15 @@ function dosomething_api_ctools_plugin_directory($module, $plugin) {
 }
 
 /**
+ * Implements hook_flush_caches().
+ *
+ * @return array
+ */
+function dosomething_api_flush_caches() {
+  return array('cache_dosomething_api');
+}
+
+/**
  * Implements hook_services_request_postprocess_alter().
  * @deprecated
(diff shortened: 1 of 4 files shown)


angaither:
:+1:

Issue #6110, opened 2016-01-29
Expose GET signup endpoint


angaither:
We're thinking that syncing all user activity data into Northstar is probably going to create more headaches for us in the future than it's worth. It seems like the best path forward is to expose this information from the Drupal API and let Northstar act as a authentication "proxy" for accessing those endpoints from the mobile app.

(In the future, the Northstar proxy code could be removed once we add some sort of [API management layer](https://github.com/DoSomething/northstar/issues/49) to let any service use the same Northstar authentication key/token.)

For Drupal to handle these endpoints, we need to add an endpoint for getting all of a given user's signups... e.g `signups/:user_id`. This will be used for the "hub" screen on the mobile app, and would return a (paginated?) collection of all signups the user has made.

```
GET signups/312412
```

``` json
{
  "data": [
    {
      "signup_id": 1234123,
      "campaign_id": 1337,
      "source": "<source>",
      "created_at": "<date>"
    },
    {
      "signup_id": 2145555,
      "campaign_id": 1338,
      "source": "<source>",
      "created_at": "<date>"
    }
  ]
}

```

It might also make sense to have an endpoint for querying a user's current signup status for a given campaign: perhaps `campaigns/:campaign_id/signups/:user_id`? Response should contain the signup the given user has made to that campaign, if they exist.

For example:

```
GET campaigns/1337/signups/312412
```

``` json
{
  "data": {
      "signup_id": 1234123,
      "campaign_id": 1337,
      "source": "<source>",
      "created_at": "<date>"
  }
}

```

angaither:
@weerd if you could work with @chloealee on this that would be :ok_woman:

weerd:
¡OLÉ! @chloealee lets have an endpoint creation party :tada:

chloealee:
@weerd yay can't wait!

aaronschachter:
This might be overkill, but here goes: if the main data source is Phoenix, any thoughts on embedding a  campaign object, like we do for the `reportbacks` and `reportback-items` endpoints?  If we also add a `reportback` property - a user's profile could be displayed in a single API request. (`user/:id/signups`)

Refs https://github.com/DoSomething/northstar/issues/210

weerd:
Yeah, I was gonna say I'd prefer a structure like this that follows along with what we're doing in our newer endpoints and Northstar:

``` json
{
  "data": [
    {
      "id": "1234123",
      "source": "<source>",
[...]

PR #6112, opened 2016-01-29
adds signup_resource.inc, SignupTransformer.php, and registers it in …

chloealee:
…dosomething_api.module
#### What does this PR do?

Creates the `signup.resource.inc` and `SignupTransformer.php` files. Registers this in `dosomething_api.module` 

`signup` should be it's own category highlighted in light blue in on the Services page: http://dev.dosomething.org:8888/admin/structure/services/list/drupalapi/resources. Dropdown should include `retrieve` and `index`.

First step for #6110. Endpoint still needs to be enabled and parameters need to be commented in.


weerd:
Remove the weird `._signup_resource.inc` hidden file that got committed.

lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
```diff
@@ -0,0 +1,113 @@
+<?php
+
+function _signup_resource_definition() {
+  $signup_resource = [];
+  $signup_resource['signup'] = [
```
weerd: make it plural! all endpoints are best to be kept plural as a general rule for REST APIs and standard to follow :)

chloealee:
ah, thanks @weerd ! For some reason that always happens to me! Deleted it :)

weerd:
Looks good! :+1: 

Didn't really look to much into the `SignupTransformer` stuff, cause it doesn't really matter just yet. Mostly just need the resource file and to update the module before we can activate the endpoint :)

PR #6113, opened 2016-02-01
runs features update to enable GET /signups endpoint

chloealee:
#### What does this PR do?

Runs features update to enable GET /signups endpoint.

Second step for #6110


weerd:
:+1:

PR #6114, opened 2016-02-01
removes ._signup.php and ._SignupTransformer.php

chloealee:
#### What does this PR do?

Removes ._Signup.php and ._SignupTransformer.php


weerd:
SSHFS :tableflip:

:+1:

PR #6115, opened 2016-02-01
Adds reportback install hook

### LINKED ISSUE #6014: Add run nid to current signups and reportbacks
after (or at the same time as) #5981 
all signups and reportbacks should have run id added to the col
currently they all default to 0
image

refs #5995 #5993

itsjoekent:
#### What's this PR do?

Adds reportback install hook which 

https://media.giphy.com/media/fDO2Nk0ImzvvW/giphy.gif

Use a query like this (replace the #'s with your own)

```
SELECT n.nid, rb.rbid
FROM node n
INNER JOIN node run ON run.nid = 1903 
INNER JOIN dosomething_reportback rb 
    ON rb.nid = n.nid 
    AND (rb.run_nid = 0 OR rb.run_nid IS NULL)
AND n.nid = 1485
```

To test for the ones which have a date range, add this line `AND from_unixtime(rb.created) BETWEEN '" . $run->start_date . "' AND '" . $run->end_date . "'` and get the run/start dates from this query:

```
SELECT DISTINCT c.entity_id as run_nid, d.field_run_date_value as start_date, d.field_run_date_value2 as end_date
                      FROM field_data_field_campaigns c
                      INNER JOIN field_data_field_run_date d ON d.entity_id = c.entity_id
                      WHERE c.bundle = 'campaign_run'
                      AND c.field_campaigns_target_id = 1485
                      ORDER BY d.field_run_date_value2 ASC
```

After running the update that query shouldnt return anything 

I mean, I think this update is pretty simple? (In all seriousness, see the million comments left in the code)

Fixes #6014 
#### Questions

I think we're gonna need to block off time again to quickly test this @angaither @mikefantini ? (also probably back the database up incase something fucks up?)


itsjoekent:
@angaither is there anything sena needs to do before we merge this? should i be targeting a new branch instead of dev?

angaither:
@deadlybutter yeah, it'll probably be better to have this be a feature branch we can deploy on it's own

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
```diff
@@ -887,3 +887,69 @@ function dosomething_reportback_update_7033() {
...
+                    FROM node n
+                    WHERE n.type = 'campaign'");
+
+  // Loop through every NID
+  foreach ($nids as $nid) {
+
+    // Get the runs for associated for this campaign & the timing
+    $runs = db_query("SELECT DISTINCT c.entity_id as run_nid, d.field_run_date_value as start_date, d.field_run_date_value2 as end_date
+                      FROM field_data_field_campaigns c
+                      INNER JOIN field_data_field_run_date d ON d.entity_id = c.entity_id
```
angaither: i think you would need to join on language here as well

itsjoekent: to prevent duplicates?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
```diff
@@ -887,3 +887,69 @@ function dosomething_reportback_update_7033() {
...
+                                AND (rb.run_nid = 0 OR rb.run_nid IS NULL)
+                             AND n.nid = " . $nid->nid);
+
+        // For reach reportback returned
+        foreach($results as $result) {
+          // Load the reportback and change its run nid to the corresponding
+          // one we've now discovered.
+          $rb = reportback_load($result->rbid);
+          $rb->run_nid = $run->run_nid;
+          entity_save('reportback', $rb);
```
angaither: you could abstract this into a helper function

angaither:
If you don't want this to be a feature branch, you could turn it into a script so we can deploy it whenever we are ready

also, it will make the deploy faster!

itsjoekent:
ah, script idea sounds good.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
angaither: no, to get the date that matches the language of the run

itsjoekent:
@angaither ok i changed a bunch of files, this should probably get re-reviewed

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.drush.inc
```diff
@@ -1,8 +1,92 @@
...
+  foreach ($nids as $nid) {
+
+    // Get the runs for associated for this campaign & the timing
+    $runs = db_query("SELECT DISTINCT c.entity_id as run_nid, d.field_run_date_value as start_date, d.field_run_date_value2 as end_date
+                      FROM field_data_field_campaigns c
+                      INNER JOIN field_data_field_run_date d ON d.entity_id = c.entity_id AND d.language = 'en'
+                      WHERE c.bundle = 'campaign_run'
+                      AND c.language = 'en'
+                      AND c.field_campaigns_target_id =" . $nid->nid ."
+                      ORDER BY d.field_run_date_value2 ASC");
```
angaither, line 44: the syntax looks strange here, might need to alternate ' "

itsjoekent, line 44: do you mean switch all of the double quotes to single? sorry not following

angaither, line 44: alternate between single and double quotes

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -217,6 +217,7 @@ function dosomething_signup_create($nid, $uid = NULL, $source = NULL, $timestamp
 
   $campaign = Campaign::get($nid);
   $run = dosomething_helpers_get_current_campaign_run_for_user($nid, $user, $campaign);
+  $run->nid = 0;
```
angaither: why is this added?

lib/modules/dosomething/dosomething_signup/includes/dosomething_signup.inc
```diff
@@ -95,8 +95,8 @@ class SignupEntityController extends EntityAPIController {
     // If the entity uid doesnt belong to current user:
     if ($entity->uid != $user->uid) {
       // And current user can't edit any reportback:
-      if (!user_access('edit any signup')) {
-        watchdog('dosomething_signup', "Attempted uid override for @entity by User @uid", 
+      if (!user_access('edit any signup') && !drupal_is_cli()) {
```
angaither, line 98: does this work? I would think it would fail out above when it checks the entity uid and the user? 
this should probably be it's own line, like [in the global module](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_global/dosomething_global.module#L24)

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
itsjoekent: leftover from testing (quick hack to make it not assign run nid upon signup), thanks for spotting that

lib/modules/dosomething/dosomething_signup/includes/dosomething_signup.inc
itsjoekent, line 98: yep this does work, @sergii-tkachenko did the same thing for reportback saves here https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_reportback/includes/ReportbackController.php#L106

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.drush.inc
itsjoekent, line 44: Yeah on line 40/41 its single quotes, but the query is surrounded by double. i'm missing something?

angaither:
:+1: 
@deadlybutter please merge, deploy and test on thor with the campaigns that have multiple runs but didn't archive data

PR #6118, opened 2016-02-02
Signups endpoint

### LINKED ISSUE #6110: Expose GET signup endpoint

chloealee:
#### What does this PR do?

Exposes `GET /signups` endpoint in API. 
1. `signup_resource.inc` - `_signup_resource_index` and `_signup_resource_retrieve` functions
2. `SignupTransformer.php` 
3. `Signup.php`
4. `dosomething_signup.query.inc`
5. SignupTransformer.php
6. Transformer.php

To `GET` an index of all signups, hit `/signups`
To `GET` signup by `sid`, hit `/signups/:sid`
To query by `user` and `campaigns`, hit `/signups?user=98&campaigns=1127,1173`

Fixes #6110

Note: we still need to add pagination to the index.


lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -406,6 +406,30 @@ protected function transformReportbackItem($data) {
...
+   *   An object containing properties of Signup data:
+   *   - id: (string) The Signup id.
+   *   - created_at: (string) Date Signup was created.
+   *   - campaign: (string) Campaign Signup belongs to.
+   *   - campaign_run: (string) Campaign run Signup belongs to.
+   * @return array
+   */
+  protected function transformSignup($data) {
+    return [
+      'signup_id' => $data->id,
```
weerd: Since this is the "signups" object, just call this `id` :)

chloealee: oh doi! :)

lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
```diff
@@ -34,26 +34,26 @@ function _signup_resource_definition() {
...
-        //     'type' => 'string',
-        //     'source' => ['param' => 'campaigns'],
-        //     'default value' => NULL,
-        //   ],
-        // ],
+        'args' => [
+          [
+            'name' => 'user',
+            'description' => 'The id of the specified user to get signups for.',
+            'optional' => FALSE,
```
weerd: this should be `TRUE`... as an index we should be able to call `/signups` and just get a paginated list of all signups. Granted it would be rare that we'd need to do that, but it shouldn't be mandatory to have to pass a user id.

lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -406,6 +406,30 @@ protected function transformReportbackItem($data) {
...
+      'source' => $data->source,
+      'created_at' => $data->created_at,
+      'campaign' => [
```
weerd: This transformer method should just be for the signup data, so we shouldn't include the campaign info here. That would be a separate call in the `SignupsTransformer` class similar to how the `ReportbackTransformer` includes the campaign data here: https://github.com/DoSomething/phoenix/blob/62106f609af0c415e97559b19b05e9394a600338/lib/modules/dosomething/dosomething_reportback/includes/ReportbackTransformer.php#L79

lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -406,6 +406,30 @@ protected function transformReportbackItem($data) {
...
+   *   - campaign: (string) Campaign Signup belongs to.
+   *   - campaign_run: (string) Campaign run Signup belongs to.
+   * @return array
+   */
+  protected function transformSignup($data) {
+    return [
+      'id' => $data->id,
+      'source' => $data->source,
+      'created_at' => $data->created_at,
+      'campaign' => [
```
weerd: This transformer method should just be for the signup data, so we shouldn't include the campaign info here. That would be a separate call in the `SignupsTransformer` class similar to how the `ReportbackTransformer` includes the campaign data here: https://github.com/DoSomething/phoenix/blob/62106f609af0c415e97559b19b05e9394a600338/lib/modules/dosomething/dosomething_reportback/includes/ReportbackTransformer.php#L79

lib/modules/dosomething/dosomething_api/includes/Transformer.php
chloealee: got it, makes sense! I wasn't sure if we needed all the information that gets returned in the campaign object so commented it out. I'll re-add now.  

Should we also include the `campaign_run` for the signup in `SignupTransformer.php`'s `transform` method? If so, I can just add that function to `Transformer.php` and call that under `$data['campaign`]...`in`SignupTransformer.php`'s`transform` method.

chloealee: To clarify the above - should we include which particular `campaign_run` the `signup` is associated with? The campaign object returns the `current campaign_runs` but not sure if this is necessarily the one associated with the `signup`.

weerd:
Delete this hidden file: **._dosomething_signup.query.inc**  --> SSHFS :(

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
```diff
@@ -26,6 +26,9 @@ function dosomething_reportback_build_reportbacks_query($params = array()) {
 
   $query->addExpression('GROUP_CONCAT(DISTINCT rbf.fid)', 'items');
 
+  print_r($query);
+  die();
```
weerd: Need to remove sneaky debug statements!

weerd:
@chloealee I think I've narrowed in on what might be causing the `flag_flag.inc` issues... there's something going on with the inclusion of the php files for the Signups class, transformer and controller...

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
chloealee: ah shoot thank you!

weerd:
@chloealee I think it's related to a dependency issue... try adding the following items to the **dosomething_signups.info** within the list of dependencies:

```
dependencies[] = dosomething_api
dependencies[] = dosomething_campaign
dependencies[] = dosomething_helpers
```

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -13,12 +13,17 @@ include_once 'dosomething_signup.theme.inc';
 include_once 'includes/dosomething_signup.mobilecommons.inc';
 include_once 'includes/dosomething_signup.variable.inc';
 include_once 'includes/dosomething_signup.cron.inc';
+include_once 'dosomething_signup.query.inc';
+
+include_once 'includes/Signup.php';
+include_once 'includes/SignupController.php';
```
weerd, line 19: Reference to **SignupController.php** which doesn't exist in the **/includes** directory is throwing errors!

lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
```diff
@@ -0,0 +1,88 @@
...
+  //   }
+  //   else {
+  //     $query->condition('s.nid', $params['campaigns']);
+  //   }
+  // }
+
+  // return $query;
+// }
+//
+//
```
weerd: Should delete the commented out code and add DocBlocks for the used functions below!

lib/modules/dosomething/dosomething_signup/includes/SignupTransformer.php
```diff
@@ -78,19 +83,13 @@ protected function transform($item) {
    * @param  array  $parameters
    * @return array
    */
-  private function setFilters($parameters) {
-    $filters = [
-      'sid' => dosomething_helpers_format_data($parameters['ids']),
-      // 'users' => dosomething_helpers_format_data($parameters['users']),
-      // 'nid' => dosomething_helpers_format_data($parameters['campaigns']),
-    ];
+  // private function setFilters($parameters) {
```
weerd: I'd recommend using the `setFilters()` function for the params... cause the list of params could grow over time and nice to split that out into its own function :smiley:

chloealee:
@weerd finally got all of the logs to disappear! I don't _think_ there was any use for the `services_resource_build_index_list` function so I got rid of it and solved our issues, still returning all the expected information for this endpoint. 

Also got rid of returning the `source` because it looks like we don't have access to that.

weerd:
@chloealee it is actually used (at least in other endpoints). It helps to generate the appropriate path URL for the endpoints for the uri property....

DFurnes:
there are so many green boxes in here!

itsjoekent:
i think we should probz squash these commits

itsjoekent:
dave just wins the internet. im done

weerd:
@chloealee the weird behavior with that services function is because it's expecting something slightly different... I'm checking to see what is returning unexpected data.

chloealee:
thanks, @weerd! Made those fixes with no errors in the log - crossing my fingers this is it!

lib/modules/dosomething/dosomething_signup/includes/SignupTransformer.php
```diff
@@ -32,12 +32,13 @@ public function index() {//$parameters) {
...
    * Display the specified resource.
    *
-   * @param string $id Resource id.
+   * @param string $id Signup id.
    * @return array
    */
   public function show($id) {
     try {
       $signup = Signup::get($id);
+      $signup = $signup[0];
```
weerd: I don't think you need this... you remove it from the array and then wrap back in array for the services function below and then pop out from array.

``` php
try {
  $signup = Signup::get($id);
  $signup = services_resource_build_index_list($signup, 'signups', 'id');
  $signup = array_pop($signup);
}
```

weerd: Oh wait, hold on a sec...

chloealee:
@weerd updated!

weerd:
Yay! Looks greeeeeat!

[!image](http://static.comicvine.com/uploads/original/11123/111231848/4602082-6248423626-tony9.jpg)

weerd:
Yay! Looks greeeeeeat!

image

chloealee:
thanks for all the help, @weerd !!

PR #6120, opened 2016-02-02
Update status on campaign save

### LINKED ISSUE #6052: Update campaign status on campaign save
When a campaign node is saved, we should update the campaign status. A user could change the campaign run on the campaign node, and we need to update the status to reflect the status based on the new campaign run and dates. 

I think we might be able to do this only when the campaign run changes.

sbsmith86:
### What's this PR do?

Updates the campaign status when someone changes the campaign run on campaign nodes. 

Also, updated the hooks I was using to update the campaign status on campaign run node save to use `hook_node_submit` so we can more reliably get the values that had just been submitted by the user. 

Start at `dosomething_campaign.module` where I use `hook_node_validate()` to get the form state of the campaign node that is submitted and allows me to update the value of fields before the form actually gets submitted. This is where we update the status. 

Fixes #6052


sbsmith86:
@angaither updated the campaign status function to check for null dates.

angaither:
:+1:

PR #6126, opened 2016-02-03
Clear cache for API when a campaign is updated.

### LINKED ISSUE #6125: Clear API cache when a campaign is updated.
To have the latest data available, we need to clear the campaign cache when a campaign is updated.

weerd:
Fixes #6125
#### What's this PR do?

This does a general clear of any campaign data in cache for the API. This also puts into use the CampaignController class that has been sitting idle. Might become more useful down the line.

I tested this a bunch of times to make sure that the API cache cleared when a campaign is updated. One thing to keep in mind, right now the API returns campaign and associated data for the made node, and includes info on what translations are available, but does not return a specific translations data to you. So you might change a field in the `us` version (with a campaign that also has a global version), but the API doesn't reflect it... while the cache _DID_ get cleared, the API is actually only showing you the field data for the main node (likely global). Just a caveat to keep in mind if it seems like the data didn't change!

Minor updates; just check the two files :)

I added the use of the `CampaignController` because it could start coming in handy. We have one for some of the other resources available through the API (mostly because they are entities and entities has it's own controller class you need to extend).

---

@angaither @DFurnes

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/includes/CampaignController.php
+++ b/lib/modules/dosomething/dosomething_campaign/includes/CampaignController.php
@@ -2,6 +2,15 @@
 
 class CampaignController {
 
-  // @TODO: handle main app controller-y things such as CRUD in here.
+  /**
+   * Update the specified Campaign in storage.
+   *
+   * @param  int  $id
+   * @return void
+   */
+  function update($id) {
+    cache_clear_all('campaigns_api_request', 'cache_dosomething_api', TRUE);
+    cache_clear_all('ds_campaign_' . $id, 'cache_dosomething_campaign', TRUE);
+  }
 
 }
--- a/lib/modules/dosomething/dosomething_campaign/includes/CampaignController.php
+++ b/lib/modules/dosomething/dosomething_campaign/includes/CampaignController.php
@@ -9,7 +9,7 @@ class CampaignController {
    * @return void
    */
   function update($id) {
-    cache_clear_all('campaigns_api_request', 'cache_dosomething_api', TRUE);
+    cache_clear_all('campaigns_index_api_request', 'cache_dosomething_api', TRUE);
     cache_clear_all('ds_campaign_' . $id, 'cache_dosomething_campaign', TRUE);
   }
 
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_campaign/includes/CampaignController.php
```diff
@@ -2,6 +2,15 @@
...
 
-  // @TODO: handle main app controller-y things such as CRUD in here.
+  /**
+   * Update the specified Campaign in storage.
+   *
+   * @param  int  $id
+   * @return void
+   */
+  function update($id) {
+    cache_clear_all('campaigns_api_request', 'cache_dosomething_api', TRUE);
```
angaither: this clears the entire api cache when any node is updated. I don't think we want that....

weerd: Well, only clears anything that begins with `campaigns_api_request*`, and rest is wildcard...

Yeah, I wasn't too keen on this, but need to keep in mind a specific campaign might be in a variety of different API related calls based on the query params...

`/campaigns?ids=362` <-- cached result
`/campaigns?ids=362,1321` <--- diff cached result
`/campaigns?staff_pick=true` <-- diff cached result that might include 362

angaither: tricky, but at the minimum this needs to be scoped only to campaign nodes https://github.com/DoSomething/phoenix/pull/6126/files#diff-ae398bb8864fc3c26317f7ccaef04c0aR1263

weerd:
@angaither I made a couple updates. I've changed the `$endpoint` parameter you can pass to the `ApiCache` to be `$action` (thanks for the naming help @DFurnes :trophy:) so when passed, you can specify not only the endpoint but the method involved, `campaign_index` or `campaign_show`. This way, when clearing cache, we can clear all `index()` collection ones, but not all `show()` ones, which are specific to that single campaign :)

weerd:
I have a further PR I'm working on to also cache the `show()` endpoints which currently are not being cached!

angaither:
kewl! :+1:

PR #6127, opened 2016-02-03
Update campaign status timing logic

### LINKED ISSUE #6058: Campaign Run end date should set campaign status to closed at midnight EST the following day
#### Current Behavior

If the end date on a campaign run is set for 1/31/16, then the campaign status of the associated campaign gets set to closed on 1/31/16. 
#### Desired Behavior

If a campaign run end date is set for 1/31/16, then on 2/1/16 at 00:00:00 EST the campaign status should be set to closed. So whether you are in New York, LA, Beijing, or anywhere else, regardless of the timezone on your user profile, the campaign should be closed at 2/1/16 at 00:00:00 EST.
#### Steps to Replicate
1. Create a campaign
2. Create a campaign run associated with that campaign and have the end date be the current date
3. Set the new run as the current run on the campaign
4. Verify that campaign status = closed
#### Why This Matters

We advertise to our users and sponsors that campaigns that close will do so at 11:59pm EST on the last day of the month. There would be lots of user confusion if the campaign closed earlier than advertised.

Campaign run with end date set for today: https://thor.dosomething.org/us/node/6464/edit

Associated campaign where status = closed:
https://thor.dosomething.org/us/node/6463/edit

sbsmith86:
#### What's this PR do?

Makes updates to the campaign timing logic: 
- Updated `dosomething_helpers_get_campaign_status()` to take in date strings instead of datetime objects. This way we can easily ensure the datetime object is created in EST which is the timezone all time comparisons should be based in.
- Add a timestamp to the end date that ensures we compare the current date and time to the end of the day on the end date. i.e. Runs set to end today, 2/3/15, don't really end until after 11:59pm.

 `dosomething_helpers_get_campaign_status()` is the main logic that was updated and then I just updated instances where we were passing datetime objects to that function to just pass the raw string.

Fixes #6058

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -1280,8 +1280,8 @@ function dosomething_campaign_node_validate($node, $form, &$form_state) {
     if ($current_run_id !== $submitted_current_run) {
       // Get the dates of the run.
       $new_run = entity_metadata_wrapper('node', $submitted_current_run);
-      $start_date = new DateTime($new_run->language($language)->field_run_date->value()['value']);
-      $end_date = new DateTime($new_run->language($language)->field_run_date->value()['value2']);
+      $start_date = $new_run->language($language)->field_run_date->value()['value'];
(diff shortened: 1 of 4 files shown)


angaither:
:+1:

PR #6129, opened 2016-02-03
Adds progress logging

itsjoekent:
#### What's this PR do?

Adds print_r in both the signup and reportback run nid update script so we know where its at

Run the script, magic! 

Kinda hard to see whats going on under the hood, so logging!

screen shot 2016-02-03 at 4 47 08 pm


angaither:
@deadlybutter instead of an index of the loop, it's probably more helpful to see which node/signup the loop is on

itsjoekent:
my mistake, how about this? 

screen shot 2016-02-03 at 4 56 24 pm

angaither:
:+1:

PR #6131, opened 2016-02-04
Enable caching of single campaign show method.

### LINKED ISSUE #6130: Cache single Campaign calls to the API
We are caching calls to index collection of `/campaigns`, but we currently are not caching calls to a single campaigns via the endpoint `/campaigns/:id`.

weerd:
Fixes #6130
#### What's this PR do?

This PR adds the capacity to cache data when requesting a single campaign via the API endpoint for. Also makes sure that the cache is cleared for that specific campaign called via API when the specified campaign is edited via the admin CMS.

Start at **CampaignTransformer.php**, then head to the changes in the **ApiCache.php** and lastly 
the **CampaignController.php** file.
#6130

---

@angaither 

cc: @DFurnes @aaronschachter @jonuy

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/ApiCache.php
+++ b/lib/modules/dosomething/dosomething_api/includes/ApiCache.php
@@ -37,18 +37,20 @@ class ApiCache {
 
     $cache = cache_get($id, 'cache_dosomething_api');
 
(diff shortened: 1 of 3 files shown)


angaither:
:+1:

Issue #6123, opened 2016-02-03
Run End date set to start date


angaither:
in the `field_data_field_run_date` on campaign runs, when you set a start date for a campaign, the end date gets set in the database to the start_date, even if you don't select "Show end date" 

dear drupal, this is not ok. 
https://www.drupal.org/node/874322

PR #6132, opened 2016-02-04
Adds second conditional to account for drupal bug

itsjoekent:
The end date is the same as the start date when not programmatically generated, thus this script was failing because it was looking for NULL values. It now accounts for cases where start == end

refs #6123


angaither:
:+1:

PR #6134, opened 2016-02-04
Updates rb script with same changes as the signup script

itsjoekent:
Explanation / changes are the same as this PR https://github.com/DoSomething/phoenix/pull/6132 but for the reportback script


angaither:
:+1:

PR #6135, opened 2016-02-04
Adds error handling and watchdog noticing to both scripts

itsjoekent:
We are experiencing some SQL errors which is resulting in large batches of reportbacks not being updated. This PR adds a try catch to make sure the script continues running and we watchdog any rbid/sid's that throw this error to allow for further investigation


angaither:
what's the opposite of DRY?

:+1:

itsjoekent:
Uh, PERFECTION!

weerd:
hahahahaha

PR #6056, opened 2016-01-15
Unarchive run activity

chloealee:
#### What does this PR do?

Un-archives run activity and moves all signups and reportbacks from run nodes back to the original campaign node. 
#### How should this be tested?

First run the 

`SELECT n.nid as campaign_node, n.language, n.title, n.created, c.field_campaigns_target_id, rb.rbid, rb.uid, rb.nid as reportback_node, rb.run_nid
FROM node n
INNER JOIN field_data_field_campaigns c on n.nid = c.entity_id
INNER JOIN dosomething_reportback rb on rb.nid = n.nid
WHERE n.type = 'campaign_run';` 

and 

`SELECT n.nid as campaign_node, n.language, n.title, n.created, c.field_campaigns_target_id, s.sid, s.uid, s.nid as signup_node, s.run_nid
FROM node n
INNER JOIN field_data_field_campaigns c on n.nid = c.entity_id
INNER JOIN dosomething_signup s on s.nid = n.nid
WHERE n.type = 'campaign_run';`

in sequel pro. Here, all `run_nid`s should be 0.

In your vagrant box, run `drush updb -y` in dosomething@dev:/var/www/dev.dosomething.org/html$. 

Reload the `dosomething_signup` and `dosomething_reportbacks` tables and `nid`s and `run_nid`s should be updated. All campaigns should be updated so their `nid`= a campaign id (e.g. Comeback Clothes 362) and `run_nid` = a campaign nid (e.g. Comeback Clothes 1227). 
#### What are the relevant tickets?

Fixes #5995

PR #6136, opened 2016-02-04
adds updates to signups and reportbacks to set run IDs to 0

chloealee:
#### What does this PR do?

Cleans up PR #6056 and sets `run_nid`s to `0`. 

See #6056 
#6056

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
@@ -902,7 +902,6 @@ function dosomething_reportback_update_7033() {
      db_update('dosomething_reportback')
        ->fields(array(
          'nid' => $reportback->field_campaigns_target_id,
-         'run_nid' => '0',
          )
        )
        ->condition('nid', $reportback->reportback_node)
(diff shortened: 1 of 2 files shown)


chloealee:
@angaither @mikefantini I'm a little confused again. When we run the queries in the functions, all the data that gets returned already have `run_nid` = `0`, which is what the goal of this script is to do, right?

angaither:
@chloealee the goal of this is to move signups and reportbacks that have been moved from `campaign_run` nodes to `campaign` nodes

chloealee:
@angaither got it. But we changed it to set the `run_nid` = `0` in order to run Joe's script first because @mikefantini said "Reason: some runs have activity on them that should actually actually be associated with a different run." Just confused about this part now everything that gets returned from the query already had a `run_nid` = `0`?

angaither:
@chloealee yeah, I guess you don't even have to touch that line
essentially you can 

```
select signups
join node
join field campaigns
where node type = campaign run

signup->nid = field_target_id
```

chloealee:
@angaither just updated it to remove the line of setting `run_nid` to `0`. Does this accomplish what @mikefantini asked for?

angaither:
:+1:

chloealee:
cool - will merge now!

PR #6138, opened 2016-02-04
don't show duplicate signups

### LINKED ISSUE #6116: Update the signup view on nodes to support languages
https://www.dosomething.org/us/node/6620/signups
image

signups are appearing to be duplicates, it's probably because the view needs to be updated to deal with joining on node id and language id (same issue as #6065)

katiecrane:
Fixes #6116 

We were seeing duplicate signups on pages like here (https://www.dosomething.org/us/node/6620/signups) stemming from multiple translations existing for the campaign.

On this page (/admin/structure/views/view/node_signups), going under "Other," clicking "Query settings" and then "distinct" stops duplicates from showing up.


angaither:
:+1:

PR #6140, opened 2016-02-04
Forces end date to be actual end of day

itsjoekent:
Signups/reportbacks made on the actual day of the end date are not included in the loop


angaither:
:+1:

PR #6142, opened 2016-02-04
Moves end date setting for RB into loop instead of outside

itsjoekent:
set the end date inside the check to see if it's set.


angaither:
:+1:

PR #6145, opened 2016-02-04
Set run date to NULL

### LINKED ISSUE #6123: Run End date set to start date

### LINKED ISSUE #6144: campaigns with current runs w/o end dates closed on Thor
Example: https://thor.dosomething.org/us/campaigns/baby-its-cold-inside

refs: https://github.com/DoSomething/phoenix/pull/6127

sbsmith86:
#### What's this PR do?

Overwrites some dumb drupal logic that sets the end date to the start date even if you don't choose "show end date" on the run date field on campaign run nodes. Now, If a user doesn't select "show run date" we will store the end date as `NULL`

Also updates `dosomething_helpers_get_campaign_status()` to account for the fact that we no longer have equal start and end dates 1) Because the end date will always be set to just before midnight so start and end dates won't be the same. 2) Because the end date will be NULL. 

here probably: `dosomething_campaign_run_node_validate()`;

Fixes #6144 
Fixes #6123

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module
+++ b/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.module
@@ -294,6 +294,29 @@ function dosomething_campaign_run_form_campaign_run_node_form_alter(&$form, &$fo
   drupal_add_css(drupal_get_path('module', 'dosomething_campaign_run') . '/campaign_run_node.css');
 }
 
+/*
+ * Implements hook_node_validate()
+ *
+ * On the run date field on campaign run nodes, we want to store the end date as NULL
+ * if the user doesn't choose to "show end date" on the form.
+ * This is needed to overwrite default drupal behavior that stores the start date in the end date
+ * if the user doesn't specify one explicity.
+ *
+ * See https://www.drupal.org/node/874322 for more on this issue.
+ */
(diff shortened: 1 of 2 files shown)


itsjoekent:
are we worried about fixing existing field values that :drupal: broke?

sbsmith86:
@deadlybutter what do you mean

angaither:
:+1:

sbsmith86:
@deadlybutter oh! do you mean like the dates that were set as equal? hmmm we could do that. @angaither ?

itsjoekent:
@sbsmith86 I thought there was rows in the DB already where start date == end date. not to block this PR merge, but i wasnt sure if at some point we should update all of these rows

angaither:
oh, not sure what you're suggestion is @deadlybutter but yes... there are currently nodes in prod that have end date set.

itsjoekent:
copying from slack

```

Joe Kent [4:35 PM] 
re: the comment in shae’s PR

[4:35] 
for all of the instances where the end date is bad because of that drupal bug

[4:35] 
shouldnt we go back and fix all of those bad rows?
```

not like a huge deal was just curious

PR #6146, opened 2016-02-04
Add support for old campaign runs under the drupal curse

### LINKED ISSUE #6144: campaigns with current runs w/o end dates closed on Thor

sbsmith86:
#### What's this PR do?

Pull Request #6145 addressed a core Drupal issue that set end dates equal to start dates by default, even if "show end date" was not selected by the admin by setting the end date to NULL explicitly. We then updated the logic to not care about start and end dates that are equal.  

HOWEVER, before this fix was in, there were about 30 runs that had been created and saved that drupal had already done this nonsense for. 

This PR, adds a conditional to `dosomething_helpers_campaign_get_status()` function that first checks if the start and end dates are equal and in the past, in which case the campaign should still be active. 

Fixes #6144


angaither:
👍🏽

PR #6152, opened 2016-02-05
Update to correct logic in passed intro images.

### LINKED ISSUE #6150: Intro container image bug
image

Bug where the template is incorrectly thinking there is an image for the intro block and forcing the text to only within the half column.

weerd:
Fixes #6150 
#### What's this PR do?

There was a bug that was returning an empty `` tag for the `$intro_image` variable, so the template was thinking there was something there and placing the text to the left column to make room for the image. But since no image was present, it was looking wonky.

This PR updates the logic to return `NULL` if there is no path to an image instead of returning an empty `` tag.

In the **\* dosomething_image.module*\* file.
#6150
#### Screenshots

**Before:**
image

**After:**
image

---

@DFurnes 

CC @angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_image/dosomething_image.module
+++ b/lib/modules/dosomething/dosomething_image/dosomething_image.module
@@ -41,12 +41,20 @@ function dosomething_image_get_themed_image($nid, $ratio, $style, $alt = '', $at
   else {
     // Get the path of the themed image.
     $path = dosomething_image_get_themed_image_url($nid, $ratio, $style);
+
+    if (!$path) {
+      return NULL;
+    }
+
     // Set drupal attributes.
     $attributes = drupal_attributes($attributes);
+
     // Construct themed image tag.
     $themed_image =  '<img src="' . check_url($path) . '" alt="' . check_plain($alt) . '"' .  $attributes .'>';
+
     // Store the themed image.
     cache_set('ds_img_' . $nid . '_' . $ratio . '_' . $style, $themed_image, 'cache_dosomething_image');
+    
     return $themed_image;
   }
 }
--- a/lib/modules/dosomething/dosomething_image/dosomething_image.module
+++ b/lib/modules/dosomething/dosomething_image/dosomething_image.module
@@ -29,8 +29,8 @@ function dosomething_image_form_image_node_form_alter(&$form, &$form_state, $for
  * @param array $attributes
  *  Optional - key/value paired array of attributes to add to the image.
  *
- * @return array
- *  A themed image tag with no set height/width.
+ * @return string|null
+ *  A themed image tag with no set height/width or null if no image path.
  */
 function dosomething_image_get_themed_image($nid, $ratio, $style, $alt = '', $attributes = array()) {
   // Attempt to get that image from cache.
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_image/dosomething_image.module
```diff
@@ -41,12 +41,20 @@ function dosomething_image_get_themed_image($nid, $ratio, $style, $alt = '', $at
   else {
     // Get the path of the themed image.
     $path = dosomething_image_get_themed_image_url($nid, $ratio, $style);
+
+    if (!$path) {
+      return NULL;
```
DFurnes, line 46: Would an empty string make more sense here, since it seems like it otherwise returns an HTML string?

DFurnes:
Totally unrelated to this PR, but it looks like the docblock should be updated to `@return string` (if I'm not misreading things).

itsjoekent:
The top IF statement doesnt return a string, I think

https://github.com/DoSomething/phoenix/blob/image-logic-bug/lib/modules/dosomething/dosomething_image/dosomething_image.module#L37

DFurnes:
@deadlybutter It looks like that cache key is just [being set from the `$themed_image` value](https://github.com/DoSomething/phoenix/blob/0fd1d0a13d186e20a9582a8a1ceae99379e44bc5/lib/modules/dosomething/dosomething_image/dosomething_image.module#L56), right?

weerd:
@DFurnes Possibly, but this is where the function is being called and actually used: https://github.com/DoSomething/phoenix/blob/657e63e61b5dacf586228738ee2e6cca0b8b8102/lib/modules/dosomething/dosomething_static_content/dosomething_static_content.module#L64

And then that assigned var is checked here: https://github.com/DoSomething/phoenix/blob/0fd1d0a13d186e20a9582a8a1ceae99379e44bc5/lib/themes/dosomething/paraneue_dosomething/templates/static_content/node--static_content.tpl.php#L20

So I feel `NULL` makes sense since it checks `isset()` and an empty string would be considered `TRUE`.

itsjoekent:
@DFurnes No I think themed image is the data of the cache
`cache_set('ds_img_' . $nid . '_' . $ratio . '_' . $style, $themed_image, 'cache_dosomething_image');`
`cache_set($cid, $data, $bin = 'cache', $expire = CACHE_PERMANENT)`

https://api.drupal.org/api/drupal/includes!cache.inc/function/cache_set/7

weerd:
What the cache returns is a string of the formatted ``.

@deadlybutter it returns the same value assigned here: https://github.com/DoSomething/phoenix/blob/657e63e61b5dacf586228738ee2e6cca0b8b8102/lib/modules/dosomething/dosomething_image/dosomething_image.module#L47

It's just that when items get cached, Drupal places them in cache object, and there's a `data` property in there that actually contains the value, whatever it might be :)

weerd:
@DFurnes and yep, I'll update docblock!

DFurnes:
It feels a bit nicer to me to always return a string, and just check `!empty()`. Buuuut I think both approaches are 100% fine, so no need to bikeshed over. If we want to conditionally return null, let's just update the return annotation to `@returns string|null`?

DFurnes:
@weerd :+1: :+1:

PR #6154, opened 2016-02-05
change isset to !empty

### LINKED ISSUE #6149: FAQs line appears in Contact Us Modal even if there are no FAQs
See https://thor.dosomething.org/us/campaigns/shower-songs as an example

Refs: #6062 and #6091

katiecrane:
Fixes #6149 

Change from checking if the FAQ array exists to checking if there are any actual FAQs in there. Last time the main example I was looking at did not have this array unless there were FAQs, but I guess that is not always the case.

To test:

As both an admin and regular user, click "Contact Us" for a variety of campaigns both with and without FAQs. There should only be a link to FAQs if the campaign has them.

cc: @mikefantini, @DFurnes


DFurnes:
:+1:

itsjoekent:
screen shot 2016-02-05 at 2 55 21 pm

you went too far with this PR, you actually broke github

PR #6161, opened 2016-02-10
Drupal dependency fixes

### LINKED ISSUE #6160: Issue with missing dependencies on resource transformer classes
```
Fatal error: Class 'SignupTransformer' not found in /var/www/staging.beta.dosomething.org/releases/20160203142447/lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
```

weerd:
Fixes #6160
#### What's this PR do?

This PR addresses some fixes to potentially help with dependency issues when Drupal is bootstrapping itself and grabbing dependencies for modules that live as files in other modules. So basically aiming to fix INCEPTION!

Just take a look at the **[Resource]Transformer.php** files and the **dosomething_[resource].module** files and make sure the specified dependencies seem correct. Also check out the **dosomething_[resource].info** files, since some of those were updated too.

Well, if you really want to, get ready to go grab some :bread: and make a sandwich, cause it'll take a while. You'll need to pull down this branch, then wipe you local _shiz_ and do a `ds build --install` and a `ds pull stage --db` making sure there isn't a ton of red :rotating_light: errors about **Transformer** related classes not being found, etc.

You could also hit a few of the Phoenix endpoints to test and make sure the API is still returning responses :smile_cat: 
#6160

---

@aaronschachter @angaither @chloealee @DFurnes

### DIFF
--- a/lib/modules/dosomething/dosomething_api/dosomething_api.module
+++ b/lib/modules/dosomething/dosomething_api/dosomething_api.module
@@ -4,17 +4,23 @@
  * Code for the dosomething_api feature.
  */
 
+module_load_include('php', 'dosomething_campaign', 'includes/CampaignTransformer');
+module_load_include('php', 'dosomething_kudos', 'includes/KudosTransformer');
+module_load_include('php', 'dosomething_reportback', 'includes/ReportbackTransformer');
+module_load_include('php', 'dosomething_reportback', 'includes/ReportbackItemTransformer');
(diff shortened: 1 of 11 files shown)


aaronschachter:
LGTM .. @sergii-tkachenko I know you've jumped in at times when we've run into similar issues - do you mind giving this a quick sanity check?

sergiitk:
Looks good.

weerd:
Yay! Thanks @aaronschachter @sergii-tkachenko

PR #6162, opened 2016-02-10
Signups endpoint 2

chloealee:

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -412,14 +412,21 @@ abstract class Transformer {
    *   An object containing properties of Signup data:
    *   - id: (string) The Signup id.
    *   - created_at: (string) Date Signup was created.
-   *   - campaign: (string) Campaign Signup belongs to.
    *   - campaign_run: (string) Campaign run Signup belongs to.
+   *   - uri: (string) API URI for Reportback Item data.
+   *   - campaign: (string) Campaign Signup belongs to.
    * @return array
    */
   protected function transformSignup($data) {
+    // print_r($data);
+    // die();
     return [
       'id' => $data->id,
       'created_at' => $data->created_at,
+      'campaign_run' => [
+        'id' => $data->campaign_run,
+        'current' => 'true',
+      ],
       'uri' => $data->uri,
     ];
   }
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -424,6 +424,10 @@ abstract class Transformer {
       'campaign_run' => [
         'id' => $data->campaign_run,
         'current' => $current,
+        'timing' => [
+          'start' => 'start-date-placeholder',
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -412,15 +412,25 @@ protected function transformReportbackItem($data) {
...
+      'campaign_run' => [
+        'id' => $data->campaign_run,
+        'current' => $current,
+        'timing' => [
+          'start' => 'start-date-placeholder',
+          'end' => 'end-date-placeholder',
+        ],
+      ],
       'uri' => $data->uri,
+      'reportbacks' => $data->reportbacks,
```
weerd: Reportbacks should be included in the `SignupTransformer` after calling `$data['campaign'] = $this->transformCampaign((object) $item->campaign);` and sending to the `$this->transformReportback()` method.

lib/modules/dosomething/dosomething_signup/includes/SignupTransformer.php
```diff
@@ -61,16 +61,27 @@ public function show($id) {
    * @return array
    */
   protected function transform($item) {
+    // print_r($item);
+    // die();
```
weerd: debug!

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
```diff
@@ -143,3 +143,26 @@ function dosomething_reportback_filter_flagged_reportbacks($params) {
...
+ *
+ * @param string $run_id Campaign run id. 
+ * @param string $sid Signup id. 
+ * @return mixed
+ */
+function dosomething_reportback_filter_reportbacks_by_signups($run_id, $sid) {
+  $reportbacks = db_query("SELECT DISTINCT s.sid, r.rbid, r.run_nid, s.run_nid
+                          FROM dosomething_signup s 
+                          INNER JOIN dosomething_reportback r on s.run_nid = r.run_nid
+                          WHERE s.uid = r.uid;");
```
weerd: So this feels like its on the right track, but you could likely do it all in the `dosomething_signup.query.inc` file, so it's all in one hit to the DB. Just add it to the query there with a join on the conditions :) Then in the **SignupTransformer.php** you can do  `$this->reportbacks = Reportback::get($rbid)` without the try catch cause it should return just one :)

chloealee: awesome, thanks Diego! Makes so much more sense! One question - could a person technically have multiple reportbacks associated with one signup?

chloealee: Also, `dosomething_reportback_filter_reportbacks_by_signups` does some logic to get rid of the reportbacks with `reportback_run` of `0` and return a user's reportbacks by signup id. 

I'm not sure where to put that logic - it seems like it'd be too much going on if we placed after `$result = $query->execute();` on line 28 in `dosomething_signup.query.inc`?

chloealee:
Closing this PR - continuing work in #6168.

PR #6165, opened 2016-02-12
adds campaign_run data

### LINKED ISSUE #6153: Include Campaign Run property in Signup response object
### Updated description

Include the Signup's Campaign Run id as well as whether Campaign Run is the Campaign's current Run. This will allow us to query for a user's "Prove it" campaigns (current campaigns which they need to complete / reportback on)
### Original description

Refs #6110

Discussed how the GET `signups` index will work with @weerd, he mentioned ability to filter by a User ID and by Campaign ID's.  The mobile app needs to know which of those Signups are for current Campaign Runs. Ideally we could support passing a query parameter like `current=true` or `active_campaigns=true` which would do some heavy lifting by joining the Signups Run on a Campaign's Current Run.

Including the Reportback information within the Signup object (#6151) would make it easy to find a User's current "Prove It" campaigns, we could just loop through a User's `current=true`  and look for Signups that don't have a Reportback.  Way less critical, but additionally could see a use for another param like `completed=false` or `reported_back=false` to let the API handle it.

chloealee:
#### What does this PR do?

Adds the `current_run` the signup is associated with and returns a boolean if the signup's current run is equal to the campaign's current run. 

When hitting `/signup` endpoints, 

```
campaign_run: {
  id: "1772",
  current: true
},
```

should be exposed. 
#### Relevant Tickets

Fixes #6153

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -1,4 +1,4 @@
-  <?php
+<?php
 
 abstract class Transformer {
 
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -1,4 +1,4 @@
-<?php
+  <?php
 
 abstract class Transformer {
 
@@ -414,12 +414,17 @@ abstract class Transformer {
    *   - created_at: (string) Date Signup was created.
    *   - campaign: (string) Campaign Signup belongs to.
    *   - campaign_run: (string) Campaign run Signup belongs to.
+   *   - uri: (string) API URI for Signup data.
    * @return array
    */
-  protected function transformSignup($data) {
+  protected function transformSignup($data, $current) {
     return [
       'id' => $data->id,
       'created_at' => $data->created_at,
+      'campaign_run' => [
+        'id' => $data->campaign_run,
+        'current' => $current,
+      ],
       'uri' => $data->uri,
     ];
   }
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -1,4 +1,4 @@
-<?php
+  <?php
```
aaronschachter: :mouse:

aaronschachter:
@chloealee Can you include in the description the results look like?  If all looks good, let's get this merged in first to close out #6153

lib/modules/dosomething/dosomething_signup/includes/SignupTransformer.php
```diff
@@ -66,7 +66,16 @@ protected function transform($item) {
...
 
-    $data += $this->transformSignup($item);
+    $campaign = (object) $item->campaign;
+    $current_run = $campaign->campaign_runs['current']['en']['id'];
+
+    $current = false;
+
+    if ($item->campaign_run == $current_run) {
+      $current = true;
+    }
```
aaronschachter: Minor, but this could just be simplified as `$current = ($item->campaign_run == $current_run);`

chloealee:
@aaronschachter sure thing! Here's the result for http://dev.dosomething.org:8888/api/v1/signups/93

```
{
  data: {
    id: "93",
    created_at: "1394754616",
    campaign_run: {
      id: "1772",
      current: true
    },
    uri: "http://dev.dosomething.org:8888/api/v1/signups/93",
    campaign: {
      id: "48",
      title: "Faces of Immigration",
      campaign_runs: {
        current: {
          en: {
            id: "1772"
          }
        },
        past: [ ]
      },
      language: {
        language_code: "en",
        prefix: "us"
      },
      translations: {
        original: "en",
        en: {
          language_code: "en",
          prefix: "us"
        }
      },
      reportback_info: {
        copy: "Show us your selfie!",
        confirmation_message: "Thanks for using your face to promote immigration reform! ",
        noun: "pictures",
        verb: "shared"
      }
    }
  }
}
```

lib/modules/dosomething/dosomething_signup/includes/SignupTransformer.php
chloealee: oh much cleaner!

aaronschachter:
LGTM :+1:

PR #6168, opened 2016-02-16
db_query works to get reportback id

### LINKED ISSUE #6151: Embed Reportback and Reportback Items within Signup response
Now that we're using Phoenix as the source of a new Signups endpoint (#6110), it'd be great to include a Signup's corresponding Reportback if it exists.

This could allow a client to render a User Profile in a single API request (like how the `user/:id/campaigns` Northstar endpoint currently does).

chloealee:
#### What does this PR do?

Exposes report back information in the /signup endpoint return value.  

Check all /signups endpoints (and combos):
- `/signups`
- `/signups?user=12&campaigns=6`
- `/signups?user=12`
- `/signups?campaigns=6`
- `/signups/1849`

Should return data that looks like the below: 

```
{
  data: {
  id: "1849",
  created_at: "1430250074",
  campaign_run: {
    id: "1227",
    current: true
  },
  uri: "http://dev.dosomething.org:8888/api/v1/signups/1849",
  campaign: {
    id: "362",
    title: "Comeback Clothes",
   campaign_runs: {
     current: {
       en: {
         id: "1227"
       }
     },
     past: [ ]
   },
  language: {
    language_code: "en",
    prefix: "us"
  },
  translations: {
    original: "en",
    en: {
      language_code: "en",
      prefix: "us"
    },
    en-global: {
      language_code: "en-global",
      prefix: null
    }
  },
  reportback_info: {
    copy: "Submit your photo before June 20 to qualify for the scholarship.",
    confirmation_message: "Going green is always in style! Thanks for helping the Earth.",
    noun: "Items",
    verb: "Recycled"
    }
  },
  reportbacks: {
    id: "3361",
    created_at: "1443482386",
    updated_at: "1454529489",
    quantity: 3,
    uri: "http://dev.dosomething.org:8888/api/v1/reportbacks/3361",
    why_participated: "It's time!",
    flagged: false,
    reportback_items: {
      total: 1,
      data: [
        {
          id: "5402",
          status: "excluded",
          caption: "It's time!",
          uri: "http://dev.dosomething.org:8888/api/v1/reportback-items/5402", 
          media: {
            uri:   "http://dev.dosomething.org:8888/sites/default/files/styles/480x480/public/reportbacks/1377/uid_1705385-nid_1377-0.jpg?itok=PzCBHPpw",
            type: "image"
          },
          created_at: "1443482386",
          kudos: {
            data: [ ]
          }
        }
      ]
    }
  }
}
}
```

Fixes #6151 

Cc @aaronschachter

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
@@ -41,8 +41,8 @@ function dosomething_signup_get_signups_query($params) {
 function dosomething_signup_build_signups_query($params = []) {
   $query = db_select('dosomething_signup', 's');
   $query->join('dosomething_reportback', 'rb', 's.nid = rb.nid AND s.run_nid = rb.run_nid');
-  $query->fields('s', ['sid', 'uid', 'nid', 'run_nid']);
-  $query->fields('rb', ['rbid', 'uid', 'nid', 'run_nid']);
+  $query->fields('s', ['sid', 'uid', 'nid', 'run_nid', 'timestamp']);
+  $query->fields('rb', ['rbid']);
   $query->where('s.uid = rb.uid');
 
   if (isset($params['user'])) {
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
@@ -17,18 +17,16 @@
  */
 function dosomething_signup_get_signups_query($params = []) {
   $query = dosomething_signup_build_signups_query($params);
-  // $offset = dosomething_helpers_isset($params, 'offset', 0);
-  // $count = dosomething_helpers_isset($params, 'count', 25);
+  $offset = dosomething_helpers_isset($params, 'offset', 0);
+  $count = dosomething_helpers_isset($params, 'count', 25);
 
-  // if ($count && $count !== 'all') {
-  //   $query->range($offset, $count);
-  // }
+  if ($count && $count !== 'all') {
+    $query->range($offset, $count);
+  }
 
   $result = $query->execute()->fetchAll();
 
   return $result;
-
-  // return array_keys($result['signup']);
 }
 
 /**
@@ -41,11 +39,6 @@ function dosomething_signup_get_signups_query($params = []) {
  * @return SelectQuery object
  */
 function dosomething_signup_build_signups_query($params) {
-// $query = db_query('SELECT *
-//                    FROM dosomething_signup s 
-//                    INNER JOIN dosomething_reportback rb on s.nid = rb.nid and rb.run_nid = s.run_nid
-//                    WHERE s.uid = rb.uid;');
-
   $query = db_select('dosomething_signup', 's');
   $query->join('dosomething_reportback', 'rb', 's.nid = rb.nid AND s.run_nid = rb.run_nid');
   $query->fields('s', ['sid', 'uid', 'nid', 'run_nid']);
@@ -56,13 +49,13 @@ function dosomething_signup_build_signups_query($params) {
     $query->condition('s.uid', $params['user'], '=');
   }
 
-  // if (isset($params['campaigns'])) {
-  //   if (is_array($params['campaigns'])) {
-  //     $query->condition('s.nid', $params['campaigns'], 'IN');
-  //   } else {
-  //     $query->condition('s.nid', $params['campaigns'], '=');
-  //   }
-  // }
+  if (isset($params['campaigns'])) {
+    if (is_array($params['campaigns'])) {
+      $query->condition('s.nid', $params['campaigns'], 'IN');
+    } else {
+      $query->condition('s.nid', $params['campaigns'], '=');
+    }
+  }
 
   return $query;
 }
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
@@ -10,6 +10,7 @@
  *
  * @param  array  $params
  *   An associative array of conditions to filter by. Possible keys:
+ *  - sid (string|array)
  *  - user (string)
  *  - campaigns (string|array)
  * @return object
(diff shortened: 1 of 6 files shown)


chloealee:
@weerd Wanted to get your :eyes: on the /signups endpoint. Everything else seems to work but for some reason when  you don't add query params, it returns no data. I found that it seems to break on potentially on line 93 in Signup.php (with the `Reportback::get($data->rbid);` function). Not sure why this would break when not adding optional query params?

weerd:
hmmm, ok. I'm taking a :eyes:

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -92,5 +90,6 @@ private function build($data) {
     $this->created_at = $data->timestamp;
```
weerd, line 90: The `timestamp` property isn't available on the returned data and it's :hankey:ing on that. If you add `timestamps` as a field to grab info from in the query here:

```
$query->fields('s', ['sid', 'uid', 'nid', 'run_nid', 'timestamp']);
```

then all results should have the timestamp available for you to use ;)

lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
```diff
@@ -38,23 +38,24 @@ function dosomething_signup_get_signups_query($params = []) {
...
  */
-function dosomething_signup_build_signups_query($params) {
-  $query = new EntityFieldQuery();
-  $query
-    ->entityCondition('entity_type', 'signup');
+function dosomething_signup_build_signups_query($params = []) {
+  $query = db_select('dosomething_signup', 's');
+  $query->join('dosomething_reportback', 'rb', 's.nid = rb.nid AND s.run_nid = rb.run_nid');
+  $query->fields('s', ['sid', 'uid', 'nid', 'run_nid']);
+  $query->fields('rb', ['rbid', 'uid', 'nid', 'run_nid']);
```
weerd: I think you can eliminate most of the fields in this call, since all we only need the `rbid` from the `rb` table, otherwise it's just duplicate data:

```
$query->fields('rb', ['rbid']);
```

Currently this is obtained for a result:

```
stdClass Object
(
    [sid] => 32
    [uid] => 12
    [nid] => 6
    [run_nid] => 1740
    [timestamp] => 1394231397
    [rbid] => 10
    [rb_uid] => 12
    [rb_nid] => 6
    [rb_run_nid] => 1740
)
```

and there's duplicate data, just with a different key name :)

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -92,5 +90,6 @@ private function build($data) {
...
     $this->campaign = Campaign::get($data->nid);
     $this->campaign_run = $data->run_nid;
+    $this->rbid = Reportback::get($data->rbid);
```
weerd: You're right on line 93 causing an issue. There seems to be a problem with some of the staging data on some super early reportbacks (like rbid 2 & 3 for example) where while there's an RB there are absolutely no RB Items in the db for it and it totally :hankey:s out :scream: However, since the `Reportback::get()` throws an exception we can catch it and deal with it. So I tried something like this, and the index collection started working correctly:

```
    $this->id = $data->sid;
    $this->campaign = Campaign::get($data->nid);

    try {
      $this->rbid = Reportback::get($data->rbid);
    }
    catch (Exception $error) {
      $this->rbid = null;
    }
  }
```

lib/modules/dosomething/dosomething_signup/includes/SignupTransformer.php
```diff
@@ -75,6 +73,8 @@ protected function transform($item) {
 
     $data['campaign'] = $this->transformCampaign((object) $item->campaign);
 
+    $data['reportbacks'] = $this->transformReportback($item->rbid);
```
aaronschachter: Can we change this to just `reportback` ? Each Signup will only have one Reportback, not multiple.

lib/modules/dosomething/dosomething_signup/includes/Signup.php
chloealee, line 90: oohh shoot I wouldn't have caught that - thanks Diego!

lib/modules/dosomething/dosomething_signup/includes/Signup.php
chloealee: ooooh this is so smart! I just started to actually understand try/catch blocks - so helpful!

chloealee:
Thanks @weerd ! Looks like it's all working now on my end!

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -92,5 +90,12 @@ private function build($data) {
     $this->created_at = $data->timestamp;
     $this->campaign = Campaign::get($data->nid);
     $this->campaign_run = $data->run_nid;
+
+    try {
+      $this->rbid = Reportback::get($data->rbid);
```
weerd: one quickie, forgot we should probs name this property `$this->reportback = ...` and not `rbid` since it'll return a reportback object, not just the rbid

chloealee: smart - updated!

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -92,5 +90,12 @@ private function build($data) {
     $this->created_at = $data->timestamp;
     $this->campaign = Campaign::get($data->nid);
     $this->campaign_run = $data->run_nid;
+
+    try {
+      $this->reportback = Reportback::get($data->rbid);
+    }
+    catch (Exception $error) {
+      $this->rbid = null;
```
weerd: this one too!

lib/modules/dosomething/dosomething_signup/includes/SignupTransformer.php
```diff
@@ -68,12 +66,10 @@ protected function transform($item) {
 
     $campaign = (object) $item->campaign;
     $current_run = $campaign->campaign_runs['current']['en']['id'];
-
     $current = ($item->campaign_run == $current_run);
-
     $data += $this->transformSignup($item, $current);
-
     $data['campaign'] = $this->transformCampaign((object) $item->campaign);
+    $data['reportback'] = $this->transformReportback((object) $item->rbid);
```
weerd: update the `$item->rbid` here too I think... should be `$item->reportback` now, right?

chloealee:
@weerd thanks for catching the logs! All set :)

weerd:
Awesome! Looks snazzy :sparkles:  :sheep: IT!

PR #6171, opened 2016-02-16
Update to return empty array response for no results in collection requests.

### LINKED ISSUE #6163: API requests with no results should return empty array instead of error message
The request is successful, but we get an error message. TouchLab had asked for an empty array instead, this may make error handling in a little easier, as this isn't an actual error.
- [x] signups 
- [x] reportbacks
- [ ] campaigns

e.g. https://staging.dosomething.org/api/v1/reportback-items?campaigns=1

screen shot 2016-02-11 at 11 12 15 am

cc @weerd @jonuy

weerd:
Fixes #6163
#### What's this PR do?

This PR updates responses from the index collection for `/reportbacks` and `/reportback-items` so that if no results is obtained, it responds with an empty array instead of an error object.

Quick :eyes: on the quick changes.

Pull down branch, try and hit index collection endpoints for `/reportbacks` and `/reportback-items` for something you know shouldn't have results and see if it returns an empty array.
#6163

_NOTE:_ added a comment about possibly sending the caught error in try catch to `watchdog()` but it would be more ideal to pass along the full query which I couldn't yet without parsing it and stringifying it (which currently is only a method available on the `ApiCache` class. :grimacing: 

---

@DFurnes @aaronschachter 

CC: @angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/includes/CampaignTransformer.php
+++ b/lib/modules/dosomething/dosomething_campaign/includes/CampaignTransformer.php
@@ -41,10 +41,9 @@ class CampaignTransformer extends Transformer {
       $total = dosomething_campaign_get_campaign_query_count($filters);
     }
     catch (Exception $error) {
+      // @TODO: Potentially log error to watchdog.
       return [
-        'error' => [
-          'message' => $error->getMessage(),
-        ],
+        'data' => [],
       ];
     }
 
--- a/lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php
+++ b/lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php
@@ -18,7 +18,9 @@ class ReportbackItemTransformer extends ReportbackTransformer {
     }
     catch (Exception $error) {
       // @TODO: Potentially log error to watchdog.
-      return [];
+      return [
+        'data' => [],
+      ];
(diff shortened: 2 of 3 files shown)


angaither:
such api, much empty wow! :+1:

aaronschachter:
@weerd Can you provide a sample response? Is the `data` property returned as an empty array, or its just simply an empty array? Not sure what standard is...

aaronschachter:
@weerd We also need this for the `signups` endpoint for rendering a User's profile.

aaronschachter:
[This thread](http://programmers.stackexchange.com/a/199831) seems to suggest keeping the `data` property but its set to an empty array (which @chloealee and I [got working by just simply removing the `if` block that throws the exception](https://github.com/DoSomething/phoenix/issues/6163#issuecomment-183473373))

It makes the client code a little simpler, don't have to check for the existence of a `responseData.data`

weerd:
I'd rather keep the caught exception in case we can use it for something later on down the road. I can just update the return to include `data` object. I just forgot that it should ;)

weerd:
@chloealee see https://github.com/DoSomething/phoenix/pull/6171#issuecomment-184874114 Need a similar response update to the `/signups` index collection response :)

weerd:
@aaronschachter updated the response ;)

aaronschachter:
:+1:

PR #6172, opened 2016-02-16
Adds run nid to view

### LINKED ISSUE #6007: Add run filtering to admin views

itsjoekent:
#### What's this PR do?

Adds run nid filter to reportback admin view

Does it filter out RB's with a diff run nid than specified?

Showed Andrea in Slack that signup filtering appears to work already
https://thor.dosomething.org/us/node/6163/signups?uid=&mail=&source=&nid=6166&page=5

Fixes #6007

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.info
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.info
@@ -21,4 +21,4 @@ features[user_permission][] = view any reportback
 features[user_permission][] = view own reportback
 features[views_view][] = reportback_files
 files[] = dosomething_reportback.test
-mtime = 1444314429
+mtime = 1446143311
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.info
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.info
@@ -21,4 +21,4 @@ features[user_permission][] = view any reportback
 features[user_permission][] = view own reportback
 features[views_view][] = reportback_files
 files[] = dosomething_reportback.test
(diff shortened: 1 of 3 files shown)


angaither:
@deadlybutter pics?

sergiitk:
:+1:

itsjoekent:
@angaither 

screen shot 2016-02-16 at 3 46 35 pm

angaither:
@deadlybutter is that on all reportback views or just the reviewed tab?

itsjoekent:
actually hmm i might have noticed something wrong, let me double check

itsjoekent:
@angaither  where else is this view found besides the review tab?

itsjoekent:
yeah this isnt working, i made the above RB (in the screenshot) have a NULL run nid and its still showing up

angaither:
@deadlybutter AFAIK it's only the inbox and reviewed tab, but to avoid issues with form api and the infamous reportback reviewing bug, i think we only want this on the reviewed tab
image

itsjoekent:
@sergii-tkachenko so @angaither said the dupes are alright, i guess ready for re-review

itsjoekent:
screen shot 2016-02-18 at 11 21 54 am

angaither:
@deadlybutter we probably want run id after `campaign` and it should link to the run node

itsjoekent:
okie dokes

itsjoekent:
@angaither done

angaither:
:+1:

PR #6174, opened 2016-02-16
Update to allow viewing unreviewed RBs if admin.

weerd:
Fixes DoSomething/northstar#279
#### What's this PR do?

This PR allows admins to retrieve unreviewed and/or flagged Reportbacks when requesting from `/reportbacks/:id` endpoint. Also updates the assignment of the `flagged` property on the Reportback class so it will now indicate `null` if a RB has not been reviewed, and `FALSE` if not flagged and `TRUE` if flagged, as opposed to just the latter two.

Check out the query edits and see if it makes sense!

This only works for the _show_ endpoint and doesn't work for the index collection endpoint. We would need to do some further work to allow unreviewed RBs to show if querying via the `/reportbacks` index collection endpoint do the the array of params that need to be accounted for (ie: such as the `status` param).

DoSomething/northstar#279

---

@DFurnes @aaronschachter 

CC: @angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
@@ -55,11 +55,11 @@ function dosomething_reportback_build_reportbacks_query($params = array()) {
     }
   }
 
-  if (!isset($params['flagged'])) {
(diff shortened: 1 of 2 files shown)


angaither:
:+1:

DFurnes:
:+1:

PR #6179, opened 2016-02-17
ADDED AN ISSUE TEMPLATE.

angaither:
so we can have this every.single.time. we make new issues


namimody:
+1 :dancer: The only PR I will ever +1.

mshmsh5000:
Oh man this PR is so important + good.

PR #6182, opened 2016-02-18
Competition signup prompt

### LINKED ISSUE #6169: Competitions signup popup
### signup form updates
- add 'competition' type to signup data form
- don't display this link in the from `stuff you need section` if competition, by making link text not required
- add fields to customize the submit and skip button text
### db updates
- add a col to `dosomething_signup` to have competition

this could be either boolean value, or a boolean and when the user actually gets divided into a group, it updates to the gladiator competition id

sbsmith86:
#### What's this PR do?
- add 'competition signup' checkbox to the sign up data form that allows admins to specify this form is to sign up for a competition
- Makes the link text to the modal not required so that it doesn't show in the "Stuff you need" section
- Add fields to customize the submit and skip button text
- Adds a `competition` column to `dosomething_signup` and when a user submits the sign up data form that is a competition sign up, we update the sign up record to flag this column as true.

Fixes #6169 - This PR should fulfill what is needed to start. But I did bring up something in that issue that we might need to think through a little further. This sorta hijacks the sign up data form to use for competitions. But, what happens when we both want to allow users to sign up for a competition but we also need them to fill out extra data as a part of the campaign (i.e teens for jeans, thumb wars)

Maybe we should make a separate issue for that?

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.install
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.install
@@ -115,6 +115,20 @@ function dosomething_signup_schema() {
         'not null' => TRUE,
         'default' => '',
       ),
+      'submit_text' => array(
+        'description' => 'The label of the submit button',
+        'type' => 'varchar',
+        'length' => 255,
+        'not null' => TRUE,
+        'default' => '',
+      ),
+      'skip_text' => array(
+        'description' => 'The label of the skip button',
+        'type' => 'varchar',
+        'length' => 255,
+        'not null' => TRUE,
+        'default' => '',
+      ),
       'form_header' => array(
         'description' => 'Form header text',
         'type' => 'varchar',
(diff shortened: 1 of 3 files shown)


angaither:
@sbsmith86 yeah, I was debating if we would want to break it out into a different signup section, but I think we would need to flesh out what the experience would be if we need both forms.

lib/modules/dosomething/dosomething_signup/dosomething_signup.install
```diff
@@ -647,3 +673,53 @@ function dosomething_signup_update_7023() {
...
+
+/**
+ * Adds competition column to the dosomething_signup table.
+ */
+function dosomething_signup_update_7026() {
+  $tbl_name = 'dosomething_signup';
+  // Load schema to get table definition.
+  $schema = dosomething_signup_schema();
+  // New field to add.
+  $field_name = 'competition';
```
angaither, line 719: these can all be moved into one update hook

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -104,6 +110,18 @@ function dosomething_signup_node_signup_data_form($form, &$form_state, $node) {
...
+  $form[$fieldset]['config'][$prefix . 'submit_text'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Submit Button Text'),
+    '#default_value' => $values['submit_text'],
+    '#description' => t('This will be used as the label of the submit button.'),
+  );
+  $form[$fieldset]['config'][$prefix . 'skip_text'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Skip button text'),
+    '#default_value' => $values['skip_text'],
```
angaither, line 122: these should probably have defaults

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -442,9 +458,10 @@ function dosomething_signup_user_signup_data_form($form, &$form_state, $signup)
     '#suffix' => '</div></div>',
   );
 
+  $submit_label = ($config['submit_text']) ? $config['submit_text'] : 'Submit';
```
angaither, line 461: ahhh, that's done here :neckbeard:

angaither:
testing locally, the form doesn't actually show up if link text isn't set

sbsmith86:
In order to get the form to show up, I had to remove my signup, re-signup,
and then reload the campaign page because the prompt is only supposed to
show if there has never been a response.

Shae Smith
Senior Software Engineer | DoSomething.org
@shaebryantdev

angaither:
@sbsmith86 yeah I did that, and it works, but if I removed the `link_text` from the config, the modal doesn't appear
image

angaither:
ahhh, it won't let me push changes to this pr b/c it's on your fork :cry:

PR #6183, opened 2016-02-18
Competition signup prompt v2

### LINKED ISSUE #6169: Competitions signup popup

angaither:
#### What's this PR do?
- add 'competition signup' checkbox to the sign up data form that allows admins to specify this form is to sign up for a competition
- Makes the link text to the modal not required so that it doesn't show in the "Stuff you need" section
- Add fields to customize the submit and skip button text
- Adds a `competition` column to `dosomething_signup` and when a user submits the sign up data form that is a competition sign up, we update the sign up record to flag this column as true.

Fixes #6169 - This PR should fulfill what is needed to start. But I did bring up something in that issue that we might need to think through a little further. This sorta hijacks the sign up data form to use for competitions. But, what happens when we both want to allow users to sign up for a competition but we also need them to fill out extra data as a part of the campaign (i.e teens for jeans, thumb wars)

Maybe we should make a separate issue for that?
### Background context

This is the same as pr #6182 with some additional changes

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.install
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.install
@@ -674,31 +674,17 @@ function dosomething_signup_update_7023() {
   }
 }
 
+
+
 /**
- * Adds competition_signup column to the dosomething_signup_data_form table.
+ * Adds fields to support competitions to the dosomething_signup_data_form table.
  */
 function dosomething_signup_update_7024() {
   $tbl_name = 'dosomething_signup_data_form';
   // Load schema to get table definition.
   $schema = dosomething_signup_schema();
-  // New field to add.
-  $field_name = 'competition_signup';
-  // If the field doesn't exist already:
-  if (!db_field_exists($tbl_name, $field_name)) {
-    // Add it per the schema field definition.
-    db_add_field($tbl_name, $field_name, $schema[$tbl_name]['fields'][$field_name]);
-  }
-}
-
-/**
- * Adds submit_text and skip_text columns to the dosomething_signup_data_form table.
- */
-function dosomething_signup_update_7025() {
-  $tbl_name = 'dosomething_signup_data_form';
-  // Load schema to get table definition.
-  $schema = dosomething_signup_schema();
   // New fields to add.
-  $field_names = ['submit_text', 'skip_text'];
+  $field_names = ['submit_text', 'skip_text', 'competition_signup'];
   foreach ($field_names as $name) {
     // If the field doesn't exist already:
     if (!db_field_exists($tbl_name, $name)) {
@@ -711,7 +697,7 @@ function dosomething_signup_update_7025() {
 /**
  * Adds competition column to the dosomething_signup table.
  */
-function dosomething_signup_update_7026() {
+function dosomething_signup_update_7025() {
   $tbl_name = 'dosomething_signup';
   // Load schema to get table definition.
   $schema = dosomething_signup_schema();
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.install
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.install
@@ -68,7 +68,6 @@ function dosomething_signup_schema() {
         'description' => 'Boolean indicating if the user signed up to a competition',
         'type' => 'int',
         'not null' => FALSE,
-        'default' => 0,
       ),
     ),
     'primary key' => array('sid'),
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.install
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.install
@@ -115,6 +115,20 @@ function dosomething_signup_schema() {
         'not null' => TRUE,
         'default' => '',
       ),
+      'submit_text' => array(
+        'description' => 'The label of the submit button',
+        'type' => 'varchar',
+        'length' => 255,
+        'not null' => TRUE,
+        'default' => '',
+      ),
+      'skip_text' => array(
+        'description' => 'The label of the skip button',
+        'type' => 'varchar',
+        'length' => 255,
+        'not null' => TRUE,
+        'default' => '',
+      ),
       'form_header' => array(
         'description' => 'Form header text',
         'type' => 'varchar',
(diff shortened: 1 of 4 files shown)


weerd:
In regards to this comment:

> This sorta hijacks the sign up data form to use for competitions. But, what happens when we both want to allow users to sign up for a competition but we also need them to fill out extra data as a part of the campaign (i.e teens for jeans, thumb wars)

I was sorta thinking that one approach (would require some initial lift upfront) would be that in the CMS all there is to indicate a campaign run has an associated competition is a toggle checkbox. If the checkbox is checked and a user loads the campaign after never having signed up for it, then because it's checked off, Phoenix knows to ask Gladiator for the related Competition signup info (button, skip text, etc) and can setup a separate "form" entirely form the signup one so that potentially both could exist without hijacking the normal signup form, while the competition signup form info is kept in Gladiator?

Just a thought, and likely for v2, but something to potentially consider :)

itsjoekent:
going off diegos idea -- couldnt gladiator just return all of the form HTML? in a v2 id love to see a JS script that fetches the HTML and displays it (and removes the drupal requirement)

lib/modules/dosomething/dosomething_signup/dosomething_signup.install
```diff
@@ -64,6 +64,12 @@ function dosomething_signup_schema() {
         'length' => 255,
         'not null' => FALSE,
       ),
+      'competition' => array(
+        'description' => 'Boolean indicating the use signed up to a competition',
```
weerd: typo `user`

lib/modules/dosomething/dosomething_signup/dosomething_signup.install
```diff
@@ -64,6 +64,12 @@ function dosomething_signup_schema() {
...
+        'type' => 'int',
+        'not null' => TRUE,
```
weerd: Don't we want it to be `null` if the user never encountered the competitions form for campaigns? `false` if they don't want to participate, `true` if they do participate, and `null` if they were never asked yet cause they haven't gone a competitions available campaign?

angaither:
@weerd yeah, that's a good idea. [Fantini mentioned](https://github.com/DoSomething/phoenix/issues/6169#issuecomment-185755715) that it isn't really a blocker because there's also an address collection form for this case when we want to collect both school and address

but I like the idea of gladiator housing the data for the signup form.
we can consider that a reach for this sprint :open_hands:

lib/modules/dosomething/dosomething_signup/dosomething_signup.install
angaither: yaaaa

lib/modules/dosomething/dosomething_signup/dosomething_signup.install
```diff
@@ -647,3 +673,39 @@ function dosomething_signup_update_7023() {
...
+  // Load schema to get table definition.
+  $schema = dosomething_signup_schema();
+  // New field to add.
+  $field_name = 'competition';
+  // If the field doesn't exist already:
+  if (!db_field_exists($tbl_name, $field_name)) {
+    // Add it per the schema field definition.
+    db_add_field($tbl_name, $field_name, $schema[$tbl_name]['fields'][$field_name]);
+  }
+}
```
weerd, line 710: Is there a reason this needs to be two separate updates instead of just adding `competitions` to the array of the update above?

angaither, line 710: it's two different tables being altered

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -442,9 +458,10 @@ function dosomething_signup_user_signup_data_form($form, &$form_state, $signup)
     '#suffix' => '</div></div>',
   );
 
+  $submit_label = ($config['submit_text']) ? $config['submit_text'] : 'Submit';
```
weerd: wrap `'Submit'` in `t()`

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -442,9 +458,10 @@ function dosomething_signup_user_signup_data_form($form, &$form_state, $signup)
...
   $form['actions']['submit'] = array(
     '#type' => 'submit',
-    '#value' => t('Submit'),
+    '#value' => $submit_label,
```
weerd, line 464: or maybe wrap here? so it wraps whatever the custom text is? but can't remember if `t()` gets angry if you wrap a variable...

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -523,7 +543,7 @@ function dosomething_signup_user_skip_signup_data_form($form, &$form_state, $sig
   );
   $form['submit'] = array(
     '#type' => 'submit',
-    '#value' => t('Skip'),
+    '#value' => $skip_label,
```
weerd, line 546: same same with `t()`

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
angaither, line 464: @weerd wait `t()` in which place around `$submit_label` or `'Submit'`

weerd:
> we can consider that a reach for this sprint :open_hands:

@angaither sounds good. might want to chat updates to the Gladiator schema if that's the case ;)

lib/modules/dosomething/dosomething_signup/dosomething_signup.install
weerd, line 710: oh duh! I'm silly. carry on :)

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
weerd, line 464: well if we want the custom text to be translateable I think maybe `$submit_label` but I think I remember `t()` having issues with translating a variable, or just there's some caveat we need to be aware of...

weerd, line 464: Translating variables with `t()`: https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/t/7

weerd:
#GLOBAL! :earth_asia:  :+1:

PR #6186, opened 2016-02-18
changes join to leftJoin

chloealee:
#### What's this PR do?

Changes the `join` in the `dosomething_signup_build_signups_query` to a `leftJoin` to return all signups (not just signups a user has reported back for).

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
@@ -41,7 +41,7 @@ function dosomething_signup_get_signups_query($params) {
  */
 function dosomething_signup_build_signups_query($params = []) {
   $query = db_select('dosomething_signup', 's');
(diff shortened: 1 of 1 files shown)


angaither:
:+1:

chloealee:
@angaither @weerd one thing that looks a little fishy - on my local `/signups/23` returns this response:
screen shot 2016-02-18 at 3 15 12 pm

but in the database, it looks like there is a reportback: 
screen shot 2016-02-18 at 3 16 10 pm

chloealee:
Solved via Slack: 

Diego Lorenzo [3:17 PM] 
ya

[3:17] 
that’s the f@*k up in the DB

[3:17] 
RBID’s # 2 & 3 exist! but they have absolutely no FIDs attached to them(edited)

[3:18 PM] 
from staging DB which is supposed to be impossible

[3:18] 
and thus why we needed to add the try/catch in the Signup class to deal with it…

PR #6191, opened 2016-02-19
Reportbacks signups

chloealee:
#### What's this PR do?

Changes the try/catch block to an if/else block in order to return correct report back associated with signup.

Hit endpoint `http://dev.dosomething.org:8888/api/v1/signups/?user=1704288` and compare with data in database to make sure it's returning correct information. 

Query: 

```
select s.sid, s.uid, s.nid, s.timestamp, r.nid, r.rbid
from dosomething_signup s 
left join dosomething_reportback r on r.uid = s.uid and r.nid = s.nid
where s.uid = 1704288
order by r.rbid DESC;
```

Fixes the in correct work done in #6168.

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
@@ -41,10 +41,10 @@ function dosomething_signup_get_signups_query($params) {
  */
 function dosomething_signup_build_signups_query($params = []) {
   $query = db_select('dosomething_signup', 's');
-  $query->leftJoin('dosomething_reportback', 'rb', 's.nid = rb.nid AND s.run_nid = rb.run_nid');
+  $query->leftJoin('dosomething_reportback', 'rb', 's.uid = rb.uid and s.nid = rb.nid'); 
   $query->fields('s', ['sid', 'uid', 'nid', 'run_nid', 'timestamp']);
   $query->fields('rb', ['rbid']);
-  $query->where('s.uid = rb.uid');
+  $query->orderBy('rb.rbid', 'DESC');
 
   if (isset($params['sid'])) {
     if (is_array($params['sid'])) {
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
@@ -44,7 +44,6 @@ function dosomething_signup_build_signups_query($params = []) {
   $query->leftJoin('dosomething_reportback', 'rb', 's.uid = rb.uid and s.nid = rb.nid'); 
   $query->fields('s', ['sid', 'uid', 'nid', 'run_nid', 'timestamp']);
   $query->fields('rb', ['rbid']);
-  $query->orderBy('rb.rbid', 'DESC');
 
   if (isset($params['sid'])) {
     if (is_array($params['sid'])) {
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
```diff
@@ -41,10 +41,10 @@ function dosomething_signup_get_signups_query($params) {
  */
 function dosomething_signup_build_signups_query($params = []) {
   $query = db_select('dosomething_signup', 's');
-  $query->leftJoin('dosomething_reportback', 'rb', 's.nid = rb.nid AND s.run_nid = rb.run_nid');
+  $query->leftJoin('dosomething_reportback', 'rb', 's.uid = rb.uid and s.nid = rb.nid'); 
   $query->fields('s', ['sid', 'uid', 'nid', 'run_nid', 'timestamp']);
   $query->fields('rb', ['rbid']);
-  $query->where('s.uid = rb.uid');
+  $query->orderBy('rb.rbid', 'DESC');
```
angaither: I don't think we actually need the orderBy in this query, that was just for when we were testing

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -91,10 +91,9 @@ private function build($data) {
     $this->campaign = Campaign::get($data->nid);
     $this->campaign_run = $data->run_nid;
 
-    try {
+    if ($data->rbid) {
```
angaither: this should probably be `if(isset($data->rbid && is_numeric($data->rbid))` to avoid warnings

lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
chloealee: For some reason, when I get rid of this orberBy in the query, it breaks the index at `/signups`. Shouldn't the below code fix that since we replaced it from the try/catch?

```
    if (isset($data->rbid) && is_numeric($data->rbid)) {
      $this->reportback = Reportback::get($data->rbid);
    } else {
      $this->reportback = null;
    }
```

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -91,10 +91,9 @@ private function build($data) {
     $this->campaign = Campaign::get($data->nid);
```
chloealee, line 91: It seems like this line is now breaking for the index of /signups?

lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
angaither: @chloealee how does it break?

chloealee: Index of /signups returns 

```
{
  data: [ ]
}
```

angaither: I think it's because the first "campaign" it tries to get is actually a fact.

chloealee: ah I see, thanks! Updated :)

lib/modules/dosomething/dosomething_signup/includes/Signup.php
angaither, line 91: yeah this is the line that breaks the results, this Campaign::get might need to check that the node exists and is of type campaign. cause we probably shouldn't be allowing signups on other node types, but apparently it has happened.

angaither:
:+1:

PR #6194, opened 2016-02-19
Forces a default run nid to be inserted into the filter

### LINKED ISSUE #6192: Reviewed reportbacks tab broke
from @agardner in slack

> It looks like in every campaign reviewed box, every single reviewed report back for every single campaign is there. So Yo Teach reviewed report backs are appearing in Get Lucky’s reviewed box.

probably due to changes in #6172

itsjoekent:
#### What's this PR do?

Makes sure that the run nid on reportback view tab has a default value corresponding to the campaigns current default.

Go to a reportback review tab, is the correct thing filled out and does it show the correct reportbacks?

I realized i didn't notice this the first time because on my machine the first ~10 reportbacks we're correct, but after that they were not. Whereas on Thor/Prod shit is just completely wrong and out of order.

Fixes #6192

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -492,6 +492,17 @@ function _dosomething_reportback_get_permalink_copy_vars($user) {
  * Alters the signup form that is attached to permalink pages.
  */
 function dosomething_reportback_form_alter(&$form, &$form_state, $form_id) {
+
+  // Make sure a default run is added to the filter
+  if ($form['#id'] == 'views-exposed-form-reportback-files-page-1') {
+    if (empty($_GET['run_nid'])) {
+      $run = dosomething_helpers_get_current_campaign_run_for_user(arg(1));
+      if (isset($run)) {
+        $form_state['input']['run_nid'] = $run->nid;
+      }
+    }
+  }
+
   // A signup/login/register form, on a reportback page.
   $forms = array('dosomething_signup_form', 'user_register_form', 'user_login', 'user_login_block');
   if (in_array($form_id, $forms) && drupal_match_path($path = current_path(), 'reportback/*')) {
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -494,7 +494,7 @@ function dosomething_reportback_form_views_exposed_form_alter(&$form, &$form_sta
   // Make sure a default run is added to the filter
   if (empty($_GET['run_nid'])) {
     $run = dosomething_helpers_get_current_campaign_run_for_user(arg(1));
-    if (isset($run)) {
+    if (!empty($run)) {
       $form_state['input']['run_nid'] = $run->nid;
     }
   }
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -485,15 +485,19 @@ function _dosomething_reportback_get_permalink_copy_vars($user) {
   );
 }
 
-
+/**
+ * Implements hook_form_FORM_ID_alter().
+ *
+ * Alters the reportback review form.
+ */
 function dosomething_reportback_form_views_exposed_form_alter(&$form, &$form_state, $form_id) {
   // Make sure a default run is added to the filter
-    if (empty($_GET['run_nid'])) {
-      $run = dosomething_helpers_get_current_campaign_run_for_user(arg(1));
-      if (isset($run)) {
-        $form_state['input']['run_nid'] = $run->nid;
-      }
+  if (empty($_GET['run_nid'])) {
+    $run = dosomething_helpers_get_current_campaign_run_for_user(arg(1));
+    if (isset($run)) {
+      $form_state['input']['run_nid'] = $run->nid;
     }
+  }
 }
 
 /**
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -486,23 +486,22 @@ function _dosomething_reportback_get_permalink_copy_vars($user) {
 }
 
 
-/**
- * Implements hook_form_alter().
- *
- * Alters the signup form that is attached to permalink pages.
- */
-function dosomething_reportback_form_alter(&$form, &$form_state, $form_id) {
-
+function dosomething_reportback_form_views_exposed_form_alter(&$form, &$form_state, $form_id) {
   // Make sure a default run is added to the filter
-  if ($form['#id'] == 'views-exposed-form-reportback-files-page-1') {
     if (empty($_GET['run_nid'])) {
       $run = dosomething_helpers_get_current_campaign_run_for_user(arg(1));
       if (isset($run)) {
         $form_state['input']['run_nid'] = $run->nid;
       }
     }
-  }
+}
 
+/**
+ * Implements hook_form_alter().
+ *
+ * Alters the signup form that is attached to permalink pages.
+ */
+function dosomething_reportback_form_alter(&$form, &$form_state, $form_id) {
   // A signup/login/register form, on a reportback page.
   $forms = array('dosomething_signup_form', 'user_register_form', 'user_login', 'user_login_block');
   if (in_array($form_id, $forms) && drupal_match_path($path = current_path(), 'reportback/*')) {


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -492,6 +492,17 @@ function _dosomething_reportback_get_permalink_copy_vars($user) {
  * Alters the signup form that is attached to permalink pages.
  */
 function dosomething_reportback_form_alter(&$form, &$form_state, $form_id) {
+
+  // Make sure a default run is added to the filter
+  if ($form['#id'] == 'views-exposed-form-reportback-files-page-1') {
+    if (empty($_GET['run_nid'])) {
+      $run = dosomething_helpers_get_current_campaign_run_for_user(arg(1));
+      if (isset($run)) {
```
sergiitk: Maybe `!empty()` will be better?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -492,6 +492,17 @@ function _dosomething_reportback_get_permalink_copy_vars($user) {
  * Alters the signup form that is attached to permalink pages.
  */
 function dosomething_reportback_form_alter(&$form, &$form_state, $form_id) {
+
+  // Make sure a default run is added to the filter
+  if ($form['#id'] == 'views-exposed-form-reportback-files-page-1') {
```
sergiitk: Please use [hook_form_FORM_ID_alter](https://api.drupal.org/api/drupal/modules%21system%21system.api.php/function/hook_form_FORM_ID_alter/7) instead.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
itsjoekent: No I only want this to run when the filter is empty. If I did !empty( it would run even when a user specifies a filter

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
itsjoekent: ok

sergiitk:
Also, instead of altering a form, would it make sense to use [one of view's hooks](https://api.drupal.org/api/views/views.api.php/group/views_hooks/7), like [hook_views_pre_build](https://api.drupal.org/api/views/views.api.php/function/hook_views_pre_build/7) or something?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
sergiitk: Sorry?

itsjoekent: This PR is setting the value for this filter 

screen shot 2016-02-19 at 3 26 59 pm

However that should only happen when the filter is empty. If a value already exists in the filter, I don't want to replace it.

itsjoekent:
I've never gotten those view build hooks to work right, I find the form_alter to be much better

sergiitk:
Ok!

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
sergiitk: Yes, but I'm talking about asserting `$run` variable's existence.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
itsjoekent: I made a new function titled `dosomething_reportback_form_exposed_form_reportback_files_page_1_alter` and did a drush cc all, but it doesnt seem to be running

sergiitk: Missed `views_`?

itsjoekent: ah whoops

sergiitk:
My only concern is `page-1` form id postfix. Would it be a case when this changes to `page-2`?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
sergiitk:
``` php
if (!empty($run)) {
```

itsjoekent:
That was my first thought too when I noticed that, but no its always page-1. I have a feeling thats an old typo someone made haha

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
itsjoekent: `dosomething_reportback_form_views_exposed_form_reportback_files_page_1_alter` still not workin :angry: :drupal:

sergiitk: Hm. `_page_1` postfix weirds me out.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
itsjoekent: OHHHHHHHHHH

the if right above is empty and i thought thats what you were referring too. 

The function im getting the run from returns NULL if it cant find it, so I dont think empty would work

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
itsjoekent: if you dpm the form name and change the page you'll see it never changes. i know its weird but it seems fine

angaither: you can set the form name in the view to be something better

sergiitk: So I dug up this a little and found out that actual form to alter is `views_exposed_form`.
Then you need to check

``` php
$view = $form_state['view'];
if ($view->name == 'my_view_machine_name' && $view->current_display == 'display_name') {
  // alter your exposed form here
}
```

`$view->name` is your view
`$view->current_display` is this mysterious page 1. You can omit this check.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
sergiitk: But `!empty(NULL)` is FALSE, as expected.

sergiitk: And to be clear, `dosomething_helpers_get_current_campaign_run_for_user()` may return not just null, but FALSE and stdClass, see `return node_load($run['id']);`.

sergiitk: and FYI

``` php
php > $false = FALSE;
php > var_dump(isset($false));
bool(true)
```

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
sergiitk: Check this out: http://drupal.stackexchange.com/a/79804

sergiitk:
26 comments - this escalated quickly.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
itsjoekent: ah i see

itsjoekent: yusss `dosomething_reportback_form_views_exposed_form_alter` works!

itsjoekent:
@sergii-tkachenko ok, i think we're good now :sleepy:

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -486,6 +486,16 @@ function _dosomething_reportback_get_permalink_copy_vars($user) {
 }
 
 
+function dosomething_reportback_form_views_exposed_form_alter(&$form, &$form_state, $form_id) {
```
sergiitk, line 493: You forgot docblock :wink:

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -486,6 +486,16 @@ function _dosomething_reportback_get_permalink_copy_vars($user) {
...
+  // Make sure a default run is added to the filter
+    if (empty($_GET['run_nid'])) {
```
sergiitk: Indentation looks wrong.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -486,6 +486,16 @@ function _dosomething_reportback_get_permalink_copy_vars($user) {
...
+      $run = dosomething_helpers_get_current_campaign_run_for_user(arg(1));
+      if (isset($run)) {
```
sergiitk: `!empty($run)`?

sergiitk: See my replies here: https://github.com/DoSomething/phoenix/pull/6194#discussion_r53516553

itsjoekent: github why you hide all the important comments

sergiitk:
Who would think this PR turned out to be so large. High five, @deadlybutter!
:+1:

itsjoekent:
hall of fame much

PR #6195, opened 2016-02-19
get the user address form to show up

### LINKED ISSUE #6158: Address fields no longer show up when collecting user address on campaign
Address fields no longer show up when you configure a campaign to collect user address via the Custom Settings on the campaign.

Example: https://thor.dosomething.org/us/node/1508/custom-settings

image

image

katiecrane:
#### What's this PR do?

There was an if statement that was unsetting all the form elements if there were fewer than 10 of them. However, we only want that to unset if only the default elements are there, and there are 2 of those.
1. Go to a campaign and go to custom settings.
2. Scroll down to "Signup Data Form"
3. Check "Collect additional signup data"
4. Expand "Configuration"
5. Expand "Form Elements"
6. Check "Collect User address"
7. Fill in the necessary fields
8. Login as a non-admin, sign up for the campaign and click the link under "Stuff you need" and the form should appear

When I was submitting my address, it was saying it was invalid. I'm not sure if it just doesn't work on local or if that has to do with the UPS api.

Fixes #6158

cc: @mikefantini, @DFurnes (also shoutout to him for help with drupal spelunking)

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
@@ -454,7 +454,7 @@ function dosomething_signup_user_signup_data_form($form, &$form_state, $signup) 
 
   // @todo Hacky solution to deal with form elements when no elements added and eliminate empty container added.
   // @see https://github.com/DoSomething/phoenix/pull/5809
-  if (count($form['elements']) <= 10) {
+  if (count($form['elements']) <= 2) {
     unset($form['elements']);
   }
 


mikefantini:
Hmmm @katiecrane did you use your NY address?

Did it just simply say it was invalid or did it suggest a way to fix it?

katiecrane:
@mikefantini no I used my parents' address in GA. It said it was invalid and did not suggest a way to fix it.

mikefantini:
Yeah @katiecrane so on production all addresses work, but for testing purposes only NY and CA addresses work.

katiecrane:
@mikefantini the more you know. I just tried with an NY address and it worked! Suggested fixes and then accepted the address.

angaither:
@katiecrane did you check the issue that this fix was added for? to make sure that didn't break?
https://github.com/DoSomething/phoenix/issues/5788

cc @weerd

katiecrane:
@angaither It looks like this now which is less extra space than in #5788, but I'm not sure if it looks the same as @weerd 's fix (but I think it should?)

image

weerd:
I'm not 100% sure why `10` was the magic number I used...? But if it's tested and works fine then seems good to me :) Any chance this effects other modals w/ form elements? Maybe that's why `10` was the magic number?

katiecrane:
@weerd it's only in the `dosomething_signup_user_signup_data_form` function so I'd think it would only mess with that modal? It's possible that the magic number is larger than 2 I guess, but it's also definitely less than 10 if we want the address form to work

weerd:
okie doke. sounds cool :dancer:

angaither:
:+1:

Issue #4526, opened 2015-05-21
Implement external authentication with Northstar


aaronschachter:
In looking to build the ["Change Email" screen in the iOS app](https://github.com/DoSomething/LetsDoThis-iOS/issues/21), I've run into a pickle.  The PUT to the Northstar User works fine and changes the email.  The Drupal User is not updated.

The opposite situation is true as well, if I update my email within Drupal, my Northstar email is not updated.

We'll run into this issue with Mobile Number and Password as well.

Not sure what best course of action is, but I believe this is functionality that @sergii-tkachenko has gotten working for the DS Canada and DS UK sites, by building the [External Auth module](https://github.com/DoSomething/drupal-external-auth), which we've talked about releasing as a contrib module on drupal.org. The module queries a third party service in order to authenticate, and sync the Drupal User with the authenticated user returned in the response.

One approach is that we implement the module for connecting to Northstar, to avoid building out custom Services calls (Northstar doing a PUT on a Drupal User, and vice versa) that would seem to duplicate the work @sergii-tkachenko has already gotten in place.

DS Canada implementation: https://github.com/DoSomething/dosomething/tree/dev/lib/modules/dosomething/dosomething_canada/includes

DS UK implementation: 
https://github.com/DoSomething/dosomething/tree/dev/lib/modules/dosomething/dosomething_uk/includes

cc @jonuy @angaither @DFurnes

jonuy:
A solution Andrea was working on before was to forward Drupal logins, registrations, and updates onto Northstar. There's a Northstar module (https://github.com/DoSomething/dosomething/pull/4434) that seems to just be waiting on Guzzle to be available. What are people's thoughts on this? Can this be a viable nearer-term solution?

Forwarding user updates from Northstar onto Drupal seems trivial and is something we're doing currently for user registrations.

Anyone know what amount of work is required to use the external auth module as a solution? What kind of dev is needed? I assume we'd also need to migrate all users from Drupal over to Northstar too before the flip?

aaronschachter:
@sergii-tkachenko -- you know the External Auth module best, can you weigh in?  

@jonuy Forwarding updates from Northstar to Drupal would still require dev work either on the Northstar end or Drupal end to transform with the various User field arrays.

aaronschachter:
It sounds like we may be able to get by in Version 1 of the mobile app without syncing shared fields such as Email, Password, and Mobile.. so, it may be worth investing in the best solution vs the quickest here.

[...]

Issue #2753, opened 2014-07-02
User username not always saved as UID


aaronschachter:
Inconsistent usernames -- not only for legacy imported users, but for new users which are  as well, sometimes they still have a random string set as their username (see #2632).

Makes it difficult to select users within the Winners field collection.

angaither:
For some reason the `db_update` in `dosomething_user_user_insert` isn't always running, maybe we can add logging in there to see what's going on?

mikefantini:
@aaronschachter @angaither is this issue still valid?

angaither:
Yeah, this for sure still happens. It's very inconsistent

namimody:
Hey @angaither could you pop a t-shirt label on this one? Thanks!

angaither:
this happens on profile save now, it's getting overwritten each time.

PR #6196, opened 2016-02-19
Verify credentials against northstar

DFurnes:
#### What's this PR do?

This PR adds "mock" authentication to Northstar after Drupal successfully authenticates a user locally.

On your local box, pull down this code and run the ol' `drush fra -y`. Log in to your account, and you _should_ be able to see your Northstar ID saved to your Drupal profile (only visible from the admin profile interface).

This is a cautious step towards _actually_ authenticating all logins against Northstar. Let's just let it run in the background for a while and hopefully nothing goes wrong!

References #4526.

---

For review: @angaither


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -459,6 +459,40 @@ function dosomething_user_user_pass_submit($form, &$form_state) {
...
+ */
+function dosomething_user_user_login(&$form_state, $account) {
+  if (!module_exists('dosomething_northstar')) return;
+
+  $northstar_user = null;
+
+  // Figure out whether user is logging in via email or mobile. Then, we'll
+  // send a request to Northstar's `user/verify` endpoint to see if Northstar
+  // is also able to successfully authenticate with the given credentials.
+  if ($account = dosomething_user_get_user_by_email($form_state['values']['name'])) {
```
angaither, line 474: I think this should be `$form_state['values']['mail']`
name is user name which is never used and only sometimes set to emails (see #2753)

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -459,6 +459,40 @@ function dosomething_user_user_pass_submit($form, &$form_state) {
...
+    $northstar_user = dosomething_northstar_verify_user([
+      'email' => $form_state['values']['name'],
+      'password' => $form_state['values']['pass'],
+    ]);
+
+  }
+  elseif ($number = $form_state['values']['name']) {
+    if ($account = dosomething_user_get_user_by_mobile($number)) {
```
angaither, line 482: also wait, why do you need to be getting the user, it should be in the param in this hook?

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -43,6 +43,30 @@ function dosomething_northstar_libraries_info() {
...
+  $client = _dosomething_northstar_build_http_client();
+
+  $response = drupal_http_request($client['base_url'] . '/auth/verify', [
+    'headers' => $client['headers'],
+    'method' => 'POST',
+    'data' => json_encode($credentials),
+  ]);
+
+  if($response->code !== '200') {
+    return null;
```
angaither, line 64: maybe we can watchdog some things here to see why it failed?

lib/modules/dosomething/dosomething_user/dosomething_user.module
DFurnes, line 482: The `$account` parameter given in the hook only includes the hashed password, and doesn't specify which "username" credential the user tried to log in with.

lib/modules/dosomething/dosomething_user/dosomething_user.module
DFurnes, line 474: The user login form doesn't differentiate between the two since they're submitted from the same form field.

DFurnes, line 474: I based this logic off [`dosomething_user_login_validate`](https://github.com/DFurnes/phoenix/blob/627a5b5e8bae0760a7ec19fde1fa228e882671da/lib/modules/dosomething/dosomething_user/dosomething_user.module#L907-L919) but let me know if there's a smarter way of handling it!

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
DFurnes, line 64: The only time this _should_ happen is if the given credentials doesn't exist in Northstar, which will return a `401`. It might be worth explicitly checking for the `401` to return `NULL` and then triggering a watchdog if any other code is returned?

angaither:
:+1:

PR #6198, opened 2016-02-22
Northstar id field fixes

### LINKED ISSUE #6197: Integrity constraint violation after merging #6196
#### Current Behavior

Things explode if you log in using an account where the username is neither email nor password.

screen shot 2016-02-22 at 11 32 15 am
#### Desired Behavior

Things should not explode.
#### Steps to Replicate

Log in to an account where the username is one of those randomly generated integers.
#### Why This Matters

Things explode.

DFurnes:
#### What's this PR do?

Fixes some issues I noticed after merging #6196:
- Removes "Northstar ID" from user profile pages, and makes it read-only for admins.
- Fixes issue where everything would explode if you tried to log in with an account that had a username that wasn't it's email address or phone number (because it would overwrite `$account` to be false). So instead…
  - Don't overwrite `$account` since we already have it as an argument from the hook
  - Use `$form_state['input']` to get raw form input rather than the incorrect `$form_state['values']` I had been using before.

If you pulled down #6196, you'd probably get a big ol' error if you tried to log in with an account that had one of those programmatic numeric usernames. Now you won't.

Fixes #6197.

---

For review: @angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -471,17 +471,17 @@ function dosomething_user_user_login(&$form_state, $account) {
   // Figure out whether user is logging in via email or mobile. Then, we'll
   // send a request to Northstar's `user/verify` endpoint to see if Northstar
   // is also able to successfully authenticate with the given credentials.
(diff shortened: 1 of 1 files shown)


angaither:
:+1:

PR #6200, opened 2016-02-22
catches errors and returns null for campaigns and reportbacks in orde…

### LINKED ISSUE #6193: /signups index doesn't return data - breaking at an error message
#### Current Behavior

`/signups` index returns no data - breaking at `no reportback data found` and `no campaign data found` error messages. 

`https://thor.dosomething.org/api/v1/signups/23` returns 

```
{
  error: { 
    message: "No reportback data found."
  }
}
```

and `https://thor.dosomething.org/api/v1/signups/1` returns 

```
{
  error: {
    message: "No campaign data found."
  }
}
```
#### Desired Behavior

If the signup is not for a campaign, should not be included in the response. 
If there is no reportback, the reportback object should return `null` instead of the error message.

chloealee:
What's this PR do?

Catches errors and returns null for campaigns and reportbacks in order to reach all /signup endpoints.

For some reason, users can sign up for some nodes that are not campaigns or sms_games which was breaking when we tried to build the campaign in Signup.php's get function.

We were hitting the No reportback found. error, again breaking the endpoint, in the situation where there was a reportback in the database but couldn't be returned if has not yet been reviewed.

How should this be manually tested?

Make sure all /signup endpoints work and don't log and errors in the Drupal log messages.

http://dev.dosomething.org:8888/api/v1/signups
http://dev.dosomething.org:8888/api/v1/signups/42
http://dev.dosomething.org:8888/api/v1/signups?user=98&campaigns=1227,1173

What are the relevant tickets?

Fixes #6193

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/includes/Signup.php
+++ b/lib/modules/dosomething/dosomething_signup/includes/Signup.php
@@ -97,13 +97,17 @@ class Signup extends Entity {
 
     $this->campaign_run = $data->run_nid;
 
+    // Only send to Reportback::get if there is a rbid. 
     if (isset($data->rbid) && is_numeric($data->rbid)) {
+      // Catch error if a reportback is not found.
+      // Reportback will not be returned if it hasn't yet been reviewed. 
       try {
         $this->reportback = Reportback::get($data->rbid);
       } catch (Exception $e) {
         $this->reportback = null;
       }
-    } else {
+    } 
+    else {
       $this->reportback = null;
     }
   }
--- a/lib/modules/dosomething/dosomething_signup/includes/Signup.php
+++ b/lib/modules/dosomething/dosomething_signup/includes/Signup.php
@@ -97,16 +97,16 @@ class Signup extends Entity {
 
     $this->campaign_run = $data->run_nid;
 
-    // Only send to Reportback::get if there is a rbid. 
+    // Only send to Reportback::get if there is a rbid.
     if (isset($data->rbid) && is_numeric($data->rbid)) {
       // Catch error if a reportback is not found.
-      // Reportback will not be returned if it hasn't yet been reviewed. 
+      // Reportback will not be returned if it hasn't yet been reviewed.
       try {
         $this->reportback = Reportback::get($data->rbid);
       } catch (Exception $e) {
         $this->reportback = null;
       }
-    } 
+    }
     else {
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_signup/includes/SignupTransformer.php
```diff
@@ -64,11 +64,15 @@ protected function transform($item) {
...
 
-    $campaign = (object) $item->campaign;
-    $current_run = $campaign->campaign_runs['current']['en']['id'];
-    $current = ($item->campaign_run == $current_run);
-    $data += $this->transformSignup($item, $current);
-    $data['campaign'] = $this->transformCampaign((object) $item->campaign);
+    if (is_null($item->campaign)) {
+      $data['campaign'] = null;
+    } else {
+      $campaign = (object) $item->campaign;
```
angaither, line 71: new line `else`

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -88,11 +88,21 @@ public static function find(array $filters = []) {
...
     $this->campaign_run = $data->run_nid;
 
     if (isset($data->rbid) && is_numeric($data->rbid)) {
-      $this->reportback = Reportback::get($data->rbid);
+      try {
+        $this->reportback = Reportback::get($data->rbid);
+      } catch (Exception $e) {
+        $this->reportback = null;
+      }
     } else {
```
angaither: same

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -88,11 +88,21 @@ public static function find(array $filters = []) {
...
+    catch (Exception $error) {
+      $this->campaign = null;
+    }
+
     $this->campaign_run = $data->run_nid;
 
     if (isset($data->rbid) && is_numeric($data->rbid)) {
-      $this->reportback = Reportback::get($data->rbid);
+      try {
+        $this->reportback = Reportback::get($data->rbid);
```
angaither, line 105: maybe add some comments from the convo to say why this is needed

chloealee, line 105: smart!

angaither:
:+1:

Issue #6170, opened 2016-02-16
forward users to gladiator


angaither:
after #6169 we need to send an HTTP request to send the user who opted into the competitions #4616
this might be a chance to get guzzle working in this repo
#4594

angaither:
@sbsmith86 i just thought of something, for now you can test this with forwarding to gladiator from local to local, then switch out the url's for each env

sbsmith86:
@angaither yea, definitely wouldn't send people to prod from my dev. But I'm not sure what endpoint im supposed to be hitting. I think @weerd was gonna set something up?

angaither:
@sbsmith86 you can use `gladiator.app:8000`!

angaither:
oh, but to the actual ENDPOINT. doi

angaither:
refs #6246
but needs more error protection
as well as potential api updates DoSomething/gladiator/issues/54

mshmsh5000:
This endpoint (`/api/v1/users/`) is crapping out with a 500 on QA with a POST payload, but only when you add the `Accept: application/json` header.

Issue #4616, opened 2015-06-24
Guzzle fails locally


angaither:
`GuzzleHttp\Exception\RequestException: cURL error 3: <url> malformed (see http://curl.haxx.se/libcurl/c/libcurl-errors.html) in GuzzleHttp\Handler\CurlFactory::createRejection() (line 162 of /var/www/dev.dosomething.org/html/profiles/dosomething/libraries/guzzle/src/Handler/CurlFactory.php).
The website encountered an unexpected error. Please try again later.`

mikefantini:
yeah this is happening on staging too fwiw

sbsmith86:
@angaither Was working on #6170 and it looks like this error was happening because we were using 

`'base_url' => array($base_url . '/{version}/', array('version' => $version))` 

but it should be `base_uri` and the value has to be a string, but it looks like it was trying to assign it to an array to do the string replacement. 

Reference: http://docs.guzzlephp.org/en/latest/quickstart.html#creating-a-client

Anyway, guzzle seems to work for me locally when I create the client correctly. So it seems like we just need to fix that stuff up in the Northstar module.

angaither:
@sbsmith86 ahh, awesome

PR #6201, opened 2016-02-22
Forward to gladiator

sbsmith86:
#### What's this PR do?

This PR starts the work of sending users to [gladiator](https://github.com/DoSomething/gladiator) when they sign up for a competition using the sign up data form. 

I didn't have the endpoint that we are actually going to be using in gladiator, so this uses a fake endpoint, but flushes out the basic logic that we will be using to do this work. 

This is isn't fully functional yet, but if you pull down this branch and then run

`drush eval 'dosomething_signup_send_user_to_gladiator('user', 'values')'` you shouldn't see any errors pop back. Also, you should be able to submit a sign up data form, that is configured to be a competition sign up, without any issues.

There was a previous issue with Guzzle working in Phoenix ( #4616 ) But I believe that was an issue because the guzzle client was being configured incorrectly. See that issue for deets.

Addresses #6170 

cc @weerd

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -65,6 +65,19 @@ function dosomething_signup_menu() {
 }
 
 /**
+ * Implements hook_libraries_info().
+ */
+function dosomething_signup_libraries_info() {
+  $libraries['guzzle'] = array(
+    'name' => 'guzzle',
+    'path' => 'vendor',
+    'files' => ['php' => ['autoload.php']],
+    'version' => 5.0,
+  );
+  return $libraries;
+}
+
+/**
  * Implements hook_entity_info().
  */
 function dosomething_signup_entity_info() {
@@ -1025,4 +1038,22 @@ function dosomething_signup_get_target_nid_for_source($source_string) {
     }
   }
   return NULL;
+}
+
+function _dosomething_signup_build_http_client() {
+  $base_url = 'http://jsonplaceholder.typicode.com';
+
+  if (libraries_load('guzzle') == TRUE) {
+    $client = new GuzzleHttp\Client(array(
+      'base_uri' => 'http://jsonplaceholder.typicode.com',
+      'defaults' => array(
+        'headers' => array(
+          'Content-Type' => 'application/json',
+          'Accept' => 'application/json'
+          ),
+        ),
+    ));
+    $res = $client->request('GET', '/posts/1');
+    echo $res->getBody();
+  }
 }
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -1040,6 +1040,12 @@ function dosomething_signup_get_target_nid_for_source($source_string) {
   return NULL;
 }
 
+/**
+ * Build the Guzzle HTTP Client to make requests to Gladiator.
+ *
+ * https://github.com/DoSomething/gladiator
+ * @TODO - use correct URL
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -1026,3 +1039,29 @@ function dosomething_signup_get_target_nid_for_source($source_string) {
...
+  if (libraries_load('guzzle') == TRUE) {
+    $client = new GuzzleHttp\Client(array(
+      'base_uri' => 'http://jsonplaceholder.typicode.com',
+      'defaults' => array(
+        'headers' => array(
+          'Content-Type' => 'application/json',
+          'Accept' => 'application/json'
+          ),
+        ),
+    ));
```
angaither: some errors if this doesn't work. maybe we need a queue? v2 stuff ;)

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -507,6 +507,10 @@ function dosomething_signup_user_signup_data_form_submit($form, &$form_state) {
   }
   // Update signup record.
   dosomething_signup_update_signup_data($values, $config);
+  // If this is a competition sign up, send it to gladiator.
+  if ($config['competition_signup']) {
+    dosomething_signup_send_user_to_gladiator($user, $values);
```
angaither, line 512: I think we will probably want some other feature flag here, but again, that can come later

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -612,3 +616,33 @@ function dosomething_signup_filter_form_submitted_copy($string, $timestamp) {
...
+ * @param object $user
+ *   The user that signup up for the competition.
+ * @param array $values
+ *   Values from the signup data form.
+ */
+function dosomething_signup_send_user_to_gladiator($user, $values) {
+  // Make sure the client is set up.
+  if (!_dosomething_signup_build_http_client()) { return; }
+
+  $client = _dosomething_signup_build_http_client();
```
angaither, line 629: can probably do the if ($client) {return;} so you don't need to call `_dosomething_signup_build_http_client` twice

sbsmith86, line 629: ack tru tru

angaither:
after [that](https://github.com/DoSomething/phoenix/pull/6201#discussion_r53677671) :+1:

weerd:
Awesome! :+1:

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -1026,3 +1039,29 @@ function dosomething_signup_get_target_nid_for_source($source_string) {
...
+
+    return $client;
+  }
+
+  return;
```
weerd, line 1066: maybe return `null` or `false` here, so when you call the function and then check if it work in **dosomething_signup.signup_data_form.inc** you have something to go off of?

sbsmith86, line 1066: I believe, when you don't specify a parameter, php returns NULL by default. 

> Note: If no parameter is supplied, then the parentheses must be omitted and NULL will be returned. Calling return with parentheses but with no arguments will result in a parse error.

http://php.net/manual/en/function.return.php

weerd, line 1066: gotcha, yeah.

PR #6207, opened 2016-02-23
Flagged bug fix

### LINKED ISSUE #6206: Expose flagged property in Reportback response
#### Current Behavior

Prior to filtering reportbacks in the response for the `/reportbacks` endpoint, we would only show certain properties `why_participated` and `flagged` if the requester had access to view those items. However, now that we do some filtering ahead of time and don't show RBs that should only be visible to admin users, we should be able to expose the `flagged` property. 
#### Desired Behavior

Show the value for the `flagged` property in the reportbacks endpoint response.
#### Why This Matters

This will also fix a slight bug with how the `flagged` property would be expected but not always available if it was an anonymous request.

weerd:
#### What's this PR do?

Made a quick fix to a tiny bug I noticed on some recent code I updated. If the endpoint was being accessed by an anonymous user than the flagged property does not exist on the returned Reportback. It should be fine to expose this property especially now that we have some filtering in place remove flagged reportbacks from being accessible by anon users.

Fixes #6206

---

@angaither @chloealee

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
@@ -68,7 +68,9 @@ function dosomething_reportback_build_reportbacks_query($params = array()) {
   $rb_fields = array('rbid', 'created', 'updated', 'quantity', 'uid', 'flagged');
 
   // Staff-only properties to add:
-  $rb_fields[] = 'why_participated';
+  if (user_access('view any reportback')) {
+    $rb_fields[] = 'why_participated';
+  }
 
   $query->fields('rb', $rb_fields);
   $query->fields('n', ['nid', 'title']);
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
```diff
@@ -65,13 +65,10 @@ function dosomething_reportback_build_reportbacks_query($params = array()) {
...
   // Public API properties to expose:
-  $rb_fields = array('rbid', 'created', 'updated', 'quantity', 'uid');
+  $rb_fields = array('rbid', 'created', 'updated', 'quantity', 'uid', 'flagged');
 
   // Staff-only properties to add:
-  if (user_access('view any reportback')) {
-    $rb_fields[] = 'why_participated';
-    $rb_fields[] = 'flagged';
-  }
+  $rb_fields[] = 'why_participated';
```
angaither: if you moved `flagged` up to the array why doesn't `why_participated` also get moved up?

weerd: I did initially and then decided against it...

weerd: Is it fine to expose what might be a sensitive description for the member user to anon users?

angaither:
:+1: makes more sense now :rocket:

PR #6208, opened 2016-02-23
Expose competition in campaign api

### LINKED ISSUE #6205: Expose competition in campaign api
To support DoSomething/gladiator/issues/14 the api needs to expose if the campaign has a competition on it.

chloealee:
#### What's this PR do?

Exposes competition in campaign object with boolean value or null if not set. 

Hit any `/campaigns` endpoint and campaigns object should return a line `competitions` under `staff_pick`. 

```
{
  data: {
    id: "1736",
  title: "Test Title",
  campaign_runs: {
    current: {
      en: {
        id: "1902"
      }
    },
    past: [ ]
  },
  language: {
    language_code: "en",
    prefix: "us"
  },
  translations: {
    original: "en",
    en: {
      language_code: "en",
      prefix: "us"
    }
  },
  tagline: "Test CTA",
  status: "active",
  type: "campaign",
  created_at: "1445282953",
  updated_at: "1456252247",
  time_commitment: 0,
  cover_image: {
    default: null,
    alternate: null
  },
  staff_pick: false,
  competition: true,
...
```

Fixes #6205

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -219,6 +219,7 @@ abstract class Transformer {
         }
 
         $output['staff_pick'] = $data->staff_pick;
+        $output['competition'] = $data->competition;
 
         $output['facts']['problem'] = $data->facts['problem'] ? $data->facts['problem']['fact'] : NULL;
         $output['facts']['solution'] = $data->facts['solution'] ? $data->facts['solution']['fact'] : NULL;
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -220,13 +220,6 @@ abstract class Transformer {
         }
 
         $output['staff_pick'] = $data->staff_pick;
-
-        if ($data->competition == '1') {
-          $data->competition = TRUE;
-        }
-        elseif ($data->competition == '0') {
-          $data->competition = FALSE;
-        }
         $output['competition'] = $data->competition;
 
         $output['facts']['problem'] = $data->facts['problem'] ? $data->facts['problem']['fact'] : NULL;
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -174,6 +174,7 @@ abstract class Transformer {
    *   - status: (string)
    *   - cover_image: (array)
    *   - staff_pick: (bool)
+   *   - competition: (bool)
    *   - facts: (array)
    *   - solutions: (array)
    *   - causes: (array)
@@ -219,6 +220,14 @@ abstract class Transformer {
         }
 
         $output['staff_pick'] = $data->staff_pick;
+
+        if ($data->competition == '1') {
+          $data->competition = TRUE;
+        }
+        elseif ($data->competition == '0') {
+          $data->competition = FALSE;
+        }
+        
         $output['competition'] = $data->competition;
 
         $output['facts']['problem'] = $data->facts['problem'] ? $data->facts['problem']['fact'] : NULL;
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -227,7 +227,6 @@ abstract class Transformer {
         elseif ($data->competition == '0') {
           $data->competition = FALSE;
         }
-        
         $output['competition'] = $data->competition;
 
         $output['facts']['problem'] = $data->facts['problem'] ? $data->facts['problem']['fact'] : NULL;
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.query.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.query.inc
@@ -4,7 +4,7 @@
 *
 * @return bool
 */
-protected function getCompetition($id) {
+public function getCompetition($id) {
 	$query = db_select('node', 'n');
 	$query->leftJoin('dosomething_signup_data_form', 'sdf', 'n.nid = sdf.nid');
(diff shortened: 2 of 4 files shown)


lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
```diff
@@ -341,6 +342,30 @@ protected function getCoverImageAlt() {
...
+
+    if ($result == '1') {
+      $result = TRUE;
+    }
+    elseif ($result == '0') {
+      $result = FALSE;
+    }
+
+    return $result;
+  }
```
weerd: For queries that need to be built out, could we move it off into a `dosomething_campaigns.query.inc` file or whatever? similar to how we do with Reportbacks, ReportbackItems, and Signups?

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
```diff
@@ -341,6 +342,30 @@ protected function getCoverImageAlt() {
...
+  protected function getCompetition($id) {
+    $query = db_select('node', 'n');
+    $query->leftJoin('dosomething_signup_data_form', 'sdf', 'n.nid = sdf.nid');
+    $query->fields('sdf', ['competition_signup']);
+    $query->condition('n.nid', $id);
+
+    $result = $query->execute()->fetchAll();
+    $result = $result[0]->competition_signup;
+
+    if ($result == '1') {
```
angaither: instead of switching these from 1 or 0 and returning true/false, you could just return the value and can't the transformer change the value to t/f?

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
chloealee: great call!

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
```diff
@@ -160,6 +160,7 @@ private function build($data, $display = 'teaser') {
 
         $this->scholarship = $this->getScholarship();
         $this->staff_pick = $this->getStaffPickStatus();
+        $this->competition = $this->getCompetition($this->id);
```
weerd: this won't still work since the current Campaign object `$this` no longer has that method. you can call the method to get competitions in the **dosomething_campaign.query.inc`something along the normal Drupal naming scheme, like`dosomething_campaign_is_competion($id)` and then you call this from the object:

```
$this->competition = dosomething_campaign_is_competition($this->id);
```

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.query.inc
```diff
@@ -0,0 +1,17 @@
+<?php
+/**
+* Get status whether this campaign has a competition or not.
+*
+* @return bool
+*/
+protected function getCompetition($id) {
```
weerd: can't be protected here since not in an object. it's just a regular function :)

lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -220,6 +221,14 @@ protected function transformCampaign($data) {
 
         $output['staff_pick'] = $data->staff_pick;
 
+        if ($data->competition == '1') {
+          $data->competition = TRUE;
+        }
+        elseif ($data->competition == '0') {
+          $data->competition = FALSE;
+        }
```
weerd: You can keep this in the build (so it's available as a boolean outside the API, so for the DS app internal stuff) and just run it through this function: 

https://github.com/DoSomething/phoenix/blob/4da588c3ffd383111e8650100dfa4673f11f46a4/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module#L127

chloealee:
thanks for all the edits @weerd ! I'm getting this error `Fatal error: Call to undefined function dosomething_campaign_is_competition() in /var/www/dev.dosomething.org/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php on line <i>164</i></th></tr>` but I think I did it correctly, following `Signup.php` and `dosomething_signup.query.inc` -- not sure why it can't get there?

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
```diff
@@ -161,6 +161,10 @@ private function build($data, $display = 'teaser') {
         $this->scholarship = $this->getScholarship();
         $this->staff_pick = $this->getStaffPickStatus();
 
+        $is_competition = dosomething_campaign_is_competition($this->id);
+        $is_competition = dosomething_helpers_convert_string_to_boolean($is_competition);
+        $this->competition = $is_competition;
```
weerd: you can condense this :)

```
$this->competition =  dosomething_helpers_convert_string_to_boolean(dosomething_campaign_is_competition($this->id));
```

or if it feels crazy long:

```
$is_competition = dosomething_campaign_is_competition($this->id);
$this->competition =  dosomething_helpers_convert_string_to_boolean($is_competition);
```

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.query.inc
```diff
@@ -0,0 +1,23 @@
...
+	$query = db_select('node', 'n');
+	$query->leftJoin('dosomething_signup_data_form', 'sdf', 'n.nid = sdf.nid');
+	$query->fields('sdf', ['competition_signup']);
+	$query->condition('n.nid', $id);
+
+	$result = $query->execute()->fetchAll();
+	$result = $result[0]->competition_signup;
+
+	return $result;
+}
```
weerd, line 23: editorconfig! indent the function body code :)

weerd:
@chloealee you'll need to include **dosomething_signup.query.inc** via the .module file, otherwise the module has no idea that file exists and when you call the `dosomething_campaign_is_competition()` function it has no idea where to look :scream:

chloealee:
@weerd ah doi!! Thanks!

weerd:
Looks sweet! :+1:

PR #6211, opened 2016-02-24
Adds stathat logging to permalink signup

### LINKED ISSUE #6075: Monitoring: Sign Up Via Permalink
#### Experience to Test:

Signing up via permalink.
#### Proxy:

No POST requests occur.
#### System:

StatHat

itsjoekent:
#### What's this PR do?

Adds stat hat logging to permalink signups

If you do a normal signup, does this log?
If you do a permalink signup, does this log?

I couldn't see a better way than using the $source parameter / URL to figure out if this was the perma link so just went with that. 

Fixes #6075

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -243,6 +243,10 @@ function dosomething_signup_create($nid, $uid = NULL, $source = NULL, $timestamp
     return FALSE;
   }
 
+  if (strpos($source, 'reportback') === 0) {
+    stathat_send_ez_count('drupal - Signup - permalink signup', 1);
+  }
+
   $values = array(
     'nid' => $nid,
     'uid' => $uid,
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -243,8 +243,10 @@ function dosomething_signup_create($nid, $uid = NULL, $source = NULL, $timestamp
     return FALSE;
   }
 
-  if (strpos($source, 'reportback') === 0) {
-    stathat_send_ez_count('drupal - Signup - permalink signup', 1);
+  if (module_exists('stathat')) {
+    if (strpos($source, 'reportback') === 0) {
+      stathat_send_ez_count('drupal - Signup - permalink signup', 1);
+    }
   }
 
   $values = array(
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -243,6 +243,10 @@ function dosomething_signup_create($nid, $uid = NULL, $source = NULL, $timestamp
     return FALSE;
   }
 
+  if (strpos($source, 'reportback') === 0) {
```
sergiitk: `if (module_exists('stathat')) {`?

angaither: why not just if `$source === 'reportback'`?

itsjoekent: @sergii-tkachenko ah good catch, we have a few places this needs to be updated then https://github.com/DoSomething/phoenix/blob/a47d9650be54a127adaa5e80f297c474f431fc01/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module#L96

@angaither the entire source param is like reportback/3423/djsfu4w89rd9 just the start of it is reportback

angaither: @joe the line above checks for stathat. https://github.com/DoSomething/phoenix/blob/a47d9650be54a127adaa5e80f297c474f431fc01/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module#L95

got it, i forgot we kept which rb got the person to sign up

itsjoekent: ahh missed that

itsjoekent:
@sergii-tkachenko alrighty should be good to go

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
sergiitk: @angaither @deadlybutter 

> why not just if `$source === 'reportback'`
> To express all sources that start with reportback, ex. `reportback-foobar`?

angaither:
@deadlybutter this should be at the end of the signup_create method, so we know it actually succeeded

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -243,6 +243,13 @@ function dosomething_signup_create($nid, $uid = NULL, $source = NULL, $timestamp
     return FALSE;
   }
 
+  if (module_exists('stathat')) {
+    // The source is often a string longer than just "reportback", this logic accounts for that.
```
sergiitk: > The source is often a string longer than just "reportback"

How longer? Should it better clarify that the source string starts with `reportback`?

itsjoekent:
@angaither moved it right before the return entity->sid

@sergii-tkachenko updated comment

sergiitk:
:+1:

PR #6212, opened 2016-02-24
Updates Drupal core to 7.43.

sergiitk:
Updates Drupal core to 7.43.
This is a security update, see https://www.drupal.org/SA-CORE-2016-001


itsjoekent:
Needs more comments

itsjoekent:
:+1:

PR #6215, opened 2016-02-24
Updates reportback review pages

### LINKED ISSUE #6213: contextual filters missing from rb views
reportback views used to have contextual filters to add which node or tax term they were viewing, and are missing now

refs #6172

image

itsjoekent:
#### What's this PR do?
- Removes form alter that was implemented because a contextual filter was missing
- Adds the contextual filter which got accidentally deleted in both reportback pages

Does the reportback review page show all reportbacks for the campaign?
Does specifying a run only display the correct reportbacks?
Does the taxonomy reportback page show the correct reportbacks?

Somehow these filters got accidentally removed in a previous PR and we were't sure what they were. The form alter here was how we patched it in the meantime, however is this is the proper solution.

Fixes #6213


itsjoekent:
@angaither alrighty it matches the old PR now https://github.com/DoSomething/phoenix/pull/6172/files

angaither:
:+1:

PR #6216, opened 2016-02-25
Address validation

### LINKED ISSUE #6214: address on 'update my profile' not validating
#### Current Behavior

you can update an address to say whatever you want, it doesn't validate
#### Desired Behavior
- address validate like on the signup data form
#### Steps to Replicate
- go to user/XXX/edit and update the address
- put in something that doesn't make sense

no errors on user/edit
image

errors on signup data form
image

This probably has something to do with the fact that `field_address` is treated different than the signup data form [here](https://github.com/DoSomething/phoenix/blob/5896b2823f770c6aa3a1e9cf46d474ef4ae60564/lib/modules/dosomething/dosomething_user/dosomething_user_valid_address.inc#L43)

chloealee:
#### What's this PR do?

Adds a validation check in the user info edit form and updates URL in admin update form to include US country code. 

On your admin account, try to edit your address and put "hi" as the zip code. An error message should pop up and not let you save. 

On a non-admin account, sign up for a campaign that asks for member's address upon signup. Again, try and add "hi" as the zip code to see if it allows you to save. 

Fixes #6214

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -258,7 +258,7 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
 
       // Add extra validation on the edit page, for addresses.
       $current_page = $_SERVER['REQUEST_URI'];
-      if (preg_match('/user\/([0-9]+)\/edit/', $current_page) && $validate_address) {
+      if (preg_match('/us/user\/([0-9]+)\/edit/', $current_page) && $validate_address) {
         $form['#validate'][] = 'dosomething_user_validate_address_field';
       }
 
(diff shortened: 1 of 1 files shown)


angaither:
:+1:

angaither:
actually, the first pattern may need to match with or without `/us/`

chloealee:
thanks, @angaither ! Would that `||` operator do the trick?

angaither:
@chloealee `||` or you could do some fancy regex if you really wanna go crazy

chloealee:
@angaither quick q - shouldn't `preg_match('/user\/([0-9]+)\/edit/', $current_page)` actually work for both cases since `/us` comes before that string and it just looks at the URL to see if `/user/:id/edit` is apart of the current url?

angaither:
@chloealee oh.yup....

chloealee:
:) Okay updated and reverted back to original `preg_match` statement!

angaither:
:+1:

PR #6155, opened 2016-02-05
Added drush php script to migrate archived activity.

angaither:
The updb from #6136 failed, lord knows. 

This script we can run again, and try to migrate the data. 
Fixes #5995

PR #6217, opened 2016-02-25
Added some logging to the run script.

### LINKED ISSUE #5995: Script to un-archive run activity
write a db_update to move all signups and reportbacks from run nodes back to the original campaign node

make sure to save the run id to the sign up and reportback as well

angaither:
#### What's this PR do?

I was running the fix for #5995 on thor and noticed some errors, this addresses that! 

run `drush --script-path=../scripts/ php-script move-signups-reportbacks-from-runs.php signup|reportback`
and watch content that was archived become unarchived

I pushed this code a while ago, but needed a mental break from runs. 
It's a good thing too, because the code I wrote in #6155 was wrong, this fixes that up! 

Fixes #5995 refs #6155

fixed up the query to include rbid, or sid
set the run_nid as well as the nid :facepalm:


DFurnes:
pancakey

PR #6218, opened 2016-02-25
unsets flagged params if admin

### LINKED ISSUE #6176: Allow admin API users to see all reportbacks on index endpoint
References discussion in DoSomething/northstar#279 and "background context" from #6174:

> This only works for the show endpoint and doesn't work for the index collection endpoint. We would need to do some further work to allow unreviewed RBs to show if querying via the /reportbacks index collection endpoint do the the array of params that need to be accounted for (ie: such as the status param).

While updating the "show" endpoint should fix the biggest issue here, it'd be good to get around to updating the `/reportbacks` index so we don't have the weird UX situation of a user submitting a reportback and not seeing it on their profile gallery.

/cc @weerd @aaronschachter

chloealee:
#### What's this PR do?

Allows admin API users to see all reportbacks on index endpoint by unsetting flagged param. Users then hit `/api/v1/reportbacks?status=pending,flagged,excluded,approved,promoted` to view all.

Use above endpoint to make sure all reportbacks are returned. 

Fixes #6176

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
@@ -133,6 +133,10 @@ function dosomething_reportback_get_reportbacks_query($params) {
  */
 function dosomething_reportback_filter_flagged_reportbacks($params) {
   if (user_access('view any reportback')) {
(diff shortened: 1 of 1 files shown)


weerd:
:+1:

PR #6223, opened 2016-02-26
API formatting updates

### LINKED ISSUE #6048: API formatting updates
From a quick offline conversation, thinking of updating some Phoenix reportback formatting to match the changes that were [recently made in Northstar](https://github.com/DoSomething/northstar/pull/253#issuecomment-170603713).
#### Proposed Changes:
- [x] Update dates to be returned as ISO-8601 string rather than UNIX timestamps
- [x] Consider moving `pagination` into a `meta` object, and formatting `links` object for next/prev URL
- [x] Remove `retrieved_count` property from response (since getting the length of an array in PHP and JavaScript are both suuuuuper fast)
- [ ] Consider namespacing filters in query-strings, for example `/api/v1/reportback-items/?filter[status]=approved,promoted&page=2`

/cc @jonuy @aaronschachter

chloealee:
#### What's this PR do?
- Updates dates to be returned as ISO-8601 string rather than UNIX timestamps
- Moves pagination into a meta object, and formats links object for next/prev URL
- Removes `retrieved_count` property from response
- Adds pagination to `/reportbacks` endpoint

Check `/reportbacks` and `/reportback-items` to make sure all of the above is changed. 

Refs #6048. Fixes bullet points 1-3.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -331,13 +331,10 @@ abstract class Transformer {
    * @return array
    */
   protected function transformReportback($data) {
-    $iso_created_at = date('c', $data->created_at);
-    $iso_updated_at = date('c', $data->updated_at);
-
     $output = [
       'id' => $data->id,
-      'created_at' => $iso_created_at,
-      'updated_at' => $iso_updated_at,
+      'created_at' => date('c', $data->created_at),
+      'updated_at' => date('c', $data->updated_at),
       'quantity' => $data->quantity,
     ];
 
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -329,10 +331,13 @@ protected function transformMedia($data, $aspect_ratio = NULL) {
...
   protected function transformReportback($data) {
+    $iso_created_at = date('c', $data->created_at);
+    $iso_updated_at = date('c', $data->updated_at);
+
     $output = [
       'id' => $data->id,
-      'created_at' => $data->created_at,
-      'updated_at' => $data->updated_at,
+      'created_at' => $iso_created_at,
+      'updated_at' => $iso_updated_at,
```
angaither: instead of creating temp vars here,  you could just set it
`'created_at' => date('c', $data->created_at)`

chloealee: ah duh - thanks Andrea!

angaither:
looks ok to me :+1:

PR #6234, opened 2016-03-02
Professional campaigns only please

### LINKED ISSUE #6233: Campaigns API: return closed status if the node is unpublished (or don't return campaign at all)
#### Current Behavior
- An unpublished campaign node can exist with `active` campaign status, resulting in a 500 error upon signup. I think this is possible as hack to allow for "mobile app" campaigns only -- campaigns that exist in the app but not on the web. That's no longer a thing.
#### Desired Behavior
- The campaign.status should be returned as `closed` because the node is unpublished, or don't return a campaign at all
#### Why This Matters

Ran into scenarios on mobile app where I tapped through someone's profile to a unpublished campaign, where its `campaign.status == 'active'`. Tried to signup and got a 500 error back.

screen shot 2016-03-02 at 9 09 50 am

Last column is "Published":
screen shot 2016-03-02 at 9 09 21 am

angaither:
#### What's this PR do?

remove unpublished campaigns from the api index. 

do unpublished campaigns show up in the index?

This fix was added in with #5175 to support the original
idea in the mobile app, where we had campaigns show up on
the app, but not on the web.
As we are no longer going that route, let's only return
published campaigns.

Fixes #6233
#### doesn't fix

> it kinda feels like always setting the campaign.status to closed for an unpublished campaign node will do the trick there 
> because even if we hide the campaign from API, that signup is still returned from `signups?user=:id`

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -464,11 +464,8 @@ function dosomething_campaign_build_campaigns_query($params = array()) {
   $query = db_select('node', 'n');
   $query->join('field_data_field_campaign_type', 't', 't.entity_id = n.nid');
 
-  // @TODO: this is a temporary fix to address the need to allow unpublished mobile_app campaigns to be retrieved, only for the mobile app.
-  if (!$params['mobile_app']) {
-    $query->condition('n.status', 1);
-  }
 
+  $query->condition('n.status', 1);
   $query->condition('t.field_campaign_type_value', $params['type']);
   $query->condition('n.type', $params['type']);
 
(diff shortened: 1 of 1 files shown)


aaronschachter:
:+1: Looks good!

Brings to mind whether to kill those Mobile App dates / checkbox fields for housekeeping...

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -464,11 +464,8 @@ function dosomething_campaign_build_campaigns_query($params = array()) {
   $query = db_select('node', 'n');
   $query->join('field_data_field_campaign_type', 't', 't.entity_id = n.nid');
 
-  // @TODO: this is a temporary fix to address the need to allow unpublished mobile_app campaigns to be retrieved, only for the mobile app.
-  if (!$params['mobile_app']) {
```
aaronschachter, line 468: We should update the wiki wiki: https://github.com/DoSomething/phoenix/wiki/api#optional-query-parameters

angaither:
@aaronschachter are all mobile app dates/query param unused?

aaronschachter:
Not using them anywhere in iOS. @jonuy?

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
angaither, line 468: this doesn't remove the param, it just makes sure all campaigns are published

jonuy:
We should be good on the Android side too.

PR #6239, opened 2016-03-03
Who needs these fields not the mobile app thats for sure

### LINKED ISSUE #6236: Remove mobile app fields on campaign
seems these aren't being used, kill em!
- [x] [mobile_app bool](https://github.com/DoSomething/phoenix/search?utf8=%E2%9C%93&q=mobile_app)
- [x] [mobile_app_dates](https://github.com/DoSomething/phoenix/search?utf8=%E2%9C%93&q=mobile_app_date)
- [x] [update wiki](https://github.com/DoSomething/phoenix/wiki/api#optional-query-parameters)

angaither:
#### What's this PR do?

remove fields that were added to support the mobile app that are no longer being used. 

does the api still work? is there no info about the mobile app, we goooood.

these were added when we were unpublishing content, to hide it from the web, but still wanted to display it in the mobile app, as that's no longer the case :killwithfire: those fields. 

also I'll remove the references to these from the wiki after merge. 

Fixes #6236

cc @jonuy @aaronschachter @chloealee


lib/modules/dosomething/dosomething_api/plugins/restful/campaigns/1.0/DoSomethingCampaignsResource.class.php
```diff
@@ -24,10 +24,6 @@ public function publicFieldsInfo() {
       'property' => 'field_call_to_action',
     ];
 
-    $public_fields['on_mobile_app'] = [
-      'property' => 'field_on_mobile_app',
-    ];
-
```
weerd, line 30: We should actually just kill all the Restful Module stuff. We don't actually use this anymore.

weerd:
Looks good! Just make sure to check the DBLog for any errors regarding mobile app date stuff cause the API usually fails silently using just warnings in the DBLog ;) Otherwise if no errors there :+1:

PR #6243, opened 2016-03-03
Password reset flow

### LINKED ISSUE #6231: Password reset flow
some testing leads us to believe that when going through the password reset flow doesn't trigger
- [x] password length validation
- [x] updating password in NS

chloealee:
#### What's this PR do?

Adds password length validation (must be 6 characters or longer) when a user changes their password. 

Log into a user account on Thor and go to edit user profile page. Try to change password with a password that is less than 6 characters. Error message should appear. 

Fixes #6231

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -9,6 +9,7 @@ include_once 'dosomething_user.features.inc';
 include_once 'dosomething_user_valid_address.inc';
 include_once 'dosomething_user.theme.inc';
 
+
 /**
  * Implements hook_preprocess_page.
  */
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -234,6 +234,7 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
     break;
 
     case 'user_profile_form':
+      dosomething_northstar_update_user($form_state);
     case 'user_register_form':
       // Gather helper text.
       _dosomething_user_register_helper_text($form);
@@ -1294,6 +1295,8 @@ function dosomething_user_info_form($form, &$form_state, $account) {
   );
 
   $form['#validate'][] = 'dosomething_user_validate_address_field';
+
+  dosomething_user_update_user($form_state);
 
   return $form;
 }
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -234,7 +234,6 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
     break;
 
     case 'user_profile_form':
-      $form['#validate'][] = 'dosomething_user_validate_password_length';
     case 'user_register_form':
       // Gather helper text.
       _dosomething_user_register_helper_text($form);
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -235,7 +235,6 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
 
     case 'user_profile_form':
       $form['#validate'][] = 'dosomething_user_validate_password_length';
-      dosomething_northstar_update_user($form_state);
     case 'user_register_form':
       // Gather helper text.
       _dosomething_user_register_helper_text($form);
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -234,6 +234,8 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
     break;
 
     case 'user_profile_form':
+      $form['#validate'][] = 'dosomething_user_validate_password_length';
+      dosomething_northstar_update_user($form_state);
```
angaither: i don't think we want to break these out to different cases
also that `dosomething_northstar_update_user` should be running as a submit hook already [here](https://github.com/chloealee/dosomething/blob/password-reset-flow/lib/modules/dosomething/dosomething_user/dosomething_user.module#L258)

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -409,6 +411,22 @@ function dosomething_user_validate_address_field($form, &$form_state) {
...
 /**
+ * Validates that password is 6 or more characters long.  
+ *
+ * @param array $form
+ *  A drupal form.
+ * @param array $form_state
+ *  A drupal form_state array.
+ *
+ */
+function dosomething_user_validate_password_length($form, &$form_state) {
```
angaither, line 422: did we not have password validation anywhere in the php already? that seems kind of shocking 
do you remember anything about this @DFurnes?

lib/modules/dosomething/dosomething_user/dosomething_user.module
chloealee: oh I didn't see that - thanks @angaither ! I'll delete that line now.

chloealee: Do you think updating the password in NS in #6231 might have been an issue with just my user account then if that is there already? @angaither

lib/modules/dosomething/dosomething_user/dosomething_user.module
chloealee, line 422: We have it client-side [here](https://github.com/DoSomething/phoenix/blob/5896b2823f770c6aa3a1e9cf46d474ef4ae60564/lib/themes/dosomething/paraneue_dosomething/js/validators/auth.js#L215) but I couldn't find anything that checked it anywhere else? @angaither @DFurnes

DFurnes, line 422: Oh. Yeah, I just tested without JS enabled & we're not doing any server-side validation. So _that_ explains _that_.

angaither, line 422: in that case, we probably want this running on both `user_profile_form` and `user_register_form`

chloealee, line 422: yes makes sense! Added.

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -234,12 +234,15 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
...
 
     case 'user_profile_form':
+      $form['#validate'][] = 'dosomething_user_validate_password_length';
     case 'user_register_form':
       // Gather helper text.
       _dosomething_user_register_helper_text($form);
       // Conditionally displays certain fields based on configuration settings.
       _dosomething_user_global_display_fields($form);
 
+      $form['#validate'][] = 'dosomething_user_validate_password_length';
```
angaither, line 243: instead of adding this in two places, it should work the way it was where both 
case 'user_profile_form' & case 'user_register_form' use the same code below

angaither, line 243: the question then is why wasn't the form[submit] to update NS getting called correctly

chloealee, line 243: Could it be something that went wrong with the users dedupe script? @angaither

angaither:
seems some stuff isn't getting sent over to NS correctly  

working
- first name
- birthday
- mobile

not working
- address fields

chloealee:
@angaither just tested on production and I'm able to get the address to update properly!

angaither:
@chloealee which page are you making those address changes to?

chloealee:
@angaither https://www.dosomething.org/us/user/:uid/edit

chloealee:
@angaither I just changed my password on Thor and then tried to sign into the mobile app with the new password and was able to sign in! It seems like my ns profile wasn't correctly synced with my information on phoenix?

angaither:
awesome :+1:

PR #6246, opened 2016-03-07
Comp comp comp it up

angaither:
#### What's this PR do?
- Creates `dosomething_gladiator.module`
- config page to set custom urls per environment
- sends post request to gladiator app after user says "YEAH" to a competition
- Add `run_nid` to signup_data_form values array

Pull down, enable `dosomething_gladiator.module`
Set the url to either your local, or qa environment
Signup for a campaign with a competition enabled & see the user created in the platform

To get this to work on local Phoenix and Gladiator environments you need to add this line to `etc/hosts` inside of phoenix's vagrant
`10.0.2.2 gladiator.app`

We also might want to add some safety if anything along this route fails. But that also requires a better response from [the gladiator endpoint](https://github.com/DoSomething/gladiator/blob/master/documentation/endpoints/users.md)

Fixes #6201 

cc @deadlybutter @weerd

### DIFF
--- a/lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.module
+++ b/lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.module
@@ -1,0 +1,75 @@
+<?php
+/**
+ * @file
+ * Code for the dosomething_gladiatior.module.
+ */
+
+include_once('dosomething_gladiator.admin.inc');
+
+function dosomething_gladiator_menu() {
+  $items = [];
+  $items['admin/config/services/gladiator'] = array(
+    'title' => 'Gladiator',
+    'description' => 'Manage Gladiator connection settings.',
+    'page callback' => 'drupal_get_form',
+    'page arguments' => array('dosomething_gladiator_config_form'),
+    'access arguments' => array('administer modules'),
+    'file' => 'dosomething_gladiator.admin.inc',
+  );
+  return $items;
+}
+/**
+ * Build the Guzzle HTTP Client to make requests to Gladiator.
+ *
+ * https://github.com/DoSomething/gladiator
+ * @TODO - use correct URL
+ */
+function _dosomething_signup_build_http_client() {
+  $base_url = 'http://jsonplaceholder.typicode.com';
+
+  if (libraries_load('guzzle') == TRUE) {
+    $client = new GuzzleHttp\Client(array(
+      'base_uri' => 'http://jsonplaceholder.typicode.com',
+      'defaults' => array(
+        'headers' => array(
+          'Content-Type' => 'application/json',
+          'Accept' => 'application/json'
+          ),
+        ),
+    ));
+
+    return $client;
+  }
+
+  return;
+}
+
+/**
+ * Sends a user to Gladiator (https://github.com/DoSomething/gladiator)
+ *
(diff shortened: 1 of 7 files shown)


lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.module
```diff
@@ -0,0 +1,75 @@
...
+    'access arguments' => array('administer modules'),
+    'file' => 'dosomething_gladiator.admin.inc',
+  );
+  return $items;
+}
+/**
+ * Build the Guzzle HTTP Client to make requests to Gladiator.
+ *
+ * https://github.com/DoSomething/gladiator
+ * @TODO - use correct URL
```
sbsmith86: can remove this TODO, right?

lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.module
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Sends a user to Gladiator (https://github.com/DoSomething/gladiator)
+ *
+ * @param object $user
+ *   The user that signup up for the competition.
+ * @param array $values
+ *   Values from the signup data form.
+ */
+function dosomething_signup_send_user_to_gladiator($user, $values) {
```
sbsmith86: should this be`dosomething_gladiator_send_user()` or something? since now it is in the `dosomething_gladiator` module?

angaither: oh yersh, good catch. copy paste errorz

itsjoekent: just to make sure you noticed, it was also accidentally removed from the other file as well

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -508,7 +513,7 @@ function dosomething_signup_user_signup_data_form_submit($form, &$form_state) {
   // Update signup record.
   dosomething_signup_update_signup_data($values, $config);
   // If this is a competition sign up, send it to gladiator.
-  if ($config['competition_signup']) {
+  if ($config['competition_signup'] && module_exists('dosomething_gladiator')) {
     dosomething_signup_send_user_to_gladiator($user, $values);
```
itsjoekent: name needs to be updated here as well I think?

lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.module
angaither: @deadlybutter what was accidentally removed?

sbsmith86:
:100:

lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.module
itsjoekent: nevermind lookin at the wrong file

itsjoekent: too many tabs blargh!

weerd:
coooool :+1:

Issue #5557, opened 2015-10-20
modules didn't turn on in global branch


angaither:
- [ ] pathauto i18n
- [ ] pathauto i18n node
- [ ] apachesolr_multilingual

angaither:
these are in the [dosomething.info](https://github.com/DoSomething/phoenix/blob/global/lib/profiles/dosomething/dosomething.info) file on the global branch and they should have been enabled when we deployed to stage

mikefantini:
@angaither is this still an issue/something we want to add to 2.1?

angaither:
I think we can close this out, we turned it on manually in prod, but we should investigate if this continues to happen

angaither:
I added `dosomething_gladiator` and it didn't automatically enable with [this script](https://github.com/DoSomething/phoenix/blob/dev/bin/enable_modules.php)

angaither:
The problem was `apachesolr_confgen` wasn't enabling because it requires a dependency we don't have installed, so all other were failing

```
The following projects have unmet dependencies:                                                                                                    [ok]
apachesolr_confgen requires querypath
```

fixed by removing that module, since we don't seem to need it

PR #6247, opened 2016-03-07
Added output to see which modules should be enabled.

angaither:
#### What's this PR do?

Adds output to the `enable_modules` script to see what's failing 

run `ds_build` and you'll see an Array of modules printed out

Refs #5557


blisteringherb:
:+1:

PR #6248, opened 2016-03-07
Remove modules that are stuck and won't enable.

angaither:
#### What's this PR do?

removes modules that are already disabled in prod from the `dosomething.info` file

Runs `ds build`

Refs #5557


blisteringherb:
:+1:

PR #6249, opened 2016-03-07
Signup count

### LINKED ISSUE #6230: Support for count query param in GET signups index endpoint
#### Current Behavior
- The Signups endpoint only returns 25 results, regardless of any `count` query parameter passed.(https://northstar-thor.dosomething.org/v1/signups?user=554c2f3f469c64ed7d8b6061&count=500). The signup in particular i'm looking for is here: https://northstar-thor.dosomething.org/v1/signups?user=554c2f3f469c64ed7d8b6061&campaigns=6621 and is not returned by the GET `signups?user=:id` query. Also tried passing [`count=all`](https://github.com/DoSomething/phoenix/blob/a64fe84b159586a638a5d15f247e79eb8a2c4ede/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc#L24) without luck.
#### Desired Behavior
- Support `count` query parameter to bring back more than 25 records 
#### Why This Matters

Power users, such as Ben Kassoy, who is the user ID mentioned above, may see incomplete profiles while using the mobile app if they have more than 25 signups. (As we close and open more campaigns, this could happen more often). I found this bug by clicking through Ben's profile from the Yo Teach campaign, but then could not find the reportback on his profile (the mobile app query is https://northstar-thor.dosomething.org/v1/signups?user=554c2f3f469c64ed7d8b606)

chloealee:
#### What's this PR do?

Adds pagination to `/signups index` endpoint. 
Adds `count` and `page` query parameters. 

Hit `/signups` endpoint and top of response should look like this: 

```
meta: {
  pagination: { 
    total: 3299,
    per_page: 25,
    current_page: 9,
    total_pages: 132,
    links: {
      prev_uri: "http://dev.dosomething.org:8888/api/v1/signups?count=25&page=8",
      next_uri: "http://dev.dosomething.org:8888/api/v1/signups?count=25&page=10"
    }
  }
},
```

`/signups?count=3&page=9` should display three signup objects and take you to page 9. To test this further, change the values of these params. 

Adds pagination to index referenced in #6118  

Fixes #6230

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -858,36 +858,6 @@ function dosomething_signup_set_presignup_message($title) {
 }
 
 /**
- * Returns total number of Signups based on supplied set of filters. 
- *
- * @param array $filters
- * @return int
- */
-function dosomething_signup_get_signup_total_by_filters(array $filters = []) {
-  $unneccessaryFilters = ['count', 'page'];
-
-  $filters = dosomething_helpers_unset_array_keys($filters, $unneccessaryFilters);
-
-  $parameters = dosomething_helpers_convert_filters_to_url_parameters($filters);
-  $parameters['cache'] = TRUE;
-
-  $cache = new ApiCache();
-
-  $total = $cache->get('signups-tally', $parameters);
-
-  if ($total) {
-    $total = $total->data;
-  }
-  else {
-    $total = dosomething_signup_get_signups_query($filters, TRUE);
-
-    $cache->set('signups-tally', $parameters, $total);
-  }
-
-  return (int) $total;
-}
-
-/**
  * Returns array of campaign nid's that a user has signed up for.
  *
  * @param int $uid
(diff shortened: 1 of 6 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -859,6 +859,37 @@ function dosomething_signup_set_presignup_message($title) {
 }
 
 /**
+* Returns total number of Signups based on supplied set of filters. 
+*
+* @param array $filters
+* @return int
+*
+*/
+function dosomething_signup_get_signup_total_by_filters(array $filters = []) {
```
angaither, line 866: i think there's something funky about the formatting here, since it isn't showing php highlighting

chloealee, line 866: So weird! I deleted it and pushed it up then re-added and pushed it up and still doesn't show php highlighting :-/

angaither:
looks ok from my end :+1: 
cc @aaronschachter @jonuy  notifying you of api updates

Issue #6184, opened 2016-02-18
Add Help Text to Admin Add Reportback Form


namimody:
#### User Story

Sometimes, leads need to manually add reportbacks for users who have missed the campaign deadline or are having trouble uploading their photo. They use this link: dosomething.org/admin/users/rb/add. However, the form is tricky and doesn't provide a lot of clarity about how to use it.
#### Additional Information (optional)

FIELD :: Additional Help Text

NODE :: The campaign node this reportback was submitted for.
USER :: Type in the UID of the user that submitted this reportback and choose the UID from the dropdown menu. If you don't see the UID in the dropdown menu, please go to the user's profile, the edit tab, and copy the username listed in the topmost field. Slowly type this into the field and select the correct user from the dropdown menu. You MUST select the user from the dropdown menu for this form to submit.
CHOOSE FILE :: The image must be at least 480 by 480 pixels. The site will crop the image on its own, so if possible submit a square image with the subject in the center of the crop.
CAPTION :: This is public facing. Do not enter private user information.
TOTAL NUMBER OF NOUNS VERBED :: Only type in the number -- like "300" or "5".
#### Why This Matters

The form should be self explanatory.

namimody:
Hey @chloealee -- looks like the admin help text for the reportback form is showing up in the user-facing view as well :(

https://www.dosomething.org/us/campaigns/elephant-tusk-force

![image](https://cloud.githubusercontent.com/assets/12396076/13641565/e774d768-e5e6-11e5-850f-451c3360a1da.png)

PR #6252, opened 2016-03-09
removes #prefix and #suffix helper text

chloealee:
#### What's this PR do?

Removes `#prefix` and `#suffix` (title and description helper texts) from file upload since it was showing up on the user view and not just the form like intended. 

Visit a campaign page to the Reportback Image section and make sure `The image must be at least 480 by 480 pixels. The site will crop the image on its own, so if possible submit a square image with the subject in the center of the crop.` helper text doesn't appear. 

`#title` and `#description` were not appearing as helper text for the the file upload section of `/admin/users/rb/add` form so I used `#prefix` and `#suffix` as alternatives. `#theme_wrappers' => array(),` seems to be the line that breaks `#title` and `#description`. 

Removes helper text from user-facing view issue in #6184. Will open separate PR to close out issue.


angaither:
:+1:

PR #6256, opened 2016-03-09
adds admin check - only show helper text if logged in as an admin

### LINKED ISSUE #6184: Add Help Text to Admin Add Reportback Form

chloealee:
#### What's this PR do?

Updates helper text in admin add reportback form only - adds a check to make sure if it's an admin. If so, it'll show helper text. If not, text will not appear. This prevents it from showing up on main campaign page in PROVE IT section. 

From an admin account, hit `/admin/users/rb/add` endpoint and make sure all helper text is visible.

From non-admin account, hit any campaign page that user is signedup for, go to PROVE IT section, and make sure helper text is not visible.

Fixes #6184

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -79,7 +79,7 @@ function dosomething_reportback_form($form, &$form_state, $entity = NULL) {
     '#suffix' => 'The image must be at least 480 by 480 pixels. The site will crop the image on its own, so if possible submit a square image with the subject in the center of the crop.',
   );
   
-  if (!user_access('administer modules')) {
+  if (!user_access('edit any reportback')) {
     unset($form['reportback_submissions']['reportback_item']['#prefix']);
     unset($form['reportback_submissions']['reportback_item']['#suffix']);
   }
@@ -211,7 +211,7 @@ function dosomething_reportback_form($form, &$form_state, $entity = NULL) {
     '#description' => t("This is public facing. Do not enter private user information."),
   );
 
-  if (!user_access('administer modules')) {
+  if (!user_access('edit any reportback')) {
     unset($form['reportback_inputs']['caption']['#description']);
   }
 
@@ -234,7 +234,7 @@ function dosomething_reportback_form($form, &$form_state, $entity = NULL) {
     '#description' => t('Only type in the number -- like "300" or "5".'),
   );
 
-  if (!user_access('administer modules')) {
+  if (!user_access('edit any reportback')) {
     unset($form['reportback_inputs']['quantity']['#description']);
   }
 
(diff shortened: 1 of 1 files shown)


chloealee:
@angaither / @namimody if this is very time sensitive, I recommend merging this after it is approved. 

Otherwise, I found a comment in the [Drupal API docs](https://api.drupal.org/api/drupal/developer!topics!forms_api_reference.html/7) that describes how to add `#title` and `#description` when using `#field_wrappers`. I anticipate that this will take me a little while to figure out properly but if this isn't time sensitive, I can work on adding this before merging - let me know!

screen shot 2016-03-09 at 5 23 47 pm

chloealee:
also looping in @weerd - do you think it'd be risky to change around `theme_wrappers` that might affect anything else? Asking because of [these comments](https://github.com/DoSomething/phoenix/issues/6257#issuecomment-194559494).

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -200,8 +208,13 @@ function dosomething_reportback_form($form, &$form_state, $entity = NULL) {
     ),
     '#title' => t("Caption"),
     '#default_value' => $caption,
+    '#description' => t("This is public facing. Do not enter private user information."),
   );
 
+  if (!user_access('administer modules')) {
```
angaither: most people don't have the `administer modules` access, you should use `edit any reportback`

namimody:
Hi @chloealee! Just curious if this works for only site "administrators" or also for site "editors" -- most campaign leads have the "editor" role on Phoenix, like so: 
image

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
chloealee: @angaither ah great - changed! 

@namimody this addresses your comment - thanks for flagging too!

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -218,8 +231,13 @@ function dosomething_reportback_form($form, &$form_state, $entity = NULL) {
       )
     ),
     '#default_value' => $entity->quantity,
+    '#description' => t('Only type in the number -- like "300" or "5".'),
   );
 
+  if (!user_access('edit any reportback')) {
+    unset($form['reportback_inputs']['quantity']['#description']);
+  }
+
```
angaither: instead of doing these if checks for each thing
can you have one if and unset it all at once?

angaither:
:+1:

Issue #6260, opened 2016-03-10
Users are not successfully being "mock" authenticated against Northstar


DFurnes:
#### Current Behavior

We're currently running "mock" authentication against Northstar's `auth/verify` endpoint with each user's credentials when they log in to Phoenix. This is a dry run for authenticating users against Northstar sometime in the future (rather than the local Phoenix DB).

Frustratingly, a lot of users (62%) are being authenticated against Phoenix but failing when we try to then use the same credentials for Northstar. We can see this because we save the user's Northstar ID to their profile on login if we can verify their credentials, but save `NONE` if something goes wrong.

The mock authentication code was added in #6196 and updated in #6198.
#### Desired Behavior

All users that can log in to Phoenix should be able to be authenticated against Northstar using the same credentials (i.e. we should never be failing and saving `NONE` to the Northstar ID field).
#### Steps to Replicate

To see users who failed the `auth/verify` step :

``` sql
SELECT * FROM field_data_field_northstar_id as n
INNER JOIN users u on u.uid = n.entity_id
WHERE n.field_northstar_id_value='NONE'
LIMIT 1000;
```
#### Why This Matters

One day, we should be authenticating all users against Northstar as our single "source of truth", rather than having two databases with overlapping user profiles. See #4526.

angaither:
let's go the log all the things route. 

it's hard to find a discernible pattern in why this isn't working. 
make a `dosomething_user_northstar_log` table that can be turned on
we can log when a request is sent to both auth & update a user

store something like

| uid | op | user_lang | user country | json_encode($values) | northstar response | json_encode(ns data) |
| --- | --- | --- | --- | --- | --- | --- |
| 10 | auth | english | us | blah | 200 | blah |

angaither:
bugs so far
- [ ] user failed with invalid credentials 
  - (uid 2881255) used mobile number to log in
  - uid 1506487 email
- [x] user failed with invalid email 
  - uid 2881255 (extra space after email) 

known & accepted
- a bunch of people are getting a 404 on registering, it seems the update hook is called before the register hook and the first request is failing b/c the user doesn't exist yet, but the registration hook passes fine.

angaither:
we are also seeing quite a few user country's not saving, even though the fastly country code is set.

DFurnes:
[...]

Issue #6159, opened 2016-02-09
Ensure GL EN Users Should Be GL EN


namimody:
#### Current Behavior

It seems like some US members have their language set as Global English.
#### Desired Behavior

Want to make sure that all users have the correct language set on their profile based on their actual location.
#### Steps to Replicate

First we need to just figure out if the GL EN language setting is actually incorrect. I think we can do this by pulling users with addresses in their profile and comparing it to what their language is set to.
#### Why This Matters

For data purposes, US users should have a profile set to US EN.
#### Relevant Screenshots + Links

UIDs: 1981451 (was GL EN, I changed it to US EN when she reported that she was having trouble logging into her account from anything but an incognito window), 2553817

@mikefantini could you add the couple other UIDs that you've seen affected? Thanks!

namimody:
2553817

mikefantini:
Yeah so for https://www.dosomething.org/us/user/2553817/edit she is 100% located in the US and yet her language is set to GL EN

blisteringherb:
@angaither We store the user's country in Northstar, correct? Would you be able to give me a hint on how to query Northstar so I can cross check users with global english set as their profile language in Drupal with US users?

angaither:
@blisteringherb yeah, we for sure store country, but not language in NS. 
lemme think about this query hereeeeeeee

DFurnes:
@blisteringherb If you want anything (other than language!) from Northstar, [Aurora](https://aurora.dosomething.org) is probably your best bet. If you log in & then ping me in Slack, I can give you admin privileges.

blisteringherb:
@DFurnes Thanks!

angaither:
@blisteringherb also, duh it's mongo so query like
`db.users.find({drupal_id:"100001"});`
`db.users.find({country:"US"});`
etc etc

blisteringherb:
I've set up Fastly logging to S3 to make sure we're getting what we expect at the edge.

blisteringherb:
@sergii-tkachenko Would you happen to have any ideas about why this might be happening?

sergiitk:
Wrong Fastly headers for mobile users? cc @deadlybutter

itsjoekent:
hmmm, im not even sure how to test that one. also seems weird thats the cause.

i get if the phone location is quite a distance off, but i feel like it should be getting country right?

sergiitk:
[...]

PR #6262, opened 2016-03-11
Log all the stars

DFurnes:
#### What's this PR do?

Hello Drupal, my old friend. This PR adds request logging for Northstar update/verify requests, in the hopes that we can figure out why so many users can't be mock authenticated. :mag: 

If you want to pull this branch down to your local and run the ol' `drush updb -y`, any attempts to login or update your profile should log the corresponding Northstar requests to the `dosomething_northstar_request_log` table. Like so:

screen shot 2016-03-11 at 10 50 11 am

Nah. See the relevant ticket below.

References #6260.

---

For review: @angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -147,6 +147,25 @@ function dosomething_northstar_update_user($form_state) {
     watchdog('dosomething_northstar', 'User not updated : ' . $response->code, NULL, WATCHDOG_ERROR);
   }
 
+  // Add to request log
+  if(variable_get('dosomething_northstar_log')) {
+    // Don't log plaintext password.
+    if (isset($ns_user['password'])) {
+      $ns_user['password'] = '*****';
+    }
+
+    db_insert('dosomething_northstar_request_log')
+      ->fields([
+        'op' => 'update_user',
+        'uid' => $user->uid,
+        'user_lang' => $user->language,
+        'user_country' => $user->field_address[LANGUAGE_NONE][0]['country'],
+        'request_values' => json_encode($ns_user),
+        'response_code' => $response->code,
+        'response_values' => $response->data,
+      ])
+      ->execute();
+  }
 }
 
 /**
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -64,7 +64,9 @@ function dosomething_northstar_verify_user($credentials) {
   // Add to request log
   if(variable_get('dosomething_northstar_log')) {
     // Don't log plaintext password.
-    $credentials['password'] = '*****';
+    if (isset($credentials['password'])) {
+      $credentials['password'] = '*****';
+    }
 
     db_insert('dosomething_northstar_request_log')
       ->fields([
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -60,6 +61,24 @@ function dosomething_northstar_verify_user($credentials) {
     'data' => json_encode($credentials),
   ]);
 
+  // Add to request log
+  if(variable_get('dosomething_northstar_log')) {
+    // Don't log plaintext password.
+    $credentials['password'] = '*****';
```
angaither: maybe just unset this?

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -60,6 +61,24 @@ function dosomething_northstar_verify_user($credentials) {
...
+
+    db_insert('dosomething_northstar_request_log')
+      ->fields([
+        'op' => 'verify user',
+        'uid' => $user->uid,
+        'user_lang' => $user->language,
+        'user_country' => $user->field_address[LANGUAGE_NONE][0]['country'],
+        'request_values' => json_encode($credentials),
+        'response_code' => $response->code,
+        'response_values' => $response->data,
```
angaither: this is already `json` ya?

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -60,6 +61,24 @@ function dosomething_northstar_verify_user($credentials) {
...
+  if(variable_get('dosomething_northstar_log')) {
+    // Don't log plaintext password.
+    $credentials['password'] = '*****';
+
+    db_insert('dosomething_northstar_request_log')
+      ->fields([
+        'op' => 'verify user',
+        'uid' => $user->uid,
+        'user_lang' => $user->language,
+        'user_country' => $user->field_address[LANGUAGE_NONE][0]['country'],
```
angaither: hmm, I wonder if it's also helpful to see if this differs from the fastly country code to help with #6159

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -148,6 +166,25 @@ function dosomething_northstar_update_user($form_state) {
...
   }
 
+  // Add to request log
+  if(variable_get('dosomething_northstar_log')) {
+    // Don't log plaintext password.
+    if (isset($ns_user['password'])) {
+      $ns_user['password'] = '*****';
+    }
+
+    db_insert('dosomething_northstar_request_log')
```
angaither: is WET the opposite of DRY :sweat_drops: :smirk:

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
DFurnes: Yep, it's a JSON string.

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
DFurnes: Yeah, we could do that. On the other hand, I wonder if it's worth doing an `isset` check and only masking the password like this if the field is in the payload... could catch if we're somehow not sending it on every request.

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
DFurnes: Mmm, that's a good idea. :+1:

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
angaither: oh interesting! I like

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
DFurnes: Added to the log in 30a2e2b!

DFurnes:
@angaither Made updates based on your comments. I also added logging to the "register" action.

angaither:
:+1:

PR #6264, opened 2016-03-11
changes code to 404 if no campaigns are found

### LINKED ISSUE #6261: API individual resources should return error code 404 when not found
#### Current Behavior
#### Desired Behavior
#### Why This Matters

_

chloealee:
#### What's this PR do?

Changes status code to a 404 if a campaign does not exists from the `campaigns/:id` endpoint. 

In postman or paw, hit `/campaigns/abc` and the response code should be a 404 along with the error message. 

Fixes #6261

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
+++ b/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
@@ -51,6 +51,7 @@ class Campaign {
     $results = node_load_multiple($ids);
 
     if (!$results) {
+      http_response_code(404);
       throw new Exception('No campaign data found.');
     }
 


aaronschachter:
:+1:

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
```diff
@@ -51,6 +51,7 @@ public static function get($ids, $display = 'teaser') {
     $results = node_load_multiple($ids);
 
     if (!$results) {
+      http_response_code(404);
       throw new Exception('No campaign data found.');
```
angaither, line 55: does this need to be returned? do we need the exception below now?

chloealee, line 55: @aaronschachter if we just `return http_response_code(404);` here's what will get returned: 

```
{
  data: {
    id: null,
    title: null, 
    campaign_runs: null,
    language: null,
    translations: null,
    tagline: null,
    status: null,
    type: null
  }
}
```

would we want the above or for the exception to be returned?

aaronschachter, line 55: I'd imagine we want to keep consistent with an `error` object with a `message`, so it looks like the `throw new Exception` does that:  http://staging.beta.dosomething.org/api/v1/signups/abc 

cc Sir RESTful @DFurnes

chloealee, line 55: sounds good! So cool to merge? @aaronschachter

aaronschachter, line 55: :ship:

DFurnes, line 55: "Sir RESTful"! :tophat:

PR #6265, opened 2016-03-11
Stathat and errors oh my

angaither:
#### What's this PR do?

Updates the requests sent to gladiator to `drupal_http_request`
Adds stathat logging for success and failed people sent to gladiator

Pull down, set gladiator URLS, signup for a competition. 

Calling `response->getBody()` in guzzle was failing

Refs #6201

### DIFF
--- a/lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.module
+++ b/lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.module
@@ -67,6 +67,9 @@ function dosomething_gladiator_send_user_to_gladiator($user, $values) {
         'campaign_run_id' => $values['run_nid'],
         ]),
       ]);
+    if (module_exists('stathat')) {
+      stathat_send_ez_count('drupal - Gladiator - user sent - count', 1);
+    }
   } catch (GuzzleHttp\Exception\RequestException $e) {
     watchdog('dosomething_gladiator', $e, array(), WATCHDOG_ERROR);
   }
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.module
```diff
@@ -56,19 +46,32 @@ function _dosomething_gladiator_build_http_client() {
...
         'campaign_id' => $values['nid'],
         'campaign_run_id' => $values['run_nid'],
-        ]),
-      ]);
-  } catch (GuzzleHttp\Exception\RequestException $e) {
-    watchdog('dosomething_gladiator', $e, array(), WATCHDOG_ERROR);
+      ]),
+  ];
+
+  if (!$client) { return; }
```
weerd: Should this go above the `$options` array, since there's a call to `$client['headers']`?

angaither: ah yes, probably. but the way this is getting made now, w/o guzzle, it's probably not needed at all

angaither:
@weerd updated!

lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.module
```diff
@@ -56,19 +46,41 @@ function _dosomething_gladiator_build_http_client() {
 function dosomething_gladiator_send_user_to_gladiator($user, $values) {
   $client = _dosomething_gladiator_build_http_client();
```
weerd: So no need for `if (!$client) { return; }` here?

weerd, line 48: So no need for `if (!$client) { return; }` here?

weerd:
If that check isn't needed no worries. Looks good otherwise! :+1:

lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.module
angaither, line 48: yeah, so the way it was constructed before was relying on guzzle to load, and creating an object, now it's just returning an array, no need!

PR #6268, opened 2016-03-14
Signups API - Join bug on Reportbacks and Drupal Logs

### LINKED ISSUE #6266: Signups API - Join bug on Reportbacks?
#### Current Behavior
- Unsure if this is a bug in the JOIN in the Signups query, or if it's a data issue. Ran into this while testing mobile app with my DS account (user id 1176550)-- My reportback (ID 60) shows up in each of my 2 signups (id's 896000 and 44209): https://www.dosomething.org/api/v1/signups?user=1176550&campaigns=362

screen shot 2016-03-12 at 11 12 21 am
#### Desired Behavior
- Each signup should have a unique Reportback ID, if it exists for that Signup's run.
#### Why This Matters

Funky profile in mobile app, see below.

May be related to https://github.com/DoSomething/LetsDoThis-iOS/issues/956

comebackclothes

chloealee:
#### What's this PR do?

Changed the query to make sure that campaign runs also equal each other, not just campaign nodes. 
Added more explicit check to make sure `!empty($northstar_response->data)` to get rid of db logs. 

Make sure that each signup has a unique Reportback ID.  (can use https://www.dosomething.org/api/v1/signups?user=1176550&campaigns=362 in prod)

Fixes #6266

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -355,7 +355,7 @@ abstract class Transformer {
       $output['flagged'] = $data->flagged;
 
       try {
-        $items = ReportbackItem::get($data->reportback_items, FALSE);
+        $items = ReportbackItem::get($data->reportback_items);
         $items = services_resource_build_index_list($items, 'reportback-items', 'id');
       }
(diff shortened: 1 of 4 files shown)


angaither:
:+1:

PR #6271, opened 2016-03-15
Fixes SMS game active view

### LINKED ISSUE #4488: active tab for SMS Games should display SMS Game layout
if a campaign is an SMS Game and the campaign is closed, then the active tab should display as if it were an active SMS game. currently it is displaying an action page as if it were an active campaign

see https://beta.dosomething.org/node/2366 as an example

related to #3220

katiecrane:
#### What's this PR do?

Fixes SMS game active view to show up as an SMS game and not a regular campaign.

As an admin, go to a closed SMS campaign and go into the active view. It should show up like an SMS game and not a campaign. It should only be calling the `dosomething_campaign_preprocess_sms_game` if you are on the active path so I don't think it should be messing anything else up.

I made the fix that @aaronschachter suggested in the issue and it seems to work.

Fixes #4488

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -49,7 +49,6 @@ function dosomething_campaign_preprocess_node(&$vars) {
     }
     // Preprocess the vars for the campaign action page.
     dosomething_campaign_preprocess_action_page($vars, $wrapper);
-
     return;
   }
 
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -44,8 +44,9 @@ function dosomething_campaign_preprocess_node(&$vars) {
   // @see dosomething_campaign_menu().
   $active_path = 'node/' . $node->nid . '/active';
   if ($current_path == $active_path) {
-    if(dosomething_campaign_get_campaign_type($node) == 'sms_game') {
+    if(dosomething_campaign_get_campaign_type($node) === 'sms_game') {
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -44,6 +44,9 @@ function dosomething_campaign_preprocess_node(&$vars) {
   // @see dosomething_campaign_menu().
   $active_path = 'node/' . $node->nid . '/active';
   if ($current_path == $active_path) {
+    if(dosomething_campaign_get_campaign_type($node) == 'sms_game') {
```
angaither: a comment inspired by @sergii-tkachenko use `===`

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -44,6 +44,9 @@ function dosomething_campaign_preprocess_node(&$vars) {
...
+      dosomething_campaign_preprocess_sms_game($vars, $wrapper);
```
angaither, line 48: hmm, would we want to return after this?

katiecrane, line 48: I think probably

angaither:
:+1: 

cc @namimody

namimody:
Thanks @angaither and @katiecrane! And does the SMS active view also appear if the SMS game is open?

katiecrane:
@namimody Yeah it shows up correctly, but just under view and not active. The active view is only available if the campaign is closed. That's my understanding at least.

namimody:
Hi @katiecrane! Got it, thanks. Going to make a separate issue for the weird /edit view that currently exists for active SMS campaigns.

PR #6276, opened 2016-03-15
adds 404 error code when error is thrown in campaign.php, signup.php,…

### LINKED ISSUE #6261: API individual resources should return error code 404 when not found

chloealee:
#### What's this PR do?

Adds `404` response code when no campaign, signup, or kudos data is found. 

Hit any of the below endpoints to make sure the status code if a `404 Not Found`

Fixes #6261 

… and kudos.php

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
+++ b/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
@@ -90,6 +90,7 @@ class Campaign {
     $results = dosomething_campaign_get_campaigns_query($filters);
 
     if (!$results) {
+      http_response_code(404);
       throw new Exception('No campaign data found.');
     }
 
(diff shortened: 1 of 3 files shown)


DFurnes:
Hmm, it'd be nice if this logic could be shared between all the different model classes.

chloealee:
@DFurnes do you mean create a function for this in the entityreference.module file (which I assume is the file that `extends Entity` references?)?

DFurnes:
I think it could make sense to add a base `Model` class that each of these inherits from, but that can definitely be saved as a refactoring task somewhere down the road.

DFurnes:
Oh, I see. Interesting! `Kudos` and `Signup` extend from `Entity` but `Campaign` doesn't. Either way, that's out-of-bounds since it's a Dr00ps built-in so no biggie for this PR. :smile: 

/cc @weerd

chloealee:
ah thanks, @DFurnes ! Definitely a good practice to think about extracting to make it DRY :)

angaither:
:+1:

weerd:
post-merge :+1:

PR #6278, opened 2016-03-15
Remove code that updates a signup run if no signup on the current run…

### LINKED ISSUE #6269: POST Signup seems to overwrite signups from the 1st run
#### Current Behavior

This one's a doozy. Testing #6268 and https://github.com/DoSomething/LetsDoThis-iOS/issues/956 on Thor:
- My user 1176550 reported back on Run 1, I see the reportback
- Upon signing up for Run 2, my Signup created date is set to my Run 1 Signup created date (in 2014)
  https://thor.dosomething.org/us/signup/388008 and the API is returning just 1 signup for me, not 2 (https://thor.dosomething.org/api/v1/signups?user=1176550&campaigns=2461).  

Here were the steps:
- Logged in as my DS user 1176550 on iOS
- From my profile, tapped on my Birthday Mail reportback, was presented Do It Now button on the Birthday Mail campaign view
- Tap Do This Now -- campaign view shows I haven't reported back. That's good (previously after signup, it displayed my first reportback from the previous run here, thinking I completed the campaign -- this was #6266  resolved by #6268 )
- Go to my profile, expecting to see Birthday Mail show up in the Doing section, but also a 2nd time in the Done section with the reportback. It shows in Doing (out of order, i'd expect it to be the first) and there's no more Birthday Mail reportback in the Done section at all.

Hitting https://thor.dosomething.org/api/v1/signups?user=1176550&campaigns=2461 in browser (logged in as admin) shows one signup without a `reportback`:
screen shot 2016-03-14 at 4 33 48 pm

The `created` date is set to `1406644593` -- June 2014 -- which corresponds to the date during the 1st run. 

Looking my signup in Run 2, 6454 in the web admin Signups view -- I see my signup with that old date, and a few signup dates outside of the Run 2 dates (Run 2 has start date set to June 10, 2015)

screen shot 2016-03-14 at 4 46 42 pm

Searching for Birthday Mail [Signups for my email](https://thor.dosomething.org/us/node/2461/signups?uid=&mail=aschachter%40dosomething.org&source=&run_nid=), I'd expect 2 records but there is only 1, which seems to be a hybrid of the current run (with no reportback) and what i'm assuming was the run 1 signup created date. 

screen shot 2016-03-14 at 4 56 53 pm

In my [web profile](https://thor.dosomething.org/us/user/1176550) things look correct, Birthday Mail shows in my doing section, but my Birthday Mail reportback exists. However on the web, assuming (but didn't check in code) that we still just query the `dosomething_reportback` table, regardless if Signups exist

angaither:
#### What's this PR do?

removes code introduced in #6098 

lord. don't get me started.
it's difficult. see background context.
you basically just need to test that signing up still works.

This was added in to support #6095 when signup run ids were null or 0
But our main man, sharron, found a crazy bug that this causes.
Essentially if you signup for a campaign, but don't reportback
We close the campaign
Open the campaign with a new run
Go to view the campaign with from the doing section on your profile
The signup you had for the previous run, will get updated to the current run.

Fixes #6269

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -357,33 +357,8 @@ function dosomething_signup_exists($nid, $run_nid = NULL, $uid = NULL, $presignu
   }
   $sid = $result->fetchField();
 
-  // If not see if they are signed up for the campaign without the run.
-  // This section is added to test for duplicate signup problems.
-  // Refs #6095.
-  if (!is_numeric($sid) && !$presignup) {
-     $result = db_select('dosomething_signup', 's')
-      ->condition('uid', $uid)
-      ->condition('nid', $nid)
-      ->fields('s', array('sid'))
-      ->execute();
-      $running_free = TRUE;
-
-    $sid = $result->fetchField();
-  }
-
   // If a sid was found, return it.
   if (is_numeric($sid)) {
-
-    // If this was a node without a run, add that in.
-    // This section is also added to fix creating duplicate signups.
-    // Refs #6095.
-    if (isset($running_free)) {
-      $query = db_update('dosomething_signup')
-               ->fields(['run_nid' => $run_nid])
-               ->condition('sid', $sid)
-               ->execute();
-    }
-
     return $sid;
   }
 


sergiitk:
:+1:

aaronschachter:
Do we need to run an update for existing activity for users who might have had the same thing happen to me in #6269 ?

PR #6279, opened 2016-03-16
Revert "adds 404 error code when error is thrown in campaign.php, signup.php,…"

angaither:
Reverts DoSomething/phoenix#6276

refs #6261


aaronschachter:
:+1:

Issue #6274, opened 2016-03-15
/reportbacks and /reportback-items individual resources should return error code 403 404 when not found 


chloealee:
#### Current Behavior

Requesting a non existing reportback or reportback item from the `reportbacks/:id` or `reportback-items/:id` endpoints result in a 200 with an error object.

Requesting a reportback from the `reportbacks/:id` endpoints result in a 200 with an error object when accessed with non-admin account and if flagged reportback exists. 
#### Desired Behavior

If there is no reportback or reportback item found, status code should be 404. 
If there is a flagged reportback but user is not an admin, status code should be 403 with error message `Access denied.`
#### Steps to Replicate

test with endpoints `/reportbacks/z` and `reportback-items/z` in Postman/Paw and 200 status code will appear. 
#### Why This Matters

Incorrect data being returned in both status codes and error messages.

chloealee:
@aaronschachter I feel like I'm over thinking this but can't think of a simple solution of how to write this and getting confused. To determine if it should be a `403` or `404` error, I'm using this check: 

```
    $results = dosomething_reportback_get_reportback_items_query(['fid' => $ids]);
    if (!user_access('administer modules') && !$results) {
      http_response_code(403);
      throw new Exception('Access denied.'); //How do we account for if there is actually no reportback data found though?
    }
    else if (!$results) {
      http_response_code(404);
      throw new Exception('No reportback data found.');
    }
```

I'm having trouble thinking through this scenario - if it's a non-admin user, how will we know there are flagged reportbacks/reportback items that exist since as a non-admin, you can't query for that `flagged`? 

With the above code, it seems like that information would be misleading because it'll throw a `403` with the error message `Access denied.` if a reportback/reportback item doesn't exist.

chloealee:
oohh I think I mis-understood - looks like you _are_ able to view flagged reportbacks as an non-admin user but aren't able to query by `rbid` and `flagged` at the same time!

chloealee:
@aaronschachter figured out a way around this and should have a PR soon!

chloealee:
reopening because non admin users are still able to access flagged reportbacks and thus, we reverted #6293.

PR #6280, opened 2016-03-16
Moves 404 error codes to Transformer level

### LINKED ISSUE #6261: API individual resources should return error code 404 when not found

chloealee:
#### What's this PR do?

Sets 404 error codes at the transformer level.

Hit the below endpoints to make sure everything returns a 404 status code. 
`/signups/z`
`/kudos/z`
`/reportbacks/z`
`/reportback-items/z`
`/campaigns/z`

Also ensure that hitting `/signups` returns a 200 status code.

Refs https://github.com/DoSomething/northstar/issues/313
Found the above bug returning a signup object but throwing a 404 status code because there was a signup but couldn't find a kudos associated with the reportback. This informed us to move status codes to transformer level. 

Fixes #6261 and fixes 404 error code in #6274 - 403 error code fix to come in separate PR.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/includes/CampaignTransformer.php
+++ b/lib/modules/dosomething/dosomething_campaign/includes/CampaignTransformer.php
@@ -40,6 +40,7 @@ class CampaignTransformer extends Transformer {
     }
     catch (Exception $error) {
       // @TODO: Potentially log error to watchdog.
+      http_response_code('404');
       return [
         'data' => [],
       ];
@@ -77,6 +78,7 @@ class CampaignTransformer extends Transformer {
       $campaign = array_pop($campaign);
     }
     catch (Exception $error) {
+      http_response_code('404');
       return [
         'error' => [
           'message' => $error->getMessage(),
--- a/lib/modules/dosomething/dosomething_campaign/includes/CampaignTransformer.php
+++ b/lib/modules/dosomething/dosomething_campaign/includes/CampaignTransformer.php
@@ -40,7 +40,6 @@ class CampaignTransformer extends Transformer {
     }
     catch (Exception $error) {
       // @TODO: Potentially log error to watchdog.
-      http_response_code('404');
       return [
         'data' => [],
       ];
--- a/lib/modules/dosomething/dosomething_kudos/includes/KudosTransformer.php
+++ b/lib/modules/dosomething/dosomething_kudos/includes/KudosTransformer.php
@@ -23,7 +23,6 @@ class KudosTransformer extends Transformer {
       $kudos = services_resource_build_index_list($kudos, 'kudos', 'id');
     }
     catch (Exception $error) {
-      http_response_code('404');
       return [
         'error' => [
           'message' => $error->getMessage(),
--- a/lib/modules/dosomething/dosomething_kudos/includes/KudosTransformer.php
+++ b/lib/modules/dosomething/dosomething_kudos/includes/KudosTransformer.php
@@ -23,6 +23,7 @@ class KudosTransformer extends Transformer {
       $kudos = services_resource_build_index_list($kudos, 'kudos', 'id');
     }
     catch (Exception $error) {
+      http_response_code('404');
       return [
         'error' => [
           'message' => $error->getMessage(),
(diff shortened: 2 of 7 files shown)


lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php
```diff
@@ -18,6 +18,7 @@ public function index($parameters) {
     }
     catch (Exception $error) {
       // @TODO: Potentially log error to watchdog.
+      http_response_code('404');
```
aaronschachter: How would we get a 404 on the index endpoint? If we're throwing an error here, shouldn't we return an `error` object instead of `data` (2 lines below?)

aaronschachter: Curious if this is similar code to #6163 -- where we want a 200 with empty array if no results found for an index endpoint. (e.g. `signups?campaigns=23`)

chloealee: @aaronschachter I forget why we'd want a 200 if there are no results found again?

DFurnes: @chloealee Since it's an index endpoint, it's still technically "successful" even if it doesn't return anything. Whereas asking for a specific thing that doesn't exist is probably an error... i.e. the difference between going to your email inbox and seeing zero messages vs. going to a specific message and it not existing.

lib/modules/dosomething/dosomething_campaign/includes/CampaignTransformer.php
```diff
@@ -40,6 +40,7 @@ public function index($parameters) {
     }
     catch (Exception $error) {
       // @TODO: Potentially log error to watchdog.
+      http_response_code('404');
```
DFurnes: This was totes a mistake on my part when we were talking things through this morning. This should still return `200` with an empty data array if there are no matching results.

lib/modules/dosomething/dosomething_kudos/includes/KudosTransformer.php
```diff
@@ -23,6 +23,7 @@ public function index($parameters) {
       $kudos = services_resource_build_index_list($kudos, 'kudos', 'id');
     }
     catch (Exception $error) {
+      http_response_code('404');
```
DFurnes: Same.

lib/modules/dosomething/dosomething_reportback/includes/ReportbackTransformer.php
```diff
@@ -28,6 +28,7 @@ public function index($parameters) {
     }
     catch (Exception $error) {
       // @TODO: Potentially log error to watchdog.
+      http_response_code('404');
```
DFurnes: :boom:

lib/modules/dosomething/dosomething_signup/includes/SignupTransformer.php
```diff
@@ -26,6 +26,7 @@ public function index($parameters) {
       $total = $this->getTotalCount($filters);
     }
     catch (Exception $error) {
+      http_response_code('404');
```
DFurnes: :chart_with_downwards_trend:

lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php
chloealee: aaah such a good analogy - thanks @DFurnes ! I get it now :)

lib/modules/dosomething/dosomething_kudos/includes/KudosTransformer.php
chloealee: @DFurnes I removed this for index but should keep this error code here for the `show`, right?

DFurnes: Yep, I misread! :no_entry_sign: :eyes:

chloealee: Cool, thanks!

DFurnes:
:+1: from me!

Issue #6283, opened 2016-03-16
Missing migrated users phoenix->gladiator


angaither:
#### Current Behavior

user count in gladiator doesn't match the number of people who opted in to the comp
#### Desired Behavior

duh, people should be migrated in there
#### Steps to Replicate

no idea
#### Why This Matters

people won't get emails and stuff
#### Relevant Screenshots + Links

we can users are failing in stathat
![image](https://cloud.githubusercontent.com/assets/645205/13828353/784f94de-eb97-11e5-9bc2-d7e6d9189320.png)

but they are not in the `dosomething_gladiator_failed_task_log`

angaither:
seeing `usernotfound exception` in gladiatior /var/www/current/storage laravel.log

angaither:
data too long to enter into response code in phoenix

angaither:
this pr #6285 fixed the issue, and all users who weren't in the log, are now migrated in correctly.

PR #6285, opened 2016-03-17
Updated gladiator response value type to a medium text.

angaither:
#### What's this PR do?

Updated `response values` from a varchar to a medium text

run `drush updb -y`

Users weren't getting added to the `failed_log` table because the response was too large for a `varchar` col, doi

Refs #6283

### DIFF
--- a/lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.install
+++ b/lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.install
@@ -94,7 +94,7 @@ function dosomething_gladiator_update_7002(&$sandbox) {
 }
 
 /**
- * Update response values to be a long text field, as the response is LONG.
+ * Update response values to be a medium text field, as the response is medium-long.
  */
 function dosomething_gladiator_update_7003(&$sandbox) {
   $field = 'response_values';
--- a/lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.install
+++ b/lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.install
@@ -49,14 +49,16 @@ function dosomething_gladiator_schema() {
         'default' => 0,
       ),
       'response_code' => array(
-        'description' => 'The HTTP response code from Northstar',
+        'description' => 'The HTTP response code from Gladiator',
         'type' => 'int',
         'default' => 0,
       ),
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.install
```diff
@@ -90,3 +92,20 @@ function dosomething_gladiator_update_7002(&$sandbox) {
     }
   }
 }
+
+/**
+ * Update response values to be a long text field, as the response is LONG.
```
sbsmith86: lol just making sure, the title of the PR says you were updating to medium text, but this says long?

angaither: oh, yeah. I can update that. I realized long was TOO GIANT.

lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.install
```diff
@@ -90,3 +92,20 @@ function dosomething_gladiator_update_7002(&$sandbox) {
...
+function dosomething_gladiator_update_7003(&$sandbox) {
+  $field = 'response_values';
+  $table = 'dosomething_gladiator_failed_task_log';
+  // Update the field to a large text.
+  db_change_field($table, $field, $field, array(
+    'description' => 'The JSON response from Gladiator',
+    'type' => 'text',
+    'length' => '500000',
+    'not null' => FALSE,
+    'default' => NULL,
```
sbsmith86, line 108: Noob question: why do you have to pass this array twice? You defined it above as `response_values` and then pass the same array into this function?

angaither, line 108: AH, no that's an excellent question! it's going to take me some digging to find the issue, but we _didn't_ do this for an update in `dosomething_reportback.install`
it seems when you have multiple webheads after the update hook runs, it'll use the `schema` to define the table.
so I remember I changed something to be default null, and found it wasn't working until the schema was also updated.

sbsmith86, line 108: ahh cool! thanks for the explanation!

sbsmith86:
:100:

PR #6286, opened 2016-03-17
adds banner error if no image submitted with reportback

### LINKED ISSUE #5141: Trying to Submit RB without Photo Should Result in Banner Error
Currently, if you try to submit an RB without a photo, the page shifts a bit, but no error message is provided as to why the RB is not submitting. Perhaps a banner error message could read, "Oops! Try submitting again when you've added a photo of you taking action."

katiecrane:
#### What's this PR do?

If a user tries to submit a reportback with no image attached, they get a "Oops! Make sure to upload a photo of you completing the campaign before submitting." in the same way that they get an error when uploading their picture if their image is too small.

Try to submit a reportback with no image. Try to submit a reportback with no image a few times. Try to submit a correct reportback to make sure the error goes away. Yay!

Fixes #5141 

cc: @weerd @DFurnes

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
@@ -2,7 +2,7 @@
 
 import ImageUploadFallback from '../images/ImageUploadFallback';
 import reportbackTplSrc from 'reportback/templates/reportback.tpl.html';
-import imageUpload from '../images/ImageUpload';
+import ImageUpload from '../images/ImageUpload';
 
 const $ = require('jquery');
 const debounce = require('lodash/function/debounce');
@@ -34,6 +34,7 @@ const Reportback = {
 
   adminAccess: null,
 
+  uploader: null,
 
   /**
    * Initialize the Reportback interface.
@@ -183,7 +184,7 @@ const Reportback = {
    * Initialize the ImageUpload interface.
    */
   imageUploadInit: function() {
-    imageUpload(this.$uploadButton);
+    this.uploader = new ImageUpload(this.$uploadButton);
 
     this.$captionField.hide();
   },
--- a/lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
@@ -1,301 +1,284 @@
-/* eslint-disable */
+/* global Modernizr */
 
-/* ----------------------------------------------------------
- * @TODO: ^ Enable linting by removing `eslint-disable`! ^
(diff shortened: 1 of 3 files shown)


angaither:
handing this over to a :js:smen

lib/themes/dosomething/paraneue_dosomething/js/validators/reportback.js
```diff
@@ -44,6 +44,23 @@ define(function(require) {
...
+    if( string !== "" ) {
+      return done({
+        success: true,
+      });
+    } else {
+      let $reportbackSubmissions = jQuery(".reportback__submissions");
+      if (!($reportbackSubmissions.find(".messages.error").length > 0)) {
+        $reportbackSubmissions.prepend("<div class='messages error'>Oops! Make sure to upload a photo of you completing the campaign before submitting.</div>");
+      }
+     return done({
```
DFurnes: Looks like indentation might be a bit off here.

DFurnes:
Not 100% necessary for merge, but I'd like to get in the habit of updating files to ES2015 syntax whenever we touch them. It _should_ be pretty straightforward to remove the `/* eslint-disable */` at the top of that file and fix whatever errors are displayed when you run `grunt test` in the theme directory.

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
```diff
@@ -1,301 +1,284 @@
...
-    /**
-     * Initialize the ImageUpload interface.
-     */
-    imageUploadInit: function () {
-      new ImageUpload(this.$uploadButton);
+  /**
+   * Initialize the ImageUpload interface.
+   */
+  imageUploadInit: function() {
+    imageUpload(this.$uploadButton);
```
DFurnes: Hmm, is this from [this rule](http://eslint.org/docs/rules/no-new)? I think if we remove the `new` operator, all the references to `this` inside the `ImageUpload` function will refer to the window object, rather than being stored on the image uploader itself. I think a good solution for now would be to store `new ImageUpload` to a property on the `Reportback` object to silence the warning without changing behavior.

katiecrane: @DFurnes Sure is. Okay I will look into that. Thanks!

DFurnes:
The rest of the changes in b9fbc51 look good! :+1: 

If you're up for it, it'd be great to remove the warnings from the `js/validators/reportback.js` that you had updated in this PR as well.

katiecrane:
@DFurnes Oh yeah, there are 3 changed files! Yeah I will go for that one as well

DFurnes:
:shipit:

PR #6293, opened 2016-03-18
Reportback 403

### LINKED ISSUE #6274: /reportbacks and /reportback-items individual resources should return error code 403 404 when not found

chloealee:
#### What's this PR do?

If a user is not an admin, a flagged reportback or reportback item will not be returned. Instead, they will see error message `Access denied.` and response will return a `403` status code.

Find a flagged `rbid` and `fid`. 

From admin account, `/reportbacks/:rbid` should return `202` status code and reportback object. 

From an anon user account, same URL should return `403` status code and error message that reads `Access denied.`

Fixes #6274

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
@@ -107,8 +107,8 @@ function dosomething_reportback_build_reportbacks_query($params = array()) {
  * @return object
  *   An executed database query object to iterate through.
  */
-function dosomething_reportback_get_reportbacks_query($params, $load_all = FALSE) {
-  $params = dosomething_reportback_filter_flagged_reportbacks($params, $load_all);
+function dosomething_reportback_get_reportbacks_query($params, $include_flagged = FALSE) {
+  $params = dosomething_reportback_filter_flagged_reportbacks($params, $include_flagged);
 
   $query = dosomething_reportback_build_reportbacks_query($params);
   $offset = dosomething_helpers_isset($params, 'offset', 0);
@@ -130,7 +130,7 @@ function dosomething_reportback_get_reportbacks_query($params, $load_all = FALSE
  * @param  array  $params
  * @return mixed
  */
-function dosomething_reportback_filter_flagged_reportbacks($params, $load_all) {
+function dosomething_reportback_filter_flagged_reportbacks($params, $include_flagged) {
   if (user_access('view any reportback')) {
     if (isset($params['flagged'])) {
       unset($params['flagged']);
@@ -139,7 +139,7 @@ function dosomething_reportback_filter_flagged_reportbacks($params, $load_all) {
     return $params;
   }
 
-  if ($load_all) {
+  if ($include_flagged) {
     $params['flagged'] = TRUE;
   } 
   else {
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
@@ -131,7 +131,7 @@ function dosomething_reportback_accessible_results ($results) {
   }
   else {
     foreach ($results as $key => $reportback) {
-      if ($reportback->flagged == 1) {
+      if ($reportback->flagged === 1) {
         unset($results[$key]);
       }
     }
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
@@ -107,8 +107,8 @@ function dosomething_reportback_build_reportbacks_query($params = array()) {
  * @return object
  *   An executed database query object to iterate through.
  */
-function dosomething_reportback_get_reportbacks_query($params, $include_flagged = FALSE) {
-  $params = dosomething_reportback_filter_flagged_reportbacks($params, $include_flagged);
+function dosomething_reportback_get_reportbacks_query($params) {
+  // $params = dosomething_reportback_filter_flagged_reportbacks($params, $include_flagged);
 
   $query = dosomething_reportback_build_reportbacks_query($params);
   $offset = dosomething_helpers_isset($params, 'offset', 0);
@@ -123,30 +123,26 @@ function dosomething_reportback_get_reportbacks_query($params, $include_flagged 
   return $result;
 }
 
-
 /**
- * Filter out requested flagged Reportbacks based on user permissions.
- *
- * @param  array  $params
- * @return mixed
+ * Determine if user has access to flagged Reportbacks based on user permission. 
+ * @param array $results
  */
-function dosomething_reportback_filter_flagged_reportbacks($params, $include_flagged) {
+function dosomething_reportback_accessible_results ($results) {
   if (user_access('view any reportback')) {
-    if (isset($params['flagged'])) {
-      unset($params['flagged']);
+    return $results;
+  }
+  else {
+    foreach ($results as $key => $reportback) {
+      if ($reportback->flagged == 1) {
+        unset($results[$key]);
+      }
     }
 
-    return $params;
+    if (!$results) {
+      return FALSE;
+    }
+    else {
+      return $results;
+    }
   }
-
-  if ($include_flagged) {
-    $params['flagged'] = TRUE;
-  } 
-  else {
-    // Only show reviewed and/or un-flagged Reportbacks.
-    $params['flagged'] = FALSE;
-  }
-
-  return $params;
 }
-
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
@@ -108,8 +108,6 @@ function dosomething_reportback_build_reportbacks_query($params = array()) {
  *   An executed database query object to iterate through.
  */
 function dosomething_reportback_get_reportbacks_query($params) {
-  // $params = dosomething_reportback_filter_flagged_reportbacks($params, $include_flagged);
-
   $query = dosomething_reportback_build_reportbacks_query($params);
   $offset = dosomething_helpers_isset($params, 'offset', 0);
   $count = dosomething_helpers_isset($params, 'count', 25);
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
@@ -107,8 +107,8 @@ function dosomething_reportback_build_reportbacks_query($params = array()) {
  * @return object
  *   An executed database query object to iterate through.
  */
-function dosomething_reportback_get_reportbacks_query($params) {
-  $params = dosomething_reportback_filter_flagged_reportbacks($params);
+function dosomething_reportback_get_reportbacks_query($params, $load_all = FALSE) {
+  $params = dosomething_reportback_filter_flagged_reportbacks($params, $load_all);
 
   $query = dosomething_reportback_build_reportbacks_query($params);
   $offset = dosomething_helpers_isset($params, 'offset', 0);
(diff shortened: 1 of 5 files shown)


angaither:
I'm going to hand this one over to mr api @weerd

lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
```diff
@@ -92,6 +92,14 @@ public static function get($ids) {
 
     $results = dosomething_reportback_get_reportbacks_query(['rbid' => $ids]);
 
+    if (!user_access('administer modules') && !$results) {
+      $results = dosomething_reportback_get_reportbacks_query(['rbid' => $ids], TRUE);
```
weerd: A little confused as to why this is happening in here and passing the `TRUE` if does _not_ have user access?

weerd: Oh wait. I was misreading it. Still can't setting it to true happen in the `dosomething_reportback_get_reportbacks_query()` function?

lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php
```diff
@@ -40,11 +40,21 @@ public function index($parameters) {
...
       $reportbackItem = services_resource_build_index_list($reportbackItem, 'reportback-items', 'id');
     }
     catch (Exception $error) {
-      http_response_code('404');
+      if ($error->getMessage() === 'Access denied.') {
+        http_response_code('403');
+      }
+      else {
+        http_response_code('404');
+      }
```
weerd, line 52: Based on this: http://php.net/manual/en/class.exception.php you could initially pass the error code to the exception when it is thrown, which would eliminate the need for the above logic since the code would be on a property in the exception and you just output it :)

lib/modules/dosomething/dosomething_reportback/includes/ReportbackTransformer.php
```diff
@@ -54,7 +54,12 @@ public function show($id) {
       $reportback = array_pop($reportback);
     }
     catch (Exception $error) {
-      http_response_code('404');
+      if ($error->getMessage() === 'Access denied.') {
+        http_response_code('403');
+      }
+      else {
+        http_response_code('404');
```
weerd, line 61: Same as earlier comment in similar code above.

lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php
chloealee, line 52: ooo that's really cool to know you can just add it as a second arg! The only thing about moving it back to the class is that sometimes, we don't want the code to be thrown ([explained in this issue](https://github.com/DoSomething/phoenix/pull/6280)) so we moved it to the Transformer level.

lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
chloealee: I don't think it can because if the user doesn't have permission to administer modules, the `flagged` param will get ignored. This is set [here](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc#L133) so that's why I used the param to change that.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
```diff
@@ -130,7 +130,7 @@ function dosomething_reportback_get_reportbacks_query($params) {
  * @param  array  $params
  * @return mixed
  */
-function dosomething_reportback_filter_flagged_reportbacks($params) {
+function dosomething_reportback_filter_flagged_reportbacks($params, $load_all) {
```
angaither: I feel a little iffy about this `load_all` business. 
1. maybe that's because the name is confusing `include_flagged` may make more sense
2. it also seems strange to be returning the flagged results when the user shouldn't be getting them.

chloealee: good call about the naming - I can definitely change to `include_flagged` for clarity! 

The user won't actually be returned the flagged results, it's more so just checking to see if there are flagged items or not and if so, then qualifies for a `403` status code.

lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php
weerd, line 52: Oh, I didn't realize that. I think I might be confused though; not sure why that was a problem.

chloealee, line 52: It was causing an issue because when hitting the index of `/signups`, a `404` status code was popping up in addition to signup objects. This was because it was looking for kudos associated with reportbacks that were being returned in the signup object - if there were no kudos, a `404` status code was returned in addition to signups so it made this very confusing. Chatting more with @angaither and @DFurnes , we decided to avoid this, the error codes should be returned at the transformer level. LMK if this is still confusing bc I'm not the best at explaining!

angaither, line 52: @weerd the issue was the api page was returning results, but also a 404 (specifically for kudos or reportbacks)

entity_translation
```diff
@@ -0,0 +1 @@
+Subproject commit 99f7b6ba722a943a5a95e411894608ca7230d5e7
```
chloealee: I think this got added by mistake since I'm also working on a patch for another issue. Is there a way I can delete this @angaither ?

lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
weerd: I think needing to do the same call but with an added param here just feels like a bad pattern and what's leading to confusion. I feel there's likely a better way of handling this, just not sure what it is offhand... maybe we could chat it out in Slack?

lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php
weerd, line 52: @angaither @chloealee not entirely sure, but couldn't the code just have been overridden and set to `200` if there were results in the response?

entity_translation
angaither: either `git reset HEAD path/to/unwanted_file`
`git rm path/to/file`

chloealee: ah perfect, thank you!

chloealee:
@weerd made the changes we talked about in slack! I _think_ that the reportback items code is okay as is?

lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php
chloealee, line 52: @weerd would this happen in the SignupTransformer?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
```diff
@@ -123,25 +121,26 @@ function dosomething_reportback_get_reportbacks_query($params) {
...
-function dosomething_reportback_filter_flagged_reportbacks($params) {
+function dosomething_reportback_accessible_results ($results) {
   if (user_access('view any reportback')) {
-    if (isset($params['flagged'])) {
-      unset($params['flagged']);
+    return $results;
+  }
+  else {
+    foreach ($results as $key => $reportback) {
+      if ($reportback->flagged == 1) {
```
weerd: Use a `===` which checks for type too. Just best to get into the habit of using triple equals; only very few exceptions ;)

lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
```diff
@@ -96,6 +96,10 @@ public static function get($ids) {
       throw new Exception('No reportback data found.');
     }
 
+    if (!dosomething_reportback_accessible_results($results)) {
+      throw new Exception('Access denied.');
+    }
```
weerd, line 101: Yaaaaisss! :dancer:

lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php
```diff
@@ -40,11 +40,21 @@ public function index($parameters) {
   public function show($id) {
     try {
       $reportbackItem = ReportbackItem::get($id);
-      $reportbackItem = $this->removeUnauthorizedResults($reportbackItem);
+
+      if (!user_access('administer modules')) {
+        $reportbackItem = $this->removeUnauthorizedResults($reportbackItem, TRUE);
+      }
```
weerd: Can you add a method similar to the one that you added to the `dosomething_reportback_accessible_results()` in reportback query but for ReportbackItems and deal with removing flagged items or their flagged parent RB? Not sure how close you can match the same approach as above with these... what do you think @chloealee?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
chloealee: ah good to note, will get in the habit of using triple equals from now on - thanks for explaining, Diego!

lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php
chloealee: @weerd I think that is a good idea! Removed the params logic and tweaked `removeUnauthorizedResults` and the `show` function in `ReportbackTransformer.php` to more closely represent what is going on in `dosomething_reportback_accessible_results()` - let me know  what you think!

lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php
```diff
@@ -40,11 +40,26 @@ public function index($parameters) {
   public function show($id) {
     try {
       $reportbackItem = ReportbackItem::get($id);
-      $reportbackItem = $this->removeUnauthorizedResults($reportbackItem);
-      $reportbackItem = services_resource_build_index_list($reportbackItem, 'reportback-items', 'id');
+
+      if (!user_access('administer modules')) {
+        $reportbackItem = $this->removeUnauthorizedResults($reportbackItem);
```
weerd: `removeUnauthorizedResults()` this method already has a check to see if `user_access('view any reportback')`... so  don't think we need another. If it turns out we actually want the rule for `administer modules` vs `view any reportback` then maybe change the conditional in `removeUnauthorizedResults()`?

lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php
```diff
@@ -40,11 +40,26 @@ public function index($parameters) {
...
+      }
+
+      if ($reportbackItem) {
+        $reportbackItem = services_resource_build_index_list($reportbackItem, 'reportback-items', 'id');
+      } else {
+        throw new Exception('Access denied.');
```
weerd: Instead of throwing the `Access denied` exception here, we could also do that in the `removeUnauthorizedResults()`. If there are no results passed to `removeUnauthorizedResults()` as the `$data`, then that means no items found so then immediately throw the `Exception('No reportback items data found.'). If there are items, but they get removed due to access restrictions, then change`Exception('No reportback items data found.')`to`Exception('Access denied.')`.

chloealee:
thanks @weerd ! Those suggestions make it waaay more easier to go through and understand!

weerd:
Looking good! If testing a variety of possibilities w/ the endpoint showed all was well, should be good to merge!

:sheep: IT!

weerd:
Noted to @chloealee IRL, we may need to take a look at whether the caching might affect revealing unauthorized RBs if first an RB is hit using admin, then as anon and grabbing same from cache. Don't remember offhand, but should be something to investigate.

chloealee:
Tested on staging and it seems like reportbacks get cached and are returned to anon user (reportback items do not). Made an issue for this [here](https://github.com/DoSomething/phoenix/issues/6318).

PR #6294, opened 2016-03-18
view user profile by northstar id

### LINKED ISSUE #6275: Use a mongo id to get to a user profile
#### User Story

As an admin I should be able to view user profile in other apps and copy the link with the mongo id, and phoenix knows to take me to the user profile page

dosomething.org/user/[crazymongoid] shows user profile
#### Why This Matters

as we rely more on mongo ids, that should mean something in phoenix

phoenix user id needs to be exposed in other apps, as well as the mongo id

katiecrane:
#### What's this PR do?

If you go to /user/[phoenix id or northstar id] it will take you to the correct user page.

Go to user/[phoenix id], user/[northstar id], and user/[anything else that exists] to make sure nothing is broken.

I did this via redirect and I'm not sure if that's the best way to approach this or not. Also, this will break if we have any pages that are user/[other words]. I didn't find any paths like that, but I'm not positive.

Fixes #6275 

cc: @DFurnes

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -46,6 +46,21 @@ function dosomething_user_init() {
       dosomething_user_non_staff_redirect();
     }
   }
+
+  // If viewing a user/[northstar id]
+  if(arg(0) == 'user') {
+    if(!is_numeric(arg(1))) {
+      $northstar_id = arg(2,request_path());
+      $query = db_select('field_data_field_northstar_id', 'f')
+              ->condition('f.field_northstar_id_value', $northstar_id)
+              ->fields('f', ['entity_id']);
+      $result = $query->execute()->fetchAll();
+      $phoenix_id = $result[0]->entity_id;
+      $user = user_load($phoenix_id);
+      $new_path = 'user/' . $phoenix_id;
+      drupal_goto($new_path);
+    }
+  }
 }
 
 /**
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -108,27 +108,32 @@ function dosomething_user_menu() {
 }
 
 function dosomething_user_menu_alter(&$items) {
-    $items['user/%'] = array(
+  unset($items['user']);
+  unset($items['user/%user']);
+  $items['user/%'] = array(
     'page callback' => 'dosomething_user_get_view',
     'page arguments' => array(1),
-    'type' => MENU_CALLBACK,
     'access callback' => TRUE,
   );
+  unset($items['user/%user']);
 
   return $items;
 }
 
 function dosomething_user_get_view($id) {
-  if(is_numeric($id)) {
-    $user = user_load($id);
-    return user_view($user);
+  if (is_numeric($id)) {
+    $account = user_load($id);
+
+    return user_view($account);
   }
   else {
     $phoenix_user = dosomething_user_get_user_by_northstar_id($id);
-    if(!$phoenix_user) {
+    if (!$phoenix_user) {
       drupal_not_found();
     }
-    return user_view($phoenix_user);
+
+    // return user_view($phoenix_user);
+    drupal_goto('user/' . $phoenix_user->uid);
   }
 }
 
@@ -1016,7 +1021,7 @@ function dosomething_user_get_user_by_northstar_id($id) {
           ->condition('f.field_northstar_id_value', $id)
           ->fields('f', ['entity_id']);
   $result = $query->execute()->fetchField();
-  if($result) {
+  if ($result) {
     return user_load($result);
   }
 
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -50,16 +50,10 @@ function dosomething_user_init() {
   // If viewing a user/[northstar id]
   if(arg(0) == 'user') {
     if(!is_numeric(arg(1))) {
-      $northstar_id = arg(2,request_path());
-      $query = db_select('field_data_field_northstar_id', 'f')
-              ->condition('f.field_northstar_id_value', $northstar_id)
-              ->fields('f', ['entity_id']);
-      $result = $query->execute()->fetchAll();
-      $phoenix_id = $result[0]->entity_id;
+      $phoenix_id = dosomething_user_get_user_by_northstar_id(arg(1));
       if(!$phoenix_id) {
         drupal_not_found();
       }
-      $user = user_load($phoenix_id);
       $new_path = 'user/' . $phoenix_id;
       drupal_goto($new_path);
     }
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -46,6 +46,24 @@ function dosomething_user_init() {
       dosomething_user_non_staff_redirect();
     }
   }
+
+  // If viewing a user/[northstar id]
+  if(arg(0) == 'user') {
+    if(!is_numeric(arg(1))) {
+      $northstar_id = arg(2,request_path());
+      $query = db_select('field_data_field_northstar_id', 'f')
```
angaither: try abstracting this `select` out into a function like `dosomething_user_get_user_by_mobile`

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -46,6 +46,24 @@ function dosomething_user_init() {
       dosomething_user_non_staff_redirect();
     }
   }
+
+  // If viewing a user/[northstar id]
+  if(arg(0) == 'user') {
```
angaither: instead of putting this in an `init` hook, it probably makes more sense in the `hook_menu`

katiecrane: @angaither it doesn't look like `dosomething_user_menu()` is getting called? No dpms from in there are printing out and when I move the code into there it no longer works

angaither: once you add a custom menu item 'user/%', make sure to `drush cc all`!

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -103,9 +103,34 @@ function dosomething_user_menu() {
...
+
+    return $items;
+}
+
+function dosomething_user_menu_alter(&$items) {
+    $items['user/%'] = array(
+    'page callback' => 'dosomething_user_get_view',
+    'page arguments' => array(1),
+    'type' => MENU_CALLBACK,
+    'access callback' => TRUE,
```
aaronschachter: Fairly certain setting this to `TRUE` may expose the user's profile publicly...

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -103,9 +103,34 @@ function dosomething_user_menu() {
...
+  );
+
   return $items;
 }
 
+function dosomething_user_get_view($id) {
+  if(is_numeric($id)) {
```
aaronschachter: its been a while since i've been writing code in phoenix, but shouldn't these `if` statements have a space?
 `if (is_numeric($id)) {`

katiecrane: yeah they should! good catch thanks!

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -103,9 +103,36 @@ function dosomething_user_menu() {
...
+    return $items;
+}
+
+function dosomething_user_menu_alter(&$items) {
+  unset($items['user']);
+  unset($items['user/%user']);
+  $items['user/%'] = array(
+    'page callback' => 'dosomething_user_get_view',
+    'page arguments' => array(1),
+    'access callback' => TRUE,
```
katiecrane: just a reminder that I still need to change this

katiecrane:
Currently, when viewing a northstar id account page like this:

```
 $phoenix_user = dosomething_user_get_user_by_northstar_id($id);  
    if (!$phoenix_user) {  
      drupal_not_found();  
    }  
 return user_view($phoenix_user);
```

you get this:
image

instead of this:
image

Unsetting `$items['user/%user']` removes the words "My Account" from the outer header, but the page still remains nested.

cc: @deadlybutter , @aaronschachter

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -103,9 +103,36 @@ function dosomething_user_menu() {
...
+  if (is_numeric($id)) {
+    $account = user_load($id);
+    return user_view($account);
+  }
+  else {
+    $phoenix_user = dosomething_user_get_user_by_northstar_id($id);
+    if (!$phoenix_user) {
+      drupal_not_found();
+    }
+    drupal_goto('user/' . $phoenix_user->uid);
```
aaronschachter, line 129: This  `drupal_goto` looks suspect, per https://github.com/DoSomething/phoenix/pull/6294#issuecomment-199860223

Does `return user_view($phoenix_user)` fix ?  If so, it's minor but the `if/else` could just set our `$account` User object, and then `return user_view($account)` outside of the if/else block

katiecrane, line 129: @aaronschachter The profile template is only wrapped in the default account page header with `return user_view($phoenix_user)`, if I use `drupal_goto` it looks correct, sorry for the confusion. 

I think ideally we'd want to use `user_view` though, in which case good call on where to put the return.

DFurnes:
@katiecrane We conditionally switch between either a "full-width" template (which lets us completely customize the markup for a node) or a default template that includes page header and gutters (which we generally use for default Drupal markup, like the "forgot password" page). 

Here's [where that logic happens](https://github.com/DoSomething/phoenix/blob/39111ee657f7325a30a5806a414aad43845a7d68/lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_page.inc#L103-L123) in the theme... not sure why that isn't triggering correctly with your changes here but I'm guessing it's somehow related.

katiecrane:
@DFurnes thanks! this definitely looks like the place to check

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -106,6 +106,29 @@ function dosomething_user_menu() {
   return $items;
 }
 
+function dosomething_user_menu_alter(&$items) {
+  $items['user/%'] = array(
+    'page callback' => 'dosomething_user_get_view',
+    'page arguments' => array(1),
+    'access callback' => 'user_access',
+    'access arguments' => array('administer modules'),
```
angaither: not all of our editors have this permission. we should use something like `view any user profile` or something that all people on staff can access

katiecrane: @angaither oooo is there a list of all of those somewhere? That one seems very appropriate though

angaither: @katiecrane you can see the breakdown of all permission on the BEAUTIFUL `/admin/people/permissions` page

katiecrane: @angaither EASILY my new favorite webpage.

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -106,6 +106,29 @@ function dosomething_user_menu() {
...
+    'page callback' => 'dosomething_user_get_view',
+    'page arguments' => array(1),
+    'access callback' => 'user_access',
+    'access arguments' => array('view any user profile'),
+  );
+  return $items;
+}
+
+function dosomething_user_get_view($id) {
+  if (is_numeric($id)) {
```
angaither, line 120: i think we can remove this test, right? drups is probably overriding this return `user_view` anyways?

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -978,6 +1001,26 @@ function dosomething_user_get_user_by_email($email) {
...
+ *   The northstar id for which you are searching.
+ *
+ * @return string/bool
+ *   The user's id in phoenix, or false if no user was found.
+ */
+function dosomething_user_get_user_by_northstar_id($id) {
+  $query = db_select('field_data_field_northstar_id', 'f')
+          ->condition('f.field_northstar_id_value', $id)
+          ->fields('f', ['entity_id']);
+  $result = $query->execute()->fetchField();
```
angaither: so nitpicky, but I would save this var as `uid` so when you call `user_load` on it it doesn't look as ~wild~

katiecrane: good call, I definitely got a little ~wild~

lib/modules/dosomething/dosomething_user/dosomething_user.module
katiecrane, line 120: doesn't look like it's overriding it, if I remove it just get a blank page

angaither:
:+1:

PR #6298, opened 2016-03-21
Adds backlog cron

### LINKED ISSUE #6287: Cron to migrate people in the failed task log
#### User Story

As a user I want to be entered into a competition. 

When users don't get entered into the gladiator app, they are stored in the `dosomething_gladiator_failed_task_log`
We should have a check for that on each cron, and migrate users over who are still in that log.

Manually make post requests from people who are in the log.

itsjoekent:
#### What's this PR do?
- Adds cron to go through backlog table and resubmit people to the gladiators
- Moves this & previous gladiator cron work into functions so its separated out

start phoenix, than gladiator (if theyre already running vagrant halt that shiz or its gonna error out with ports)
setup some campaign to have a signup data form and submit to competitions
make a bad url in the gladiator module so it cant submit people
use chrome incognito on that campaign and signup for competition
check you've been placed in the table for people that failed joining competition
go into gladiator and make sure there is a contest that matches the nid/run_nid and is open for business
run cron on drupal
10 minutes later
check the user is no longer in the table
check the user is in the proper contest waiting room in gladiator

no

Fixes #6287

### DIFF
--- a/lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.cron.inc
+++ b/lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.cron.inc
@@ -4,6 +4,11 @@
  * Implements hook_cron()
  */
  function dosomething_gladiator_cron() {
+   dosomething_gladiator_update_campaigns();
+   dosomething_gladiator_retry_failed_users();
+ }
+
+ function dosomething_gladiator_update_campaigns() {
    $client = _dosomething_gladiator_build_http_client();
 
    $options = [
@@ -30,5 +35,25 @@
        ])
        ->condition('nid', $nid)
        ->execute();
+   }
+ }
+
+ function dosomething_gladiator_retry_failed_users() {
+   $task_log = db_select('dosomething_gladiator_failed_task_log', 't')
+    ->fields('t')
+    ->execute()
+    ->fetchAll();
+
+   foreach ($task_log as $task) {
+      $user = user_load($task->uid);
+      $values = [
+        'nid' => $task->nid,
+        'run_nid' => $task->run_nid,
+      ];
+      db_delete('dosomething_gladiator_failed_task_log')
+        ->condition('uid', $task->uid)
+        ->condition('run_nid', $task->run_nid)
+        ->execute();
+      dosomething_gladiator_send_user_to_gladiator($user, $values);
    }
  }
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.cron.inc
```diff
@@ -32,3 +37,23 @@
...
+    ->execute()
+    ->fetchAll();
+
+   foreach ($task_log as $task) {
+      $user = user_load($task->uid);
+      $values = [
+        'nid' => $task->nid,
+        'run_nid' => $task->run_nid,
+      ];
+      db_delete('dosomething_gladiator_failed_task_log')
```
angaither: we should wait to remove the failed logs after we get a `200` from gladiator

itsjoekent: if it fails again you just get put back in the log as a new row

angaither: oh duh.

itsjoekent: its ok i didnt realize it did this until i accidentally did it haha

angaither: oh, you could just [truncate](https://api.drupal.org/api/drupal/includes!database!database.inc/function/db_truncate/7) the whole table after grabbing everything instead of removing each row at a time

itsjoekent:
@angaither good to merge?

lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.cron.inc
itsjoekent: what if a user pops in after i started looping?

angaither: it can be done right after you fetch all rows from the table, not after the loop

itsjoekent: oh yeah that would work

itsjoekent:
added truncation. I think its good to go now @angaither

angaither:
:+1:

PR #6299, opened 2016-03-22
Log exported users

DFurnes:
#### What's this PR do?

:memo: Adds migrated users to the [Northstar request log](https://github.com/DoSomething/phoenix/pull/6262), so we can track if we run into any errors.

:1234: Store the returned Northstar ID to the Drupal profiles of migrated users for future use.

:boom: Fixes a crash I ran into when re-testing this where it was trying to migrate the "anonymous" user and crashing in a big ball of flames.

Feel free to pull down this branch and test against your local environment.

Noooooooope.

References #6260.

---

For review: @angaither

### DIFF
--- a/scripts/export-users-to-northstar.php
+++ b/scripts/export-users-to-northstar.php
@@ -24,7 +24,7 @@ foreach ($users as $user) {
   // Create json object
   $user = user_load($user->uid);
   $ns_user = build_northstar_user($user);
-  
+
   // Don't "forward" the anonymous user.
   if($user->uid == 0) {
     continue;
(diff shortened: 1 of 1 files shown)


scripts/export-users-to-northstar.php
```diff
@@ -25,13 +25,25 @@
   $user = user_load($user->uid);
   $ns_user = build_northstar_user($user);
 
+  // Don't "forward" the anonymous user.
+  if($user->uid == 0) {
```
angaither, line 29: instead of doing this, I set the `last_user_migrated` var to 0, either way

angaither:
just a minor comment, not needed for merge, but :+1:

scripts/export-users-to-northstar.php
DFurnes, line 29: Oh clever, I didn't think to do that.

PR #6305, opened 2016-03-22
adds entity translation patch

### LINKED ISSUE #5876: Add node status to translate tab
Currently the translate tab shows original translation as published, even if the node is not
Add node status to that tab, to make it less confusing

image
### Original report

**What is happening?**
When a campaign is created in x language, saved, and then the "translate" tab is viewed, the translation just created is noted as "published" even thought it was never published and 404s for anon users.

**What should be happening?**
Published statuses in the "translate" tab should reflect the actual published status...

**How can someone replicate it?**
Create a campaign, note the language. Save. Check the "translate" tab.

**When did it start/when did you notice it?**
I noticed it today.

chloealee:
#### What's this PR do?

Updates translation tabs to include the node status and changes published/not published to translated/not translated. 

As a logged in admin user, click on a campaign. 
At the top of the page, click "Translate"
Above the table, you should see "Node status:" and either "Published" if campaign is published or "Not published" if campaign is not yet published. 
"STATUS" column should only have "Translated" or "Not translated" values in column's rows. 

Fixes #5876

### DIFF
--- a/lib/profiles/dosomething/drupal-org.make
+++ b/lib/profiles/dosomething/drupal-org.make
@@ -65,6 +65,7 @@ projects[entity_autocomplete][subdir] = "contrib"
 ; Dev branch used: 2015-Aug-16
 projects[entity_translation][version] = "1.x-dev"
 projects[entity_translation][subdir] = "contrib"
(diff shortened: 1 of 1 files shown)


angaither:
@chloealee pix plx

chloealee:
screen shot 2016-03-22 at 4 30 06 pm

angaither:
cc @namimody

sergiitk:
:+1: The patch has also been reviewed by me.

PR #6311, opened 2016-03-23
Non admin users can see their own profiles

### LINKED ISSUE #6310: Non admin login is 404'ed
#### Current Behavior

After a non-admin logs in, the user is 404'ed
#### Desired Behavior

user is taken to profile
#### Steps to Replicate

log in as a non-admin user, or create a new account and try to view your profile

image

katiecrane:
#### What's this PR do?

Now using `user_view_access` for access callback. This is a special access function just for views that will allow users to see their own profiles, but no others. It also allows admins/other users that can see all profiles to do that too

Login as an admin and make sure you can see user profiles. Throw a northstar id in there just to make sure. As a normal user, make sure you can see your own profile and NOT other profiles.

This was caused by #6294 

Fixes #6310

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -110,8 +110,8 @@ function dosomething_user_menu_alter(&$items) {
   $items['user/%'] = array(
     'page callback' => 'dosomething_user_get_view',
     'page arguments' => array(1),
-    'access callback' => 'user_access',
-    'access arguments' => array('access user profiles'),
+    'access callback' => 'user_view_access',
+    'access arguments' => array(1),
   );
   return $items;
 }


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -110,8 +110,8 @@ function dosomething_user_menu_alter(&$items) {
   $items['user/%'] = array(
     'page callback' => 'dosomething_user_get_view',
     'page arguments' => array(1),
-    'access callback' => 'user_access',
-    'access arguments' => array('access user profiles'),
+    'access callback' => 'user_view_access',
```
angaither, line 113: hmm, does this expose more than we want though? can users see `edit` now instead of `edit info` for instance?

katiecrane, line 113: @angaither /user/{id}/edit gives them the normal update profile page and /user/{id}/info 404s

katiecrane, line 113: that's the normal behavior right?

angaither, line 113: @katiecrane sorry wait
https://staging.beta.dosomething.org/us/user/[id]/info should be the user info edit page

regular users should not see/have access to 
https://staging.beta.dosomething.org/us/user/[id]/edit

katiecrane, line 113: @angaither are you sure that's not the other way around? I'm seeing it the other way around on the live site and the same on my local

katiecrane, line 113: aka from user/[id] clicking "update my profile" goes to the /edit page. but it's not the same /edit page that an admin would see

angaither, line 113: I thought the info form was a [different path](https://github.com/DoSomething/phoenix/blob/8fc28b63d4657d6a71ce6d79a9b2e38a1a1a7085/lib/modules/dosomething/dosomething_user/dosomething_user.module#L89)

katiecrane, line 113: @angaither okay yeah info is a path that uses [these permissions](https://github.com/DoSomething/phoenix/blob/ef62285b87f37c914cc5a35a6a6fdf0efca94bd1/lib/modules/dosomething/dosomething_user/dosomething_user.features.user_permission.inc#L345) and I don't think anything we did should mess with that or /edit?

angaither:
:+1:

PR #6313, opened 2016-03-23
Adds Member Age to Profile

### LINKED ISSUE #6291: Adding Member Age to Profile
#### User Story

There are many occasions where a campaign lead needs a member's age for a wrap report, sponsor updates, the gala, conference, or social post. It would be great if member age could be part of their admin-facing profile.

This age will need to always consider the current date so it is not static.
#### Why This Matters

Will save campaign leads time.

A campaign lead does this: 
1. Go to their profile: https://www.dosomething.org/us/user/2319718
2. Go to edit info: https://www.dosomething.org/us/user/2319718/info
3. Grab their date of birth. 
4. Go to someplace like this: http://www.myagecalculator.com/

How feasible is this @angaither?

chloealee:
#### What's this PR do?

Adds member age to profile e.g. https://www.dosomething.org/us/user/2319718/info

Visit a user profile following the URL above. Next to birthday, you should see "User age:"
screen shot 2016-03-23 at 4 52 23 pm

Fixes #6291

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -1264,6 +1264,7 @@ function dosomething_user_non_staff_redirect() {
  * @see dosomething_user_menu()
  */
 function dosomething_user_info_form($form, &$form_state, $account) {
+
   $form['uid'] = array(
     '#type' => 'hidden',
     '#value' => $account->uid,
@@ -1328,11 +1329,15 @@ function dosomething_user_info_form($form, &$form_state, $account) {
     '#default_value' => $zipcode,
   );
   $birthday = dosomething_user_get_field_birthdate(dosomething_user_get_field('field_birthdate', $account), 'm/d/Y');
+  $birthDate = explode("/", $birthday);
+  $age = (date("md", date("U", mktime(0, 0, 0, $birthDate[0], $birthDate[1], $birthDate[2]))) > date("md") ? ((date("Y") - $birthDate[2]) - 1) : (date("Y") - $birthDate[2]));
+
   $form['info']['birthdate'] = array(
     '#title' => "Birthdate",
     '#type' => 'textfield',
     '#default_value' => $birthday,
     '#description' => t("Format: M/D/YYYY"),
+    '#field_suffix' => 'User age: ' . $age, 
   );
   $form['picture'] = array(
     '#type' => 'fieldset',
@@ -1356,7 +1361,6 @@ function dosomething_user_info_form($form, &$form_state, $account) {
   );
 
   $form['#validate'][] = 'dosomething_user_validate_address_field';
-
   return $form;
 }
 
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -1218,6 +1218,23 @@ function dosomething_user_format_string($string, $format) {
         return strtolower($string);
     }
 }
+/**
+ * Returns given user's age.
+ *
+ * @param string $field_value
+ *   The value of the field from `dosomething_user_get_field` to format.
+ * @param string $format
+ *   Optional- format to return. http://www.php.net/manual/en/function.date.php.
+ *   If NULL, the raw timestamp is returned.
+ *
+ * @return string
+ *   Returns the date in specified $format.
+ */
+function dosomething_user_get_age($birthdate) {
+  $today = new DateTime('now');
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -1328,11 +1329,15 @@ function dosomething_user_info_form($form, &$form_state, $account) {
     '#default_value' => $zipcode,
   );
   $birthday = dosomething_user_get_field_birthdate(dosomething_user_get_field('field_birthdate', $account), 'm/d/Y');
+  $birthDate = explode("/", $birthday);
+  $age = (date("md", date("U", mktime(0, 0, 0, $birthDate[0], $birthDate[1], $birthDate[2]))) > date("md") ? ((date("Y") - $birthDate[2]) - 1) : (date("Y") - $birthDate[2]));
+
```
chloealee: this was taken from stack overflow - should I include the link in a comment for credit?

chloealee: also,  should this be extracted into a new function in this file? Wasn't sure if it should be since it would be only two lines...

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -1328,11 +1329,15 @@ function dosomething_user_info_form($form, &$form_state, $account) {
     '#default_value' => $zipcode,
   );
   $birthday = dosomething_user_get_field_birthdate(dosomething_user_get_field('field_birthdate', $account), 'm/d/Y');
+  $birthDate = explode("/", $birthday);
+  $age = (date("md", date("U", mktime(0, 0, 0, $birthDate[0], $birthDate[1], $birthDate[2]))) > date("md") ? ((date("Y") - $birthDate[2]) - 1) : (date("Y") - $birthDate[2]));
```
angaither: yeah this looks pretty insane. you should be able to call `diff` on a DateTime object and just grab the value from `y`
like

```
today = new DateTime("now")
bday = new DateTime($birthday_value);
age = bday->diff(today)->y;
```

lib/modules/dosomething/dosomething_user/dosomething_user.module
angaither: it would probably be helpful for this to be a function on it's own. we have `dosomething_user_get_field_birthdate` we could use/modify that to also return age? or create a new function

angaither:
also note, I think this info form is exposed to all users, but we would want this only displayed for admins

chloealee:
@angaither I checked with a non admin account and wasn't able to hit this page - I _think_ it's only available to admins?

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -1328,11 +1346,14 @@ function dosomething_user_info_form($form, &$form_state, $account) {
...
   );
   $birthday = dosomething_user_get_field_birthdate(dosomething_user_get_field('field_birthdate', $account), 'm/d/Y');
+  $age = dosomething_user_get_age($birthday);
+
   $form['info']['birthdate'] = array(
     '#title' => "Birthdate",
     '#type' => 'textfield',
     '#default_value' => $birthday,
     '#description' => t("Format: M/D/YYYY"),
+    '#field_suffix' => 'User age: ' . $age,
```
chloealee, line 1352: @angaither do you think I should add `. ' years old.'` here so when it is posted on the user profile, it reads "User age: 20 years old." instead of "User age: 20" or is it explicit enough as is?

angaither, line 1352: this looks :ok:

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -1218,6 +1218,23 @@ function dosomething_user_format_string($string, $format) {
         return strtolower($string);
     }
 }
+/**
+ * Returns given user's age.
+ *
+ * @param string $field_value
```
angaither: the params & return looks like a copy/paste mistake

lib/modules/dosomething/dosomething_user/dosomething_user.module
chloealee, line 1352: k I'll leave as is!

lib/modules/dosomething/dosomething_user/dosomething_user.module
chloealee: totally right - fixed!

angaither:
:+1:

PR #6321, opened 2016-03-24
Build assets with Webpack

DFurnes:
#### What's this PR do?

This PR updates Phoenix to use the [shared Webpack config](https://github.com/DoSomething/webpack-config) and [ESLint config](https://github.com/DoSomething/eslint-config) that were extracted from Forge in DoSomething/forge#535. Because the Modernizr CLI & Webpack now handle building all of our front-end assets, this also removes Grunt as a build dependency.

If you want to manually test: pull down this branch, run `npm install` to get the new build dependencies, and then `npm start` to build & watch, or `npm run build` to build for production.

(The `ds grunt` alias that's used in our deploy script was also updated, although the command name is probably a bit suspect now!)

References DoSomething/forge#535.

---


bin/ds
```diff
@@ -447,9 +447,9 @@ function grunt {
     then
       echo -e "\n\e[4mBuilding SCSS/JS with Grunt, and watching for changes...\e[0m"
       echo -e "\e[30;43m(Press Ctrl-C to stop.)\e[0m\n"
-      eval "/usr/bin/grunt"
+      eval "/usr/bin/npm start"
```
weerd, line 446:
```
npm activate
```

```
npm commence
```

```
npm let'er rip
```

```
npm launch
```

```
npm initiate launch sequence
```

```
npm go
```

```
npm npmification
```

```
npm gotta catch em all
```

just tossing those out 😼

lib/themes/dosomething/paraneue_dosomething/scss/content/_finder.scss
```diff
@@ -213,7 +213,7 @@
   }
 
   .no-result {
-    background: $off-black asset-url("images/placeholder.png") center center;
+    background: $off-black center center;
```
weerd: why'd the default image get removed?

DFurnes: It [doesn't exist](https://github.com/DoSomething/phoenix/tree/dev/lib/themes/dosomething/paraneue_dosomething/images)! Since Webpack loads assets as well (to potentially inline as Data URIs), it was failing because the path didn't resolve to anything.

lib/themes/dosomething/paraneue_dosomething/scss/content/_win-module.scss
```diff
@@ -1,13 +1,11 @@
 .win-module {
   .win-module__progress {
     margin-bottom: $base-spacing;
-    background-image: asset-url("images/WinBackgroundS.svg");
```
angaither, line 4: rip win

DFurnes, line 4: Oh, although this one actually _does_ exist. We just don't have a loader configured for SVGs. Is the "win module" still a thing?

lib/themes/dosomething/paraneue_dosomething/scss/content/_finder.scss
weerd: So it's fine to use the shorthand CSS property without defining an image path?

angaither:
@DFurnes can you update the [wiki](https://github.com/DoSomething/phoenix/wiki/Front-End-Development) does `ds grunt --watch` still work? 
we should probs change the name if it's a :suspect:

lib/themes/dosomething/paraneue_dosomething/scss/content/_win-module.scss
angaither, line 4: I don't think so if this is winning from a hot module. we still have the code for hot module, but we aren't using it anymoar
but there's a win state on campaigns... dunno which this is.

DFurnes:
@angaither I've left it as `ds grunt` and `ds grunt --watch` just so we don't trip anyone up. I think it could make sense to rename to something more tool-agnostic like `ds front-end` or `ds assets` maybe?

angaither:
@DFurnes as long as documentation is updated, it should be fine. esp if it doesn't make sense to leave it named that
my 2 :moneybag:

weerd:
yeah, rename it! `ds assets` works for me but might be too generic... `ds fender`?

lib/themes/dosomething/paraneue_dosomething/scss/content/_finder.scss
DFurnes: I think so, although it doesn't really make any sense to set background positioning now. :confused:

DFurnes:
@angaither @weerd Updated! I left in compatibility with running `ds grunt` in case people have muscle-memory for that, but will update the docs once I merge! :smile:

lib/themes/dosomething/paraneue_dosomething/scss/content/_finder.scss
weerd: yar :\

bin/ds
```diff
@@ -644,9 +640,9 @@ fi
 #----------------------------------------------------------------------
 # grunt
 #----------------------------------------------------------------------
-if [[ $1 == "grunt" ]]
+if [[ $1 == "grunt" ]] || [[ $1 == "assets" ]]
```
weerd, line 643: :wink:

lib/themes/dosomething/paraneue_dosomething/scss/content/_finder.scss
```diff
@@ -213,7 +213,7 @@
   }
 
   .no-result {
-    background: $off-black asset-url("images/placeholder.png") center center;
+    background: $off-black;
```
weerd: Can we make this `background-color`? Would make it easy to customize adding to it if needed without having to override a bunch of the default properties the shorthand otherwise adds...

DFurnes: Good call! Done!

weerd:
:+1:

weerd:
:sheep: IT! :dancers:

PR #6322, opened 2016-03-25
Adds competition param to signup index

### LINKED ISSUE #6316: Update Signups endpoint for competition support
#### User Story

As, gladiatior? I want signup data to be returned in a more sustainable format. 
#### Additional Information

Would like to grab signups from a campaign & a specific run
Add params for including only signups where `competition = 1`
And the default should be to order by RB quantity 
#### Why This Matters

so we don't have to spend our lives waiting for pages to load. :older_man: :older_woman:

itsjoekent:
#### What's this PR do?

Adds competition parameter to the signup index endpoint, this does two things.
1- Only include signups that are marked as competition signups.
2- Order the signups by quantity.

Checkout `http://dev.dosomething.org:8888/api/v1/signups?competition=1`

nope

Fixes #6316

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
@@ -69,11 +69,11 @@ function _signup_resource_definition() {
           ],
           [
             'name' => 'competition',
-            'description' => 'Only competition signups, defaults to 0.',
+            'description' => 'Only competition signups, defaults to false.',
             'optional' => TRUE,
-            'type' => 'int',
+            'type' => 'boolean',
             'source' => ['param' => 'competition'],
-            'default value' => 0,
+            'default value' => FALSE,
           ],
         ],
         'access callback' => '_signup_resource_access',
@@ -96,7 +96,7 @@ function _signup_resource_access() {
  * @param string $campaigns
  * @param int $count
  * @param int $page
- * @param int $competition
+ * @param boolean $competition
  * @return array
  */
 function _signup_resource_index($user, $campaigns, $count, $page, $competition) {
--- a/lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
@@ -67,6 +67,14 @@ function _signup_resource_definition() {
             'source' => ['param' => 'page'],
             'default value' => 1,
           ],
+          [
+            'name' => 'competition',
+            'description' => 'Only competition signups, defaults to 0.',
+            'optional' => TRUE,
+            'type' => 'int',
+            'source' => ['param' => 'competition'],
+            'default value' => 0,
+          ],
         ],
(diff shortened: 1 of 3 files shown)


weerd:
Just fyi, don't use integers in the examples or what gets passed as params. More obvious to use `true/false` booleans.

itsjoekent:
gotcha!

lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
```diff
@@ -67,6 +67,14 @@ function _signup_resource_definition() {
             'source' => ['param' => 'page'],
             'default value' => 1,
           ],
+          [
+            'name' => 'competition',
+            'description' => 'Only competition signups, defaults to 0.',
```
weerd: `defaults to false.`

itsjoekent: true false true false

lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
```diff
@@ -67,6 +67,14 @@ function _signup_resource_definition() {
...
+            'optional' => TRUE,
+            'type' => 'int',
```
weerd: `boolean`

lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
```diff
@@ -67,6 +67,14 @@ function _signup_resource_definition() {
...
+            'source' => ['param' => 'competition'],
+            'default value' => 0,
```
weerd: `FALSE`... just indicated the changes based on my comment regarding using booleans instead :smile:

angaither:
@deadlybutter does this play nice with other params? can we request competition signups for just one campaign?

angaither:
ahh, sorry hit close instead of comment!

itsjoekent:
SHAME.

itsjoekent:
@angaither let me give that a try actually, but i think it should

chloealee:
:+1: from me!

itsjoekent:
can confirm it works with the other params @angaither also @weerd updated your stuffs

weerd:
K. Rechecking :dancer:

lib/modules/dosomething/dosomething_signup/includes/SignupTransformer.php
```diff
@@ -111,6 +111,7 @@ private function setFilters($parameters) {
       'campaigns' => dosomething_helpers_format_data($parameters['campaigns']),
       'count' => (int) $parameters['count'] ?: 25,
       'page' => (int) $parameters['page'],
+      'competition' => (int) $parameters['competition'] ?: NULL,
```
weerd: No need to cast to `(int)` here.

itsjoekent:
already fixed @weerd haha

weerd:
Cool! :+1:

PR #6324, opened 2016-03-25
Adds northstar id to signups api response

itsjoekent:
#### What's this PR do?

adds northstar user id to the signups response so gladiator can use it to do magical things

competitions=true means only return signups that are for a competition, and order them by reportback quantity.

`http://dev.dosomething.org:8888/api/v1/signups?competition=true`

:cake: 

Fixes #nope

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -427,7 +427,6 @@ abstract class Transformer {
     return [
       'id' => $data->id,
       'created_at' => $data->created_at,
-      'northstar_id' => $data->northstar_id,
       'campaign_run' => [
         'id' => $data->campaign_run,
         'current' => $current,
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -427,6 +427,7 @@ abstract class Transformer {
     return [
       'id' => $data->id,
       'created_at' => $data->created_at,
+      'northstar_id' => $data->northstar_id,
       'campaign_run' => [
         'id' => $data->campaign_run,
         'current' => $current,
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -445,7 +445,8 @@ abstract class Transformer {
    */
   protected function transformUser($data) {
     return [
-      'id' => $data->id,
+      'id' => $data->drupal_id,
+      'northstar_id' => $data->northstar_id,
       'first_name' => $data->first_name,
       'last_initial' => $data->last_initial,
       'photo' => $data->photo,
--- a/lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
+++ b/lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
@@ -170,11 +170,11 @@ class Reportback extends Entity {
 
     $this->user = [
       'drupal_id' => $data->uid,
-      'id' => dosomething_helpers_isset($northstar_user, 'id'),
-      'first_name' => dosomething_helpers_isset($northstar_user, 'first_name'),
-      'last_initial' => dosomething_helpers_isset($northstar_user, 'last_initial'),
-      'photo' => dosomething_helpers_isset($northstar_user, 'photo'),
-      'country' => dosomething_helpers_isset($northstar_user, 'country'),
+      'northstar_id' => dosomething_helpers_isset($northstar_user->id, 'id'),
+      'first_name' => dosomething_helpers_isset($northstar_user->first_name, 'first_name'),
+      'last_initial' => dosomething_helpers_isset($northstar_user->last_initial, 'last_initial'),
+      'photo' => dosomething_helpers_isset($northstar_user->photo, 'photo'),
+      'country' => dosomething_helpers_isset($northstar_user->country, 'country'),
     ];
   }
 
--- a/lib/modules/dosomething/dosomething_reportback/includes/ReportbackItem.php
+++ b/lib/modules/dosomething/dosomething_reportback/includes/ReportbackItem.php
@@ -120,11 +120,11 @@ class ReportbackItem extends Entity {
 
     $this->user = [
       'drupal_id' => $data->uid,
-      'id' => dosomething_helpers_isset($northstar_user, 'id'),
-      'first_name' => dosomething_helpers_isset($northstar_user, 'first_name'),
-      'last_initial' => dosomething_helpers_isset($northstar_user, 'last_initial'),
-      'photo' => dosomething_helpers_isset($northstar_user, 'photo'),
-      'country' => dosomething_helpers_isset($northstar_user, 'country'),
+      'northstar_id' => dosomething_helpers_isset($northstar_user->id, 'id'),
+      'first_name' => dosomething_helpers_isset($northstar_user->first_name, 'first_name'),
+      'last_initial' => dosomething_helpers_isset($northstar_user->last_initial, 'last_initial'),
+      'photo' => dosomething_helpers_isset($northstar_user->photo, 'photo'),
+      'country' => dosomething_helpers_isset($northstar_user->country, 'country'),
     ];
   }
 
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
@@ -54,8 +54,10 @@ function dosomething_signup_get_signups_query($params, $tally = FALSE) {
 function dosomething_signup_build_signups_query($params = []) {
   $query = db_select('dosomething_signup', 's');
   $query->leftJoin('dosomething_reportback', 'rb', 's.uid = rb.uid and s.nid = rb.nid and s.run_nid = rb.run_nid');
+  $query->leftJoin('field_data_field_northstar_id', 'ns', 'ns.entity_id = s.uid');
   $query->fields('s', ['sid', 'uid', 'nid', 'run_nid', 'timestamp']);
   $query->fields('rb', ['rbid']);
+  $query->fields('ns', ['field_northstar_id_value']);
 
   if (isset($params['sid'])) {
     if (is_array($params['sid'])) {
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
@@ -71,6 +71,8 @@ function dosomething_signup_build_signups_query($params = []) {
 
   if (isset($params['competition'])) {
     $query->condition('s.competition', 1, '=');
+    $query->leftJoin('field_data_field_northstar_id', 'ns', 'ns.entity_id = s.uid');
+    $query->fields('ns', ['field_northstar_id_value']);
     $query->orderBy('rb.quantity', 'DESC');
   }
 
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
@@ -54,8 +54,10 @@ function dosomething_signup_get_signups_query($params, $tally = FALSE) {
 function dosomething_signup_build_signups_query($params = []) {
   $query = db_select('dosomething_signup', 's');
   $query->leftJoin('dosomething_reportback', 'rb', 's.uid = rb.uid and s.nid = rb.nid and s.run_nid = rb.run_nid');
+  $query->leftJoin('field_data_field_northstar_id', 'ns', 'ns.entity_id = s.uid');
   $query->fields('s', ['sid', 'uid', 'nid', 'run_nid', 'timestamp']);
   $query->fields('rb', ['rbid']);
+  $query->fields('ns', ['field_northstar_id_value']);
 
   if (isset($params['sid'])) {
     if (is_array($params['sid'])) {
@@ -71,8 +73,6 @@ function dosomething_signup_build_signups_query($params = []) {
 
   if (isset($params['competition'])) {
     $query->condition('s.competition', 1, '=');
-    $query->leftJoin('field_data_field_northstar_id', 'ns', 'ns.entity_id = s.uid');
-    $query->fields('ns', ['field_northstar_id_value']);
     $query->orderBy('rb.quantity', 'DESC');
   }
 
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
@@ -54,10 +54,8 @@ function dosomething_signup_get_signups_query($params, $tally = FALSE) {
 function dosomething_signup_build_signups_query($params = []) {
   $query = db_select('dosomething_signup', 's');
   $query->leftJoin('dosomething_reportback', 'rb', 's.uid = rb.uid and s.nid = rb.nid and s.run_nid = rb.run_nid');
-  $query->leftJoin('field_data_field_northstar_id', 'ns', 'ns.entity_id = s.uid');
   $query->fields('s', ['sid', 'uid', 'nid', 'run_nid', 'timestamp']);
   $query->fields('rb', ['rbid']);
-  $query->fields('ns', ['field_northstar_id_value']);
 
   if (isset($params['sid'])) {
     if (is_array($params['sid'])) {
--- a/lib/modules/dosomething/dosomething_signup/includes/Signup.php
+++ b/lib/modules/dosomething/dosomething_signup/includes/Signup.php
@@ -92,7 +92,7 @@ class Signup extends Entity {
     $user = user_load($data->uid);
     $this->user = [
       'drupal_id' => $data->uid,
-      'id' => dosomething_helpers_isset($user->uuid, 'id'),
+      'northstar_id' => dosomething_helpers_isset($user->uuid, 'id'),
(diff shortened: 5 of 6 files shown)


lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -88,6 +88,8 @@ public static function find(array $filters = []) {
   private function build($data) {
     $this->id = $data->sid;
     $this->created_at = $data->timestamp;
+    $this->northstar_id = $data->field_northstar_id_value;
```
angaither: hmm, does this work if you're asking for signups w/o a competition?

itsjoekent:
@angaither updated

itsjoekent:
happy monday! is this PR good to merge?

lib/modules/dosomething/dosomething_signup/includes/Signup.php
itsjoekent: for context, this comment was made because before the northstar_id was only added with the competition parameter, but we decided to make it always added (and thus this isnt a problem anymore)

lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -427,6 +427,7 @@ protected function transformSignup($data, $current) {
     return [
       'id' => $data->id,
       'created_at' => $data->created_at,
+      'northstar_id' => $data->northstar_id,
```
weerd: So this isn't necessarily a data property that should be on the signup object. Placing it here for convenience is a bad precedent to set. I would say that this `northstar_id` should be a property of the user object and not of the signup. The `user` object can be nested into the `signup` object in the `SignupTransformer`, which would make more sense since the signup is associated with a particular user. Plus, then the NS id would be a regular property on the user object.

DFurnes: +1, this is definitely confusing to have on the signup object.

lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
```diff
@@ -54,8 +54,10 @@ function dosomething_signup_get_signups_query($params, $tally = FALSE) {
 function dosomething_signup_build_signups_query($params = []) {
   $query = db_select('dosomething_signup', 's');
   $query->leftJoin('dosomething_reportback', 'rb', 's.uid = rb.uid and s.nid = rb.nid and s.run_nid = rb.run_nid');
+  $query->leftJoin('field_data_field_northstar_id', 'ns', 'ns.entity_id = s.uid');
```
DFurnes, line 57: Just as a heads up, this field is not guaranteed to be populated for every account. We're currently running a script that's setting it on each user, but it won't be done for a while yet.

itsjoekent, line 57: @angaither is it even worth using this to solve our problem in gladiator then?

angaither, line 57: @deadlybutter yes

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -88,6 +88,22 @@ public static function find(array $filters = []) {
   private function build($data) {
     $this->id = $data->sid;
     $this->created_at = $data->timestamp;
+
+    $northstar_response = dosomething_northstar_get_northstar_user($data->uid);
```
angaither: this seems super wild to me. why are we pinging NS for every user signup?

itsjoekent: cause thats how we do it for reportbacks & reportback items https://github.com/DoSomething/phoenix/blob/09a6a60ae359882228bfed9b6b01cb8bb97f623e/lib/modules/dosomething/dosomething_reportback/includes/Reportback.php#L147

angaither: That's only if the full user is needed, but all we need here is the ID, correct?

weerd:
update the description... you changed it to be a boolean right? ----> `http://dev.dosomething.org:8888/api/v1/signups?competition=1`

lib/modules/dosomething/dosomething_signup/includes/Signup.php
weerd: yeah, if all we need is the NS id and that's available in Drupal, lets not go to NS for that each time @_@ Add it to the query in the signups as something to grab along w/ the singup info so you have it in the results as you're parsing through them in the `build()`

itsjoekent:
@weerd im not sure what youre suggesting, that is the URL in the description?

itsjoekent:
also @weerd & @angaither i removed the northstar use, just loading user from db now

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -88,6 +88,17 @@ public static function find(array $filters = []) {
   private function build($data) {
     $this->id = $data->sid;
     $this->created_at = $data->timestamp;
+
+    $user = user_load($data->uid);
+    $this->user = [
+      'drupal_id' => $data->uid,
```
angaither, line 95: do we say `drupal_id` in northstar or `phoenix_id` and `northstar_id` ? 
cc @DFurnes

DFurnes, line 95: We've been using `drupal_id` for this in Northstar, and I think it'd be a bit of an ordeal to change now. I'd definitely been trying to preference using "Phoenix" whenever possible otherwise though.

itsjoekent, line 95: so drupal_id and northstar_id ?

DFurnes, line 95: @deadlybutter Yeah, I think that's good.

DFurnes, line 95: Hopefully :v: :twisted_rightwards_arrows: we can start to deprecate using Phoenix IDs sometime in the near-ish future in favor of Northstar IDs everywhere though.

itsjoekent:
alrighty changed the names @DFurnes @angaither

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -88,6 +88,17 @@ public static function find(array $filters = []) {
...
+      'northstar_id' => dosomething_helpers_isset($user->uuid, 'id'),
```
weerd: still not `uuid` @deadlybutter haha. You :heart: this whole `uuid` thing for some reason :wink:

itsjoekent: why do i keep doing that.

weerd: `¯\_(ツ)_/¯`

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -88,6 +88,17 @@ public static function find(array $filters = []) {
...
+      'first_name' => dosomething_helpers_isset($user->field_first_name[LANGUAGE_NONE][0]['value'], 'first_name'),
+      'last_initial' => dosomething_helpers_isset($user->field_last_name[LANGUAGE_NONE][0]['value'], 'last_initial'),
+      'photo' => dosomething_helpers_isset($user->photo, 'photo'),
+      'country' => dosomething_helpers_isset($user->field_address[LANGUAGE_NONE][0]['country'], 'country'),
```
weerd, line 100: you should be able to pass the field first through `dosomething_helpers_extract_field_data()` which deals with language stuff w/out hardcoding it, and then run the `isset(...)` around it :)

itsjoekent:
something i didnt think about, how many apps are going to break now that im chaging the transformUser function to use drupal & northstar _id instead of id

DFurnes:
@deadlybutter Nothing should break, since you're only adding additional fields here... right? We should definitely try to avoid changing any existing fields that client applications would expect to see.

itsjoekent:
well, you guys suggested I use drupal_id which is not what it looks like right now in the transform user function 

https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_api/includes/Transformer.php#L446

DFurnes:
Ah, gotcha. Sorry, I think I misunderstood what code we were discussing.

Let's definitely _not_ change that – we can keep `id` for the Drupal ID so we don't break the existing "contract" and then add an additional `northstar_id` field to that transformer for the Northstar ID.

itsjoekent:
gotcha, ill do that then

lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -445,7 +445,8 @@ protected function transformSignup($data, $current) {
    */
   protected function transformUser($data) {
     return [
-      'id' => $data->id,
+      'id' => $data->drupal_id,
```
weerd, line 448: So this might actually cause a bunch of problems. A while back, we decided that in the API response `id` would be the NS id and we would not expose the Drupal ID since we're moving away from from Drupal IDs as NS takes over. So I think this might mess a few things up (particularly with the `/reportback` and `/reportback-items` endpoints when passing the [`load_user=true`](https://github.com/DoSomething/phoenix/wiki/API#reportbacks) param that returns the NS user and populates the prior `$data->id`... I'm worried this might break things already in play (ie: mobile app)

weerd, line 448: What we could do is add a `legacy_id` property and populate the Drupal one there and leave the NS one as just `id` like it was before...
@DFurnes thoughts?

DFurnes, line 448: @weerd Oh, is `$data->id` already the Northstar ID in this case? Yeah, let's not change that then.

DFurnes, line 448: Its not my favorite naming, but I'd probably vote for using `drupal_id` (over `legacy_id`) for the Drupal ID just for consistency with how we refer to it elsewhere. And then just make sure to document that it's not a preferred way to identify a user & may be removed in the future.

itsjoekent, line 448: @DFurnes @weerd so this?

```
      'id' => $data->northstar_id,
      'drupal_id' => $data->drupal_id,
```

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -95,8 +95,8 @@ private function build($data) {
       'drupal_id' => $data->uid,
       'northstar_id' => dosomething_helpers_isset($data->field_northstar_id_value, 'id'),
       'first_name' => dosomething_helpers_isset(dosomething_helpers_extract_field_data($user->field_first_name), 'first_name'),
-      'last_initial' => dosomething_helpers_isset($user->field_last_name[LANGUAGE_NONE][0]['value'], 'last_initial'),
-      'photo' => dosomething_helpers_isset($user->photo, 'photo'),
+      'last_initial' => dosomething_helpers_isset(dosomething_helpers_extract_field_data($user->field_last_name), 'last_initial'),
+      'photo' => dosomething_helpers_isset(dosomething_helpers_extract_field_data($user->photo), 'photo'),
```
DFurnes, line 99: The `dosomething_helpers_extract_field_data` [should check if the field is set](https://github.com/DoSomething/phoenix/blob/b280f55934058d7b8a4d599acf877be2cfae880a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module#L161-L162) so I think this could be simplified:

``` php
'photo' => dosomething_helpers_extract_field_data($user->photo),
```

lib/modules/dosomething/dosomething_api/includes/Transformer.php
DFurnes, line 448: @deadlybutter Yes, although it looks like the Northstar ID is stored under `$data->id`.

lib/modules/dosomething/dosomething_signup/includes/Signup.php
DFurnes, line 99: Oh, never mind. That's totally wrong. :no_good:

lib/modules/dosomething/dosomething_api/includes/Transformer.php
weerd, line 448: Yep, `$data->id` should already be the NS id, but doesn't really get populated unless the `load_user` query param is added and set to `true`. However, it only really applies to the Reportback endpoint(s). @deadlybutter this shouldn't drastically change your PR I don't think... are you actually using the Drupal ID or were you just adding to differentiate between the two?

itsjoekent, line 448: No its $data->northstar_id now because it was confusing me what ID's were what. See the changes i made: https://github.com/DoSomething/phoenix/pull/6324/files#diff-d0a29b0a298822b8c127c0c3ce12f46dR123

itsjoekent, line 448: @weerd Im adding northstar ID because I needed that for signups, which I was not getting before.

Im adding drupal ID because it was confusing  what "Id" was

weerd, line 448: Leave the drupal one out if we don't actually need it. No need to add that param if no one is using it and eventually will be deprecated anyway.

lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
```diff
@@ -170,11 +170,11 @@ private function build($data, $full = false) {
...
-      'id' => dosomething_helpers_isset($northstar_user, 'id'),
-      'first_name' => dosomething_helpers_isset($northstar_user, 'first_name'),
-      'last_initial' => dosomething_helpers_isset($northstar_user, 'last_initial'),
-      'photo' => dosomething_helpers_isset($northstar_user, 'photo'),
-      'country' => dosomething_helpers_isset($northstar_user, 'country'),
+      'northstar_id' => dosomething_helpers_isset($northstar_user->id, 'id'),
+      'first_name' => dosomething_helpers_isset($northstar_user->first_name, 'first_name'),
+      'last_initial' => dosomething_helpers_isset($northstar_user->last_initial, 'last_initial'),
+      'photo' => dosomething_helpers_isset($northstar_user->photo, 'photo'),
+      'country' => dosomething_helpers_isset($northstar_user->country, 'country'),
```
weerd, line 177: These should be breaking... for the `dosomething_helpers_isset()`, if it's an object, you pass it in and then indicate the property as a string, like `dosomething_helpers_isset($northstar_user, 'first_name')`... you don't pass the object and arrow property `$northstar_user-> first_name`

lib/modules/dosomething/dosomething_reportback/includes/ReportbackItem.php
```diff
@@ -120,11 +120,11 @@ private function build($data, $full = false) {
...
-      'id' => dosomething_helpers_isset($northstar_user, 'id'),
-      'first_name' => dosomething_helpers_isset($northstar_user, 'first_name'),
-      'last_initial' => dosomething_helpers_isset($northstar_user, 'last_initial'),
-      'photo' => dosomething_helpers_isset($northstar_user, 'photo'),
-      'country' => dosomething_helpers_isset($northstar_user, 'country'),
+      'northstar_id' => dosomething_helpers_isset($northstar_user->id, 'id'),
+      'first_name' => dosomething_helpers_isset($northstar_user->first_name, 'first_name'),
+      'last_initial' => dosomething_helpers_isset($northstar_user->last_initial, 'last_initial'),
+      'photo' => dosomething_helpers_isset($northstar_user->photo, 'photo'),
+      'country' => dosomething_helpers_isset($northstar_user->country, 'country'),
```
weerd, line 127: Same as above for these.

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -88,6 +88,18 @@ public static function find(array $filters = []) {
...
+
+    $user = user_load($data->uid);
+
+    $this->user = [
+      'drupal_id' => $data->uid,
+      'northstar_id' => dosomething_helpers_isset($data->field_northstar_id_value, 'id'),
+      'first_name' => dosomething_helpers_isset(dosomething_helpers_extract_field_data($user->field_first_name), 'first_name'),
+      'last_initial' => dosomething_helpers_isset(dosomething_helpers_extract_field_data($user->field_last_name), 'last_initial'),
+      'photo' => dosomething_helpers_isset(dosomething_helpers_extract_field_data($user->photo), 'photo'),
+      'country' => dosomething_helpers_isset($user->field_address[LANGUAGE_NONE][0]['country'], 'country'),
```
weerd, line 100: @deadlybutter did the `dosomething_helpers_extract_field_data()` not work because specifically looking for `country`?

itsjoekent:
im just gonna restart this entire PR its a mess

lib/modules/dosomething/dosomething_signup/includes/Signup.php
weerd, line 100: Also just to note here, not sure if it's an issue or not, but we're grabbing `user` object data from DS here, but elsewhere in the other endpoints where user data is populated, it's always from NS. I know the point of this update is to make the NS id available without doing more calls to NS. Thoughts on maybe just including the NS id and saying null to everything thing else otherwise? Alternatively, a future update could be a tiny UserResolver class or something that can check if NS has the data and if not then check Drupal and if not, then `null`. Maybe it allows passing a boolean so that we tell it specifically to _not_ check NS if we want to keep the search localized to just in Drupal too... just some ramblings and thoughts.

PR #6329, opened 2016-03-28
Address validation

### LINKED ISSUE #6317: Address validation triggered when not required on user profile edit form.
#### Current Behavior
- When hitting attempting to save the edit view of the user profile (e.g.  https://www.dosomething.org/us/user/2319718/info), address validation is triggered even though address is not required.

screen shot 2016-03-24 at 3 11 45 pm
#### Desired Behavior
- User should be able to save without inputting address.
- Address validation should only be triggered when an address is present. 
#### Why This Matters
- Doesn't allow for user profile to be edited.

chloealee:
#### What's this PR do?

Fixes bug on `/user/:uid/info` form that validates address correctly and doesn't throw an error if address is not present (since it is not required). 

Visit  `/user/:uid/info` and:
- delete address - should save profile. 
- add address and input `z` as the zipcode - should throw error. 

Fixes #6317

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -284,7 +284,7 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
       // Add extra validation on the edit page, for addresses.
       $current_page = $_SERVER['REQUEST_URI'];
       if (preg_match('/user\/([0-9]+)\/edit/', $current_page) && $validate_address) {
-        $form['#validate'][] = 'dosomething_user_validate_address_field';
+        $form['#validate'][] = 'dosomething_user_validate_edit_form_address';
       }
 
 
@@ -421,7 +421,7 @@ function dosomething_user_validate_edit_form_address ($form, &$form_state) {
   $last_name = $form_state['input']['field_last_name'][LANGUAGE_NONE][0]['value'];
   $address = $form_state['input']['field_address'][LANGUAGE_NONE][0];
 
-  dosomething_user_validate_address_field($first_name, $last_name, $address);
+  dosomething_user_validate_address_field($first_name, $last_name, $address, $form, $form_state);
 }
 
 /**
@@ -449,7 +449,12 @@ function dosomething_user_validate_address_field($first_name, $last_name, $addre
     }
     // We have a full address, save it!
     else {
-      dosomething_user_set_address_values_from_info_form($form, $form_state, $formatted_address);
+      if ($form['#form_id'] === 'dosomething_user_info_form') {
+        dosomething_user_set_address_values_from_info_form($form, $form_state, $formatted_address);
+      }
+      else {
+        dosomething_user_set_address_values_from_edit_form($form, $form_state, $formatted_address);
+      }
     }
   }
 }
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -284,7 +284,7 @@ function dosomething_user_form_alter(&$form, $form_state, $form_id) {
       // Add extra validation on the edit page, for addresses.
       $current_page = $_SERVER['REQUEST_URI'];
       if (preg_match('/user\/([0-9]+)\/edit/', $current_page) && $validate_address) {
-        $form['#validate'][] = 'dosomething_user_validate_edit_form_address';
+        $form['#validate'][] = 'dosomething_user_validate_address_field';
       }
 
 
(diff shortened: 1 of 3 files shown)


angaither:
@chloealee I think we already handled the two different types of address in [here](https://github.com/DoSomething/phoenix/blob/5896b2823f770c6aa3a1e9cf46d474ef4ae60564/lib/modules/dosomething/dosomething_user/dosomething_user_valid_address.inc#L43)

chloealee:
@angaither should I change this line to use that callback to handle the two different types of addresses here? Looks like right now, it just grabs the form on the page (`'page callback' => 'drupal_get_form'`).

chloealee:
@angaither hmm I tried that and it didn't seem to work, I'm not sure I understand how to incorporate that callback to this scenario?

angaither:
@chloealee you shouldn't have to change these to handle both the `form_address_field` and the manually created `user_address` without all the language none stuff
I would say you can take a look at how addresses are validated in the shipment form, or the signup data form

you can
1. update the `dosomething_user_validate_user_address` function to handle both JS and form submissions
OR
2. create your own `validation` function that is a mix of setting form errors, but using the field the way the `user_address` is used instead of the LANGUAGE_NONE stuff
OR 
3. refactor what already exists to DRY up the logic out that's the same across both of these things

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -429,11 +433,55 @@ function dosomething_user_validate_address_field($form, &$form_state) {
...
     // We have a full address, save it!
     else {
-      dosomething_user_set_address_values($form, $form_state, $formatted_address);
+      if ($form['#form_id'] === 'dosomething_user_info_form') {
+        dosomething_user_set_address_values_from_info_form($form, $form_state, $formatted_address);
+      }
+      else {
+        dosomething_user_set_address_values_from_edit_form($form, $form_state, $formatted_address);
+      }
     }
```
chloealee, line 433: @angaither instead of this if block, I tried to just `return TRUE;` and then write 

```
if (function dosomething_user_validate_address_field($first_name, $last_name, $address, $form, &$form_state)) {
  dosomething_user_set_address_values_from_edit_form($form, $form_state, $formatted_address);
}
```

as the last statement in `function dosomething_user_validate_edit_form_address` (and similar for `function dosomething_user_validate_info_form_address`) but when I did that, it for some reason wouldn't save the address in the address fields - it would successfully validate but would erase these fields and leave them blank on save. 

Tried this way instead to get it to work but not sure your thoughts on it - lmk!

chloealee:
@angaither also found another bug with the edit form - it looks like it's not sending the updated address to northstar anymore. Should I create a new issue for this/check with the mobile app team if they've noticed any bugs with this?

lib/modules/dosomething/dosomething_user/dosomething_user.module
angaither, line 433: @chloealee I'm having a hard time reading this diff, but this seems like a confusing approach. 
You should be able to leave the `address_field` validation the way it was as that sets and validates correctly 

Since the info form is a different type of field that we build manually you can use either the [menu hook](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_user/dosomething_user.module#L65) to validate the way [JS does](https://github.com/DoSomething/phoenix/blob/5896b2823f770c6aa3a1e9cf46d474ef4ae60564/lib/themes/dosomething/paraneue_dosomething/js/validators/address.js#L125), or create a new validation to handle the manually created fields that sets form errors and saves the fields

but you shouldn't touch the validation that validates the actually address field with all the `LANGUAGE_NONE` fields

chloealee, line 433: @angaither I'm not sure I quite understand how to use the menu hook to validate the way JS does so tried to go the second route you suggested of creating a new validation to handle the manually created fields. To create this new validation, I noticed that I would use a lot of the same code that is in `dosomething_user_validate_address_field` so I just created two new functions to set the different fields in the info vs. edit form and then send all to `dosomething_user_validate_address_field` to actually do the work of setting form errors and saving the fields. I _think_ this follows the route of creating a new validation but could be completely misunderstanding what you're saying?

angaither, line 433: @chloealee yeah, but `dosomething_user_validate_address_field` is meant to deal with the actual `address_field` type and you have to do a lot of work to turn the manually created form items into the type to match the address field format, which is probably why it isn't saving/updating

chloealee, line 433: @angaither it works as is right now to both save and update the form so when a user submits, the validated address is populated in the fields.

chloealee, line 433: @angaither I can totally change though to go about it a different route! Just not sure/don't understand how to follow the menu hook to validate the way JS does and I thought going about it this way was what you meant about creating a new validation/re-factoring?

chloealee:
@angaither separated the two functions out again - ready for another review!

angaither:
approach looks good, but you've got quite a few extra spaces/files changed that look like they shouldn't be included in this. please clean that up

chloealee:
all set!

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -410,8 +410,8 @@ function dosomething_user_add_validation_attributes_to_address_fields($form, &$f
  *
  */
 function dosomething_user_validate_address_field($form, &$form_state) {
-  $first_name = $form_state['input']['field_first_name'][LANGUAGE_NONE][0]['value'];
-  $last_name = $form_state['input']['field_last_name'][LANGUAGE_NONE][0]['value'];
+  $first_name = $form_state['input']['field_first_name'][LANGUAGE_NONE][0]['value'];
```
angaither: gah, extra spaces. do you have the [`.editorconfig`](https://github.com/DoSomething/phoenix/blob/dev/.editorconfig) setup? these should automatically be trimmed

chloealee: ah sorry I missed that! I should have the `.editorconfig` setup - I think @DFurnes set that up for me last week so not sure why it did that? Maybe because I just copied and pasted?

angaither:
:+1:

PR #6333, opened 2016-03-29
updating  field collection to beta version 11

### LINKED ISSUE #6332: Deploys are failing b/c of filed collection
#### Current Behavior

field collection deploy causes and error 

```
Fatal error: Cannot redeclare field_collection_update_7007() (previously declared in /var/www/staging.beta.dosomething.org/releases/20160329135241/html/profiles/dosomething/modules/contrib/field_collection/field_collection.install:336) in 
```

angaither:
#### What's this PR do?

Lock field collection module to `1.0-beta11` version 

run `ds build` and check that everything looks :ok: and there's no data loss
### Any background context

The patch we are using to support entity translation and the dev branch of this module are not compatible

Fixes #6332

### DIFF
--- a/lib/profiles/dosomething/drupal-org.make
+++ b/lib/profiles/dosomething/drupal-org.make
@@ -80,9 +80,6 @@ projects[features][version] = "2.2"
 projects[features][subdir] = "contrib"
 
 ; Field Collection
-; Updated to dev to be compatible with latest field_collection patch.
-; Dev branch used: 2015-Mar-26
-; !! Users are cautioned that data loss has been reported – please test carefully !!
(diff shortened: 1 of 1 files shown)


lib/profiles/dosomething/drupal-org.make
```diff
@@ -83,7 +83,7 @@ projects[features][subdir] = "contrib"
 ; Updated to dev to be compatible with latest field_collection patch.
 ; Dev branch used: 2015-Mar-26
```
sergiitk: Let's remove this comment then. And the one below.

sergiitk:
:+1:

PR #6335, opened 2016-03-29
Signups API modifications TAKE 2!

### LINKED ISSUE #6316: Update Signups endpoint for competition support

itsjoekent:
#### What's this PR do?
- Allows you to request multiple signups by ID
- Adds a plural 'users' param because Diego likes plural things.

Ive been using this locally
`http://dev.dosomething.org:8888/api/v1/signups?competition=true&users=1705752,1700226&campaigns=1485`

:cake: 

Fixes #6316

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
@@ -45,7 +45,7 @@ function _signup_resource_definition() {
           ],
           [
             'name' => 'users',
-            'description' => 'The id of the specified user to get signups for.',
+            'description' => 'The ids of the specified users to get signups for.',
             'optional' => TRUE,
             'type' => 'string',
             'source' => ['param' => 'users'],
--- a/lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
@@ -44,6 +44,14 @@ function _signup_resource_definition() {
             'default value' => NULL,
           ],
           [
+            'name' => 'users',
+            'description' => 'The id of the specified user to get signups for.',
+            'optional' => TRUE,
+            'type' => 'string',
+            'source' => ['param' => 'users'],
+            'default value' => NULL,
+          ],
+          [
             'name' => 'campaigns',
             'description' => 'The ids of the specified campaigns to get signups for.',
             'optional' => TRUE,
@@ -99,7 +107,12 @@ function _signup_resource_access() {
  * @param boolean $competition
  * @return array
  */
-function _signup_resource_index($user, $campaigns, $count, $page, $competition) {
+function _signup_resource_index($user, $users, $campaigns, $count, $page, $competition) {
+
+  if (isset($users)) {
+    $user = $users;
+  }
+
   $parameters = [
     'user' => $user,
     'campaigns' => $campaigns,
--- a/lib/modules/dosomething/dosomething_signup/includes/Signup.php
+++ b/lib/modules/dosomething/dosomething_signup/includes/Signup.php
@@ -91,7 +91,7 @@ class Signup extends Entity {
 
     $user = user_load($data->uid);
     $this->user = [
-      'id' => $data->field_northstar_id_value,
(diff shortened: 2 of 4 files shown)


weerd:
LOL. Fix sample URL in description! `?competitions=true...`  :stuck_out_tongue_winking_eye:

lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
```diff
@@ -44,6 +44,14 @@ function _signup_resource_definition() {
             'default value' => NULL,
           ],
           [
+            'name' => 'users',
+            'description' => 'The id of the specified user to get signups for.',
```
weerd: Change description to imply a single or multiple ids :)

itsjoekent:
@weerd okie dokes all updated

lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
```diff
@@ -99,7 +107,12 @@ function _signup_resource_access() {
  * @param boolean $competition
  * @return array
  */
-function _signup_resource_index($user, $campaigns, $count, $page, $competition) {
+function _signup_resource_index($user, $users, $campaigns, $count, $page, $competition) {
+
+  if (isset($users)) {
+    $user = $users;
```
weerd, line 113: I think my recommendation would be for it to be the other way around, so eventually we could deprecate the singular version and stay with the plural form.

weerd, line 113: Granted that probs means more work. Maybe create an issue for it, and it can be a future refactor? Or add a `@TODO` around this part?

itsjoekent, line 113: We can still deprecate the singular version if we wanted, but by doing 3 lines of code here I save myself from changing user --> users in many more places :wink:

weerd:
Is there any transforming/fixing that needs to be done to account for when retrieving a single user vs a collection of them and their signups? What's the sample output look like. Can you post in Slack since it might have sensitive user info?

lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
weerd, line 113: Oh yeah, I got that, but eventually we should change those lines of code in other places so it doesn't lead to confusion. And follows the patterns with the approach everywhere else with params of similar function.

DFurnes, line 113: +1 updating variable names for clarity.

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -88,6 +88,16 @@ public static function find(array $filters = []) {
   private function build($data) {
     $this->id = $data->sid;
     $this->created_at = $data->timestamp;
+
+    $user = user_load($data->uid);
+    $this->user = [
+      'id' => $data->field_northstar_id_value,
```
weerd: should wrap in the `dosomething_helpers_isset()` just in case...

itsjoekent: this returns "NONE" if it doesnt exist

itsjoekent: but added the isset check anyway

weerd: oh, well I think we want this to be `null` though if there is no id and not `"NONE"`... maybe don't use the `dosomething_helpers_isset()` and just use a simple ternary that turns "NONE" to `null`.

weerd:
Just fix up the StyleCI error!

weerd:
:+1:

PR #6337, opened 2016-03-30
adds code back in for 403/404 on reportbacks and reportback items aft…

### LINKED ISSUE #6274: /reportbacks and /reportback-items individual resources should return error code 403 404 when not found

chloealee:
#### What's this PR do?

Adds code back in from reverted merge in #6293. 
Fixes bug that exposed flagged reportbacks to non admin users. 

Find a flagged `rbid` and `fid`.

From admin account, `/reportbacks/:rbid` should return `202` status code and reportback object.

From an anon user account, same URL should return `403` status code and error message that reads `Access denied.`

`===` in line 134 of `dosomething_reportback.query.inc` was breaking the code because it was also checking the type to see if it was a boolean. Since `$reportback->flagged` is a string, it returned false and never unset. 

Would it be better to use `==` in this case or still use `===` and change that line to `if ($reportback->flagged === "1") {`?

Fixes #6274 

…er revert. Also fixes bug so non admins can no longer view flagged rbs


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
```diff
@@ -123,25 +121,26 @@ function dosomething_reportback_get_reportbacks_query($params) {
...
-function dosomething_reportback_filter_flagged_reportbacks($params) {
+function dosomething_reportback_accessible_results ($results) {
   if (user_access('view any reportback')) {
-    if (isset($params['flagged'])) {
-      unset($params['flagged']);
+    return $results;
+  }
+  else {
+    foreach ($results as $key => $reportback) {
+      if ($reportback->flagged == 1) {
```
chloealee: @weerd this is the line I am referring to in my "Any background context you want to provide?" comment :)

chloealee: @weerd actually, probably more helpful to have that comment here!  

The `===` was breaking the code because it was also checking the type to see if it was a boolean. Since `$reportback->flagged` is a string, it returned false and never unset.

Would it be better to use `==` in this case or still use `===` and change this to 
`if ($reportback->flagged === "1") {`?

weerd: ooo, I have a function for that... [`dosomething_helpers_convert_string_to_boolean()`](https://github.com/DoSomething/phoenix/blob/f0caf06c5ccacc05c9fceb659d1b51736186de83/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module#L127). Returns a _true_ boolean value from a string :)

Then you can do:

``` php
if (dosomething_helpers_convert_string_to_boolean($reportback->flagged) === true) {
```

sergiitk: Why just not `if ($reportback->flagged)`?

weerd: Hmmm, just maybe confirm it works with a string of `"1"` as well. It might just be for string `"true"`.

If not, then maybe you can update the helper function to handle both a string "true/false" and a string "0/1" :wink: 

Would definitely come in handy!

sergiitk: Check this out: http://php.net/manual/en/types.comparisons.php

chloealee: ooo @weerd @sergii-tkachenko it works with just `if ($reportback->flagged)` - is this the best way to go?

sergiitk: I think that's the best way in this case.
FYI `if ($reportback->flagged)` is just a short form of `if ($reportback->flagged == 1)`.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
```diff
@@ -123,25 +121,26 @@ function dosomething_reportback_get_reportbacks_query($params) {
...
+    }
   }
-
-  // Only show reviewed and/or un-flagged Reportbacks.
-  $params['flagged'] = FALSE;
-
-  return $params;
-}
-
+}
```
sergiitk: Note: all files must end with a newline.

sergiitk: http://stackoverflow.com/questions/5813311/no-newline-at-end-of-file

chloealee: Yes, @DFurnes and @weerd helped me to get the EditorConfig do that for me automatically moving forward but interesting to know reasoning behind it - thanks!

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
```diff
@@ -123,25 +121,26 @@ function dosomething_reportback_get_reportbacks_query($params) {
...
-function dosomething_reportback_filter_flagged_reportbacks($params) {
+function dosomething_reportback_accessible_results ($results) {
   if (user_access('view any reportback')) {
-    if (isset($params['flagged'])) {
-      unset($params['flagged']);
+    return $results;
+  }
+  else {
+    foreach ($results as $key => $reportback) {
+      if ($reportback->flagged) {
```
weerd, line 134: Is this always returned as a string though? Then it'll _always_ be true.... `"1"` and `"0"` are both technically true.

If it's not and is actually returned as an integer then the above should be fine. Just can't remember what the return is... I thought it was a string though...

sergiitk, line 134: "0" is FALSE, see http://php.net/manual/en/types.comparisons.php, "Loose comparisons with ==".

chloealee, line 134: Confirmed - with a reportback that is not flagged (`reportback->flagged = 0`), `0` is a string. It was also returned to the non-admin user as expected (instead of hitting `403` status code with error message `Access denied.`).

sergiitk, line 134: Again `'0'` as a string returns FALSE, as we need here:

```
php > var_dump('0' == TRUE);
bool(false)
```

weerd, line 134: Cool. I just wasn't sure if it was doing loose comparison or strict comparison when presented as 

``` php
if ("0") {
  // do the things
}
```

sergiitk, line 134: It's loose comparison by default.

weerd:
¡OLÉ!
:sheep: it!

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
weerd, line 134: Cool! Good to know!

image

PR #6340, opened 2016-03-31
returns empty array if non admin tries to hit /reportbacks?flagged=tr…

### LINKED ISSUE #6319: Need to filter out restricted reportbacks in index view
#### Current Behavior
- Currently, restricted reportbacks are not filtered out from the index view. 
- `/reportbacks?flagged=true` returns flagged RBs
#### Desired Behavior
- Depending on permissions, restricted reportbacks should be filtered out for anon users. 
- If only one reportback is found or `/reportbacks?flagged=true` is hit from an anon user account, status code should be 403 and `Access denied.` should be returned. 

**Question** Should this actually return an empty array, similar to the empty array that is returned when hitting `/reportback-items?status=flagged` from a non admin account?
#### Steps to Replicate
- Visit https://staging.dosomething.org/api/v1/reportbacks?flagged=true with an anon user account. 
- You'll see flagged RBs and a 200 status code. 
#### Why This Matters
- Only admins should be able to view flagged reportbacks.

chloealee:
#### What's this PR do?

Adds check to see if user has permission to access flagged reportbacks. If not, empty array will be returned to user trying to hit `/reportbacks?flagged=true` with a `200` status code. 

As a non admin user, try to hit  `/reportbacks?flagged=true`. Empty array and `200` status code should be returned. 

Fixes #6319

…ue and a 200 status code

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
+++ b/lib/modules/dosomething/dosomething_reportback/includes/Reportback.php
@@ -131,6 +131,10 @@ class Reportback extends Entity {
       throw new Exception('No reportback data found.');
     }
 
+    if (!dosomething_reportback_accessible_results($results)) {
+      throw new Exception('Access denied.');
+    }
+
     foreach($results as $item) {
       // @TODO: remove need for passing variable for constructor check.
       $reportback = new static(['ignore' => true]);


weerd:
Sweet! :+1:

PR #6343, opened 2016-04-04
adds field for campaign run id

### LINKED ISSUE #6338: Does the Admin Add RB Form Need a Field for Campaign Run?
#### User Story

Leads and the impact team use this form decently often to add reportbacks for members. Does this page need a field for run ID?

https://thor.dosomething.org/admin/users/rb/add
#### Why This Matters

Reportbacks must be associated with the correct run.

chloealee:
#### What's this PR do?

Adds a new field for Campaign Run ID in the reportback form. 
- Visit `/admin/users/rb/add`
- Put `362` into the `Node` field.
- Try to add `d` into the `Campaign Run ID` field - this should throw an error - `The specified entity d does not match requirements.`
- Put `1227` into the `Campaign Run ID` field - it should not only save but also update the field to read `Comeback Clothes 2014 [id:1227]` after saving. 

Fixes #6338

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -37,6 +37,15 @@ function dosomething_reportback_form($form, &$form_state, $entity = NULL) {
       '#required' => TRUE,
       '#default_value' => $entity->nid,
     );
+    $form['camapgin_run_nid'] = array(
+      '#type' => 'entity_autocomplete',
+      '#title' => t('Campaign Run ID'),
+      '#description' => t("The campaign run node this reportback was submitted for."),
+      '#entity_type' => 'node',
+      '#bundles' => array('campaign_run'),
+      '#required' => TRUE,
+      '#default_value' => $entity->nid,
+    );
     if (!empty($entity->uid)) {
       $account = user_load($entity->uid);
       $username = $account->name;


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -37,6 +37,15 @@ function dosomething_reportback_form($form, &$form_state, $entity = NULL) {
...
       '#default_value' => $entity->nid,
     );
+    $form['camapgin_run_nid'] = array(
+      '#type' => 'entity_autocomplete',
+      '#title' => t('Campaign Run ID'),
+      '#description' => t("The campaign run node this reportback was submitted for."),
+      '#entity_type' => 'node',
+      '#bundles' => array('campaign_run'),
+      '#required' => TRUE,
+      '#default_value' => $entity->nid,
```
angaither, line 47: does this default work?

chloealee, line 47: When hitting this form, both `Node` and `Campaign Run ID` fields are not populated with anything. Should I delete these lines because that will actually never work since we're creating a new entity and thus they shouldn't have a `#default_value`, right?

angaither, line 47: @chloealee I'm not sure it might be there if there's an error on the form and it saves the values so you don't have to refill out the entire form

chloealee, line 47: @angaither ah yes - it does work then in that case! If there's an error on the form, it saves the field to `Comeback Clothes 2014 [id:1227]` so you don't have to fill this out again.

angaither:
:+1:

PR #6341, opened 2016-04-01
Cleanup settings.local.php.

sergiitk:
#### What's this PR do?

Cleans up local settings. All settings provided in `default.settings.local.php` (that becomes `settings.local.php` through `ds build`) were either defaults or already provided in general `settings.php`.

The only exception was `$conf['dosomething_search_finder_url']`, but since solr is set [after](https://github.com/DoSomething/phoenix/blob/dev/config/settings.php#L109-L126) including environment staging, it's possible to preset it thought env variable.

In addition, it turned out that `$conf['cache_backends']` was overriding [redis configuration](https://github.com/DoSomething/phoenix/blob/dev/config/settings.php#L68) because it reinitialized itself as an empty array.
#### How should this be manually tested?

Rebuild site as normal.
#### What are the relevant tickets?

A part of #5921

Issue #5921, opened 2015-12-08
Cleanup build errors


sergiitk:
##### `ds build --install`
- [x] `WD Apache Solr: HTTP Status: 404; Message: Not Found: Not Found; Response: [error]
  Error 404 Not Found`
- [x] `WD Apache Solr: HTTP 404; Not Found: Not Found in apachesolr_cron [error]`
- [x] `WD Apache Solr: No Solr instance available during indexing. [error]`
- [ ] `WD varnish: Unable to connect to server socket 127.0.0.1:6082: Connection refused [error]`
- [x] `npm WARN engine dosomething-validation@0.2.2: wanted: {"node":"0.10.x"} (current: {"node":"0.12.7","npm":"2.11.3"})`
- [x] `npm WARN engine dosomething-modal@0.3.0: wanted: {"node":"0.10.x"} (current: {"node":"0.12.7","npm":"2.11.3"})`
- [ ] `npm WARN engine @dosomething/forge@6.6.2: wanted: {"node":"0.12.x"} (current: {"node":"4.4.0","npm":"2.14.20"})`
- [x] `npm WARN optional dep failed, continuing fsevents@1.0.5`
- [ ] `npm WARN prefer global babel@5.8.34 should be installed with -g`
##### `ds build; ds pull stage --db`
- [ ] `Drush status:
  <h1>Additional uncaught exception thrown while handling exception.</h1><h2>Original</h2><p>PDOException: SQLSTATE[42S02]: Base table or view not found: 1146 Table &amp;#039;dosomething.blocked_ips&amp;#039; doesn&amp;#039;t exist: SELECT 1 FROM {blocked_ips} WHERE ip = :i`
- [ ] `exception 'PDOException' with message 'SQLSTATE[42S02]: Base table or view not found: 1146 Table 'dosomething.system' doesn't exist' in /var/www/dev.dosomething.org/html/includes/database/database.inc:2171`

deezone:
`$ ds build`

```
...
Found makefile: services.make                                                         [ok]
Invalid argument supplied for foreach() make.utilities.inc:101          [warning]
```

DFurnes:
> npm WARN engine dosomething-validation@0.2.2: wanted: {"node":"0.10.x"} (current: {"node":"0.12.7","npm":"2.11.3"})
[...]

PR #6346, opened 2016-04-05
Fixes logic error in ds build install.

sergiitk:
#### What's this PR do?
- Fixes logic error in ds build install
- Implicitly sets up local solr environment
- As a result solr build errors are fixed: `WD Apache Solr: HTTP Status: 404; Message: Not Found: Not Found; Response: [error] Error 404 Not Found`, see #5921.

Rebuild site as usual.

There was a logic problem in `ds build --install` script. It copied global and local settings file AFTER profile installation. Without settings were installed incorrectly and configuration through features was incomplete. Also, the only reason why Drupal even knew how to access the database without settings file was that connection details were set in [`ds.aliases.drushrc.php`](https://github.com/DoSomething/tower/blob/master/roles/dosomething.phoenix/templates/drush/ds.aliases.drushrc.php.j2#L3).

Ref #6341
A part of #5921

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -216,27 +216,9 @@ function build_helper {
   # Clear caches and Run updates
   cd "$WEB_PATH/sites/$SITE"
 
-  if [[ $2 == "--install" ]]
-  then
-    echo "Installing"
-    install
-  fi
-
-  if [[ $3 == "--dev" ]]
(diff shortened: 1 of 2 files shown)


angaither:
:+1:

PR #6353, opened 2016-04-05
Bulk exclude backlogged reportbacks on node 362

### LINKED ISSUE #6344: Bulk exclude backlogged reportbacks on node 362
#### User Story

As an admin, I want all fids on node 362 with a quantity under 200 to be marked as excluded before it relaunches on Thursday.

Campaigns team was trying to go through the backlog and the number of reportbacks went from ~600 to ~1500. Got these numbers from Lizzy. 600 felt doable, 1500 seems unreasonable between now and then. Also a few reports of "Illegal choice detected" error when intern was running through things.
#### Why This Matters

Since campaign relaunches on Thursday and reportbacks are listed as oldest to newest the campaigns team won't realistically be able to get to current reportbacks without this being cleared out. Only other option I can think of is that we adjust the script to include only reportbacks that match one of the previous run ids which would buy us a little time I suppose and allow us to get the work done later this week.

@angaither let's chat about this in the AM

chloealee:
#### What's this PR do?

Script to bulk exclude any fids for node 362 that are pending and whose reportback has a quantity of less than 200. 

When ssh'd in your vagrant environment, run `drush updb -y` in `dosomething@dev:/var/www/dev.dosomething.org/html$`

Check `http://dev.dosomething.org:8888/us/node/362/rb/` and only `fids` with a quantity over 200 should still be pending and appear in the inbox. 

Fixes #6344

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
@@ -915,7 +915,7 @@ function dosomething_reportback_update_7033() {
  * Bulk excludes backlogged reportbacks on node 362.
  * Updates any fid with a pending status to be marked as excluded if the quantity in the reportback is less than 200.  
  */   
-function dosomething_reportback_update_7051() {
+function dosomething_reportback_update_7034() {
   $fids_to_be_marked_as_excluded = db_query("SELECT rb.nid, rb.run_nid, rb.quantity, rbf.status, rbf.fid
                                             FROM dosomething_reportback rb
                                             JOIN dosomething_reportback_file rbf on rb.rbid = rbf.rbid
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
@@ -915,11 +915,11 @@ function dosomething_reportback_update_7033() {
  * Bulk excludes backlogged reportbacks on node 362.
  * Updates any fid with a pending status to be marked as excluded if the quantity in the reportback is less than 200.  
  */   
-function dosomething_reportback_update_7034() {
-  $fids_to_be_marked_as_excluded = db_query("SELECT rb.nid, rb.quantity, rbf.status, rbf.fid
+function dosomething_reportback_update_7051() {
+  $fids_to_be_marked_as_excluded = db_query("SELECT rb.nid, rb.run_nid, rb.quantity, rbf.status, rbf.fid
                                             FROM dosomething_reportback rb
                                             JOIN dosomething_reportback_file rbf on rb.rbid = rbf.rbid
-                                            WHERE rb.nid = 362 AND rb.quantity < 200 AND rbf.status = 'pending';");
+                                            WHERE rb.nid = '362' AND rb.run_nid = '5439' OR rb.run_nid = '2341' AND rb.quantity < 200 AND rbf.status = 'pending';");
 
   foreach ($fids_to_be_marked_as_excluded as $fid) {
     db_update('dosomething_reportback_file')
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
```diff
@@ -910,3 +910,24 @@ function dosomething_reportback_update_7033() {
...
+                                            WHERE rb.nid = 362 AND rb.quantity < 200 AND rbf.status = 'pending';");
+
+  foreach ($fids_to_be_marked_as_excluded as $fid) {
+    db_update('dosomething_reportback_file')
+      ->fields(array(
+          'status' => 'excluded',
+        )
+      )
+      ->condition('fid', $fid->fid)
+      ->execute();
```
angaither, line 931: I think you'll also want to update the `flagged` status on the reportback to 0 as well

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
```diff
@@ -910,3 +910,24 @@ function dosomething_reportback_update_7033() {
 // }
 // COMMENTED OUT by AG -- causing duplicate row errors and replaced by
 // https://github.com/DoSomething/phoenix/pull/6155
+
+/**
+ * Bulk excludes backlogged reportbacks on node 362.
+ * Updates any fid with a pending status to be marked as excluded if the quantity in the reportback is less than 200.  
+ */   
+function dosomething_reportback_update_7039() {
```
angaither: I think this is skipping a few updates. looks like prod is on `7033`

chloealee: ah yes, forgot to change that back - thanks!

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install
chloealee, line 931: @angaither per the third bullet point in [Nami's comment](https://github.com/DoSomething/phoenix/issues/6344#issuecomment-205929056) would this be misleading if other fids associated with this rb are marked differently?

angaither, line 931: @chloealee `flagged` is applied to the entire reportback if ONE item is flagged. if 0 items are `flagged` that property is set to `0`

chloealee, line 931: @angaither if the `fid` is marked as `excluded`, does that mean the reportback is `flagged`?

angaither, line 931: @chloealee no, only if one item is marked as `flagged`

the logic [here](https://github.com/DoSomething/phoenix/blob/7b2a84aa399c9bd2eb9a91ddc6ad60bf50e5ccd1/lib/modules/dosomething/dosomething_reportback/includes/Reportback.php#L356) might help

chloealee, line 931: @angaither wait nevermind about the comment above. 

The next step would be to do another query to get all the `fids` associated with the reportback and see if any of those are `flagged` - if so, change the `flagged` status of the reportback to `1` and if not, change the `flagged` status to `0`?

angaither, line 931: instead of doing a `db_update` like this you may be able to use the function i linked to above if you load the reportback item

chloealee, line 931: @angaither ah thank you!

angaither:
@chloealee also this campaign reopens tomorrow, so you'll want to only target reportbacks with `run_nid = 5439`

mikefantini:
@chloealee @angaither there might also be reportbacks with `run_nid = 2341`

chloealee:
thanks @angaither and @mikefantini ! Added those `run_nid`s to target along with the logic to change the reportback `flagged` status.

angaither:
@chloealee how have you tested this? did you pull rbs from a different campaign? I'd love to deploy just this branch to thor to take a look to make sure it does what's expected

chloealee:
@angaither yup! I was testing on a different campaign on my local and added a reportback with 2 reportback items. I ran the script and saw that both reportback items statuses were changed to `excluded` and the parent reportback `flagged` status changed to `0`. 

Then, I added two more reportback items and changed one `status` to `flagged`. I ran the script and the second reportback item's status updated to `excluded` and the reportback `flagged` status remained as `1`.

angaither:
@chloealee kool, let's deploy this to thor to test the branch with the real content then!

PR #6359, opened 2016-04-08
Campaign run param

### LINKED ISSUE #6328: Add campaign run to signups/reportback params
#### User Story

As an external service I want to grab activity for a particular campaign & a particular run only

currently we can look for at a signup endpoint for a particular campaign, but old runs might be included in this  `https://www.dosomething.org/api/v1/signups?campaigns=6661&competition=true`
we need to add `run=id` to the url 
#### Why This Matters

As we bulid out competitions, future users might signup on a campaign that has had multiple runs and the data will be confusing

chloealee:
#### What's this PR do?

Adds a `runs` param to `/signups` and `/reportbacks` to be able to query for each by campaign run. 

`http://dev.dosomething.org:8888/api/v1/reportbacks?campaigns=1631&runs=1874` should return an empty data array (there are no reportbacks for this campaign run. 

`http://dev.dosomething.org:8888/api/v1/reportbacks?campaigns=1631&runs=1903` should return reportback with `rbid = 3543`

`http://dev.dosomething.org:8888/api/v1/signups?campaigns=1631&runs=1903` should return only signups with `campaign run id = 1903` (this will show in the returned results. 

Fixes #6328

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/reportback_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/reportback_resource.inc
@@ -99,6 +99,14 @@ function _reportback_resource_definition() {
             'source' => ['param' => 'flagged'],
             'default value' => FALSE,
           ],
+          [
+            'name' => 'runs',
+            'description' => 'The ids of the specified campaign runs to get reportbacks for.',
+            'optional' => TRUE,
+            'type' => 'string',
+            'source' => ['param' => 'runs'],
+            'default value' => NULL,
+          ],
         ],
         'access callback' => '_reportback_resource_access',
         'access arguments' => ['index'],
@@ -144,7 +152,7 @@ function _reportback_resource_access($op) {
  * @param  bool    $flagged
  * @return array
  */
-function _reportback_resource_index($ids, $campaigns, $status, $count, $random, $page, $load_user, $flagged) {
+function _reportback_resource_index($ids, $campaigns, $status, $count, $random, $page, $load_user, $flagged, $runs) {
   $parameters =  [
     'ids' => $ids,
     'campaigns' => $campaigns,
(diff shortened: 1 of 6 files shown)


aaronschachter:
This looks good @chloealee -- can you include some sample responses and results into this PR? I don't have a local Phoenix setup too handy.

Also just checking in that we've performed sanity checks that nothing breaks when
- The `runs` parameter is not included
- An invalid Run value is passed to `runs` like, `abc`

chloealee:
@aaronschachter sure thing! Screen shots below. Also - confirmed that nothing breaks when `runs` param is not included and invalid run value returns empty data array. 
screen shot 2016-04-11 at 4 21 13 pm
screen shot 2016-04-11 at 4 21 45 pm
screen shot 2016-04-11 at 4 23 33 pm
screen shot 2016-04-11 at 4 24 08 pm
screen shot 2016-04-11 at 4 24 30 pm

aaronschachter:
Thanks @chloealee ! 🚢

chloealee:
yikes @aaronschachter looks like there are a few issues when testing this on thor:
- `/reportbacks` break for the `campaigns` and `runs` param
- `/signups?campaigns=362&runs=z` returns a signup for a `run_nid` of `0`

Do you think I should revert this before tomorrow morning's deploy?

aaronschachter:
> `/reportbacks` break for the `campaigns` and `runs` param

@chloealee - What breaks exactly?  An error thrown?

chloealee:
@aaronschachter we chatted in slack but just documenting :) 

`/reportbacks` would give an empty data array when using the `campaigns` or `runs` params. When I cleared cache on thor however, this seemed to work okay and return the data that I expected (not an empty data array).

PR #6360, opened 2016-04-08
Only load campaign nodes in Campaign.php.

### LINKED ISSUE #6357: Campaigns endpoint returns random nodes if id matches
#### Current Behavior
- A request to the `/campaigns` endpoint with a seemingly valid id such as:

```
https://staging.dosomething.org/api/v1/campaigns/504
```

Will return a response, however the above is a node id for a fact node and not a campaign.
#### Desired Behavior
- The campaigns endpoint should return an appropriate not found API error response if it id is not for an actual campaign node.
#### Steps to Replicate
- Hit the above endpoint. You get a response with lots of `null` values, when you shouldn't get a response at all.
#### Why This Matters

It's affecting test data on Gladiator at the moment, because we think we're getting valid results with autogenerated campaign ids, which should not be happening.

angaither:
#### What's this PR do?

Only loads node in the campaign class of `type = campaign`

pull down
`drush cc all`
check campaign pages & api endpoints, only get campaigns, not facts, not home pages, not NUTTIN

duh.

Fixes #6357

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
+++ b/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
@@ -48,7 +48,8 @@ class Campaign {
       $single_campaign = TRUE;
       $ids = [$ids];
     }
-    $results = node_load_multiple($ids);
+    // Only load campaign types in this class.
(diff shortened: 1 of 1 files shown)


weerd:
👍

Issue #6166, opened 2016-02-12
Include timing within Signup.Campaign Run property


aaronschachter:
Refs https://github.com/DoSomething/phoenix/issues/6153#issuecomment-182519054.

This informs a client if the Signup for a Current Run is for a closed campaign (which is when `timing.end` is in the past).

Sample response:

```
{
  "data": {
    "id": "2",
    "created_at": "1393010906",
    "campaign_run": {
        "id": "1982",
        "current" true,
        "timing": {
            start: "2015-07-18T00:00:00+0000",
            end: "2015-07-25T00:00:00+0000"
        },
    },
```

aaronschachter:
This won't block our 1.1 release, as we should be able to use the `campaign.status` property to determine whether the Campaign is closed.

PR #6361, opened 2016-04-08
sorts by signup ids in descending order

### LINKED ISSUE #6301: Signups index endpoint is not sorted
#### Current Behavior
- Our Signups index endpoint returns data in random order, because it's missing an [`ORDER BY`](https://github.com/DoSomething/phoenix/blob/ffe9863131196a65e355adc3c2e7b265aa19a0c7/lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc#L54)
#### Desired Behavior
- Sort by descending Signup ID at a minimum. Don't believe we have a current need for ascending
#### Why This Matters

On iOS we're sorting the Signups back in code (not yet in [Android](https://github.com/DoSomething/LetsDoThis-Android/issues/242)), and asking for a large number of Signups (count=200) to fully render a Profile. Ideally this count on iOS should be smaller, to improve performance and also to avoid invalid JSON data upon a large count set (refs https://github.com/DoSomething/LetsDoThis-iOS/issues/972). However, the data in paginated requests won't work until the response we get back is sorted.

Side note / issue is that it looks like there may be caching issues with pagination object returned by the SIgnups endpoint (see thread that begins at https://github.com/DoSomething/LetsDoThis-iOS/issues/956#issuecomment-197138927)

cc @jonuy

chloealee:
#### What's this PR do?

Adds an `orderBy` to `dosomething_signup_build_signups_query` to order by signup ids in descending order. 

Hit `/signups` and `ids` should be sorted in descending order. 

Fixes #6301

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/includes/SignupTransformer.php
+++ b/lib/modules/dosomething/dosomething_signup/includes/SignupTransformer.php
@@ -85,7 +85,11 @@ class SignupTransformer extends Transformer {
       $campaign = (object) $item->campaign;
       $current_run = $campaign->campaign_runs['current']['en']['id'];
       $current = ($item->campaign_run == $current_run);
-      $data += $this->transformSignup($item, $current);
+      $campaign_run = node_load($current_run);
+      $current_run_start = $campaign_run->field_run_date['en'][0]['value'];
+      $current_run_end = $campaign_run->field_run_date['en'][0]['value2'];
+
+      $data += $this->transformSignup($item, $current, $current_run_start, $current_run_end);
       $data['campaign'] = $this->transformCampaign((object) $item->campaign);
     }
 
--- a/lib/modules/dosomething/dosomething_signup/includes/SignupTransformer.php
+++ b/lib/modules/dosomething/dosomething_signup/includes/SignupTransformer.php
@@ -85,7 +85,6 @@ class SignupTransformer extends Transformer {
       $campaign = (object) $item->campaign;
       $current_run = $campaign->campaign_runs['current']['en']['id'];
       $current = ($item->campaign_run == $current_run);
-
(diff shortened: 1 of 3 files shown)


chloealee:
cc @jonuy

lib/modules/dosomething/dosomething_signup/includes/SignupTransformer.php
```diff
@@ -85,7 +85,11 @@ protected function transform($item) {
       $campaign = (object) $item->campaign;
       $current_run = $campaign->campaign_runs['current']['en']['id'];
       $current = ($item->campaign_run == $current_run);
-      $data += $this->transformSignup($item, $current);
+      $campaign_run = node_load($current_run);
+      $current_run_start = $campaign_run->field_run_date['en'][0]['value'];
+      $current_run_end = $campaign_run->field_run_date['en'][0]['value2'];
```
aaronschachter: @chloealee -- It's possible the Run may not have an end date set -- which would cause this to throw some exceptions. Can you confirm?

aaronschachter:
Overall looks good - think it needs some existence checks per comment.

lib/modules/dosomething/dosomething_signup/includes/SignupTransformer.php
chloealee: @aaronschachter oh shoot - I didn't mean for this to be apart of this PR - I thought I switched branches but looks like I didn't! This is meant for #6166 . I'm going to delete this stuff and just have changes made in `dosomething_signup.query.inc.

chloealee:
@aaronschachter updated - sorry about the confusion!

aaronschachter:
:+1: :ship:

PR #6367, opened 2016-04-12
adds check to make sure that campaign run id is a number

### LINKED ISSUE #6371: /signups?campaigns=362&runs=z returns signup for campaign run nid = 0
#### Current Behavior
- `/signsups?campaigns=362&runs=z` returns a signup with the run nid of `0`
#### Desired Behavior
- `/signsups?campaigns=362&runs=z` should return an empty data array 
#### Steps to Replicate
- Hit `www.dosomething.org/api/v2/signsups?campaigns=362&runs=z`
#### Why This Matters
- Return value is not what we expect and gives incorrect data. 
- We originally reverted this code when we saw this issue on Thor but decided to [re-add it](https://github.com/DoSomething/phoenix/pull/6369) so it isn't a blocker for Gladiator (right now, this bug shouldn't block anything for Gladiator). 
- Working PR open to check to make sure param is a number, like we expect: https://github.com/DoSomething/phoenix/pull/6367 

screen shot 2016-04-13 at 2 48 35 pm

cc @aaronschachter @deadlybutter - let me know if I missed anything in the notes!

chloealee:
#### What's this PR do?

Adds check to make sure that `$params['run']` is a number. 

Same as testing in #6359 , screen shots below: 

screen shot 2016-04-12 at 2 08 11 pm
screen shot 2016-04-12 at 2 08 45 pm
screen shot 2016-04-12 at 2 09 42 pm
screen shot 2016-04-12 at 2 10 19 pm
screen shot 2016-04-12 at 2 10 36 pm
screen shot 2016-04-12 at 2 11 12 pm
screen shot 2016-04-12 at 2 11 56 pm

When testing on thor, `/signups?campaigns=362&runs=z` returned a signup with the `run_id` of `0`. No other params have ids of `0` in the production database so do not need this check (should we include anyway to be safe and not sorry?). 

Fixes #6371

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -707,10 +707,11 @@ function dosomething_helpers_unset_array_keys(array $array = [], array $keys = [
 
 /**
 * Checks to see if items are numeric. Unsets anything non-numeric and returns array of numbers.
+* @param string|array
 */
 function dosomething_helpers_unset_non_numeric_values($items) {
   if (is_string($items)) {
-      $items = explode(' ', $items);
+    $items = [$items];
   }
   
   foreach ($items as $key => $item) {
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -521,6 +521,8 @@ function dosomething_helpers_floor_decimal($value, $decimals = 1) {
   return floor($value * pow(10, $decimals)) / pow(10, $decimals);
 }
 
+
+
 /**
  * Adds custom GA event to page
  *
@@ -701,6 +703,25 @@ function dosomething_helpers_unset_array_keys(array $array = [], array $keys = [
   }
 
   return $array;
+}
+
+/**
+* Checks to see if items are numeric. Unsets anything non-numeric and returns array of numbers.
+*/
+function dosomething_helpers_unset_non_numeric_values($items) {
+  if (is_string($items)) {
+      $items = explode(' ', $items);
+  }
+
+  $numeric_items = [];
+  
+  foreach ($items as $item) {
+    if (is_numeric($item)) {
+      array_push($numeric_items, $item);
+    } 
+  }
+
+  return $numeric_items;
 }
 
 /**
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -712,16 +712,14 @@ function dosomething_helpers_unset_non_numeric_values($items) {
   if (is_string($items)) {
       $items = explode(' ', $items);
   }
-
-  $numeric_items = [];
   
-  foreach ($items as $item) {
-    if (is_numeric($item)) {
-      array_push($numeric_items, $item);
+  foreach ($items as $key => $item) {
+    if (! is_numeric($item)) {
+      unset($items[$key]);
     } 
   }
 
-  return $numeric_items;
+  return $items;
 }
 
 /**
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -706,10 +706,11 @@ function dosomething_helpers_unset_array_keys(array $array = [], array $keys = [
 }
 
 /**
-* Checks to see if items are numeric. Unsets anything non-numeric and returns array of numbers.
+* Checks to see if items are numeric. Unsets anything non-numeric and returns array of numbers (or empty array).
 * @param string|array
+* @return array
 */
-function dosomething_helpers_unset_non_numeric_values($items) {
+function dosomething_helpers_unset_non_numeric_array_values($items) {
   if (is_string($items)) {
     $items = [$items];
   }
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
@@ -10,10 +10,10 @@
  *
  * @param array $params
  *   An associative array of conditions to filter by. Possible keys:
- *   - nid: (string) A campaign node nid to filter by.
- *   - rbid: (string) A reportback rbid to filter by.
+ *   - nid: (string or array) A campaign node nid to filter by.
+ *   - rbid: (string or array) A reportback rbid to filter by.
  *   - random: (boolean) If set, randomly sort the results.
- *   - runs: (string) Campaign run nid(s) to filter by. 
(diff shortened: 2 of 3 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
```diff
@@ -10,10 +10,10 @@
  *
  * @param array $params
  *   An associative array of conditions to filter by. Possible keys:
- *   - nid: (string) A campaign node nid to filter by.
- *   - rbid: (string) A reportback rbid to filter by.
+ *   - nid: (string or array) A campaign node nid to filter by.
+ *   - rbid: (string or array) A reportback rbid to filter by.
  *   - random: (boolean) If set, randomly sort the results.
- *   - runs: (string) Campaign run nid(s) to filter by. 
+ *   - runs: (string or array) Campaign run nid(s) to filter by.
```
chloealee: is this okay to write? If it's just one run_id, then it's a string. If it's multiple, it's an array.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
```diff
@@ -66,11 +66,19 @@ function dosomething_reportback_build_reportbacks_query($params = array()) {
...
-    } else {
-      $query->condition('rb.run_nid', $params['runs'], '=');
+    if (is_string($params['runs'])) {
+      $params['runs'] = explode(' ', $params['runs']);
     }
+
+    $campaign_runs = [];
+    
+    foreach ($params['runs'] as $campaign_run) {
+      if (preg_match('/^[0-9]+$/', $campaign_run)) {
```
aaronschachter: [`is_numeric`](http://php.net/manual/en/function.is-numeric.php) would make this a litlte more readable, although not sure what's exactly best practice / what we use the most -- cc @sergii-tkachenko ?

sergiitk: Yes, is_numeric() is the best practice here.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
aaronschachter: I believe these are all strings because we pass a string`12,234,89,23` when we want multiple runs, or campaigns... but then line 70 below looks like we are actually no longer passing an array...

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
```diff
@@ -66,11 +66,19 @@ function dosomething_reportback_build_reportbacks_query($params = array()) {
   }
 
   if (isset($params['runs'])) {
-    if (is_array($params['runs'])) {
```
aaronschachter, line 69: not sure why we're removing this. lets chat in Slack about this when you get a chance - i’m a little confused exactly on purpose of this PR and why we need to check the id’s passed are numeric..  refs https://github.com/DoSomething/phoenix/pull/6359#issuecomment-209046454

chloealee, line 69: @aaronschachter I tried to change line 68 to add `&& is_numeric($params['runs'])` when testing with `runs=z` but since it wasn't numeric, it skipped this entire conditional and still returned results when we want it to return an empty data object. 

In order to weed out the param values that aren't numeric for `runs`, I think it makes most sense to iterate through them all and put the valid ones in an array and then run the conditional clause against those that we know are definitely numbers? 

Refresher - on production and on thor, when `runs=z`, we get a signup object with the `run_id` of `0` returned as well - seems like the version of php matches the numbers to pass through to `0` for some reason.

aaronschachter, line 69: I'm still a little lost on why we remove this line -- can dive a little deeper into the functions but @weerd curious if this is a quick one to review?

weerd:
I can take a look in a bit! Reviewing some other PRs but will 👀  soon!

weerd:
So the `runs` url param can accept a single id or a csv of ids, right?

In the past, this has been dealt with by passing using the respective by the transformer for the resource by `setFilters()` methods specific to the resource... for parameters such as this they are passed through the [`formatData()` method](https://github.com/DoSomething/phoenix/blob/cd7ef101835655174ab160f82b3080b4c0eaf99a/lib/modules/dosomething/dosomething_campaign/includes/CampaignTransformer.php#L115) which will split the csv of ids into an array of id strings. If it's a single id than it just leaves it alone.

Then in the query, we checked to see if it's an array so we can decide what time of query to use... to use an `in` option. This type of query if passed a random string like `z` will just ignore that since it won't find any RBs with an id of `z` but shouldn't fail. So we defs should keep the `is_array()` check in the query.

If we want to make sure that the passed values are ids than might be better to have a helper function that you pass the array to (after the formatting) and it checks all the items to make sure they're numeric and `unset()` the ones that aren't. But as a helper function it'd be available to all code, not just nested in the reportback query code. Would be better to have it universal 😉 

@chloealee lemme know if you wanna chat this over further IRL!

chloealee:
ah thanks @aaronschachter and @weerd ! That's smart to make it a helper function to check if it's numeric - I'll work on that now!

chloealee:
@weerd @aaronschachter extracted the `Is_numeric` code/check into a helper function! I don't think we need the `is_array()` check still since I'm changing everything, even if a single run is passed, to an array in `dosomething_helpers_unset_non_numeric_values` in order to go through each id that is passed and ensure that it is a number. Since everything will be an array, we'll only need to query with the `in` option.

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -704,6 +706,24 @@ function dosomething_helpers_unset_array_keys(array $array = [], array $keys = [
 }
 
 /**
+* Checks to see if items are numeric. Unsets anything non-numeric and returns array of numbers.
+* @param string|array
+*/
```
weerd, line 712: Add a `@returns` so we know it aims to return an array ;)

lib/modules/dosomething/dosomething_signup/dosomething_signup.query.inc
```diff
@@ -49,8 +49,8 @@ function dosomething_signup_get_signups_query($params, $tally = FALSE) {
  *   An associative array of conditions to filter by. Possible keys:
  *   - uid: (string) User id(s) to filter by.
  *   - campaigns: (string) Campaign nid(s) to filter by.
- *   - competition: (boolean) Only return competition signups & order by reportback quantity.
- *   - runs: (string) Campaign run nid(s) to filter by.
+ *   - competition: (boolean) Only return competition signups & order by reportback quantity. 
+ *   - runs: (string) Campaign run nid(s) to filter by.
```
weerd: `(string|array)`

weerd:
Just a couple docblock edits, but looks good! 
:sheep: it!

PR #6368, opened 2016-04-12
Adds run nid query

itsjoekent:
#### What's this PR do?

Adds ability to specify run nid in signup query index

http://dev.dosomething.org:8888/api/v1/signups?run_nid=1903

We're making some complicated functions / logic in Gladiator that could be greatly simplified we allowed specifying run nid's

Fixes # 🍰


weerd:
For one thing, we don't want the param to be the drupal specific `run_nid`. We want it to be `campaign_run_id`

itsjoekent:
chloe is working on this already whoops

aaronschachter:
Refs https://github.com/DoSomething/phoenix/pull/6359

PR #6369, opened 2016-04-12
revert "Revert "Campaign run param""

### LINKED ISSUE #6328: Add campaign run to signups/reportback params

chloealee:
#### What's this PR do?

Re-adds code from #6359 

http://dev.dosomething.org:8888/api/v1/reportbacks?campaigns=1631&runs=1874 should return an empty data array (there are no reportbacks for this campaign run.

http://dev.dosomething.org:8888/api/v1/reportbacks?campaigns=1631&runs=1903 should return reportback with rbid = 3543

http://dev.dosomething.org:8888/api/v1/signups?campaigns=1631&runs=1903 should return only signups with campaign run id = 1903 (this will show in the returned results.

see more tests from #6359 

When I merged in #6359, I had some problems testing on thor - `/signups?campaigns=362&runs=z` gave a signup object with the `run_nid` of `0`. 

`/reportbacks` broke on thor when using the `campaigns` and/or the `runs` param (returned empty data objects when I expected many reportback objects to be returned) but when I cleared the cache, everything worked as expected.

Fixes #6328 

cc @deadlybutter 

This reverts commit c158b9e2df898bc11f4e797976247c1db123da7f.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/reportback_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/reportback_resource.inc
@@ -99,6 +99,14 @@ function _reportback_resource_definition() {
             'source' => ['param' => 'flagged'],
             'default value' => FALSE,
           ],
+          [
+            'name' => 'runs',
+            'description' => 'The ids of the specified campaign runs to get reportbacks for.',
+            'optional' => TRUE,
+            'type' => 'string',
+            'source' => ['param' => 'runs'],
+            'default value' => NULL,
+          ],
         ],
         'access callback' => '_reportback_resource_access',
         'access arguments' => ['index'],
@@ -144,7 +152,7 @@ function _reportback_resource_access($op) {
  * @param  bool    $flagged
  * @return array
  */
-function _reportback_resource_index($ids, $campaigns, $status, $count, $random, $page, $load_user, $flagged) {
+function _reportback_resource_index($ids, $campaigns, $status, $count, $random, $page, $load_user, $flagged, $runs) {
   $parameters =  [
     'ids' => $ids,
     'campaigns' => $campaigns,
@@ -154,6 +162,7 @@ function _reportback_resource_index($ids, $campaigns, $status, $count, $random, 
     'page' => $page,
     'load_user' => $load_user,
     'flagged' => $flagged,
+    'runs' => $runs,
   ];
 
   return (new ReportbackTransformer)->index($parameters);
--- a/lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
@@ -83,6 +83,14 @@ function _signup_resource_definition() {
             'source' => ['param' => 'competition'],
             'default value' => FALSE,
           ],
+          [
+            'name' => 'runs',
+            'description' => 'The ids of the specified campaign runs to get signups for.',
+            'optional' => TRUE,
+            'type' => 'string',
+            'source' => ['param' => 'runs'],
+            'default value' => NULL,
+          ],
         ],
         'access callback' => '_signup_resource_access',
         'access arguments' => ['index'],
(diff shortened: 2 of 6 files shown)


aaronschachter:
Code still looks good👍 , let's open an issue to track why we reverted (and re-reverted this) refs #6367

chloealee:
@aaronschachter @deadlybutter this failed on thor  :-/ I'll make an issue for the below now - thanks to both of you for all the help! 

screen shot 2016-04-12 at 6 03 04 pm
screen shot 2016-04-12 at 6 02 54 pm

itsjoekent:
the only one that scares me is the one where it returned the wrong run/signup when you gave a number

itsjoekent:
whats different between this & the code I wrote? cause that seemed to work fine

chloealee:
mine also seemed to work fine until thor! I'll make a comment on some of the differences I saw in the files @deadlybutter

lib/modules/dosomething/dosomething_api/resources/reportback_resource.inc
```diff
@@ -99,6 +99,14 @@ function _reportback_resource_definition() {
             'source' => ['param' => 'flagged'],
             'default value' => FALSE,
           ],
+          [
+            'name' => 'runs',
+            'description' => 'The ids of the specified campaign runs to get reportbacks for.',
+            'optional' => TRUE,
+            'type' => 'string',
```
chloealee: @deadlybutter your put `integer` here for the `type`

lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
```diff
@@ -83,6 +83,14 @@ function _signup_resource_definition() {
             'source' => ['param' => 'competition'],
             'default value' => FALSE,
           ],
+          [
+            'name' => 'runs',
+            'description' => 'The ids of the specified campaign runs to get signups for.',
+            'optional' => TRUE,
+            'type' => 'string',
```
chloealee, line 90: @deadlybutter you put `integer` here for the `type`

lib/modules/dosomething/dosomething_api/resources/signup_resource.inc
```diff
@@ -83,6 +83,14 @@ function _signup_resource_definition() {
...
+            'source' => ['param' => 'runs'],
+            'default value' => NULL,
```
chloealee, line 92: @deadlybutter you have `FALSE` for this line

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
```diff
@@ -64,6 +65,14 @@ function dosomething_reportback_build_reportbacks_query($params = array()) {
...
 
+  if (isset($params['runs'])) {
+    if (is_array($params['runs'])) {
+      $query->condition('rb.run_nid', $params['runs'], 'IN');
+    } else {
+      $query->condition('rb.run_nid', $params['runs'], '=');
+    }
+  }
+
   // Public API properties to expose:
```
chloealee, line 76: @deadlybutter for this, your PR only has it query for `=` since only allowing to check for one run_nid, which makes sense

lib/modules/dosomething/dosomething_signup/includes/SignupTransformer.php
```diff
@@ -130,6 +130,7 @@ private function setFilters($parameters) {
       'count' => (int) $parameters['count'] ?: 25,
       'page' => (int) $parameters['page'],
       'competition' => $parameters['competition'] ?: NULL,
+      'runs' => dosomething_helpers_format_data($parameters['runs']),
```
chloealee, line 133: @deadlybutter you use  `'run_nid' => (int) $parameters['run_nid'],`, maybe I should try this too?

PR #6378, opened 2016-04-15
Fixes northstar id not displaying on some requests

itsjoekent:
#### What's this PR do?

Instead of using the northstar id table (which @DFurnes says is not properly working, and Ive confirmed) we decided make a call to northstar to get the ID, like the reportbacks endpoint.

Ive been using http://dev.dosomething.org:8888/api/v1/signups?users=1700226 (My user id)

Some people being exported in gladiator did not have any reportback activity, even though theye have reportedback. I believe this is because the ID is null in the response for them (and when you visit the profile NORTHSTAR_ID isnt set). because the ID is null we cant associate data with members in gladiator. 

Fixes # 🍰

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -189,6 +189,9 @@ function dosomething_helpers_extract_field_data($field, $language = LANGUAGE_NON
         // Date values, so need to extract using custom method.
         $values[] = dosomething_helpers_extract_field_dates($content);
       }
+      else if (in_array('country', $keys)) {
+        $values[] = $content['country'];
+      }
       elseif ($keys[0] === 'value') {
         // Text value, so need to extract using custom method.
         $values[] = dosomething_helpers_extract_field_text($content);
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -190,6 +190,7 @@ function dosomething_helpers_extract_field_data($field, $language = LANGUAGE_NON
         $values[] = dosomething_helpers_extract_field_dates($content);
       }
       else if (in_array('country', $keys)) {
+        //TODO: Build function to get all address data, not just country.
         $values[] = $content['country'];
       }
       elseif ($keys[0] === 'value') {
--- a/lib/modules/dosomething/dosomething_signup/includes/Signup.php
+++ b/lib/modules/dosomething/dosomething_signup/includes/Signup.php
@@ -91,7 +91,7 @@ class Signup extends Entity {
 
     $drupal_user = user_load($data->uid);
 
-    if (!isset($data->field_northstar_id_value)) {
+    if (!isset($data->field_northstar_id_value) || $data->field_northstar_id_value === "NONE") {
       $northstar_response = dosomething_northstar_get_northstar_user($data->uid);
       $northstar_response = json_decode($northstar_response);
 
--- a/lib/modules/dosomething/dosomething_signup/includes/Signup.php
+++ b/lib/modules/dosomething/dosomething_signup/includes/Signup.php
@@ -97,13 +97,19 @@ class Signup extends Entity {
 
       if (!empty($northstar_response->data) && !isset($northstar_response->error)) {
         $northstar_user = $northstar_response->data;
+        user_save($drupal_user, ['field_northstar_id' => [LANGUAGE_NONE => [0 => ['value' => $northstar_user->id]]]]);
       }
+    }
+
+    $last_initial = substr(dosomething_helpers_extract_field_data($drupal_user->field_last_name), 0, 1);
+    if ($last_initial === FALSE) {
+      $last_initial = NULL;
     }
 
     $this->user = [
       'id' => isset($northstar_user) ? $northstar_user->id : $data->field_northstar_id_value,
       'first_name' => dosomething_helpers_extract_field_data($drupal_user->field_first_name),
-      'last_initial' => dosomething_helpers_extract_field_data($drupal_user->field_last_name),
+      'last_initial' => $last_initial,
       'photo' => dosomething_helpers_extract_field_data($drupal_user->photo),
       'country' => dosomething_helpers_extract_field_data($drupal_user->field_address),
     ];
--- a/lib/modules/dosomething/dosomething_signup/includes/Signup.php
+++ b/lib/modules/dosomething/dosomething_signup/includes/Signup.php
@@ -89,14 +89,21 @@ class Signup extends Entity {
     $this->id = $data->sid;
     $this->created_at = $data->timestamp;
 
-    $user = user_load($data->uid);
+    $drupal_user = user_load($data->uid);
+
+    $northstar_response = dosomething_northstar_get_northstar_user($data->uid);
+    $northstar_response = json_decode($northstar_response);
+
+    if (!empty($northstar_response->data) && !isset($northstar_response->error)) {
+      $northstar_user = $northstar_response->data;
+    }
 
     $this->user = [
-      'id' => $data->field_northstar_id_value === 'NONE' ? NULL : $data->field_northstar_id_value,
-      'first_name' => dosomething_helpers_extract_field_data($user->field_first_name),
-      'last_initial' => dosomething_helpers_extract_field_data($user->field_last_name),
-      'photo' => $user->picture,
-      'country' => dosomething_helpers_extract_field_data($user->field_address),
+      'id' => isset($northstar_user) ? $northstar_user->id : NULL,
+      'first_name' => dosomething_helpers_extract_field_data($drupal_user->field_first_name),
+      'last_initial' => dosomething_helpers_extract_field_data($drupal_user->field_last_name),
+      'photo' => dosomething_helpers_extract_field_data($drupal_user->photo),
+      'country' => dosomething_helpers_extract_field_data($drupal_user->field_address),
     ];
 
     try {
--- a/lib/modules/dosomething/dosomething_signup/includes/Signup.php
+++ b/lib/modules/dosomething/dosomething_signup/includes/Signup.php
@@ -91,19 +91,21 @@ class Signup extends Entity {
 
     $drupal_user = user_load($data->uid);
 
-    $northstar_response = dosomething_northstar_get_northstar_user($data->uid);
-    $northstar_response = json_decode($northstar_response);
+    if (!isset($data->field_northstar_id_value)) {
+      $northstar_response = dosomething_northstar_get_northstar_user($data->uid);
+      $northstar_response = json_decode($northstar_response);
 
-    if (!empty($northstar_response->data) && !isset($northstar_response->error)) {
-      $northstar_user = $northstar_response->data;
+      if (!empty($northstar_response->data) && !isset($northstar_response->error)) {
+        $northstar_user = $northstar_response->data;
+      }
     }
 
     $this->user = [
-      'id' => isset($northstar_user) ? $northstar_user->id : NULL,
-      'first_name' => dosomething_helpers_isset($northstar_user, 'first_name'),
-      'last_initial' => dosomething_helpers_isset($northstar_user, 'last_initial'),
-      'photo' => dosomething_helpers_isset($northstar_user, 'photo'),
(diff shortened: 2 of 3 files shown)


lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -89,14 +89,21 @@ private function build($data) {
...
+    }
 
     $this->user = [
-      'id' => $data->field_northstar_id_value === 'NONE' ? NULL : $data->field_northstar_id_value,
-      'first_name' => dosomething_helpers_extract_field_data($user->field_first_name),
-      'last_initial' => dosomething_helpers_extract_field_data($user->field_last_name),
-      'photo' => $user->picture,
-      'country' => dosomething_helpers_extract_field_data($user->field_address),
+      'id' => isset($northstar_user) ? $northstar_user->id : NULL,
+      'first_name' => dosomething_helpers_extract_field_data($drupal_user->field_first_name),
```
DFurnes, line 111: These should be coming from the Northstar response, [like in the Reportback](https://github.com/DoSomething/phoenix/blob/728b9b8184f69fcd55df1d7f701d3cac4bba881c/lib/modules/dosomething/dosomething_reportback/includes/Reportback.php#L180-L185).

itsjoekent, line 111: Oh I can switch those out too no problem, one sec

itsjoekent:
@DFurnes fixed!

weerd:
So not sure how this would help when needing to request for like 300 or so users. I feel this might still hangup and timeout regardless...

What if instead, we keep the NS field search thing... then if that field is `null` or whatever, _then_ we do a request to Northstar to figure it out, AND we populate it in Drupal. Then next time the NS id would exist and no need for the call to NS to go an retrieve it.

itsjoekent:
diego with the smart ideas as always

DFurnes:
@weerd That sounds like a good idea. :+1:

weerd:
haha, well if it works out then we can 🔪  two 🐦 's with one 🗿 (<---stone?)

itsjoekent:
@weerd @DFurnes ok should be good to go

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -89,14 +89,23 @@ private function build($data) {
...
     $this->user = [
-      'id' => $data->field_northstar_id_value === 'NONE' ? NULL : $data->field_northstar_id_value,
-      'first_name' => dosomething_helpers_extract_field_data($user->field_first_name),
-      'last_initial' => dosomething_helpers_extract_field_data($user->field_last_name),
-      'photo' => $user->picture,
-      'country' => dosomething_helpers_extract_field_data($user->field_address),
+      'id' => isset($northstar_user) ? $northstar_user->id : $data->field_northstar_id_value,
+      'first_name' => dosomething_helpers_extract_field_data($drupal_user->field_first_name),
+      'last_initial' => dosomething_helpers_extract_field_data($drupal_user->field_last_name),
+      'photo' => dosomething_helpers_extract_field_data($drupal_user->photo),
```
weerd, line 113: Just curious why wasn't this using the helpers function before? Wondering if something w/ the result was erroring out or not? Not really sure.

DFurnes, line 113: Wait, do we store the same `photo` on Drupal profiles? I thought that was just for MCC spotlights, rather than the avatar that is normally included in the Northstar response.

weerd:
Is there any chance we can potentially `->save()` the NS info that's collected if missing to the actual `field_northstar_id_value` in the Drupal DB? So next time it's present?

lib/modules/dosomething/dosomething_signup/includes/Signup.php
itsjoekent, line 113: i dunno, it was here before i touched it haha. I can just 💥  it if we dont want it

itsjoekent:
@weerd the NS get call function thing is supposed to do a cache set automatically, dunno if it works

itsjoekent:
line 216 of ds northstar module

```
        cache_set('northstar_user_info_' . $id, $northstar_user_info, 'cache_dosomething_northstar', REQUEST_TIME + 60*60*24);
```

itsjoekent:
216*

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -89,14 +89,23 @@ private function build($data) {
...
+      'country' => dosomething_helpers_extract_field_data($drupal_user->field_address),
```
DFurnes, line 114: Are you sure that `$drupal_user->field_address` returns the country code?

I think it might be `$drupal_user->field_address['country']`.

itsjoekent, line 114: screen shot 2016-04-15 at 2 26 03 pm

weerd:
@deadlybutter maybe add a `@TODO` cause if it only caches it not as ideal. Would be nice to actually fill in the missing data in the Drupal database while we're at it :)

itsjoekent:
would you like me to do that now? i dont think it should be too hard

lib/modules/dosomething/dosomething_signup/includes/Signup.php
weerd, line 114: Just to be sure, @deadlybutter can you do a 

``` php
print_r($drupal_user->field_address);
die();
```

Curious what the output is on that. I think it should be right, but can't remember. Wondering if the function is only returning part of it

itsjoekent, line 114: yeah sure one second!

DFurnes, line 114: Magic! I think it might be pulling the first element from the array (alphabetically `country`), because `field_address` is definitely an array elsewhere.

itsjoekent, line 114: @weerd
screen shot 2016-04-15 at 2 30 48 pm

weerd:
@deadlybutter yeah, might as well if it's not hard. Would help down the line!

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -89,14 +89,23 @@ private function build($data) {
...
 
     $this->user = [
-      'id' => $data->field_northstar_id_value === 'NONE' ? NULL : $data->field_northstar_id_value,
-      'first_name' => dosomething_helpers_extract_field_data($user->field_first_name),
-      'last_initial' => dosomething_helpers_extract_field_data($user->field_last_name),
-      'photo' => $user->picture,
-      'country' => dosomething_helpers_extract_field_data($user->field_address),
+      'id' => isset($northstar_user) ? $northstar_user->id : $data->field_northstar_id_value,
+      'first_name' => dosomething_helpers_extract_field_data($drupal_user->field_first_name),
+      'last_initial' => dosomething_helpers_extract_field_data($drupal_user->field_last_name),
```
DFurnes: Hmm, `last_initial` is a computed property in Northstar. You're returning the full last name. I think it'd be good to either "compute" that here as well so we're not unnecessarily exposing user's full names.

itsjoekent:
@weerd saving works!
@DFurnes updated to just have last initial!

DFurnes:
@deadlybutter Looks good. Can you update `$drupal_user->field_address` to `$drupal_user->field_address['country']` like we discussed in the diff?

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -89,14 +89,29 @@ private function build($data) {
...
-      'id' => $data->field_northstar_id_value === 'NONE' ? NULL : $data->field_northstar_id_value,
-      'first_name' => dosomething_helpers_extract_field_data($user->field_first_name),
-      'last_initial' => dosomething_helpers_extract_field_data($user->field_last_name),
-      'photo' => $user->picture,
-      'country' => dosomething_helpers_extract_field_data($user->field_address),
+      'id' => isset($northstar_user) ? $northstar_user->id : $data->field_northstar_id_value,
+      'first_name' => dosomething_helpers_extract_field_data($drupal_user->field_first_name),
+      'last_initial' => $last_initial,
+      'photo' => dosomething_helpers_extract_field_data($drupal_user->photo),
+      'country' => dosomething_helpers_extract_field_data($drupal_user->field_address),
```
weerd, line 114: Maybe to help protect against this working weird use:

``` php
dosomething_helpers_isset($drupal_user->field_address, 'country')
```

itsjoekent:
@DFurnes that returns null

lib/modules/dosomething/dosomething_signup/includes/Signup.php
itsjoekent, line 114: oh i can try that

itsjoekent, line 114: also returns null

DFurnes, line 114: Ah yeah, that's a good idea.

DFurnes:
@deadlybutter Ah yeah, from the screenshot you posted it looks like it's `[0][LANGUAGE_NONE]['country']` but @weerd's suggestion is probably the right approach. :wink:

itsjoekent:
@DFurnes no @weerd 's suggestion returns null :/

weerd:
So turns out it's luck that pulls the first item and happens to be the country:

https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module#L198

Might want to update the helpers function to call a diff function if it happens to be address related data. I know this is not code you specifically worked on @deadlybutter but would help to correct it, or make it more clear :)

itsjoekent:
maybe im missing something, why cant i just put $drupal_user->field_address[0][LANGUAGE_NONE]['country']

if thats null ^ then it doesnt really matter?

weerd:
That can probs work, just not sure w/ all things #global if it's affected if it's not actually `LANGUAGE_NONE` aka `und`.... and maybe a global user...?

itsjoekent:
address has never had a diff language AFAIK

DFurnes:
@deadlybutter I think you're safe to directly access the field here, but it'd be good to fix the bug in `dosomething_helpers_extract_field_data` while we're here (or at the very least, create an issue describing it) so it doesn't trip us up again in the future.

weerd:
☝️

itsjoekent:
im not sure what the bug is? accessing the first element? based on where it is in the if else chain i'd say thats intended functionality as a last resort

DFurnes:
That's definitely not expected functionality.

itsjoekent:
what should it be doing then? just nothing?

DFurnes:
As a counter example, imagine we'd merged the code where it was implictly grabbing `country` since it's the first element. If we later added a `apt` field to the address, it'd unexpectedly switch to grabbing that field instead.

DFurnes:
Right, it should either return null or throw an error since it's being given unexpected input.

weerd:
The bug is that it's pure luck that we got the right value. It very well could have been one of the other properties first (or if it gets sorted or something) and then you get a different result. Ideally it would call another function (like it does with dates, etc) and returns the array of addresses or deals with it appropriately.

itsjoekent:
gotcha gotcha, i can do that now then

itsjoekent:
@weerd better? or do we need a function for all of the fields under address?

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -189,6 +189,9 @@ function dosomething_helpers_extract_field_data($field, $language = LANGUAGE_NON
         // Date values, so need to extract using custom method.
         $values[] = dosomething_helpers_extract_field_dates($content);
       }
+      else if (in_array('country', $keys)) {
+        $values[] = $content['country'];
```
weerd, line 194: I think this should be good for now, but add a `@TODO` or some comment to help explain what's going on and maybe that this needs a dedicated function later on (like `dosomething_helpers_extract_field_location_data()` or something) for extracting all the info at a later date.

itsjoekent, line 194: done

weerd:
👍

Issue #6372, opened 2016-04-13
Runs param on /reportbacks breaks with pagination and returning correct data


chloealee:
#### Current Behavior
- When using the `runs` param with `/reportbacks`, each page returns same exact reportback objects. This breaks because only values for one campaign run id are returned if you pass multiple ids in the `run` param e.g. `https://www.dosomething.org/api/v1/reportbacks?campaigns=1508&runs=5654,3182`
  ![screen shot 2016-04-13 at 3 11 59 pm](https://cloud.githubusercontent.com/assets/9019452/14506134/55415c08-018a-11e6-8085-b232806f7de5.png)
  ![screen shot 2016-04-13 at 3 14 02 pm](https://cloud.githubusercontent.com/assets/9019452/14506143/61706686-018a-11e6-8f6d-c56a2863d1c9.png)
- Correct data is not returned - the below returns no reportbacks when there are in fact reportbacks in the database.
  ![screen shot 2016-04-13 at 3 18 30 pm](https://cloud.githubusercontent.com/assets/9019452/14506302/0e3610fa-018b-11e6-8f87-7ca31e1e35f3.png)
  ![screen shot 2016-04-13 at 3 19 30 pm](https://cloud.githubusercontent.com/assets/9019452/14506380/7c9ae566-018b-11e6-85c6-a296e0f81adf.png)
#### Desired Behavior
- Pagination should return different reportback objects on each page. 
- Data should be returned if it exists in the database. 
#### Steps to Replicate
- Hit URLs in screenshots. 
- Make sure that your Language Settings in your user profile are set to English (not English, Global).
#### Why This Matters
- Currently, incorrect data is being returned when using the `runs` param.
#### Countries Affected (optional)
- It seems like `/reportbacks?campaigns=362` (or any campaign nid) + params only gets returned if your language is set to English. Is this okay @aaronschachter ?

chloealee:
cc @angaither @weerd @sbsmith86 @deadlybutter too as an FYI for gladiator work - is this blocking anyone significantly?

weerd:
@chloealee I don't think so at the moment... I think we're only really using the `/signups` endpoint, but if this also affects that endpoint then maybe?

itsjoekent:
well we are still facing a bug in gladiator regarding the runs / signup thing, but i dont believe it has anything to do with this phoenix

chloealee:
[...]

PR #6381, opened 2016-04-18
Runs bug

### LINKED ISSUE #6373: Runs param on /signups returns incorrect data with pagination
#### Current Behavior
- When using the runs param with `/signups`, each page returns same exact signup objects. This returns incorrect information - only values for one campaign run id are returned if you pass multiple ids in the run param e.g. https://www.dosomething.org/api/v1/signups?campaigns=362&runs=2341,5439
#### Desired Behavior
- Pagination should return different reportback objects on each page.
#### Steps to Replicate
- Hit https://www.dosomething.org/api/v1/signups?campaigns=362&runs=2341,5439 and click on `next_uri` link -  you'll see that the results are the same as the previous and next pages.
#### Why This Matters
- Incorrect data is being returned when using multiple campaign runs. 
- Pagination is broken.

chloealee:
#### What's this PR do?

Added missing `setOffset` to `ReportbackTransformer.php` and `SignupTransformer.php` to fix pagination issues. 

Updates `build` function in Signup.php to only return `drupal_id` - the rest of the user information gets filled in by [Northstar proxy](https://github.com/DoSomething/northstar/pull/343). 

Make sure that `prev_uri` and `next_uri` links return different data on each page for `/signups` and `/reportbacks`.

Fixes #6373 
Refs #6372 - fixes pagination error for this issue.

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/includes/Signup.php
+++ b/lib/modules/dosomething/dosomething_signup/includes/Signup.php
@@ -89,19 +89,8 @@ class Signup extends Entity {
     $this->id = $data->sid;
     $this->created_at = $data->timestamp;
 
-    $drupal_user = user_load($data->uid);
-
-    if (!isset($data->field_northstar_id_value) || $data->field_northstar_id_value === 'NONE') {
-      $northstar_response = dosomething_northstar_get_northstar_user($data->uid);
-      $northstar_response = json_decode($northstar_response);
-
-      if (!empty($northstar_response->data) && !isset($northstar_response->error)) {
-        dosomething_northstar_save_id_field($drupal_user, $northstar_response);
-      }
-    }
-
     $this->user = [
-      'id' => isset($northstar_user) ? $northstar_user->id : $data->field_northstar_id_value,
+      'drupal_id' => $data->uid,
     ];
 
     try {
--- a/lib/modules/dosomething/dosomething_signup/includes/Signup.php
+++ b/lib/modules/dosomething/dosomething_signup/includes/Signup.php
@@ -100,17 +100,8 @@ class Signup extends Entity {
       }
     }
 
-    $last_initial = substr(dosomething_helpers_extract_field_data($drupal_user->field_last_name), 0, 1);
-    if ($last_initial === FALSE) {
-      $last_initial = NULL;
-    }
-
     $this->user = [
       'id' => isset($northstar_user) ? $northstar_user->id : $data->field_northstar_id_value,
-      'first_name' => dosomething_helpers_extract_field_data($drupal_user->field_first_name),
-      'last_initial' => $last_initial,
-      'photo' => isset($drupal_user->picture) ? $drupal_user->picture : NULL,
-      'country' => dosomething_helpers_extract_field_data($drupal_user->field_address),
     ];
 
     try {
--- a/lib/modules/dosomething/dosomething_signup/includes/Signup.php
+++ b/lib/modules/dosomething/dosomething_signup/includes/Signup.php
@@ -109,7 +109,7 @@ class Signup extends Entity {
       'id' => isset($northstar_user) ? $northstar_user->id : $data->field_northstar_id_value,
       'first_name' => dosomething_helpers_extract_field_data($drupal_user->field_first_name),
       'last_initial' => $last_initial,
-      'photo' => dosomething_helpers_extract_field_data($drupal_user->photo),
+      'photo' => isset($drupal_user->picture) ? $drupal_user->picture : NULL,
       'country' => dosomething_helpers_extract_field_data($drupal_user->field_address),
     ];
 
--- a/lib/modules/dosomething/dosomething_signup/includes/Signup.php
+++ b/lib/modules/dosomething/dosomething_signup/includes/Signup.php
@@ -90,12 +90,7 @@ class Signup extends Entity {
     $this->created_at = $data->timestamp;
 
     $this->user = [
-      'id' => NULL,
       'drupal_id' => $data->uid,
-      'first_name' => NULL,
-      'last_initial' => NULL,
-      'photo' => NULL,
-      'country' => NULL,
     ];
 
     try {
--- a/lib/modules/dosomething/dosomething_signup/includes/Signup.php
+++ b/lib/modules/dosomething/dosomething_signup/includes/Signup.php
@@ -90,7 +90,12 @@ class Signup extends Entity {
     $this->created_at = $data->timestamp;
 
     $this->user = [
+      'id' => NULL,
       'drupal_id' => $data->uid,
+      'first_name' => NULL,
+      'last_initial' => NULL,
+      'photo' => NULL,
+      'country' => NULL,
     ];
 
     try {
@@ -109,11 +114,11 @@ class Signup extends Entity {
       try {
         $this->reportback = Reportback::get($data->rbid);
       } catch (Exception $e) {
-        $this->reportback = null;
+        $this->reportback = NULL;
       }
     }
     else {
-      $this->reportback = null;
+      $this->reportback = NULL;
     }
   }
 }
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -451,12 +451,12 @@ abstract class Transformer {
    */
   protected function transformUser($data) {
     return [
-      'id' => $data->id,
-      'drupal_id' => $data->drupal_id,
-      'first_name' => $data->first_name,
-      'last_initial' => $data->last_initial,
-      'photo' => $data->photo,
-      'country' => $data->country,
+      'id' => dosomething_helpers_isset($data, 'id'),
+      'drupal_id' => dosomething_helpers_isset($data, 'drupal_id'),
+      'first_name' => dosomething_helpers_isset($data, 'first_name'),
+      'last_initial' => dosomething_helpers_isset($data, 'last_initial'),
+      'photo' => dosomething_helpers_isset($data, 'photo'),
+      'country' => dosomething_helpers_isset($data, 'country'),
     ];
   }
 
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -452,6 +452,7 @@ abstract class Transformer {
   protected function transformUser($data) {
     return [
       'id' => $data->id,
+      'drupal_id' => $data->drupal_id,
       'first_name' => $data->first_name,
       'last_initial' => $data->last_initial,
       'photo' => $data->photo,
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
@@ -68,7 +68,8 @@ function dosomething_reportback_build_reportbacks_query($params = array()) {
   if (isset($params['runs'])) {
     if (is_array($params['runs'])) {
       $query->condition('rb.run_nid', $params['runs'], 'IN');
-    } else {
+    } 
+    else {
       $query->condition('rb.run_nid', $params['runs'], '=');
     }
   }
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.query.inc
@@ -74,7 +74,7 @@ function dosomething_reportback_build_reportbacks_query($params = array()) {
   }
 
   // Public API properties to expose:
-  $rb_fields = array('rbid', 'created', 'updated', 'quantity', 'uid', 'flagged');
+  $rb_fields = array('rbid', 'created', 'updated', 'quantity', 'uid', 'flagged', 'run_nid');
 
   // Staff-only properties to add:
   if (user_access('view any reportback')) {
--- a/lib/modules/dosomething/dosomething_reportback/includes/ReportbackTransformer.php
+++ b/lib/modules/dosomething/dosomething_reportback/includes/ReportbackTransformer.php
@@ -109,7 +109,7 @@ class ReportbackTransformer extends Transformer {
       'flagged' => dosomething_helpers_convert_string_to_boolean($parameters['flagged']),
       'runs' => dosomething_helpers_format_data($parameters['runs']),
     ];
-    
+
     $filters['offset'] = $this->setOffset($filters['page'], $filters['count']);
 
     // Unset False boolean values that affect the query builder.
--- a/lib/modules/dosomething/dosomething_reportback/includes/ReportbackTransformer.php
+++ b/lib/modules/dosomething/dosomething_reportback/includes/ReportbackTransformer.php
@@ -21,6 +21,7 @@ class ReportbackTransformer extends Transformer {
    */
   public function index($parameters) {
     $filters = $this->setFilters($parameters);
+
     try {
       $reportbacks = Reportback::find($filters);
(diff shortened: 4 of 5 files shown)


lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -110,7 +110,7 @@ private function build($data) {
       'id' => isset($northstar_user) ? $northstar_user->id : $data->field_northstar_id_value,
       'first_name' => dosomething_helpers_extract_field_data($drupal_user->field_first_name),
       'last_initial' => $last_initial,
-      'photo' => dosomething_helpers_extract_field_data($drupal_user->photo),
+      'photo' => isset($drupal_user->picture) ? $drupal_user->picture : NULL,
```
aaronschachter: Question about this change -- for #mobile-app, we need the User's Northstar avatar, not their default Drupal User.picture property...

itsjoekent: We get the northstar object above this now but only for ppl that dont have a northstar ID saved in the DB.

chloealee: oooh so right - thanks for catching @aaronschachter! So should we always load the northstar user then?

itsjoekent: doing this has been under debate - mainly because when we're bulk fetching users it will take _forever_ to get a response back. debating the best way to handle this... @weerd ?

chloealee: or @deadlybutter, should we just change back to `'photo' => dosomething_helpers_isset($northstar_user, 'photo')` instead of your change [here](https://github.com/DoSomething/phoenix/commit/2a79a94b2e8760f5eba04576faa5b61099474385)?

itsjoekent: @chloealee no this wont work 100% of the time, northstar user is only created when we dont have the northstar ID saved in the database https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_signup/includes/Signup.php#L99

chloealee: ahh i see thanks @deadlybutter !

chloealee: discussion between @deadlybutter and I on slack [here](https://dosomething.slack.com/archives/phoenix/p1461004451000027). @weerd do you think it makes sense to always load the user object so we have access to the avatar?

itsjoekent: *only for requests with 1 user

weerd: Maybe that's something that Northstar could add when it's getting proxied. Kinda like how with Gladiator, we send the `/signups` request to Gladiator and it will convert all the NS IDs we pass to Phoenix IDs to send to Phoenix. So what could happen is NS receives the request to `/signups` proxies to Phoenix, when Phoenix sends it back to NS, it takes control and adds the NS avatar  info then, before proceeding with sending the full response back to client? Avoids having to request from NS only to then send back to NS anyways.... @DFurnes thoughts?

DFurnes: It's definitely a good short-term fix. I'd like to be able to remove those two proxy endpoints from Northstar sometime in the near-ish future though, and I worry a bit that this would be making them more tightly coupled. Not sure what a better solution is though. 😅

itsjoekent: removing the proxies would break gladiator, we dont have drupal id's

DFurnes: Yep, hence sometime in the near-ish future, rather than immediately and without warning. 😉 Once Phoenix has Northstar IDs in it though, that part of the proxy can be removed.

chloealee: @DFurnes @weerd I'm a little confused about how this would play out.
@weerd - you said "Kinda like how with Gladiator, we send the /signups request to Gladiator and it will convert all the NS IDs we pass to Phoenix IDs to send to Phoenix." - can you send me the code for this so I can see by example to wrap my head around?

DFurnes: @chloealee Here's the [`signups/` proxy route](https://github.com/DoSomething/northstar/blob/dev/app/Http/Controllers/SignupController.php#L32-L58) in Northstar.

I think @weerd's solution is a good fix for now, but I do want to think a bit on how we can better handle situations like this so we're not piping requests through an ever-growing number of services when we need info from multiple places.

weerd: Yeah, my main worry is having a bunchof back-and-forth API calls to grab all data and it totally killing the processing time due to all the HTTP requests. I'm also not 100% sure I know _what_ the user avatar is and how it differs from the profile picture (which I think was just used for MCC stuff)...

chloealee: thanks @DFurnes and @weerd ! Sorry for my confusion on my end but are the steps to take to also send back the NS avatar picture in the [proxy route](https://github.com/DoSomething/northstar/blob/dev/app/Http/Controllers/SignupController.php#L32-L58) and to call this in [Signup.php's `build` function](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_signup/includes/Signup.php#L88)?

@weerd the avatar picture is confusing! It's the avatar picture on the mobile app that gets saved to aws. As I understand, on the site, the only people who have profile pictures are admins? cc @aaronschachter in case I got any of that incorrect.

DFurnes: @chloealee Yeah I think that's correct! I wonder if we could "merge" those two properties and just reformat the MCC campaign badge to use a rounded avatar instead...? It is pretty confusing to have both. 😕 

If we add the user property in the Northstar proxy, we can just remove the `$this->user` business in the `build` function and then "append" it to each signup in the result in the Northstar proxy method (since then it could happen as a single database query, rather than multiple HTTP requests).

chloealee: awesome, thanks for everyone's input! Talked more with @DFurnes in slack but for documentation, here's the plan: 
1. update the code in phoenix for approval
2. open a new issue in northstar to address changes (adding user info, including avatar to pass back) 
3. make a new pr with changes in northstar
4. once approved, merge northstar and deploy
5. merge phoenix changes and deploy

aaronschachter: Chiming in on Avatar vs Drupal user picture -- assuming nothing's changed over the past year or so -- yep the only place we use the Drupal `user.picture` is when we have a MCC - our staff uploads an image we prepare with the Member on a transparent background. Deprecating the Drupal `user.picture` would be ideal but likely requires tweaks to the Member Created designs... if we even use them anymore? 

`destroy` 🔥 🚒

weerd: Awesome! Sounds good to me @chloealee I like this direction :)

chloealee:
@weerd @DFurnes Working through this more, I think we'll need to still keep the `$this->user` business on the phoenix side because otherwise, in ns, we don't have access to the user if it's not returned in the results here. 

In ns, I'm taking the results from hitting the phoenix endpoint, iterating through each one to get the user, find the user avatar image and append it to the user object. 

Just want to run by you both quickly to make sure I'm not going down a rabbit hole! Please let me know if this makes sense to you both or if I'm overlooking something obvious!

DFurnes:
@chloealee Ahhhh, yeah that makes sense. It may be best to remove all the fields aside from `drupal_id` then so that it's more clear that the user content gets re-written elsewhere?

chloealee:
@DFurnes smart - that makes sense!

weerd:
Yep yep! Defs don't remove the `user` object on the Phoenix side. There's other things that need it, etc. Just getting rid of the `user->photo` or whatever should be fine, and then like you mention, having NS add that data and property :)

DFurnes:
@weerd Do we have any services that aren't using the Northstar proxy for those endpoints? If not, I'd make a vote for _not_ including those fields in the original Phoenix response and then replacing them on Northstar just so it's clear that they're coming from elsewhere.

weerd:
@DFurnes I don't think there's any service using `/signups` without going via the proxy. 

Which fields do you mean? I thought it's just the photo related one?

chloealee:
@weerd the `user` object also returns  `first_name`, `last_initial`, and `country` in addition to `photo` and `id`. I think @DFurnes is suggesting to leave `id` (which I believe is the northstar id) and then just append all the rest into the user object coming from NS.

weerd:
Oh gotcha! Yep yep. Thanks for clarifying @chloealee 💃

DFurnes:
> I think @DFurnes is suggesting to leave id (which I believe is the northstar id) and then just append all the rest into the user object coming from NS.

Yep! The reason I say that is so that there aren't two separate scenarios where `first_name` (or other fields) could be filled in with potentially different data (depending on where you hit the endpoint from).

If we only return `user: { drupal_id: 123456 }` on the response, I feel like it makes it a bit more clear that it's a "placeholder" for someone else down the line to fill-in the real user data in the response.

weerd:
Cool. We'll likely need to update other endpoints that pull in the user info too. It'll probs just be an update to the user transformer :)

chloealee:
@weerd  - @DFurnes helped to refactor this to make sure nothing breaks! We just added `drupal_id` in the `transformUser` function in the Transformer. This way, nothing should change elsewhere and data just gets overridden by the northstar proxy. 

This is all set for review whenever you get a chance @weerd !

weerd:
Right-on! Checking now 💃

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -89,29 +89,8 @@ private function build($data) {
...
-        user_save($drupal_user, ['field_northstar_id' => [LANGUAGE_NONE => [0 => ['value' => $northstar_user->id]]]]);
-      }
-    }
-
-    $last_initial = substr(dosomething_helpers_extract_field_data($drupal_user->field_last_name), 0, 1);
-    if ($last_initial === FALSE) {
-      $last_initial = NULL;
-    }
-
     $this->user = [
```
weerd, line 92: Ok, so just to be sure on this (since Gladiator highly depends on this part at the moment), the response sends back to Northstar the drupal id, and Northstar figures out the rest to add the user info, etc?

Gonna cc @deadlybutter cause he worked on the above fix for Gladiator. Once this is merged we should definitely test out Gladiator QA w/ Phoenix Staging!

weerd, line 92: Also, this _only_ happens for when it's a NS proxied endpoint to Phoenix, correct?

itsjoekent, line 92: nearly had a heart attack, but read top of PR & now I get it haha

yar, lets test test test when this stuff gets deployed.

will deploying the northstar PR first break anything? feels safer to do that first to make sure everything works on that end, then deploy phoenix

DFurnes, line 92: Yep, so Phoenix would always return _just_ `drupal_id` for the user response since that's all it can reliably say about a user. And then for any requests going through the Northstar proxy, Northstar would add the relevant `first_name`, `last_initial`, etc etc to each `user` object.

weerd, line 92: ok, cause in the user transformer above, it's not _just_ returning the `drupal_id`, but other items as well: https://github.com/DoSomething/phoenix/pull/6381/files#diff-2f3c6ef3f5ca436621af7d7d225af381R454

DFurnes, line 92: @weerd Is that `transformUser` method used elsewhere? We should probably remove those fields if not.

weerd, line 92: Yeah, it's definitely used on other endpoints, like `/reportback` etc. Not just the `/signups` endpoint.

DFurnes, line 92: Mmmm, yeah! Let's not break those then. 😄

chloealee:
@DFurnes @weerd quick q! Trying to test and debug this and I'm not sure it's actually getting sent through to NS when I thought it already was hooked up - where does the NS proxy connect/come into play here?

DFurnes:
@chloealee Northstar is the "proxy" for the request –– so clients would make their request to Northstar (like `https://northstar.dosomething.org/v1/signups`) and then Northstar makes the corresponding Phoenix API request on their behalf (and makes any modifications before/after the request)!

So Phoenix never talks to Northstar here. :)

chloealee:
@DFurnes oooohhh that makes so much more sense, thank you for clarifying that! That's why when `/signups` testing on dev.dosomething.org, all user info isn't fully returned - mystery solved! Thank you!

lib/modules/dosomething/dosomething_signup/includes/Signup.php
```diff
@@ -89,28 +89,13 @@ private function build($data) {
...
-      'last_initial' => $last_initial,
-      'photo' => dosomething_helpers_extract_field_data($drupal_user->photo),
-      'country' => dosomething_helpers_extract_field_data($drupal_user->field_address),
+      'id' => NULL,
+      'drupal_id' => $data->uid,
+      'first_name' => NULL,
+      'last_initial' => NULL,
+      'photo' => NULL,
+      'country' => NULL,
     ];
```
chloealee, line 94: @weerd added these properties back in and set them to `NULL` to prevent errors in the db logs!

chloealee:
@weerd going to merge the [NS code ](https://github.com/DoSomething/northstar/pull/343) and then after would love to merge this! I tested with `/reportbacks?load_user=true` along with testing pagination and nothing broke on my local! Would you mind taking a final look for merge?

lib/modules/dosomething/dosomething_signup/includes/Signup.php
weerd, line 94: Cool. Are they always `null` now? When do they get populated? Is there a reason for changing from the above if those items existed? Been a few days since I looked at this and brain is kinda fuzzy on it.

chloealee, line 94: These get populated from the [NS proxy](https://github.com/DoSomething/northstar/pull/343). I set them to `NULL` because otherwise it throws an error in the db logs saying `Notice: Undefined property: stdClass::$country in Transformer->transformUser() (line 459 of /var/www/dev.dosomething.org/lib/modules/dosomething/dosomething_api/includes/Transformer.php).` for each of these properties.

weerd, line 94: Gotcha. An alternative option could be to use the `dosomething_helpers_isset()` function in the actual [`transformUser()`](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_api/includes/Transformer.php#L452)

Not a huge deal now, but it's something to consider if `Signup` class is used internally for stuff and doesn't call out to NS for DS Web App stuff. (Kinda like the `Campaign` class does). I know it's not the case now, but could be in the future. Maybe it's worth adding a comment to note that and not necessarily worry about changing it now.

chloealee, line 94: ah that is smart! I can make that change now to the `transformUser()` function now so we don't have to worry about it later :)

weerd:
¡OLÉ!
:sheep: it!

PR #6395, opened 2016-04-26
Update to include affiliates info in Campaign class build.

### LINKED ISSUE #6393: Expose sponsor info in campaigns endpoint
# _FEATURE OVERVIEW_

---
#### User Story

As an api consumer, I should know which campaigns are sponsored, to give love on the app. 

"sponsor & partner display"

```
affiliates: {
  partners: [
    {
      name: "H&M",
      sponsor: true,
      media: {
        uri: https://www.dosomething.org/sites/default/files/styles/wmax-423px/public/partners/HM-1.png?itok=N8vhbM3I,
        type: "image"
      },
      uri:  http://hm.com
    }
 ]
}
```

"partner display"

```
affiliates: {
  partners: [
    {
      name: "H&M",
      sponsor: false,
      media: {
        uri: null,
        type: null
      },
      uri:  http://hm.com
    }
  ]
}
```
#### Why This Matters

they pay us, their logo/name should be exposed. 
### Example

adding a bool value https://github.com/DoSomething/phoenix/pull/6208/files
### Sample nodes

Staging:
- 362
- 1334

Thor:
- 362
- 1508
- 5775

Prod:
- 362
- 6692

weerd:
#### What's this PR do?

This PR include a Campaign's affiliates information (such as partners in this particular case) to the `Campaign` class build function when a new `Campaign` is instantiated and also exposes this data to the API.

Example output:
image

Load a few campaigns both with and without partners (and partners that are also sponsors) via the API and see if the affiliate property on the response object is appropriately filled out.

If a partner does not have a logo "media" available, instead of making the `uri` and `type` be `NULL` I decided to make the entire `media` object `NULL`. Just seemed to make more sense to me that when planning on outputting the path to the logo, better to check the `media` object if that's something valid first. But we can nerd 🤓 battle over this if others feel strongly with a different opinion!

Also, since this is on the `Campaign` class object, the partner information is now nicely formatted and available for use in future Campaign template updates via the web interface as well, wink wink 😉 

Fixes #6393

---

@aaronschachter @angaither @chloealee 

cc: @mikefantini

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
+++ b/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
@@ -30,6 +30,7 @@ class Campaign {
   public $tags;
   public $timing;
   public $reportback_info;
+  public $affiliates;
   public $services;
 
   /**
@@ -184,6 +185,8 @@ class Campaign {
         $timing = $this->getTiming();
         $this->timing = $timing;
 
+        $this->affiliates = $this->getAffiliates();
+
         $this->services = $this->getServices();
       }
 
@@ -220,6 +223,14 @@ class Campaign {
     }
 
     return $data;
+  }
+
+  protected function getAffiliates()
+  {
+    // @TODO: Only grabs partners for now. Update with further affiliates as needed.
+    return [
+      'partners' => $this->getPartners(),
+    ];
   }
 
   /**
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
```diff
@@ -444,6 +455,49 @@ protected function getMobileCommonsData() {
...
+      $partner['name'] = $data_taxonomy->name;
+      $partner['sponsor'] = (bool) dosomething_helpers_extract_field_data($data_collection->field_is_sponsor);
+      $partner['copy'] = dosomething_helpers_extract_field_data($data_collection->field_partner_copy);
+      $partner['uri'] = dosomething_helpers_extract_field_data($data_collection->field_partner_url);
+
+      $logo = dosomething_helpers_extract_field_data($data_taxonomy->field_partner_logo);
+
+      if ($logo) {
+        $partner['media'] = [
+          'uri' => image_style_url('wmax-423px', $logo['uri']),
```
angaither, line 489: I don't think we enforce minimums on partner logos, so I have a feeling this might be a little large

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
```diff
@@ -444,6 +455,49 @@ protected function getMobileCommonsData() {
     ];
   }
 
+  protected function getPartners() {
+    $data = [];
+
+    // @TODO: only accounts for single partner at the moment.
+    // Update extract function to grab array of partners instead of just first id.
```
aaronschachter, line 462: Can we create a new Phoenix issue for this `TODO` ? Nothing we need right away but this one could be easy to forget that we never added in vLater.

aaronschachter:
Code looks good to me -- I think a `null` media object makes sense when we don't have a logo.

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
weerd, line 489: Yeah, but it's already what we use in the web app. On the DS we use CSS to help us tame the size of partner logos. I referenced the image "style" used in the ds_taxonomy module.

lib/modules/dosomething/dosomething_campaign/includes/Campaign.php
weerd, line 462: Yeah, sure. But I find issues are sometimes as susceptible to being forgotten. Would also like to leave the `@TODO` cause then it reminds us when we're inside the code. I could create an issue and reference it in the todo comment as well?

aaronschachter, line 462: Sounds good 👍

PR #6397, opened 2016-04-26
writes script to batch exclude reportbacks

### LINKED ISSUE #6392: Batch Exclude Reportbacks to Clear Campaign Inboxes
#### User Story

As a campaign lead, I want all fids on the nodes listed below with a quantity under _x_ to be marked as excluded before Thursday of this week.

Campaign (Node ID) // Threshold for Exclusion // Runs Affected
# SuperStressFace (2710) //  threshold: 10 // Run: 3110, 6660, 6470

Notes from Shawn (2805) // threshold: 250 // Run: 3494, 5828
Shower Songs (3590) //  threshold: 100 // Run: 5278, 6511
Trash Scavenger Hunt (2929) // threshold: 200 // Run: 3111
#### Why This Matters

The campaigns team doesn't have the bandwidth to clear out these inboxes. The most pressing concern is that Shower Songs is reopening in May and the inbox needs to be cleared before it launches again -- currently there are approx. 11,800 images that need to be reviewed.
#### Order of Operations
- [x] Nami provides impact threshold for four campaigns + verifies with campaign leads
- [x] Chloe writes code
- [x] Chloe and Nami test code on Thor
- [x] Ask Sena to backup production
- [x] Deploy to production and confirm!

cc @angaither @mikefantini

chloealee:
#### What's this PR do?

Provides a new custom drush command to batch exclude reportbacks to clear campaign inboxes. 
Command can be used for any campaign and campaign run by taking the following parameters: 
-`nid` - campaign nid 
-`run_nid` - campaign run nid 
-`quantity_limit` - threshold for exclusion - reportback quantity will be less than this number

run `drush batch-exclude-reportbacks` command with above parameters 
check campaign reportbacks and make sure only reportbacks that are under the quantity threshold remain in the reportback inbox

e.g. Run `drush batch-exclude-reportbacks 1321 1837 200`. On your local, visit http://dev.dosomething.org:8888/us/node/1321/rb and make sure only reportbacks with quantities over `200` remain in the inbox.

Fixes #6392


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.drush.inc
```diff
@@ -104,3 +116,55 @@ function drush_dosomething_reportback_ds_reportback_run_nid_fix() {
...
+ * Callback for batch-exclude-reportbacks command.
+ */
+function drush_dosomething_reportback_batch_exclude_reportbacks($nid, $run_nid, $quantity_limit) {
+    // Find all the fids that should be marked as excluded, based on params passed through. 
+    $fids_to_be_marked_as_excluded = db_query("SELECT rb.rbid, rb.nid, rb.run_nid, rb.quantity, rbf.status, rbf.fid, rbf.rbid
+                                               FROM dosomething_reportback rb
+                                               JOIN dosomething_reportback_file rbf on rb.rbid = rbf.rbid
+                                               WHERE rb.nid = :nid AND rb.run_nid = :run_nid AND rb.quantity < :quantity_limit AND rbf.status = 'pending'",
+                                               array(
+                                                ':nid' => $nid,
```
angaither: let's add some checks to make sure nid/run_id are actual nodes before running this

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.drush.inc
```diff
@@ -104,3 +116,55 @@ function drush_dosomething_reportback_ds_reportback_run_nid_fix() {
...
+                                                ':run_nid' => $run_nid,
+                                                ':quantity_limit' => $quantity_limit,
+                                               ));
+
+    // Change fids status to excluded 
+    foreach ($fids_to_be_marked_as_excluded as $fid) {
```
angaither: this chunk lives in a function somewhere, right? can we call that?

chloealee: the only other place this is in is in [this update](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.install#L918-L922) and is a tiny bit different because it doesn't include the array with the variables. 

But yes, all the below should be the same! Is it possible to call the code from the hook? Or, I can refactor maybe to write that all into one function and just call the function in both places?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.drush.inc
```diff
@@ -104,3 +116,63 @@ function drush_dosomething_reportback_ds_reportback_run_nid_fix() {
...
+
+/**
+ * Callback for batch-exclude-reportbacks command.
+ */
+function drush_dosomething_reportback_batch_exclude_reportbacks($nid, $run_nid, $quantity_limit) {
+
+    $campaign = node_load($nid);
+    $campaign_run = node_load($run_nid);
+    $campaign_exists = ($campaign != FALSE);
+    $campaign_run_exists = ($campaign_run != FALSE);
```
angaither: you can condense these four lines, and the if statement into one like this

```
// Load the nodes, and only if they are matching the type you're expecting.
if (node_load($nid, 'campaign') && node_load($run_nid, 'campaign_run')) {
 // query
}
else {
 drush_set_error('what are you doing, wild thing, those aren't the right nodes')
}
```

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.drush.inc
angaither: well, my feeling on this is since we are building out a review app, and we want to eventually POST to update `reportback_items` this should def be a function you can call

chloealee: ah I see! Would the best place to put this function be in `dosomething_reportback.query.inc`?

angaither:
👍

PR #6401, opened 2016-05-02
Fix bugs with Northstar auth verification.

DFurnes:
#### What's this PR do?

This pull request includes cleanup and fixes to the Northstar auth verification & sync code:

🐛 The "verify" check was previously being run on newly registered users (_before_ their profiles were synced to Northstar), resulting in "failed" validation and a bunch of users with `NONE` saved to their Northstar ID field. This PR moves this call to occur in a submit handler on the login form, rather than a Drupal hook. Also now just sends "username" and lets Northstar infer which type. (4a655c7)

📄 Updates docblocks to try to make it clearer when different hooks and submit handlers are run, since it was ambiguous from looking at the method names without plopping `dpm` hooks everywhere. (ab3ec96)

👉 Adds a helper method to save the "Northstar ID" profile field, since the syntax is kinda wonky and it's nice to not have the same code duplicated all over. (71c70e8)

🍃 Removes some commented out code from the Northstar Guzzle implementation, since it was making that file harder to skim. Once we revisit #5932, hopefully we can start using the Northstar PHP library here instead! (3f86eb6)

🍂 General tidying up. (c502379, a7df8ff, 939b0bb)

It's probably easiest to just test this on staging TBH. But there should be significantly fewer cases where users have `NONE` for their Northstar ID field, since it's now properly set on registration.

👻 

References #6260.

---

For review: @angaither @weerd

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -72,6 +72,17 @@ function dosomething_northstar_verify_user($credentials) {
 }
 
 /**
+ * Save the user's Northstar ID to their local profile.
+ *
+ * @param $user - Drupal user
+ * @param $northstar_response - Northstar user JSON response
+ */
+function dosomething_northstar_save_id_field($user, $northstar_response) {
+  $northstar_id = !empty($northstar_response->data->id) ? $northstar_response->data->id : 'NONE';
+  user_save($user, ['field_northstar_id' => [LANGUAGE_NONE => [0 => ['value' => $northstar_id]]]]);
+}
+
+/**
  * Send user registration events to northstar.
  */
 function dosomething_northstar_register_user($form_state) {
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -72,44 +72,50 @@ function dosomething_northstar_verify_user($credentials) {
 }
 
 /**
+ * Save the user's Northstar ID to their local profile.
+ *
+ * @param $user - Drupal user
+ * @param $northstar_response - Northstar user JSON response
+ */
+function dosomething_northstar_save_id_field($user, $northstar_response) {
```
angaither, line 80: we have some [getters](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_user/dosomething_user.module#L1171) and [setters](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_user/dosomething_user.module#L1235) in the user module, maybe this should live there?

DFurnes, line 80: Updated to use that method here in b27921d!

angaither:
👍

DFurnes:
Could've sworn I already posted this somewhere, but dropping it here for future reference.

Logging in to an existing account (from homepage or campaign signup) will verify user against `POST auth/verify` endpoint, and save returned ID to local field. This results in the following methods being called:

```
        dosomething_user_login_submit
        dosomething_user_northstar_login
        dosomething_northstar_verify_user
        dosomething_northstar_save_id_field
```

Registering a new user will send new user to the `POST users/` endpoint, and save returned ID to local field. This is the result of the following methods being called:

```
        dosomething_user_new_user
        dosomething_northstar_register_user
        dosomething_northstar_save_id_field
        dosomething_northstar_update_user
```

Going to the “edit profile” page will send update to the `POST users/drupal_id/:drupal_id` endpoint:

```
        dosomething_northstar_update_user
```

PR #6402, opened 2016-05-02
adds ->fetchAll(); after db_query to return the expected array of user objects

chloealee:
#### What's this PR do?

Added `$users = $users->fetchAll();` after `db_query` to return an array of user objects. 

Without this, when I did `print_r` for `$users`, it returned a `DatabaseStatementBase Object`. With this update, it will return an array of the user objects.

Check by adding `print_r($users);` and `die();` after these two new lines and run the script. Make sure an array of uses is returned. 

To run script: cd in vagrant box and within `/var/www/dev.dosomething.org/html` run `drush --script-path=../scripts/ php-script export-users-to-northstar.php`

### DIFF
--- a/scripts/export-users-to-northstar.php
+++ b/scripts/export-users-to-northstar.php
@@ -13,11 +13,13 @@ if ($last_saved) {
   $users = db_query("SELECT u.uid
             FROM users u
             WHERE uid > $last_saved");
+  $users = $users->fetchAll();
 }
 else {
   // Get all the users!
   $users = db_query('SELECT u.uid
                    FROM users u');
+  $users = $users->fetchAll();
 }
 
 foreach ($users as $user) {


DFurnes:
So strange! This works without the added `->fetchAll()` call on my local.

It _looks_ like the [`DatabaseStatementInterface`](https://api.drupal.org/api/drupal/includes%21database%21database.inc/interface/DatabaseStatementInterface/7.x) returned by the `db_query` function implements PHP's [`Traversable`](http://php.net/manual/en/class.traversable.php) interface so it should be able to be traversed via a `foreach`... what's the error it's giving?

chloealee:
ah okay - that makes sense, thanks @DFurnes ! I can close this out if all works correctly since it wasn't necessarily throwing an error, just didn't seem to export the users to NS. Now that I think of it though, I had to change the service to send to my local instead of NS qa so that was probably it and not this - sorry about that!

DFurnes:
All good! I was totally confused why it _was_ working at first haha. Re-open if you run into any issues! 😄

Issue #6406, opened 2016-05-04
Users aren't (always) getting ported to NS 


angaither:
#### Current Behavior

seeing a few (3) users who aren't being ported into gladiator for competitions and it looks like that's b/c the user doesn't exist in NS. (UIDS: `3812330`, `3814753`, `3844143`)
#### Desired Behavior

people should exist in NS after their accounts were created. 
#### Steps to Replicate

no idea.
#### Why This Matters

people can't live the lives they choose, and do things like sign up for the app, or get competitions messaging. 
#### Countries Affected (optional)

the whole damn 🌍

DFurnes:
Super weird. I see `3812330` and `3844143` in the request log with two 404s because it couldn't update a Northstar record with that UID... no records of having any issues migrating or creating them.

The first two accounts were registered on 04/08/2016, and the third on 4/17/2016... so they _should_ exist in the logs (although I did clear out a bunch of 200s at one point since the table was getting super large).

Hmmmm.

DFurnes:
And none of those Drupal IDs or emails exist in Northstar. I think it'd be a good idea to try to manually re-migrate each of those records to see if anything goes wrong, but I'm still at a bit of a loss for the root issue.

angaither:
I thought the update in northstar also created a new user if on didn't exist?

DFurnes:
@angaither We've been using the `POST users/drupal_id/:uid` route to explicitly update that particular Northstar account from Phoenix if their profile changes, so it doesn't "upsert" if the given UID isn't found.

angaither:
@DFurnes should the catch handle if the UID isn't found?

DFurnes:
Yeah, that'd probably be a good idea.

DFurnes:
When a user tries to update their profile, Phoenix will now catch 404s and try to create that missing Drupal ID (see #6411). I'm going to keep this open since we've probably got other root causes to patch up for this.

angaither:
@DFurnes I'm still seeing lots of 404s in the `gladiator_failed_task_log` 
I'm wondering if we do need to handle this is another way. those users aren't updating, they are normally either registering or signing up for a campaign and they are now always stuck in the failed log, unless we manually port them over.

angaither:
actually looking into these users I might have found a pattern they all have registration from niche
additionally these bugs are true for everyone I've checked so far
- they all are `en-global`users even though the address is United states (#6159)
[...]

PR #6411, opened 2016-05-06
More Northstar sync fixes

DFurnes:
#### What's this PR do?

This PR fixes a number of smaller issues with Northstar migration and syncing. The highlights:
- Fixes issue where `1234445555@mobile` "placeholder" emails were being sent with mobile users and causing validation errors on Phoenix. These are now ignored when forming the payload. (cc3640b)
- When forwarding new registrations, set `phoenix` as the signup source. This was previously only happening for batch migrated users. (0096db3)
- I also did a little refactoring to extract the `build_ns_user` call _out_ of the methods which make the HTTP requests to Northstar (c946d7b). This allows us to easily chain another request to register a user if the "update" request failed without having to try to re-build the whole user object. (9492787)
- Outputs progress to the console when migrating now! (d100755)

I'd recommend reviewing commit-by-commit since otherwise some of the changes get jumbled.

I've tested by manually checking that users are created/updated against my local Northstar, and by adding `dpm` calls to trace which hooks/methods get called.

I'm not 100% sure that triggering a "register" whenever an "update" call fails will work... I feel like there might be some sort of conflict that's preventing those users from being registered in the first place. _But_ it'll help surface that issue if so, and if not it's a nice way to "stop the bleeding" while continuing to investigate to figure out the underlying issue. 🔍 

References #6406 and #6260.

---

For review: @angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -87,25 +87,24 @@ function dosomething_northstar_save_id_field($user, $northstar_response) {
 
 /**
  * Send user registration events to northstar.
+ *
+ * @param $user - Drupal user object
+ * @param $payload - Array of data to be sent to Northstar
  */
-function dosomething_northstar_register_user($form_state) {
-  global $user;
-  // Build a user that Northstar expects.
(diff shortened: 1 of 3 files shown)


angaither:
👍

PR #6420, opened 2016-05-10
Replaced continue with a return.

### LINKED ISSUE #6416: PHP 7.0 Compatibility - Continue keyword
#### Current Behavior
- `'Error: 'continue' not in the 'loop' or 'switch' context in
  /var/www/www.dosomething.org/releases/20160510153234/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module, line 1212` error on deploy after upgrading to PHP 7.0. 
#### Desired Behavior
- There shouldn't be a PHP fatal error when running any code
#### Steps to Replicate
- It happens on deploy, but chances are it will happen whenever the reportback code is run.
#### Why This Matters
- We can't upgrade to PHP 7.0 until we fix all places where we use the `continue` keyword outside of a switch

https://jenkins-thor.dosomething.org/job/Deploy-Thor/405/console
https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module#L1212

angaither:
#### What's this PR do?

changes the `continue` inside of `dosomething_reportback_mark_reportback_as_excluded` to a  `return`

👀 

causing php upgrade to fail
#### Relevant tickets

Fixes #6416

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -1209,10 +1209,10 @@ function dosomething_reportback_mark_reportback_as_excluded($fid) {
   $reportback = reportback_load($fid->rbid);
 
   if (empty($reportback)) {
-    continue;
+    return;
   }
 
-  // Get all reportback item files associated with reportback. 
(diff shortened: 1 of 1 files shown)


blisteringherb:
👍

PR #6424, opened 2016-05-10
Codename: Cyclops

DFurnes:
#### What's this PR do?

👀 - 1. Y'know it's a cyclops.
#### How should this be reviewed?

😉 
#### Any background context you want to provide?

🐻 🔪 
#### Relevant tickets

See the [Cyclops](https://github.com/DoSomething/phoenix/milestones/Cyclops%20) milestone.
#### Checklist
- [ ] Documentation added for new features/changed endpoints.

PR #6435, opened 2016-05-13
Let those kudos do their thing.

DFurnes:
#### What's this PR do?

It was looking for a URL parameter, rather than the ol' JSON data. This PR fixes that. 🐛 

Look at those changes. 👀 

We were also sending an `Accepts` header which like _isn't_ a thing. jQuery seems to automatically sends the correct `Accept` header so no need to try to override that here.
#### Relevant tickets

References a bug in #6424.
#### Checklist


angaither:
LOVELY 💩

PR #6452, opened 2016-05-17
Don't save unhashed password on profile form.

DFurnes:
#### What's this PR do?

Fixes a bug with the user profile form saving incorrect data to the user's account.
#### How should this be reviewed?

Does it only save the things we want it to?
#### Any background context you want to provide?

🍃 
#### Relevant tickets

Fixes #6444.
#### Checklist
- [ ] Documentation added for new features/changed endpoints.

PR #6453, opened 2016-05-17
Refactor save ID method to load user object via UID.

DFurnes:
#### What's this PR do?

Refactors the `dosomething_northstar_save_id_field` method to ask for a UID rather than user object, and then load that unmodified user account in order to save the updated Northstar ID to it.

This should still work the same way, but will no longer cause [warnings on undefined fields](https://cloud.githubusercontent.com/assets/583202/15331858/433a4814-1c31-11e6-9e6c-a1415951a90a.png).

References #6452.
#### Relevant tickets

🍃
#### Checklist


angaither:
👍

Issue #6444, opened 2016-05-16
Extra Perms on import


angaither:
#### Current Behavior
- imported users are getting treated extra special
#### Steps to Replicate
- how can you make it happen every time?
#### Why This Matters
- everyone should be authenticated, but nothing more
#### Relevant Screenshots + Links

[check this doc for patterns](https://docs.google.com/spreadsheets/d/1RkGdi17BS-iBEU9IFZK6a-T86Dg-JZtYmFIrgAfnG_U/edit#gid=0)

PR #6455, opened 2016-05-17
Move the randomly created `name` field to registration form only

### LINKED ISSUE #2753: User username not always saved as UID

angaither:
#### What's this PR do?

move the random string name thing to user registration only, since this resets the user name to a random string any time a user updates their profile

👀 

noticed this while debugging #6444 
#### Relevant tickets

Fixes #2753


DFurnes:
👍

PR #6456, opened 2016-05-17
Watchdog logging and sending email for user with added perms

blisteringherb:
#### What's this PR do?

I've added code for watchdog logging and sending an email to devops when a user adds any permissions other than authenticated user.

You can add perms to a user to see if you receive the email and check to make sure the watchdog entry has been logged.

Users had the ability to give themselves additional permissions via the API, which is now fixed, but this gives us some additional insight so we can take action immediately if a problem like this occurs again.
- Logs via watchdog with which permissions were added and includes
  a link to the user
- Sends an email to devops with which roles got added and a link to the
  user that was added.

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -831,37 +831,3 @@ function dosomething_helpers_get_current_campaign_run_for_user($nid, $user = NUL
   // @TODO: may want to investigate how it loads the proper translation...
   return node_load($run['id']);
 }
-
-function dosomething_helpers_user_update(&$edit, $account, $category) {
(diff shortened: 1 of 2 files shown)


sheyd:
👍

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -831,3 +831,37 @@ function dosomething_helpers_get_current_campaign_run_for_user($nid, $user = NUL
   // @TODO: may want to investigate how it loads the proper translation...
   return node_load($run['id']);
 }
+
+function dosomething_helpers_user_update(&$edit, $account, $category) {
```
angaither: can we move this to `dosomething_users.module` also, 

``````
/**
 * comments plz
 */```
``````

PR #6462, opened 2016-05-18
Oh behave, API.

### LINKED ISSUE #6450: Able to create duplicate mobile users via API.
#### Current Behavior

Requests to the `api/v1/users` endpoint will allow you to create duplicate records for the mobile field.
#### Desired Behavior

It should not let you do this.
#### Steps to Replicate
1. Make an API request to the `api/v1/users` endpoint.
2. Then, do it again with the same mobile.
3. Repeat as many times as desired.
4. React in horror to the disaster you have wrought! 😨 ☎️ 
#### Why This Matters

We don't want dupes. Dupes are bad.

👮

DFurnes:
#### What's this PR do?

This fixes some bugs with the `POST api/v1/users` endpoint:
1. We previously didn't have any validation to prevent creating duplicate mobile users via the API.
2. We also didn't previously "clean" mobile numbers to strip out non-numeric characters like we do from the web registration form.

Try hitting those endpoints. You should no longer be able to do the aforementioned bad things.

It'd be great if we could share the same business/validation logic anywhere we create a user. I'm not sure if there's any idiomatic way of doing that in Drupal, or if it's worth taking the risk of refactoring things too much though.
#### Relevant tickets

Fixes #6450.
#### Checklist

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
@@ -174,7 +174,7 @@ function _member_resource_create($account) {
     'first_name' => $account['first_name'],
     'country' => dosomething_settings_get_affiliate_country_code(),
     'last_name' => !empty($account['last_name']) ? $account['last_name'] : NULL,
-    'mobile' => !empty($account['mobile']) ? $account['mobile'] : NULL,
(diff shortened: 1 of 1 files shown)


angaither:
we need some setters! 

👍

DFurnes:
Yeah, that'd be great. Merging, but we should continue thinking about this. 💧

sheyd:
lmg4y_f-thumbnail-100-0

PR #6463, opened 2016-05-18
Update user endpoint to accept country code and language.

### LINKED ISSUE #6442: Externally created users should have the right language
#### Current Behavior
- users who are created programaticaaly get assigned `global-english` even if we have their address as being in the US
- because global english is the default, they all are being assigned the default language
#### Desired Behavior
- allow external services to either set a language, or have drupal infer the correct language if an address is provided

refs #6159

DFurnes:
#### What's this PR do?

This updates the API endpoint to _not_ try to use the Fastly header to set a user's country code. Since these requests are (almost?) exclusively coming from our own internal services, this will always just set the user's country to the location of the server (US).

The endpoint now accepts a `country` value, which will be saved to the user's address field. If nothing is provided for that field, it defaults to `US`.

Hit that endpoint. Does it now save a country code and language? It should.

Should it default to US? Or should we try to save `null` to that user's country? Also, should we consider allowing this endpoint to accept a full address?
#### Relevant tickets

Fixes #6442.
#### Checklist

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
@@ -180,19 +180,8 @@ function _member_resource_create($account) {
 
   dosomething_user_set_fields($edit, $fields);
 
-  // Set user's language if provided, or infer from country code
-  if(! empty($account['language'])) {
-    $languages = array_keys(variable_get('dosomething_global_language_map', ''));
-
-    // Only allow whitelisted languages.
-    if(! in_array($account['language'], $languages)) {
-      return services_error(t('The language @language is not supported.', ['@language' => $account['language']]), 422);
-    }
-
-    $edit['language'] = $account['language'];
-  } else {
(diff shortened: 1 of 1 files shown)


angaither:
@DFurnes mind knocking out #6442 along with this?

angaither:
oh, i see you said fixes that, but this just sets the country, and not the matching language, right?

DFurnes:
Oh, I failed reading comprehension there. Will add that part into this PR too. 😁

DFurnes:
@angaither Updated that in 1c77cd8. Let me know if you think there's a better approach.

Also, will update wiki documentation once this is merged.

angaither:
@DFurnes do you think we should allow a language input? I was thinking we should just set it based on the country

DFurnes:
Oh, that seems fine too! And is a bit more consistent with how we do things on the web.

DFurnes:
@angaither And removed in dce0cf9! Cleans things up quite nicely.

angaither:
👍 🚢 🌍

PR #6465, opened 2016-05-18
Fix warnings on pulling staging database.

### LINKED ISSUE #6312: Warnings when pulling from staging
#### Current Behavior

When you pull updates from the staging database, some warnings are fired because null is given as the second argument to `array_key_exists` [on line 80 in `preprocess_node.inc`](https://github.com/DoSomething/phoenix/blob/8fc28b63d4657d6a71ce6d79a9b2e38a1a1a7085/lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc#L80), since the Fastly header is not set and so neither is `$country_contact_emails`.

```
array_key_exists() expects parameter 2 to be array, null given preprocess_node.inc:80
```
#### Desired Behavior

There should not be warnings.
#### Steps to Replicate

When you run `ds pull stage --db`, you'll see a bunch of errors which shouldn't be there.
#### Why This Matters

Warnings are ugly. :warning: :no_good: 

screen shot 2016-03-23 at 5 14 55 pm

DFurnes:
#### What's this PR do?

This only attempts to check the `$country_contact_emails` array when there actually is a country code to check for, silencing warnings during the DB pull command.

If you run `ds pull stage --db` then you shouldn't see errors.

🌊 
#### Relevant tickets

Fixes #6312.
#### Checklist

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
@@ -69,15 +69,13 @@ function paraneue_dosomething_add_info_bar(&$vars) {
   if (module_exists('dosomething_global')) {
     $country_code = dosomething_global_get_current_prefix();
 
-    if ($country_code) {
-      $country_contact_emails = array(
-        'br' => 'ajuda@dosomething.org',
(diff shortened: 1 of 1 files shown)


angaither:
:shipit:

Issue #6409, opened 2016-05-04
Fix issue where multiple Drupal users share the same email.


DFurnes:
#### Current Behavior

There are currently a bunch of users (both old and new) that are registered with the same value in the `mail` column in the users table. This is kinda a bummer since we should really only have one account per email. 😞 📬 
#### Desired Behavior

We need to do some more research on how Drupal is currently treating users with duplicate emails (like when you try to log in with that email, which account does it pick?), and then remove the duplicate users who are not being actively used (and perhaps merge their activity into the "active" accounts?)
#### Steps to Replicate

Try running this real outrageous 🐬 query:

``` sql
SELECT group_concat(u.uid), u.name, u.mail, m.field_mobile_value, s.field_user_registration_source_value, from_unixtime(u.access), count(u.mail)
FROM users u
LEFT JOIN field_data_field_mobile m on m.entity_id = u.uid 
left join field_data_field_user_registration_source s on s.entity_id = u.uid
GROUP BY u.mail
HAVING COUNT(u.mail) > 1
ORDER BY u.mail
```
#### Why This Matters

When trying to sync users with Northstar, we're getting validation errors because we're trying to make multiple users that all share the same "unique" field. It's also super confusing how this even _works_ in Phoenix so it'd be nice to figure out that mystery!

📪 🌵

DFurnes:
New query that seems to be a bit more helpful:

``` sql
SELECT u.uid, u.mail, from_unixtime(u.created), from_unixtime(u.access), m.field_mobile_value, s.field_user_registration_source_value FROM users u
LEFT JOIN field_data_field_mobile m ON m.entity_id = u.uid 
LEFT JOIN field_data_field_user_registration_source s ON s.entity_id = u.uid
WHERE mail IN (SELECT mail FROM users GROUP BY mail HAVING COUNT(mail) > 1);
```

mshmsh5000:
In case we haven't posted the explanation elsewhere:
- There's no `UNIQUE` constraint on `mail` in the `users` table, so MySQL's not gonna stop you.
- Drupal's default user form validation will [catch duplicate email addresses](https://github.com/drupal/drupal/blob/7.x/modules/user/user.module#L1273), so these are most likely from when we've bypassed either that validation or form submission entirely (API calls?).

angaither:
Most (if not all) of the cases appear to be two (or more) users created right in a row. the ol' double ping. the clickly-click. 

[...]

PR #6475, opened 2016-05-19
Add script to remove duplicate emails.

DFurnes:
#### What's this PR do?

This PR adds a simple script to remove any duplicate emails in the database.

Check out the query/script and make sure I'm not doing something silly. I've checked a random sampling of the emails for accounts matched by the query, and none of them are the accounts that are being resolved by Drupal on login.

Once we run this to remove existing dupes, we should update the schema to make this column have a `UNIQUE` constraint to make sure this can't happen again.
#### Relevant tickets

References #6409.
#### Checklist

### DIFF
--- a/scripts/remove-duplicate-emails.php
+++ b/scripts/remove-duplicate-emails.php
@@ -6,7 +6,9 @@
  * drush --script-path=../scripts/ php-script remove-duplicate-emails.php
  */
 
-$users = db_query('SELECT * FROM users u WHERE mail IN (SELECT mail FROM users GROUP BY mail HAVING COUNT(mail) > 1);');
+$users = db_query('SELECT * FROM users 
+  WHERE mail IN (SELECT mail FROM users GROUP BY mail HAVING COUNT(mail) > 1)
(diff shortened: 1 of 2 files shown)


scripts/remove-duplicate-emails.php
```diff
@@ -0,0 +1,29 @@
+<?php
+/**
+ * Script to remove users with
+ *
+ * to run:
+ * drush --script-path=../scripts/ php-script remove-duplicate-emails.php
+ */
+
+$users = db_query('SELECT * FROM users
```
angaither: two things
1. I would say only select UID (since we don't need the whole dang row)
2. I had trouble running [this](https://github.com/DoSomething/phoenix/blob/dev/scripts/export-users-to-northstar.php) script because that table is giant on prod, so I would suggest setting a var on the last user deleted and order by UID to start off where it died

DFurnes: 1. Good call. ❌ 
2. Ooof, yeah 203 seconds. Since it's iterating over a heavily filtered set of UIDs (rather than the whoooooole thing like the other script), I'm not sure saving a UID would help out too much. But on the bright side, we can easily just keep running this script over and over until there's no more dupes.

angaither:
👍  works for meeee

PR #6476, opened 2016-05-20
User params

### LINKED ISSUE #6430: Detect if a sign up is from a new account or an old one
We could display different experiences for people who just created an account to sign up for a campaign vs those who have already been members.

**New Folks:** If a user signs up for an account and then signs up for a campaign we could display more of an onboarding experience for them. 

**Old Folks:** If a member has already had an account, we could cut right to showing them other campaigns to sign up for.

chloealee:
#### What's this PR do?
- Adds new function to determine if this is a user's first signup for a campaign. 
  - If it is a user's first signup, the signup takeover experience will be enabled. 
  - If a user has previously signed up for a campaign, they'll get the regular "You're sign up! Get started below." messaging. 
- Passes `userStatus` param front backend to JS to indicate if this is a user's first signup for a campaign. 
  - Two options for this param:
    - `user-status-existing`
    - `user-status-new`
- From an existing account, signup for a campaign.
  - You should get the "You're signed up!" messaging in an alert at the top of the page. 
  - Open up the dev tools console. Type in `Drupal.settings.dosomethingUser` and the `userStatus` should be `user-status-existing`
- From a new user account, sign up for a campaign. 
  - The signup takeover should be enabled. 
  - `Drupal-settings-dosomethingUser`'s `userStatus` should be `user-status-new`. 

In the future, we can change the experience of the existing user (maybe they still get the signup takeover screen but they just skip to the multi-signup page instead of the onboarding experience). This is just a prototype and sets up params to pass user information from backend to JS. 
#### Checklist

Fixes #6430

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -10,7 +10,7 @@
  */
 function dosomething_campaign_preprocess_node(&$vars) {
   if ($vars['type'] != 'campaign' || $vars['view_mode'] != 'full') return;
-  dosomething_signup_set_takeover();
+  // dosomething_signup_set_takeover();
   $node = $vars['node'];
   $vars['campaign'] = dosomething_campaign_load($node);
 
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -416,6 +416,8 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
     array(
       'dosomethingCampaign' => array(
         'recommendedCampaigns' => $signup_takeover_recommended_campaigns,
+      ),
+      'dosomethingUser' => array(
         'userStatus' => $user_status,
       ),
     ),
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -10,7 +10,6 @@
  */
 function dosomething_campaign_preprocess_node(&$vars) {
   if ($vars['type'] != 'campaign' || $vars['view_mode'] != 'full') return;
-  // dosomething_signup_set_takeover();
   $node = $vars['node'];
   $vars['campaign'] = dosomething_campaign_load($node);
 
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -387,6 +387,7 @@ function dosomething_campaign_preprocess_media_vars(&$vars, &$wrapper) {
  */
 function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
   $campaign = $vars['campaign'];
+  $user = $vars['user'];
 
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -401,12 +401,24 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
     array_push($signup_takeover_recommended_campaigns, $campaign_array);
   }
 
-  // Send 3 campaigns to signup takeover, prompting user to signup for multiple campaigns.
+  // Check to see if this is a user's first signup for a campaign.
+  $existing_user = dosomething_signup_not_first_signup($user->uid);
+  if ($existing_user) {
+    $user_status = 'repeat-signup-existing';
```
sbsmith86: I wonder if this should just be more general `existing-user-signup` and then `new-user-signup` cause it doesn't really matter if it is a repeat signup until we get into the js.

sbsmith86: or even `user-status-new` vs `user-status-existing`?

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -581,8 +581,33 @@ function dosomething_signup_user_signup($nid, $account = NULL, $source = NULL) {
...
+      dosomething_signup_set_takeover();
+    }
+  }
+}
+
+/**
+ * Returns true if user this is not a user's first signup.
+ * @param int $user uid
+ */
+function dosomething_signup_not_first_signup($uid) {
```
sbsmith86: This might be a little nitpicky - I feel like this should be flipped like `dosomething_signup_first_signup()` and return `true` if this is the users first sign up. 

I think it would make for more readable code to have to not check for the false positive when checking if this is the users first sign up. So you can do...

```
if (dosomething_signup_first_signup) {
dosomething interesting and different.
} else {
dosomething else. 
}
```

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
chloealee: yeah good call then it can be used for other things too!

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
chloealee: that makes total sense! I was even getting a little confused when writing it / had to double check it a ton. I'll change that now!

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -581,8 +581,31 @@ function dosomething_signup_user_signup($nid, $account = NULL, $source = NULL) {
...
+    else {
+      dosomething_signup_set_signup_message();
+    }
+  }
+}
+
+/**
+ * Returns true if user this is a user's first signup.
+ * @param int $user uid
+ */
```
sbsmith86, line 600: Probably want to doc with `@return bool`

sbsmith86:
One more little comment but otherwise looks good!

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
chloealee, line 600: ah right! Thanks!

chloealee:
Thanks, @sbsmith86 ! Cool to merge now?

sbsmith86:
@chloealee yas

Issue #6477, opened 2016-05-20
Warning on array_diff_key when calling user_save.


DFurnes:
#### Current Behavior

Modifying a user _without_ changing their role results in a warning.
#### Desired Behavior

No warnings. 🔕  
#### Steps to Replicate

If you edit a user and the `$edit` array does not contain the `roles` key, you get the following warning triggered by [this line attempting to read a value from an array before checking that the key exists](https://github.com/dosomething/phoenix/blob/beff1f11738d478d9fedcce0b178056d0804f37d/lib/modules/dosomething/dosomething_user/dosomething_user.module#L1713-L1713):

```
array_diff_key(): Argument #1 is not an array dosomething_user.module:1713
```
#### Why This Matters

Because warnings.

PR #6478, opened 2016-05-20
Fix mobile fields

### LINKED ISSUE #6408: Fix users with `@mobile` email but no phone stored in mobile field.
#### Current Behavior

A couple of people have xxxxxx@mobile generated emails, but nothing set in the `field_mobile_value` field on their profile.
#### Desired Behavior

Since we know these user's phone numbers (but don't have it properly recorded), we should fix this by attaching the proper phone number to each of these profiles.

I think the most likely solution would be to write a simple Drush script that could iterate over these users (based on the UIDs returned from the below query) and fix each account.
#### Steps to Replicate

Check out this 🐬 query:

``` sql
SELECT u.uid, u.name, u.mail, u.access, m.field_mobile_value FROM users u
left join field_data_field_mobile m on u.uid = m.entity_id
WHERE mail LIKE '%@mobile'
AND m.field_mobile_value IS NULL
```
#### Why This Matters

Since we check against the `field_mobile_value` when letting people log in with their mobile numbers, these users cannot log in. We also can't synchronize them to Northstar because we don't have a valid email or mobile number without doing this extra work.

💃

DFurnes:
#### What's this PR do?

This pull request adds a script which attempts to fix users who have a `xxxxx@mobile` email address but nothing set for their mobile field. A quick spot-check shows these users do exist in Mobile Commons, so it'd be good to fix our data for them on our end.

This pull request also includes a small fix for #6477.

Read through that script and consider how it might work/not work when run against the results of that query on the production database. Are there things I'm missing?

☎️ 
#### Relevant tickets

Fixes #6408 and #6477.
#### Checklist

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -1710,6 +1710,18 @@ function dosomething_user_views_pre_render(&$view) {
  * Implements hook_user_update().
  */
 function dosomething_user_user_update(&$edit, $account, $category) {
+  if (!empty($edit['roles'])) {
+    _dosomething_user_audit_roles($edit, $account);
+  }
+}
+
+/**
+ * Audit changes to user roles and send an alert to DevOps.
+ *
+ * @param $edit
+ * @param $account
+ */
+function _dosomething_user_audit_roles(&$edit, $account) {
   $added_rids = array_diff_key($edit['roles'], $edit['original']->roles);
   if ($added_rids) {
     $added_roles = array();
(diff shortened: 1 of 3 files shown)


angaither:
code looks 🆗  
but, I wonder if/what we should set the mail to? Does NS ignore that mail field?

DFurnes:
Yes! ...well, sort of! We remove that field when building the user object [in the export script](https://github.com/DoSomething/phoenix/blob/6f3d919ba657acd5dd0dba4a55dc6499854c6eab/scripts/export-users-to-northstar.php#L100-L105) but I forgot to add the same thing when syncing people on demand. Will make an issue for that. 😓

DFurnes:
Added #6480 to track that.

PR #6484, opened 2016-05-23
Composer

DFurnes:
### What's this PR do?

This pull request revisits the Composer experiments that we tried out in #5932, sans crazy merge conflicts. This lets us more easily include Composer packages (yay!) and also means that Composer can resolve version ranges and check dependency conflicts like it's supposed to!

This moves (most) Composer dependencies into the gitignored `vendor/` directory and installs them in a single step during the `bin/ds` install process. Drupal then uses Composer's `autoload.php` to load all those classes in one fell swoop.

Pull down the branch and run `ds build --install`. It should install dependencies & work as expected.

We should probably add this to the docs somewhere, but (if merged) this means that dependencies can be installed by simply adding them to `composer.php` (y'know like you'd do in any other modern PHP project) rather than updating a Drupal makefile.

Drupal modules, of course, still work the same as they always have.
#### Relevant tickets

🎼 
#### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,5 @@ npm-debug.log
 Gemfile.lock
 .fuse_hidden*
 lib/themes/dosomething/paraneue_dosomething/dist/
-lib/themes/dosomething/paraneue_dosomething/bower_components/
-lib/themes/dosomething/paraneue_dosomething/vendor/
 tmp/
-.sass-cache
+vendor/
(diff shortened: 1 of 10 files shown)


DFurnes:
Not quite sure why, but StyleCI seems to be having issues running on this branch. 😞

DFurnes:
I tested a fresh install on my local, and Zendesk works as expected (ticket showed up in the Help Center). We should also test this pretty thoroughly on Staging & Thor if we decide to merge.

PR #6485, opened 2016-05-23
Use Composer for library dependencies.

DFurnes:
### What's this PR do?

This pull request revisits the Composer experiments that we tried out in #5932, sans crazy merge conflicts. This lets us more easily include Composer packages (yay!) and also means that Composer can resolve version ranges and check dependency conflicts like it's supposed to!

This moves (most) Composer dependencies into the gitignored `vendor/` directory and installs them in a single step during the `bin/ds` install process. Drupal then uses Composer's `autoload.php` to load all those classes in one fell swoop.

Pull down the branch and run `ds build --install`. It should install dependencies & work as expected.

We should probably add this to the docs somewhere, but (if merged) this means that dependencies can be installed by simply adding them to `composer.php` (y'know like you'd do in any other modern PHP project) rather than updating a Drupal makefile.

Drupal modules, of course, still work the same as they always have.
#### Relevant tickets

🎼 
#### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,5 @@ npm-debug.log
 Gemfile.lock
 .fuse_hidden*
 lib/themes/dosomething/paraneue_dosomething/dist/
-lib/themes/dosomething/paraneue_dosomething/bower_components/
-lib/themes/dosomething/paraneue_dosomething/vendor/
 tmp/
-.sass-cache
+vendor/
--- a/bin/ds
+++ b/bin/ds
@@ -406,24 +406,14 @@ function deploy {
 function composer {
   if hash composer 2>/dev/null;
   then
-    cd $WEB_PATH/profiles/dosomething/libraries/zendesk
-    echo -e "\e[4mInstalling composer dependencies for Zendesk...\e[0m"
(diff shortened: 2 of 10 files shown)


DFurnes:
Remade this from #6484 since that PR was borked on StyleCI and the branch wasn't accepting new commits. Also from the other PR:

> I tested a fresh install on my local, and Zendesk works as expected (ticket showed up in the Help Center). We should also test this pretty thoroughly on Staging & Thor if we decide to merge.

🚥 💁

DFurnes:
Oh, also worth mentioning and documenting if merged: this PR also adds a `lib/app` directory which is used as a PSR-4 namespace root for any `Phoenix\` classes we may want to add (for example, for trying to pull some domain logic out of Drupal hook functions).

DFurnes:
@angaither @mshmsh5000 Would love each of your Drupal eyes on this whenever you have time!

angaither:
oooo-la-la. 
I'm in

mshmsh5000:
Build looks good and works! 👍 🍔

weerd:
conan-lever-5-o

DFurnes:
✌️🔀

PR #6489, opened 2016-05-23
Return pass reset link + login.

### LINKED ISSUE #6470: Fix Niche Flow
#### Current flow
1. User is created via the api
2. Sent a link to reset password
3. Click the dumb login link
4. Reset password/edit profile
#### Easier flow
1. User is created via the api
2. Sent a link to reset password
3. Skip the login thing, and just reset password/edit profile

[There's a module for this](https://www.drupal.org/project/simple_pass_reset) but it looks like it has some serious excess code. 
We could use that, or write something ourself, and override the edit form to display more relevant things
#### Why This Matters

The current flow ~sucks~
Also, we could probably make #1003 while we are at it

angaither:
#### What's this PR do?

changes the `password_reset_url` to return a login in link instead of the `are you sure you want to login` page

👀 

@aaronschachter not sure if you're using this anywhere in the mobile app! 
#### Relevant tickets

Fixes #6470 

you get this
image

instead of
image

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
@@ -305,6 +305,6 @@ function _member_resource_get_activity_result($uid, $nid) {
  */
 function _member_resource_reset($uid) {
   $account = user_load($uid);
-  $reset_url = user_pass_reset_url($account);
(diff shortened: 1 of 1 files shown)


aaronschachter:
👀 👍 

The mobile app's "Reset password" buttons open up the `user/password` URL in a web browser outside of app.

PR #6493, opened 2016-05-24
Updates to remove unnecessary button from finder on mobile view.

### LINKED ISSUE #6375: Remove Find a Campaign button on mobile homepage
It doesn't functionally do anything since we results update as soon as you select them...

cc: @DFurnes

weerd:
#### What's this PR do?

This PR removes the "Find A Campaign" button that would only show up in the narrow/mobile view of the campaign finder on the homepage. Along with removing the button it also removes the JavaScript code that enabled the button.

Run the site, try the finder out in a narrow/mobile view and see if the button shows up. Does the finder still work? Great!

It seems that [at some point](https://github.com/DoSomething/phoenix/blob/da7fd6194bf36c402bd1f18a2978f352b12448fd/lib/themes/dosomething/paraneue_dosomething/js/finder/FormView.js#L115) the button was meant to behave as a necessary item when viewing the site on mobile devices and to be able to properly submit the form. However, since then, the finder has changed a bit and it will auto-filter and update whenever any selection in the dropdowns is made both on mobile and desktop, thus the button is no longer needed. If JavaScript fails for whatever reason, the entire finder would not work and the button would serve no purpose since the filtering is handled entirely on the client side.
#### Relevant tickets

Fixes #6375

---

@DFurnes @angaither

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/finder/FormView.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/finder/FormView.js
@@ -27,9 +27,6 @@ define(function(require) {
     // The div where the form goes
     $div: null,
 
-    // <button> that mobile users click to initiate the search
-    $searchButton: null,
-
     // The field groups that contain the checkmarks
     $fields: {},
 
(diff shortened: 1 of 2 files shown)


DFurnes:
👍

PR #6495, opened 2016-05-24
Fix syncing `@email` users.

### LINKED ISSUE #6480: Properly handle `@@mobile` emails when syncing in the Northstar module.
#### Current Behavior

We remove the "email" field when it matches the `@@mobile` pattern [in the export script](https://github.com/DoSomething/phoenix/blob/6f3d919ba657acd5dd0dba4a55dc6499854c6eab/scripts/export-users-to-northstar.php#L100-L105) but don't do that when syncing users on-demand in the Northstar module.
#### Desired Behavior

The `dosomething_northstar_build_ns_user` should also use the same logic to remove that field if the user has a fake "mobile" email address.
#### Steps to Replicate

Register a user through the API that has a `@@mobile` email address. It will be properly registered in Drupal but will cause a validation error when syncing to Northstar.
#### Why This Matters

Currently we're not syncing those users to Northstar.

📶

DFurnes:
#### What's this PR do?

This pull request fixes syncing `@mobile` users, which previously only got transformed when manually migrating users. It also extracts shared code between those two into a shared "transformation" function that then gets called for each unique use case, since there was like a ton of duplication.

Make sure I didn't miss anything when extracting shared logic into the new `dosomething_northstar_transform_user` function, and that each of the original functions still add their special sauce.

🎎 
#### Relevant tickets

Fixes #6480.
#### Checklist

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -158,7 +158,7 @@ function dosomething_northstar_update_user($user, $payload) {
 
   // Add to request log if enabled.
   dosomething_northstar_log_request('update_user', $user, $payload, $response);
-  
+
   if ($response->code === '200' && module_exists('stathat')) {
     stathat_send_ez_count('drupal - Northstar - user updated - count', 1);
   }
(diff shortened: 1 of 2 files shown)


angaither:
👍

PR #6497, opened 2016-05-25
Adds missing user and node variables

### LINKED ISSUE #6460: Undefined variables in _dosomething_user_add_signup_data
#### Current Behavior

In `_dosomething_user_add_signup_data`, the variables `$user` and `$node` are [referenced on line 895](https://github.com/dosomething/phoenix/blob/39233a8d142dd6fd021b1472fce1070a8a6fffee/lib/modules/dosomething/dosomething_user/dosomething_user.module#L894-L895) but don't seem to exist in the current scope. It looks like the `dosomething_global_get_language` function handles this fine by being extra careful about parameters passed to it, but it'd be nice to clean up. 🍃 
#### Desired Behavior

We should not reference variables that don't exist. 👀

itsjoekent:
#### What's this PR do?

Adds two variables missing in the add signup data function.

Signup on all of the things, are the params added properly?

This has been working because the function call to get_language is pretty good at non failing from NULL values, but as Dave noted in the issue we shouldn't be passing non existent vars intentionally!
#### Relevant tickets

Fixes #6460

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -892,6 +892,8 @@ function _dosomething_user_add_signup_data(&$form) {
     }
 
     // Use the global url as the form action so the user gets redirected to the right page.
+    global $user;
+    $node = node_load($nid);
     $lang_code = dosomething_global_get_language($user, $node);
 
     $url_options = array(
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -892,6 +892,8 @@ function _dosomething_user_add_signup_data(&$form) {
     }
 
     // Use the global url as the form action so the user gets redirected to the right page.
+    global $user;
```
angaither: this should be at the top of the function

itsjoekent: gotcha, didn't see it used anywhere else so figured it was ok

itsjoekent: fixed!

angaither:
👍

PR #6500, opened 2016-05-25
Organ donation module

### LINKED ISSUE #6491: Set up organ donation module
#### User Story

As dosomething devs, staff, and other interested parties when a user is presented with the organ donation modal we want to track the user's registration. 
#### Implementation deets
- Set up a module `dosomething_organ_donation`
- Create a new table `dosomething_organ_donation`
  - sid
  - uid
  - nid (?)
  - run_nid (?)
  - registration_timestamp (store time of valid registration)
  - Screen they dropped out off 
  - Other stuff maybe? (Dependent on what we get back from the Organize api)
- Add menu hook to new module with an endpoint the frontend forms can post to with a callback that will store the data we need. 
#### Why This Matters

We think it would be best to set up a new module/table to track organ donation stuff rather then say, appending more columns to the signup table because 1) this is so specific to this one case and 2) we might need to track different things

sbsmith86:
#### What's this PR do?

This PR adds a new module `dosomething_organ_donation`. 

This module sets up a new table that we will use for tracking organ donation registration data. With this PR we will be able to track: 
- The signup id of the registration
- The user id
- the timestamp of when they registered
- the screen the user was on when the registration modal closes

TBD if we will need to track more thangs.

This module also sets up an endpoint, `organ-donation/registration` that we can hit from js to store that data. It accepts two parameters so far `sid` and `uid`

TBD what else we will be passing to this endpoint.

Pull down, enable the new module, make a get request with the required params, `sid` and `uid` and you should get a response back that the registration was stored...

Also you should see your info in the DB in the `dosomething_organ_donation` table.

Still waiting on a response from Organize about getting access to their api so we can see what other info we might want to track. 
#### Relevant tickets

Fixes #6491 

cc @chloealee 
#### Checklist

### DIFF
--- a/lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
+++ b/lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
@@ -42,12 +42,15 @@ function dosomething_organ_donation_store_registration() {
       $response = [
         'status' => 'Required parameters not found.',
       ];
+
+      watchdog('dosomething_organ_donation', $response, NULL, WATCHDOG_ERROR);
     }
 
     drupal_json_output($response);
   }
   catch (Exception $e) {
     drupal_json_output($e);
+    watchdog('dosomething_organ_donation', $e, NULL, WATCHDOG_ERROR);
   }
 
   return;
--- a/lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
+++ b/lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
@@ -22,7 +22,7 @@ function dosomething_organ_donation_menu() {
 }
 
 /*
- * Stores meta data about the organ registration in the dosomething_organ_registration table.
+ * Stores meta data about the organ registration in the dosomething_organ_donation table.
  *
  */
 function dosomething_organ_donation_store_registration() {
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
```diff
@@ -0,0 +1,57 @@
...
+/*
+ * Stores meta data about the organ registration in the dosomething_organ_registration table.
+ *
+ */
+function dosomething_organ_donation_store_registration() {
+  try {
+    if ($_GET['sid'] && $_GET['uid']) {
+      db_insert('dosomething_organ_donation')
+        ->fields([
+          'sid' => $_GET['sid'],
```
angaither: should use `drupal_get_query_parameters`

sbsmith86: Ahh good to know. will do

sbsmith86:
@angaither updated

angaither:
in order to turn the module on with the deployment, you need to add the module to the [make file](https://github.com/DoSomething/phoenix/blob/dev/lib/profiles/dosomething/dosomething.make) and the [info file](https://github.com/DoSomething/phoenix/blob/dev/lib/profiles/dosomething/dosomething.info)

sbsmith86:
@angaither AHHH thanks! updated.

angaither:
👍

PR #6502, opened 2016-05-25
Add some validation to the users endpoint.

DFurnes:
#### What's this PR do?

This pull request should wrap up the last fixes to the `POST api/v1/users/` endpoint.

🍎 Validates email addresses (so you can't register with an email that'll later be rejected elsewhere).

☎️ Return validation errors for invalid phone numbers (rather than silently discarding the field).

🔏 Don't allow blank password field to be set. I don't think this should cause any issues since a [fake password is set](https://github.com/DoSomething/mbc-user-import/blob/89050bd5235e9f61a7fce30f88ec434f01fda66b/src/MBC_UserImport_Source_Niche.php#L91) in the Niche importer, and any "blank" passwords would never be able to be used because empty strings aren't hashed in `user_save`.

🔢 Save UID as the user's "name" like we do elsewhere.

Try to create bad accounts from that endpoint. Hopefully you can't! 😬

This should be the last step to prevent badly formatted data from sneaking in again, so that our user data cleanup attempts don't all go to waste! 
#### Relevant tickets

References #6503.
#### Checklist

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
@@ -168,7 +168,7 @@ function _member_resource_create($request) {
   // Initialize array to pass to user_save().
   $edit = [
     'mail' => $email,
-    'name' => $email,
+    'name' => user_password(), // Overridden below.
     'status' => 1,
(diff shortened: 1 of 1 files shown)


angaither:
👍

Issue #6488, opened 2016-05-23
Error on Log In and Log Out on GL EN


namimody:
#### Current Behavior

Clicked Log In on GL EN homepage (in Canada via TorGuard), logged in with non-admin account, was brought to the screenshot below. I cleared cache and cookies, and got the same error. Hitting back in the browser brought me back to the homepage which looked fine, and I was logged in. If I tried to log out, I was shown the same screen as when I tried to log in.
#### Desired Behavior

Homepage should simply refresh if I log in from the homepage.
#### Steps to Replicate

Go to somewhere outside the US (not Brazil or MX) and try to log in via the homepage.
#### Why This Matters

Users should be able to log in and out.
#### Countries Affected (optional)

At least GL EN, need to test BR and MX. US EN seems fine.

Need to test if this happens from pitch page as well.
#### OS/Browser (optional)

Chrome.
#### Relevant Screenshots + Links

https://www.dosomething.org/user/logout
![image](https://cloud.githubusercontent.com/assets/12396076/15480654/b26f4f82-20f4-11e6-80dc-225aca54885e.png)

May be related to: https://github.com/DoSomething/phoenix/issues/5924

cc @mnelson338

weerd:
So one suspicion I had is that somewhere in the code, because it's a country we don't specifically account for, there's probably an error with it trying to do something with the Canada prefix and it can't because technically that should just be global.

Did some testing locally and this seems to be the case. On loading any page, even within the CMS, I see errors for:

```
Notice: Undefined index: en-ca in dosomething_global_get_prefix_for_language() (line 477 of /var/www/dev.dosomething.org/lib/modules/dosomething/dosomething_global/dosomething_global.module).
```

Not necessarily sure if this is the root cause of the above issue, but could be related. Needs to be fixed anyhoo!

weerd:
So it looks like Canada is a very specific case. 

I believe that for languages that we don't support on the site they get designated to `en-global`, so for example when we've tested spoofing loading the site from Japan which doesn't have `en` in its language code, we default that to `en-global` somewhere in the code and then all works fine with that. But for Canada which can have a language code of `en-ca` along with probably something like `fr-ca` it seems to throw our logic for a loop and doesn't default it to `en-global`. Instead it tries to find `en-ca` in our list and craps out.

[...]

PR #6507, opened 2016-05-26
Global fixes

weerd:
#### What's this PR do?

This PR updates the `dosomething_global_get_prefix_for_language()` function to allow it to fallback to no prefix returned for the URL (thus a global URL) if the passed `$language` isn't found in the list of supported languages. Prior to this, if somehow a language was passed that wasn't supported, we'd get lots of logged error notices.

Pull down the branch, use ModHeader chrome extension to pretend you're coming in from Canada, load the site and check the logs. If there are no errors regarding `Undefined index: en-ca in dosomething_global_get_prefix_for_language()` then all should be good.

This is only part of a fix that may be related to the larger issue in #6488 
#### Relevant tickets

Refs #6488

---

@angaither @joe


angaither:
👍

PR #6508, opened 2016-05-26
Fix issue where mobile was being validated if not passed.

DFurnes:
#### What's this PR do?

This fixes an issue introduced in #6502, where the contents of the `mobile` field would always be validated even if the field was not passed in the request. It is now only validated if the field exists.

You should be able to create a user without passing a valid `mobile`.

💃
#### Relevant tickets

🌵 
#### Checklist


angaither:
:ship: 📴

PR #6510, opened 2016-05-26
Forward users unless query-string is set.

DFurnes:
#### What's this PR do?

Users added through Phoenix's `POST /api/v1/users` endpoint are now automatically forwarded to Northstar, unless the `?forward=` query string is passed with a falsy value.

Is there a better way to get, cast, and set default values for query parameters?

The default behavior is now to forward users to Northstar. This way we don't need to make any changes to consumers of this API (e.g. Message Broker, SMS responder, maybe other things?) but can specifically opt-out of making an extra "circular" request when hitting this via Northstar.
#### Relevant tickets

References #6406.
#### Checklist


angaither:
Taps-Aff 🇬🇧 
↗️ ⭐

PR #6511, opened 2016-05-26
Fix for undefined index bug on pitch pages.

### LINKED ISSUE #6509: Undefined index notice on every pitch page.
#### Current Behavior

We are getting this error in the logs:

```
Notice: Undefined index: signup_cta in dosomething_campaign_preprocess_pitch_page() (line 423 of /var/www/dev.dosomething.org/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc).
```
#### Desired Behavior

We shouldn't be getting an error.
#### Steps to Replicate

Load a pitch page and check the logs. The error/notice will show up there.
#### Why This Matters

It's a small bug with a very quick and easy fix that has been around for a while. This should be fixed so that it doesn't pollute the list of more serious bugs that can be found in the logs that need to be addressed.

weerd:
#### What's this PR do?

This PR fixes a bug with the pitch pages with an `undefined index` with the `signup_cta`. 

We create the `signup_cta` item in the `$vars` array when preprocessing the pitch page, but instead of adding the item to the array, we were using the `.=` operator which otherwise requires the `signup_cta` to exist before it can be concatenated. But that item doesn't exist prior to the spot when we use the `.=` so it was always throwing an error that the item was undefined.

This PR also adds some spacing between lines of code in the `dosomething_campaign_preprocess_pitch_page()` function because FOR THE LOVE OF GOD AND ALL THAT IS HOLY that code is making me claustrophobic! Why are we afraid to add spacing to Drupal code @_@

Load a few pitch pages and see if `Notice: Undefined index: signup_cta in dosomething_campaign_preprocess_pitch_page()` still pops up in the logs.
#### Relevant tickets

Fixes #6509 

---

@angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -395,34 +395,46 @@ function dosomething_campaign_preprocess_shipment_form(&$vars) {
 function dosomething_campaign_preprocess_pitch_page(&$vars, &$wrapper) {
   // Use the pitch page template to theme.
   $vars['theme_hook_suggestions'][] = 'node__' . $vars['type'] . '__pitch';
+
   // Track that we're viewing a pitch page.
   $vars['campaign']->is_pitch_page = TRUE;
+
   // Check for a signup button copy override.
   $label = $vars['campaign']->variables['signup_form_submit_label'];
+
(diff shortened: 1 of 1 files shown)


DFurnes:
> This PR also adds some spacing between lines of code

Yessss, so much nicer. 😌

angaither:
🚢

PR #6512, opened 2016-05-26
Adds email masking on RB perma link

### LINKED ISSUE #5997: Reportback display: Mask usernames that are actually email addresses
If username matches an email regex, mask the string.

username-email

itsjoekent:
#### What's this PR do?

Hides a users email address on RB perma link page

First set the "non owners title backup" in the permalink settings 
`admin/config/dosomething/dosomething_campaign`

Then go to a RB perma link page as logged out user.
screen shot 2016-05-26 at 1 23 59 pm

After that change the first name to something that resembles an email
screen shot 2016-05-26 at 2 07 02 pm

run `drush cc all`

Visit that perma link page again and witness magic
screen shot 2016-05-26 at 2 06 49 pm

In theory I could have figured out a complex regex which made sure the string was a true email, and not just a string with @ in it. But why in the world would someone put @ in there first name?
#### Relevant tickets

Fixes #5997 
#### Checklist
- [ ] Set the backup non owner title in `admin/config/dosomething/dosomething_campaign` (I have it as Join DoSomething.org instead of Join [first name] and DoSomething.org)

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
@@ -247,6 +247,13 @@ function dosomething_campaign_admin_config_form($form, &$form_state) {
     '#required' => TRUE,
     '#default_value' => t(variable_get('dosomething_campaign_permalink_nonowners_page_title')),
   );
+  $form['permalink']['nonowners']['dosomething_campaign_permalink_nonowners_page_title_backup'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Permalink non-owners page title backup'),
+    '#description' => t('If the title contains sensitive user information this is used as a backup.'),
+    '#required' => TRUE,
+    '#default_value' => t(variable_get('dosomething_campaign_permalink_nonowners_page_title_backup')),
+  );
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
```diff
@@ -247,6 +247,13 @@ function dosomething_campaign_admin_config_form($form, &$form_state) {
     '#required' => TRUE,
     '#default_value' => t(variable_get('dosomething_campaign_permalink_nonowners_page_title')),
   );
+  $form['permalink']['nonowners']['dosomething_campaign_permalink_nonowners_page_title_backup'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Permalink non-owners page title backup'),
+    '#description' => t('If the title contains sensitive user information this is used as a backup.'),
+    '#required' => TRUE,
+    '#default_value' => t(variable_get('dosomething_campaign_permalink_nonowners_page_title_backup')),
```
angaither: can you set a default here, so we don't have to do this in every environment

itsjoekent: sure! is the current message I have "Join DoSomething.org" ok?

angaither:
👍

PR #6517, opened 2016-05-27
Remove leftover markup from Optimizely test.

DFurnes:
#### What's this PR do?

Removes markup for an old Optimizely test. The styles that had been hiding this markup were removed in #6494, but not the associated markup, so they started appearing on the page again.
#### How should this be reviewed?

👀
#### Any background context you want to provide?

🙈 
#### Relevant tickets

Fixes #6516.
#### Checklist
- [ ] Documentation added for new features/changed endpoints.

PR #6494, opened 2016-05-24
Spring cleaning

DFurnes:
#### What's this PR do?

This PR removes a lot of things that don't need to be there anymore: the hot/win module, snowflakes, international affiliate styles, and finished Optimizely tests. 🍃 
#### How should this be reviewed?

Nothing should visibly change since all of these features have been disabled for a long time.
#### Any background context you want to provide?

This reduces our stylesheet and scripts bundles by 64kb.
#### Checklist
- [ ] Documentation added for new features/changed endpoints.

PR #6519, opened 2016-05-27
Remove submit handler for removed function.

### LINKED ISSUE #6518: Fatal error when trying to save theme settings
#### Current Behavior

Trying to save theme settings results in a fatal error:

```
Fatal error: Call to undefined function paraneue_dosomething_theme_settings_handle_files() in /var/www/dev.dosomething.org/html/includes/form.inc on line 1519
```
#### Desired Behavior

It should not result in a fatal error.
#### Steps to Replicate

Go to the theme settings and try to save them.
#### Why This Matters

Fatal errors are bad.

DFurnes:
#### What's this PR do?

Fixes #6518 by removing the form submit handler for a function removed in #6494.

Try to save the theme settings, you should not get a fatal error.

💣 🙅 
#### Relevant tickets

Fixes #6517.
#### Checklist

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/theme-settings.php
+++ b/lib/themes/dosomething/paraneue_dosomething/theme-settings.php
@@ -42,11 +42,6 @@ function paraneue_dosomething_form_system_theme_settings_alter(&$form, &$form_st
   _paraneue_dosomething_theme_settings_footer($form, $form_state);
   _paraneue_dosomething_theme_settings_user($form, $form_state);
 
-  if (!isset($form['#submit'])) {
-    $form['#submit'] = [];
-  }
-  $form['#submit'][] = 'paraneue_dosomething_theme_settings_handle_files';
-
   // Work-around for this bug: https://drupal.org/node/1862892.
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/theme-settings.php
```diff
@@ -42,11 +42,6 @@ function paraneue_dosomething_form_system_theme_settings_alter(&$form, &$form_st
   _paraneue_dosomething_theme_settings_footer($form, $form_state);
   _paraneue_dosomething_theme_settings_user($form, $form_state);
 
-  if (!isset($form['#submit'])) {
-    $form['#submit'] = [];
```
angaither, line 46: does this part also wanna be gone?

DFurnes, line 46: We don't use that array anywhere else that I can tell (although who knows 😬).

angaither, line 46: I mean as long as it submits and saves values, I'm 🆒

DFurnes, line 46: Livin' dangerously! 💣

PR #6520, opened 2016-05-27
Switches the variable being referenced for non owner title

itsjoekent:
#### What's this PR do?

The variable being used to generate the non owner title on RB perma link

clear cache, go to a RB page that aint yours. then change the name to be an email and see if it hides it

used the array of field_first_name instead of the first name i extracted
#### Relevant tickets

Fixes #nope


angaither:
👍

PR #6522, opened 2016-05-27
Updates to address the annoying notices we always see due to #global.

### LINKED ISSUE #6521: Lots of locale_language_url_rewrite_url errors in dblog
#### Current Behavior
- The Drupal logs are constantly reporting numerous notices regarding:

```
Notice: Trying to get property of non-object in locale_language_url_rewrite_url() (line 433 of /var/www/dev.dosomething.org/html/includes/locale.inc).
```
#### Why This Matters

These notices overpopulate and mask more important warnings and notices that the system is reporting. It would be helpful to remove these, and may also highlight other issues in the code when the above notice is addressed.

weerd:
#### What's this PR do?

This PR fixes a long running issue with a dblog notice report that gets reported in the logs like 5 or more times during _any_ page load. With the help of @DFurnes and some XDebugging, we were able to localize the spots in our code that were causing the swarm of notices.

The main culprit is that when we were passing an array of url options to the `url()` function, if the `language` item was present we were passing it as a language code string, but the function required it to be passed as the Drupal language object.

```
Notice: Trying to get property of non-object in locale_language_url_rewrite_url() (line 433 of /var/www/dev.dosomething.org/html/includes/locale.inc).
```

Pull down the code, load a few pages (maybe spoof a different country other than US to test language stuff) and check the DB Logs to see if the above notice is reported!

While fixing this issue it seems like our code base in regards to languages is a bit confusing, with both confusing function names and confusing variable names.

Functions like `dosomething_global_get_language()` don't return the language object, but instead return the language code string. Variables like `$language` sometimes contain the Drupal language object (which seems to be the standard for what the variable should always hold) but at other times contains the language code string. This is very confusing and can complicate what the expected return value is from a function or variable as the variable is passed around. This is also what led to the issue this PR addresses.

I propose we standardize what we as a team consider these functions to return (and name them appropriately) along with a set of variable names that mean what they are expected to contain.

For example:

**Functions:**
Name change reflects that it returns a language code string

```
dosomething_global_get_language() -> dosomething_global_get_language_code() 
```

Name change reflects that it returns a language code string and also, the parameter is now more obvious that it needs to be the country code string

```
dosomething_global_convert_country_to_language($country) -> dosomething_global_convert_country_to_language_code($country_code)
```

**Variables:**
`$language` = stdClass object; the Drupal language object

`$language_code` = the [ISO639-1 language code](https://www.wikiwand.com/en/Language_localisation#/Language_tags_and_codes) (e.g. en, en-global, es-mx, pt-br)

`$country_code` = the [ISO-3166-1 2 letter code for a country](https://www.wikiwand.com/en/ISO_3166-1#/Officially_assigned_code_elements) (e.g. US, BR, MX)

`$prefix` = similar to the country code above but when used specifically in urls like http://dosomething.org/br/campaigns or whatever
#### Relevant tickets

Fixes #6521 

---

@angaither @DFurnes 

CC: @namimody @mikefantini

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -565,27 +565,38 @@ function dosomething_campaign_get_campaigns_doing($uid = NULL) {
  *   An array of variables.
  */
 function dosomething_campaign_get_campaign_block_vars($nid, $image_size = '740x480', $source = NULL) {
+  global $user;
+
   $node = node_load($nid);
   $clc = dosomething_helpers_get_current_language_content_code();
 
-  global $user;
   $language = dosomething_global_get_language($user, $node);
+
+  if (! is_object($language)) {
+    dosomething_global_get_language_by_language_code($language);
+  }
+
   // Build a global URL and remove the first slash returned
-  $path_alias = preg_replace('/\//', '', dosomething_global_url('node/' . $nid, array('language' => $language)), 1);
+  $path_alias = preg_replace('/\//', '', dosomething_global_url('node/' . $nid, ['language' => $language]), 1);
 
   if ($source) {
     $path_alias .= '?source=' . $source;
   }
+
   $image = NULL;
   $image_nid = NULL;
+
   if (!empty($node->field_image_campaign_cover)) {
     $image_nid = $node->field_image_campaign_cover[$clc][0]['target_id'];
     $image = dosomething_image_get_themed_image_url($image_nid, 'landscape', $image_size);
   }
+
   $cta = NULL;
+
   if (!empty($node->field_call_to_action)) {
     $cta = $node->field_call_to_action[$clc][0]['value'];
   }
+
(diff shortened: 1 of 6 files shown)


DFurnes:
👏! The naming suggestions you gave seem 100% more understandable to me.

It also seems like it'd make sense to always request or pass around those "language" objects, and then get the prefix/string code from the object when necessary (and avoid ever handling those string values directly).

weerd:
@DFurnes agreed. I think that would help resolve a lot of the issues with the complexity of #global

We might need to add a property or two to the Language object, but making that a class or something with some getters and setters with some nice defaults would make things easier and always know what to expect!

angaither:
👍

itsjoekent:
@weerd should 

`dosomething_global_convert_country_to_language_code($country_code)`

be `dosomething_global_convert_country_code_to_language_code($country_code)`

weerd:
@deadlybutter oh yeah, good call. I didn't actually implement that in this PR, just some suggestions for standardizing the craziness, but yeah, defs. Good call :)

PR #6523, opened 2016-05-27
Organ donation share link

### LINKED ISSUE #6483: share campaign
As a user I get a "unique" url that I can share with friends so other people can sign up from me
- create a custom URL that is a link to the campaign's pitch page with a UID attached
- eg dosomething.org/campagins/organ?source=[uid]

chloealee:
#### What's this PR do?
- Created custom share link that appends `source=[uid]` to share link.
- Adds share bar to organ donation module with custom share links. 
- Go to a campaign and enable organ donation in the custom settings form. 
- Sign up for the campaign from a user that has not signed up for this before. 
- Fill out the form until you get to the share page on the module. Click on share buttons and you should see `source=[uid]` as first param after the campaign URL. 
- Hover over or click on the share link and you should also see `source=[uid]` as the first param after campaign URL. 

Note: design/copy is not complete but these are just placeholders with correct info. for now. 
#### Relevant tickets

Fixes #6483 
#### Checklist

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -301,33 +301,6 @@ function dosomething_helpers_format_data($data) {
 }
 
 /**
- * Utility funciton to get base URL.
- *
- * @return string
- */
-function dosomething_helpers_get_base_url(){
-  return url(base_path(), array('absolute'=> TRUE, 'language' => 'en-global'));
-}
-
-/**
- * Utility function to get country prefix.
- *
- * @return string
- */
-function dosomething_helpers_get_country_prefix() {
-  return (dosomething_global_get_current_prefix()) ? strtoupper(dosomething_global_get_current_prefix()) : 'global';
-}
-
-/**
- * Utiliy function to get currnt URL path.
- *
- * @return string
- */
-function dosomething_helpers_get_current_url_path() {
-  return url(current_path(), array('absolute' => TRUE));
-}
-
-/**
  * Utility function to extract data translated in the specified language from an
  * array of available translations.
  *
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -301,6 +301,33 @@ function dosomething_helpers_format_data($data) {
 }
 
 /**
+ * Utility funciton to get base URL.
+ *
+ * @return string
+ */
+function dosomething_helpers_get_base_url(){
```
sbsmith86: I always love the idea of DRY'ing up code, but I sorta feel like these helpers are a little redundant. 

For example:
- `dosomething_helpers_get_base_url()` replaces the `url()` call which is already a helper that drupal provides. So in these cases we would really just be replacing one line of code with the other. It does make it a little bit more readible. One way of maybe making this more helpful is to take in the `$language_prefix` as a parameter so that it can be more dynamic? 
- A function like `dosomething_helpers_get_country_prefix()` already exists in `dosomething_global_get_current_country_code()` so I'm not sure we need this here as well. It might not have existed when we initially did share stuff so that it why it wasn't used. But totally understand the confusion there.
- `dosomething_helpers_get_current_url_path()` really just replaces one line of code with another. 

One thing we could think about is a helper that does all of [this](https://github.com/DoSomething/phoenix/blob/bb4fa06179e98a3b76cc14dd5a4791edfc8cf3e1/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc#L129-L154) in one function since you have to sorta follow all these steps when making the share bar...but a lot of the "steps" you need are specific to each configuration of the bar so maybe not.

chloealee: Yeah, I totally see your point here, @sbsmith86 ! That's what I was thinking too to extract it all into a helper but some variables, like you mentioned, are just a tad bit different. 

I'll change it back to before now!

sbsmith86:
💯 👍

PR #6514, opened 2016-05-26
Remove campaign group signup

angaither:
#### What's this PR do?

removes the ability to signup on campaign groups 
#### How should this be reviewed?
- 👀 
- test signups on campaign nodes
#### Any background context you want to provide?

this is a less radical approach than #6496, which for some reason was returning a WSOD for anon users on campaign pages. 
So I kept the signup config [fields on the admin section of campaign collections, for now](https://github.com/DoSomething/phoenix/pull/6496/commits/3e258360e776659d59a544d0b5055ea98fc046c3) 
#### Relevant tickets

Refs #5914

Issue #5914, opened 2015-12-07
Campaign Collection Redirects Needed


mikefantini:
We need redirects for the following campaign collections:
- [ ] https://www.dosomething.org/volunteer/globalgoals -> https://www.dosomething.org/us/volunteer/globalgoals
- [ ] https://www.dosomething.org/volunteer/suspended-what -> https://www.dosomething.org/us/volunteer/suspended-what
- [ ] https://www.dosomething.org/volunteer/tune-and-take-action -> https://www.dosomething.org/us/volunteer/tune-and-take-action
- [ ] https://www.dosomething.org/volunteer/manitou-service-weekend -> https://www.dosomething.org/us/volunteer/manitou-service-weekend
- [ ] https://www.dosomething.org/volunteer/boosomethingorg -> https://www.dosomething.org/us/volunteer/boosomethingorg
- [ ] https://www.dosomething.org/volunteer/bleep-out-bullying -> https://www.dosomething.org/us/volunteer/bleep-out-bullying
- [ ] https://www.dosomething.org/volunteer/stop-climate-change -> https://www.dosomething.org/us/volunteer/stop-climate-change

Perhaps we do this in Fastly as we did for Old School Impact and Seniors and Smartphones. Perhaps we come up with a different way to do it. See slack conversation for the Fastly config here: https://dosomething.slack.com/archives/phoenix/p1449530407002014

@mshmsh5000 @blisteringherb @sergii-tkachenko

mikefantini:
Also to be clear, the reason this is important is because currently users who try to go to that page get taken to 
![image](https://cloud.githubusercontent.com/assets/6330971/11643637/509cfeea-9d13-11e5-9fa6-e7e6b81d26ce.png)

mshmsh5000:
This is super lame to do in Fastly because of the interface: you have to set up a Content (request) rule with a condition, then a Response rule with a duplicate, identical condition.

Content rule:

![content-rule](https://cloud.githubusercontent.com/assets/149811/11722008/a8ac55f4-9f34-11e5-9fb8-9e24bef72919.png)

Content condition:

![content-condition](https://cloud.githubusercontent.com/assets/149811/11722046/dadcc66c-9f34-11e5-9117-a36c6df2b5c9.png)

Response rule:

![response](https://cloud.githubusercontent.com/assets/149811/11722071/ff74d032-9f34-11e5-9ab7-29e1886d1017.png)

Response condition:

[...]

PR #3832, opened 2015-01-28
Kills flagged_reportback Flag and related fields

aaronschachter:
Removes the `flagged_reportback` flag module, eliminating all traces of Flag module and the fields associated with the Flags we used.  Replaced by the Reportback Entity's `flagged` and `flagged_reason` properties and the new `dosomething_reportback_files_form`.

Closes #3689

PR #6524, opened 2016-05-27
Remove signup fields from campaign group

angaither:
#### What's this PR do?

removes the signup fields config from the campaign group node form.
- `drush updb -y`
- `drush fra -y`
- fields should be 💥 

This is the rest of the PR from #6514
#### Relevant tickets

refs #5914


lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.features.field_base.inc
```diff
@@ -221,48 +126,5 @@ function dosomething_campaign_group_field_default_field_bases() {
     'type' => 'text',
   );
 
-  // Exported field_base: 'field_signup_confirm_msg'.
-  $field_bases['field_signup_confirm_msg'] = array(
```
sergiitk, line 225: This field is still referenced here: https://github.com/DoSomething/phoenix/blob/e4961813d27a19bca90cf9015bd5fb609e732cf1/lib/modules/dosomething/dosomething_signup/dosomething_signup.module#L576.
I think this `if` statement should be removed.

lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.install
```diff
@@ -36,3 +36,20 @@ function dosomething_campaign_group_update_7003(&$sandbox) {
...
+  $fields = [
+    'field_automatic_parent_signup',
+    'field_display_signup_form',
+    'field_post_signup_copy',
+    'field_post_signup_title',
+    'field_signup_confirm_msg',
+    'field_transactional_email_copy',
+  ];
+  foreach($fields as $field) {
+    field_delete_field($field);
```
sergiitk, line 53: I think it's not safe. How `field_delete_field` would delete a field if it's base and instance is removed from code?

sergiitk, line 53: So if something (like `ds build`) invokes `drush fra -y` pripor this update, it could become a huge pain.

angaither, line 53: @sergii-tkachenko this is how we have removed fields in the past  (see #3832)

sergiitk, line 53: Yes, I'm testing it now and it looks like it works just fine. Sorry.

lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.features.field_base.inc
angaither, line 225: good catch, removed in cb765ab

lib/modules/dosomething/dosomething_campaign_group/dosomething_campaign_group.install
sergiitk, line 53: I think this is possible because field bases are still stored in the DB.

sergiitk:
👍 💥 ♻️

PR #6528, opened 2016-05-31
Only displays partners if translated

### LINKED ISSUE #6379: If partner term translation does not exist hide "In partnership with"
#### Current Behavior
- Not showing who the partner is for the Elephant Tusk Force campaign. 
#### Desired Behavior
- Partner name should be displayed. 
#### Steps to Replicate
- 1. While signed in, navigate to https://www.dosomething.org/node/5411
- 2. Scroll all the way down
#### Why This Matters
- Partners should be displayed if they are present. If not, the line should be hidden.

screen shot 2016-04-16 at 3 38 18 am

### LINKED ISSUE #6488: Error on Log In and Log Out on GL EN

itsjoekent:
#### What's this PR do?
- Does not display "Partners" in the info bar if the partner term has not been translated
- Minor fix to global get lang logic

First visit a campaign page that has a partner term set.
screen shot 2016-05-31 at 1 34 14 pm

Then translate this campaign, but NOT the partner term. Does it show in the info bar still?
screen shot 2016-05-31 at 1 34 32 pm

Also because I changed the global get lang function, make sure the site routing still seems ok.
#### Relevant tickets

Fixes #6379

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -514,12 +514,7 @@ function dosomething_global_get_language($account, stdClass $node = NULL, $respe
     // Respect the URL first.
     $request_path = explode('/', request_path());
     if (!empty($request_path[0])) {
-
-      // If the string is longer than a country code
-      if (strlen($request_path[0]) > 2) {
-        return DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE;
-      }
-
+      
       // Otherwise try & match it against configured languages
       $languages = language_list();
       foreach ($languages as $langcode => $lang_data) {
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.module
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.module
@@ -514,6 +514,13 @@ function dosomething_global_get_language($account, stdClass $node = NULL, $respe
     // Respect the URL first.
     $request_path = explode('/', request_path());
     if (!empty($request_path[0])) {
+
+      // If the string is longer than a country code
+      if (strlen($request_path[0]) > 2) {
+        return DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE;
+      }
+
+      // Otherwise try & match it against configured languages
       $languages = language_list();
       foreach ($languages as $langcode => $lang_data) {
         if ($lang_data->prefix == $request_path[0]) {
--- a/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
+++ b/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
@@ -207,7 +207,7 @@ function dosomething_taxonomy_get_partners_data($field_partners, $logo_style = '
  *   Multi-dimensional array of partners/sponsors, or NULL if empty.
  *
  */
-function dosomething_taxonomy_get_partners_vars($field_partners = array(), $language_code = NULL) {
+function dosomething_taxonomy_get_partners_vars($field_partners = array(), $language_code = 'en-global') {
 
   if (empty($field_partners)) {
     return;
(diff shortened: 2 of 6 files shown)


lib/modules/dosomething/dosomething_global/dosomething_global.module
```diff
@@ -514,6 +514,13 @@ function dosomething_global_get_language($account, stdClass $node = NULL, $respe
     // Respect the URL first.
     $request_path = explode('/', request_path());
     if (!empty($request_path[0])) {
+
+      // If the string is longer than a country code
```
angaither: what is this case for?

lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
```diff
@@ -199,12 +199,15 @@ function dosomething_taxonomy_get_partners_data($field_partners, $logo_style = '
...
+ * @param string $language_code
+ *   The optional language code which indicates to only return partners
+ *   translated in that lang.
  *
  * @return mixed
  *   Multi-dimensional array of partners/sponsors, or NULL if empty.
  *
  */
-function dosomething_taxonomy_get_partners_vars($field_partners = array()) {
+function dosomething_taxonomy_get_partners_vars($field_partners = array(), $language_code = NULL) {
```
angaither: we probably don't want this to be `NULL` default, right? if we call this w/o a `language_code` it will do what we currently see on the site now

lib/modules/dosomething/dosomething_global/dosomething_global.module
itsjoekent: `campaigns` -- but if the default language was Spanish/Mexico, this would be `voluntario`

basically, 

/campaigns/something = global
/us/campaigns/something = not global

in the first case the [0] is the campaigns, and this new line of code catches that

lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
itsjoekent: true, apart of me was wondering if we'd have a scenario where we didn't want to filter by language. but if you think its safe to remove the default then 👍

angaither: shouldn't the default be `en-global`?

itsjoekent: yeah that should work fine. it feels weird because most of our partners will be in "en" but #global....

lib/modules/dosomething/dosomething_global/dosomething_global.module
angaither: we should be able to follow the way we determine language in [other places](https://github.com/DoSomething/phoenix/blob/bd9b5fc7a3a585481cbff4a1435a7804bdd8f14e/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc#L93) instead of updating this

itsjoekent:
@angaither seems to work with the way you suggested, removed the global change and also updated the default param to be en-global

angaither:
👍

PR #6531, opened 2016-06-01
Issue6439 campaign signup transactional flag

### LINKED ISSUE #6439: Optional transactional for /api/v1/campaigns/[nid]/signup
#### User Story

When the Campaigns Signup (https://github.com/DoSomething/phoenix/wiki/API#campaign-signup) endpoint is called a transactional message is triggered. This message needs to be optionally suppressed.
#### Additional Information
- calls to `https://www.dosomething.org/api/v1/campaigns/[nid]/signup` result in the campaign signup transactional being sent:

image
#### Why This Matters
- In the case of user imports that include a campaign signup it's desired that a campaign signup transactional is not sent in favour or other messaging in a welcome email.
- not possible, all calls to this endpoint result in a transactional.
- as a possible workaround, the `source` is defined in the call to the endpoint which could be used to disable the sending the transactional when `source = niche`. The `source` value would have to be looked up in the `signup` table and add the `source` value to the request message payload.
#### Desired Behavior
- an optional parameter `transactional=1 | 0` in addition to `uid` and `source` results in the campaign signup transactional not / being sent.
#### Additional Info
- Current transactional message payload:

```
(
    [activity] => campaign_signup
    [email] => jjensen+nicheTest-101@dosomething.org
    [uid] => 3882606
    [merge_vars] => Array
        (
            [MEMBER_COUNT] => 5.3 million
            [FNAME] => James
            [CAMPAIGN_TITLE] => Birthday Mail
            [CAMPAIGN_LINK] => https://www.dosomething.org/campaigns/birthday-mail?source=node/2461
            [CALL_TO_ACTION] => Make a birthday card for a child experiencing homelessness.
            [STEP_ONE] => Make Your Card!
            [STEP_TWO] => Snap a Pic
            [STEP_THREE] => Snail Mail It
        )
    [user_country] => US
    [user_language] => en
    [campaign_country] => global
    [subscribed] => 1
    [event_id] => 2461
    [email_tags] => Array
        (
            [0] => 2461
            [1] => drupal_campaign_signup
        )

    [activity_timestamp] => 1463156998
    [application_id] => US
)
```

deezone:
#### What's this PR do?

Adds support for transactional parameter in call to **Campaign Signup** endpoint: https://github.com/DoSomething/phoenix/wiki/API#campaign-signup that will result in transactional messaging being able to be prevented on campaign signups via Phoenix API calls.

```
curl http://dev.dosomething.org:8888/api/v1/campaigns/23/signup -X POST 
--header "Content-type: application/json" 
--header "Accept: application/json" 
--header "X-CSRF-Token: G136HF5yB5ZZawvrsOfU4gw0poUOaQygPrsJlaFakMU" 
--header "Cookie:SESSd57f2aef87e6d4352ce5db4659184fa7=mKI5_yfoXYBz3r4o95utui4fwBV_lUO1JNN1nEVsPRg" 
-d '{
  "uid": 123uid456,
  "source": "niche",
  "tranasactional": 0
}'
```

Should result in a campaign signup `dosomething_mbp` message generated with a `transactional` value of `FALSE / 0`. This message value will be consumed by `mbc-transactional-email` and `mbc-registration-mobile` to suppress sending transactional messaging.

Niche user imports include automatic signup to certain campaigns. The messaging sent to Niche users includes details about being signed up to a campaign. The additional generic campaign signup messages needs to be disabled to prevent redundant messaging.
#### Relevant tickets

Fixes #6439
#### Checklist
- [ ] cURL call to `/api/v1/campaigns/23/signup` results in `transactional` value in mbp message.

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.install
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.install
@@ -715,3 +715,19 @@ function dosomething_signup_update_7024() {
     db_add_field($tbl_name, $field_name, $schema[$tbl_name]['fields'][$field_name]);
   }
 }
+
+/**
+ * Adds transactional column to the dosomething_signup table.
+ */
+function dosomething_signup_update_7025() {
+  $tbl_name = 'dosomething_signup';
+  // Load schema to get table definition.
+  $schema = dosomething_signup_schema();
+  // New field to add.
+  $field_name = 'transactional';
+  // If the field doesn't exist already:
+  if (!db_field_exists($tbl_name, $field_name)) {
+    // Add it per the schema field definition.
+    db_add_field($tbl_name, $field_name, $schema[$tbl_name]['fields'][$field_name]);
+  }
+}
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.install
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.install
@@ -69,6 +69,11 @@ function dosomething_signup_schema() {
         'type' => 'int',
         'not null' => FALSE,
       ),
+      transactional' => array(
+        'description' => 'Boolean indicating if transactional messaging should be sent',
+        'type' => 'int',
+        'not null' => TRUE,
+      ),
     ),
     'primary key' => array('sid'),
     'unique keys' => array(
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -537,8 +537,11 @@ function dosomething_signup_third_party_subscribe($account, $node) {
     $var_name = 'dosomething_signup_mobilecommons_opt_in_path_general_campaign';
     $opt_in = variable_get($var_name);
   }
+
+  $var_name  = 'transactional';
+  $transactional = dosomething_helpers_get_variable('node', $node->nid, $var_name);
   // Send message broker request.
-  dosomething_signup_mbp_request($account, $node, $opt_in);
+  dosomething_signup_mbp_request($account, $node, $opt_in, $transactional);
 }
 
 /**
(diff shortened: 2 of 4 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.install
```diff
@@ -710,3 +715,19 @@ function dosomething_signup_update_7024() {
     db_add_field($tbl_name, $field_name, $schema[$tbl_name]['fields'][$field_name]);
   }
 }
+
+/**
+ * Adds transactional column to the dosomething_signup table.
+ */
+function dosomething_signup_update_7025() {
```
angaither, line 722: This will create a `transactional` field on every signup, I don't think we need to store if the signup was transactional on each individual signup record?

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -534,8 +537,11 @@ function dosomething_signup_third_party_subscribe($account, $node) {
     $var_name = 'dosomething_signup_mobilecommons_opt_in_path_general_campaign';
     $opt_in = variable_get($var_name);
   }
+
+  $var_name  = 'transactional';
+  $transactional = dosomething_helpers_get_variable('node', $node->nid, $var_name);
```
angaither, line 542: Where are you getting this helper variable? It's not set anywhere

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -229,11 +229,13 @@ function dosomething_signup_get_query_source() {
...
  *   (optional) The timestamp of the signup.
  *   If not provided, uses @dries time.
+ * @param bool $transactional
+ *   (optional) Flag to trigger sending transactional messages related to signup.
  *
  * @return mixed
  *   The sid of the newly inserted signup, or FALSE if error.
  */
-function dosomething_signup_create($nid, $uid = NULL, $source = NULL, $timestamp = NULL) {
+function dosomething_signup_create($nid, $uid = NULL, $source = NULL, $timestamp = NULL, $transactional = TRUE) {
```
angaither, line 238: I don't think we want the default here to be true, that means each time this is called throughout the codebase, it would be logged as a transactional signup

angaither, line 238: just saw this comment
`+  // transactional = FALSE will result in neither email or SMS transactional messages from being sent.`
so maybe that should be in the block above?

deezone:
Closing after @angaither pointed out the flaws in how I was implementing my solution.

Issue #6482, opened 2016-05-23
post request to api


angaither:
- send first request to the organize api with the zip code

angaither:
closing for #6547

PR #6533, opened 2016-06-01
Refactor takeover modal

sbsmith86:
#### What's this PR do?

Refactor some of what I'm calling the "takeover" modal into a class that we can use for various things. In this case, I also set up an `OrganDonation` class that extends the `Takeover` class and lets use use the modal functionality for the custom organ donation flow. 
#### Relevant tickets

Address #6482

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/app.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/app.js
@@ -80,9 +80,9 @@ $(document).ready(function() {
   // Add fallback for HTML5 video on mobile tiles
   Tile.init();
 
-  if ($('.takeover-container').length) {
-    let donation = new OrganDonation($('.takeover-container'));
-  }
+  // Initialize organ donation modal if it is on the page.
+  OrganDonation.init();
+
   /**
    * We use the Filament Group's fixed-sticky library to polyfill the
    * CSS `position: sticky` property.
--- a/lib/themes/dosomething/paraneue_dosomething/js/app.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/app.js
@@ -21,7 +21,6 @@ import Donate from 'donate/Donate';
 import Finder from 'finder/Finder';
 import Reportback from 'reportback/Reportback';
 import SchoolFinder from 'campaign/SchoolFinder';
-// import Takeover from 'takeover/Takeover';
 import OrganDonation from 'campaign/OrganDonation';
 
 // Patterns
@@ -77,9 +76,6 @@ $(document).ready(function() {
 
   // Initialize "show more" button on permalink pages.
   ShowMore.init();
-
-  // Intialize takeover modal.
-  // Takeover.init();
 
   // Add fallback for HTML5 video on mobile tiles
   Tile.init();
--- a/lib/themes/dosomething/paraneue_dosomething/js/app.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/app.js
@@ -21,7 +21,8 @@ import Donate from 'donate/Donate';
 import Finder from 'finder/Finder';
 import Reportback from 'reportback/Reportback';
 import SchoolFinder from 'campaign/SchoolFinder';
-import Takeover from 'takeover/Takeover';
+// import Takeover from 'takeover/Takeover';
+import OrganDonation from 'campaign/OrganDonation';
 
 // Patterns
 import Gallery from 'patterns/Gallery';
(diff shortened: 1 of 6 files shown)


lib/themes/dosomething/paraneue_dosomething/js/app.js
```diff
@@ -78,11 +79,12 @@ $(document).ready(function() {
   ShowMore.init();
 
   // Intialize takeover modal.
-  Takeover.init();
+  // Takeover.init();
```
DFurnes: 👾

lib/themes/dosomething/paraneue_dosomething/js/app.js
```diff
@@ -78,11 +79,12 @@ $(document).ready(function() {
...
 
   // Add fallback for HTML5 video on mobile tiles
   Tile.init();
 
+  let donation = new OrganDonation($('.takeover-container'));
```
DFurnes: Does mean we'll try to instantiate the `OrganDonation` takeover on every page now? If so, it might be worth exposing an `init` function from this module like some of the others do that checks if we're on the right page first.

lib/themes/dosomething/paraneue_dosomething/js/campaign/OrganDonation.js
```diff
@@ -0,0 +1,84 @@
+const $ = require('jquery');
+import setting from '../utilities/Setting';
+import Takeover from 'takeover/Takeover';
+
+class OrganDonation extends Takeover {
+  constructor($takeoverContainer) {
+
+    super($takeoverContainer);
+
+    this.zipCodeButton = $('.submit-postal-code');
```
DFurnes: Should these be scoped to the given container, like `$takeoverContainer.find('.submit-postal-code')`?

lib/themes/dosomething/paraneue_dosomething/js/campaign/OrganDonation.js
```diff
@@ -0,0 +1,84 @@
...
+    this.backButton = $('.takeover__screen .back-button');
+
+    this.set();
+  }
+
+  /*
+   * Handles the zip code request screen.
+   */
+  submitZipCode() {
+    this.zipCodeButton.off('click');
```
DFurnes, line 21: Why do each of these methods remove the associated event handler?

lib/themes/dosomething/paraneue_dosomething/js/campaign/OrganDonation.js
```diff
@@ -0,0 +1,84 @@
...
+   * Handles the registration request screen.
+   */
+  submitRegistration() {
+    this.registrationButton.off('click');
+    const sid = setting('dosomethingSignup.sid');
+
+    this.handleSubmit(this.registrationButton, {
+      'button' : this.registrationButton,
+      'method' : 'GET',
+      'url'    : this.baseUrl + '/organ-donation/registration',
```
DFurnes: Totally stylistic, but you could use a template string to make this prettier!

lib/themes/dosomething/paraneue_dosomething/js/takeover/Takeover.js
```diff
@@ -1,80 +1,91 @@
...
+   *
+   * @param  {string}   direction
+   */
+  changeScreens(direction) {
+    // Check if an allowed direction was passed in.
+    if ($.inArray(direction, ['next', 'previous']) === -1) {
+      return;
+    }
+
+    const $current = $(`.takeover__screen.${this.activeClass}`);
```
DFurnes, line 72: It could be nice to store the container (like as `this.$el` or something like that) when instantiating this class and make DOM lookups relative to that. Probably a little performance boost, and reduces the risk of there somehow being a conflicting classname elsewhere.

lib/themes/dosomething/paraneue_dosomething/js/campaign/OrganDonation.js
sbsmith86, line 21: @DFurnes cause I want to reset all the click handlers when a screen is active. I'm using CSS transitions so nothing ever leaves the DOM so, if you go back and forth between screens, I wanted to reset the state of the buttons.

lib/themes/dosomething/paraneue_dosomething/js/app.js
sbsmith86: @DFurnes not sure i understand how the init function would work with classes...it seems we should check if the element we are looking for is available before instantiating the class like...

```
if ($('.takeover-container')) {
let donation = new OrganDonation($('.takeover-container'));
}
```

Or are you saying the init function gets called in the constructor of these classes, so we still instantiate the class on every page, but none of the code runs.

DFurnes: If the `OrganDonation` class isn't used outside the module, it might make sense to only export an `init` function from that module rather than exporting the class:

``` js
class OrganDonation {
  // ...
}

export default {
  init($el = $('.takeover-container')) {
    new OrganDonation($el);
  }
}
```

sbsmith86: ahhh i see i see!

sbsmith86:
@DFurnes Updates made! Can you take another look.

DFurnes:
Looks good to me! 🚢

PR #6534, opened 2016-06-02
Issue6439 api campaign signup transactionals

### LINKED ISSUE #6439: Optional transactional for /api/v1/campaigns/[nid]/signup

deezone:
#### What's this PR do?

Adds support for `transactionals` and `source` parameters in call to **Campaign Signup** endpoint: https://github.com/DoSomething/phoenix/wiki/API#campaign-signup that will result in transactional messaging being able to be prevented on campaign signups via Phoenix API calls as well as respond to `source` values.

```
curl http://dev.dosomething.org:8888/api/v1/campaigns/23/signup -X POST 
--header "Content-type: application/json" 
--header "Accept: application/json" 
--header "X-CSRF-Token: G136HF5yB5ZZawvrsOfU4gw0poUOaQygPrsJlaFakMU" 
--header "Cookie:SESSd57f2aef87e6d4352ce5db4659184fa7=mKI5_yfoXYBz3r4o95utui4fwBV_lUO1JNN1nEVsPRg" 
-d '{
  "uid": 123456,
  "source": "niche",
  "tranasactionals": 0
}'
```

Should result in a campaign signup `dosomething_mbp` message generated with a `transactional` value of `FALSE / 0`. This message value will be consumed by `mbc-transactional-email` and `mbc-registration-mobile` to suppress sending transactional messaging.

Niche user imports include automatic signup to certain campaigns. The messaging sent to Niche users includes details about being signed up to a campaign. The additional generic campaign signup messages needs to be disabled to prevent redundant messaging.
#### Relevant tickets

Fixes #6439
#### Checklist
- [x] cURL call to `/api/v1/campaigns/23/signup` results in `transactionals` and `source` value in mbp message.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -264,6 +264,7 @@ function _campaign_resource_index($ids, $staff_pick, $term_ids, $count, $random,
  *   - uid: The user uid (int).  Optional, uses global $user if not set.
  *   - source (string).
  *   - transactionals (bool). Optional flag to disable sending transactional messages.
+ *       Defaults to sending messages.
  */
 function _campaign_resource_signup($nid, $values) {
   global $user;
@@ -274,7 +275,7 @@ function _campaign_resource_signup($nid, $values) {
     $values['source'] = NULL;
   }
   if (!isset($values['transactionals'])) {
-    $values['transactionals'] = NULL;
+    $values['transactionals'] = TRUE;
   }
   if (DOSOMETHING_SIGNUP_LOG_SIGNUPS) {
     watchdog('dosomething_api', '_campaign_resource_signup values:' . json_encode($values));
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -263,6 +263,7 @@ function _campaign_resource_index($ids, $staff_pick, $term_ids, $count, $random,
  *   The signup data to post. Expected keys:
  *   - uid: The user uid (int).  Optional, uses global $user if not set.
  *   - source (string).
+ *   - transactionals (bool). Optional flag to disable sending transactional messages.
  */
 function _campaign_resource_signup($nid, $values) {
   global $user;
@@ -272,13 +273,14 @@ function _campaign_resource_signup($nid, $values) {
   if (!isset($values['source'])) {
     $values['source'] = NULL;
   }
+  if (!isset($values['transactionals'])) {
+    $values['transactionals'] = NULL;
+  }
   if (DOSOMETHING_SIGNUP_LOG_SIGNUPS) {
     watchdog('dosomething_api', '_campaign_resource_signup values:' . json_encode($values));
   }
-  // @todo: Pass parameter into signup_create whether or not to send SMS.
-  // Since SMS campaign signups would hit this endpoint, would not want
-  // to send an additional "You've signed up text".
-  return dosomething_signup_create($nid, $values['uid'], $values['source']);
+  // transactionals = FALSE will result in neither email or SMS transactional messages being sent.
+  return dosomething_signup_create($nid, $values['uid'], $values['source'], NULL, $values['transactionals']);
 }
 
 
--- a/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
+++ b/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
@@ -105,26 +105,25 @@ function dosomething_mbp_get_transactional_payload($origin, $params = NULL) {
     ),
   );
 
-  if (isset($params['transactionals'])) {
-    $payload['transactionals'] = $params['transactionals'];
+  // Optional values
+  $optional_params = [
+    'transactionals',
+    'source',
+    'user_language',
+    'campaign_language',
+    'campaign_country',
+  ];
+  foreach ($optional_params as $op) {
+    if (isset($params[$op])) {
+      $payload[$op] = $params[$op];
+    }
   }
-  if (isset($params['source'])) {
-    $payload['source'] = $params['source'];
-  }
+
   if (isset($params['user_country'])) {
     $payload['user_country'] = $params['user_country'];
   }
   else {
     $payload['user_country'] = dosomething_settings_get_geo_country_code();
-  }
-  if (isset($params['user_language'])) {
-    $payload['user_language'] = $params['user_language'];
-  }
-  if (isset($params['campaign_language'])) {
-    $payload['campaign_language'] = $params['campaign_language'];
-  }
-  if (isset($params['campaign_country'])) {
-    $payload['campaign_country'] = $params['campaign_country'];
   }
 
   // If email template wasn't set use request details to define template name.
--- a/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
+++ b/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
@@ -104,6 +104,10 @@ function dosomething_mbp_get_transactional_payload($origin, $params = NULL) {
       'MEMBER_COUNT' => dosomething_user_get_member_count(TRUE),
     ),
   );
+
+  if (isset($params['transactionals'])) {
+    $payload['transactionals'] = $params['transactionals'];
+  }
   if (isset($params['user_country'])) {
     $payload['user_country'] = $params['user_country'];
   }
--- a/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
+++ b/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
@@ -108,6 +108,9 @@ function dosomething_mbp_get_transactional_payload($origin, $params = NULL) {
   if (isset($params['transactionals'])) {
     $payload['transactionals'] = $params['transactionals'];
   }
+  if (isset($params['source'])) {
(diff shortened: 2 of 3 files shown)


deezone:
@angaither Another pass based on our discussion today.

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -272,13 +273,14 @@ function _campaign_resource_signup($nid, $values) {
   if (!isset($values['source'])) {
     $values['source'] = NULL;
   }
+  if (!isset($values['transactionals'])) {
+    $values['transactionals'] = NULL;
```
sergiitk: Why `NULL`? Would `FALSE` be more explicit?

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -619,11 +638,15 @@ function dosomething_signup_mbp_request($account, $node, $opt_in) {
  *   Details about the node that the signup was made on.
  * @param int $opt_in
  *   The opt-in path to use for opt-in.
+ * @param bool $transactionals
+ *   (optional) Control if transactional messaging should be sent.
+ * @param bool $transactionals
```
sergiitk: Did you mean to document `$source` here?

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.module
```diff
@@ -104,6 +104,13 @@ function dosomething_mbp_get_transactional_payload($origin, $params = NULL) {
       'MEMBER_COUNT' => dosomething_user_get_member_count(TRUE),
     ),
   );
+
+  if (isset($params['transactionals'])) {
```
angaither: just a refactor note, this whole block is pretty repetitive, might be easier to have an array of values and add them to the `$payload` if they are set

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -508,6 +512,8 @@ function dosomething_signup_entity_insert($entity, $type) {
 
   $account = user_load($entity->uid);
   $node = node_load($entity->nid);
+  $transactionals = $entity->transactionals;
```
angaither: does this work? I don't think you're ever attaching it to the entity?

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -518,13 +524,22 @@ function dosomething_signup_entity_insert($entity, $type) {
   }
 
   // Send relevant third-party subscribe requests.
-  dosomething_signup_third_party_subscribe($account, $node);
+  dosomething_signup_third_party_subscribe($account, $node, transactionals, $source);
```
angaither: missing `$` in front of `transactionals`

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
deezone: Line 263 of `dosomething_signups.module` adds `transactionals` value to `$values` submitted to `create_entity()`.

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -229,11 +229,14 @@ function dosomething_signup_get_query_source() {
...
  *   If not provided, uses @dries time.
+ * @param bool $transactionals
+ *   (optional) A flag to disable sending transactional messaging.
+ *   The default is to send messaging.
  *
  * @return mixed
  *   The sid of the newly inserted signup, or FALSE if error.
  */
-function dosomething_signup_create($nid, $uid = NULL, $source = NULL, $timestamp = NULL) {
+function dosomething_signup_create($nid, $uid = NULL, $source = NULL, $timestamp = NULL, $transactionals = NULL) {
```
sergiitk: Mybe set `$transactionals = FALSE` by default?

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -518,13 +524,22 @@ function dosomething_signup_entity_insert($entity, $type) {
...
+ *    User details who signed up for a campaign.
+ * @param object $node
+ *   Details about the campaign the user signed up to.
+ * @parm bool $trasactionals
+ *    A flag to prevent sending transactionals when FALSE. Defaults to TRUE when NULL.
+ * @parm string $source
+ *    A value defining where the request to sign a user up to a campaign originated.
  */
-function dosomething_signup_third_party_subscribe($account, $node) {
+function dosomething_signup_third_party_subscribe($account, $node, $transactionals = NULL, $source = NULL) {
```
sergiitk: Let's keep `$source` argument before `$transactionals`, as it is originally in `dosomething_signup_create()`;

sergiitk:
I really, really, don't like all these looong function declarations with most arguments default to NULL.
It's getting very messy and inconsistent, compare:

`function dosomething_signup_create($nid, $uid = NULL, $source = NULL, $timestamp = NULL, $transactionals = NULL) {`

`function dosomething_signup_third_party_subscribe($account, $node, $transactionals = NULL, $source = NULL) {`

Somewhere `$transactionals` is the third arg, somewhere — the fourth, etc, etc. It'll soon look like win32 api: `doSomething(1, nil, MY_CONSTANT, nil, DONT_TOUCH_THIS, true, 'actual data')`.

I recommend moving all optional arguments to `options` arg. Example:

``` php
function dosomething_signup_create($nid, $options = []) {
```

``` php
function dosomething_signup_third_party_subscribe($account, $node, $options = []) {
```

And then just check `$options` for necessary keys. It'd be also much easier to pass options around.

What do you think, @angaither?

deezone:
@sergii-tkachenko I support your suggestion to use `$options = []` as a param. I've adjusted `dosomething_signup_third_party_subscribe($account, $node, $options)`.

But `dosomething_signup_create()` is called in 9 places throughout the code base. I suggest the params refactoring be done in a separate issue.

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -517,14 +519,30 @@ function dosomething_signup_entity_insert($entity, $type) {
...
+ *   Details about the campaign the user signed up to.
+ * @parm array $options
+ *   Possible options used for the subscribe request.
+ *   - source
+ *     A value defining where the request to sign a user up to a campaign originated.
+ *   - trasactionals
+ *     A flag to prevent sending transactionals when FALSE. Defaults to TRUE when NULL.
  */
-function dosomething_signup_third_party_subscribe($account, $node) {
+function dosomething_signup_third_party_subscribe($account, $node, $options) {
```
sergiitk: Let's set $options to an empty array be default. `$options = []`.

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -584,8 +602,13 @@ function dosomething_signup_user_signup($nid, $account = NULL, $source = NULL) {
...
  * @param int $opt_in
  *   The opt-in path to use for opt-in.
+ * @param array $options
+ *   - transactionals
+ *   (optional) A flag to disable sending transactional messages.
+ *   - source
+ *   (optional) A idendify name of the original source that requested the campaign signup.
  */
-function dosomething_signup_mbp_request($account, $node, $opt_in) {
+function dosomething_signup_mbp_request($account, $node, $opt_in, $options) {
```
sergiitk: Let's declare $options as an empty array everywhere, like you did in `dosomething_signup_third_party_subscribe()`.

deezone: Good point - done.

sergiitk:
👍

PR #6538, opened 2016-06-02
Adding in command to restart php7 after deploy

blisteringherb:
#### What's this PR do?

I've added a command to restart PHP7 that's analogous to the command to restart PHP5

On a box running PHP7, You can check the PID for PHP7 fpm first by running `ps aux | grep php` run `bin/ds restart_php7` and then make sure the PID has changed. There won't be any output. Or run this deploy on Thor.

This is in preparation for upgrading the Drupal servers to PHP7. After this is merged, we should plan to halt any deploys until the servers have been updated to PHP7. It won't block deploys, but if the servers haven't been upgraded we'll need to manually restart php-fpm on all of the servers.
- Leaving in php5 restart command in case we need to roll back

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -551,6 +551,24 @@ function restart_php5_fpm {
   fi
 }
 
+#=== FUNCTION ==================================================================
+# NAME: restart_php7_fpm
+# DESCRIPTION: Restarts php7.0-fpm on servers running php7.
+#===============================================================================
+function restart_php7_fpm {
+  PHP_FPM_STATUS="$(sudo service php7.0-fpm status)"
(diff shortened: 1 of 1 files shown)


angaither:
What's the time frame for updating the servers to PHP7? Should we put the php version in an `env` var until they are all updated so we don't need to manually restart `php-fpm`?

>  After this is merged, we should plan to halt any deploys until the servers have been updated to PHP7.

blisteringherb:
I'd like to complete the upgrade today.

angaither:
vunderful! 🚢

PR #6540, opened 2016-06-02
Reportback source

### LINKED ISSUE #6385: Posting Reportback to API does not store source
#### Current Behavior
- Neither the `dosomething_reportback` or `dosomething_reportback_file` tables contain a source column, which means we're not tracking the platform a Reportback is submitted for -- even though we're sending a `source` value when posting from the [mobile app](https://github.com/DoSomething/LetsDoThis-iOS/blob/6448d07491cb7ca42c597f9af9b85b3353518953/Lets%20Do%20This/Networking/DSOAPI.m#L210). 

The [`campaigns/:id/reportback` endpoint](https://github.com/DoSomething/phoenix/blob/18f5e064e135cdc70d5b1bf0df60091c37fb8a50/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc#L131) calls [`dosomething_reportback_save`](https://github.com/DoSomething/phoenix/blob/18f5e064e135cdc70d5b1bf0df60091c37fb8a50/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc#L352), which calls [`dosomething_reportback_set_properties`](https://github.com/DoSomething/phoenix/blob/b2706b012d415a2d94796bb6df776da538dcd2e4/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module#L731), which [doesn't check for a `source`](https://github.com/DoSomething/phoenix/blob/b2706b012d415a2d94796bb6df776da538dcd2e4/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module#L767)  -- it wouldn't have a column to set it to.. unless we hijack the `dosomething_reportback_files.review_source` column. That feels quite 😷  but a little less than work than creating a new column -- plus would it go on the Reportback or the Reportback Files table?
#### Desired Behavior
- Store the posted Reportback `source`, either in DS Reportback or DS Reportback Files
#### Why This Matters

Don't have a clean (or any?) way to tell if a Reportback was submitted via mobile app, web, or SMS. I think the hack we used (and likely still use?) is checking the IP in `remote_addr` for a Mobile Commons IP (cc @jonuy )

katiecrane:
#### What's this PR do?
- added a column for `source` on `reportback_files`
- set the source when posting a reportback (if it is included)
  - the mobile app is already [sending this in](https://github.com/DoSomething/LetsDoThis-iOS/blob/6448d07491cb7ca42c597f9af9b85b3353518953/Lets%20Do%20This/Networking/DSOAPI.m#L210) 
- ~~return the source when getting a reportback~~
- ~~currently `source` is public (along with `caption` and `status`)~~

[Post a reportback](https://github.com/DoSomething/phoenix/wiki/API#campaign-reportback) and include a source as one of the parameters.
[Retrieve that reportback](https://github.com/DoSomething/phoenix/wiki/API#retrieve-a-specific-reportback) and make sure you DO NOT see the source.
Try the same thing with no sources.
#### Relevant tickets

Fixes #6385
#### Checklist

### DIFF
--- a/documentation/endpoints/campaigns.md
+++ b/documentation/endpoints/campaigns.md
@@ -240,3 +240,53 @@ curl -X GET \
 }
 ```
 
+## Campaign reportback
+
+Creates or updates a User Reportback for the given Campaign nid.
+
+**POST** `https://www.dosomething.org/api/v1/campaigns/[nid]/reportback`
+
+Must be authenticated to post.
+
+  - **uid**: (int).
+    The user uid reporting back.  If not included, the uid of the current user will be used.
+  - **nid**: (int) required.
+    The node nid the user is reporting back to.
+  - **quantity**: (int) required.
+    The number of reportback nouns verbed.
+  - **why_participated**: (string) required.
+    The reason why the user participated.
+  - **file_url**: (string) required if `file` is not provided.
+    An image URL for the reportback.
+  - **file**: (string) required if `file_url` is not provided.
+    Base64 encoded file string to save.
+  - **filename**: (string) required if `file` is provided.
+    Necessary for Drupal File API save (see `dosomething_reportback_get_file_dest`).
+  - **caption**: (string).
+    Corresponding caption for the reportback image.
+  - **source**: (string).
+    Where the reportback file was submitted from.
+
+Response: The reportback rbid if success, `FALSE` if not.
+
+Example request:
+````
+curl https://www.dosomething.org/api/v1/campaigns/23/reportback -X POST 
+--header "Content-type: application/json" 
+--header "Accept: application/json" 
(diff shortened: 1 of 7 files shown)


angaither:
@katiecrane do you mind taking this wonderful opportunity to add (and update) the [reportback endpoints](https://github.com/DoSomething/phoenix/wiki/API#reportbacks) to the [documentation folder](https://github.com/DoSomething/phoenix/tree/dev/documentation)

katiecrane:
@angaither I have already handwritten a note to do so! 😄

katiecrane:
I should probably read that I'm still not good at it

angaither:
👍 🚢

Issue #4128, opened 2015-03-06
API member count is the wrong request type


angaither:
`/users/member_count`
This is a `POST` request, it should be a `GET`

aaronschachter:
This is because it's using a Services `targeted_action` and I was following examples of existing modules.  can be refactored to use a  Services `relationship` but will need to give a heads up to @DeeZone if/when this happens because email things use this endpoint.

deezone:
Easy to update: https://github.com/DoSomething/mb-toolbox/blob/master/src/MB_Toolbox.php#L179-L210 

The deployment would be the challenge. Any chance of doing this in steps where a GET endpoint could be added while the POST continues to work for awhile? Perhaps GET `/users/member_count` as having "get_" in the URL is not sounding very "RESTful" anyways.

angaither:
also, true @DeeZone 

I don't know the details of the services thing, but I would assume if we created the new endpoint without the `get` in the url, both would work at the same time.

weerd:
:+1: @DeeZone 

Could even simplify it as `/users/count`. We shouldn't have verbs other than the HTTP verb in our enpoints!

aaronschachter:
Worth noting that in the world of DS, the Members Count is different than Users count.  `users/count` looks like total # of users -- this endpoint returns total # of members.

weerd:
`users/members/count` ?

Just tossing out ideas.

DFurnes:
:+1: to changing to a `GET` and  removing `get_` from the endpoint URL. Also makes sense to "deprecate" the old `POST` endpoint but not fully remove for a while.

As for users vs. members, perhaps `users/count` and `members/count`?

weerd:
:+1: @DFurnes I like that better. `/members/count` :dancer:

itsjoekent:
Yes to the deprecating but keeping for a while. The current app that does the member counting for the homepage & dashboard relies on the post _but_ I'm discontinuing that app and migrating it's functionality into dashboard 2.0 so I don't need it for too long

deezone:
My impression is to be even more "RESTful" there would be a parameter in the root GET `/count` request:
- GET `/count?type=user`
- GET `/count?type=member`

But this is stuff is like a religion, everyone has an opinion.

weerd:
@DeeZone lol. I think I'm slowly finding that out, haha.

itsjoekent:
[...]

PR #6541, opened 2016-06-03
Fix get_member_count API documentation

sergiitk:
#### What's this PR do?

`POST /api/v1/users/get_member_count` documentation: fix api endpoint path. Add `/users/` prefix.
It doesn't work without it. See [source](https://github.com/DoSomething/phoenix/blob/8841506956bb800c5a22ec7ff44c1bd6fa6861e4/lib/modules/dosomething/dosomething_api/resources/member_resource.inc#L80).

```
curl -X POST http://staging.beta.dosomething.org/api/v1/users/get_member_count                                                                                                                                                                               init phoenix-js


{"formatted":"3,596,441","readable":"3.5 million"}
```

Also, it's better to change the request to GET, but this PR is to fix existing docs.

### DIFF
--- a/documentation/endpoints/users.md
+++ b/documentation/endpoints/users.md
@@ -141,7 +141,12 @@ curl -H "Accept: application/json" http://staging.beta.dosomething.org/api/v1/us
 ## Get Member Count
 
 ```
-POST /api/v1/get_member_count
+POST /api/v1/users/get_member_count
+```
+
+**Example Request:**
+```sh
+curl -X POST http://staging.beta.dosomething.org/api/v1/users/get_member_count
 ```
 
 **Example Response:**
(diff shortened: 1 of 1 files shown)


angaither:
👍

sergiitk:
@angaither Just a note: I tried to make this available via GET, but [it](https://www.drupal.org/node/783254) is [impossible](https://github.com/kylebrowning/services/blob/7.x-3.x/servers/rest_server/includes/RESTServer.inc#L567). Services module is stupid. They decided REST 'action' members can be `POST` only.

angaither:
haha, okay, thanks for the note @sergii-tkachenko 
 I closed out the issue for that (#4128)

itsjoekent:
good ol drupal

PR #6549, opened 2016-06-06
Remove tests

### LINKED ISSUE #6486: Remove tests?
#### Current Behavior

The functional and unit tests are out of date and have not been updated (or run) in a very long time.
#### Proposed Behavior

Since these aren't being used, we should consider removing them from the repository.
#### Why This Matters

We shouldn't have code in the repository that isn't being used, since it is confusing for onboarding.
#### Relevant Tickets

References DoSomething/devops#52.

angaither:
#### What's this PR do?
- removes Casper tests
- removes simple tests

👓 

we don't use these, nor do they probably run 💥 
#### Relevant tickets

Fixes #6486


DFurnes:
RIP. 💀 Two other things that could be removed: the [project-wide `package.json`](https://github.com/DoSomething/phoenix/blob/remove-tests/package.json) (which is only used for test suite dependencies) & the [test runner](https://github.com/DoSomething/phoenix/blob/remove-tests/bin/test) and it's references in the `bin/ds` script.

angaither:
ahh, yis, removed those as well @DFurnes

DFurnes:
The test runner is also referenced in the `bin/ds` script! [Here](https://github.com/DoSomething/phoenix/blob/cdd284c18962523730c46cb333f651025077f1a1/bin/ds#L32-L36), [here](https://github.com/DoSomething/phoenix/blob/cdd284c18962523730c46cb333f651025077f1a1/bin/ds#L104-L112), and [here](https://github.com/DoSomething/phoenix/blob/cdd284c18962523730c46cb333f651025077f1a1/bin/ds#L658-L666).

DFurnes:
Looks good to me, burn it down! 🔥 🚒

PR #6553, opened 2016-06-06
Add organize config form

### LINKED ISSUE #6537: Organize config options
#### User Story

As an admin, I would like to switch between using Organize's testing api and their prod api without a code deploy. We can currently do this for the API tokens, but not the base URLs.

During development/testing we should use Organize's testing server at: `'https://fiftythree-dev.organize.org/api/v2/` which we have a token for.

When we move to deploy this, we need to use their production server located at : `'https://fiftythree.organize.org/api/v2/` with a new token they will eventually provide for us.
#### Why This Matters

If we hardcode these values in the code, we will have to do a code deploy to switch them out. If we add them as admin settings, then we can switch them out in the Custom Settings screen. 

Hard coding values in the code.

angaither:
#### What's this PR do?
- remove api token settings from a single node
- add api token and api url settings to a form @ `/admin/config/services/organs`
- updated calls in the zipcode GET to use the new config form settings
- add an Auth token to the `/admin/config/services/organs` form and then
- load up  `http://dev.dosomething.org:8888/organ-donation/send-postal-code?postal_code=11217` and see that it returns data
#### Relevant tickets

Fixes #6537

### DIFF
--- a/lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.admin.inc
+++ b/lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.admin.inc
@@ -20,7 +20,7 @@ function dosomething_organize_config_form($form, &$form_state) {
     '#title' => t('Organize URL'),
     '#description' => t('Without the trailing slash'),
     '#required' => TRUE,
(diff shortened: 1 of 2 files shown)


chloealee:
so cool! 👍

Issue #6547, opened 2016-06-06
Create backend api calls


angaither:
- [x] `GET`  with zipcode
- [x] `POST` with registration

PR #6558, opened 2016-06-06
Update the json output, to take out the response.

angaither:
#### What's this PR do?

update the json response from the organize api to return the data object on the response

👀 

we don't want to include the whole response from the HTTP request
#### Relevant tickets

Refs #6547


chloealee:
👍

PR #6560, opened 2016-06-07
Fixes for Kudos DBLog notices/warnings/errors

weerd:
#### What's this PR do?

This PR fixes a bunch of notices/warnings/errors that were showing up in the DBLog for a bunch of kudos related things, particularly for when a placeholder is used in the gallery instead of an actual Reportback Item. Placeholders shouldn't be treated as RB items otherwise a bunch of values that are expected to exist on the placeholder lead to errors when they're not there.

Load a campaign with reportbacks and one that's fresh and has a bunch of placeholders still in the gallery, then check the dblog in the Drupal CMS.
#### Relevant tickets

Fixes # 💅 

---

@DFurnes @angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
+++ b/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
@@ -195,8 +195,9 @@ function dosomething_kudos_get_totals_by_taxonomy_term($data) {
  * @return int
  */
 function dosomething_kudos_get_term_id_by_name($name) {
-  return array_pop(taxonomy_get_term_by_name($name, 'kudos'))->tid;
+  $term = taxonomy_get_term_by_name($name, 'kudos');
 
+  return array_pop($term)->tid;
 }
 
 /**
@@ -233,3 +234,62 @@ function dosomething_kudos_get_kudos_for_user_by_file($fid) {
 
   return $kudos;
 }
+
+/**
+ * Check to see if the kudos term has been selected by the user on specified
+ * reportback.
+ *
+ * @param  [type] $id
+ * @param  object $reportbackItem
+ * @return bool
+ */
+function dosomething_kudos_term_is_selected($id, $reportbackItem) {
+  if (! $reportbackItem->existing_kids) {
+    return false;
+  }
+
+  if (! isset($reportbackItem->existing_kids[$id])) {
+    return false;
+  }
+
+  return true;
+}
+
+/**
+ * Get the icon associated with the kudos term id.
+ *
+ * @param  string $id
+ * @return string
+ * @TODO: this will eventually be changed to refere to SVG icon paths for our custom kudos icons!
+ */
+function dosomething_kudos_get_icon_by_term_id($id) {
+  settype($id, 'string');
+
+  $icons = [
+    '641' => '&#128150;',
+    '646' => '&#128169;',
+  ];
+
+  return $icons[$id];
+}
+
+/**
+ * Get the count for a kudos term on a Reportback item by the
+ * kudos term id.
+ *
+ * @param  string $id
+ * @param  object $reportbackItem
+ * @return string
+ */
+function dosomething_kudos_get_count_by_term_id($id, $reportbackItem) {
+  settype($id, 'string');
+
+  $count_term = [
+    '641' => 'likes',
+    '646' => 'poos',
+  ];
+
+  return $reportbackItem->$count_term[$id];
+}
+
+
--- a/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
+++ b/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
@@ -239,16 +239,16 @@ function dosomething_kudos_get_kudos_for_user_by_file($fid) {
  * Check to see if the kudos term has been selected by the user on specified
  * reportback.
  *
- * @param  [type] $id
- * @param  object $reportbackItem
+ * @param  string $id
+ * @param  object $reportback_item
  * @return bool
  */
-function dosomething_kudos_term_is_selected($id, $reportbackItem) {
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -233,3 +234,62 @@ function dosomething_kudos_get_kudos_for_user_by_file($fid) {
...
+
+/**
+ * Check to see if the kudos term has been selected by the user on specified
+ * reportback.
+ *
+ * @param  [type] $id
+ * @param  object $reportbackItem
+ * @return bool
+ */
+function dosomething_kudos_term_is_selected($id, $reportbackItem) {
```
angaither: we are 🐫 s in 🐘 , not 🐍 s

weerd: We're 🐍 in PHP in Droooops though 😞

DFurnes: Not suggesting as something for this PR, but I'd love to consider adopting our PSR-inspired code style in Phoenix as well, since it is can be pretty confusing switching back and forth from Dries-mode.

weerd: wait, did you mean the function name or the parameter?

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -233,3 +234,62 @@ function dosomething_kudos_get_kudos_for_user_by_file($fid) {
 
   return $kudos;
 }
+
+/**
+ * Check to see if the kudos term has been selected by the user on specified
+ * reportback.
+ *
+ * @param  [type] $id
```
DFurnes: I think you're missing a type here. `[type]` 👀

weerd: it's type _type_, hehe. fixing!

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -233,3 +234,62 @@ function dosomething_kudos_get_kudos_for_user_by_file($fid) {
...
+
+/**
+ * Get the count for a kudos term on a Reportback item by the
+ * kudos term id.
+ *
+ * @param  string $id
+ * @param  object $reportbackItem
+ * @return string
+ */
+function dosomething_kudos_get_count_by_term_id($id, $reportbackItem) {
```
angaither, line 284: we shouldn't need this function if we can call `dosomething_kudos_get_total_for_file_by_term_name`

weerd, line 284: oh cool. Didn't know about that one. Will give it a whirl :)

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -233,3 +234,62 @@ function dosomething_kudos_get_kudos_for_user_by_file($fid) {
...
+ *
+ * @param  string $id
+ * @return string
+ * @TODO: this will eventually be changed to refere to SVG icon paths for our custom kudos icons!
+ */
+function dosomething_kudos_get_icon_by_term_id($id) {
+  settype($id, 'string');
+
+  $icons = [
+    '641' => '&#128150;',
```
angaither, line 269: these `TIDs` (since currently we are only using one sitewide) should probably be defined at the top of the file as a constant. I would imagine once we start having overrides per campaign, we will want to pull this from a global variable, but we can just hardcode the one for now.

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
angaither: I meant it the other way, 🐍  not 🐫 . for the `reportbackItem` variable.

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -233,3 +234,62 @@ function dosomething_kudos_get_kudos_for_user_by_file($fid) {
...
+
+/**
+ * Get the icon associated with the kudos term id.
+ *
+ * @param  string $id
+ * @return string
+ * @TODO: this will eventually be changed to refere to SVG icon paths for our custom kudos icons!
+ */
+function dosomething_kudos_get_icon_by_term_id($id) {
+  settype($id, 'string');
```
DFurnes: Is it necessary to cast the variable since we ask for a string? If so, would it be clearer to cast `$id = (string) $id`?

angaither: 👏

weerd: @DFurnes I wanted to cask just in case, since we know that sometimes values come in as different types from the Droops. I can change; I just learned about `settype()` and felt pretty readable but if the casting approach is clearer I can update to that.

DFurnes: Personally, I think I'd rather we trigger warnings so we don't sweep mistakes under the rug. Maybe once all of our environments are updated to PHP 7 we can consider starting to use scalar type hints.

weerd: I agree but we also need to get into the habit of checking the DBLogs more often to see and address the warnings, cause a bunch of them were being thrown with similar things that were ignored before 😬 Plus, some are less noticeable without a slew of tests that we can't do in Droopy Droops.

PR #6563, opened 2016-06-07
Fixes for Kudos DBLog notices/warnings/errors (take 2)

weerd:
#### What's this PR do?

This PR fixes a bunch of notices/warnings/errors that were showing up in the DBLog for a bunch of kudos related things, particularly for when a placeholder is used in the gallery instead of an actual Reportback Item. Placeholders shouldn't be treated as RB items otherwise a bunch of values that are expected to exist on the placeholder lead to errors when they're not there.

Load a campaign with reportbacks and one that's fresh and has a bunch of placeholders still in the gallery, then check the dblog in the Drupal CMS.

Ref #6560 This PR takes a simpler approach to only account for a single kudos item on RB items in gallery at the moment, instead of accounting for multiple.
#### Relevant tickets

Fixes # 💅 

---

@DFurnes @angaither


DFurnes:
🚢 ✅

PR #6570, opened 2016-06-08
Add magic login link functionality.

### LINKED ISSUE #6555: Create a login endpoint
#### User Story

As a user I can get sent a link (via the mobile app) and then click that link and be magically logged in to the phoenix site

We need to expose a new endpoint on the phoenix api that sends Admin users a magic link to log a person in. 
#### Why This Matters

Organs.

DFurnes:
#### What's this PR do?

This PR adds an endpoint that allows admins to request single-use "magic" login links via the API, and the corresponding route that logs a user into the application when given the magic link.

This is heavily inspired by Drupal's built-in reset password link functionality (and uses the exact same [`user_pass_rehash`](https://api.drupal.org/api/drupal/modules%21user%21user.module/function/user_pass_rehash/7.x) method to generate the secure token), but isn't specifically coupled to sending a user to the profile page.

Check `member_resource.inc` for the admin endpoint that creates the login links, and `dosomething_user.module` for the route that logs people in. I'd also love extra special attention to the `dosomething_user_check_magic_login_validity` to make sure there aren't any mistakes in there! 👀 

🍡 
#### Relevant tickets

Closes #6555.
#### Checklist

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -1806,7 +1806,7 @@ function dosomething_user_magic_session($uid, $timestamp, $login_token) {
   // which prevents further use of the one-time login link.
   user_login_finalize();
 
-  watchdog('user', 'User %name used one-time login link at time %timestamp.', ['%name' => $account->name, '%timestamp' => $timestamp]);
+  watchdog('user', 'User %uid used one-time login link at time %timestamp.', ['%name' => $account->uid, '%timestamp' => $timestamp]);
   drupal_goto($redirect, ['external' => FALSE]);
 }
 
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -109,6 +109,11 @@ function dosomething_user_menu() {
     'access callback' => 'user_view_access',
     'access arguments' => array(1),
   );
+  $items['user/magic/%'] = array(
+    'page callback' => 'dosomething_user_magic_session',
+    'page arguments' => array(2, 3, 4),
+    'access callback' => TRUE,
+  );
   return $items;
 }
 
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -1762,3 +1767,77 @@ function dosomething_user_mail($key, &$message, $params) {
...
+  }
+
+  // Looks good! Set the new user via the global.
+  $user = $account;
+
+  // Use user_login_finalize() to set the login timestamp of the user,
+  // which prevents further use of the one-time login link.
+  user_login_finalize();
+
+  watchdog('user', 'User %name used one-time login link at time %timestamp.', ['%name' => $account->name, '%timestamp' => $timestamp]);
```
angaither: let's use `uid` here `name` can be nutty #2753

DFurnes: Good call. Changed! 👍

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -1762,3 +1767,77 @@ function dosomething_user_mail($key, &$message, $params) {
...
+ * @param $login_token
+ * @return bool
+ */
+function dosomething_user_check_magic_login_validity($account, $timestamp, $login_token) {
+  if(! $account) {
+    return false;
+  }
+  
+  $timeout = variable_get('user_password_reset_timeout', 86400);
+  $has_expired = REQUEST_TIME - $timestamp > $timeout;
```
angaither, line 1828: maybe we should custom expire these? (cc @mshmsh5000)

mshmsh5000, line 1828: Because we might want a different value than `user_password_reset_timeout`?

angaither, line 1828: Doesn't look like we actually have a value for that, so `86400` is 24 hours? that seems okay. I just remember our logout time was longggg, so I didn't want these hanging out forevah

DFurnes, line 1828: @angaither Yep, that's the same default that the password reset links use.

angaither:
🚢  MAGIC 🎱

lib/modules/dosomething/dosomething_user/dosomething_user.module
mshmsh5000, line 1828: I think we can 👍 for now as 24 hrs seems fine for both, and add a separate expiry value here if we ever want to manage them separately.

DFurnes, line 1828: Makes sense to me!

Issue #6458, opened 2016-05-18
set default kudos for entire site


angaither:
- replace the `heart` emoji with the SVGs provided

angaither:
@lkpttn we are using the heart and here is the SVGs
https://dosomething.slack.com/files/piercey/F1EQSUM9P/gray-heart.svg
https://dosomething.slack.com/files/piercey/F1EP7FJHZ/pink-heart.svg

lkpttn:
Here's what it looks like in practice:

![kudos gallery](https://cloud.githubusercontent.com/assets/1865372/15870725/31f3b146-2cbf-11e6-99a6-e14850ab59a0.png)

Icon is 30px wide. Font is the gamma heading, which should be roughly 18px.
12px to the left and bottom edges. The text is normal #4a4a4a.

angaither:
@deadlybutter mind picking this up next?

itsjoekent:
@angaither sure

itsjoekent:
talked w/ the 😎 fenders -- going to 

1- use CSS to do this
2- switch the <a> to a button because that makes more sense

PR #6577, opened 2016-06-09
Updating kudos based on new design

### LINKED ISSUE #6543: Update Kudos config
instead of enabling per campaign, this should be enabled by default and then the box will be disable per campaign

image

itsjoekent:
~~not ready for review yet~~
#### What's this PR do?
- Adds SVG's for the heart instead of character code
- Removes the button from the markup
  - I was starting to override nearly everything in the .button & etc classes / markup I removed. It made much more sense to start fresh I think.
- small style tweaks based on the new mock that can be found [here](https://github.com/DoSomething/phoenix/issues/6458#issuecomment-224378410)

When page loads, does the button accurately reflect the correct state?
If you click the button, does it change correctly?

screen shot 2016-06-09 at 12 27 39 pm

screen shot 2016-06-09 at 12 27 46 pm
#### Relevant tickets

Fixes #6543


lib/themes/dosomething/paraneue_dosomething/templates/system/patterns/photo.tpl.php
```diff
@@ -17,7 +17,7 @@
   <?php if ($content->id && !$content->disable_reactions): ?>
     <ul class="form-actions -inline photo__actions">
       <li>
-        <a class="button -mini js-kudos-button <?php print dosomething_kudos_term_is_selected($content, 'heart') ? 'is-active' : '' ?>" data-kudo-id="<?php print array_pop($content->allowed_reactions); ?>">&#128150;</a>
+        <img class="js-kudos-button <?php print dosomething_kudos_term_is_selected($content, 'heart') ? 'is-active' : '' ?>" data-kudo-id="<?php print array_pop($content->allowed_reactions); ?>" src="https://files.slack.com/files-pri/T024GV2BW-F1EQSUM9P/download/gray-heart.svg" >
```
weerd: you should still wrap it in a button tag cause that's what it is.

weerd: I would still vote for jut adding the svg via CSS.

weerd: And why are we grabbing the image from Slack?

weerd:
Sorry, missed the "not ready for review" title. But added some comments.

lib/themes/dosomething/paraneue_dosomething/templates/system/patterns/photo.tpl.php
itsjoekent: PR at top says "not ready for review yet" for this reason, was easier to just use the slack url in my clipboard

itsjoekent: also it does change it via CSS, this is just the default svg

weerd:
So one cool thing is you can actually _style_ elements on an SVG. So what could be cooler is to add the SVG via CSS... just one item, and then when the `is-active` class is added, actually use CSS to change the color of the icon and remove the border. Just a thought and might not be feasible moving forward with other icons whenever that happens... Just thinking out loud.

itsjoekent:
yeah thats how D3 works (styling svg's), im gonna say 👎 only because any design changes in the future would require a developer to update CSS, and it could get insane if we go beyond hearts. (eg: imagine doing a thumb...)

weerd:
true true.

itsjoekent:
@weerd if i wrap this in a button, even if I dont add our Forge button classes it looks like this

screen shot 2016-06-09 at 1 39 13 pm

which means id now need to add more CSS to remove the default button styling (if we added .btn its the same problem). it seems very silly to create more work & code to maintain just for the sake of having a button tag wrap this

itsjoekent:
also @weerd its now fully using CSS to set the SVG's, used the :not selector to handle the grey heart

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_counter.scss
```diff
@@ -1,12 +1,35 @@
...
+}
+
+.photo__actions li {
+  * {
+    display: inline;
+    vertical-align: middle;
+  }
+
+  img {
+    width: 30px !important;
```
DFurnes: Does this need to be `!important`?

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_counter.scss
```diff
@@ -1,12 +1,35 @@
...
+  padding-left: 4px;
 
   &.is-active {
-    color: $purple;
+    color: $dark-gray;
+  }
+}
+
+.photo__actions li {
+  * {
```
DFurnes: We should avoid using the `*` selector if possible... this could cause unexpected bugs if we ever nested anything else within this container.

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_counter.scss
itsjoekent: sadly yes, default value it was using is 100%

screen shot 2016-06-09 at 1 51 29 pm

also cant really make it much more specific than this without it looking weird (I think?).

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_counter.scss
```diff
@@ -1,12 +1,35 @@
...
+
+    &:hover {
+      cursor: pointer;
+    }
+
+    &.is-active {
+      content: url("../images/pink-heart.svg");
+    }
+
+    &:not(.is-active) {
```
DFurnes: I don't believe IE8 or IE9 support the `:not` operator, and (unfortunately!) we still support those. Would it be possible to have this as the default styling for the element and just override it if the `.is-active` class is applied?

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_counter.scss
itsjoekent: gotcha, looked prettier than element comma element so just went with that. i can easily switch this though

DFurnes:
The SVGs you're using here are pretty massive (~1MB each)... could you try running them through some sort of optimizer (like [SVGOMG](https://jakearchibald.github.io/svgomg/)) to see if you can cut them down to a more reasonable size?

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_counter.scss
itsjoekent: I can take a screenshot if this doesn't make sense, but this was actually the first thing i tried. the problem was if the page loaded and the is-active class was already applied, and you unclicked the active kudo, the default gray one didn't show.

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_counter.scss
DFurnes: Maybe we could update that `.photo.-framed img` rule to only apply to `img` which are direct children of the `.wrapper` (or better yet, add a class to that image and style that!)

itsjoekent:
@DFurnes sure i can give that a whirl

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_counter.scss
DFurnes: Can you set the SVG as a `background-image` of the button? I'm not sure the `content` property is intended to be used outside of `:before` and `:after` pseudo-elements so there might be bugs.

DFurnes:
> which means id now need to add more CSS to remove the default button styling (if we added .btn its the same problem). it seems very silly to create more work & code to maintain just for the sake of having a button tag wrap this

It sounds like we're not properly resetting browser default styling on `<button>` elements in our Forge CSS reset. I think it'd make sense to fix that in Phoenix and make an issue to port it back to Forge.

itsjoekent:
tried using SVGOMG but it had an error trying to read the files. any other tools you know of? @DFurnes

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_counter.scss
itsjoekent: Added a custom class but this rule is still overriding the 30px width. I'll track down where the .photo.-framed img is coming from...

itsjoekent: hmmm, looks like this is doing it https://github.com/DoSomething/phoenix/blob/c430155c6a8636fd4eb3d5fc0ac803912ec6714e/lib/themes/dosomething/paraneue_dosomething/scss/patterns/_photo.scss#L27

itsjoekent: great fixed that rule w/ your suggestion

DFurnes:
I wasn't able to import it into SVGOMG or open it in Illustrator either... I think the file may be corrupted. I was able to open it and re-save for web in Sketch, and got it down to 379 bytes.

itsjoekent:
@DFurnes can you send me those files? i dont have sketch

DFurnes:
@deadlybutter Here you go: [icons.zip](https://github.com/DoSomething/phoenix/files/307765/icons.zip)

itsjoekent:
@DFurnes thanks!

itsjoekent:
@DFurnes I think i've fixed all the styling / markup. Wanna take another look? ~~(still need to import the new SVG's you sent me but I wanted to fix the merge conflict first)~~

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_photo.scss
```diff
@@ -23,7 +23,7 @@
       padding: ($base-spacing / 2);
     }
 
-    img {
+    .wrapper img {
```
DFurnes: Let's move this `img { ... }` to be nested under the `> .wrapper { ... }` rule above. The `>` is important whenever targeting `.wrapper` since we use "wrapper" as a standard "child" class in a bunch of different patterns.

DFurnes:
One more comment, but everything else looks good to me! 👍

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_photo.scss
itsjoekent: Ah yeah good catch, wouldn't wanna break of images unintentionally

weerd:
Nice work. Looks better :)

itsjoekent:
@DFurnes fixed!

DFurnes:
Ship it! :shipit:

itsjoekent:
CHOO CHOO 🚂 🚋 🚋 💥

weerd:
lol

PR #6585, opened 2016-06-09
Exposing kudos reactions on view more RB items.

### LINKED ISSUE #6544: Allow kudos actions on 'view more' reportbacks
Reportback items rendered on the frontend after clicking `view more` need to have kudos options on them

### LINKED ISSUE #6575: Add property to global DS JS object to indicate whether kudos reactions active on campaign
Need to add a boolean property that can be checked by the "view more" script in the campaign galleries to know if it needs to append kudos data for the Reportback Items it loads via AJAX.

weerd:
#### What's this PR do?

This PR exposes the full functionality for kudos reactions for the Reportback Items that are dynamically added via AJAX request to the API.

Pull down code, load a campaign and check the gallery. Click on "view more" to load more items into the gallery and see if the kudos appear (if kudos are activated for that campaign). Click on a kudos, etc and see if that all seems to work right.

We'll likely need to address https://github.com/DoSomething/phoenix/issues/6594 for how kudos are accounted for and rendered in the gallery and moving forward if lots of kudos reactions get added to items. Refer to that issue for more on that...

_Also_ I need to update the JS template to account for the new changes to the kudos structure that was completed in #6577
#### Relevant tickets

Fixes #6544
Fixes #6575

---

@DFurnes 

cc: @angaither

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
@@ -18,6 +18,13 @@ const Reportback = {
 
   apiUrl: '/api/v1/reportback-items',
 
+  reactions: {
+    approved: ['heart'],
+    inventory: {
+      heart: '641', // @TODO: totally temporary! Will be obtained via global object passed in via PHP.
+    },
+  },
+
   nid: null,
 
   itemIds: null,
@@ -179,6 +186,38 @@ const Reportback = {
     });
   },
 
+  /**
+   * Get Kudos Reactions for a Reportback item.
+   *
+   * @param  {array} data  Array of Kudos objects from API.
+   * @return {array}
+   */
+  getReactions: function(data) {
+    console.log(data);
+    let reactions = [];
+
+    for (let i = 0, count = this.reactions.approved.length; i < count; i++) {
+      const term = this.reactions.approved[i];
+      const termId = this.reactions.inventory[term];
+
+      const reaction = {
+        term: term,
+        termId: termId,
+        total: 0,
+        userReacted: false,
+      };
+
+      for (let j = 0, count = data.length; j < count; j++) {
+        if (data[j].term.name === term) {
+          reaction.total = data[j].kudos_items.total;
+        }
+      }
+
+      reactions.push(reaction);
+    }
+
+    return reactions;
+  },
 
   /**
    * Initialize the ImageUpload interface.
@@ -238,6 +277,8 @@ const Reportback = {
         }
       },
     }).done(function(response) {
+      // console.log(response);
+
       _this.apiUrl = response.meta.pagination.links.next_uri || null;
 
       _this.templatize(response);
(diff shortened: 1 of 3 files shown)


lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
```diff
@@ -179,6 +186,38 @@ const Reportback = {
...
+   * Get Kudos Reactions for a Reportback item.
+   *
+   * @param  {array} data  Array of Kudos objects from API.
+   * @return {array}
+   */
+  getReactions: function(data) {
+    console.log(data);
+    let reactions = [];
+
+    for (let i = 0, count = this.reactions.approved.length; i < count; i++) {
```
weerd: @DFurnes was considering pulling in [lodash's `forEach`](https://lodash.com/docs#forEach) for some of these loops I have. Thoughts?

DFurnes: Sounds good to me! :thumbsup: I think we might even already use it elsewhere in the codebase. You might also want to look into `map` since you're returning a mutated version of each of the elements in the `this.reactions` array.

``` js
const reactions = this.reactions.approved.map((reaction) => {
  const term = this.reactions.approved[i];
  // ...
  return reaction;
});
```

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
```diff
@@ -179,6 +187,48 @@ const Reportback = {
...
   },
 
+  /**
+   * Get Kudos Reactions for a Reportback item.
+   *
+   * @param  {array} data  Array of Kudos objects from API.
+   * @return {array}
+   */
+  getReactions: function(data) {
+    const _this = this;
```
DFurnes: You can use arrow functions to preserve lexical `this` now! :)

``` js
forEach(this.reactions.approved,(term) => {
  // ...
  forEach(data,(record) => {
    // ...

    if (this.reactions.user) {
      // ...
    }

});
```

lib/themes/dosomething/paraneue_dosomething/js/reportback/templates/reportback.tpl.html
```diff
@@ -1,10 +1,21 @@
...
   <figure class="wrapper">
     <img src="data:image/gif;base64,R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs=" alt="<%= caption %>" data-src="<%= image %>" />
     <figcaption class="photo__copy">
       <p class="photo__caption"><%= caption %></p>
     </figcaption>
   </figure>
+
+  <% if (reactions.enabled) { %>
+  <ul class="form-actions -inline photo_actions">
+    <% for (i = 0, count = reactions.data.length; i < count; i++) { %>
```
DFurnes: I think you still need a `var` or `let` here to avoid this being made into an implicit global, right? I'd also vote for using the `reactions.data.length` in the condition clause for readability, since it isn't a big performance hit to re-check the length property on each iteration.

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
weerd: Gawd I keep forgetting @_@ Thanks for the reminder :)

lib/themes/dosomething/paraneue_dosomething/js/reportback/templates/reportback.tpl.html
weerd: Yep yep! Caveat from jumping over from PHP to JS @_@ haha

DFurnes: It's such a struggle! 😆

weerd: image

DFurnes:
:+1:

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
```diff
@@ -265,11 +313,16 @@ const Reportback = {
 
     for (let i = 0, count = items.length; i < count; i++) {
```
itsjoekent, line 314: not blocking this PR, more of a question. how ES6-y are we trying to be? eg: this could be

```
for (let item of items) {
  ...
}
```

weerd:
Oh, nice @deadlybutter. I can do a later update to address that :)

Issue #6548, opened 2016-06-06
Create front end registration form


angaither:
- Create form containing results of fields returned from #6547's `GET` request 
- This should have our JS validation we already have
- Should `POST` to #6547's `POST` endpoint

angaither:
closed with #6630

PR #6587, opened 2016-06-09
Organ post zip

sbsmith86:
#### What's this PR do?

zip-code-get-example

This starts to chip off at the flow for organ donation. With this PR: 
- A user is prompted to enter their zip code. 
- Client-side validation is implemented around the zip code field which should throw errors for incorrect zip codes, empty field, or the zip isn't found. 
- Based on the response from the Organize API, we output a form with the necessary fields the user should fill out.

⚠️  This doesn't address a bunch of things, but want to start breaking this up into smaller pieces:

Does not:
- Handle submission of the registration form
- Handle validation on the registration form

There are probably a bunch of places I can clean this code up and I will be working on that as I go. This PR is already getting large...
#### Relevant tickets

Ref #6548


lib/themes/dosomething/paraneue_dosomething/js/takeover/Takeover.js
```diff
@@ -89,55 +71,21 @@ class Takeover {
...
-        _this.toggleSpinner(request.button);
-      },
-      complete: function() {
-        _this.toggleSpinner(request.button);
+        'X-CSRF-Token': $('meta[name="csrf-token"]').attr('content')
       }
-    });
+    })
+    .then(response => response.json())
+    .then(json => request.done(json));
```
DFurnes: Instead of passing a callback in the `request` parameter here, you can return the Promise from `fetch()` and continue chaining `.then()` on it. So you'd be able to call this method like:

``` js
this.sendRequest({
  url    : `${this.postalCodeUrl}?postal_code=${zipCode}`,
  method : 'GET'
}).then(json => {
  // ...
});
```

sbsmith86: ooo cool. ya.

lib/themes/dosomething/paraneue_dosomething/js/takeover/Takeover.js
```diff
@@ -161,6 +109,78 @@ class Takeover {
...
+
+  /*
+   * Resize an elemen based on the height of it's contents
+   *
+   * @param  {jQuery}   $container
+   * @param  {jQuery}   $referenceHeight
+   * @param  {int}      padding
+   */
+  resizeContainerHeight($container, $referenceHeight, padding = null) {
+    padding = (padding === null) ? 100 : padding;
```
DFurnes: Can you use `100` as the default `padding` parameter here rather than `null` and the ternary?

sbsmith86: lol yep def.

lib/themes/dosomething/paraneue_dosomething/js/campaign/OrganDonation.js
```diff
@@ -1,84 +1,136 @@
 const $ = require('jquery');
+const forEach = require('lodash/collection/forEach');
+const template = require('lodash/string/template');
+
 import setting from '../utilities/Setting';
 import Takeover from 'takeover/Takeover';
 
+// Templates
+import organDonationTpl from 'campaign/templates/OrganDonation/takeover.tpl.html';
```
DFurnes, line 9: Not needed for this PR, but I'd love to think about loading these from a `<script type="text/html">` on the action page so we don't need to include all this markup in our site-wide script bundle. (Or maybe starting to break things into different entry points so we only load all these scripts on pages where they're used.)

sbsmith86:
@DFurnes Thanks! Addressed the comments you made so far. I work on loading the scripts better as we get closer to finishing this project up.

DFurnes:
Sounds good! 👍

Issue #6583, opened 2016-06-09
Display unique share link in campaign template


angaither:
#### User Story

As a user, when I finish registering to be an organ donor, I want to be able to access my social share links if I revisit the campaign page. 
#### Additional Information (optional)

This should display in the "stuff you need" section, if the share link is enabled for that campaign
#### Why This Matters

because Adam said so.

ngjo:
Full conversation in Slack, but for recap:
- Put the unique social shares in the "Stuff You Need" section
  -If the double social call outs (The Problem and Stuff You Need sections) is confusing, we can hide the Problem share buttons via optimizely

Thanks!

angaither:
currently this is tied to the organ donation campaign & it shouldn't be 💯

ngjo:
@angaither we're able to edit the instructional text that gives context to the custom social link? if so, is that something the campaign lead can do on their own or does it require dev support?

angaither:
@ngjo the link has custom social share language which is in the `custom settings` section of the node

ngjo:
woohoooo! i'll try to test that out once this is un-tied from organz!

PR #6595, opened 2016-06-10
Custom setting unique social share links and text 

### LINKED ISSUE #6584: Create custom setting to enable & configure unique social share link
#### User Story

As an admin I would like to enable a unique social share link per campaign.
As an admin I would like to provide custom text to share with the link. 

This should be a toggle in the `custom settings` of a campaign.
There should also be a text field (it would be cool it if also had a counter)
#### Why This Matters

We can customize the share per campaign.

chloealee:
#### What's this PR do?

Creates a custom setting to enable and configure unique social share links. 
- Click on a campaign 
- Go to the "CUSTOM SETTINGS" tab
- Click the "Create unique social share links for each user" check box under "CUSTOM SOCIAL SHARING" 
- Add dummy text in "Custom social share language" field
- Save and `drush cc all`. 
- Click "VIEW" tab. 
- Click on Facebook share - the link should have this pattern `https://www.facebook.com/dialog/feed?app_id=1374507732873066&display=popup&link=http%3A//dev.dosomething.org%3A8888/us/campaigns/test-campaign%3Fsource%3D1704953%` etc. with `source` = your `uid`. 
- Click on Twitter share - you should see your dummy text in the caption pop pre-populated along with a URL with the following pattern: `http://dev.dosomething.org:8888/us/campaigns/test-campaign?source=1704953?` etc. with `source` = your `uid`.
- Tumblr share should be the same (Tumblr won't allow us to can't test on dev) 
- Go back to the "CUSTOM SOCIAL SHARING" tab. 
- Uncheck box and delete your custom social share language. 
- Save and `drush cc all`
- Click on Facebook share - you should no longer see `source` with your `uid` in the URL. 
- Click on Twitter share - again, you shouldn't see `source` with your `uid` in the URL and caption should just be campaign's CTA. 
#### Relevant tickets

Fixes #6584

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -126,6 +126,15 @@ function dosomething_campaign_preprocess_common_vars(&$vars, &$wrapper) {
     $country_prefix = (dosomething_global_get_current_prefix()) ? strtoupper(dosomething_global_get_current_prefix()) : 'global';
     $problem_share_image = $base_url . 'sites/default/files/images/problem-' . $vars['nid'] . '-' . $country_prefix . '.png';
 
+    if (isset($campaign->variables['social_share_custom_text'])) {
+      $tweet = $campaign->variables['social_share_custom_text'];
+      $tumblr_caption = $campaign->variables['social_share_custom_text'];
+    }
+    else {
+      $tweet = isset($campaign->fact_problem) ? $campaign->fact_problem['fact'] : NULL;
+      $tumblr_caption = t("This is REAL, do something about this with me: ");
+    }
+
     $share_types = array(
       'facebook' => array(
         'type' => 'feed_dialog',
@@ -134,16 +143,22 @@ function dosomething_campaign_preprocess_common_vars(&$vars, &$wrapper) {
         ),
       ),
       'twitter' => array(
-        'tweet' => (isset($campaign->fact_problem)) ? $campaign->fact_problem['fact'] : NULL,
+        'tweet' => $tweet,
       ),
       'tumblr' => array(
         'posttype' => 'photo',
         'content' => $problem_share_image,
-        'caption' => t("This is REAL, do something about this with me: "),
+        'caption' => $tumblr_caption,
       ),
     );
 
-    $vars['share_bar'] = dosomething_helpers_share_bar($campaign_path, $share_types, 'problem_shares', 'social-menu -with-callout');
+    if ($campaign->variables['social_share_unique_link'] == 1) {
+      $custom_share_link = $campaign_path . '?source=' . $vars['user']->uid;
+      $vars['share_bar'] = dosomething_helpers_share_bar($custom_share_link, $share_types, 'problem_shares', 'social-menu -with-callout');
+    }
+    else {
+      $vars['share_bar'] = dosomething_helpers_share_bar($campaign_path, $share_types, 'problem_shares', 'social-menu -with-callout');
+    }
   }
 }
 
@@ -393,7 +408,7 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
     ),
   );
 
-  $custom_share_link = $campaign_path . '?source=' . $vars['user']->uid;;
+  $custom_share_link = $campaign_path . '?source=' . $vars['user']->uid;
 
   $organ_donations_share_button_markup = dosomething_helpers_share_bar($custom_share_link, $organ_donation_share_types, 'organ_donation_referral', 'social-menu');
 
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -153,7 +153,7 @@ function dosomething_campaign_preprocess_common_vars(&$vars, &$wrapper) {
     );
 
     if ($campaign->variables['social_share_unique_link'] == 1) {
-      $custom_share_link = $campaign_path . '?source=' . $vars['user']->uid;
+      $custom_share_link = $campaign_path . '?source=user/' . $vars['user']->uid;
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -134,16 +143,22 @@ function dosomething_campaign_preprocess_common_vars(&$vars, &$wrapper) {
         ),
       ),
       'twitter' => array(
```
angaither, line 136: hmmm, I don't think we want to replace the default shares with this copy

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -393,7 +408,7 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
     ),
   );
 
-  $custom_share_link = $campaign_path . '?source=' . $vars['user']->uid;;
+  $custom_share_link = $campaign_path . '?source=' . $vars['user']->uid;
```
angaither: I didn't catch this before but I think we want the `source` to be `source=user/[uid]`

chloealee: I'll change that now!

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
chloealee, line 136: @angaither a little confused here, but maybe I misunderstood the [issue](https://github.com/DoSomething/phoenix/issues/6584)? Is this correct? We want to give admins the ability to change social share text but if they don't customize it, we just populate the share language with the CTA.

angaither, line 136: @chloealee right, but this belongs in it's own section see #6583

chloealee, line 136: @angaither ahh okay got it. Are social shares always displayed in STEP 1 of the campaign? Should we keep this as well and also add it to the "stuff you need" section?

angaither, line 136: AKAIK problem fact shares should be left as is, this should be a new share steps just in the "Stuff you need" section

chloealee, line 136: @angaither one more question! I'm assuming we want to have the share bar here again with the icons but might look weird in the "stuff you need" section since the format of that is text in a bulleted list. Should we just add the share bar in this "STEP 2: PLAN IT" section, perhaps under the "Find a Location" section or just give the user their unique share URL (that might look very ugly just as text)?

angaither, line 136: Yeah, @chloealee the more I thought about it, the more I'm not sure. I think it's worth seeing some mock-ups for different options around this

chloealee, line 136: @angaither okay cool! I can do both and send screen shots and send to the #organ-donation room (just saw you posted in there too - thanks!).

chloealee, line 136: Hi @angaither @ngjo @mikefantini ! Exploring this a little more, adding this to the "Stuff You Need" section is a little more difficult than I thought. Since there isn't a set format (bullet points vs. regular text) and the length will be different for each of the campaigns, it's hard to append to the end of this string: 
screen shot 2016-06-13 at 12 00 16 pm

I also think that without a share bar, sending a user their unique link might be a little weird b/c it's so ugly and long!

I would recommend that we add another share about in this "STEP 2: PLAN IT" section or customizing the share bar at top and guiding the users there to share their particular links. Thoughts?

angaither, line 136: @chloealee you can just remove the changes made to this file and we can figure out this portion with the other issue (#6583)

`git checkout origin/dev -- lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc`

mikefantini:
@chloealee @angaither @ngjo maybe a bad idea but I'll throw it out there:

What if in Stuff You Need we listed as a bullet "Your personal link to get people registered as an organ donor" and this was linked and when clicked opened up a modal that pulled in the share prompt with the social buttons?

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
angaither, line 136: Also, I was thinking we add the changes to the content array [here](https://github.com/DoSomething/phoenix/blob/72e94a3639cff8c75cd167118975a59b591904b8/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc#L210) and not to the backend field

chloealee, line 136: @angaither sounds good, I'll update and remove those changes in `lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc`. Thanks for the github help to do that :) 

Yes - I was trying to add changes to the content [here](https://github.com/DoSomething/phoenix/blob/72e94a3639cff8c75cd167118975a59b591904b8/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc#L210) as well but that is actually a string and not an array so would be difficult to inject to the end without a set pattern (each campaign will have a different amount of bullet points, the content here might not be bullet points etc.). 

Unless, we want this to only be specific for the organ donation campaign, I can do a check to see if the organ donation module is enabled, add a bullet point at the end of this string?

angaither, line 136: @chloealee we wouldn't want this tied specifically to organ-donation checkbox, we would want it tied to the checkbox you created in this PR

angaither:
👍

PR #6606, opened 2016-06-13
changes source param to source=user/[uid]

chloealee:
#### What's this PR do?

Changes custom share link to read `source=user/[uid]` instead of just `source=[uid]`. 

This got deleted in #6595 when we didn't add changes made to `dosomething_campaign.theme.inc`


angaither:
👍

PR #6612, opened 2016-06-20
Some quick kudos cleanup & error fixes

### LINKED ISSUE #6605: User Profile Kudos - Neg Numbers
#### Current Behavior

Kudos going into negative numbers on user profile. 
#### Desired Behavior

Kudos numbers should toggle positively by 1. 
#### Steps to Replicate
- Navigate to user account page
- Click on heart.

screen shot 2016-06-13 at 5 27 57 pm

itsjoekent:
#### What's this PR do?
- Fixes spammy index warnings being thrown
- Prevents Kudos from going negative if the response is bad
- Prevents object/property warnings from being thrown

Can you do Kudos from the user profile page?
Does Kudos still work on the campaign template?

Had the wrong attribute in the markup. #duh
#### Relevant tickets

Fixes #6605

### DIFF
--- a/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
+++ b/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
@@ -245,7 +245,10 @@ function dosomething_kudos_get_kudos_for_user_by_file($fid) {
 function dosomething_kudos_term_is_selected($reportback_item, $term) {
   // We need to check both array & objects b/c preprocess_node works w/ RB Item objects
   // and preprocess_user works w/ content arrays.
-  if ((is_array($reportback_item) && empty($reportback_item['existing_kids'])) && ! $reportback_item->existing_kids) {
+  if (is_array($reportback_item) && empty($reportback_item['existing_kids'])) {
+    return false;
+  }
+  else if (!is_array($reportback_item) && !$reportback_item->existing_kids) {
     return false;
   }
 
--- a/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
+++ b/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
@@ -243,12 +243,9 @@ function dosomething_kudos_get_kudos_for_user_by_file($fid) {
  * @return bool
  */
 function dosomething_kudos_term_is_selected($reportback_item, $term) {
-  // We need to check both array & objects b/c preprocess_node works w/ RB Item objects
-  // and preprocess_user works w/ content arrays.
-  if (is_array($reportback_item) && empty($reportback_item['existing_kids'])) {
-    return false;
-  }
-  else if (!is_array($reportback_item) && !$reportback_item->existing_kids) {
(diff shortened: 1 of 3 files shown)


itsjoekent:
ready for review

itsjoekent:
never mind, always clear cache.

itsjoekent:
ok, i think this is actually ready for review now.

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -245,7 +245,10 @@ function dosomething_kudos_get_kudos_for_user_by_file($fid) {
 function dosomething_kudos_term_is_selected($reportback_item, $term) {
   // We need to check both array & objects b/c preprocess_node works w/ RB Item objects
   // and preprocess_user works w/ content arrays.
-  if ((is_array($reportback_item) && empty($reportback_item['existing_kids'])) && ! $reportback_item->existing_kids) {
+  if (is_array($reportback_item) && empty($reportback_item['existing_kids'])) {
+    return false;
+  }
+  else if (!is_array($reportback_item) && !$reportback_item->existing_kids) {
```
weerd: Just a thought, but maybe we could cast `$reportback_item` to an object here and then we can just avoid the logic of having to check if it's an array it has the item, or if it's a object and it has the property. Then we always just check one thing and keep the logic clean and concise.

``` php
$reportback_item = (object) $reportback_item;

if (! $reportback_item->existing_kids) {
  return false;
}
```

thoughts @joe @angaither ?

itsjoekent: Can we "uncast" it back to an array? otherwise we'd have to rewrite the template to use objects

weerd: This function only returns a boolean, so there wouldn't be a need. It'd just be cast inside the function and then forgotten cause all it sends back is `true/false`.

itsjoekent: OH, this is true. sure yeah this makes sense then

lib/themes/dosomething/paraneue_dosomething/js/campaign/prototype.js
```diff
@@ -54,11 +54,17 @@ $(document).on('click', '.js-kudos-button', function(event) {
...
   $el.attr('disabled','disabled');
+  const dataKid = $el.attr('data-kid');
+  const method = dataKid ? 'DELETE' : 'POST'
 
-  makeKudosRequest($el, $el.attr('data-kid'))
+  makeKudosRequest($el, dataKid)
     .done((response) => {
       const kid = array_get(response, '0.kid', '');
 
+      if (method == 'POST' && !kid) {
```
weerd: `===` trippple equals!

itsjoekent:
@weerd all fixed, profile & campaign kudos still work

weerd:
¡OLÉ!
:sheep: it!

PR #6618, opened 2016-06-21
Adds Kudos to perma link

### LINKED ISSUE #6545: Add kudos on the permalink
Enable kudos on the permalink page
image

itsjoekent:
#### What's this PR do?

Adds Kudos to the perma link

Can you Kudo from the permalink?
Does Kudos still work on campaign template / user profile?
Did the [`Civil Defense Siren`](https://en.wikipedia.org/wiki/Civil_defense_siren) go off? (Note: If this actually happened please follow the instructions given to us earlier during the fire drill & proceed down the stairs)
(I have no idea why I included this in the PR)

not really
#### Relevant tickets

Fixes #6545

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -523,11 +523,13 @@ function dosomething_reportback_view_entity($entity) {
     $fid = array_pop($entity->fids);
   }
 
+
+  // @TODO Can we templatize/DRY some of this? It's fairly similiar code in multiple places.
   $disable_reactions = dosomething_kudos_disable_reactions($entity->nid);
 
   if (! $disable_reactions) {
     $kudos = [];
-    $kudos['id'] = $entity->rbid;
+    $kudos['rbid'] = $entity->rbid;
     $kudos['disable_reactions'] = $disable_reactions;
     $kudos['allowed_reactions'] = [dosomething_kudos_get_term_id_by_name('heart')];
     $kudos['reaction_totals'] = ['heart' => dosomething_kudos_get_total_for_file_by_term_name($entity->rbid, 'heart')];
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -514,12 +514,26 @@ function dosomething_reportback_form_alter(&$form, &$form_state, $form_id) {
 function dosomething_reportback_view_entity($entity) {
   $rb_simple_user = dosomething_user_get_simple_user($rb_user = user_load($entity->uid));
   global $user;
+
   $is_owner = ($user->uid == $rb_simple_user->uid) ? 1 : 0;
   $fid = $_GET['fid'];
+
   if (!$fid) {
     // Get the most recently uploaded item.
     $fid = array_pop($entity->fids);
   }
+
+  $disable_reactions = dosomething_kudos_disable_reactions($entity->nid);
+
+  if (! $disable_reactions) {
+    $kudos = [];
+    $kudos['id'] = $entity->rbid;
+    $kudos['disable_reactions'] = $disable_reactions;
+    $kudos['allowed_reactions'] = [dosomething_kudos_get_term_id_by_name('heart')];
+    $kudos['reaction_totals'] = ['heart' => dosomething_kudos_get_total_for_file_by_term_name($entity->rbid, 'heart')];
+    $kudos['existing_kids'] = dosomething_kudos_get_kudos_for_user_by_file($entity->rbid);
+  }
+
   // Try to get the permalink from cache.
   if ($cached_permalink = cache_get('ds_permalink_' . $entity->rbid . '_' . $fid . '_' . $is_owner, 'cache_dosomething_reportback')) {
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -514,12 +514,26 @@ function dosomething_reportback_form_alter(&$form, &$form_state, $form_id) {
...
   if (!$fid) {
     // Get the most recently uploaded item.
     $fid = array_pop($entity->fids);
   }
+
+  $disable_reactions = dosomething_kudos_disable_reactions($entity->nid);
+
+  if (! $disable_reactions) {
+    $kudos = [];
+    $kudos['id'] = $entity->rbid;
```
weerd: Is it weird that the `$kudos['id']` here is actually the reportback-item id? maybe change the name to be `$kudos['reportback_item_id']` or something that would cause less confusion, since there are ids on Kudos...

weerd:
Mind adding a screenshot to see how this looks on the page when working properly?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -514,12 +514,26 @@ function dosomething_reportback_form_alter(&$form, &$form_state, $form_id) {
...
+
   $is_owner = ($user->uid == $rb_simple_user->uid) ? 1 : 0;
   $fid = $_GET['fid'];
+
   if (!$fid) {
     // Get the most recently uploaded item.
     $fid = array_pop($entity->fids);
   }
+
+  $disable_reactions = dosomething_kudos_disable_reactions($entity->nid);
```
weerd, line 528: Mind adding a `@TODO` for a future refactor run to DRY up some of this code. Seems like there's overlap between what is going on here and in the [preprocess_node.inc](https://github.com/DoSomething/phoenix/blob/280b2937a41c600e001916ba7b6dcd35fadc81a1/lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc#L253) and [helpers.inc](https://github.com/DoSomething/phoenix/blob/280b2937a41c600e001916ba7b6dcd35fadc81a1/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc#L354) etc... could be good to consolidate into a single function that preps that stuff if possible!

itsjoekent, line 528: yeah good idea

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
itsjoekent: good point, I think this could be changed without breaking anything

itsjoekent:
@weerd ah, went to post the mock/screenshot but realized this isnt in the place the mock wanted it... 🎨 time

itsjoekent:
Pushing updates up now, heres what it looks like 

screen shot 2016-06-22 at 1 47 40 pm

angaither:
👍 from me 
@DFurnes do you wanna :eyes: the 🎨 ??

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_counter.scss
```diff
@@ -10,6 +10,18 @@
   }
 }
 
+.card__column {
+  .photo__actions, .caption {
```
DFurnes: We shouldn't be nesting a `.photo__action` element inside a `.card`. Either this should be a new child element of the `.card`, or we should see if `.photo` makes sense to use for the markup for this interface as well.

lib/themes/dosomething/paraneue_dosomething/templates/reportback/reportback-permalink.tpl.php
```diff
@@ -18,6 +18,7 @@
  * - $user->first_name                   : User's first name
  * - $node->fact_problem['fact']
  * - $node->fact_solution['fact']
+ * $Kudos                                : Kudos data
```
DFurnes: This is missing a `-` bullet point, and shouldn't be capitalized.

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_counter.scss
itsjoekent: Whats the rationale for not adding photo actions in cards?

DFurnes: The `.photo__action` is a [BEM element](http://getbem.com/introduction/) which indicates that it's semantically tied to the `.photo` block. If we start using that class elsewhere (without explicitly indicating that it's a standalone component), we make our styles more fragile to change.

itsjoekent: I think should we rename photo__actions to something else then -- right now its being used in photo, figure, and was about to be added to cards here.

itsjoekent: gonna go with `.kudos` & `.kudos__icon`

itsjoekent:
@DFurnes updated da 🎨

DFurnes:
☑️

itsjoekent:
🚂 🚋 🚋

PR #6624, opened 2016-06-22
Store something better than null in competition signup data

### LINKED ISSUE #6617: Competitions row is null/1 and no 0s
#### Current Behavior
- Queried `dosomething_signup` and all values are either `1` or `null` for `dosomething_signup.competition`
#### Desired Behavior
- `dosomething_signup.competition` should return `1` (yes to comp), `0` (no to comp) or `null`
#### Steps to Replicate
- query this: `select
  ds.competition, 
  count(*)
  from
  dosomething.dosomething_signup ds
  group by ds.competition
  `
#### Why This Matters

From an analytics perspective, makes reporting on competitions impossible. Need accurate data that differentiates between yes and no. Blocker for current reporting. 

From an operational perspective, I have no idea how/if this is impacting user experience or gladiator administration

cc @angaither

angaither:
#### What's this PR do?

Moved the competition signup thing to outside of the `response` block.
That is only triggered if the user says yes, which was not saving any `0` values if a person opted out of a competition. 
This allows for either a 0/1 to be stored here if the user interacts with the modal at all.

👀 
#### Relevant tickets

Fixes #6617


sbsmith86:
👍

PR #6634, opened 2016-06-23
Move organ templates

sbsmith86:
#### What's this PR do?

I had different screens of the organ donation modal live in lodash templates, but we don't really need that since most of the markup is fine with being on the page on pageload. This PR moves all the mark up for the modal into the `tpl.php` and then we just use a template for the fields because they are dynamically added to the page after page load. 

You should still see the modal and move through the experience when organ donation is turned on in a node.


lib/themes/dosomething/paraneue_dosomething/js/campaign/OrganDonation.js
```diff
@@ -7,8 +7,6 @@ import Takeover from 'takeover/Takeover';
 
 // @TODO - Move this to the drupal tpl. The only template we need is for the fields
```
DFurnes: Can this `@TODO` be removed now?

sbsmith86: ah yep

lib/themes/dosomething/paraneue_dosomething/js/takeover/Takeover.js
```diff
@@ -5,9 +5,9 @@ const $ = require('jquery');
 const template = require('lodash/string/template');
 
 class Takeover {
-  constructor($takeoverContainer, tpl) {
+  constructor($takeoverContainer) {
     this.$el = $takeoverContainer;
-    this.template = tpl;
+    // this.template = tpl;
```
DFurnes: And this comment? ✏️

DFurnes:
👍

PR #6636, opened 2016-06-24
Moved the node load up above so it checks the campaign type.

### LINKED ISSUE #6635: Disable Signup Email on SMS Games for logged in members
#### Current Behavior
- A member received a transactional email after signing up for context clues. 
- The message they received was our standard signup template, meaning it should have provided three steps to help complete the campaign.
- However, since it is an SMS nudge/game and not a DS campaign, there are no tips. 
- This email is incredibly misleading for two reasons (1)  the "Get Started” button leads them back to the webpage, when the entire experience is meant to be via SMS and (2) we ask them to “snap a pic” which results in the member being completely confused as to what type of campaign it is 
#### Desired Behavior
- They should receive clear instructions on how to complete the SMS Game or no email at all (to prevent them from bouncing between platforms)
- As discussed in Slack, the interim solution is to disable all SMS Game signup emails. 
#### Steps to Replicate
- Sign up for an SMS Game while logged in to your DS account
#### Why This Matters
- The email creates a poor user experience by confusing the member on how to report back. 

screen shot 2016-06-23 at 5 53 54 pm

https://www.dosomething.org/us/campaigns/context-clues

angaither:
#### What's this PR do?

Re-fix the bug that `sms-game` campaign nodes were getting transactional emails.

😍 

This was already an issue #2315 that was fixed [a long time ago](https://github.com/DoSomething/phoenix/commit/0ad46b569a61a4ba4c4f17262bc177ffc62727db) but in refactoring it seems the `node_load` was moved below where it should be. So we were never returning out of this function since `$node` was not defined. 
#### Relevant tickets

Fixes #6635

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module
@@ -510,7 +510,7 @@ function dosomething_signup_views_data() {
 function dosomething_signup_entity_insert($entity, $type) {
   // If not a signup, exit out of function.
   if ($type != 'signup') { return; }
-
+  $node = node_load($entity->nid);
   // If this is a SMS Game Campaign node:
   if (module_exists('dosomething_campaign') && dosomething_campaign_get_campaign_type($node) == 'sms_game') {
     // Third party subscription is handled elsewhere,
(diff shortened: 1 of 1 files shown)


DFurnes:
🐎

PR #6642, opened 2016-06-27
Kudos endpoint new setup

### LINKED ISSUE #6594: Update Kudos API to accept user query to help not kill the API on gallery
Had a quick chat with @DFurnes regarding Kudos and the potential for the API to get overwhelmed...

With the advent of the new Kudos Reactions being activated, right now when viewing more items hits the `/reportback-items` endpoint and nests a [`kudos` object](https://gist.github.com/weerd/1f351d23468c5096b332d5317af1579b) which contains an array of the data. In the data is the term info, and then all the kudos items that were applied to that term along with the user information, etc. 

Lets say Kudos is very popular, and there end up being RB items with like 900 kudos hearts and variations in quantities among the rest. That is A LOT of data passed in a JSON response, when all we really care about is the term and the quantity and if the current user responded with a kudos to a particular RB item in the gallery. So something more to the effect of:

``` json
"kudos": {
  "data": [
    {
      "term": {
        "id": "641",
        "name": "heart",
        "total": 900,
      }
      "user_reacted": true,
    },
    {
      "term": {
        "id": "646",
        "name": "poo",
        "total": 102,
      }
      "user_reacted": false,
    }
  ]
}
```

@DFurnes suggested that maybe in the AJAX request we pass along the user id as a parameter, and then when querying for kudos, it can take into account the user id and add a boolean to the response to indicate whether they already reacted with that kudos term on a RB item. I think this would definitely help reduce the length of the JSON response _significantly_ if kudos proves to be popular.

Thoughts?

@angaither @DFurnes 

__NOTE: Currently I'm working on a interim solution for dealing with the client knowing if current user reacted to a *view more_ RB item added dynamically, by filtering through the list of kudos items in the current response and trying to match the user id. Totally not ideal, but will be much easier if we implement the above suggested approach and I can update this once the above is working._

Checklist:
- [x] Allow passing current user id as a URL parameter.
- [ ] Update the `/kudos` endpoint
- [ ] Update the `/reportback-items` endpoint which includes kudos data

weerd:
#### What's this PR do?

This PR does some initial setup work for upcoming changes to the Kudos endpoint. It allows passing a current user ID that Kudos can later use (next PR) for the new Kudos response which will be more efficient and let you know if just the current user specified has reacted on a ReportbackItem.

I decided to update the names of the methods in the **ReportbackItemTransformer** class to be more obvious regarding what they are doing, now that an `$options` property was added.

`setDatabaseQueryFilters` will set the filters that get used in the database query for retrieving RB items.

`setTransformerOptions` will set any additional options (like the current user) that get passed to the transformer if it needs to take these into account for some additional logic.

It also fixes a random bug w/ the Northstar query and loading user data (not sure if NS response changed or what, but this fixes a DBLog error that was popping up).

This also sets up _one_ universal place to dictate the approved reactions for Kudos (currently only "heart"). But it's easy to retrieve them for JS or PHP via a call to `Kudos::getApprovedReactions()`.

Are kudos borked? ... no? Great!

I'm breaking my work on #6594 into smaller digest-able PRs because @angaither bullies me when I try to submit epic PRs. It's probably warranted though 😬 

These updates were done in a way that should not break any current Kudos functionality. Tested it a bit 😉 
#### Relevant tickets

Refs #6594

---

@angaither @DFurnes

### DIFF
--- a/lib/modules/dosomething/dosomething_kudos/includes/Kudos.php
+++ b/lib/modules/dosomething/dosomething_kudos/includes/Kudos.php
@@ -119,7 +119,7 @@ class Kudos extends Entity {
       $northstar_response = json_decode($northstar_response);
 
       if ($northstar_response && !isset($northstar_response->error)) {
-        $northstar_user = (array) $northstar_response->data;
+        $northstar_user = $northstar_response->data;
       }
     }
 
--- a/lib/modules/dosomething/dosomething_kudos/includes/Kudos.php
+++ b/lib/modules/dosomething/dosomething_kudos/includes/Kudos.php
@@ -4,6 +4,13 @@ class Kudos extends Entity {
 
   protected $entity;
 
+  /**
+   * Approved reactions that are accessible and viewable.
+   * Edit this array to allow for additional kudos reactions.
+   * @var array
+   */
+  protected $approved_reactions = ['heart'];
+
   public $id;
   public $term;
   public $reportback_item;
@@ -86,6 +93,16 @@ class Kudos extends Entity {
   }
 
   /**
+   * Get the approved reactions.
+   * @return array
+   */
+  public static function getApprovedReactions() {
+    $kudos = new static;
+
+    return $kudos->approved_reactions;
+  }
+
+  /**
    * Build out the instantiated Kudos class object with supplied data.
    *
    * @param $data
@@ -102,7 +119,7 @@ class Kudos extends Entity {
       $northstar_response = json_decode($northstar_response);
 
       if ($northstar_response && !isset($northstar_response->error)) {
-        $northstar_user = array_shift($northstar_response->data);
+        $northstar_user = (array) $northstar_response->data;
       }
     }
 
--- a/lib/modules/dosomething/dosomething_kudos/includes/Kudos.php
+++ b/lib/modules/dosomething/dosomething_kudos/includes/Kudos.php
@@ -6,10 +6,11 @@ class Kudos extends Entity {
 
   /**
    * Approved reactions that are accessible and viewable.
-   * Edit this array to allow for additional kudos reactions.
+   * Edit this array to allow for additional kudos reactions across the platform.
+   *
    * @var array
    */
-  protected $approved_reactions = ['heart'];
+  protected static $approved_reactions = ['heart'];
 
   public $id;
   public $term;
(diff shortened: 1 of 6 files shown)


lib/modules/dosomething/dosomething_kudos/includes/Kudos.php
```diff
@@ -86,6 +93,16 @@ public static function find(array $filters = []) {
   }
 
   /**
+   * Get the approved reactions.
+   * @return array
+   */
+  public static function getApprovedReactions() {
+    $kudos = new static;
```
DFurnes: Why not make `$approved_reactions` static so you don't have to make and discard an instance here?

lib/modules/dosomething/dosomething_kudos/includes/Kudos.php
```diff
@@ -102,7 +119,7 @@ private function build($data, $full = false) {
       $northstar_response = json_decode($northstar_response);
 
       if ($northstar_response && !isset($northstar_response->error)) {
-        $northstar_user = array_shift($northstar_response->data);
+        $northstar_user = (array) $northstar_response->data;
```
DFurnes: Would it make sense to use the [`$assoc` parameter of `json_decode`](http://php.net/manual/en/function.json-decode.php) here to avoid having to cast it?

lib/modules/dosomething/dosomething_kudos/includes/Kudos.php
weerd: Oh, true. I wasn't sure if it's normally a best practice to be retrieving property values on a class using getter methods, rather than grabbing a static property directly... Maybe that's not actually a thing? haha.

I can update!

DFurnes:
Two tiny Qs, but looks good to me! 👍

weerd:
@DFurnes for the JSON thingy, turns out we actually wanted an object so it works out fine. Especially since the `dosomething_helpers_isset()` which uses the items in the object can grab stuff whether it's an array or object :) So I just removed the casting.

DFurnes:
Tricks! Sounds good. 👍

lib/modules/dosomething/dosomething_kudos/includes/Kudos.php
DFurnes: Yeah I'm not 100% sure what the bestest of practices is, but I've made the property `protected static` and the getter function `public` in the past & not been too upset with myself.

weerd: Nice, yeah. Just did the same thing :) Tested and works fine.

PR #6644, opened 2016-06-27
Registration validation

### LINKED ISSUE #6632: Validation and Error handing on Organize Registration form
- Add validation to the social security and terms of service field on the organize registration form. 
- Don't allow submission if there are front end errors on the form. 
- display errors if we get validation issues back from organize after submitting.

sbsmith86:
#### What's this PR do?

Added some custom validation to the registration form and updated the flow to handle errors better. 

Some of the validation rules are already defined in our js but I had to [add new ones](https://github.com/DoSomething/phoenix/compare/registration-validation?expand=1#diff-1cb7dc4f0cb778df2c6a67f31b80cca4R5) to be able to message around the validation in a way that make sense for the user at the point they are in this form. 

I also fixed the way I was [checking for different errors](https://github.com/DoSomething/phoenix/commit/e1f8d3f3c6a7c1815fef583dc9c15503fead7651) in the responses we were getting back from Organize that were allowing people to submit forms even when we were getting back server errors, etc.

Also fixed the issue with validation messages not showing and/or breaking the layout because I forgot to add the [right classes to different fields](https://github.com/DoSomething/phoenix/compare/registration-validation?expand=1#diff-27d1575dfbde2abaa023ecae2d5a323aR398)

I think this would best be tested on staging by multiple users to make sure validation is happening sanely. 
#### Relevant tickets

Fixes #6632 
#### Checklist

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/app.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/app.js
@@ -38,6 +38,7 @@ import 'validators/auth';
 import 'validators/reportback';
 import 'validators/address';
 import 'validators/donate';
+import 'validators/organDonation';
 
 import 'campaign/prototype';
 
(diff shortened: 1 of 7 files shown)


DFurnes:
👍

PR #6646, opened 2016-06-28
pre-fill the organ donation form

### LINKED ISSUE #6556: Pre-fill organ forms fields
We have user info for the zip, birthday, email, etc for most users.
When that user goes to fill out the organs form, that stuff should be pre-filled if we have the info

refs #6548

katiecrane:
#### What's this PR do?

For the organ donation form, we are now pre-filling the form with any relevant info we already have. I'm using the zipcode from the first screen since we know for sure that we have that.

If any of the following fields for your account are filled out, they should be being pulled in. Zip Code should always be pulled in from the first screen.
- first name
- last name
- birthdate
- street address line 1
- street address line 2
- city
- state
#### Relevant tickets

Fixes #6556
#### Checklist

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -219,15 +219,17 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
       $materials['content'] .= '<ul><li>Copy and paste your custom link to invite friends and family to become an organ donor: <code>' . $custom_organ_share_link . '</code></li></ul>';
       drupal_add_js(
         array('dosomethingUser' =>
-          array(
+          array('userInfo' =>
+            array(
             'email' => $vars['user']->mail,
             'first_name' => dosomething_user_get_field('field_first_name'),
             'last_name' => dosomething_user_get_field('field_last_name'),
-            'birthday' => dosomething_user_get_field('field_birthdate', NULL, 'm/d/Y'),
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -217,6 +217,21 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
     // if organ donation campaign, add in the share link here
     if ($campaign->variables['register_organ_donors']) {
       $materials['content'] .= '<ul><li>Copy and paste your custom link to invite friends and family to become an organ donor: <code>' . $custom_organ_share_link . '</code></li></ul>';
+      drupal_add_js(
```
sbsmith86, line 220: not a huge deal, but another option is to store all the user stuff in a single object like. 

```
drupal_add_js([
 'dosomethingUser' => [
   'userInfo' => [
    'email' => $email,
     ...
   ]
 ]
]);
```

that way  in the js, you just need to set like 
`this.prefillvalues = setting('dosomethingUser.userInfo');`

 instead of having to grab them all individually and you can just loop through that object when you need the properties.

katiecrane, line 220: a-ha! clever, good call!

katiecrane, line 220: @sbsmith86 that's updated now 😄

sbsmith86:
💯 👍

angaither:
🚢

PR #6649, opened 2016-06-29
Update to response data structure for kudos in RB items.

### LINKED ISSUE #6594: Update Kudos API to accept user query to help not kill the API on gallery

weerd:
#### What's this PR do?

This PR uses some of the work completed in #6642 to update the response when Kudos reaction data is included in the `/reportback-items` response. It makes the response much more useful (and waaaay more compact) for how we actually use the Kudos data (particularly in the campaign galleries), and also takes into account the current user and indicates in the response data whether they reacted on each Reportback Item, removing a lot of the logic we currently have to do on the client-side.

Example of the new response:
image

Hit the `/reporback-items` endpoint and make sure the nested kudos objects in the data for each RB item looks correct.

Since the response data for the reportback items is changing I will be double-checking to make sure nothing gets broken where we use Kudos now.

_NOTE:_ I think there maybe be a better approach in general for how the above is accomplished, but by modifying the query for when RB items are retrieved. Instead of getting all the Kudos IDs for each RB item, gathering totals for each item on the fly. Not sure if this is more efficient or not... someone with better SQL chops would know better 😣
#### Relevant tickets

Fixes #6594 

---

### DIFF
--- a/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
+++ b/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
@@ -205,7 +205,8 @@ function dosomething_kudos_resolve_current_user_activity($current_user, $data, $
 }
 
 /**
- * [dosomething_kudos_compare_current_user_activity description]
+ * Compare current user activity against specified Reportback Item data.
+ *
  * @param  object  $data
  * @param  object  $current_user
  * @return null|array
--- a/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
+++ b/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
@@ -98,15 +98,10 @@ function dosomething_kudos_sort($data, $options = []) {
   // group the kudos based on taxonomy term, along with providing a total
   // count of kudos for each term.
 
-  // @TODO: Not sure if necessary, but setting for now to make sure not to break things.
-  if (!isset($options['display'])) {
-    $options['display'] = 'teaser';
-  }
-
   usort($data, 'dosomething_kudos_sort_by_taxonomy_term');
 
   $data = dosomething_kudos_group_by_taxonomy_term($data, $options);
-  $data = dosomething_kudos_get_totals_by_taxonomy_term($data);
+
   return $data;
 }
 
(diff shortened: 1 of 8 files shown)


weerd:
@deadlybutter @aaronschachter pinging you early on this PR for a quick heads up on the response structure changes in case it affects stuff you might be aware of...

@deadlybutter probs affects the Profile kudos stuff
@aaronschachter mobile app?

itsjoekent:
@weerd sorry I didn't touch much of the JS work around this, mostly just 🎨 so not sure what will/could break. I'm assuming you've tested giving a kudos on the user profile + gallery and it works ok?

weerd:
@deadlybutter testing that out now. Updating the JS cause it definitely affects it. Just wanted to give a heads up and ask to make sure I'm aware of everything to test, etc :) I'll keep ya posted!

aaronschachter:
We're not using Kudos in iOS at all... I know Android doesn't display them, but there is for sure [non-user-facing code the repo that references of them](https://github.com/DoSomething/LetsDoThis-Android/search?utf8=%E2%9C%93&q=kudos).

@lkpttn -- It'd be quite nice to run a test build of the Android app on staging or thor before we to prod to sanity check whether this could break anything in the production Android app..

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -131,67 +126,120 @@ function dosomething_kudos_sort_by_taxonomy_term($a, $b) {
   return $result;
 }
 
+/**
+ * Get current user from supplied data array or global user.
+ *
+ * @param  array  $data
+ * @return null|string
+ */
+function dosomething_kudos_get_current_user($data = []) {
```
angaither: this is a super 🐟 y function

itsjoekent: 🎣 (I've always wanted to use this emoji)

weerd: Yeah, not super keen on it, but added it to take into account if a user is passed via a URL param (which it now supports) versus using the current global user. Chatted w/ @DFurnes about this and we both kind of agreed it might just be best to remove the newly added `?current_user=120193` url param on the `/reportback-items` endpoint and just use the `global $user`. Eventually as we integrate the new OAuth2 stuff FURRRRRRRNS has worked on, we can add some logic to if no `global $user` to check against a `JSON Web Token` or whatever the OAuth-y approach is :)

angaither: hrm two things
1. this seems more like a global or helper function
2. it can/probably should be done in way less checks

weerd: Yeah, if I remove it as a URL param that gets passed in and along to numerous functions (I'm passing an `options` array around that had to contain the user id that could be passed via `/reportback-items?current_user=1219129`), then I can just do away with all this to begin with and rely on solely the `global $user` which is way simpler :)

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
```diff
@@ -281,15 +281,16 @@ function paraneue_dosomething_preprocess_reportback_gallery(&$vars, $count = 6)
   drupal_add_js(
     [
       'dsKudosReactions' => [
+        'approved' => Kudos::getApprovedReactions(),
```
weerd, line 284: Now we _only_ need to define the `approved` Kudos in the Kudos class property and that's it. One property defines the full approved listing, and it's retrieved via `Kudos::getApprovedReactions()`!

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -131,67 +126,105 @@ function dosomething_kudos_sort_by_taxonomy_term($a, $b) {
...
-    if ($options['display'] === 'full') {
-      $output['reportback_item'] = $item->reportback_item;
+    if (!isset($output[$item->term['id']]['term']['total'])) {
+      $output[$item->term['id']]['term']['total'] = 0;
     }
 
-    if (isset($item->uri)) {
-      $output['uri'] = $item->uri;
-    }
+    $output[$item->term['id']]['term']['total'] += 1;
```
DFurnes, line 161: Does this still have to query all of the individual Kudos and then iterate over each one to get the count? I feel like we might be able to do this more efficiently with a SQL `COUNT()` query.

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
DFurnes, line 284: 👏

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
```diff
@@ -281,15 +281,16 @@ function paraneue_dosomething_preprocess_reportback_gallery(&$vars, $count = 6)
...
         'enabled' => !$disable_reactions,
-        'user' => isset($user->uid) ? $user->uid : null,
         'terms' => [
           // @TODO: Ideally this would call a function that retrieves all
           // kudos reaction terms and sets up the associative array, but
           // for the time being we are only using a single term, "heart",
           // so let's keep it simple ;)
           'heart' => dosomething_kudos_get_term_id_by_name('heart'),
         ],
+        'user' => isset($user->uid) ? $user->uid : null,
```
DFurnes, line 293: We probably don't need a `user` value specifically scoped under `dsKudosReactions`. Would it be possible to use `drupal-user-id` meta tag, or maybe set a root-level `uid` property that everything can share?

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
weerd, line 161: See my _NOTE_ in **Any background context you want to provide?** I agree, but there's more shit to be done there and some caveats regarding then having still query against each item to see if the current user reacted, or grab the collection of current user reactions and compare those when each RB item is processed. There's more to consider and didn't want to break everything. This provides _a_ solution for right now, but ultimately I'd like to go down the SQL counting route.

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
weerd, line 293: That was there from before... I just moved it down so it's in alphabetical order. Hehe. But yeah, I was thinking root-level current user id would be ideal... but maybe something for a different PR.

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
DFurnes, line 161: Alright, makes sense. Definitely still a good performance win getting this merged in. :)

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
DFurnes, line 293: Kk, sounds fine to me.

DFurnes:
👍

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
weerd, line 161: Defs agree with you though. The aim for this was to ultimately help with efficiency in what gets sent back in a JSON response so the kudos object isn't crazy long with like 900 things. SQL query performance would be a next step :)

aaronschachter:
@weerd Can you provide example of what the current response to get a better idea of the upcoming changes?

weerd:
Oh shiza. Forgot to update the docs. The example is in the screenshot in the description, but I will update the docs right now!

angaither:
BUT IT'S A CHECKBOX IN THE TEMPLATE @weerd!

weerd:
I know. I accidentally deleted it...

image

aaronschachter:
I meant what the response looks like before this was merged. Screenshot from prod:
screen shot 2016-06-30 at 10 40 41 am

weerd:
#### Checklist
- [x] Documentation added for new features/changed endpoints. 😬

PR #6656, opened 2016-06-29
Organ Registration Form Bugs

### LINKED ISSUE #6655: Organ form submission issues.
#### Current Behavior

1) As an Anonymous user, we can't submit the zip code form.

2) The registration form seems to be hanging sometimes, only showing a loading gif instead of taking the user to the share declaration page.  There seems to be an issue where the form is submitting twice which **could** be messing things up. Also the above permission issue could have also been causing issues. 
#### Desired Behavior

1) If I am signed up as an anonymous user, I should be able to submit the zip code form and see the registration form for my zip code.

2) I should be able to submit the registration form successfully without it hanging. 
#### Steps to Replicate

1) Create and account or log in as a non-admin. Try to submit the zip code form. 

2) I could not replicate this issue consistently. My one hunch is that it is related to the permissions issue where our api wrapper endpoints are not allowing access. The other is that the form is submitting twice and that is messing up submission.

sbsmith86:
#### What's this PR do?

1) Changes access to the api wrapper endpoints so that logged in users can use them. The `user_access` permission was not allowing non-admins to submit forms.

2) Updates how I was handling submission of the registration form so it worked with our validation package better. The validation package [binds a submit handler to all forms](https://github.com/DoSomething/validation/blob/master/src/validation.js#L264) and that was conflict with our more specific handler causing use to submit info to organize twice. 

The update I made is a bit of a hack that just stops the body form handler from refreshing the page when it is done, but still runs the validation. And then, I subscribe to the successful validation event that that validation package throws to know when to actually submit the info to Organize. This seemed like the simplest solution. But i'm open to other suggestions, i'm sure there is some fancy propagation stuff we could do. 

Overall, I think a combo of both of these things will fix an issue we were seeing with the form hanging sometimes.
#### Relevant tickets

Fixes #6655

### DIFF
--- a/lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
+++ b/lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
@@ -24,14 +24,14 @@ function dosomething_organ_donation_menu() {
 
   $items['organize/registration'] = [
     'page callback' => 'dosomething_organ_donation_post_registration',
-    'access callback' => 'user_access',
+    'access callback' => 'user_is_logged_in',
     'access arguments' => array('administer modules'),
     'type' => MENU_CALLBACK,
   ];
 
   $items['organize/postal-code'] = [
     'page callback' => 'dosomething_organ_donation_send_postal_code',
-    'access callback' => 'user_access',
+    'access callback' => 'user_is_logged_in',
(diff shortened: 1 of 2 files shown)


angaither:
👍  from a 🐘 's perspective

DFurnes:
Seems like a good solution to me. 👍 🚢

PR #6658, opened 2016-06-30
store info from organ registration form on the user

### LINKED ISSUE #6557: Store user fields after a user submits organ reg form
We should be able to fill out on a user once they submit the organ donation form
- [ ] `field_birthdate`
- [ ] `field_mobile`
- [ ] `field_address`
- [ ] `field_last_name`

katiecrane:
#### What's this PR do?

Updates user name, address, and birthday with the information from the organ donation registration form.

Register to give us your organs and change some of your data. Then check to make sure it updated.
#### Relevant tickets

Fixes #6557

### DIFF
--- a/lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
+++ b/lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
@@ -90,30 +90,7 @@ function dosomething_organ_donation_post_registration() {
 
   $response = drupal_http_request($client['base_url'] . '/v2/registrations/', $options);
 
-  // Load the user so we can make updates
-  global $user;
-
-  // Update address (there doesn't seem to be a way to update the whole address using dosomething_user_set_fields)
-  $new_field_address = [
-    'administrative_area' => $params['state'],
-    'locality' => $params['city'],
-    'postal_code' => $params['postal_code'],
-    'thoroughfare' => $params['street_address'],
-    'premise' => $params['street_address_2'],
-  ];
-  $user->field_address[LANGUAGE_NONE][0] = $new_field_address;
-
-  // Update name and birthday
-  $new_fields = [
-    'first_name' => $params['first_name'],
-    'last_name' => $params['last_name'],
-    'birthdate' => date('Y-m-d', strtotime($params['birthdate'])),
-  ];
-  $formatted_fields = [];
-  dosomething_user_set_fields($formatted_fields, $new_fields);
-
-  // Save the updates
-  user_save($user, $formatted_fields);
+  dosomething_organ_donation_update_user($params);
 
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
```diff
@@ -90,6 +90,32 @@ function dosomething_organ_donation_post_registration() {
 
   $response = drupal_http_request($client['base_url'] . '/v2/registrations/', $options);
 
+  // Load the user so we can make updates
+  global $user;
+  $account = user_load($user->uid);
```
angaither: `global $user` should be the full loaded object, you shouldn't need to re-load here

lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
```diff
@@ -90,6 +90,32 @@ function dosomething_organ_donation_post_registration() {
...
+  $account->field_address[LANGUAGE_NONE][0] = $new_field_address;
+
+  // Update name and birthday
+  $new_fields = [
+    'first_name' => $params['first_name'],
+    'last_name' => $params['last_name'],
+    'birthdate' => date('Y-m-d', strtotime($params['birthdate'])),
+  ];
+  $formatted_fields = [];
+  dosomething_user_set_fields($formatted_fields, $new_fields);
```
angaither: this function works if you pass an empty array?

lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
```diff
@@ -90,6 +90,32 @@ function dosomething_organ_donation_post_registration() {
 
   $response = drupal_http_request($client['base_url'] . '/v2/registrations/', $options);
 
+  // Load the user so we can make updates
```
angaither: we should break this whole chunk out into it's own function

lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
katiecrane: aha! you're right, something else must have fixed the problem that caused me to add this

lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
katiecrane: @angaither it sure does. I got it to work but I'm still confused by that function

angaither: understandable

angaither:
lookin 🚢 able

PR #6669, opened 2016-06-30
add magic link copy to custom campaign settings and api response

### LINKED ISSUE #6641: Create and expose 'view on web' link
#### User Story

As a user on the mobile app, I would like the option to view the web content only on the web without re-logging in. 
#### Additional Information
- Create a custom setting to hold `magic_link_copy` 
- Expose the the copy in the API

included in the `/campaigns` api response

```
magic_link_copy: click here to fill out the form on the web
```
#### Why This Matters

Some experiences (i.e. organ donation, filling out address forms) is only available on the web, instead of building this experience out for mobile, let's just forward the user to be authenticated to that campaign in a browser.

katiecrane:
#### What's this PR do?

Added a new type of custom setting on campaigns. Include the `magic_link_copy` in the api response to get a campaign. 

Add some magic link copy and make sure you get it when you grab a campaign via the api. 
#### Relevant tickets

Fixes #6641 
#### Checklist

cc: @aaronschachter & @DFurnes

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -266,6 +266,9 @@ abstract class Transformer {
         $output['uri'] = $data->uri;
       }
 
+      if (dosomething_helpers_get_variable('node', $data->id, 'magic_link_copy')) {
+        $output['magic_link_copy'] = dosomething_helpers_get_variable('node', $data->id, 'magic_link_copy');
+      }
     }
 
     return $output;
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -266,8 +266,9 @@ abstract class Transformer {
         $output['uri'] = $data->uri;
       }
 
-      if (dosomething_helpers_get_variable('node', $data->id, 'magic_link_copy')) {
-        $output['magic_link_copy'] = dosomething_helpers_get_variable('node', $data->id, 'magic_link_copy');
+      $magic_link_copy = dosomething_helpers_get_variable('node', $data->id, 'magic_link_copy');
+      if ($magic_link_copy) {
+        $output['magic_link_copy'] = $magic_link_copy;
       }
     }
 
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_helpers/dosomething_helpers.variable.inc
```diff
@@ -154,6 +154,21 @@ function dosomething_helpers_variable_form($form, &$form_state, $node) {
...
+    $form['magic_link'] = [
+      '#type' => 'fieldset',
+      '#title' => t('Magic Link'),
+      '#collapsible' => TRUE,
+      '#collapsed' => TRUE,
+    ];
+    $form['magic_link']['magic_link_copy'] = [
+      '#type' => 'textfield',
+      '#title' => t('Magic Link Copy'),
+      '#description' => t('Custom text to explain the magic link.'),
```
angaither: maybe a lil helper info about what the magic link is 🔮

lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -266,6 +266,9 @@ protected function transformCampaign($data) {
         $output['uri'] = $data->uri;
       }
 
+      if (dosomething_helpers_get_variable('node', $data->id, 'magic_link_copy')) {
```
angaither: probably need an `isset` here to avoid warnings? then you don't have to re-get the variable on the line below as well

katiecrane: @angaither In the new commit I updated it to grab the variable first, and then it's either `FALSE` or the value. Does that work?

angaither: @katiecrane the value in the API is `FALSE`? 
the api response should def be `NULL` but maybe I'm not understanding what what is the 'it' in this case?

katiecrane: @angaither `dosomething_helpers_get_variable('node', $data->id, 'magic_link_copy')` is either `FALSE` or the `magic_link_copy` text. So in the new commit it either returns the `magic_link_copy` value or leaves out that attribute. Lemme know if I used too many pronouns again haha

angaither: ahh! lovely!

angaither:
🚢

PR #6674, opened 2016-07-05
add help text to campaign run timing

### LINKED ISSUE #6659: Change "Show End Date" on Campaign Run to "Choose End Date"
#### User Story

On the campaign run, the checkbox to select an end date says "show end date" -- this is misleading because:
1. there is a show end date checkbox on the campaign node, which makes the end date appear in the header photo, and this checkbox serves a different purpose.
2. if someone clicks this checkbox on the campaign run, it shows the same date as the start date. this appears as though the end date has actually been set, but Drupal is really just populating the field.

We should:
1. change the text from "show end date" to "select end date"
2.  if possible, the end date field should not autopopulate with the start date listed on the campaign run. (if this is not possible, we'd like to amend the help text as cited below -- please let us know if that's the case!)
3. Add help text to the Timing part of the form

> These dates determine whether the campaign is open or closed. If the run has no end date selected, it will appear open. If the campaign does have an end date, the campaign will automatically switch to the closed state at 12:05 AM the day after the end date. If the campaign is translated and open across different translations, then the start and end dates must be correctly set for _every_ translation of the run."

https://www.thor.dosomething.org/us/node/6810/edit/en
image
#### Why This Matters

It's confusing to have to set the end dates on both the run and the campaign, but this should help clear up that distinction.

cc @ngjo

katiecrane:
#### What's this PR do?

Added a description to the timing form on campaign runs. Looks like this now:
image

The text is what @namimody [provided in the issue](https://github.com/DoSomething/phoenix/issues/6659#issuecomment-230494722).

Go to edit a campaign run and make sure the text shows up in the timing box.

This fixes point 3 of the issue. It sounds like 1 can't be done and 2 is a problem with the Drupal Date Module.
#### Relevant tickets

Refs #6659

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.field_group.inc
+++ b/lib/modules/dosomething/dosomething_campaign_run/dosomething_campaign_run.field_group.inc
@@ -88,7 +88,9 @@ function dosomething_campaign_run_field_group_info() {
     'format_settings' => array(
       'formatter' => 'collapsible',
       'instance_settings' => array(
-        'description' => '',
+        'description' => 'These dates determine whether the campaign is open or closed, and group sign up and reportback activity. If the run has no end date selected, the campaign will stay active. If the campaign run does have an end date, the campaign will automatically switch to the closed state at 12:05 AM the day after the end date. </br></br>
+
+          To select an end date, click "Show End Date" and use the dropdown menu to select a date. Be sure the date is correct, and not simply the date that autofills in the field. If the campaign is translated and open across different translations, then the start and end dates must be correctly set for every translation of the run.',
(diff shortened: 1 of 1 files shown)


angaither:
@katiecrane I think we may be able to edit the text of the "show end date" checkbox. 
and maybe do some digging on the third autofill issue. 
but if you use `refs #` instead of `fixes` you can merge this PR and continue looking into the other things

katiecrane:
@angaither okay cool, I did some light digging into those but wasn't sure if we wanted to full dig. I'll get out the tunneling gear.

angaither:
👍 

kewl, but you can merge this while digging 🔨

PR #6677, opened 2016-07-05
Update so "EVERYBODY gets a kudos!" --Oprah

### LINKED ISSUE #6661: Reviewing reportbacks applies a kudo
#### User Story

As an admin when I `approve`, `promote` or `exclude` a reportback item, that item gets automatically kudo'ed

weerd:
#### What's this PR do?

This PR updates the Admin review process, so if an admin marks a Reportback Item as "approved" or "promoted" it automagically ✨ gets a kudos created for that item by the admin.

Try reviewing a few Reportback Items and then check the gallery to make sure they have 1 kudos already (or check the database for that `fid` and confirm a new record was created)!

image
#### Relevant tickets

Fixes #6661

---

@angaither @DFurnes

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
@@ -249,7 +249,8 @@ function dosomething_reportback_files_form_submit($form, &$form_state) {
     $kudos = [
       'fid' => $fid,
       'uid' => $user->uid,
-      'tid' => dosomething_kudos_get_term_id_by_name(array_pop(Kudos::getApprovedReactions())),
+      // @TODO: use array_pop(Kudos::getApprovedReactions()) once new Kudos refactor is fully approved.
+      'tid' => dosomething_kudos_get_term_id_by_name('heart'),
       'created' => time(),
     ];
 
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
@@ -238,12 +238,20 @@ function dosomething_reportback_files_form($form, &$form_state, $status = NULL, 
  * Form submit callback for dosomething_reportback_files_form_submit.
  */
 function dosomething_reportback_files_form_submit($form, &$form_state) {
+  $user = dosomething_helpers_get_current_user();
+
   $rb_files = $form_state['values']['rb_files'];
   if (empty($rb_files) || !is_array($rb_files)) {
     return;
   }
 
   foreach ($rb_files as $fid => $values) {
+    $kudos = [
+      'fid' => $fid,
+      'uid' => $user->uid,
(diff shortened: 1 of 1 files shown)


DFurnes:
👏 👏 👏 🐝

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
```diff
@@ -261,6 +269,11 @@ function dosomething_reportback_files_form_submit($form, &$form_state) {
       form_set_error(t("An error has occurred."));
     }
 
+    // Save kudos reaction from admin on RB item.
+    if ($values['status'] === 'promoted' || $values['status'] === 'approved') {
```
angaither: we want `excluded` in here as well

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
```diff
@@ -238,12 +238,20 @@ function dosomething_reportback_files_form($form, &$form_state, $status = NULL,
...
   $rb_files = $form_state['values']['rb_files'];
   if (empty($rb_files) || !is_array($rb_files)) {
     return;
   }
 
   foreach ($rb_files as $fid => $values) {
+    $kudos = [
+      'fid' => $fid,
+      'uid' => $user->uid,
+      'tid' => dosomething_kudos_get_term_id_by_name(array_pop(Kudos::getApprovedReactions())),
```
angaither: in theory... if we have multiple kudo types, this will just grab the first thing in the array..which might not always be a heart. maybe add a note for that in the function? or here?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
weerd: ah jes. ok!

weerd:
I made an update to how it sets the Kudos term to use, since we might revert the code from last week. added a TODO to set it back if it doesn't get reverted.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
```diff
@@ -249,7 +249,8 @@ function dosomething_reportback_files_form_submit($form, &$form_state) {
     $kudos = [
       'fid' => $fid,
       'uid' => $user->uid,
-      'tid' => dosomething_kudos_get_term_id_by_name(array_pop(Kudos::getApprovedReactions())),
+      // @TODO: use array_pop(Kudos::getApprovedReactions()) once new Kudos refactor is fully approved.
+      'tid' => dosomething_kudos_get_term_id_by_name('heart'),
```
angaither, line 253: I still feel like this is more secure, even if the kudos refactor is merged in. Since we will always want a `heart`, and not necessarily the first thing in an array

angaither:
yas yas 🚢 🐝 🐝 🐝 🐝

weerd:
hahahah. that glare 👀

PR #6678, opened 2016-07-05
in some environments register_organ_donors isn't in the campaign variables

katiecrane:
#### What's this PR do?

In some environments the `register_organ_donors` variable does not exist, so now we make sure it's in the array before pulling the value.

If you were seeing this error before on global campaigns, make sure this gets rid of it.

@deadlybutter and @weerd were seeing this. I was unable to recreate locally but @deadlybutter wasn't seeing the error anymore with this line changed.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -215,7 +215,7 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
     $materials['title'] = t('Stuff You Need');
     $materials['content'] = $campaign->items_needed;
     // if organ donation campaign, add in the share link here
-    if ($campaign->variables['register_organ_donors']) {
+    if (array_key_exists('register_organ_donors', $campaign->variables) && $campaign->variables['register_organ_donors']) {
       $materials['content'] .= '<ul><li>Copy and paste your custom link to invite friends and family to become an organ donor: <code>' . $custom_organ_share_link . '</code></li></ul>';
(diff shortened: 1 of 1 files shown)


angaither:
🚢

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -215,7 +215,7 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
     $materials['title'] = t('Stuff You Need');
     $materials['content'] = $campaign->items_needed;
     // if organ donation campaign, add in the share link here
-    if ($campaign->variables['register_organ_donors']) {
+    if (array_key_exists('register_organ_donors', $campaign->variables) && $campaign->variables['register_organ_donors']) {
```
weerd, line 218: would a simple `isset($campaign->variables['register_organ_donors'])` work just as well?

katiecrane, line 218: I thiiink so but @deadlybutter had specifically tested with this and since I wasn't seeing the error I felt better about using `array_key_exists`. Either way we'd need both conditions though

itsjoekent, line 218: in this case isset by itself should also work fine, i could try it out if you want

weerd, line 218: could just be a later update. Just feels cleaner and easier to grok w/ a simple `isset()`. Only reason I mentioned it 😄

PR #6679, opened 2016-07-05
Adding Symfony Var-Dumper package via composer.

weerd:
#### What's this PR do?

This PR includes the Symfony Var-Dumper package to make `var_dump()` results much easer to understand and see all top level items in one easily viewable, collapsed collection. It is added as a dev dependency via composer. 

Instead of using `var_dump($variable)` to see something like this that you have to scroll forever on:
screen shot 2016-07-05 at 5 21 46 pm

 You can now simply use `dump($variable)` to get this:
screen shot 2016-07-05 at 5 22 15 pm

Try it out!

Not super keen on the colors, but we could maybe do some editing to it kinda like Laravel does, when it uses the `dd()` function.
#### Relevant tickets

Fixes 💅 

---

@angaither @DFurnes

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,5 +14,8 @@
   "require": {
     "stripe/stripe-php": "^1.18.0",
     "zendesk/zendesk_api_client_php": "^1.2.0"
+  },
+  "require-dev": {
+    "symfony/var-dumper": "^3.1"
   }
 }
--- a/composer.lock
+++ b/composer.lock
@@ -1,10 +1,11 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "eea86c837efda216671b620e473afb0d",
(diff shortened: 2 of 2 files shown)


DFurnes:
Ahhh yesssss! ✨ This is great.

> Not super keen on the colors, but we could maybe do some editing to it kinda like Laravel does

It looks like it wouldn't be too hard to make a custom dumper with [Laravel's styles](https://github.com/laravel/framework/blob/5.2/src/Illuminate/Support/Debug/HtmlDumper.php).

weerd:
Yep yep! I was checking that out. Next step was to maybe create a `dosomething_debug.module` with classes etc in there that behave similar to how Laravy does things etc.

weerd:
I'm aiming to get a `dd()` into Droooopal.

weerd:
Or a `dsd()` since @DFurnes mentioned that Devel has a `dd()` function 😱  (that is different than what we're used to the Laravel one doing).

angaither:
I'm not fussed, but why? `dpm` doesn't work for you?

weerd:
`dpm()` works, but there are times when I need to dump stuff out outside of `dpm()` or in a particular spot (whereas dpm waits to output into the `messages` array or whatever). Plus, I'd like to try and get this working for when I'm dumping stuff out for API development. Right now I'm forced to use `print_r()` for all api output while working on stuff.

angaither:
`print_r()` is the only way to develop apis.

Issue #6647, opened 2016-06-29
Store data in dosomething_organ_donation table


sbsmith86:
#### User Story

As admins, we would like to store data about how the user is using the registration form in the `dosomething_organ_donation` table.

Store that the user has completed the registration experience when they click the `Done` button
#### Why This Matters

We need this for tracking user registrations on our end.

PR #6684, opened 2016-07-06
store organ info, make modal disappear when you click done

katiecrane:
#### What's this PR do?
- We keep the `uid`, `sid`, and `timestamp` of the registration in `dosomething_organ_donation` table if the user get through the whole flow. 
- When you click `Done` the modal should disappear now.
- I also threw in checking if we collected the `street_address_2` from the user since (a) it seems like we are not being fed that field anymore and (b) I *think* it should be optional anyway.

Register to give your organs and make sure the data shows up in the table.
#### Relevant tickets

Refs #6647

### DIFF
--- a/lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
+++ b/lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
@@ -250,7 +250,7 @@ function dosomething_organ_donation_update_user($params) {
     'thoroughfare' => $params['street_address'],
   ];
   // If this field is present I think it would be optional, so check if we have it first
-  if(array_key_exists('street_address_2', $params)) {
+  if (array_key_exists('street_address_2', $params)) {
     $new_field_address['premise'] = $params['street_address_2'];
   }
   $user->field_address[LANGUAGE_NONE][0] = $new_field_address;
--- a/lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
+++ b/lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
@@ -130,15 +130,12 @@ function dosomething_organ_donation_store_registration() {
   // Grab this to get the sid
   $params = drupal_get_query_parameters();
 
-  // Grab this to get the uid
-  global $user;
-
   try {
-    if ($params['sid'] && $user->uid) {
+    if ($params['sid'] && $params['uid']) {
       db_insert('dosomething_organ_donation')
         ->fields([
           'sid' => $params['sid'],
-          'uid' => $user->uid,
+          'uid' => $params['uid'],
           'registration_form_timestamp' =>  REQUEST_TIME,
       ])->execute();
 
--- a/lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
+++ b/lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
@@ -127,14 +127,18 @@ function dosomething_organ_donation_send_postal_code() {
  *
  */
 function dosomething_organ_donation_store_registration() {
+  // Grab this to get the sid
   $params = drupal_get_query_parameters();
 
+  // Grab this to get the uid
+  global $user;
+
   try {
-    if ($params['sid'] && $params['uid']) {
+    if ($params['sid'] && $user->uid) {
       db_insert('dosomething_organ_donation')
         ->fields([
           'sid' => $params['sid'],
-          'uid' => $params['uid'],
+          'uid' => $user->uid,
           'registration_form_timestamp' =>  REQUEST_TIME,
       ])->execute();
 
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
```diff
@@ -244,8 +248,11 @@ function dosomething_organ_donation_update_user($params) {
     'locality' => $params['city'],
     'postal_code' => $params['postal_code'],
     'thoroughfare' => $params['street_address'],
-    'premise' => $params['street_address_2'],
   ];
+  // If this field is present I think it would be optional, so check if we have it first
+  if(array_key_exists('street_address_2', $params)) {
```
sbsmith86: 👾

lib/themes/dosomething/paraneue_dosomething/js/campaign/OrganDonation.js
```diff
@@ -16,7 +16,9 @@ class OrganDonation extends Takeover {
 
     this.postalCodeUrl       = `${this.baseUrl}/organize/postal-code`;
     this.registrationUrl     = `${this.baseUrl}/organize/registration`;
+    this.shareUrl            = `${this.baseUrl}/organ-donation/registration`;
```
sbsmith86: I feel like `shareUrl` is a little confusing here since it feels like that would relate to social sharing...maybe like `storeRegistrationUrl` or something more explicit?

lib/themes/dosomething/paraneue_dosomething/js/campaign/OrganDonation.js
```diff
@@ -137,6 +139,13 @@ class OrganDonation extends Takeover {
     this.loadScreen(this.$shareScreen);
 
     // @TODO - Handle submission of "done" button that posts to our db info about the submission being done.
+    // This happens whether they hit "done" or "skip" since at this point those are the same
```
sbsmith86, line 142: I'm not sure these are the same....would we want to not show the modal anymore if they chose to skip it? I mean we could cause it would be annoying, but they could just skip out of the form for other reasons and we would want to show the modal experience until they actually go through the flow. Could be a product question.

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -433,7 +433,7 @@
                   <p>Copy and paste your share link:</p>
                   <code><?php print $custom_organ_share_link ?></code>
                   <div class="form-item -padded submit-done-container">
-                    <input type="submit" class="button submit-done" value="Done" />
+                    <input type="submit" class="button js-close-modal" value="Done" />
```
sbsmith86, line 436: Again, we might want to add some clarity about whether we want to log skipping the form, and getting through the flow and clicking done, differently.

lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
```diff
@@ -127,14 +127,18 @@ function dosomething_organ_donation_send_postal_code() {
...
 function dosomething_organ_donation_store_registration() {
+  // Grab this to get the sid
   $params = drupal_get_query_parameters();
 
+  // Grab this to get the uid
+  global $user;
+
   try {
-    if ($params['sid'] && $params['uid']) {
+    if ($params['sid'] && $user->uid) {
```
angaither: this should be in the params still, right?

lib/themes/dosomething/paraneue_dosomething/js/campaign/OrganDonation.js
katiecrane, line 142: Ah sorry @sbsmith86  this is kind of unclear. This is referring specifically to the last page of the modal. It only goes away forever if they hit `Done` or `skip` on the last page.

sbsmith86:
Looks good to me 👍

lib/modules/dosomething/dosomething_organ_donation/dosomething_organ_donation.module
katiecrane: @angaither as far as I can tell we don't have it in `Drupal.settings` so I don't thiiiink we have access to throw it in as a param?

sbsmith86: it's actually on a meta tag on the page, we have grabbed it from there in the past. I think it should actually be in the Drupal.settings object...if you want to make that update.

sbsmith86: It's added as a meta tag https://github.com/DoSomething/phoenix/blob/e5be0dd592c52264f7465971c3c56591ed507aea/lib/themes/dosomething/paraneue_dosomething/templates/system/html.tpl.php#L21

katiecrane: @sbsmith86 yeah that seems like a good idea. Should that get added in [here](https://github.com/DoSomething/phoenix/blob/df9950e89038a155ae365ecdc69d5a6fa0aab3fc/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc#L220-L236) as well or somewhere else?

sbsmith86: @katiecrane that seems like a good place! Love having one big user object.

katiecrane: @sbsmith86 I added that into this PR, can you 👀 with the latest commit? thank you!

sbsmith86:
Nice!

sbsmith86:
:shipit:

PR #6689, opened 2016-07-07
Fixes kudos on profile/perma link + bender refactoring

### LINKED ISSUE #6685: Discrepancy in # of Kudos
#### Current Behavior
1. Likes on action page gallery are not matching likes on permalink page or profile
2. Likes on permalink page and profile are being stored incorrectly (i.e. liking something on the permalink page is not adding to the overall kudos count)
#### Desired Behavior
1. If a photo has 9 likes it should have 9 likes everywhere we display kudos.
2. If I like a photo on the permalink or profile I want it to be liked everywhere
#### Steps to Replicate
- Look at a photo in the gallery. 
- Check to see if the # of kudos matches on the permalink for that photo or the profile of the user.
- Confirm that it does not.
#### Why This Matters
- The data should be the same across the site. It's confusing for users to see different # of likes in different places.
- Profile: https://www.dosomething.org/us/user/2956739

image
- Permalink: https://www.dosomething.org/reportback/97544

image
- Gallery: 
  https://www.dosomething.org/us/campaigns/thumb-wars

image

cc: @jessleenyc

itsjoekent:
#### What's this PR do?

Refactors backend Kudo usage
In doing so fixed some usage bugs on the user profile & perma link 

I've tested off the top of my head
- Kudo's (Kudo, unkudo) work on campaign template, user profile, reportback perma link
- verify the state reflects properly across all of ^

1- Code was duplicated all over the place, hence bugs & bad maintenance problems (my bad)
2- Was unclear when I originally wrote this what ID was so I used rbid instead of the fid. All is good now.
#### Relevant tickets

Fixes #6685

### DIFF
--- a/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
+++ b/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
@@ -313,7 +313,7 @@ function dosomething_kudos_disable_reactions($nid) {
 function dosomething_kudos_get_kudos_data_for_fid($fid, $nid) {
   $disable_reactions = dosomething_kudos_disable_reactions($nid);
   if ($disable_reactions) {
-    return;
+    return NULL;
   }
 
   $kudos = [];
@@ -324,6 +324,7 @@ function dosomething_kudos_get_kudos_data_for_fid($fid, $nid) {
   $kudos['existing_kids'] = dosomething_kudos_get_kudos_for_user_by_file($fid);
   $kudos['reaction_totals'] = ['heart' => dosomething_kudos_get_total_for_file_by_term_name($fid, 'heart')];
   $kudos['allowed_reactions'] = [dosomething_kudos_get_term_id_by_name('heart')];
+  $kudos['disable_reactions'] = $disable_reactions;
 
   return $kudos;
 }
--- a/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
+++ b/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
@@ -315,7 +315,11 @@ function dosomething_kudos_disable_reactions($nid) {
  * @param  int $fid Reportback file id
  * @return array Values needed to render Kudos markup
  */
-function dosomething_kudos_get_kudos_data_for_fid($fid) {
+function dosomething_kudos_get_kudos_data_for_fid($fid = NULL) {
+  if (! $fid) {
+    return [];
+  }
+  
   $kudos = [];
   $kudos['fid'] = $fid;
   $kudos['reaction_totals'] = [
--- a/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
+++ b/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
@@ -310,6 +310,12 @@ function dosomething_kudos_disable_reactions($nid) {
   return !variable_get('dosomething_reportback_kudos_enabled', FALSE) || dosomething_helpers_get_variable('node', $nid, 'disable_reactions');
 }
 
+/**
+ * Grab all Kudos data required to render the markup for a given reportback fid
+ * @param  int $fid Reportback file id
+ * @param  int $nid Campaign id for this reportback
+ * @return array Values needed to render Kudos markup
+ */
 function dosomething_kudos_get_kudos_data_for_fid($fid, $nid) {
   $disable_reactions = dosomething_kudos_disable_reactions($nid);
   if ($disable_reactions) {
@@ -324,7 +330,6 @@ function dosomething_kudos_get_kudos_data_for_fid($fid, $nid) {
   $kudos['existing_kids'] = dosomething_kudos_get_kudos_for_user_by_file($fid);
   $kudos['reaction_totals'] = ['heart' => dosomething_kudos_get_total_for_file_by_term_name($fid, 'heart')];
   $kudos['allowed_reactions'] = [dosomething_kudos_get_term_id_by_name('heart')];
-  $kudos['disable_reactions'] = $disable_reactions;
 
   return $kudos;
 }
--- a/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
+++ b/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
@@ -309,3 +309,21 @@ function dosomething_kudos_term_is_selected($reportback_item, $term) {
 function dosomething_kudos_disable_reactions($nid) {
   return !variable_get('dosomething_reportback_kudos_enabled', FALSE) || dosomething_helpers_get_variable('node', $nid, 'disable_reactions');
 }
+
+function dosomething_kudos_get_kudos_data_for_fid($fid, $nid) {
+  $disable_reactions = dosomething_kudos_disable_reactions($nid);
+  if ($disable_reactions) {
+    return;
+  }
+
+  $kudos = [];
+  $kudos['fid'] = $fid;
+  $kudos['reaction_totals'] = [
+    'heart' => dosomething_kudos_get_total_for_file_by_term_name($fid, 'heart'),
+  ];
+  $kudos['existing_kids'] = dosomething_kudos_get_kudos_for_user_by_file($fid);
+  $kudos['reaction_totals'] = ['heart' => dosomething_kudos_get_total_for_file_by_term_name($fid, 'heart')];
+  $kudos['allowed_reactions'] = [dosomething_kudos_get_term_id_by_name('heart')];
+
+  return $kudos;
+}
--- a/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
+++ b/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
@@ -307,7 +307,7 @@ function dosomething_kudos_term_is_selected($reportback_item, $term) {
  * @return bool
  */
 function dosomething_kudos_disable_reactions($nid) {
-  return !variable_get('dosomething_reportback_kudos_enabled', FALSE) || dosomething_helpers_get_variable('node', $nid, 'disable_reactions');
+  return dosomething_helpers_get_variable('node', $nid, 'disable_reactions');
 }
 
 /**
@@ -317,11 +317,6 @@ function dosomething_kudos_disable_reactions($nid) {
  * @return array Values needed to render Kudos markup
  */
 function dosomething_kudos_get_kudos_data_for_fid($fid, $nid) {
-  $disable_reactions = dosomething_kudos_disable_reactions($nid);
-  if ($disable_reactions) {
-    return NULL;
-  }
-
   $kudos = [];
   $kudos['fid'] = $fid;
   $kudos['reaction_totals'] = [
--- a/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
+++ b/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
@@ -313,10 +313,9 @@ function dosomething_kudos_disable_reactions($nid) {
 /**
  * Grab all Kudos data required to render the markup for a given reportback fid
  * @param  int $fid Reportback file id
- * @param  int $nid Campaign id for this reportback
  * @return array Values needed to render Kudos markup
  */
-function dosomething_kudos_get_kudos_data_for_fid($fid, $nid) {
+function dosomething_kudos_get_kudos_data_for_fid($fid) {
   $kudos = [];
   $kudos['fid'] = $fid;
   $kudos['reaction_totals'] = [
--- a/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
+++ b/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
@@ -319,7 +319,7 @@ function dosomething_kudos_get_kudos_data_for_fid($fid = NULL) {
   if (! $fid) {
     return [];
   }
-  
+
   $kudos = [];
   $kudos['fid'] = $fid;
   $kudos['reaction_totals'] = [
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -599,7 +599,7 @@ function dosomething_reportback_view_entity($entity) {
       'kudos' => $kudos,
     );
 
-    // cache_set('ds_permalink_' . $entity->rbid . '_' . $fid . '_' . $is_owner, $permalink_vars, 'cache_dosomething_reportback');
+    cache_set('ds_permalink_' . $entity->rbid . '_' . $fid . '_' . $is_owner, $permalink_vars, 'cache_dosomething_reportback');
   }
   // Don't ever cache the signup button, as this changes/blocks submission.
   if($node->status == 'closed') {
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -523,7 +523,9 @@ function dosomething_reportback_view_entity($entity) {
     $fid = array_pop($entity->fids);
   }
 
-  $kudos = dosomething_kudos_get_kudos_data_for_fid($fid, $entity->nid);
+  if (dosomething_kudos_disable_reactions($entity->nid)) {
+    $kudos = dosomething_kudos_get_kudos_data_for_fid($fid);
+  }
 
   // Try to get the permalink from cache.
   if ($cached_permalink = cache_get('ds_permalink_' . $entity->rbid . '_' . $fid . '_' . $is_owner, 'cache_dosomething_reportback')) {
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -523,7 +523,7 @@ function dosomething_reportback_view_entity($entity) {
     $fid = array_pop($entity->fids);
   }
 
-  if (dosomething_kudos_disable_reactions($entity->nid)) {
+  if (! dosomething_kudos_disable_reactions($entity->nid)) {
     $kudos = dosomething_kudos_get_kudos_data_for_fid($fid);
   }
 
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -523,18 +523,7 @@ function dosomething_reportback_view_entity($entity) {
     $fid = array_pop($entity->fids);
   }
 
-
-  // @TODO Can we templatize/DRY some of this? It's fairly similiar code in multiple places.
-  $disable_reactions = dosomething_kudos_disable_reactions($entity->nid);
-
-  if (! $disable_reactions) {
-    $kudos = [];
-    $kudos['rbid'] = $entity->rbid;
-    $kudos['disable_reactions'] = $disable_reactions;
-    $kudos['allowed_reactions'] = [dosomething_kudos_get_term_id_by_name('heart')];
-    $kudos['reaction_totals'] = ['heart' => dosomething_kudos_get_total_for_file_by_term_name($entity->rbid, 'heart')];
-    $kudos['existing_kids'] = dosomething_kudos_get_kudos_for_user_by_file($entity->rbid);
-  }
+  $kudos = dosomething_kudos_get_kudos_data_for_fid($fid, $entity->nid);
 
   // Try to get the permalink from cache.
   if ($cached_permalink = cache_get('ds_permalink_' . $entity->rbid . '_' . $fid . '_' . $is_owner, 'cache_dosomething_reportback')) {
@@ -610,7 +599,7 @@ function dosomething_reportback_view_entity($entity) {
       'kudos' => $kudos,
     );
 
-    cache_set('ds_permalink_' . $entity->rbid . '_' . $fid . '_' . $is_owner, $permalink_vars, 'cache_dosomething_reportback');
+    // cache_set('ds_permalink_' . $entity->rbid . '_' . $fid . '_' . $is_owner, $permalink_vars, 'cache_dosomething_reportback');
   }
   // Don't ever cache the signup button, as this changes/blocks submission.
   if($node->status == 'closed') {
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -523,7 +523,7 @@ function dosomething_reportback_view_entity($entity) {
     $fid = array_pop($entity->fids);
   }
 
-  if (! dosomething_kudos_disable_reactions($entity->nid)) {
+  if (dosomething_kudos_disable_reactions($entity->nid)) {
     $kudos = dosomething_kudos_get_kudos_data_for_fid($fid);
   }
 
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -523,6 +523,7 @@ function dosomething_reportback_view_entity($entity) {
     $fid = array_pop($entity->fids);
   }
 
+  $kudos = [];
   if (! dosomething_kudos_disable_reactions($entity->nid)) {
     $kudos = dosomething_kudos_get_kudos_data_for_fid($fid);
   }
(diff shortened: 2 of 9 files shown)


lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -309,3 +309,22 @@ function dosomething_kudos_term_is_selected($reportback_item, $term) {
 function dosomething_kudos_disable_reactions($nid) {
   return !variable_get('dosomething_reportback_kudos_enabled', FALSE) || dosomething_helpers_get_variable('node', $nid, 'disable_reactions');
 }
+
+function dosomething_kudos_get_kudos_data_for_fid($fid, $nid) {
```
angaither: docblock me!

angaither: also the name of this function seems off, it does more than `get_kudos_data_for_fid` right?

itsjoekent: doh! 

and I wasn't sure of a good name, the point was to get all kudos data for a given fid so I just literally called the function that

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -309,3 +309,22 @@ function dosomething_kudos_term_is_selected($reportback_item, $term) {
...
+
+  $kudos = [];
+  $kudos['fid'] = $fid;
+  $kudos['reaction_totals'] = [
+    'heart' => dosomething_kudos_get_total_for_file_by_term_name($fid, 'heart'),
+  ];
+  $kudos['existing_kids'] = dosomething_kudos_get_kudos_for_user_by_file($fid);
+  $kudos['reaction_totals'] = ['heart' => dosomething_kudos_get_total_for_file_by_term_name($fid, 'heart')];
+  $kudos['allowed_reactions'] = [dosomething_kudos_get_term_id_by_name('heart')];
+  $kudos['disable_reactions'] = $disable_reactions;
```
angaither: is this line needed? it looks like we return `NULL` if the `disable_reactions` var is set

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -309,3 +309,22 @@ function dosomething_kudos_term_is_selected($reportback_item, $term) {
 function dosomething_kudos_disable_reactions($nid) {
   return !variable_get('dosomething_reportback_kudos_enabled', FALSE) || dosomething_helpers_get_variable('node', $nid, 'disable_reactions');
 }
+
+function dosomething_kudos_get_kudos_data_for_fid($fid, $nid) {
+  $disable_reactions = dosomething_kudos_disable_reactions($nid);
```
weerd: Not sure if the disable reactions part should go in this function. Would also simplify the parameter listing...

angaither:
Another cleanup task that might be worth getting into this PR is removing the checkbox to `enable_kudos` globally, since I don't think that's something we will want to turn off.

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
angaither: yeah I think kuos module is struggling with many functions that are named similar/do sort of the same thing in this file

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
itsjoekent: We'd then have to still put duplicate code everywhere,

```
$disable = kudos_disable(nid)
if (!disable)
 get_kudos_data
```

with this its just 1 line of code if you want to grab Kudos data & forget about it

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
itsjoekent: hm, I guess the template could just check if the array was set instead of using the key/val pair.

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
angaither: hm maybe like `prepare_kudos_data` oyy, perhaps writing the docblock will help come up with a title

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -523,18 +523,7 @@ function dosomething_reportback_view_entity($entity) {
...
-  $disable_reactions = dosomething_kudos_disable_reactions($entity->nid);
-
-  if (! $disable_reactions) {
-    $kudos = [];
-    $kudos['rbid'] = $entity->rbid;
-    $kudos['disable_reactions'] = $disable_reactions;
-    $kudos['allowed_reactions'] = [dosomething_kudos_get_term_id_by_name('heart')];
-    $kudos['reaction_totals'] = ['heart' => dosomething_kudos_get_total_for_file_by_term_name($entity->rbid, 'heart')];
-    $kudos['existing_kids'] = dosomething_kudos_get_kudos_for_user_by_file($entity->rbid);
-  }
```
weerd, line 537: Was it a problem before that the `$kudos` var would only be created if reactions weren't disabled... but then at the bottom the `$kudos` var was passed regardless of whether it was created or not? I wonder if the check for disabled reactions can just be on its own like:

``` php
$kudos = NULL;  // or []

if (! dosomething_kudos_disable_reactions($entity->nid)) {
  $kudos = dosomething_kudos_get_kudos_data_for_fid($fid);
}
```

particularly if other things need to know if reactions are disabled, tucking away `dosomething_kudos_disable_reactions()` inside of `dosomething_kudos_get_kudos_data_for_fid()` would prevent other things from knowing kudos are disabled and require the need to call `dosomething_kudos_disable_reactions()` to figure that out... :shrug:

itsjoekent, line 537: right but going back to what i wrote here: https://github.com/DoSomething/phoenix/pull/6689/files/e839e044531f01b9fe7635bbdf77a4b560469836#r69947073

we would then need to write this code 3 times. the point in self enclosing everything was to make it 1 line everywhere, no room for error. if you'd still prefer having the disable check in each file I will go back to that.

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
weerd: Yeah, but it keeps other logic from knowing if kudos are disabled... just future thinking... would help keep the name of the function simple if it's just doing one thing.

You also don't need to save the result in `$disable`... just use the function in the conditional.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
weerd, line 537: That's also assuming this is the _only_ function that will need to know if Kudos is disabled. As more functionality is added I'm guessing other things will need to know this, regardless of whether they are trying to grab kudos data or not.

weerd, line 537: Just an alternative way of thinking that considers future functionality. The code isn't wrong, just providing an additional perspective with other considerations 😄

itsjoekent, line 537: I'll give it a shot & see if it makes sense!

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -360,13 +360,7 @@ function paraneue_dosomething_get_themed_reportback_item($data) {
...
   }
 
-  if ($data->id && !$data->disable_reactions) {
-    $data->reaction_totals = [
-      'heart' => dosomething_kudos_get_total_for_file_by_term_name($data->id, 'heart'),
-    ];
-
-    $data->existing_kids = dosomething_kudos_get_kudos_for_user_by_file($data->id);
-  }
+  $data->kudos = dosomething_kudos_get_kudos_data_for_fid($data->id, $data->nid);
```
weerd: You probs still need to work with an `if` conditional here if `$data->id` doesn't exist... there's no check or that in `dosomething_kudos_get_kudos_data_for_fid()`...

itsjoekent: hm, does it make more sense to add the check here or in the kudos function?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
weerd, line 537: Coolio. Yeah, just consider that _good_ functions are ones that ideally do 1 thing, and do that really well. When they turn into functions that do _a lot_ that means they're turning into Drupal-style functions... :trollface:

itsjoekent:
@angaither removed ✅  

~~edit: wait nope didnt commit that~~

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -307,5 +307,24 @@ function dosomething_kudos_term_is_selected($reportback_item, $term) {
  * @return bool
  */
 function dosomething_kudos_disable_reactions($nid) {
-  return !variable_get('dosomething_reportback_kudos_enabled', FALSE) || dosomething_helpers_get_variable('node', $nid, 'disable_reactions');
```
weerd, line 310: Did we need this as an `||` statement?

~~also, remove the 👾 after the `return`.~~ mis-read the diff!

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -307,5 +307,24 @@ function dosomething_kudos_term_is_selected($reportback_item, $term) {
...
+  return dosomething_helpers_get_variable('node', $nid, 'disable_reactions');
+}
+
+/**
+ * Grab all Kudos data required to render the markup for a given reportback fid
+ * @param  int $fid Reportback file id
+ * @param  int $nid Campaign id for this reportback
+ * @return array Values needed to render Kudos markup
+ */
+function dosomething_kudos_get_kudos_data_for_fid($fid, $nid) {
```
weerd: After removing the "disabled reactions" check, you no longer need the `$nid` at all. One less param and simpler function!

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
itsjoekent, line 310: all good

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
itsjoekent: oh crap i rewrote all the other function calls without but forgot to actually remove it here

itsjoekent:
@weerd I think it's good to go

itsjoekent:
wait something broke locally, let me see whats goin on

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -523,17 +523,8 @@ function dosomething_reportback_view_entity($entity) {
...
-  $disable_reactions = dosomething_kudos_disable_reactions($entity->nid);
-
-  if (! $disable_reactions) {
-    $kudos = [];
-    $kudos['rbid'] = $entity->rbid;
-    $kudos['disable_reactions'] = $disable_reactions;
-    $kudos['allowed_reactions'] = [dosomething_kudos_get_term_id_by_name('heart')];
-    $kudos['reaction_totals'] = ['heart' => dosomething_kudos_get_total_for_file_by_term_name($entity->rbid, 'heart')];
-    $kudos['existing_kids'] = dosomething_kudos_get_kudos_for_user_by_file($entity->rbid);
+  if (dosomething_kudos_disable_reactions($entity->nid)) {
```
weerd: Shouldn't this be `if (! dosomething_kudos_disable_reactions($entity->nid)) { ... }`?

itsjoekent: literally just as i was commiting this, yeah thats why it broke locally. good thing i double checked

itsjoekent: blame @lkpttn i was trying to code & learn about the magical card marketplace at the same time lol

itsjoekent:
sweet everything works locally now

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -360,12 +360,8 @@ function paraneue_dosomething_get_themed_reportback_item($data) {
     return NULL;
   }
 
-  if ($data->id && !$data->disable_reactions) {
-    $data->reaction_totals = [
-      'heart' => dosomething_kudos_get_total_for_file_by_term_name($data->id, 'heart'),
-    ];
-
-    $data->existing_kids = dosomething_kudos_get_kudos_for_user_by_file($data->id);
+  if (! dosomething_kudos_disable_reactions($data->campaign->id)) {
```
weerd: I still think you need the `$data->id && ...` in the conditional as before.

itsjoekent: Oh I added it in the template logic, I can do it here since that makes more sense

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -307,5 +307,23 @@ function dosomething_kudos_term_is_selected($reportback_item, $term) {
...
+  return dosomething_helpers_get_variable('node', $nid, 'disable_reactions');
+}
+
+/**
+ * Grab all Kudos data required to render the markup for a given reportback fid
+ * @param  int $fid Reportback file id
+ * @return array Values needed to render Kudos markup
+ */
+function dosomething_kudos_get_kudos_data_for_fid($fid) {
+  $kudos = [];
```
weerd, line 323: as a safety measure, I would suggest settinf `$fid = NULL` as a default for the param and:

``` php
if (! $fid) {
  return [];
}
```

itsjoekent, line 323: yar i asked before if that made sense but i think the comment chain got closed out before you had a chance to see it https://github.com/DoSomething/phoenix/pull/6689#discussion_r69950096

itsjoekent:
@weerd added the FID check both in the $data->id conditional and in the new kudos function just to be safe on all fronts

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -360,12 +360,8 @@ function paraneue_dosomething_get_themed_reportback_item($data) {
     return NULL;
   }
 
-  if ($data->id && !$data->disable_reactions) {
-    $data->reaction_totals = [
-      'heart' => dosomething_kudos_get_total_for_file_by_term_name($data->id, 'heart'),
-    ];
-
-    $data->existing_kids = dosomething_kudos_get_kudos_for_user_by_file($data->id);
+  if (! dosomething_kudos_disable_reactions($data->campaign->id) && $data->id) {
```
weerd: Bit picky, but I would put the `$data->id` first, because if that doesn't exist then there's no point in calling the function to check if kudos reactions are disabled. If the first item in an `&&` statement is false, then the whole thing is false and the second check doesn't even bother happening.

itsjoekent:
@weerd like that?

https://github.com/DoSomething/phoenix/pull/6689/files#diff-dec3a9ad5859cdb163537532751f59e8R363

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
```diff
@@ -249,22 +249,14 @@ function paraneue_dosomething_preprocess_reportback_gallery(&$vars, $count = 6)
...
-    $item->disable_reactions = $disable_reactions;
-    $item->allowed_reactions = $allowed_tids;
+    $id = $item->id;
     if ($item instanceof ReportbackItem) {
-      $gallery['item_ids'][] = $item->id;
+      $gallery['item_ids'][] = $id;
     }
-    $gallery['items'][] = paraneue_dosomething_get_themed_reportback_item($item);
+
+    $item = paraneue_dosomething_get_themed_reportback_item($item);
```
weerd: you're overwriting `$item` from the loop with a new var `$item`... it's also not really necessary since you only use it again right below. Just keep it

``` php
$gallery['items'][] = paraneue_dosomething_get_themed_reportback_item($item);
```

With `$id` above it, you only use it in one place, so no point in assigning it to a variable, just keep it:

``` php
$gallery['item_ids'][] = $item->id;
```

itsjoekent: sorry originally i was passing item + id into the kudos function, and then adding it to the array. however i decided to just ditch that later & forgot to fix this

lib/themes/dosomething/paraneue_dosomething/templates/reportback/reportback-permalink.tpl.php
```diff
@@ -40,10 +40,10 @@
               <?php print filter_xss($rb['caption']); ?> - <?php print check_plain($user->first_name); ?>
             </div>
           <?php endif; ?>
-          <?php if (isset($kudos) && !$kudos['disable_reactions']): ?>
-            <ul class="form-actions -inline kudos" <?php print $kudos['rbid'] ? 'data-reportback-item-id="' . $kudos['rbid'] . '"' : ''; ?>>
+          <?php if (isset($content['kudos']) && isset($content['kudos']['fid'])): ?>
+            <ul class="form-actions -inline kudos <?php print json_encode($kudos) ?>" <?php print $kudos['fid'] ? 'data-reportback-item-id="' . $kudos['fid'] . '"' : ''; ?>>
```
weerd: why are you `json_econde()`ing `$kudos`?

itsjoekent: arghh i removed it from the other templates but forgot this one i guess

weerd: Would not some of this stuff come up if you're testing locally?

itsjoekent: I wasn't really looking at the markup much after i did the template updates, just making sure the functionality worked

lib/themes/dosomething/paraneue_dosomething/templates/reportback/reportback-permalink.tpl.php
```diff
@@ -40,10 +40,10 @@
               <?php print filter_xss($rb['caption']); ?> - <?php print check_plain($user->first_name); ?>
             </div>
           <?php endif; ?>
-          <?php if (isset($kudos) && !$kudos['disable_reactions']): ?>
-            <ul class="form-actions -inline kudos" <?php print $kudos['rbid'] ? 'data-reportback-item-id="' . $kudos['rbid'] . '"' : ''; ?>>
+          <?php if (isset($content['kudos']) && isset($content['kudos']['fid'])): ?>
+            <ul class="form-actions -inline kudos" <?php print $kudos['fid'] ? 'data-reportback-item-id="' . $kudos['fid'] . '"' : ''; ?>>
```
weerd, line 44: since this is now in the `$content` variable... how does `$kudos['fid']` know that it's referring to `$content['kudos']['fid']`?

weerd, line 44: Double check the DBLogs... this is probably throwing a few notices!

itsjoekent:
just went through and checked everything again, functionality seems ok, code looks good

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -523,17 +523,8 @@ function dosomething_reportback_view_entity($entity) {
...
-  $disable_reactions = dosomething_kudos_disable_reactions($entity->nid);
-
-  if (! $disable_reactions) {
-    $kudos = [];
-    $kudos['rbid'] = $entity->rbid;
-    $kudos['disable_reactions'] = $disable_reactions;
-    $kudos['allowed_reactions'] = [dosomething_kudos_get_term_id_by_name('heart')];
-    $kudos['reaction_totals'] = ['heart' => dosomething_kudos_get_total_for_file_by_term_name($entity->rbid, 'heart')];
-    $kudos['existing_kids'] = dosomething_kudos_get_kudos_for_user_by_file($entity->rbid);
+  if (! dosomething_kudos_disable_reactions($entity->nid)) {
```
weerd, line 527: maybe set:

``` php
$kudos = [];
```

above the conditional otherwise you'll have an error here: ~~https://github.com/DoSomething/phoenix/pull/6689/files#diff-6cd2bd158ecf1b2419b491fc3dae7ceaR601~~ if `$kudos` is not set @_@

Apparently, link is incorrect. @deadlybutter check line 601 in **dosomething_reportback.module** where the `$permalink_vars` are set:

``` php
'kudos' => $kudos,
```

weerd, line 527: It's something from a while back, seems like, and just didn't get caught 😞

itsjoekent, line 527: did you link to the right place? I'm not sure I see what you're talking about.

lib/themes/dosomething/paraneue_dosomething/templates/system/patterns/photo.tpl.php
```diff
@@ -13,12 +13,11 @@
...
     </figcaption>
   </figure>
-
-  <?php if ($content->id && !$content->disable_reactions): ?>
+  <?php if (isset($content->kudos) && isset($content->kudos['fid'])): ?>
     <ul class="form-actions -inline kudos">
       <li>
-        <button class="js-kudos-button kudos__icon <?php print dosomething_kudos_term_is_selected($content, 'heart') ? 'is-active' : '' ?>" data-kudos-term-id="<?php print $content->allowed_reactions[0]; ?>" data-kid="<?php print dosomething_helpers_isset($content->existing_kids[$content->allowed_reactions[0]], 'kid') ?>"></button>
-        <span class="counter"><?php print $content->reaction_totals['heart']; ?></span>
+        <button class="js-kudos-button kudos__icon <?php print dosomething_kudos_term_is_selected($content->kudos, 'heart') ? 'is-active' : '' ?>" data-kudos-term-id="<?php print $content->kudos['allowed_reactions'][0]; ?>" data-kid="<?php print dosomething_helpers_isset($content->kudos['existing_kids'][$content->kudos['allowed_reactions'][0]], 'kid') ?>"></button>
```
weerd, line 19: kinda weird that kudos in `$content` is referenced as an object in this tpl but an array in other. Not a big deal, but maybe something to universalize at a later date :)

weerd:
Looks good. Let's give this a test on Staging. 👍

lib/themes/dosomething/paraneue_dosomething/templates/system/patterns/photo.tpl.php
itsjoekent, line 19: Thats just how its been, even before Kudos :/

itsjoekent:
🚂 🚋 🚋

itsjoekent:
Cool everything works on staging but I am noticing one problem which isn't related to these changes ^

When the cache for the perma link is created, whatever state Kudos is in, its basically locked in that state on the permalink. the UI seems to change fine but its not reflected on the backend.

@weerd @angaither

angaither:
@deadlybutter confused by this... kudos are not posting when applied from a permalink page? Or the number never reflects more even though the row is created in the kudos table?

itsjoekent:
if you open a reportback perma link, and the kudos for that RB are 10 + yourself, that will show on the permalink. You can then click the button again & it will change, but if you refresh its the same view from the first page load. The only way to make this page update is by clearing cache

angaither:
@deadlybutter we had a similar issue with the signup button and [moved the things that shouldn't be cached](https://github.com/DoSomething/phoenix/blob/0fffbd859c1cac6767bdc236b2a4392a0baddd0e/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module#L607) to the outside of the cache, that should fix the issue you're seeing

weerd:
@deadlybutter You can harness the `ApiCache` class to help you do this on saving a new Kudos record... I suspected this might happen since the response w/ prior info is cached.

@angaither I can help with it, or assign me an issue

itsjoekent:
ah, gotcha, I can make an issue for that

edit: or andrea has it i guess?

weerd:
Just FYI: it's slightly different if it's going through the API...

weerd:
If this entirely a permalink cache issue though, then nevermind...

PR #6701, opened 2016-07-08
Fixes a bug in template and adds utility function.

### LINKED ISSUE #6700: Repeated notices regarding undefined offset 641
#### Current Behavior
- We are getting repeated notices of an undefined offset in the DB Log Reports

```
Notice: Undefined offset: 641 in include() (line 19 of /var/www/staging.beta.dosomething.org/releases/20160707205512/lib/themes/dosomething/paraneue_dosomething/templates/system/patterns/photo.tpl.php).
```
#### Desired Behavior
- Not get any notices and all working without errors logging.
#### Steps to Replicate
- Load any campaign action page.

weerd:
#### What's this PR do?

This PR fixes an issue where on every campaign action page load, there are about 6 or less (because of the number of initial gallery items) notices regarding an undefined offset `641`. The issue is that the function in the template that is called to find a particular _nested_ array item won't work because the `dosomething_helpers_isset()` function cannot handle dealing with grabbing values if the value is nested. 

Now we use the handy `data_get()` helper function brought in by the **tightenco/collect** package (amongst a bunch of other awesome stuff that package brings in).

This PR also cleans up: 
- `dosomething_helpers_isset()` function to rename a param to help make it more obvious what it is.
- `dosomething_kudos_term_is_selected()` since the param name no longer reflected what was being passed in to it.. it is no longer a `$reportback_item` object, but a `$kudos` array.

Load a campaign action page. If you don't see any `Notice: Undefined offset: 641` errors, then all is well!
#### Relevant tickets

Fixes #6700
#### Checklist
- [ ] Tested on staging.

---

@deadlybutter @angaither @jessleenyc

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -29,22 +29,56 @@ function dosomething_helpers_array_is_associative(array $data) {
  *
  * @param string|array|object $var  Variable to check.
  * @param string $key  If $var is an array or object, key to check.
- * @param null $value  Set a default value to assign if $var not set.
+ * @param null $default  Set a default value to assign if $var not set.
  *
  * @return array|object|string|integer|null
  *
  * @todo rename this function to dosomething_helpers_get_value()
  */
-function dosomething_helpers_isset($var, $key = NULL, $value = NULL) {
+function dosomething_helpers_isset($var, $key = NULL, $default = NULL) {
   if (is_array($var) && $key) {
-    return isset($var[$key]) ? $var[$key] : $value;
+    return isset($var[$key]) ? $var[$key] : $default;
   }
 
   if (is_object($var) && $key) {
-    return isset($var->$key) ? $var->$key : $value;
+    return isset($var->$key) ? $var->$key : $default;
   }
 
-  return isset($var) ? $var : $value;
+  return isset($var) ? $var : $default;
+}
+
+/**
+ * Utility function to check and obtain a nested value in an array.
+ *
+ * @param  array $data
+ * @param  string $key
+ * @param  string $parent_key
+ * @return string|null
+ *
+ * @todo  should update to check if array or object and just make it smarter in general.
+ */
+function dosomething_helpers_get_nested_value($data, $key, $parent_key = NULL) {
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -29,22 +29,56 @@ function dosomething_helpers_array_is_associative(array $data) {
...
  * @param string $key  If $var is an array or object, key to check.
- * @param null $value  Set a default value to assign if $var not set.
+ * @param null $default  Set a default value to assign if $var not set.
  *
  * @return array|object|string|integer|null
  *
  * @todo rename this function to dosomething_helpers_get_value()
  */
-function dosomething_helpers_isset($var, $key = NULL, $value = NULL) {
+function dosomething_helpers_isset($var, $key = NULL, $default = NULL) {
```
itsjoekent, line 38: ooo default value 💯

itsjoekent:
👍  🚢

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -275,16 +275,16 @@ function dosomething_kudos_get_kudos_for_user_by_file($fid) {
...
  *
- * @param  object $reportback_item
+ * @param  array $kudos
  * @param  string $term
  * @return bool
  */
-function dosomething_kudos_term_is_selected($reportback_item, $term) {
-  $reportback_item = (object) $reportback_item;
+function dosomething_kudos_term_is_selected($kudos, $term) {
+  $kudos = (object) $kudos;
```
DFurnes: Why do you cast to an object here?

weerd: I had it from before when we were passing the RB object... but now it's just a kudos array; but not 100% sure if it's ever a Kudos object. Also, I just prefer arrow notation over array notation 😖

DFurnes: Haha okay! It might be worth type-hinting `array` or the classname in the future so we can ensure the right input is being passed.

DFurnes:
🏁 🚀

PR #6702, opened 2016-07-08
Adds framework agnostic Collections and helper functions

weerd:
#### What's this PR do?

This adds the tightenco/collect composer package which is specifically created to allow adding some of the sweet sugary Collections and helper functions that the Illuminate/Support package adds, but in a framework agnostic fashion and without a bunch of the other dependencies that Illuminate/Support normally brings with it.

Now we'll have access to the [Collection methods](https://laravel.com/docs/5.2/collections) from the Collections class, simply use the `collect()` helper method or instantiate the class.

Also have access to the [Arr methods](https://laravel.com/docs/5.2/helpers#available-methods) from the Arr class.

👀 

Thanks @DFurnes for finding this package! Apart from not using **Illuminate/Support** because it brings with it a few dependencies that we likely won't be using on the Drupal site, it brings the full list of helper functions like `dd()` which conflict and throw a fatal exception with the Drupal Devel module `dd()` function.

Make sure to run `composer install` to download and install this new composer package.
#### Relevant tickets

Refs #6700 (will help provide a better fix for this issue)

---

@DFurnes @angaither 

Pinging you all below so that you're also aware that once merged we'll have access to some fancier shiz that you might be accustomed to in Laravel, but now in Drupal.
CC: @deadlybutter @sbsmith86 @DeeZone @aaronschachter @katiecrane

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,8 @@
   },
   "require": {
     "stripe/stripe-php": "^1.18.0",
-    "zendesk/zendesk_api_client_php": "^1.2.0"
+    "zendesk/zendesk_api_client_php": "^1.2.0",
+    "tightenco/collect": "^5.2"
   },
   "require-dev": {
     "symfony/var-dumper": "^3.1"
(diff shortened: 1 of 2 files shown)


DFurnes:
🚢 👏

Issue #6682, opened 2016-07-06
Set up onboarding experience in fancy JS world


angaither:
Hardcode copy/slides in :js:

itsjoekent:
base work required to do this issue: https://github.com/DoSomething/phoenix/pull/6706

PR #6706, opened 2016-07-11
Laying the foundations for react / onboarding experience

itsjoekent:
#### What's this PR do?

adds some starter react components for onboarding experience

`ds assets`

go to `/admin/config/dosomething/dosomething_campaign`

click standard campaigns --> action page --> Enable onboarding

very rough, but the point is to get the base up so Diego can jump on & we have a direction to work from
#### Relevant tickets

Fixes #~oops shoulda made an issue~
#6682

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
@@ -198,6 +198,15 @@ function paraneue_dosomething_preprocess_node_campaign(&$vars) {
   // If this is not the pitch page:
   if (!isset($vars['is_pitch_page'])) {
     $vars['reportbacks_gallery'] = paraneue_dosomething_preprocess_reportback_gallery($vars);
+
+    if (variable_get('dosomething_campaign_enable_onboarding', FALSE)) {
+      drupal_add_js(PARANEUE_PATH . '/dist/onboarding.js', [
+        'group'      => JS_THEME,
+        'weight'     => 1000,
+        'every_page' => FALSE,
+        'preprocess' => FALSE,
+      ]);
+    }
   }
 
   if (isset($vars['is_pitch_page'])) {
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
@@ -206,14 +206,7 @@ function paraneue_dosomething_preprocess_node_campaign(&$vars) {
         'every_page' => FALSE,
         'preprocess' => FALSE,
       ]);
-      drupal_add_js(
-        array('dsOnboarding' =>
-          array(
-            'enabled' => true
-          )
-        ),
-        'setting'
-      );
+      drupal_add_js(['dsOnboarding' => ['enabled' => true]], 'setting');
     }
   }
 
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
@@ -206,6 +206,14 @@ function paraneue_dosomething_preprocess_node_campaign(&$vars) {
         'every_page' => FALSE,
         'preprocess' => FALSE,
       ]);
+      drupal_add_js(
+        array('dsOnboarding' =>
+          array(
+            'enabled' => true
+          )
+        ),
+        'setting'
+      );
     }
   }
 
--- a/lib/themes/dosomething/paraneue_dosomething/js/app.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/app.js
@@ -42,6 +42,11 @@ import 'validators/organDonation';
 
 import 'campaign/prototype';
 
+// Load React components
+import 'components/onboard';
+import 'components/carousel';
+import 'components/context_slide';
+
 /**
  * Let's go!
  */
--- a/lib/themes/dosomething/paraneue_dosomething/js/app.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/app.js
@@ -42,11 +42,6 @@ import 'validators/organDonation';
 
 import 'campaign/prototype';
 
-// Load React components
-import 'components/onboard';
-import 'components/carousel';
-import 'components/context_slide';
-
 /**
  * Let's go!
  */
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/carousel.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/carousel.js
@@ -1,12 +1,14 @@
 const React = require('react');
 
-const Carousel = React.createClass({
-  getInitialState: function() {
-    return {
-      activeSlide: 0,
+class Carousel extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      activeSlide: 0
     };
-  },
-  render: function() {
+  }
+
+  render() {
     return (
       <div className="carousel">
         {this.props.slides[this.state.activeSlide]}
@@ -16,6 +18,6 @@ const Carousel = React.createClass({
       </div>
     );
   }
-});
+}
 
 export default Carousel;
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/carousel.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/carousel.js
@@ -1,24 +1,15 @@
 const React = require('react');
 
-import ContextSlide from './context_slide.js';
-
 const Carousel = React.createClass({
   getInitialState: function() {
     return {
-      activeSlide: 1,
+      activeSlide: 0,
     };
   },
   render: function() {
-    var slide;
-    switch (this.state.activeSlide) {
-      case 1: slide = <ContextSlide />; break;
-      case 2: // slide = Y
-      case 3: // slide = Z
-    }
-
     return (
       <div className="carousel">
-        {slide}
+        {this.props.slides[this.state.activeSlide]}
         <div className="carousel__controls">
           buttons go here
         </div>
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/carousel.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/carousel.js
@@ -1,0 +1,30 @@
+const React = require('react');
+
+import ContextSlide from './context_slide.js';
+
+const Carousel = React.createClass({
+  getInitialState: function() {
+    return {
+      activeSlide: 1,
+    };
+  },
+  render: function() {
+    var slide;
+    switch (this.state.activeSlide) {
+      case 1: slide = <ContextSlide />; break;
+      case 2: // slide = Y
+      case 3: // slide = Z
+    }
+
+    return (
+      <div className="carousel">
+        {slide}
+        <div className="carousel__controls">
+          buttons go here
+        </div>
+      </div>
+    );
+  }
+});
+
+export default Carousel;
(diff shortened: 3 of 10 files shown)


lib/themes/dosomething/paraneue_dosomething/js/app.js
```diff
@@ -42,6 +42,11 @@ import 'validators/organDonation';
 
 import 'campaign/prototype';
 
+// Load React components
+import 'components/onboard';
```
DFurnes: Let's move these to a separate entry point (that we conditionally load via the feature flag), so we're not loading a bunch of code that isn't getting used for every user. You can add a new entry-point in the [webpack config](https://github.com/DoSomething/phoenix/blob/dba088f6e7c1bc48e830019a4dfd5fcccdca9d33/lib/themes/dosomething/paraneue_dosomething/webpack.config.js#L6-L7), and then add it to the action page via `drupal_add_js`.

lib/themes/dosomething/paraneue_dosomething/js/components/carousel.js
```diff
@@ -0,0 +1,30 @@
...
+const Carousel = React.createClass({
+  getInitialState: function() {
+    return {
+      activeSlide: 1,
+    };
+  },
+  render: function() {
+    var slide;
+    switch (this.state.activeSlide) {
+      case 1: slide = <ContextSlide />; break;
```
DFurnes: I think it might make more sense to have the slides in an array here, and display them based on an offset.

itsjoekent: Ah so pass the slides in as a property & keep activeSlide on the state, then do something like `this.props.slides[this.state.activeSlide]`

lib/themes/dosomething/paraneue_dosomething/js/components/carousel.js
```diff
@@ -0,0 +1,30 @@
+const React = require('react');
+
+import ContextSlide from './context_slide.js';
+
+const Carousel = React.createClass({
```
DFurnes: Let's use ES6 classes for our stateful components. The `React.createClass` syntax is more-or-less deprecated:

``` js
class Carousel extends React.Component {
   constructor(props) {
     super(props);

     this.state = { activeSlide : 1 };
   }

   render() {
     // ...
   }
}
```

itsjoekent: sure thing 👍

lib/themes/dosomething/paraneue_dosomething/js/app.js
itsjoekent: is react.js fine? or should it be onboarding specific?

lib/themes/dosomething/paraneue_dosomething/js/components/onboard.js
```diff
@@ -0,0 +1,21 @@
...
+  render: function() {
+    return (
+      <div className="onboard">
+        <Carousel />
+      </div>
+    );
+  }
+});
+
+if (localStorage.getItem('dosomething_enable_onboarding') == "true") {
```
DFurnes: Why `localStorage` over a `Drupal.setting` value?

lib/themes/dosomething/paraneue_dosomething/js/components/onboard.js
```diff
@@ -0,0 +1,21 @@
...
+  ReactDOM.render(
```
DFurnes: Definitely not necessary for this pull request, but let's think about having a central place to handle initializing an rendering components like we do [in the voting app](https://github.com/DoSomething/voting-app/blob/04234aebff2cf60114e3d0c36dd02d25d2fe555b/resources/assets/js/app.js#L54-L60).

lib/themes/dosomething/paraneue_dosomething/js/components/onboard.js
itsjoekent: connivence of not having to touch drupal haha, but if we're going to make drupal logic now anyway ill just switch this for a feature flag

lib/themes/dosomething/paraneue_dosomething/templates/system/page--full-width.tpl.php
```diff
@@ -12,6 +12,7 @@
 <?php if (!empty($tabs['#primary'])): ?><nav class="admin-tabs"><?php print render($tabs); ?></nav><?php endif; ?>
 <?php print $messages; ?>
 
+<div id="onboarding" class="wrapper"></div>
```
DFurnes: It might be nice to think about a convention for React component containers to make it clear what this is used for, maybe `#jsx-onboarding` or `#react-onboarding`?

lib/themes/dosomething/paraneue_dosomething/js/components/carousel.js
weerd: @DFurnes didn't they just deprecate the last established method for component creation @_@ haha They're moving quick!

lib/themes/dosomething/paraneue_dosomething/templates/system/page--full-width.tpl.php
DFurnes: Also not absolutely necessary for this PR, but it feels iffy to have a campaign-specific element in the global "page" templates. Maybe we should be specifically overriding the page template for the campaign node?

itsjoekent: yeah i was wondering about this myself when i was writing it

lib/themes/dosomething/paraneue_dosomething/js/components/carousel.js
DFurnes: Yep!

lib/themes/dosomething/paraneue_dosomething/js/components/carousel.js
itsjoekent: facebook, everyone

DFurnes: `React.createClass` was the original method for component creation. Not officially deprecated yet, but Facebook are [moving away from using it internally](https://github.com/reactjs/core-notes/blob/9d503233bdfacb09c28ac29f4da227873e40067d/2016-06/june-30.md#update-on-createclass--es-classes) and it's reasonable to assume it might be removed at some point in the future.

lib/themes/dosomething/paraneue_dosomething/js/app.js
DFurnes: I think it'd make sense to make a `onboarding.js` entry point. If we end up having other bundles that also use React, it'll get pulled into `lib.js` and shared between bundles automatically.

lib/themes/dosomething/paraneue_dosomething/js/components/onboard.js
itsjoekent: actually i just realized @DFurnes -- Diego is already working on this setting, id rather not conflict with him

weerd: Yeah, I'm working on adding a `dsUser` to the `Drupal.settings` object so we have access to a current user object via that settings object.

@deadlybutter What we talked about using local storage for was to store whether the member user already saw the onboarding experience or not. It would be a simple implementation and not needing a feature flag and database record for storing that a user has seen the item already for this beta. If it turns out we go a different direction with the beta for onboarding, then we'd have some Drupal cleanup and a random database record to deal with, whereas the cleanup with using local storage would be virtually nothing.

For _actually_ showing the onboarding or not across all campaigns and user... yeah, that should be a feature flag in the Drupes.

itsjoekent: right that was the idea I suggested, this here was just me trying to avoid touching drupal to get it out the door

sbsmith86: Just a note... @katiecrane already started adding user information to `Drupal.settings` for the organ donation work we were doing...I think we should try to consolidate them as to not have two user objects...we really just need one so that anything can use it as needed. https://github.com/DoSomething/phoenix/blob/973e7cbeaf72abf05bd16c30cc7851d2e75b080e/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc#L222

itsjoekent: @weerd ^

DFurnes: Using `window.localStorage` for low-fi tracking onboarding makes sense to me. 👍 You should be able to remove the client-side feature flag altogether if you conditionally include the bundle based on the Drupal feature flag.

weerd: @katiecrane @sbsmith86 gotcha. good to know. Maybe we can consolidate after? I'm actually using a call to Northstar to grab the user info based on the user id, so that we are pulling the User info from our main User info pool. I'm also adding it as a helper function so we can use that in the API (we repeat a few calls to Northstar there for the info) and so moving forward we just call 1 function and the output could be placed in the Drupal user object in the JS or in the API response via PHP. Thoughts?

sbsmith86: @weerd that's fine. just wanted to flag that.

weerd: Yeah, defs thanks @sbsmith86. I'll ping you and @katiecrane in my PR for it so we can chat some more there about it. Thanks for the heads up!

sbsmith86: @weerd but also one thing that would be nice to have is user activity information in that object...some of it might be outside the scope of just northstar. Just something to think about.

weerd: Yeah defs. As a helper function it calls Northstar to grab the info, but then processes it, so it's not just a call to NS and that's it. We could defs throw in user activity as well, and as a helper function can be used by other code outside of just the JS settings object 💃

lib/themes/dosomething/paraneue_dosomething/js/components/carousel.js
itsjoekent: for reference this was pretty helpful https://toddmotto.com/react-create-class-versus-component/

itsjoekent:
@DFurnes Okie dokes I think I got it all

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
```diff
@@ -198,6 +198,23 @@ function paraneue_dosomething_preprocess_node_campaign(&$vars) {
...
+
+    if (variable_get('dosomething_campaign_enable_onboarding', FALSE)) {
+      drupal_add_js(PARANEUE_PATH . '/dist/onboarding.js', [
+        'group'      => JS_THEME,
+        'weight'     => 1000,
+        'every_page' => FALSE,
+        'preprocess' => FALSE,
+      ]);
+      drupal_add_js(
+        array('dsOnboarding' =>
```
DFurnes: Short array syntax! `array(...)` → `[...]`

itsjoekent: oh yeah thats a thing :facepalm:

DFurnes:
One nitpick, everything else looks good to me! 🚢

itsjoekent:
🚂 🚋 🚋

itsjoekent:
Not working on stage, @DFurnes any ideas what this error is about?

screen shot 2016-07-11 at 3 43 29 pm
screen shot 2016-07-11 at 3 44 07 pm

DFurnes:
Do you get that error if you run a production build on your local, via `ds assets` or `npm run build`?

itsjoekent:
@DFurnes ds assets is fine, npm run build throws an error

edit: facepalm, running this in the wrong directory
npm run build works fine

DFurnes:
Strange, I was able to reproduce the error locally with `npm run build`! I pushed up a fix in DoSomething/webpack-config#1 and will make a patch release as soon as it's merged. :)

DFurnes:
And fixed on staging. ✅

PR #6709, opened 2016-07-11
Extracts user info collection to a universal helper function.

weerd:
#### What's this PR do?

This PR updates the `dosomething_helpers_get_current_user()` function and adds a `dosomething_helpers_get_user_info($drupal_id)` function. Now we can call the former function and if there is a current user logged in, then it will pass the drupal id to the latter function and actually retrieve all the user information needed from our Northstar user API, process the info a bit to only show what is needed and then return it as an array.

This is used in the new `dsUser` object that is located in the `Drupal.settings` JS object on the front-end, but can also be used for DRYing up the current code in the Phoenix API when the user object is created [here](https://github.com/DoSomething/phoenix/blob/6c02a9e93670fb56c9401789ce135c8a3fca20b5/lib/modules/dosomething/dosomething_kudos/includes/Kudos.php#L126), [here](https://github.com/DoSomething/phoenix/blob/78f318426a2f64d987c5e94e9f2b9081e8ffd45a/lib/modules/dosomething/dosomething_reportback/includes/ReportbackItem.php#L122), and [here](https://github.com/DoSomething/phoenix/blob/78f318426a2f64d987c5e94e9f2b9081e8ffd45a/lib/modules/dosomething/dosomething_reportback/includes/Reportback.php#L179).

The `dsUser` object currently has 2 properties, one for `info` with the processed user info from NS, and `activity`, which is currently `null` but could include activity info for the user in the future as [suggested by](https://github.com/DoSomething/phoenix/pull/6706#discussion_r70301872) @sbsmith86. This could be done with a separate helper function call.

Try it out. Load a campaign, open the dev console and type `Drupal.settings` and hit return. Do you see the new `dsUser` object with info if signed in? Also, did Kudos break?! Hopefully not!

We should consolidate this and the code [here](https://github.com/DoSomething/phoenix/blob/973e7cbeaf72abf05bd16c30cc7851d2e75b080e/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc#L222) and only have one user object in the `Drupal.settings`.
#### Relevant tickets

Refs #6682 
#### Checklist
- [ ] Tested on staging.

---

@DFurnes @deadlybutter @angaither 

cc: @sbsmith86 @katiecrane

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -688,7 +688,7 @@ function dosomething_helpers_get_current_language_content_code() {
 /**
  * Get the current, logged in user.
  *
- * @return null|object
+ * @return null|array
  */
 function dosomething_helpers_get_current_user() {
   global $user;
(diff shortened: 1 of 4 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
```diff
@@ -203,6 +203,16 @@ function paraneue_dosomething_preprocess_node_campaign(&$vars) {
   if (isset($vars['is_pitch_page'])) {
     $vars['show_persistent_signup'] = theme_get_setting('show_persistent_signup');
   }
+
+  drupal_add_js(
+    [
+      'dsUser' => [
+        'info' => dosomething_helpers_get_current_user(),
+        'activity' => NULL,
```
DFurnes, line 213: Should we add a `@TODO` so we remember this is just a placeholder (and not something that's filled in by a hook elsewhere in the codebase)?

weerd, line 213: oh ya!

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -697,7 +697,41 @@ function dosomething_helpers_get_current_user() {
...
+  $northstar_response = json_decode($northstar_response);
+
+  if ($northstar_response && isset($northstar_response->error)) {
+    return NULL;
+  }
+
+  $northstar_user = $northstar_response->data;
+
+  return [
+    'id' => dosomething_helpers_isset($northstar_user, 'id'),
```
DFurnes: Would it make sense to use `data_get` here?

DFurnes:
Two subjective suggestions. The rest looks good to me! 👍

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
weerd: oh yeah. Should we just start phasing out `dosomething_helpers_isset()`?

DFurnes: Definitely not necessary for merge, but might make sense to mark `dosomething_helpers_isset` as deprecated and get out of the habit of using it now that we have the battle-tested tools included in the Collect package.

DFurnes: Hah, was typing that follow-up as you commented. Yeah I think that's a good idea! 😄

weerd: sounds good to me!

DFurnes:
🚢!

PR #6711, opened 2016-07-11
Fix error with getting user info if Northstar module not enabled.

DFurnes:
#### What's this PR do?

This fixes a white-screen on the campaign page if the `dosomething_northstar` module is disabled.

If the `dosomething_northstar` module is disabled, you should not see a white screen on the campaign page.

We disable the `dosomething_northstar` module when pulling the database from staging, so this function is not reliably defined in local environments. 
#### Relevant tickets

Fixes 💥.
#### Checklist
- [ ] Tested on staging.


weerd:
👍

PR #6714, opened 2016-07-12
Don't throttle queries

### LINKED ISSUE #6581: Able to select multiple search filters at once
#### Current Behavior

-Currently able to select multiple causes/times/types on campaign search
#### Desired Behavior
- Should only be able to select item from each list
#### Steps to Replicate
1. navigate to home page
2. click cause
3. quickly click on different categories. observe multiple categories highlighted. 
4. repeat for other time & type
#### Why This Matters

-Selecting multiple items is not currently a feature we offer. 
#### OS/Browser (optional)

Safari & Chrome

screen shot 2016-06-09 at 1 45 10 pm

sbsmith86:
#### What's this PR do?

Switched from using the `throttledQuery()` method in `SolarAdapter.js` to just use the `query()` method directly.

The `throttledQuery()` method set a delay of 400ms before sending a query to solr. Because we disable fields once we get results back from solr, the 400ms delay was allowing a user to select multiple fields within that short amount of time. 

Using the `query()` method directly removes that delay so multiple selections can not be made. 

Use the campaign finder and try to select multiple filters very quickly, you should not be able to.

I opted to just use a new method instead of removing the throttle method entirely. If we every want to allow multiple selections, then this could be useful. Open to peoples feelings about just yanking that code, but this seemed easy enough.
#### Relevant tickets

Fixes #6581 
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/finder/Finder.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/finder/Finder.js
@@ -53,10 +53,10 @@ define(function(require) {
       // Run the Solr query
       if(!queryOffset) {
         // If no offset, we display a new set of results
-        SolrAdapter.throttledQuery(checkedFields, queryOffset, Finder.displayResults);
+        SolrAdapter.query(checkedFields, queryOffset, Finder.displayResults);
       } else {
         // Otherwise, append to existing result set
-        SolrAdapter.throttledQuery(checkedFields, queryOffset, ResultsView.appendResults);
+        SolrAdapter.query(checkedFields, queryOffset, ResultsView.appendResults);
       }
 
       // Mark the results pane as loading


DFurnes:
Removing the throttled version of the query method makes sense to me. Are users able to select results while the network request is being made (like on a slow connection or if it times out)?

sbsmith86:
@DFurnes just did some lightweight network throttling and saw some delay on a 2G connection that would allow a user to select multiple things and if connection was completely lost then buttons would never be disabled. These seem like edge cases to me?

DFurnes:
Yeah, I think ideally we could make everything unclickable while a request is in progress but I agree that's an edge case. Let's merge this and see if we still have people reporting issues? 🚢

sbsmith86:
Sounds good!

PR #6715, opened 2016-07-12
Cleanup up code and naming and setting up styles.

weerd:
#### What's this PR do?

This PR does a little bit of cleanup for the initial work on the Onboarding experience and sets up the start of the CSS styles.

Some quick 👀  to make sure all looks swell.

Switched up slightly how the Onboarding component is attached to the DOM, so instead of prepending it as the first item in the `.chrome` element, it places it right before the `.wrapper` in the `.chrome` which works a bit better when viewing the page as an admin w/ the Drupal admin bar and admin nav.
#### Relevant tickets

Refs #6682
#### Checklist
- [ ] Tested on staging.

---

@DFurnes 
cc @angaither


lib/themes/dosomething/paraneue_dosomething/js/components/ContextSlide.js
```diff
@@ -1,5 +1,9 @@
 const React = require('react');
 
+/**
+ * ContextSlide Component
+ * <ContextSlide />
+ */
 class ContextSlide extends React.Component {
   constructor(props) {
```
DFurnes, line 8: You can exclude the `constructor` if you're not overriding anything (e.g. setting initial state, etc).

lib/themes/dosomething/paraneue_dosomething/js/lab.js
```diff
@@ -0,0 +1,24 @@
+import Onboarding from './components/Onboarding';
+import ContextSlide from './components/ContextSlide';
+
+const $ = require('jquery');
+const React = require('react');
+const ReactDom = require('react-dom');
+
+/**
+ * Welcome to the Laboratory.
```
DFurnes, line 9: Haha, I love it. 😈 💥

lib/themes/dosomething/paraneue_dosomething/js/lab.js
```diff
@@ -0,0 +1,24 @@
+import Onboarding from './components/Onboarding';
+import ContextSlide from './components/ContextSlide';
+
+const $ = require('jquery');
+const React = require('react');
+const ReactDom = require('react-dom');
```
DFurnes, line 6: Hmm, I totes get making this camel-cased for consistency but it looks so weird. I could get used to it. 💭

DFurnes:
Looks good to me! Ship it! ⛵

lib/themes/dosomething/paraneue_dosomething/js/components/ContextSlide.js
weerd, line 8: It was in there from before... Might leave for now if that's cool, since I didn't delve into building farther on that yet. Not sure if I will have to or not.

lib/themes/dosomething/paraneue_dosomething/js/lab.js
weerd, line 6: yeah, I hear ya. I've forced myself to like it for the sake of simplicity, and consistency in other projects 😖

lib/themes/dosomething/paraneue_dosomething/js/components/ContextSlide.js
DFurnes, line 8: Yep, that's totes fine by me.

weerd, line 8: but good to know! ;)

PR #6720, opened 2016-07-14
Update for creating multiple slides and passing props to each slide.

### LINKED ISSUE #6721: Allow passing data props to new component slides
When creating a new component slide (or multiple slides) in the onboarding/slideshow component, need to be able to pass along some dynamic properties pertaining to the current campaign and current user through to each slide.

weerd:
#### What's this PR do?

This PR allows creating multiple slide components while also passing along props with data regarding the user and current campaign to each slide.

Does the code 👀  reasonable?
#### Relevant tickets

Fixes #6721
#### Checklist
- [ ] Tested on staging.

---

@DFurnes @deadlybutter

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/Slideshow.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/Slideshow.js
@@ -1,6 +1,7 @@
 import SlideshowController from './SlideshowController'
 
 const React = require('react');
+const ReactDom = require('react-dom');
 
 /**
  * Slideshow Component
@@ -13,16 +14,28 @@ class Slideshow extends React.Component {
     this.state = {
       activeSlide: 0
     };
+
+    this.renderSlide = this.renderSlide.bind(this);
   }
 
   moveSlide() {
     //
   }
 
+  renderSlide(slide) {
+    console.log('a slide');
+  }
+
   render() {
+    const user = this.props.user;
+    const campaign = this.props.campaign;
+
     return (
       <div className="slideshow">
-        {this.props.slides[this.state.activeSlide]}
+        {/* Why does this.renderSlide not work? */}
+        {this.props.slides.map(function (slide, index) {
+          return React.createElement(slide, {key: index, campaign: campaign, user: user});
+        })}
 
         <SlideshowController moveSlide={this.moveSlide}/>
       </div>
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/Slideshow.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/Slideshow.js
@@ -23,9 +23,13 @@ class Slideshow extends React.Component {
     const user = this.props.user;
     const campaign = this.props.campaign;
 
+    const slides = this.props.slides.map((slide, index) => {
+      return React.createElement(slide, {key: index, campaign: campaign, user: user});
+    });
+
     return (
       <div className="slideshow">
-        {React.createElement(this.props.slides[this.state.activeSlide], {campaign: campaign, user: user})}
+        {slides[this.state.activeSlide]}
 
         <SlideshowController moveSlide={this.moveSlide}/>
       </div>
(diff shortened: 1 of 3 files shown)


lib/themes/dosomething/paraneue_dosomething/js/components/Slideshow.js
```diff
@@ -1,6 +1,7 @@
 import SlideshowController from './SlideshowController'
 
 const React = require('react');
+const ReactDom = require('react-dom');
```
DFurnes: Is this being used?

itsjoekent: Why did we re-require this? I don't see it used anywhere

lib/themes/dosomething/paraneue_dosomething/js/components/Slideshow.js
```diff
@@ -20,9 +21,16 @@ class Slideshow extends React.Component {
...
+    const campaign = this.props.campaign;
+
+    const slides = this.props.slides.map((slide, index) => {
+      return React.createElement(slide, {key: index, campaign: campaign, user: user});
+    });
+
     return (
       <div className="slideshow">
-        {this.props.slides[this.state.activeSlide]}
+        {slides}
```
itsjoekent: I'm not sure how I feel about this. The reason I had it the previous way was, if you're on slide 1, you only need to render slide 1. With this it seems like we would rerender the entire slideshow every state change, which seems very expensive with no real upside to it.

lib/themes/dosomething/paraneue_dosomething/js/components/Slideshow.js
weerd: Shit. Left over form experiment. Sorry. :facepalm:

lib/themes/dosomething/paraneue_dosomething/js/components/Slideshow.js
itsjoekent: talked with @weerd , came to the suggestion we can keep `{slides}` in the render loop so all of the React elements are available on the page for CSS styling. however the `const slides` should be moved into the constructor and set as a property of the Slideshow, rather than re-creating every slide on every `render()`. @DFurnes thoughts?

DFurnes: I think it makes sense to only render the visible slide into the Virtual DOM, so we can use [ReactCSSTransitionGroup](https://facebook.github.io/react/docs/animation.html) to handle transitions and our interface stays declarative.

lib/themes/dosomething/paraneue_dosomething/js/components/Slideshow.js
```diff
@@ -20,9 +20,12 @@ class Slideshow extends React.Component {
   }
 
   render() {
+    const user = this.props.user;
+    const campaign = this.props.campaign;
+
     return (
       <div className="slideshow">
-        {this.props.slides[this.state.activeSlide]}
+        {React.createElement(this.props.slides[this.state.activeSlide], {campaign: campaign, user: user})}
```
DFurnes: I think you still want to create each of the slides with a unique key (like you had before, but then just grabbing the appropriate slide) so React knows about them all when transitioning things into and out of the DOM.

weerd: oh.... 😖

itsjoekent: I honestly made this mistake for a while at first, I didn't even know key was a thing you had to do until I read it in a blog post. I think this is one of those gotchas that isn't explained well

DFurnes:
:ship:

lib/themes/dosomething/paraneue_dosomething/js/components/Slideshow.js
weerd: yeah... just a very different way to think about things from what I'm used to.

PR #6728, opened 2016-07-18
add goal fields to custom settings form

### LINKED ISSUE #6699: Add signup and impact goal fields on runs
#### User Story

As an admin, I want the option to add signup goals, new member goals, and impact goals for each campaign. 

These should be three backend fields that are non-required. 
#### Why This Matters

Pulling these numbers directly from Phoenix will give us on-demand access to see whether or not each campaign is meeting their signup, member, and impact goals through our new visualization tool. The current process is not streamlined and leaves more room for human error.

sbsmith86:
#### What's this PR do?

Adds field to the custom settings page of a campaign that allows admins to set sign up and impact goals. 

image

Go to a campaign, go to the custom settings page, and test that you can set goals. 

I put these fields on the custom settings page because they are not supposed to be translatable and this seems like a more "global" (excuse the term) way of setting these goals across all translations. See the discussion on #6699 about not making these fields translatable. 
#### Relevant tickets

Fixes #6699 
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.variable.inc
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.variable.inc
@@ -86,6 +86,42 @@ function dosomething_helpers_variable_form($form, &$form_state, $node) {
     '#default_value' => $vars['signup_form_submit_label'],
     '#size' => 20,
   );
+  $form['campaign_goals'] = [
+    '#type' => 'fieldset',
+    '#title' => t('Campaign Goals'),
+    '#collapsible' => TRUE,
+    '#collapsed' => FALSE,
+  ];
+  $form['campaign_goals']['external_signup_goal'] = [
+    '#type' => 'textfield',
+    '#title' => t('External Signup Goal'),
+    '#default_value' => $vars['external_signup_goal'],
+  ];
+  $form['campaign_goals']['external_new_member_goal'] = [
+    '#type' => 'textfield',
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_helpers/dosomething_helpers.variable.inc
```diff
@@ -227,7 +263,13 @@ function dosomething_helpers_get_variable_names() {
     'count_flagged',
     'count_pending',
     'count_promoted',
+    'external_signup_goal',
+    'external_new_member_goal',
+    'external_impact_goal',
     'goal',
```
angaither: I know you didn't write this line (I probably did) but this can be removed since this was hot module goals

sbsmith86: 👍

angaither:
this looks 👍  my concern about these being custom settings is I think the data team is the peeeps who want this info, so we will need to explain how to get custom settings ;)

sbsmith86:
@angaither ya i was thinking about that... it just seemed more intuitive from the admin perspective to put it there.

angaither:
@sbsmith86 totes

🚢

angaither:
OH BUT WAIT

these are on campaigns, not runs? we want them to be on runs, me thinks

sbsmith86:
@angaither fuck i moved to fast...reopened the issue and will set it based on run....

PR #6731, opened 2016-07-18
Update to expose current campaign data to Drupal JS object.

### LINKED ISSUE #6730: Expose Campaign infomation to Drupal JS object
The React components need access to current campaign data on the pitch/action pages exposed via the `Drupal.settings` object.

weerd:
#### What's this PR do?

This PR updates the `paraneue_dosomething_preprocess_node_campaign()` function to expose current campaign data for pitch and action pages in the `Drupal.settings` object as a `dsCampaign` property.

Load a few campaigns and check if you see the `dsCampaigns` property with data regarding the current campaign.
#### Relevant tickets

Fixes #6730
#### Checklist
- [ ] Tested on staging.

---

@DFurnes @deadlybutter 

cc: @jessleenyc

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
@@ -112,7 +112,6 @@ function paraneue_dosomething_add_info_bar(&$vars) {
  * @see paraneue_dosomething_preprocess_node().
  */
 function paraneue_dosomething_preprocess_node_campaign(&$vars) {
-
   // List of partials to add as $vars.
   $partials = array(
     'campaign_creator',
@@ -228,6 +227,14 @@ function paraneue_dosomething_preprocess_node_campaign(&$vars) {
     $vars['show_persistent_signup'] = theme_get_setting('show_persistent_signup');
   }
 
+  // Get prepared Campaign data.
+  try {
+    $current_campaign = Campaign::get($vars['nid']);
+  }
+  catch (Exception $error) {
+    $current_campaign = NULL;
+  }
+
   drupal_add_js(
     [
       'dsUser' => [
@@ -235,7 +242,8 @@ function paraneue_dosomething_preprocess_node_campaign(&$vars) {
         // @TODO: add a way to include user activity if it is a feature that is needed.
         // Maybe a function like: dosomething_helpers_get_current_user_activity();
         'activity' => NULL,
-      ]
+      ],
+      'dsCampaign' => $current_campaign,
     ],
     'setting'
   );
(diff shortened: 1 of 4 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
```diff
@@ -228,14 +227,23 @@ function paraneue_dosomething_preprocess_node_campaign(&$vars) {
     $vars['show_persistent_signup'] = theme_get_setting('show_persistent_signup');
   }
 
+  // Get prepared Campaign data.
+  try {
+    $current_campaign = Campaign::get($vars['nid']);
```
DFurnes, line 232: Does this re-fetch the campaign from the database? If so, would it make sense to have a way to "hydrate" a Drupal node into a `Campaign` class without hitting the DB another time?

weerd, line 232: From what I've come to understand, if Drupal already loaded the node, it's cached and not re-fetched from the DB, so I figured it wouldn't do it harm. I was thinking about that too, but I think it's supposed to be ok.

Maybe "hydrating" is something to consider for a future refactor?

lib/themes/dosomething/paraneue_dosomething/js/lab.js
```diff
@@ -15,10 +15,7 @@ $(document).ready(function() {
   if (Drupal.settings.dsOnboarding.enabled) {
     const jsxElementHook = 'jsx-onboarding';
     const slides = [ContextSlide, ContextSlide] // @TODO: create other slide components.
-    const campaign = {
-      // @TODO: switch to use Drupal.settings.dsCampaign once it is available.
-      title: 'Awesome Campaign Placeholder'
-    }
+    const campaign = Drupal.settings.dsCampaign || null;
```
itsjoekent, line 18: Not sure how others feel about this, but i generally prefer using `undefined` over `null` in :js:

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
DFurnes, line 232: Okay, let's keep an eye on it. 👓

lib/themes/dosomething/paraneue_dosomething/js/lab.js
DFurnes, line 18: I think `null` makes more sense here, since it's explicitly stating that the variable exists but is empty. I think `undefined` should generally be reserved for undefined variables.

weerd, line 18: I agree. `undefined` just means a variable was declared but is not assigned yet. I'm specifically assigning a `null` value to indicate there is no campaign data available.

DFurnes:
👍 🚢

PR #6733, opened 2016-07-18
Campaign goals per run

### LINKED ISSUE #6699: Add signup and impact goal fields on runs

sbsmith86:
#### What's this PR do?

Adds fields on the campaign run node that allow admins to set signup and impact goals.

NOTE: I'm not sure why there are updates to the `field_description_field` but it looks like it just moved where it was in the feature export file. I didn't make any updates to that field. 

Edit a campaign run and check out the new fields. 
#### Relevant tickets

Fixes #6699 
#### Checklist
- [x] Tested on staging.


angaither:
> NOTE: I'm not sure why there are updates to the `field_description

You need to run `drush fra -y` before because that field will be removed with this

sbsmith86:
@angaither ok i don't know what i'm doing wrong here. I rebuilt my dev machine to reset everything, I ran `drush fra -y`, I added the new goal fields and did `drush fu dosomething_campaign_run` and I get the same diff. 

I see that the description field is removed from `dosomething_campaign_run.features.field_base.inc` but it does seem to still be present in `dosomething_campaign_run.features.field_instance.inc` (just moved up to the beginning of the file. Doesn't that mean that the field still exists?

angaither:
@sbsmith86 just checked on prod and it seems that feature is overridden there as well, so it's not a thing you're doing wrong, features is acting whacky.

If we do a refresh thor/prod and you deploy this there, we can test to make sure the description field doesn't get removed & then clean up the features

sbsmith86:
@angaither ok so merge and deploy this to staging?

angaither:
yaaa

Issue #6662, opened 2016-06-30
Display onboarding on a campaign signup


angaither:
#### User Story

As a user, after I signup for a campaign I would like to see onboarding message box 
#### Additional Information (optional)

This should happen the first time and only the first time the authenticated user sees the action page.
This should appear on every single signup the user makes

sbsmith86:
@angaither is this still a thing?

angaither:
@sbsmith86 I mostly assigned this to you to see if any of the functionality you and @chloealee created could work for this

sbsmith86:
@angaither ahh i can work on it...but wasn't sure how it fit in with the way @weerd and @deadlybutter were building this, since it seems they are rebuilding from the bottom up.

angaither:
@sbsmith86 nah, that's fine I'll unassign you.

PR #6734, opened 2016-07-19
Adding current user signups to Drupal JS object.

weerd:
#### What's this PR do?

This PR exposes current user campaign signups to the `Drupal.settings.dsUser` JS object for use in front-end work. It provides the signups as an array of ids as a property, `Drupal.settings.dsUser.activity.signups`.

Load a campaign action page and see if the above property on the Drupal JS object contains signup IDs. If not authenticated, then the value for `Drupal.settings.dsUser.activity` is `NULL`.

This will be useful for the Onboarding component feature.
#### Relevant tickets

Refs #6662
#### Checklist
- [ ] Tested on staging.

---

@DFurnes @deadlybutter @angaither 

cc: @jessleenyc


itsjoekent:
🍨

PR #6736, opened 2016-07-19
Adds post request on RB review

itsjoekent:
#### What's this PR do?

Sends a post request to Vision when a RB is reviewed

Does it send the request successfully?
Is the payload ok?

🏃 ::message-broker: 
(thats me running away)
#### Relevant tickets

Fixes #SNACK_WEEK
#### Checklist
- [ ] Test on staging but leave it off afterwards

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/includes/ReportbackItem.php
+++ b/lib/modules/dosomething/dosomething_reportback/includes/ReportbackItem.php
@@ -198,6 +198,17 @@ class ReportbackItem extends Entity {
       $this->deleteFile();
     }
 
+    // Send external message request
+    $vision_url = variable_get('dosomething_settings_vision_url');
+    if (isset($vision_url)) {
+      $options = [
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_reportback/includes/ReportbackItem.php
```diff
@@ -198,6 +198,17 @@ public function review($values) {
       $this->deleteFile();
     }
 
+    // Send external message request
+    $vision_url = variable_get('dosomething_settings_vision_url');
```
angaither, line 202: this should probably be in it's own function

angaither:
this is fine, but let's not merge stuff from snack week.
if you wanna continue branching off of this branch and make PRs against that, that would be 👍

itsjoekent:
sounds like branch mania

angaither:
global 2.0

PR #6738, opened 2016-07-20
Added mapping for signup badges.

angaither:
#### What's this PR do?

Added in some mapping for signup badges.
Things are named terrrrrribly.
But it works, mostly.
Enjoy.

it reallllly shouldn't


katiecrane:
👍 💖

PR #6741, opened 2016-07-21
Moving to array_slice to grab all the badges.

angaither:
#### What's this PR do?
- using `array_slice` to get from 1-howevermanythingsthepersonhas
- use handler functions to grab badges of each type. 

👀 

This seems a lot less wild than using a range, then array mapping each badge onto the correct section.


katiecrane:
👀 🍗 👍

PR #6744, opened 2016-07-21
Badges with real badges

angaither:
#### What's this PR do?
- adds in the real assets to the theme
- this breaks the emoji display because they are printing images, so we need a default now.
- not all of them are named with standards in mind

👀


angaither:
oh, also the files are probably too damn large and we will want to make them smalllller before actually using them 😉

katiecrane:
🎨 👍

PR #6754, opened 2016-07-25
Prevents MX & BR users from getting the onboarding flow

### LINKED ISSUE #6663: Suppress onboarding experience for global users
#### User Story

Do not display onboarding messaging if a user comes from `/mx/` `/br/`

This should only be displayed for `/us` and the global english experience
#### Why This Matters

It's super wonky to display this as english text on non-english campaigns

itsjoekent:
#### What's this PR do?

_what the title says_

do you get onboarding flow as US user?
what about MX/BR?

nope
#### Relevant tickets

Fixes #6663 
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
@@ -13,9 +13,10 @@ const ReactDom = require('react-dom');
 $(document).ready(function() {
   const campaign = Drupal.settings.dsCampaign || null;
   const user = Drupal.settings.dsUser || null;
+  const disabled_langs = ['mx', 'br']
   let localStorageKey = null;
 
-  if (campaign && user) {
+  if (campaign && user && disabled_langs.indexOf(campaign.language.language_code) == -1) {
     localStorageKey = `onboarding:campaign=${campaign.id}-user=${user.info.id}`;
   }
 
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
```diff
@@ -13,9 +13,10 @@ const ReactDom = require('react-dom');
 $(document).ready(function() {
   const campaign = Drupal.settings.dsCampaign || null;
   const user = Drupal.settings.dsUser || null;
+  const disabled_langs = ['mx', 'br']
   let localStorageKey = null;
 
-  if (campaign && user) {
+  if (campaign && user && disabled_langs.indexOf(campaign.language.language_code) == -1) {
```
DFurnes: Could you pull this into a named boolean? Otherwise this conditional is pretty heavy to read.

``` js
const isDisabledForLanguageCode = ['mx', 'br'].indexOf(campaign.language.language_code) == -1;

if (campaign && user && isDisabledForLanguageCode) {
  // ...
```

itsjoekent: works for me

DFurnes: Also does this mean the on-boarding experience will only ever be disabled by that local storage key if the user has one of the language codes? Should the conditional be `|| isDisabledForLanguageCode` instead?

itsjoekent: The logic im about to push up is,

localStorageKey = null
if the page has the campaign data, user data, and they're not MX/BR
set the storage key variable

DFurnes: Oh okay, I just looked at the rest of the file and I see how it's being used.

itsjoekent:
@DFurnes updated, i think this should be ok

DFurnes:
👍

PR #6770, opened 2016-07-27
Removes spaces before <?php tag.

sergiitk:
#### What's this PR do?

Removes spaces before `<?php` tag.
#### Why it matters

Everything before `<?php` tag is treated as a template and will be printed out to the output buffer.
If HTTP headers are issued after that, they won't work.


angaither:
🚢

PR #6771, opened 2016-07-27
Checks if modal is open before assigning key

itsjoekent:
#### What's this PR do?
- Check if a modal is open before assigning a local storage key.
- Otherwise if the modal triggers a page refresh the onboarding flow is tricked & thinks you already saw it, meanwhile you didn't.
- Adds an onboarding version to the key.

Setup a campaign with a competition modal, check if onboarding works after page refresh.
Confirm normal signups also still work

The onboarding version is used so we can make sure users see onboarding again in the future if we make a major change to it (eg: new slide).
#### Relevant tickets

Fixes #hm_nope
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
@@ -8,6 +8,8 @@ const React = require('react');
 const ReactDom = require('react-dom');
 const Modal = require('dosomething-modal');
 
+const ONBOARDING_VERSION = '1.0';
+
 /**
  * Welcome to the Laboratory.
  * Run your experiments below.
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
```diff
@@ -17,7 +18,7 @@ $(document).ready(function() {
   const disabledByLanguage = ['mx', 'br'].indexOf(campaign.language.language_code) != -1;
   let localStorageKey = null;
 
-  if (campaign && user && !disabledByLanguage) {
+  if (campaign && user && !disabledByLanguage && !Modal.isOpen()) {
```
itsjoekent, line 23: This conditional is getting long, not sure if we should break it out into a function of its own or just leave it

DFurnes:
Okie dokie. ⛵

PR #6787, opened 2016-08-01
Onboarding fixes and tweaks

### LINKED ISSUE #6763: Add Hover Color For 'X'
#### User Story

As a user, I want to be indicated that the 'x' is clickable.
#### Why This Matters

Other smasher notifications have 'x' buttons change opacity upon hover. We should keep this consistent.

weerd:
#### What's this PR do?

This PR does an update to remove the CSS hack used for the instruction video, in favor of the `media-video` pattern from Forge. It also adds a hover color change effect on the close button in the onboarding feature.

use some 👀 
#### Relevant tickets

Fixes #6763
#### Checklist
- [ ] Tested on staging.

---

@DFurnes @deadlybutter 

cc: @jessleenyc

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/InstructionSlide.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/InstructionSlide.js
@@ -27,9 +27,11 @@ class InstructionSlide extends React.Component {
 
             <div className="container__row">
               <div className="container__block -narrow -centered">
-                <div className="beta-card">
-                  <video className="beta-card__video" poster="https://static.dosomething.org/onboarding/poster.png" src="https://static.dosomething.org/onboarding/CouldBeYou_720p.mp4" loop autoPlay controls width="100%"></video>
-                </div>
+                {/*<div className="beta-card">*/}
+                  <div className="media-video">
+                    <video poster="https://static.dosomething.org/onboarding/poster.png" src="https://static.dosomething.org/onboarding/CouldBeYou_720p.mp4" loop autoPlay controls width="100%"></video>
+                  </div>
+                {/*</div>*/}
               </div>
             </div>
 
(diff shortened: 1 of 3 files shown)


lib/themes/dosomething/paraneue_dosomething/js/components/InstructionSlide.js
```diff
@@ -27,8 +27,8 @@ class InstructionSlide extends React.Component {
 
             <div className="container__row">
               <div className="container__block -narrow -centered">
-                <div className="beta-card">
-                  <video className="beta-card__video" poster="https://static.dosomething.org/onboarding/poster.png" src="https://static.dosomething.org/onboarding/CouldBeYou_720p.mp4" loop autoPlay controls width="100%"></video>
+                <div className="media-video">
```
weerd, line 30: We didn't really need the `beta-card` here _just_ for the border-radius. Chatted with @lkpttn and looks like we want to use border-radius moving forward on a lot of our "media" so maybe we add it to the `media-*` patterns. Left a `@TODO` in the comments.

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_card.scss
```diff
@@ -52,11 +52,6 @@
   overflow: hidden;
   background-color: $white;
 
-  .beta-card__video {
-    // The video element incorrectly sits in the beta card, quick hack to fix it. 
-    margin-bottom: -10px;
```
weerd, line 57: @deadlybutter FYI don't need this anymore, but the issue wasn't a margin thing... you just needed to make the video display as block instead of the default of inline. Adding `display: block` also fixes this and in a slightly better way since it's not messing w/ margin which could affect other items. Just an FYI for future reference 😄

lib/themes/dosomething/paraneue_dosomething/js/components/InstructionSlide.js
itsjoekent, line 30: gotcha 👍

DFurnes:
👍 ⛵

PR #6797, opened 2016-08-01
add missing photo modal to profile and campaign page

katiecrane:
#### What's this PR do?

Added modals telling members about the image situation on action pages and user profiles. Members can access modals from the `You Did` section of their profiles and the `Prove It` section of campaigns that they have reported back on.

Make sure to check as a non-admin as well.

`1.` Go to your profile, and scroll down to the `You Did` section (this section should only appear if a user has reported back before). It should look like this:

pasted image at 2016_08_01 04_55 pm-1

`2.` Click the link and make sure the modal looks okay.

pasted image at 2016_08_01 04_55 pm

`3.` Go to a campaign that you have reported back on. Scroll down to `Prove It` and you should see this:

pasted image at 2016_08_01 04_55 pm-2

`4.` Click on that link and make sure you see the same modal.

`5.` Go to a campaign you have NOT reported back on and make sure you don't see that link.

[`dosomething_reportback_exists($campaign->nid)`](https://github.com/DoSomething/phoenix/blob/743c97e0d37360739c63d731e3a078e6fb599499/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module#L669-L711) is being used to determine whether or not the current user has a reportback for the campaign. We are not specifying a run which means it checks the current run. I *think* that is what we want, but we can change it.
#### Relevant tickets

we skipped this part as far as I know
#### Checklist
- [x] Tested on staging.

cc: @mikefantini @lkpttn

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
@@ -331,6 +331,9 @@
             <?php if (isset($reportback_copy)): ?>
               <p class="copy inline-alt-text-color"><?php print $reportback_copy; ?></p>
             <?php endif; ?>
+            <?php if(dosomething_reportback_exists($campaign->nid)): ?>
+              <a href="#" data-modal-href="#modal-missing-photos">Is your photo not showing up?</a>
+            <?php endif; ?>
           </div>
       </div>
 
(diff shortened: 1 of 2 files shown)


mikefantini:
@katiecrane two copy edits:

First:

Let's change "Last week" to "The other day"

Second:

Can we we please have the bulleted list under good news be:
- We’ve fixed the issue so new photos will not be deleted.
- Even though your photo isn’t showing up, your impact is still being counted.
- And if the campaign had a scholarship opportunity, don't worry! Your entry into the scholarship was still counted.

Also did you and @lkpttn play around with placement of this in the profile? Does it make sense to put it below the campaigns in the You Did section? Or does this compete with Show More for users who have reported back on 4 or more campaigns?

katiecrane:
@mikefantini I'm not sure if @lkpttn played around with placement beforehand, but "Show More" is at the bottom of the section and this new message is at the top. Placing it in this section also puts it as close as possible to where any missing images should have showed up. We can really easily drop that link anywhere in the profile though, just my $0.02 😄

katiecrane:
here's the current version:

image

lkpttn:
@mikefantini @katiecrane Yeah I think this placement is appropriate. It shows up in proximity to where users would encounter the error and won't interfere with the reportback images, regardless of how many they have.

DFurnes:
👍 from me!

angaither:
only note is if the text is the same it should maybe be an include
🚢

PR #6800, opened 2016-08-02
Fix for video getting cut off.

### LINKED ISSUE #6790: Fix &ndash; html entity not being escaped
image

The text is outputting the string instead of the html entity.

### LINKED ISSUE #6791: Full Video Not Appearing
#### Current Behavior

Looks like the bottom portion of the video frame is cut off. Also, a gray bar (that sort of looks like part of a scroll bar or the top part of player controllers) appear upon hover at the bottom of the video. See screenshot below. 
#### Desired Behavior

The full video and its border should be present. 
#### Steps to Replicate

1) Navigate to thor. 
2) Sign up for a campaign you haven't signed up for previously.

screen shot 2016-08-01 at 2 38 54 pm

weerd:
#### What's this PR do?

This PR fixes an issue with the onboarding video for the instruction slide where the bottom part of the video, including the controls, were getting cut off. It also fixes the html entity for `&ndash;` and changes it to `\u2013` so it renders properly within the JS string.

👀 

The video was getting cut-off because the `media-video` pattern which is very useful for displaying responsive video across browsers and mobile is dependent on a typical ratio for a video, similar to that of a video on youtube or vimeo. The video we are using has a different aspect ratio and thus the % padding that helps define the ration was incorrect.
#### Relevant tickets

Fixes #6791
Fixes #6790
#### Checklist
- [ ] Tested on staging.

---

@DFurnes @deadlybutter 

cc: @jessleenyc

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/InstructionSlide.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/InstructionSlide.js
@@ -1,7 +1,7 @@
 const React = require('react');
 
 const title = "This is how you do it.";
-const body = "We are PUMPED to see your photos. Complete these simple steps and upload a photo of yourself in action. Your pics will inspire others to join the movement &ndash; so don't forget to share them!";
+const body = "We are PUMPED to see your photos. Complete these simple steps and upload a photo of yourself in action. Your pics will inspire others to join the movement \u2013 so don't forget to share them!";
 
 /**
  * InstructionSlide Component
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
@@ -13,6 +13,7 @@
   // so maybe we add the border-radius to all media-* classes?
   & .media-video {
     border-radius: 4px;
+    padding-bottom: 60.4%;
   }
 }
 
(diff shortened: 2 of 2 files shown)


lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
```diff
@@ -13,6 +13,7 @@
   // so maybe we add the border-radius to all media-* classes?
   & .media-video {
     border-radius: 4px;
+    padding-bottom: 60.4%;
```
DFurnes, line 19: Hmm, can you add a comment for this? Also makes me wonder if we make this calculation a mixin.

weerd, line 19: Yeah, was thinking there might be a better approach. Will add comment for now to get this shipped.

DFurnes:
👍

lib/themes/dosomething/paraneue_dosomething/js/components/InstructionSlide.js
```diff
@@ -1,7 +1,7 @@
 const React = require('react');
 
 const title = "This is how you do it.";
-const body = "We are PUMPED to see your photos. Complete these simple steps and upload a photo of yourself in action. Your pics will inspire others to join the movement &ndash; so don't forget to share them!";
+const body = "We are PUMPED to see your photos. Complete these simple steps and upload a photo of yourself in action. Your pics will inspire others to join the movement \u2013 so don't forget to share them!";
```
DFurnes, line 4: Oh sorry, just noticed this when the issue was closed. Why `\u2013` rather than just `–`?

PR #6803, opened 2016-08-02
Onboarding  slides feature flags

### LINKED ISSUE #6788: Feature Flag Each Screen
#### User Story

As an admin, I want all screens to have a feature flag that we can toggle as we build. 
- [x] toggle screen 1
- [x] toggle screen 2
- [x] toggle screen 3
#### Why This Matters

We don't want users to see parts of onboarding that isn't production-ready.

weerd:
#### What's this PR do?

This PR allows toggling each slide as a feature flag in the CMS. It applies to all campaigns, so it's not something to control on a campaign by campaign basis (just FYI). 

Make sure you see the toggles in the `.../admin/config/dosomething/dosomething_campaign` settings page and as you toggle items, they show up as expected. Keep in mind we don't have pagination working on the Onboarding slideshow just yet, so maybe just toggle one on and off as you test 😉 

image

@DFurnes @deadlybutter in the `Drupal.settings.dsOnboarding.slides` array, I have each activated slide as a string. It'd be way cleaner if I could somehow just pass this array and set it as the value for `slides` variable, but since the component names are strings in the array, they can't call the associated class. Do either of you know if there's a way to use a string to somehow call a class in JS? I was playing around with this, but couldn't get a suitable solution working.
#### Relevant tickets

Fixes #6788 
#### Checklist
- [ ] Tested on staging.

---

@DFurnes @deadlybutter 

cc: @jessleenyc

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
@@ -28,13 +28,9 @@ $(document).ready(function() {
 
   // Enable the Onboarding experiment.
   if (Drupal.settings.dsOnboarding.enabled && localStorage.getItem(localStorageKey) !== 'true') {
-    const slideComponents = {ContextSlide, InstructionSlide, ReportbackItemKudosSlide};
     const jsxElementHook = 'jsx-onboarding';
-    let slides = [];
-
-    slides = Drupal.settings.dsOnboarding.slides.map(function (slide) {
-      return slideComponents[slide];
-    });
+    const slideComponents = {ContextSlide, InstructionSlide, ReportbackItemKudosSlide};
+    const slides = Drupal.settings.dsOnboarding.slides.map(name => slideComponents[name]);
 
     if (slides.length > 0) {
       $('.chrome').find('> .wrapper').before(`<div id="${jsxElementHook}"></div>`);
--- a/lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
@@ -28,21 +28,12 @@ $(document).ready(function() {
 
   // Enable the Onboarding experiment.
   if (Drupal.settings.dsOnboarding.enabled && localStorage.getItem(localStorageKey) !== 'true') {
+    const slideComponents = {ContextSlide, InstructionSlide, ReportbackItemKudosSlide};
     const jsxElementHook = 'jsx-onboarding';
     const slides = [];
 
     Drupal.settings.dsOnboarding.slides.map(function (slide) {
-      if (slide === 'ContextSlide') {
-        slides.push(ContextSlide);
-      }
-
-      if (slide === 'InstructionSlide') {
-        slides.push(InstructionSlide);
-      }
-
-      if (slide === 'ReportbackItemKudosSlide') {
-        slides.push(ReportbackItemKudosSlide);
-      }
+      slides.push(slideComponents[slide]);
     });
 
     if (slides.length > 0) {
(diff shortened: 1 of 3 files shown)


lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
```diff
@@ -29,16 +29,32 @@ $(document).ready(function() {
   // Enable the Onboarding experiment.
   if (Drupal.settings.dsOnboarding.enabled && localStorage.getItem(localStorageKey) !== 'true') {
     const jsxElementHook = 'jsx-onboarding';
-    const slides = [InstructionSlide]; // ReportbackItemKudosSlide, ContextSlide
+    const slides = [];
```
weerd: @DFurnes should this be `let`? I wasn't getting any errors. I'm not changing the value from an array to some other type, just adding to the items... so I wasn't sure.

DFurnes: Yeah it's kinda confusing. You can safely use `const` here since you're not changing the binding of what variable is assigned to `slides` (just adding/removing objects to that same array).

weerd: cool cool. that's what I thought, but wanted to make sure 💃

itsjoekent:
> Do either of you know if there's a way to use a string to somehow call a class in JS?

`eval()` might be what you're lookin for

weerd:
@deadlybutter cool! will try that!

DFurnes:
I'd caution against using `eval` unless we reaaaaaaaally need it, since it can be pretty dangerous to evaluate an untrusted string. You could probably get most of what you're looking for like this:

``` js
const allSlides = {ContextSlide, InstructionSlide, ReportbackItemKudosSlide};
const slides = Drupal.settings.dsOnboarding.slides.map(name => allSlides[name]);
```

weerd:
@DFurnes used your example and tweaked it a bit! Works!

lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
```diff
@@ -28,17 +28,24 @@ $(document).ready(function() {
...
   // Enable the Onboarding experiment.
   if (Drupal.settings.dsOnboarding.enabled && localStorage.getItem(localStorageKey) !== 'true') {
+    const slideComponents = {ContextSlide, InstructionSlide, ReportbackItemKudosSlide};
     const jsxElementHook = 'jsx-onboarding';
-    const slides = [InstructionSlide]; // ReportbackItemKudosSlide, ContextSlide
+    const slides = [];
 
-    $('.chrome').find('> .wrapper').before(`<div id="${jsxElementHook}"></div>`);
-    $('.messages:not(.error)').remove();
+    Drupal.settings.dsOnboarding.slides.map(function (slide) {
```
DFurnes: No need to use `.map` if you're not returning anything.

lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
```diff
@@ -28,17 +28,24 @@ $(document).ready(function() {
...
   // Enable the Onboarding experiment.
   if (Drupal.settings.dsOnboarding.enabled && localStorage.getItem(localStorageKey) !== 'true') {
+    const slideComponents = {ContextSlide, InstructionSlide, ReportbackItemKudosSlide};
     const jsxElementHook = 'jsx-onboarding';
-    const slides = [InstructionSlide]; // ReportbackItemKudosSlide, ContextSlide
+    let slides = [];
 
-    $('.chrome').find('> .wrapper').before(`<div id="${jsxElementHook}"></div>`);
-    $('.messages:not(.error)').remove();
+    slides = Drupal.settings.dsOnboarding.slides.map(function (slide) {
```
DFurnes: I know this is nit-picky, but I'm confused why you're setting `slides` to an empty array on line 33 and then immediately re-writing it on line 35 with the return value of `.map()`. Wouldn't it make more sense to remove the original empty array definition and just define it as the result of the map?

``` js
const slides = Drupal.settings.dsOnboarding.slides.map(name => allSlides[name]);
```

DFurnes:
🚢 🚢 🚢

Issue #6773, opened 2016-07-27
Different flow experience for +18 and under 18


ngjo:
#### User Story

As a user, if I will be 18 by the election, I want to get the registration form and be able to register.

As a user, if I will be under 18 by the election, I want to get the competition and social share experience.
#### Additional Information (optional)

Think this will be looking at their birthday (in the DS system). Election is on November 8, 2016 so they must be born on November 8, 1998 or earlier to by 18 by the election day.
#### Why This Matters

So the user gets the right flow

PR #6805, opened 2016-08-02
add voter reg modal and share modal (ish)

### LINKED ISSUE #6774: Voter widget as a pop-up modal
#### User Story

As a user, when I sign-up for the campaign, I want to be greeted with the registration modal (or the under 18 experience -- see #6773) as a pop-up!

So they don't have to scroll down to click the "Register to Vote" in the Stuff You Need section -- we still want to keep the link in that section too.  (Will get exact wording for what that link should be!)
#### Why This Matters

So the user can be prompted immediately

katiecrane:
#### What's this PR do?

This is one step forward (I don't want it to become a super huge PR). Current behavior is that when a user comes to a campaign with voter registration enabled, we check how old they are and give them either the share modal or the voter reg modal. I haven't added in any logic on when to stop serving them a modal. In the `Stuff You Need` section they are able to get directly to the voter reg module as well (perhaps we want to hide this if they are too young?). 

Go to a campaign with voter reg enabled. Try with an 18+ user and a not-18-by-election-day user. Make sure the behavior is as described above.

VOTIN'
#### Relevant tickets

Fixes #6774
Refs #6773
#### Checklist
- [x] Tested on staging.

cc: @ngjo


ngjo:
hey @katiecrane i tried enabling it on here https://staging.dosomething.org/us/node/1283/custom-settings but it doesn't seem to be popping up or showing up in "stuff you need" -- do i need to do something else to enable it?

katiecrane:
@ngjo It hasn't been merged yet 😄 what we have so far is turned on on the space jam campaign though!

ngjo:
doyyyyy @katiecrane will try it out there!

ngjo:
@katiecrane i just tried it on space jam campaign with someone born in 2000 but when i click "register to vote" the registration form popped up

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -413,27 +413,34 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
...
+  $share_button_markup = dosomething_helpers_share_bar($custom_organ_share_link, $social_share_types, 'organ_donation_referral', 'social-menu');
 
   $vars['custom_organ_share_link'] = $custom_organ_share_link;
-  $vars['organ_share_bar'] = $organ_donations_share_button_markup;
+  $vars['organ_share_bar'] = $share_button_markup;
+  $vars['social_share_bar'] = $share_button_markup;
+  $vars['custom_social_share_link'] = $custom_social_share_link;
+
+  // If we are registering voters check to see if the user will be 18 by the election
+  if (array_key_exists('enable_voter_registration', $campaign->variables) && $campaign->variables['enable_voter_registration']) {
```
angaither, line 442: we should add a `@todo` to see if ( name of feautre is enabled) because we use this chuck of code all over the place

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -413,27 +413,34 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
...
+    $vars['can_vote'] = dosomething_user_will_be_eightteen_by_date('Nov 8 2016', $person);
```
angaither: where are you getting this `$person` var?

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -757,6 +757,29 @@ function dosomething_user_is_under_thirteen($user = NULL) {
...
 
 /**
+ * Checks that user will be 18 by a certain date
+ *
+ * @param $date
+ *
+ * @return bool
+ *  Returns true if user will be 18 by the given date, else false.
+ */
+function dosomething_user_will_be_eightteen_by_date($date, $user = NULL) {
```
angaither: I think we have a function that gets the date diff somewhere already...

angaither:
@jen also wait, I am confused by this PR because I thought that the flow was 
over 18 + ---> voter reg
under 18 ----> competitions (not share)

ngjo:
@angaither it's either a share modal and then the competitions modal OR the competitions modal ideally w/  the custom social share link included in the text (if not, then just text directing them to the custom social share) -- when i was talking about it with @katiecrane i proposed both and left it up to her to see which was the easiest

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
katiecrane, line 442: @angaither yeah it comes up a whole lot. good call

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
katiecrane: OOPS that is left over from a test and just so happens to have no a(e?)ffect. thanks!

katiecrane:
@angaither @ngjo yup what Jen said is what I was aiming for. I started with just share (main goal being to have something different come up for under 18 users) and plan to either smush that and competitions into one modal or follow it up with the competitions one

lib/modules/dosomething/dosomething_user/dosomething_user.module
katiecrane: @angaither I couldn't find that function but I switched this one to be more general. Now it just returns the age of a user on the given date. Seems like that might be worth having but I'm down to remove

angaither:
🚢

PR #6806, opened 2016-08-02
Updating the close button to work better on mobile.

### LINKED ISSUE #6804: Extra padding on mobile - 'X'
#### Current Behavior

'X' looks like it have extra padding-left on mobile. 
#### Desired Behavior

X should appear aligned to the right of the screen. 
#### Steps to Replicate
1. Navigate to staging.dosomething.org
2. Sign up for a campaign
3. Observe onboarding 
#### Why This Matters

Having a floating X is poor design -- and not what our mockups display!

img_3656

weerd:
#### What's this PR do?

This PR addresses the extra spacing for the "x" close button on the Onboarding message feature when viewed on mobile.

Mobile:
image

Desktop:
image

This takes a slightly different approach and I think it works well because the extra padding around the button makes the button a nice clickable size, which is ideal for mobile. Then only on tablet does the button get some additional spacing from the right edge. This is also using the multiple of `12` that the forge pattern library aims to achieve with spacing.
#### Relevant tickets

Fixes #6804 
#### Checklist
- [ ] Tested on staging.

---

@DFurnes 

cc: @jessleenyc

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
@@ -28,9 +28,14 @@
   display: block;
   font-size: 42px;
   line-height: 1;
+  padding: ($base-spacing/2);
   position: absolute;
-  right: $base-spacing;
+  right: 0;
   top: 0;
+
+  @include media($tablet) {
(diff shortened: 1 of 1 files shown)


DFurnes:
👍

PR #6808, opened 2016-08-03
Display competition confirm msg in onboarding flow 

### LINKED ISSUE #6801: Show Competitions Notificiation In Onboarding Container
#### User Story

As a user, I would like affirmation that I've successfully joined a competition. In addition, I would like to know that I'll be contacted with next steps for the competition. 
#### Additional Info

The notification text should be in italics above the header. The competition notification should only appear if a member joins a competition.
#### Why This Matters

This is important because competition doesn't currently live anywhere on the action page. Without this notifications, users will not know what to expect after they've joined a competition.

itsjoekent:
#### What's this PR do?

^ title

If you signup for a competition is the confirm_msg  for that campaigns signup data form inserted into the onboarding flow?

not really
#### Relevant tickets

Fixes #6801 
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -401,7 +401,7 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
       $competition = ['joined' => $signup->competition];
       if ($competition['joined']) {
         $signup_form = dosomething_signup_get_signup_data_form_info($signup->nid);
-        $competition['confirm_msg'] = $signup_form['confirm_msg'];
+        $competition['messages']['confirmation'] = $signup_form['confirm_msg'];
       }
 
       drupal_add_js(
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -396,12 +396,21 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
   // Pass the signup id to the front end.
   if (module_exists('dosomething_signup')) {
     if ($sid = dosomething_signup_exists($vars['nid'])) {
+      // Get competition data
+      $signup = signup_load($sid);
+      $competition = ['joined' => $signup->competition];
+      if ($competition['joined']) {
+        $signup_form = dosomething_signup_get_signup_data_form_info($signup->nid);
+        $competition['confirm_msg'] = $signup_form['confirm_msg'];
+      }
+
       drupal_add_js(
-        array('dosomethingSignup' =>
-          array(
+        ['dosomethingSignup' =>
+          [
             'sid' => $sid,
-          ),
-        ),
+            'competition' => $competition
+          ],
+        ],
         'setting'
       );
     }
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
@@ -513,15 +513,6 @@ function dosomething_signup_user_signup_data_form_submit($form, &$form_state) {
   }
   // Display the signup_data_form's confirm_msg field.
   drupal_set_message($config['confirm_msg']);
-  drupal_add_js(
-    ['dosomethingSignup' =>
-      [
-        'competition' => true,
-        'confirm_msg' => $config['confirm_msg']
-      ],
-    ],
-    'setting'
-  );
 }
 
 /**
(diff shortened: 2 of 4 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -396,12 +396,21 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
...
+      $competition = ['joined' => $signup->competition];
+      if ($competition['joined']) {
+        $signup_form = dosomething_signup_get_signup_data_form_info($signup->nid);
+        $competition['confirm_msg'] = $signup_form['confirm_msg'];
+      }
+
       drupal_add_js(
-        array('dosomethingSignup' =>
-          array(
+        ['dosomethingSignup' =>
```
weerd, line 408: ~~can we change this to `dsSignup`. Moving forward we've just been prefixing Drupal.settings objects w/ `ds`. If this exists already, still make it `dsSignup` and we can eventually update the old one.~~

Errr, wait. This is the old one. Crap. ok, nevermind.

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -513,6 +513,15 @@ function dosomething_signup_user_signup_data_form_submit($form, &$form_state) {
   }
   // Display the signup_data_form's confirm_msg field.
   drupal_set_message($config['confirm_msg']);
+  drupal_add_js(
+    ['dosomethingSignup' =>
```
itsjoekent: whoops this needs to go

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
itsjoekent, line 408: yeah that confused me as well when debugging, for like 5 minutes i lost my mind

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -396,12 +396,21 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
...
+          [
             'sid' => $sid,
-          ),
-        ),
+            'competition' => $competition
```
weerd, line 411: this isn't a competition object or id or anything right? just the confirmation message? Maybe structure it as such, so 

``` php
`competition` => [
  'messages' => [
    'confirmation' => '$confirmation_message,
  ]
]
```

it's a bit more verbose, but if we need to expand on this object later it will afford us the opportunity to do so! 

Up to you if you want to make `messages` array, maybe it's just one property as `confirmation_message` or something and not a nested array inside the `competition` array.

itsjoekent, line 411: it also has the "joined" parameter, we can do this but i dont think we'll ever have more than this 1 message (for competitions)

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
weerd, line 408: yeah, not ideal. Moving forward we're prefixing everything in the JS appended script that is an object as `dsObjectName`. But harder for ones already named using the full dosomething prefix and already in use 😞

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
itsjoekent, line 411: @weerd updated with this ^ structure

weerd, line 411: Gotcha. Missed that you were passing in an array to begin with @_@ Anyhoo, yeah, I just prefer to make a logical hierarchy of properties so that if we amend these things or add to them, we're not in a weird place with property names, etc. Run into that before and it gets ugly.

lib/themes/dosomething/paraneue_dosomething/js/components/InstructionSlide.js
```diff
@@ -21,6 +23,7 @@ class InstructionSlide extends React.Component {
             <div className="container__row">
               <div className="container__block -narrow -centered">
                 <h2 className="heading -gamma">{title}</h2>
+                <p>{competition.joined ? competition.messages.confirmation : ''}</p>
```
weerd: Looks good. But do we want the default to be something related if the message is not present instead of just empty string?

weerd:
Add a default confirmation message instead of empty string! Other than that, cool to merge! 🚢

itsjoekent:
@weerd confirmed there should be no default message here instead

weerd:
okie dokes!

PR #6812, opened 2016-08-03
Minor updates to the order of things in a campaign run node

angaither:
#### What's this PR do?

Updating campaign run node feature to
- put the title before the description
- collapse the 'goal' group
- re-order goals to near the date, since that should be filled out when you make the node

it looks like this now
image

it used to look like this 
image
#### Relevant tickets

there are none, this was just annoying me. and it was easier than image recovery. 
#### Checklist
- [x] Tested on staging.


blisteringherb:
👍

PR #6821, opened 2016-08-04
Updates to feature card

### LINKED ISSUE #6818: Blank onboarding experience, for experienced users
#### Current Behavior

Onboarding experience appears blank if user is signed up for all staff picks
image
#### Desired Behavior

During the onboaring experience if a user is is signed up for all staff pick campaigns, pull in other non staff pick campaigns in the order described in #6665
- display open, published
- boost staff pick campaigns
- boost if has a partner that is a sponsor  
- boost by `field_active_hours` (smallest values first) 
#### Steps to Replicate
- check on staging on an account you've used a couple times.

itsjoekent:
#### What's this PR do?
- Fetches all of the campaigns for the feature card
- Adds signup request to the features array

If you're a super user who already signed up for every staff pick campaign, do you now see more staff pick campaigns?

See: https://github.com/DoSomething/phoenix/issues/6818#issuecomment-237594809

> For the sake of performance / load time, I think the best approach is to only retrieve the staff pick campaigns to initialize the card. After the card is initially rendered, we can then do a larger pagination to retrieve all non staff pick campaigns & apply the same boost logic. Since most users are not going to be signed up for every staff pick campaign, this should be incredibly seamless.

in practice it seems to work as I thought it would ^^
#### Relevant tickets

Fixes #6818 
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
@@ -22,9 +22,19 @@ function checkForSponsor(campaign) {
 }
 
 function sortCampaigns(a, b) {
-  const aScore = a.time_commitment - (checkForSponsor(a) ? 1 : 0);
-  const bScore = b.time_commitment - (checkForSponsor(b) ? 1 : 0);
-  return aScore - bScore;
+  // const aScore = a.time_commitment - (checkForSponsor(a) ? 1 : 0) + (a.staff_pick ? 2 : 0);
+  // const bScore = b.time_commitment - (checkForSponsor(b) ? 1 : 0) + (a.staff_pick ? 2 : 0);
+  // return aScore - bScore;
+
+  // something like
+  // if only one is a sponsor
+  // return 1 or -1
+  // else if one is a corporate
+  // return 1 or -1
+  // else
+  // return based on least time commit
+
+  return 0;
 }
 
 /**
@@ -37,26 +47,21 @@ class FeatureCard extends React.Component {
 
     this.requests = [];
     this.state = {
-      campaigns: [],
-      current_campaign: '',
-      signup_animation: ''
+      signupAnimation: '',
+      campaignPage: 1,
+      campaigns: []
     };
   }
 
   componentDidMount() {
     // First load staff pick campaigns to quickly initialize the card
-    this.getCampaigns({staffPick: true, campaigns: []}, 1, campaigns => {
+    this.getCampaigns(1, campaigns => {
       this.setState({
-        campaigns: campaigns,
-        current_campaign_index: 0
+        campaigns: campaigns
       });
 
-      // Then backfill with all of the other campaigns
-      this.getCampaigns({staff_pick: 0, campaigns: campaigns}, 1, campaigns => {
-        this.setState({
-          campaigns: campaigns
-        });
-      });
+      // For our crazy super users, double check they actually have campaigns to see!
+      this.checkCampaignsRemaining();
     });
   }
 
@@ -68,42 +73,42 @@ class FeatureCard extends React.Component {
     }
   }
 
-  getCampaigns(data, page, cb) {
-    this.requests.push($.get(`${API_URL}campaigns?staff_pick=${data.staff_pick}&page=${page}`, result => {
-      let rawCampaigns = result.data;
+  checkCampaignsRemaining() {
+    // If we're almost out of campaigns let's get more
+    if (this.state.campaigns.length < 4) {
+      // First update the API page
+      this.setState({campaignPage: this.state.campaignPage + 1});
+      // Then grab campaigns for that page & update the state
+      this.getCampaigns(this.state.campaignPage, campaigns => {
+        // this.setState({
+        //   campaigns: campaigns
+        // })
+      });
+    }
+  }
+
+  getCampaigns(page, cb) {
+    // Grab all campaigns for given page
+    // The staff_pick=true is embedded within the template string intentionally,
+    // it seems like the precense alone of the param causes it to execute..
+    this.requests.push($.get(`${API_URL}campaigns?page=${page}`, result => {
+      let rawCampaigns = result.data;console.log(result);
 
       // Filter out campaigns users have signed up for
       const userSignups = Drupal.settings.dsUser.activity.signups;
       rawCampaigns = rawCampaigns.filter(campaign => userSignups.indexOf(parseInt(campaign.id)) == -1);
 
-      // Remove any non staff picks
-      if (!data.staffPick) {
-        rawCampaigns = rawCampaigns.filter(campaign => campaign.staff_pick == false);
-      }
-
       // Finally sort by weight values
       rawCampaigns = rawCampaigns.sort(sortCampaigns)
 
       // Join previous campaigns with newly retrieved
-      const campaigns = data.campaigns.concat(rawCampaigns);
-
-      if (result.pagination.total_pages > page) {
-        data.campaigns = campaigns;
-        this.getCampaigns(data, page + 1, cb);
-        return;
-      }
+      const campaigns = this.state.campaigns.concat(rawCampaigns);
 
       cb(campaigns);
     }));
   }
 
-  signup() {
-    const campaign = this.state.campaigns[this.state.current_campaign_index];
-
-    if (!campaign) {
-      return;
-    }
-
+  signup(campaign) {
     this.requests.push($.ajax({
       url: `${API_URL}campaigns/${campaign.id}/signup`,
       type: 'POST',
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -43,32 +44,67 @@ class FeatureCard extends React.Component {
...
+        rawCampaigns = rawCampaigns.filter(campaign => campaign.staff_pick == false);
+      }
+
+      // Finally sort by weight values
+      rawCampaigns = rawCampaigns.sort(sortCampaigns)
+
+      // Join previous campaigns with newly retrieved
+      const campaigns = data.campaigns.concat(rawCampaigns);
+
+      if (result.pagination.total_pages > page) {
```
DFurnes: Wait, so does this load all 266 campaigns every time a user is shown the onboarding flow? If so, that seems awfully excessive. Can we limit the number of pages it loads, or lazy-load as the user scrolls through the cards?

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -43,32 +44,67 @@ class FeatureCard extends React.Component {
...
+    this.requests.push($.get(`${API_URL}campaigns?staff_pick=${data.staff_pick}&page=${page}`, result => {
+      let rawCampaigns = result.data;
+
+      // Filter out campaigns users have signed up for
+      const userSignups = Drupal.settings.dsUser.activity.signups;
+      rawCampaigns = rawCampaigns.filter(campaign => userSignups.indexOf(parseInt(campaign.id)) == -1);
+
+      // Remove any non staff picks
+      if (!data.staffPick) {
+        rawCampaigns = rawCampaigns.filter(campaign => campaign.staff_pick == false);
```
DFurnes: Is this necessary since you're doing this above with the `?staff_pick=` query parameter?

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
itsjoekent: loading as they go through seems like a simple enough approach. it only took ~1-2 seconds to load everything so I didn't think it would matter too much

itsjoekent: then again im also connecting locally, im sure out in the wild it would take longer

DFurnes: Yeah, this is making eleven ~100kb HTTP requests. It'll definitely be slower in the wild.

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
itsjoekent: thats what i thought as well -- turns out this parameter probably needs an issue.

if you pass FALSE it seems to be treated as `1` and returns staff pick only. If you pass `0` it returns both staff pick & non staff pick...

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -43,32 +44,67 @@ class FeatureCard extends React.Component {
...
+      }
     }
   }
 
+  getCampaigns(data, page, cb) {
+    this.requests.push($.get(`${API_URL}campaigns?staff_pick=${data.staff_pick}&page=${page}`, result => {
+      let rawCampaigns = result.data;
+
+      // Filter out campaigns users have signed up for
+      const userSignups = Drupal.settings.dsUser.activity.signups;
```
DFurnes, line 98: Let's use the [`setting()`](https://github.com/DoSomething/phoenix/blob/dev/lib/themes/dosomething/paraneue_dosomething/js/utilities/Setting.js) helper whenever we need to load something from the settings object.

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
DFurnes: Okay, can you make that issue and add a `@TODO: ...` comment here explaining this fix?

DFurnes:
Not necessary for this pull request, but let's consider breaking this `FeatureCard` into [separate data-fetching component and presentational components](https://medium.com/@learnreact/container-components-c0e67432e005#.86z23p82v) since this file is getting pretty hefty.

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
angaither: this is, I think where the api breaks down as the appropriate tool. we only need header image, title and cta, correct? we don't need to load the entire campaign.

itsjoekent: well if we do the page by page approach it's a pretty normal way to handle this problem, i just took the insane approach to make it quicker

DFurnes: @angaither Yeah, it kinda feels like we should be making a custom query and including the results in the page markup. Or this would be a beautiful task for GraphQL in the future. 📈

itsjoekent:
Did some refactoring to make it per page, still need to rewrite the sort function so 
1) its more clear how it works 
2) accepts all of these requirements 

Also I just removed all of the staff pick request logic for now, made that simpler. 

As for thoughts on implementation
@DFurnes totally agree this file is yuuge and breaking this AJAX logic out into a different component would allow the feature card to do more than just this in the future.
@angaither writing the solr query scared both diego & I haha, plus we'd have to either
A) query all of the campaign data for every campaign when onboarding is displayed
or
B) make API requests anyway to backfill

if we notice performance problems with the new page based implementation we can totally try solr, but I think doing it this way saves a lot of dev time for now

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -22,9 +22,19 @@ function checkForSponsor(campaign) {
 }
 
 function sortCampaigns(a, b) {
-  const aScore = a.time_commitment - (checkForSponsor(a) ? 1 : 0);
-  const bScore = b.time_commitment - (checkForSponsor(b) ? 1 : 0);
-  return aScore - bScore;
+  // const aScore = a.time_commitment - (checkForSponsor(a) ? 1 : 0) + (a.staff_pick ? 2 : 0);
```
DFurnes, line 25: Is this intended to be commented out?

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -37,26 +47,21 @@ class FeatureCard extends React.Component {
 
     this.requests = [];
     this.state = {
-      campaigns: [],
-      current_campaign: '',
-      signup_animation: ''
+      signupAnimation: '',
```
DFurnes, line 50: This should be a boolean state (like `isSignedUp`) and then the animation can be handled in the render method.

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -68,42 +73,42 @@ class FeatureCard extends React.Component {
...
-    this.requests.push($.get(`${API_URL}campaigns?staff_pick=${data.staff_pick}&page=${page}`, result => {
-      let rawCampaigns = result.data;
+  checkCampaignsRemaining() {
+    // If we're almost out of campaigns let's get more
+    if (this.state.campaigns.length < 4) {
+      // First update the API page
+      this.setState({campaignPage: this.state.campaignPage + 1});
+      // Then grab campaigns for that page & update the state
+      this.getCampaigns(this.state.campaignPage, campaigns => {
+        // this.setState({
```
DFurnes, line 83: Also this should probably be uncommented?

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -68,42 +73,42 @@ class FeatureCard extends React.Component {
...
+      });
+    }
+  }
+
+  getCampaigns(page, cb) {
+    // Grab all campaigns for given page
+    // The staff_pick=true is embedded within the template string intentionally,
+    // it seems like the precense alone of the param causes it to execute..
+    this.requests.push($.get(`${API_URL}campaigns?page=${page}`, result => {
+      let rawCampaigns = result.data;console.log(result);
```
DFurnes, line 95: `console.log` 👾

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -119,31 +124,37 @@ class FeatureCard extends React.Component {
...
     }));
   }
 
   viewAnother() {
-    let newIndex = this.state.current_campaign_index + 1;
-    if (newIndex >= this.state.campaigns.length) {
-      newIndex = 0;
-    }
+    // Remove current campaign
+    this.state.campaigns.splice(0, 1);
```
DFurnes, line 135: This seems pretty confusing (and potentially harder to change, for example if we we wanted a "go back" button in the future). What's the advantage over using an index to pick which campaign to display?

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -119,31 +124,37 @@ class FeatureCard extends React.Component {
...
-    const campaign = this.state.campaigns[this.state.current_campaign_index];
+    const campaign = this.state.campaigns[0];
 
     if (!campaign) {
       return null;
     }
 
+    // Local campaign files are sometimes borked, quick fix (DS Logo)
+    if (campaign.cover_image.default == undefined) {
+      campaign.cover_image.default = {sizes: {square: {uri: ''}}};
```
DFurnes, line 155: Looks like this needs a `@TODO` for the image URI?

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
itsjoekent, line 135: The problem I was running into was we need to remove campaigns from the campaign array when you sign up. If you do that the state changes & things get out of sync

I figured I could either remove the index or store campaigns off the state, wasn't sure which was better.

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
itsjoekent, line 83: I left a fee things still commented out / undone because it's a wip, didn't mean to ask for review yet!

itsjoekent:
different branch / PR Incoming, closing this one

PR #6829, opened 2016-08-08
Adding new Button component.

### LINKED ISSUE #6819: Feature card signup animation
#### Current Behavior
- When signing up for a campaign via the feature card, the button wobbles.
#### Desired Behavior
- after discussion with Luke we decided it should actually look something like this https://dribbble.com/shots/2325445-Listen
#### Steps to Replicate
- When signing up for a campaign via the feature card, the button wobbles.

weerd:
#### What's this PR do?

This PR adds a new reusable `Button` component that accepts a few different `props`. It adds the new Button component to the `FeatureCard` and also enables an animation on the button to indicate to the user that they have successfully signed up to the campaign indicated on the card.

Example animation:

image

The button gets disabled when it's in the green "success" state, to avoid further clicking since it's no longer really a button but a confirmation indicator.

To help with testing and not actually signup to each campaign indicated on the card, I switched the AJAX request to hit a specific campaign and just do a GET request :trollface: 

We can tweak the animation if needed. I used CSS `@keyframes`, and both the checkmark is a separate pseudo element that can be animated separately. Might be useful to eventually move this into Forge, and/or maybe optimize to allow for more versatility.
#### Relevant tickets

Fixes #6819
#### Checklist
- [ ] Tested on staging.

---

@DFurnes @deadlybutter 

cc: @jessleenyc

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
@@ -7,15 +7,17 @@ const React = require('react');
  * @return {[type]}       [description]
  */
 const Button = (props) => {
+    // console.log(props);
     const classes = ['button'].concat(props.classes);
 
     return (
-      <button className={classes.join(' ')} onClick={props.action}>{props.text}</button>
+      <button className={classes.join(' ')} onClick={props.action()}>{props.text}</button>
     );
 }
 
 Button.defaultProps = {
   loading: false,
+  success: false,
   text: 'Submit',
 };
 
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
@@ -5,7 +5,6 @@ const React = require('react');
  * <Button />
  */
 const Button = (props) => {
-    console.log(props.children);
     const classes = ['button'].concat(props.classes);
 
     if (props.loading === true) {
@@ -14,6 +13,14 @@ const Button = (props) => {
 
     if (props.success === true) {
       classes.push('is-successful');
+    }
+
+    if (props.type) {
+      classes.push(props.type);
+    }
+
+    if (props.reactive) {
+      classes.push('-reactive');
     }
 
     return (
@@ -32,6 +39,8 @@ Button.propTypes = {
   disabled: React.PropTypes.bool,
   loading: React.PropTypes.bool,
   onClick: React.PropTypes.func.isRequired,
+  reactive: React.PropTypes.bool,
+  type: React.PropTypes.oneOf(['-secondary', '-tertiary']),
 };
 
 export default Button;
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
@@ -16,7 +16,7 @@ const Button = (props) => {
     }
 
     if (props.type) {
-      classes.push(props.type);
+      classes.push('-' + props.type);
     }
 
     if (props.reactive) {
@@ -40,7 +40,7 @@ Button.propTypes = {
   loading: React.PropTypes.bool,
   onClick: React.PropTypes.func.isRequired,
   reactive: React.PropTypes.bool,
-  type: React.PropTypes.oneOf(['-secondary', '-tertiary']),
+  type: React.PropTypes.oneOf(['secondary', 'tertiary']),
 };
 
 export default Button;
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
@@ -3,27 +3,34 @@ const React = require('react');
 /**
  * Button Component
  * <Button />
- * @param  {[type]} props [description]
- * @return {[type]}       [description]
  */
 const Button = (props) => {
-    // console.log(props);
     const classes = ['button'].concat(props.classes);
 
+    if (props.loading === true) {
+      classes.push('is-loading');
+    }
+
+    if (props.success === true) {
+      classes.push('is-successful');
+    }
+
     return (
-      <button className={classes.join(' ')} onClick={props.action()}>{props.text}</button>
+      <button className={classes.join(' ')} onClick={props.onClick} disabled={props.disabled}>{props.text}</button>
     );
 }
 
 Button.defaultProps = {
+  disabled: false,
   loading: false,
   success: false,
   text: 'Submit',
 };
 
 Button.propTypes = {
-  action: React.PropTypes.func.isRequired,
+  disabled: React.PropTypes.bool,
   loading: React.PropTypes.bool,
+  onClick: React.PropTypes.func.isRequired,
   text: React.PropTypes.string,
 };
 
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
@@ -5,6 +5,7 @@ const React = require('react');
  * <Button />
  */
 const Button = (props) => {
+    console.log(props.children);
     const classes = ['button'].concat(props.classes);
 
     if (props.loading === true) {
@@ -16,7 +17,7 @@ const Button = (props) => {
     }
 
     return (
-      <button className={classes.join(' ')} onClick={props.onClick} disabled={props.disabled}>{props.text}</button>
+      <button className={classes.join(' ')} onClick={props.onClick} disabled={props.disabled}>{props.children}</button>
     );
 }
 
@@ -24,14 +25,13 @@ Button.defaultProps = {
   disabled: false,
   loading: false,
   success: false,
-  text: 'Submit',
+  children: 'Submit',
 };
 
 Button.propTypes = {
   disabled: React.PropTypes.bool,
   loading: React.PropTypes.bool,
   onClick: React.PropTypes.func.isRequired,
-  text: React.PropTypes.string,
 };
 
 export default Button;
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
@@ -1,0 +1,28 @@
+const React = require('react');
+
+/**
+ * Button Component
+ * <Button />
+ * @param  {[type]} props [description]
+ * @return {[type]}       [description]
+ */
+const Button = (props) => {
+    const classes = ['button'].concat(props.classes);
+
+    return (
+      <button className={classes.join(' ')} onClick={props.action}>{props.text}</button>
+    );
+}
+
+Button.defaultProps = {
+  loading: false,
+  text: 'Submit',
+};
+
+Button.propTypes = {
+  action: React.PropTypes.func.isRequired,
+  loading: React.PropTypes.bool,
+  text: React.PropTypes.string,
+};
+
+export default Button;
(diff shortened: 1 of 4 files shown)


lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
```diff
@@ -0,0 +1,30 @@
...
+ * <Button />
+ * @param  {[type]} props [description]
+ * @return {[type]}       [description]
+ */
+const Button = (props) => {
+    // console.log(props);
+    const classes = ['button'].concat(props.classes);
+
+    return (
+      <button className={classes.join(' ')} onClick={props.action()}>{props.text}</button>
```
DFurnes: Try switching this out to `onClick={props.action}`.

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -188,7 +201,7 @@ class FeatureCard extends React.Component {
               <h2 className="heading -delta">{campaign.title}</h2>
               <p className="tagline">{campaign.tagline}</p>
               <ul className="form-actions">
-                {(campaign.signedUp && !this.state.isSignedUp) ? <h1>You're signed up already!</h1> : <li><button className={`${this.state.isSignedUp ? 'tada' : ''} button`} onClick={()=>this.signup()}>Sign Up</button></li>}
+                <li><Button loading={this.state.isLoading} success={this.state.isSignedUp} action={this.signup}></Button></li>
```
DFurnes: And then here you can pass the bound function: `action={this.signup.bind(this)}`

DFurnes: It might be cleaner to bind all the event handlers to the component instance in the constructor (like [this](https://github.com/DoSomething/voting-app/blob/04234aebff2cf60114e3d0c36dd02d25d2fe555b/resources/assets/js/components/Tile.js#L10)), but either way should work fine.

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -100,6 +102,11 @@ class FeatureCard extends React.Component {
   }
 
   signup() {
+    console.log(this);
```
itsjoekent: 👀

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -44,9 +45,13 @@ class FeatureCard extends React.Component {
...
 
     this.state = {
+      actionDisabled: false,
       campaignIndex: -1,
-      isSignedUp: ''
+      isSignedUp: false,
+      isLoading: false,
     };
+
+    this.signup = this.signup.bind(this);
```
itsjoekent, line 54: wait this works? I just tried it in both a node console & browser window, both threw an error for bind of undefined. (am I not seeing where this is defined beforehand?)

couldn't this just be `this.signup = this;`? bind feels unnecessary here

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -120,13 +125,20 @@ class FeatureCard extends React.Component {
       },
       success: (data) => {
         this.campaigns[this.state.campaignIndex].signedUp = true;
-        this.setState({ //TODO: Fancy animation
+        this.setState({
```
itsjoekent, line 128: ahh now i get what you did with the loading state stuff, 👍 👍

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_button.scss
```diff
@@ -1,14 +1,73 @@
+%animated {
```
itsjoekent: Im gonna ask a possibly dumb question, what is this selector..?

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_button.scss
```diff
@@ -1,14 +1,73 @@
...
+    &::after {
+      content: '\02713';
+      display: block;
+      font-size: $font-large;
+      left: 50%;
+      line-height: 1;
+      opacity: 0;
+      position: absolute;
+      text-align: center;
+      transform: translate(-50%, -50%);
```
itsjoekent, line 40: Hm, we're setting left to 50%, top 50%, and then translating it back -50% in both dimensions?

lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
```diff
@@ -0,0 +1,37 @@
+const React = require('react');
+
+/**
+ * Button Component
+ * <Button />
+ */
+const Button = (props) => {
+    const classes = ['button'].concat(props.classes);
```
DFurnes, line 8: I'd recommend removing `props.classes`, since it breaks encapsulation of the component's styles (and there probably shouldn't be any need to pass arbitrary classes in that wouldn't be better served as part of the button).

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
DFurnes, line 54: This [binds](function.prototype.bind.apply%28...%29 is not a constructor) that event handler function's `this` so that it points to the component instance, rather than the DOM element. If you replaced this line with `this.signup = this`, that would overwrite the method on the object to just be a circular reference to the object itself.

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -194,7 +207,7 @@ class FeatureCard extends React.Component {
               <h2 className="heading -delta">{campaign.title}</h2>
               <p className="tagline">{campaign.tagline}</p>
               <ul className="form-actions">
-                {(campaign.signedUp && !this.state.isSignedUp) ? <h1>You're signed up already!</h1> : <li><button className={`${this.state.isSignedUp ? 'tada' : ''} button`} onClick={()=>this.signup()}>Sign Up</button></li>}
+                <li><Button classes={['-reactive']} text="Sign Up" loading={this.state.isLoading} disabled={this.state.actionDisabled} success={this.state.isSignedUp} onClick={this.signup}></Button></li>
                 <li><a className="button -tertiary" onClick={()=>this.viewAnother(true)}>Show me another</a></li>
```
DFurnes: This can also use the Button component:

``` jsx
<li><Button onClick={()=>this.viewAnother(true)} text="Show me another" /></li>
```

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
itsjoekent, line 54: Oh right this is within the constructor, that went over my head

lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
```diff
@@ -0,0 +1,37 @@
...
+    if (props.loading === true) {
+      classes.push('is-loading');
+    }
+
+    if (props.success === true) {
+      classes.push('is-successful');
+    }
+
+    return (
+      <button className={classes.join(' ')} onClick={props.onClick} disabled={props.disabled}>{props.text}</button>
```
DFurnes: I think you can probably just use `props.children` for the text, so the component can be used a bit more naturally: `<Button>Sign Up!</Button>`. And then you can validate that `props.children` must be a string in the PropTypes.

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_button.scss
DFurnes: I _thiiiiink_ you should be able to use Forge's existing [`%animated`](https://github.com/DoSomething/forge/blob/dc8d2a60ce5b09c448007a2ad5b1d353e1cd901a/scss/_base/_animations.scss#L1-L5) placeholder selector here.

weerd: it's the [placeholder selector!](http://thesassway.com/intermediate/understanding-placeholder-selectors).

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_button.scss
weerd, line 40: it's a trick for centering both vertically and horizontally when you don't know the width & height.

lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
DFurnes, line 8: And then the `.-reactive` class can be applied based on a prop, and stays as part of the Button's implementation.

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
weerd: it can't cause it's supposed to be an anchor link and not a button... unless, it doesn't need to be...

lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
weerd, line 8: well, I pass in the `-reactive` class because I need to differentiate between a button w/ the checkmark stuff added to it, and one without.

lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
weerd: ah cool! Did not know :)

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_button.scss
weerd: I can't cause I'm changing the animation duration. I probs don't need the placeholder anymore since it's only one rule now... before I was using both the fill-mode and the duration...

weerd: Actually, I can just use the Forge one and then just override where necessary. Good call @DFurnes 💃

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
DFurnes: Yeah, I don't think this one should be an anchor link.

lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
DFurnes, line 8: But that class should be applied based on a prop, like any of the others. It shouldn't be the `FeatureCard`'s concern what the actual classname is.

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
weerd: yar. changing it :)

lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
```diff
@@ -0,0 +1,46 @@
...
+  success: false,
+  children: 'Submit',
+};
+
+Button.propTypes = {
+  disabled: React.PropTypes.bool,
+  loading: React.PropTypes.bool,
+  onClick: React.PropTypes.func.isRequired,
+  reactive: React.PropTypes.bool,
+  type: React.PropTypes.oneOf(['-secondary', '-tertiary']),
```
DFurnes: This is such a nitpick, but it'd be cool to have the type be `secondary` or `tertiary`, and then add the `-` prefix when translating it into a CSS class:

``` js
    if (props.type) {
       classes.push('-' + props.type);
    }
```

DFurnes:
One super unimportant nitpicky suggestion, but looks great! 🚢

lib/themes/dosomething/paraneue_dosomething/js/components/Button.js
weerd: nitpick indeed! but a solid one. will update!

PR #6835, opened 2016-08-08
Update script to safely de-activate, and keep latest active.

DFurnes:
#### What's this PR do?

So funny story, apparently I had made this script in #6475 and completely forgot about it. 👴 However, I think there were some smart things that we talked about today so I updated it based on that discussion!
1. It now keeps the most recently accessed account for each duplicated email.
2. Instead of deleting accounts, the script now deactivates the dupes & renames them to a "valid" but obviously fake email so they don't conflict when syncing to Northstar.

screen shot 2016-08-08 at 5 57 05 pm

👀 

🔥 
#### Relevant tickets

References #6409.
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/scripts/remove-duplicate-emails.php
+++ b/scripts/remove-duplicate-emails.php
@@ -6,24 +6,37 @@
  * drush --script-path=../scripts/ php-script remove-duplicate-emails.php
  */
 
-$users = db_query('SELECT uid FROM users 
-  WHERE mail IN (SELECT mail FROM users GROUP BY mail HAVING COUNT(mail) > 1)
-  AND uid NOT IN (SELECT MIN(uid) FROM users GROUP BY mail HAVING COUNT(mail) > 1)');
+$dupes = array_keys(db_query('SELECT mail FROM users GROUP BY mail HAVING COUNT(mail) > 1')->fetchAllKeyed());
+$removed = 0;
 
-foreach ($users as $user) {
-  $user = user_load($user->uid);
+foreach ($dupes as $mail) {
+  // Load all users with that duped email address, with the most recently accessed first.
+  $users = db_query('SELECT uid FROM users WHERE mail = :mail ORDER BY access DESC', [':mail' => $mail]);
+  $canonical_uid = 0;
 
-  if($user->access == 0) {
-    print 'Deleting ' . $user->uid . ' (' . $user->mail . ')' . PHP_EOL;
+  foreach ($users as $key => $user) {
+    $user = user_load($user->uid);
 
-    $northstar_id = $user->field_northstar_id[LANGUAGE_NONE][0]['value'];
-
-    if ($northstar_id !== 'NONE') {
-      print 'User ' . $user->uid . ' has a Northstar profile: ' . $northstar_id;
+    if ($key == 0) {
+      print 'Keeping ' . $user->uid . ' for ' . $user->mail . '.' . PHP_EOL;
+      $canonical_uid = $user->uid;
+      continue;
     }
 
-    user_delete($user->uid);
-  } else {
-    print 'Ignoring duplicate ' . $user->uid . ' (' . $user->mail . ')' . PHP_EOL;
+    // Set the new email for the deactivated user.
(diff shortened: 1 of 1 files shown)


scripts/remove-duplicate-emails.php
```diff
@@ -6,24 +6,37 @@
...
+  $users = db_query('SELECT uid FROM users WHERE mail = :mail ORDER BY access DESC', [':mail' => $mail]);
+  $canonical_uid = 0;
 
-  if($user->access == 0) {
-    print 'Deleting ' . $user->uid . ' (' . $user->mail . ')' . PHP_EOL;
+  foreach ($users as $key => $user) {
+    $user = user_load($user->uid);
 
-    $northstar_id = $user->field_northstar_id[LANGUAGE_NONE][0]['value'];
+    if ($key == 0) {
```
angaither, line 29: instead of adding this chunk to skip user 0, can we update the query to skip the first user?

scripts/remove-duplicate-emails.php
```diff
@@ -6,24 +6,37 @@
...
+    $new_email = 'duplicate-' . $canonical_uid . '-' . $key . '@dosomething.invalid';
+    print ' - Removing ' . $user->uid . ' (' . $user->mail . ' --> ' . $new_email . ')' . PHP_EOL;
+    user_save($user, ['mail' => $new_email, 'status' => 0]);
+    $removed++;
+
+    // Finally, make a note if that user has a Northstar profile, so we can clean that up.
+    $northstar_id = $user->field_northstar_id[LANGUAGE_NONE][0]['value'];
     if ($northstar_id !== 'NONE') {
-      print 'User ' . $user->uid . ' has a Northstar profile: ' . $northstar_id;
+      print '   ** User ' . $user->uid . ' has a Northstar profile: ' . $northstar_id;
```
angaither: should we store this in a temp table to clean up the users in NS in a different script?

scripts/remove-duplicate-emails.php
```diff
@@ -6,24 +6,37 @@
...
+    if ($key == 0) {
+      print 'Keeping ' . $user->uid . ' for ' . $user->mail . '.' . PHP_EOL;
+      $canonical_uid = $user->uid;
+      continue;
+    }
 
+    // Set the new email for the deactivated user.
+    $new_email = 'duplicate-' . $canonical_uid . '-' . $key . '@dosomething.invalid';
+    print ' - Removing ' . $user->uid . ' (' . $user->mail . ' --> ' . $new_email . ')' . PHP_EOL;
+    user_save($user, ['mail' => $new_email, 'status' => 0]);
```
angaither, line 38: does this save with the `$user` pass in 🆗 ? 😧

scripts/remove-duplicate-emails.php
DFurnes, line 29: Hmm, I think we need it in the query so we can store the canonical UID on the duped accounts (for the `duplicate-12345-1@dosomething.invalid` placeholder email). Is there a way to get the UID but skip the rest of the record?

scripts/remove-duplicate-emails.php
DFurnes, line 38: Yep, just did a spot check and they've still got the hashed password. The problem we ran into last time was that the changes were made to the `$user` object and then just persisted with `user_save($user)`, rather than putting the individual changes in the edits array.

scripts/remove-duplicate-emails.php
DFurnes: Yeah, I think that's a good idea! 📝

scripts/remove-duplicate-emails.php
angaither, line 29: Ah, I misunderstood what this was doing 🌵

scripts/remove-duplicate-emails.php
DFurnes: Added a table for these in 1ea35b1!

angaither:
👯 🎱 🔥 💃

PR #6836, opened 2016-08-08
remove image requirement if rb is being updated

### LINKED ISSUE #6762: Updating quantity on reportbacks
#### Current Behavior
- tell us what happened
  when someone tries to update the quantity on their RB, we saw in our meeting that you are required to upload a new photo. this is happening for the user and on the admin edit view of the alpha permalink
#### Desired Behavior
- what should have happened
#### Steps to Replicate
1. sign up for a campaign
2. RB 
3. Try to update quantity
#### Why This Matters

so users can update their quantity (maybe they wrote it wrong) and so admins can adjust the quantity if it's wrong.

screen shot 2016-07-26 at 3 24 01 pm

katiecrane:
#### What's this PR do?

If we're updating a reportback, don't validate it to make sure it has an image. No image necessary. 

👀 
1. Step 4: Prove It
2. return to the action page
3. update your quantity or caption or get real wild and update BOTH
4. step 3 should work and you should end up on the permalink page 🎉 
#### Relevant tickets

Fixes #6762
#### Checklist
- [x] Tested on staging. (by @katiecrane and @ngjo)

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -103,6 +103,7 @@ function dosomething_reportback_form($form, &$form_state, $entity = NULL) {
   else {
     // Output reportback images with date last updated.
     $form['reportback_submissions']['reportback_item']['#attributes']['is_supersized'] = FALSE;
+    $form['reportback_submissions']['reportback_item']['#attributes']['data-validate'] = '';
 
     $reportback_submissions_data = array();
     // Specify number of prior reportbacks to retrieve, including the latest one to show prominently.
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -103,7 +103,9 @@ function dosomething_reportback_form($form, &$form_state, $entity = NULL) {
   else {
     // Output reportback images with date last updated.
     $form['reportback_submissions']['reportback_item']['#attributes']['is_supersized'] = FALSE;
-    $form['reportback_submissions']['reportback_item']['#attributes']['data-validate'] = '';
+
+    // do not require a new image to be uploaded on update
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -103,6 +103,7 @@ function dosomething_reportback_form($form, &$form_state, $entity = NULL) {
   else {
     // Output reportback images with date last updated.
     $form['reportback_submissions']['reportback_item']['#attributes']['is_supersized'] = FALSE;
+    $form['reportback_submissions']['reportback_item']['#attributes']['data-validate'] = '';
```
angaither: normally we would just `unset` these values, but I think we still want to validate the item on update, just don't require an image, right?

katiecrane: wow `unset` duhhhh. that's why it was being funky.

Previously `data-validate` was using [`hasImage`](https://github.com/DoSomething/phoenix/blob/7b2a84aa399c9bd2eb9a91ddc6ad60bf50e5ccd1/lib/themes/dosomething/paraneue_dosomething/js/validators/reportback.js#L34-L48) whose description is `make sure an image is uploaded`, so I think that's all that this part was handling?

angaither:
@katiecrane if the validate logic should be using `hasImage`, shouldn't that be fixed instead of not validating updates?

> Previously data-validate was using hasImage whose description is make sure an image is uploaded, so I think that's all that this part was handling?

DFurnes:
I think removing the validation rule makes sense, since it's basically just a `required` validator and that rule only applies to new reportbacks. 👍 I believe we validate dimensions as part of the actual image uploader codebase (since files aren't supported in the validation code).

angaither:
okie! 🚢

katiecrane:
@angaither you raise an excellent point. everything else seems to still be working though, I get the expected "image too small" if I try to put up something tiny. So the update is being validated, just the specific validation that forces there to be an image is not called on updates.

PR #6841, opened 2016-08-09
Fixes onboarding close button

### LINKED ISSUE #6838: Need to click "X" twice to close out of onboarding
#### Current Behavior

Tested several onboarding experiences and need to click 'x' twice to close the container. The first click shows a blue box surrounding the element. See below screenshot. 
#### Desired Behavior

The container should close with one click.

screen shot 2016-08-09 at 9 42 47 am

itsjoekent:
#### What's this PR do?

Fixes the on boarding close button

Can you close on boarding properly now?

On first click the feature card was not unmounting correctly. This is because we were using a `for in` loop, which returned the index of the xhr object, not the object itself. Thus when we called .abort() it threw an error. 

By switching to a `for of` loop, we can get the actual xhr object. More here on this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of#Difference_between_for...of_and_for...in
#### Relevant tickets

Fixes #6838 
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
@@ -93,7 +93,7 @@ class FeatureCard extends React.Component {
 
   componentWillUnmount() {
     if (this.requests) {
-      for (var request of this.requests) {
+      for (let request of this.requests) {
         request.abort();
       }
     }
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
@@ -93,7 +93,7 @@ class FeatureCard extends React.Component {
 
   componentWillUnmount() {
     if (this.requests) {
-      for (var request in this.requests) {
+      for (var request of this.requests) {
         request.abort();
       }
     }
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -93,7 +93,7 @@ class FeatureCard extends React.Component {
 
   componentWillUnmount() {
     if (this.requests) {
-      for (var request in this.requests) {
+      for (let request of this.requests) {
```
DFurnes: You could do this a bit more concisely with the `this.request` array's forEach method:

``` js
this.requests.forEach(request => request.abort());
```

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -93,7 +93,7 @@ class FeatureCard extends React.Component {
 
   componentWillUnmount() {
     if (this.requests) {
```
DFurnes: Is this check necessary, since `this.requests` is defined as an array in the constructor?

DFurnes:
Two thoughts, otherwise ⛵!

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
itsjoekent: this is truth

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
itsjoekent: technically not necessary! left over from before when it wasn't defined up there already

PR #6842, opened 2016-08-09
And phone cleanup script

DFurnes:
#### What's this PR do?

Alongside #6835, this should clear up the last of our duplicate data problems so we can update the schema to make those fields unique! The `unset-duplicate-mobiles` finds all users where there are duplicates for `field_mobile`, and removes the field for all but the most recently active user.

screen shot 2016-08-09 at 2 36 56 pm

👀

📴
#### Relevant tickets

References #6409.
#### Checklist
- [ ] Tested on staging.


angaither:
awyis 🚢

PR #6843, opened 2016-08-09
Attempt to fix setTimeout issue.

weerd:
#### What's this PR do?

This PR provides an attempt at a fix for the a bug with the delay to auto-show a new campaign after a successful signup from the `FeatureCard`. There is a `setTimeout` that waits 5 seconds before it auto-loads a new campaign. However, if the "show me another" is clicked prior to the 5 seconds completing, a new campaign immediately shows and then when the time completes, it shows another immediately after. So we need to clear the timeout whenever the `viewAnother` method is called.

Um, 👀  and let me know if this seems like a good approach or not! I needed to be able to save the `setTimeout()` to a variable or property I could call directly in the `clearTimeout()`
#### Relevant tickets

Fixes 💅  🐛 
#### Checklist
- [ ] Tested on staging.

---

@DFurnes @deadlybutter

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
@@ -51,6 +51,7 @@ class FeatureCard extends React.Component {
       isLoading: false,
     };
 
+    this.delay = null;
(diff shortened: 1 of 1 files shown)


DFurnes:
Seems like a decent fix. 👢 🐛

PR #6848, opened 2016-08-10
Onboarding  button animation tweak

weerd:
#### What's this PR do?

This PR makes a quick update to eliminate the flash of text on the button as it changes from the loading animation, back to its original state, and then transitions to the successful action button animation.

👀 
#### Relevant tickets

Fixes 💅 
#### Checklist
- [ ] Tested on staging.

---

@DFurnes

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
@@ -95,10 +95,8 @@ class FeatureCard extends React.Component {
     this.campaignsToExclude.push(campaign.id);
 
     this.requests.push($.ajax({
-      // url: `${API_URL}campaigns/${campaign.id}/signup`,
-      // type: 'POST',
-      url: `${API_URL}campaigns/1283`,
-      type: 'GET',
+      url: `${API_URL}campaigns/${campaign.id}/signup`,
+      type: 'POST',
       headers: {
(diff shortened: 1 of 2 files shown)


DFurnes:
Does this work in IE? I seem to remember `color: transparent` being a bit of a pain but I forget whether that was just an IE 8 thing or not.

itsjoekent:
some interesting notes on that ^ here: http://stackoverflow.com/questions/3171586/color-transparent-is-not-working-in-internet-explorer

weerd:
ah good call. What IE are we still supporting? Been out of the fender game for Phoenix for a bit and can't remember. 

@deadlybutter good find. I'll switch it to `visibility`.

weerd:
errr. wait. I can't switch to visibility. That'll affect the entire button. Color specifically targets the text. I'd have to wrap it in a `span` which I'd like to avoid @_@

DFurnes:
I think this is still good to merge, but you might need to add a fallback solid color like we do for the [`.is-loading` modifier](https://github.com/DoSomething/forge/blob/dc8d2a60ce5b09c448007a2ad5b1d353e1cd901a/scss/_components/_forms/_button.scss#L122-L123) in Forge.

> What IE are we still supporting? Been out of the fender game for Phoenix for a bit and can't remember.

Let's concentrate on supporting IE10+, but if it's easy to fix stuff up for IE8/IE9 then why not.

PR #6850, opened 2016-08-10
Add script to clean out bad Northstar data.

DFurnes:
#### What's this PR do?

Three things:

🚳 In the migration script, add Northstar IDs to the `dosometing_northstar_delete_queue` table if they conflict on the `drupal_id` field. This is necessary because not all of the duped users from the other two scripts had their Northstar ID field filled in.

🔥 Adds a `delete-bad-users-in-northstar` script to empty out that delete queue by deleting the offending Northstar users with extreme prejudice.

🍡 Reset the Northstar settings to their default values when running `ds pull stage --db`, since otherwise the Northstar module will still be configured to connect to Northstar QA if enabled.

👀

👻 
#### Relevant tickets

References #6409.
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -464,6 +464,12 @@ function pull_db {
       echo "Disabling external modules locally."
       drush -y dis optimizely dosomething_northstar dosomething_gladiator
 
+      # Also reset Northstar settings, so enabling the module doesn't default to Northstar QA.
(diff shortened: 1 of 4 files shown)


angaither:
ah lovely!

PR #6859, opened 2016-08-10
Improved "delete bad users" script.

DFurnes:
#### What's this PR do?

This PR includes some improvements to the "delete bad users" and "export to Northstar" scripts:

🔁 After deleting a "conflicted" Northstar ID from the `dosomething_northstar_delete_queue` table, try to re-upload that Drupal ID's profile via the `POST users/` endpoint (same method as in the export script). This way we won't have to run the whooooole export process again after the expected validation errors.

:vs: Fixes a minor issue where I was trying to do a strict comparison of the status code (as an integer), but Drupal returns it as a float! Weird!

Does this make sense as the most efficient way to clear out bad Northstar profiles and re-fill them from the "canonical" Northstar data?

I noticed this after trying the "cleanup" process on Staging, and then realizing there was no way to fix the users that were removed via the "delete queue" without re-syncing _eeeeveryone_. So with this change, the process for Thor/Production:
1. Run the `php artisan northstar:clean_drupal_ids` script on Northstar. This will delete any profiles with duplicate Drupal IDs on Northstar (and attempt to keep the "correct" profile based on choosing the one with the earliest creation date).
2. Run the `remove-duplicate-emails` and `unset-duplicate-mobiles` Drush scripts to fix duplicated records in Phoenix before re-syncing. This will mark changed Drupal IDs (and if field is populated, Northstar IDs) in the `dosomething_northstar_delete_queue` table.
3. Run the `delete-bad-users-in-northstar` Drush script. It'll delete the marked Northstar IDs, and try to re-sync the marked Phoenix IDs (and if that results in a conflict, it'll add the conflicting Northstar IDs to the queue). Run it again for newly added Northstar IDs.
4. Finally, re-run the `export-users-to-northstar` script as a final spot-check. The only users that should not have a successful response code are users with invalid emails.

:cactus: 
#### Relevant tickets

References #6409.
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/scripts/delete-bad-users-in-northstar.php
+++ b/scripts/delete-bad-users-in-northstar.php
@@ -12,6 +12,8 @@ $records = db_query('SELECT id, uid, northstar_id FROM dosomething_northstar_del
 
 foreach ($records as $record) {
   $client = _dosomething_northstar_build_http_client();
+
+  // Delete the user with the provided conflicting Northstar ID.
   $response = drupal_http_request($client['base_url'] . '/users/' . $record->northstar_id, [
     'headers' => $client['headers'],
     'method' => 'DELETE',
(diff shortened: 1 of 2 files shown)


angaither:
⚓

PR #6862, opened 2016-08-11
Fixes GA on admin/prod

itsjoekent:
#### What's this PR do?

tell us, what have you changed?

what else?
#### Relevant tickets

Fixes #
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
@@ -86,7 +86,10 @@ class FeatureCard extends React.Component {
 
   signup() {
     const campaign = this.campaigns[this.state.campaignIndex];
-    ga('send', 'event', 'Onboarding', 'Signup');
+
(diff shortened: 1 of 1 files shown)


DFurnes:
Okie dokie.

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -133,7 +136,7 @@ class FeatureCard extends React.Component {
     let newIndex = this.state.campaignIndex + 1;
 
     // Only log to GA if someone actually pressed the button
-    if (buttonTrigger) {
+    if (buttonTrigger && typeof ga !== 'undefined' && ga !== null) {
```
weerd, line 139: `typeof ga !== 'undefined' && ga !== null`.... If this stays as a thing, maybe we turn into a helper ga function to call that returns a boolean so we're not repeating this code in places :)

DFurnes, line 139: Yeah, I'd really like to do that. 👍

PR #6811, opened 2016-08-03
Slide title change.

weerd:
#### What's this PR do?

This PR make a quick change to the title for slide 2.
#### How should this be reviewed?

👀 
#### Relevant tickets

Fixes #6793
#### Checklist
- [ ] Tested on staging.

---

@deadlybutter 

cc: @jessleenyc

PR #6866, opened 2016-08-11
Switches competition messaging to slide 1

### LINKED ISSUE #6864: Update first line of slide one with competitions notification
#### User Story

Similar to what we did for slide 2, let's move the competition notification into slide one underneath the header. 

This line should only pull in if a user joins a competition. If the user does not join, no text should appear. 
#### Why This Matters

Currently, our site doesn't acknowledge competitions so it's important that we get messaging to our members who enter competitions.

### LINKED ISSUE #6865: Update Slide 2 Copy
#### User Story

Now that we've launched slide 1, we should return slide 2 to its normal copy. Please update the header to say "This is how you do it" and remove "Whoohoo! You're signed up." Also remove any text related to competitions. 
#### Why This Matters

If we didn't update the text, once we have both screens up the messaging will be confusing!

itsjoekent:
#### What's this PR do?

see da title
also fixes member count to 5.3 mill

does the competition messaging get added properly

no
#### Relevant tickets

Fixes #6865
Fixes #6864 
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/InstructionSlide.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/InstructionSlide.js
@@ -1,6 +1,6 @@
 const React = require('react');
 
-const title = "Whoo! You're Signed Up.";
+const title = "This is how you do it.";
 const body = "We are PUMPED to see your photos. Complete these simple steps and upload a photo of yourself in action. Your pics will inspire others to join the movement \u2013 so don't forget to share them!";
 
 /**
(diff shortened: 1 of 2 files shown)


lib/themes/dosomething/paraneue_dosomething/js/components/InstructionSlide.js
```diff
@@ -1,6 +1,6 @@
 const React = require('react');
 
-const title = "Whoo! You're Signed Up.";
+const title = "This is how you do it.";
```
weerd, line 3: Didn't we _just_ switch this? https://github.com/DoSomething/phoenix/pull/6811 Is it cause the slide order showing is changing?

itsjoekent, line 3: Cause the slide order is changing :D

weerd, line 3: Lovely.

weerd:
👍

PR #6867, opened 2016-08-11
Update to provide new url helper functions file and us only onboarding.

### LINKED ISSUE #6860: Disable Onboarding for Global English
#### User Story

Currently, onboarding is pulling in campaigns without translations on the featurecard. Let's disable onboarding for global english and create a fix for it after we've shipped all three slides for US english first. 

An example: Smiles for soldiers appeared in onboarding, when I signed up for it I saw this page: 

screen shot 2016-08-10 at 5 38 51 pm

Upon investigation, drupal shows that this campaign isn't translated at all so the campaign should not appear in onboarding. 

screen shot 2016-08-10 at 5 38 57 pm
#### Why This Matters

Signing members up for campaigns that don't exist on global creates a poor user experience. Our current focus is on US so let's disable until we can accommodate.

weerd:
#### What's this PR do?

This PR provides a new `url/helpers.js` file which contains a useful `getPathnameSegments()` function to grab a specified segment from a provided path (or defaults to `window.location.pathname`). Setup as a helper file so it can be used easily in other components!

This allows us to make sure we only show the Onboarding feature to users in the US.

👀  Can test by seeing if you see onboarding while coming from US and not see it if coming from outside the US.
#### Relevant tickets

Fixes #6860
#### Checklist
- [ ] Tested on staging.

---

@DFurnes @deadlybutter

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/url/helpers.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/url/helpers.js
@@ -1,0 +1,14 @@
+/**
+ * Get a specified segment from the pathname.
+ *
+ * @param  {String} pathname
+ * @param  {Number} segment
+ * @return {Bool}
+ */
+export function getPathnameSegment(pathname = window.location.pathname, segmentIndex = 0) {
+  let segments = pathname.split('/');
+
+  segments.splice(0,1);
+
+  return segments[segmentIndex];
+}
--- a/lib/themes/dosomething/paraneue_dosomething/js/url/helpers.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/url/helpers.js
@@ -3,7 +3,7 @@
  *
  * @param  {String} pathname
  * @param  {Number} segment
- * @return {Bool}
+ * @return {Boolean}
  */
 export function getPathnameSegment(pathname = window.location.pathname, segmentIndex = 0) {
   let segments = pathname.split('/');
--- a/lib/themes/dosomething/paraneue_dosomething/js/url/helpers.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/url/helpers.js
@@ -7,8 +7,7 @@
  */
 export function getPathnameSegment(pathname = window.location.pathname, segmentIndex = 0) {
   let segments = pathname.split('/');
-
-  segments.splice(0,1);
+  segmentIndex++;
 
   return segments[segmentIndex];
 }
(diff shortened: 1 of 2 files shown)


lib/themes/dosomething/paraneue_dosomething/js/url/helpers.js
```diff
@@ -0,0 +1,14 @@
...
+ * Get a specified segment from the pathname.
+ *
+ * @param  {String} pathname
+ * @param  {Number} segment
+ * @return {Bool}
+ */
+export function getPathnameSegment(pathname = window.location.pathname, segmentIndex = 0) {
+  let segments = pathname.split('/');
+
+  segments.splice(0,1);
```
itsjoekent: Im confused why this is here?

lib/themes/dosomething/paraneue_dosomething/js/url/helpers.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * Get a specified segment from the pathname.
+ *
+ * @param  {String} pathname
+ * @param  {Number} segment
+ * @return {Bool}
```
DFurnes: `{Boolean}` 👾

lib/themes/dosomething/paraneue_dosomething/js/url/helpers.js
itsjoekent: oh i see, just tried it in my browser console

lib/themes/dosomething/paraneue_dosomething/js/url/helpers.js
weerd: Arg! Laravel PHP docs vs JS docs @_@

lib/themes/dosomething/paraneue_dosomething/js/url/helpers.js
DFurnes: Can we just increase the index by one below, rather than splicing the array?

weerd: Yeah. Was worried that would muddle things, but I guess not really since not returning the full array of segments and just a single piece.

DFurnes:
I feel like relying on the URL path seems a little fragile. Could we store this in the `Drupal.settings` object perhaps?

weerd:
Hmmm, well yeah. Also turns out `Drupal.settings.pathPrefix` is already set in there @_@. Will switch to using that and just leave the URL helper.js in case it's useful @DFurnes

DFurnes:
👍

PR #6868, opened 2016-08-11
Added a script to retrofix user languages.

### LINKED ISSUE #6748: Backfilling the language setting for niche people who were set to GL EN
#### Current Behavior

When users were created via API after we made `global-english` the default language , we have lots of users who have GL EN set as their language, when they're actually US EN.
#### Desired Behavior

Appropriately assign/backfill the language for users who should be US EN instead of GL EN. 

**Discussed Possible Solutions**
1. Look at address w/ Niche people and if they're in the US, assign US EN instead of GL EN. Niche is just in the US, so we can assume members via niche are in the US.
2. Get a subset of the possible sources from Dee for niche users so we can identify those members and assign US EN
3. Using address information for non-Niche people
4. To discuss other potential areas we can accurately determine the source language
#### Why This Matters

Members should have the right site experience

Ref: #6159

angaither:
#### What's this PR do?

Update user language from `en-global` to `en` if
1. we have the address for the user, and they live in the US
2. the user came from niche, or the mobile apps.

🔍 

When we switched the default language from `en` to `en-global` users who were created via the API were assigned the wrong language. This means if they come back to the site, they get a bad experience
#### Relevant tickets

Fixes #6748
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/scripts/correct-language-for-users.php
+++ b/scripts/correct-language-for-users.php
@@ -1,0 +1,30 @@
+<?php
+/*
+ * Script to correct languages on users.
+ *
+ * If we have the address for a user, and it's the US, set the lang to english.
+ * Or if the user comes from niche, or a mobile app, set the lang to english as well.
+ *
+ * drush --script-path=../scripts php-script correct-language-for-users.php
+ *
+ */
+
+
+$global_peeps = db_query("SELECT u.uid, s.field_user_registration_source_value as source, a.field_address_country as country
+                          FROM users u
+                          LEFT JOIN field_data_field_user_registration_source s on s.entity_id = u.uid
+                          LEFT JOIN field_data_field_address a on a.entity_id = u.uid
+                          WHERE u.language = 'en-global';");
+
+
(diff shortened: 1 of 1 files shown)


scripts/correct-language-for-users.php
```diff
@@ -0,0 +1,30 @@
...
+                          FROM users u
+                          LEFT JOIN field_data_field_user_registration_source s on s.entity_id = u.uid
+                          LEFT JOIN field_data_field_address a on a.entity_id = u.uid
+                          WHERE u.language = 'en-global';");
+
+
+$possible_sources = ['niche', 'Nice', 'niche-import-service', 'letsdothis_ios', 'mobileapp_android'];
+
+foreach($global_peeps as $peep) {
+  if ($peep->country == 'US' || in_array($peep->field_address_country, $possible_sources)) {
```
DFurnes: Should this be `in_array($peep->field_source, $possible_sources`?

angaither: 🆒 👀 s

DFurnes:
👍

PR #6870, opened 2016-08-11
Scrub dat mobile data

### LINKED ISSUE #6409: Fix issue where multiple Drupal users share the same email.

angaither:
#### What's this PR do?

Adds a script to cleans up mobile data for users ported from the old days database.

📜 

nah. 
#### Relevant tickets

Fixes #6409
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/scripts/sanitize-mobile-numbers.php
+++ b/scripts/sanitize-mobile-numbers.php
@@ -20,12 +20,12 @@ foreach($wild_typers as $wilder) {
       $user = user_load($wilder->uid);
       $edit = ['field_mobile' => [ LANGUAGE_NONE => [ 0 => [ 'value' => $fresh_and_clean_digits ] ] ] ];
       user_save($user, $edit);
-      print "Updated user " . $wilder->uid . "\n";
+      print 'Updated user ' . $wilder->uid . "\n";
     }
     else {
-      print "something terrible with " . $wilder->uid . "\n";
+      print 'something terrible with ' . $wilder->uid . "\n";
     }
   }
 }
 
-print "done";
+print 'done';
--- a/scripts/sanitize-mobile-numbers.php
+++ b/scripts/sanitize-mobile-numbers.php
@@ -1,0 +1,31 @@
+<?php
+/*
+ * Sanitize mobile numbers on user accounts from ages ago.
+ *
+ *
+ *
+ * drush --script-path=../scripts php-script sanitize-mobile-numbers.php
+ *
+ */
+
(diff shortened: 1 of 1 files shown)


scripts/sanitize-mobile-numbers.php
```diff
@@ -0,0 +1,31 @@
...
+                        FROM field_data_field_mobile
+                        WHERE field_mobile_value NOT REGEXP '^[0-9]+$';");
+
+
+foreach($wild_typers as $wilder) {
+  if ($wilder->mobile) {
+    $fresh_and_clean_digits = dosomething_user_clean_mobile_number($wilder->mobile);
+    if ($fresh_and_clean_digits) {
+      $user = user_load($wilder->uid);
+      $edit = ['field_mobile' => [ LANGUAGE_NONE => [ 0 => [ 'value' => $fresh_and_clean_digits ] ] ] ];
```
DFurnes, line 21: `[ => [ => [ => [ =>`

this is _classic_ droops

DFurnes:
🚢

scripts/sanitize-mobile-numbers.php
angaither, line 21: 😂

weerd, line 21: sweep it under the rug!

PR #6874, opened 2016-08-12
More aggressively try to sanitize bad numbers, and delete any we can'…

DFurnes:
#### What's this PR do?

This PR updates the `sanitize-mobile-numbers` so that it removes any non-numeric characters before sending a malformed mobile field to the `dosomething_user_clean_mobile_number` function. 

This will let us salvage especially nasty fields like `+1234561234` or `123=456-1234`, and it'll just remove the field for that user if it's something _too_ crazy like `DONT- CAL` or `no` (real values).

👀 

🌵 
#### Relevant tickets

References #6409.
#### Checklist
- [ ] Tested on staging.

…t fix.


angaither:
🐑 it

PR #6875, opened 2016-08-15
Progress on slide transitions.

### LINKED ISSUE #6820: Onboarding gallery controller
#### User Story

As a user, I should be able to navigate between the different onboarding slides 
#### Why This Matters

We can't launch with more slides if this does not exist

Only enable 1 slide at a time

weerd:
#### What's this PR do?

This PR includes progress on the slide transitions animation for the Onboarding experience.

Pull down branch and test out the animation.

@deadlybutter talk to @DFurnes so he can help you figure out some of the CSS related to the slide transition. I ran out of time on Friday trying to figure out a specific issue where when the new slide is added in, it bumps the slide you're on, down (since it's added as an element) and then the animation runs. Ideally, instead of having the first slide, bump down, it just moves left with the new one moving in from the left and over.

Other issue, is that we need to add an additional class so that we know when the animation needs to move from right to left (clicking "next'), vs left to right (clicking "prev"). Right now it only moves left to right.
#### Relevant tickets

Fixes #6820
#### Checklist
- [ ] Tested on staging.

---

@deadlybutter @DFurnes 

cc: @jessleenyc

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/Slideshow.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/Slideshow.js
@@ -12,8 +12,7 @@ class Slideshow extends React.Component {
     super(props);
 
     this.state = {
-      currentSlide: 0,
-      totalSlides: this.props.slides.length
+      currentSlide: 0
     };
 
     this.moveBackward = this.moveBackward.bind(this);
@@ -46,11 +45,11 @@ class Slideshow extends React.Component {
 
     return (
       <div className="slideshow">
-        <ReactCSSTransitionGroup className="slideshow__carousel" transitionName="slide" component="div" transitionEnterTimeout={10000} transitionLeaveTimeout={10000}>
+        <ReactCSSTransitionGroup className="slideshow__carousel" transitionName="slide" component="div" transitionEnterTimeout={500} transitionLeaveTimeout={500}>
           {slides[this.state.currentSlide]}
         </ReactCSSTransitionGroup>
 
-        <SlideshowController currentSlide={this.state.currentSlide} totalSlides={this.state.totalSlides} moveBackward={this.moveBackward} moveForward={this.moveForward} />
+        <SlideshowController currentSlide={this.state.currentSlide} totalSlides={this.props.slides.length} moveBackward={this.moveBackward} moveForward={this.moveForward} />
       </div>
     );
   }
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/Slideshow.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/Slideshow.js
@@ -1,7 +1,6 @@
 import SlideshowController from './SlideshowController';
 
 const React = require('react');
-const ReactCSSTransitionGroup = require('react-addons-css-transition-group');
 
 /**
  * Slideshow Component
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/Slideshow.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/Slideshow.js
@@ -1,6 +1,7 @@
-import SlideshowController from './SlideshowController'
+import SlideshowController from './SlideshowController';
 
 const React = require('react');
+const ReactCSSTransitionGroup = require('react-addons-css-transition-group');
 
 /**
  * Slideshow Component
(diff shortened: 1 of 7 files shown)


lib/themes/dosomething/paraneue_dosomething/js/components/Slideshow.js
```diff
@@ -29,9 +46,11 @@ class Slideshow extends React.Component {
 
     return (
       <div className="slideshow">
-        {slides[this.state.activeSlide]}
+        <ReactCSSTransitionGroup className="slideshow__carousel" transitionName="slide" component="div" transitionEnterTimeout={10000} transitionLeaveTimeout={10000}>
```
weerd: I made the timeout `10000` so it's easier to see how it was behaving by moving it slowly. Change this to around `500` once rest of it figured out :)

lib/themes/dosomething/paraneue_dosomething/js/components/SlideshowController.js
```diff
@@ -6,15 +6,18 @@ const React = require('react');
  */
 class SlideshowController extends React.Component {
   render() {
+    const disablePrevButton = this.props.currentSlide === 0 ? true : false;
+    const disableNextButton = this.props.currentSlide === (this.props.totalSlides - 1) ? true : false;
```
DFurnes: Nitpick: the ternary doesn't do anything here (since it's just setting `true` if true, and `false` if false).

DFurnes:
@deadlybutter Definitely up for chatting implementation ideas if you want!

It may end up that we can't use ReactCSSTransitionGroup since we need to animate differently based on the "direction" of the slide offset, but I think you should be able to keep mostly the same declarative approach by applying left/right animations based on the key of the active slide.

itsjoekent:
Yeah I think we'll need to use just Transition Group

itsjoekent:
Ill swing by in a sec

DFurnes:
So recap from our IRL conversation: using [`componentWillReceiveProps`](https://facebook.github.io/react/docs/component-specs.html#updating-componentwillreceiveprops) in a child "slide view" component seems like an elegant solution for this, since it allows us to compare the changed `currentSlide` prop and set a "forward" or "backward" animation state for the `render` method.

itsjoekent:
◀️ ▶️

lib/themes/dosomething/paraneue_dosomething/js/components/Slideshow.js
```diff
@@ -1,6 +1,7 @@
-import SlideshowController from './SlideshowController'
+import SlideshowController from './SlideshowController';
 
 const React = require('react');
+const ReactCSSTransitionGroup = require('react-addons-css-transition-group');
```
itsjoekent: whoops i should remove this!

lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
```diff
@@ -42,19 +42,25 @@
     cursor: pointer;
   }
 }
-
-.slideshow {
-  //
+.slideshow__carousel {
+  overflow: hidden;
```
itsjoekent: this was left over from me experimenting with something, i should just remove the entire selector since it wasnt being used

itsjoekent:
@DFurnes in sprint sync we decided to get the controls launched & do animations in the next PR. I have most of the groundwork setup for the transition anyway in this PR, but the controls work 👍  (gif incoming)

itsjoekent:
I was too lazy to make another gif of the markup changing, but the -left & -right are correctly appended as well, just need to make some animation 🎨

lib/themes/dosomething/paraneue_dosomething/js/components/Slideshow.js
```diff
@@ -11,27 +11,53 @@ class Slideshow extends React.Component {
...
   }
 
   render() {
     const user = this.props.user;
     const campaign = this.props.campaign;
 
     const slides = this.props.slides.map((slide, index) => {
-      return React.createElement(slide, {key: index, campaign: campaign, user: user});
+      const isActive = this.state.currentSlide == index;
+      const displayClasses = ['slideshow__slide'];
```
DFurnes: This seems iffy to me. The slides shouldn't require a `Slideshow` in order to be styled as slides. Ideally the slideshow just passes props (either booleans or enum-style values via a PropTypes contract) rather than explicitly applying classes to its children.

itsjoekent: Yeah thats what i originally did, problems i ran into

1) was not DRY
2) the class string was very messy to look at

I guess we could make a helper to generate the class strings? It just felt overkill for this

DFurnes: > 1) was not DRY

Don't think of DRY as the number of times you type a string. Imagine that we used one of these slide components elsewhere in the app. It shouldn't be necessary to remember that there's a block of code in this parent component that injects specific classes into the child.

itsjoekent: gotcha 👍

DFurnes: I think it could be an indicator that the slide content should be separated from the common "slide" aspects shared between them, like `<Slide><InstructionContent /></Slide>` and `<Slide><ContextContent /></Slide>`. This is definitely something that can be broken into another PR though, if you want.

itsjoekent:
@DFurnes hows that?

DFurnes:
🚢

PR #6880, opened 2016-08-15
Don't try to dedupe the anonymous user.

DFurnes:
#### What's this PR do?

It turns out we have a couple of users (414 of them!) who have an empty-string as their email, which is crazy because I didn't think Drupal allowed that. 👮 Anyways, the existing de-dupe script would try to update the anonymous user which wasn't _at all_ what we wanted, so this fixes that.

👀

With the `AND uid != 0`, this seems to work as expected when tested on Thor.
#### Relevant tickets

References #6409.
#### Checklist
- [ ] Tested on staging.


angaither:
🚢

Issue #6861, opened 2016-08-11
Unified array / helper for disabling features for non US users


itsjoekent:
#### User Story

As a US user, I want to see new experiences, as a global user, I should not as they often are buggy. 
#### Additional Information (optional)

Let's make a Drupal + JS function to verify that the user country is US
We have something like this in onboarding.js `const disabledByLanguage = ['MX', 'BR'].indexOf(user.info.country) != -1;`
#### Why This Matters

We few features either in the works or coming up which might need this
#### Current Workaround w/o Feature (if applicable)

See line of code above ^

Relevant issue: https://github.com/DoSomething/phoenix/issues/6860

itsjoekent:
Potentially easiest option, create a Drupal function in helpers (or maybe in settings with the other lang code grab utilities). In the page preprocessor call that function & embed a JS settings variable like "enable_bleeding_edge". 
Then onboarding would retrieve this setting in the frontend
And any bender work could also retrieve it by calling the helper function

angaither:
refs #6860 refs #6849

angaither:
also refs voter registration modal, and organ donation modals

PR #6881, opened 2016-08-15
voter reg and custom sharing

### LINKED ISSUE #6773: Different flow experience for +18 and under 18

### LINKED ISSUE #6775: User don't keep being asked to register if they hit skip
#### User Story

As an admin, if the user has signed up and get the registration form and they click "skip" we shouldn't give them the form anymore.
#### Why This Matters

If they skip they SKIP.

### LINKED ISSUE #6780: Adding "Join"  competiton modal to the under 18 flow
#### User Story

As an admin, I'd like the "join competition" prompt to come up on screen for those in the under 18 flow  #6773 

Like we talked about, if this is possible, editing the competition modal to have their custom social link in here and that they should share it to join the competition and win things
#### Why This Matters

Getting people to join the competitions can have big impact

Manual competitions

(added this ticket so our meeting/thoughts is captured somewhere!)

### LINKED ISSUE #6832: Social stuff on its own modal
#### User Story

Pull the social stuff (custom social share and social buttons) on its own modal. This modal should be referenced in the Stuff You Need section.

I'm assuming we'll want this same share modal for the over and under 18 experience
#### Why This Matters

Better user flow, can use this for future campaigns.

### LINKED ISSUE #6872: Update custom social share
#### Thing that should happen

Currently the social share link uses the phoenix id as the source.
Since we will be using external auth soon, #4526 we should actually use the Northstar id as the share link. 

`ds.org/campaigns/[campaign-title]?source=user/[northstar_id]`
#### Why This Matters

~the future~

katiecrane:
#### What's this PR do?

Lots of voter reg and custom sharing 🔮 !
- If voter reg is enabled and the user is eligible to register, the voter reg form will popup upon signup AND a link to open the modal is added to "Stuff You Need"
- Custom social share link now uses Northstar ID (uuid) instead of phoenix ID
- If custom social sharing is enabled and voter reg is enabled, we include the custom social link in the signup form modal
- Changed some names of custom social share variables from being specific to organs
- Added function to check if a campaign feature is enabled (for instance, voter reg or custom social sharing)
- If custom social sharing is enabled, include a link to the share modal in "Stuff You Need" instead of just text including the custom link

Peruse for logic flaws.

Make sure that all features worked as described above. Be sure to test on campaigns with custom sharing and voter reg both disabled and enabled, and with one on and the other off. 

You know how sometimes in Drupal you work on something for a long time and then there is a lot less actual code than it felt like? Everything is also very closely related so I had trouble trying to break it out into separate PRs 😬 
#### Relevant tickets

Fixes #6780
Fixes #6775
Fixes #6773
Fixes #6832
Fixes #6872
#### Checklist
- [ ] Tested on staging.
- [ ] Pinged a PM if this is a larger PR that would benefit from some additional testing love. @ngjo

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -1205,10 +1205,20 @@ function dosomething_campaign_node_validate($node, $form, &$form_state) {
    *   Campaign variables
    * @param string $feature
    *   Checking if this feature is on
+   * @param bool $us_only
+   *   If true, make sure the user is in the US
+   *
+   * @return bool
    */
-  function dosomething_campaign_feature_on($campaign, $feature) {
+  function dosomething_campaign_feature_on($campaign, $feature, $us_only = FALSE) {
     if (array_key_exists($feature, $campaign->variables) && $campaign->variables[$feature]) {
-      return TRUE;
+      if ($us_only && dosomething_global_get_current_prefix() == 'us') {
+        return TRUE;
+      }
+      elseif (!$us_only) {
+        return TRUE;
+      }
     }
+
     return FALSE;
   }
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -1199,3 +1199,16 @@ function dosomething_campaign_node_validate($node, $form, &$form_state) {
     }
   }
 }
+
+  /**
+   * @param object $campaign
+   *   Loaded campaign node to check if given feature is on
+   * @param string $feature
+   *   Checking if this featuer is on
+   */
+  function dosomething_campaign_feature_on($campaign, $feature) {
+    if (array_key_exists($feature, $campaign->variables) && $campaign->variables[$feature]) {
+      return TRUE;
+    }
+    return FALSE;
+  }
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -1202,9 +1202,9 @@ function dosomething_campaign_node_validate($node, $form, &$form_state) {
 
   /**
    * @param object $campaign
-   *   Loaded campaign node to check if given feature is on
+   *   Campaign variables
    * @param string $feature
-   *   Checking if this featuer is on
+   *   Checking if this feature is on
    */
   function dosomething_campaign_feature_on($campaign, $feature) {
     if (array_key_exists($feature, $campaign->variables) && $campaign->variables[$feature]) {
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -439,7 +439,7 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
   $vars['register_voters'] = dosomething_campaign_feature_on($campaign, 'enable_voter_registration') ? true : false;
 
   // If we are registering voters check to see if the user will be 18 by the election
-  if ($vars['register_voters']) {
+  if ($vars['register_voters'] && (dosomething_global_get_current_prefix() == 'us')) {
     // Calculate the user's age on election day
     $election_day_age = dosomething_user_age_on_date('Nov 8 2016');
     $vars['can_vote'] = (!($election_day_age < 18)) ? true : false;
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -436,10 +436,10 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
   $vars['custom_social_share_link'] = $custom_social_share_link;
 
   // Pass along whether or not voter registration is enabled
-  $vars['register_voters'] = dosomething_campaign_feature_on($campaign, 'enable_voter_registration') ? true : false;
+  $vars['register_voters'] = (dosomething_campaign_feature_on($campaign, 'enable_voter_registration') && dosomething_global_get_current_prefix() == 'us') ? true : false;
 
   // If we are registering voters check to see if the user will be 18 by the election
-  if ($vars['register_voters'] && (dosomething_global_get_current_prefix() == 'us')) {
+  if ($vars['register_voters']) {
     // Calculate the user's age on election day
     $election_day_age = dosomething_user_age_on_date('Nov 8 2016');
     $vars['can_vote'] = (!($election_day_age < 18)) ? true : false;
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -213,7 +213,7 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
   // 1) Materials
 
   // Build the unique share link.
-  $share_path = 'user/' . $vars['user']->uuid;
+  $share_path = 'user/' . dosomething_user_get_field('field_northstar_id');
   $custom_social_share_link = dosomething_global_url(current_path(), array('absolute' => TRUE,'query' => array('source' => $share_path)));
 
   if (isset($campaign->items_needed)) {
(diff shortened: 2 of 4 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -213,17 +213,17 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
   // 1) Materials
 
   // Build the unique share link.
-  $share_path = 'user/' . $vars['user']->uid;
+  $share_path = 'user/' . $vars['user']->uuid;
```
angaither: `uuid` is actually not the same as NS id. it's a drupaly thing
`field_data_field_northstar_id` is where you wanna grab the right id

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -1199,3 +1199,16 @@ function dosomething_campaign_node_validate($node, $form, &$form_state) {
...
   }
 }
+
+  /**
+   * @param object $campaign
+   *   Loaded campaign node to check if given feature is on
+   * @param string $feature
+   *   Checking if this featuer is on
+   */
+  function dosomething_campaign_feature_on($campaign, $feature) {
```
angaither: yais!

another thing we might want to pull into this is if the feature is enabled for a more than US experiences see #6861

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -1199,3 +1199,16 @@ function dosomething_campaign_node_validate($node, $form, &$form_state) {
     }
   }
 }
+
+  /**
+   * @param object $campaign
+   *   Loaded campaign node to check if given feature is on
+   * @param string $feature
+   *   Checking if this featuer is on
```
angaither: typo `feature`

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -1199,3 +1199,16 @@ function dosomething_campaign_node_validate($node, $form, &$form_state) {
     }
   }
 }
+
+  /**
+   * @param object $campaign
+   *   Loaded campaign node to check if given feature is on
```
angaither: smmmalll thing, but I don't think the `campaign` is actually the node. I think it's the campaign class object from [`Campaign::get`](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php#L53) I think passing the node object here would break things

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -409,37 +409,34 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
 
   // Pass custom share links to the front end. Unique by UID.
   $base_url = url(base_path(), array('absolute'=> TRUE, 'language' => 'en-global'));
-  $country_prefix_for_organ_donation_share = (dosomething_global_get_current_prefix()) ? strtoupper(dosomething_global_get_current_prefix()) : 'global';
-  $organ_donation_share_image = $base_url . 'sites/default/files/images/problem-' . $campaign->nid . '-' . $country_prefix_for_organ_donation_share . '.png';
+  $country_prefix_for_share = (dosomething_global_get_current_prefix()) ? strtoupper(dosomething_global_get_current_prefix()) : 'global';
+  $share_image = $base_url . 'sites/default/files/images/problem-' . $campaign->nid . '-' . $country_prefix_for_share . '.png';
```
angaither, line 413: 💭 hmmm does this override the problem shares (f that, share this stat) or whatever? I thought we didn't want to do that?

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -173,8 +173,10 @@
                     <li><a href="#" data-modal-href="#modal-shipment-form"><?php print $shipment_form_link; ?></a></li>
                   <?php endif; ?>
 
-                  <?php if (array_key_exists('enable_voter_registration', $campaign->variables) && $campaign->variables['enable_voter_registration']) : ?>
-                    <li><a href="#" data-modal-href="#modal-voter-registration">Register to Vote</a></li>
+                  <?php if (dosomething_campaign_feature_on($campaign, 'enable_voter_registration')) : ?>
```
angaither: the fact that this feature is enabled should be stored in the `vars` array so you don't have to check if it's enabled a bunch of times in the theme

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -391,33 +403,30 @@
...
-              <iframe src="https://register2.rockthevote.com/registrants/map/?source=iframe&partner=35164" width="100%" height="1200" marginheight="0" frameborder="0"></iframe>
-            </div>
-          </div>
-        <?php else : ?>
-          <?php // @TODO: fix the styling in hurr ?>
+
+        <?php // Voter Registration Modal // ?>
+        <?php if (dosomething_campaign_feature_on($campaign, 'enable_voter_registration')) : ?>
           <div data-modal id="modal-voter-registration" role="dialog">
+            <iframe src="https://register2.rockthevote.com/registrants/map/?source=iframe&partner=35164" width="100%" height="1000" marginheight="0" frameborder="0"></iframe>
```
angaither: might be worth sending the modal/signup form to the theme so we don't have to repeat this chuck here and above

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
katiecrane, line 413: it sure doesn't! this is a separate thing that was made originally for organs. all I did here was change the variable names from being organ specific, which maybe we don't want that? but the regular problem shares happen [here](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc#L119-L153)

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
katiecrane: @angaither is it safe to assume that everyone has a northstar id?

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
katiecrane: yeah I agree it's not actually the node. it's like the campaign variables of the node, which might be equivalent to what you said but I'm not positive. regardless, not the node.

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
katiecrane: SO MUCH NICER

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
angaither: @katiecrane moooooost do. but no, not everyone.

katiecrane: @angaither do you happen to know what we would want the link to be if they don't have one?

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -220,10 +220,20 @@
...
+          <?php endif; ?>
+          <?php if ($register_voters && dosomething_campaign_feature_on($campaign, 'social_share_unique_link')) : ?>
+            <div class="modal__block">
+              <p>Copy and paste your share link:</p>
+              <code><?php print $custom_social_share_link ?></code>
+            </div>
+          <?php endif; ?>
           <?php if (isset($skip_signup_data_form)): ?>
-          <div><?php print render($skip_signup_data_form); ?></div>
+            <div><?php print render($skip_signup_data_form); ?></div>
```
angaither, line 236: 👾

angaither:
this probably needs to only be enabled on `/us/` and not other experiences 🇺🇸 ☑️

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
katiecrane, line 236: wait don't we want that to be indented like that though since it's in an if?

angaither, line 236: mah bad 👀

katiecrane, line 236: they made it hard to read 👓

katiecrane:
@angaither ^ is that the best way to do enable only on `/us/`?

angaither:
@katiecrane hmm @deadlybutter might have some tips for this

itsjoekent:
seems good to me 👍 

i thought we stored the lang code in the preprocess but apparently not, so this is fine

katiecrane:
the new commit sets `register_voters` to false if not in the US since, instead of just checking if they are in the US in that one `if` statement

angaither:
@katiecrane sure sure. 
Ideally we will have a list of features (#6861) that are enabled only in the us, so we don't need to do this evrrrrywhere. and that I could see being useful in the `dosomething_campaign_feature_on` function

katiecrane:
@angaither AH I see what you're saying. I can add an optional parameter to `dosomething_campaign_feature_on` to also check if the user is on `/us/`

angaither:
🍧 🇺🇸

PR #6887, opened 2016-08-16
Really, really sanitize those mobiles.

### LINKED ISSUE #6886: Strange behavior with the sanitize-mobile-numbers script.
#### Current Behavior

Running the `sanitize-mobile-numbers` script seems to iterate over all users with a mobile field (rather than just the "bad" ones). Interestingly, running the same exact query (see below) from Sequel Pro returns only the expected "bad" user rows.

The script _does_ update the iterated user's mobile fields with cleaned numbers.
#### Desired Behavior

It should find users with poorly formatted mobile numbers (anything that's not a string of exactly 10 digits), and process them with the `dosomething_user_clean_mobile_number` function.
#### Steps to Replicate

Run `drush --script-path=../scripts php-script sanitize-mobile-numbers.php` on Thor. It will iterate over all users with a mobile field (rather than just ones with a broken mobile field).

The script on Thor has a slightly modified query (which will also include numeric fields with the wrong length), which I'll put in a PR once we've figured out the problem outlined above.

``` diff
 $wild_typers = db_query("SELECT entity_id as uid, field_mobile_value as mobile
                         FROM field_data_field_mobile
-                        WHERE field_mobile_value NOT REGEXP '^[0-9]+$';");
+                        WHERE field_mobile_value NOT REGEXP '^[0-9]{10}$';");
```
#### Why This Matters

The sanitize mobile numbers script should only sanitize numbers that are not sanitary, not ones that are. Since it seems to be iterating over _all_ mobile fields, it's hard to verify that it's worked (and will take significantly longer to finish).

I'll send screenshots over Slack, since they contain phone numbers.

DFurnes:
#### What's this PR do?

This PR uses a bastardized form of the `^[0-9]{10}$` regex used in the issue, since we can't apparently use brackets with `db_query` without them getting find-replaced as table prefixes.

It also adds logging of "before and after" values for sanitized numbers so we can easily verify.

🌵. Blahhhh. It's not pretty, but at least it seems to work.

🐮 
#### Relevant tickets

Fixes #6886.
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/scripts/sanitize-mobile-numbers.php
+++ b/scripts/sanitize-mobile-numbers.php
@@ -14,8 +14,8 @@ $wild_typers = db_query('SELECT entity_id as uid, field_mobile_value as mobile
 
 foreach($wild_typers as $wilder) {
   if ($wilder->mobile) {
-    $mobile = preg_replace('[^0-9]+', '', $wilder->mobile);
-    $fresh_and_clean_digits = dosomething_user_clean_mobile_number($mobile);
+    $mobile = $wilder->mobile;
+    $fresh_and_clean_digits = dosomething_user_clean_mobile_number(preg_replace('[^0-9]', '', $mobile));
     if ($fresh_and_clean_digits) {
       print 'Updated user ' . $wilder->uid . "\n";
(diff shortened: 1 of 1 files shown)


angaither:
🚢

PR #6888, opened 2016-08-16
Updating onboarding script to account for possible null user.info.

### LINKED ISSUE #6884: Onboarding not appearing - DS user id is null
#### Current Behavior

Onboarding not appearing for some users. Looks like their DS user ids are coming back as null -- could be a Northstar sync issue. 
#### Desired Behavior

Onboarding should appear for all members. 
#### Why This Matters

We want all users to see onboarding so they have the opportunity to sign up for additional campaigns. 

screen shot 2016-08-16 at 9 55 46 am
screen shot 2016-08-16 at 9 56 55 am

weerd:
#### What's this PR do?

This PR adds a fix to account for the possibility that the `user.info` returned from Northstar _might_ have responded with an error and thus the value is `null`. If so, let's not rely on having `user.info.id` be present and instead handle that a little better.

If you get a `null` response from Northstar (you can try and fake it for testing), do the Onboarding experience still show for a logged in user?

Used `Object.keys()` since we're also added the campaign run to the localStorage key. However, these run ids are nested within objects whose keys are the language codes available for a translated run. To help avoid errors with adding the campaign run ID, I'm nabbing the first item with a key (typically `en`) and grabbing that run ID.
#### Relevant tickets

Fixes #6884
#### Checklist
- [ ] Tested on staging.

---

@DFurnes @deadlybutter 

cc: @jessleenyc

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
@@ -10,9 +10,13 @@ const ReactDom = require('react-dom');
 const Modal = require('dosomething-modal');
 
 function getLocalStorageKey(user, campaign) {
+  // @TODO: probably need a better way to get the run ID consistently.
+  const languageCode = (Object.keys(campaign.campaign_runs.current)[0]);
+
   let key = null;
+
   if (campaign && user) {
-    key = `onboarding:campaign=${campaign.id}-user=${user.info.id}`;
+    key = `onboarding:campaign=${campaign.id}&run=${campaign.campaign_runs.current[languageCode].id}${user.info ? "&user=" + user.info.id :  ''}`;
   }
 
   return key;
--- a/lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
@@ -10,13 +10,10 @@ const ReactDom = require('react-dom');
 const Modal = require('dosomething-modal');
 
 function getLocalStorageKey(user, campaign) {
-  // @TODO: probably need a better way to get the run ID consistently.
-  const languageCode = (Object.keys(campaign.campaign_runs.current)[0]);
-
   let key = null;
 
   if (campaign && user) {
-    key = `onboarding:campaign=${campaign.id}&run=${campaign.campaign_runs.current[languageCode].id}${user.info ? "&user=" + user.info.id :  ''}`;
+    key = `onboarding:campaign=${campaign.id}&user=${user.info ? user.info.id :  'null'}`;
   }
 
   return key;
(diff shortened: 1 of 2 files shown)


lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
```diff
@@ -10,9 +10,13 @@ const ReactDom = require('react-dom');
 const Modal = require('dosomething-modal');
 
 function getLocalStorageKey(user, campaign) {
+  // @TODO: probably need a better way to get the run ID consistently.
+  const languageCode = (Object.keys(campaign.campaign_runs.current)[0]);
```
weerd: Not a fan of this but seemed to work pretty well and more reliable than just relying on `campaign.campaign_runs.current.en.id`... thoughts?

itsjoekent: hmmm, is `en` always the top one? feels sketchy

lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
```diff
@@ -10,9 +10,13 @@ const ReactDom = require('react-dom');
...
+
   let key = null;
+
   if (campaign && user) {
-    key = `onboarding:campaign=${campaign.id}-user=${user.info.id}`;
+    key = `onboarding:campaign=${campaign.id}&run=${campaign.campaign_runs.current[languageCode].id}${user.info ? "&user=" + user.info.id :  ''}`;
```
itsjoekent: Feels weird to me having two types of keys looking at it, should it just be one or the other? or even something as simple as this?

`campaign=${campaign.id}&user=${user.info ? user.info.id :  -1}`

lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
weerd: Well, the Onboarding isn't happening for non-US, so yeah. I know still not ideal, but it's some duct-tape in the meantime.

lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
weerd: Well, if the campaign happens again and user hasn't cleared localStorage and different run ID then no onboarding then... but maybe that's a non-issue...

weerd: Maybe we don't really need the run ID at the moment 🤔

weerd:
@deadlybutter decided to best go the simple route. good call. we can wait and see if campaign runs are necessary if Onboarding is a feature that sticks ;)

itsjoekent:
sweets 
🚂 🚋 🚋

PR #6890, opened 2016-08-16
Potentially cleaner dedupe and one bugfix

DFurnes:
#### What's this PR do?

This is a potentially mistaken optimization and a bug fix.

💹 **The potentially mistaken optimization:** I was thinking it might make more sense to use the `PUT /v1/users/drupal_id/:uid` route to make the corresponding change in Northstar each time we update a user in Phoenix. It'd mean we wouldn't have to build up a queue of "borked" accounts to delete and re-create.

This _should_ be a safe call since each of these scripts should always leave us with valid "updates" for Northstar's data (if run in this order `remove-duplicate-emails` → `unset-duplicate-mobiles` → `sanitize-mobile-numbers` → `unset-duplicate-mobiles`).

And the `dosomething_northstar_update_user` method already falls back to using the `POST /v1/users` endpoint if the given Drupal ID returns a 404.

🐛 **The bug fix:** the `dosomething_northstar_register_user` was logging an error to Watchdog on successful responses. It should do the _opposite_ of that.

Let's refresh Thor Phoenix & Thor Northstar from prod, and then try running the modified scripts from this branch. If we're faced with a ton of errors then it's probably not a good improvement. 😇 

🌅
#### Relevant tickets

References #6409.
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -92,7 +92,7 @@ function dosomething_northstar_register_user($user, $payload) {
   // Add to request log if enabled.
   dosomething_northstar_log_request('register_user', $user, $payload, $response);
 
-  if (_dosomething_northstar_is_successful_response($response)) {
+  if (! _dosomething_northstar_is_successful_response($response)) {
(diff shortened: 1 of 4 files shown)


angaither:
a very reasonable pull request. 🐦 🔀 🌟

Issue #6666, opened 2016-06-30
Display reportback photos from current campaign during onboarding


angaither:
#### User Story

As a user, when I reach the end of the onboarding experience I see reportback photos from the campaign I'm currently signed up for. 
#### Additional Information (optional)
- ~~Display at least one promoted photo, 4 photos total~~ see #6909
- [x] The rest of the photos are random, approved or promoted photos
- ~~You can kudos these photos~~ see #6910

PR #6893, opened 2016-08-17
ApiClient class & Reportback Items Slide

weerd:
#### What's this PR do?

This PR adds a new `ApiClient` class to help remove the need to duplicate A LOT of API related code, and universalize how we do API calls to our services 🎉, based on a couple conversations that @DFurnes and I had as we saw how we're doing more API calls in JS and none of them are standardized/repeating code. This leads to more bugs instead of having a single class with sole purpose to handle API calls.

Currently the API calls only apply to Phoenix, but can eventually be made to apply to our other services. It uses the `whatwg-fetch` library recommended by @DFurnes which is super easy to use and use the modern approach to making API calls on the web with JS Promises.

This PR also includes a new `ReportbackItem` component that we can ideally have handle all representations of a Reportback Items across the site. It is used specifically for Slide 3 of the Onboarding feature, which is also built out in this PR.

screen shot 2016-08-18 at 3 45 30 pm

Download the code and run it. Make sure the other slides in Onboarding didn't break, and then check the 3rd slide to see that all looks as expected.

The ReportbackItems outputted in Slide 3 do not include Reactions on them. The data is available and can be easily output, but I need to add a `post()` method to the `ApiClient` class to support registering these reactions. That is next on my list.

The way the RB items are retrieved, in this PR but will be updated soon. It grabs 4 RB items for specified campaign from promoted or approved RBs and does so at random.
#### Relevant tickets

Refs #6666
#### Checklist
- [ ] Tested on staging.

---

@deadlybutter @DFurnes 

cc: @jessleenyc

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc
@@ -327,8 +327,8 @@ function dosomething_helpers_add_js_onboarding() {
       $onboardingSettings['slides'][] = 'InstructionSlide';
     }
 
-    if (variable_get('dosomething_campaign_onboarding_reportback_item_kudos_slide', FALSE)) {
-      $onboardingSettings['slides'][] = 'ReportbackItemKudosSlide';
+    if (variable_get('dosomething_campaign_onboarding_reportback_items_slide', FALSE)) {
+      $onboardingSettings['slides'][] = 'ReportbackItemsSlide';
     }
 
     drupal_add_js(['dsOnboarding' => $onboardingSettings], 'setting');
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
@@ -59,8 +59,6 @@ class FeatureCard extends React.Component {
   }
 
   getCampaigns(staffPickOnly, paginate, page, callback) {
-    console.log('requesting...');
-
     this.requests.push(this.serverRequest = $.get(`${this.apiClient.url}campaigns?${staffPickOnly ? 'staff_pick=true&' : ''}${page != -1 ? 'page=' + page : ''}`, result => {
       // Get campaigns ordered by relevance & handle exclusions.
       const campaigns = sortCampaignsByRelevance(result.data).filter(campaign => this.campaignsToExclude.indexOf(parseInt(campaign.id)) == -1);
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
@@ -19,7 +19,7 @@ class FeatureCard extends React.Component {
     this.campaigns = [];
     this.campaignsToExclude = setting('dsUser.activity.signups', []);
     this.campaignsApiPage = 1;
-    this.apiClient = new ApiClient(window.location.hostname, '/api/v1/', window.location.port);
+    this.apiClient = new ApiClient('v1');
 
     this.state = {
       actionDisabled: false,
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
@@ -5,6 +5,8 @@ import Button from './Button';
 import setting from '../utilities/Setting';
 import { sortCampaignsByRelevance } from '../campaign/helpers';
 
+import ApiClient from '../utilities/ApiClient';
+
 const API_HOST = window.location.hostname;
 const API_PORT = window.location.port || undefined;
 const API_VERSION = '/api/v1/';
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
@@ -3,14 +3,8 @@ const React = require('react');
 
 import Button from './Button';
 import setting from '../utilities/Setting';
-import { sortCampaignsByRelevance } from '../campaign/helpers';
-
 import ApiClient from '../utilities/ApiClient';
-
-const API_HOST = window.location.hostname;
-const API_PORT = window.location.port || undefined;
-const API_VERSION = '/api/v1/';
-const API_URL = `//${API_HOST}${API_PORT ? ':' + API_PORT : ''}${API_VERSION}`;
+import { sortCampaignsByRelevance } from '../campaign/helpers';
 
 /**
  * Card Component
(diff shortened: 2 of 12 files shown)


lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -23,6 +19,7 @@ class FeatureCard extends React.Component {
     this.campaigns = [];
     this.campaignsToExclude = setting('dsUser.activity.signups', []);
     this.campaignsApiPage = 1;
+    this.apiClient = new ApiClient(window.location.hostname, '/api/v1/', window.location.port);
```
itsjoekent: Do you ever see a scenario where we don't pass in these `window.location.hostname` & `window.location.port`? I feel like those could potentially be defaults

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -54,7 +51,7 @@ class FeatureCard extends React.Component {
   }
 
   getCampaigns(staffPickOnly, paginate, page, callback) {
-    this.requests.push(this.serverRequest = $.get(`${API_URL}campaigns?${staffPickOnly ? 'staff_pick=true&' : ''}${page != -1 ? 'page=' + page : ''}`, result => {
+    this.requests.push(this.serverRequest = $.get(`${this.apiClient.url}campaigns?${staffPickOnly ? 'staff_pick=true&' : ''}${page != -1 ? 'page=' + page : ''}`, result => {
```
itsjoekent, line 54: little confused, wasn't the point of writing the new Api Client to replace these get requests?

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
weerd: Yeah, the `port` does have a default in the ApiClass constructor and was kinda wondering if the host one should default to the `window.location.hostname` as well. Maybe even the `version` has a default too? Wasn't sure how crazy to get.

The port made the most sense for sure. If it's not present it's passed as an empty string to the constructor and then I do a check there to normalize.

lib/themes/dosomething/paraneue_dosomething/js/onboarding.js
```diff
@@ -9,6 +9,13 @@ const React = require('react');
 const ReactDom = require('react-dom');
 const Modal = require('dosomething-modal');
 
+/**
+ * Get the local storage key for Onboarding.
+ *
+ * @param  {Object} user
+ * @param  {Object} campaign
+ * @return {String}
+ */
```
deadlybutter: whoops mah bad

itsjoekent, line 18: wait yeah, i think i wrote this, whoops!

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
weerd, line 54: Yes, and I tested the calls with simpler stuff via the `FeatureCard` but wanted to do the full switcheroo in a different PR to minimize the testing and changes. Upcoming PR will have those changed. Mostly setting the groundwork for the `ApiClient` class and defs used in `ReportbackItemKudos` component... and slowly extending it out elsewhere ;)

lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
```diff
@@ -0,0 +1,106 @@
...
+  get(path, query = {}, callback) {
+    const url = `${this.url}${path}${this.stringifyQuery(query)}`;
+
+    fetch(url)
+      .then(this.checkStatus)
+      .then(this.parseJson)
+      .then(function(data) {
+        return callback(data);
+      }).catch(function(data) {
+        console.log('um, there was an error! handle that shizzzzz!');
```
itsjoekent, line 85: 😆

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
itsjoekent: I think version could remain user defined, in the event we ever start using /v2/

itsjoekent:
lots of changes, i think overall its ready to go. should we make an issue for finishing the api client work?

itsjoekent:
gotta implement the post request & converting the remaining requests

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
weerd: Ya. Was also debating a bit, whether we let the dev specify full thing `/api/v1/` or just a `v1` and we fill out the rest in the constructor when building the url? I see benefits with either... not something that can't be easily changed anyhoo.

weerd: also means I'd likely change the argument order from:

``` javascript
(host, version, port = null)
```

to: 

``` javascript
(version, host = window.location.hostname, port = null)
```

weerd:
Yeah. I'm going to go ahead and make the issues soon :)

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
itsjoekent: I think that order makes sense, also im 👍 to just doing `v1`, `v2` etc

weerd: Cool. I'll make those updates right now and push them up!

weerd:
@deadlybutter updated! Just check the last [two commits](https://github.com/DoSomething/phoenix/pull/6893/commits) if that's easier! Made some naming udpates since Kudos are now called Reactions, but technically it's a slide with RB items, that just happen to have reactions on them ;)

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc
```diff
@@ -327,8 +327,8 @@ function dosomething_helpers_add_js_onboarding() {
       $onboardingSettings['slides'][] = 'InstructionSlide';
     }
 
-    if (variable_get('dosomething_campaign_onboarding_reportback_item_kudos_slide', FALSE)) {
-      $onboardingSettings['slides'][] = 'ReportbackItemKudosSlide';
+    if (variable_get('dosomething_campaign_onboarding_reportback_items_slide', FALSE)) {
```
itsjoekent, line 330: not needed for this PR, but a more dynamic way of doing this feels like a good idea

itsjoekent:
yeah last two commits look good

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc
weerd, line 330: yar!

weerd:
💃

PR #6900, opened 2016-08-18
Adds more onboarding options

### LINKED ISSUE #6817: Create a field that can toggle onboarding by campaign.
#### User Story

As an admin, I want the ability to turn off onboarding for specific campaigns. 

Instead of doing a one off for voter registration, we're creating a setting so we can turn off onboarding for future campaigns (like organ donation) if desired.

itsjoekent:
#### What's this PR do?
- Adds a developer override mode. This is because we're constantly clearing local storage & commenting out code when developing so we can refresh our changes.
- Adds a per campaign disable option. This is for vcard and other campaigns in the future which have a more in depth signup process. 

If you enable onboarding dev, does on boarding show up every time?
If you disable ^, then disable onboarding, does onboard show up?
If you enable onboarding, then disable it on a campaign, does it show up?
If you enable it on the campaign, does it show up?

Everything is described above I think
#### Relevant tickets

Fixes #6817 
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
@@ -323,10 +323,6 @@ function dosomething_campaign_admin_custom_settings_page($node) {
   $third_party_form = drupal_get_form($form_id, $node);
   $output .= render($third_party_form);
 
-  $form_id = 'dosomething_campaign_onboarding_config_form';
-  $onboarding_config_form = drupal_get_form($form_id, $node);
-  $output .= render($onboarding_config_form);
-
   if (module_exists('dosomething_signup')) {
     $signup_data_config_form = drupal_get_form('dosomething_signup_node_signup_data_form', $node);
     $output .= render($signup_data_config_form);
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
@@ -89,6 +89,13 @@ function dosomething_campaign_admin_config_form($form, &$form_state) {
     '#default_value' => variable_get('dosomething_campaign_enable_onboarding'),
   );
 
+  $form['campaign']['action_page']['dosomething_campaign_onboarding_developer_mode'] = [
+    '#type' => 'checkbox',
+    '#title' => t('Onboarding developer mode'),
+    '#description' => t('Obviously don\'t enable this on prod'),
+    '#default_value' => variable_get('dosomething_campaign_onboarding_developer_mode'),
+  ];
+
   $form['campaign']['action_page']['dosomething_campaign_onboarding_context_slide'] = [
     '#type' => 'checkbox',
     '#title' => t('Display Context Slide'),
@@ -315,6 +322,10 @@ function dosomething_campaign_admin_custom_settings_page($node) {
   $form_id = 'dosomething_helpers_third_party_variable_form';
   $third_party_form = drupal_get_form($form_id, $node);
   $output .= render($third_party_form);
+
+  $form_id = 'dosomething_campaign_onboarding_config_form';
+  $onboarding_config_form = drupal_get_form($form_id, $node);
+  $output .= render($onboarding_config_form);
 
   if (module_exists('dosomething_signup')) {
     $signup_data_config_form = drupal_get_form('dosomething_signup_node_signup_data_form', $node);
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc
@@ -305,12 +305,12 @@ function dosomething_helpers_add_js_campaign_and_user_data($vars) {
  * @return void
  */
 function dosomething_helpers_add_js_onboarding($nid) {
-  $global_enable = variable_get('dosomething_campaign_enable_onboarding', FALSE);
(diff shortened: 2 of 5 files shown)


lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc
```diff
@@ -304,9 +304,13 @@ function dosomething_helpers_add_js_campaign_and_user_data($vars) {
  *
  * @return void
  */
-function dosomething_helpers_add_js_onboarding() {
+function dosomething_helpers_add_js_onboarding($nid) {
+  $globalEnable = variable_get('dosomething_campaign_enable_onboarding', FALSE);
```
angaither: `snake_case` in drooopz 🐘

itsjoekent: table flip

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
```diff
@@ -316,6 +323,10 @@ function dosomething_campaign_admin_custom_settings_page($node) {
   $third_party_form = drupal_get_form($form_id, $node);
   $output .= render($third_party_form);
 
+  $form_id = 'dosomething_campaign_onboarding_config_form';
```
angaither: curious why this is was just added? how did it work before without it?

itsjoekent: old code from the first way I tried doing it, good catch

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc
```diff
@@ -304,9 +304,13 @@ function dosomething_helpers_add_js_campaign_and_user_data($vars) {
  *
  * @return void
  */
-function dosomething_helpers_add_js_onboarding() {
+function dosomething_helpers_add_js_onboarding($nid) {
+  $global_enable = variable_get('dosomething_campaign_enable_onboarding', FALSE);
```
weerd: `global_enable` sounds like it could be confusing... can we just call it `onboarding_enabled`. And on that regard, nitpicky, but can we either go with `noun_enabled` or `enable_noun`, where _noun_ is either `developer`, `onboarding`, or `campaign`. Again the latter two just sound better when reading it with an associated `true/false` boolean.

weerd:
So does it have to be something that is per campaign? Couldn't we just have an "Enable Debug Mode" globally, and then in each respective feature we build we check if it is on or not? Otherwise to add more Debug Modes for other features we always need to add a form item for each one. Unless I'm misunderstanding something. 

Ideally it's something we can just toggle on and off for the site and each dev can use at their discretion whether they need it for testing their particular feature or not.

itsjoekent:
@weerd There is 3 options now with this PR

1) Allow admins to turn onboarding on / off for the entire site
2) Allow admins to disable onboarding for a specific campaign (EG: Voter reg)
3) Allow developers to override all of this plus local storage, to make our lives easier when refreshing the page.

I'll rename the variables now!

itsjoekent:
@weerd should be good to go now, everything appears to still work

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc
```diff
@@ -304,9 +304,13 @@ function dosomething_helpers_add_js_campaign_and_user_data($vars) {
  *
  * @return void
  */
-function dosomething_helpers_add_js_onboarding() {
+function dosomething_helpers_add_js_onboarding($nid) {
+  $onboarding_enabled = variable_get('dosomething_campaign_enable_onboarding', FALSE);
+  $developer_mode_enabled = variable_get('dosomething_campaign_onboarding_developer_mode', FALSE);
```
weerd, line 309: 💃 reads so lovely 😄

itsjoekent, line 309: 👀 🎢

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc
```diff
@@ -315,8 +319,10 @@ function dosomething_helpers_add_js_onboarding() {
     ]);
 
     $onboardingSettings = [
-        'enabled' => true,
-        'slides' => [],
+      'globalEnabled' => $onboarding_enabled,
+      'campaignEnabled' => $campaign_enabled,
+      'developerEnabled' => $developer_mode_enabled,
```
weerd: `developerModeEnabled`

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc
```diff
@@ -315,8 +319,10 @@ function dosomething_helpers_add_js_onboarding() {
     ]);
 
     $onboardingSettings = [
-        'enabled' => true,
-        'slides' => [],
+      'globalEnabled' => $onboarding_enabled,
```
weerd: `globallyEnabled`

weerd:
2 naming things. once set 👍

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.theme.inc
itsjoekent: I changed it to onboardingEnabled cause I did it before your comment, I assume thats alright lol

PR #6901, opened 2016-08-18
fix signup source from social share

### LINKED ISSUE #6891: Signup source is saving utm params as well
#### Current Behavior

Currently when we store signup source it's saved with all the UTM params if they are there. we don't want that
`user/[northstar_id&utm=blahblahblah`
#### Desired Behavior

cut all sources at the `&` so we don't save UTM params in the signup source
#### Steps to Replicate
- signup with a source and utm params, you'll see a lot of nonsense in the table

katiecrane:
#### What's this PR do?

Previously, when building social share buttons to link to the campaign with the custom social share source parameter, the link was built incorrectly. When we added other utm parameters associated with social shares, we assumed there were no other parameters and tacked on a `?` to the url regardless of if we already had one or not, which meant that gobbledigook was getting saved in the source column. NOW we give the function that adds the utm parameters the raw campaign link and tack on the custom source after the utm parameters if there is one, which saves everything correctly.

👀 To test, make sure the regular problem share links work as expected. The social share buttons are added to the custom share link modal, so those can be tested from there.
#### Relevant tickets

Fixes #6891
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -418,6 +418,7 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
 
   // Pass custom share links to the front end. Unique by UID.
   $base_url = url(base_path(), array('absolute'=> TRUE, 'language' => 'en-global'));
+  $campaign_path = url(current_path(), array('absolute' => TRUE));
   $country_prefix_for_share = (dosomething_global_get_current_prefix()) ? strtoupper(dosomething_global_get_current_prefix()) : 'global';
   $share_image = $base_url . 'sites/default/files/images/problem-' . $campaign->nid . '-' . $country_prefix_for_share . '.png';
 
@@ -439,7 +440,7 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
     ),
   );
 
-  $share_button_markup = dosomething_helpers_share_bar($custom_social_share_link, $social_share_types, $campaign->nid . '_referral', 'social-menu');
+  $share_button_markup = dosomething_helpers_share_bar($campaign_path, $social_share_types, $campaign->nid . '_referral', 'social-menu', $share_path);
 
   $vars['social_share_bar'] = $share_button_markup;
   $vars['custom_social_share_link'] = $custom_social_share_link;
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.share.inc
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.share.inc
@@ -110,7 +110,6 @@ function dosomething_helpers_utm_share_paths($share_types, $path, $utm_campaign,
         'utm_campaign' => $utm_campaign,
       )
     );
-
     $share_paths[$type] = $path . $utm_params;
 
     // Append the custom share source to the URL if we have one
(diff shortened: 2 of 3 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -418,6 +418,7 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
 
   // Pass custom share links to the front end. Unique by UID.
   $base_url = url(base_path(), array('absolute'=> TRUE, 'language' => 'en-global'));
+  $campaign_path = url(current_path(), array('absolute' => TRUE));
   $country_prefix_for_share = (dosomething_global_get_current_prefix()) ? strtoupper(dosomething_global_get_current_prefix()) : 'global';
   $share_image = $base_url . 'sites/default/files/images/problem-' . $campaign->nid . '-' . $country_prefix_for_share . '.png';
```
angaither, line 423: do we need all four of these wild paths?

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.share.inc
```diff
@@ -109,6 +111,11 @@ function dosomething_helpers_utm_share_paths($share_types, $path, $utm_campaign)
       )
     );
     $share_paths[$type] = $path . $utm_params;
+
+    // Append the custom share source to the URL if we have one
+    if ($custom_share_source) {
+      $share_paths[$type] .= '&source=' . $custom_share_source;
```
angaither, line 117: I think the `&` will get encoded strange here, might be best to pass this through `http_build_query` like `dosomething_helpers_utm_parameters` does

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
katiecrane, line 423: `$base_url` and `$country_prefix_for_share` are used in `$share_image`. currently that is just being built the same way it is elsewhere

angaither, line 423: But we don't use that share image in this custom link, right?

katiecrane, line 423: The share image is the image that comes up on facebook or tumblr, so currently for the custom social share buttons that's just the same image as if you used the problem share buttons at the top of the page. Do we want to put something else in there for now?

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.share.inc
katiecrane, line 117: Yeah I feel like it _should_ end up being funky somehow as it is but it seems like it's actually okay? The link that comes up in the tweet is something like `http://dev.dosomething.org:8888/us/campaigns/test-campaign?utm_source=dosomething&utm_medium=twitter&utm_campaign=1485_referral&source=user/5639066ba59dbfe6598b4567` which works great as far as I can tell?

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
angaither, line 423: oh i have noy idea about that

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.share.inc
angaither, line 117: 🎱

katiecrane, line 117: 🔮

angaither:
seems okay-ish, we can leave the share image as is for meow and see if @ngjo wants to change it to other things

PR #6904, opened 2016-08-19
make a new variable for the date when checking age

katiecrane:
#### What's this PR do?

Before, it was trying to set the timestamp as just a blank DateTime object. Needed to make that a fresh variable.

To test, make sure no warnings related to `DateTime::setTimestamp()` come up if voter reg is enabled.
#### Checklist
- [ ] Tested on staging.


angaither:
💯

PR #6920, opened 2016-08-22
Only display the competition modal for US

### LINKED ISSUE #6849: We don't let the world into competitions, but we show them the modal anyways
#### Current Behavior

Currently we display the competition modal across all translations. 
this causes issues like 
1. users see content that isn't in their language
2. users aren't actually ported into gladiator b/c the runs aren't the same

(n.b. normally users who opt into this are global english users who were wrongfully assigned the language) 
#### Desired Behavior

We need to either 
- not offer the modal across all experiences
- always match global english runs with english runs when offering a competition
- fix #6748 
#### Steps to Replicate
- opt in to a compeitions via the modal if you are global english and you won't get a transaction 'welcome to comp' email.  
#### Why This Matters

People want in, and they aren't getting in. 
if you look in the `dosomething_gladiator_failed_task_log` every time there's a `500` it means that user wasn't added to gladiator b/c of a run mis-match

itsjoekent:
#### What's this PR do?

Displays the competition modal for US users only

Test the competition flow for both US + MX, BR, ETC

Heres what it looks like on BR

Heres what it looks like on the same campaign but for US

Nope!
#### Relevant tickets

Fixes #6849 
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -681,7 +681,7 @@ function dosomething_campaign_node_view($node, $view_mode, $langcode) {
     }
 
     // Add signup_data_form variables if needed.
-    dosomething_campaign_add_signup_data_form_vars($node);
+    dosomething_campaign_add_signup_data_form_vars($node, $langcode);
 
     // Flag if we need to register an organ donor on the page.
     if (module_exists('dosomething_organ_donation')) {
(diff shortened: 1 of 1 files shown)


angaither:
ah! I know this is built in a different way but is there anyway to use the new function @katiecrane made? `dosomething_campaign_feature_on`
https://github.com/DoSomething/phoenix/blob/1ab9cac2507c6cf04fe69a8bc4c3b5479bbfcaa6/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module#L1212

itsjoekent:
@angaither from a quick glance it looks like id have to refactor it a bit if we were to use it for this

itsjoekent:
let me quickly check something...

itsjoekent:
yeah this would complicate things a bit unfortunately to be able to make the function call

angaither:
@deadlybutter hrm, bummer. it's annoying to have all these exceptions in the code all over the place
#🌐 

🚢

itsjoekent:
basically

PR #6932, opened 2016-08-23
Changed the classes on this thing to match the styles from steps.

### LINKED ISSUE #6913: Changing the language in the upload photo section
#### User Story

As a member, I'd like to have more context about what the photo uploader is and that someone is reviewing my photo. As an admin, I want the member to know that the photo will be reviewed and that the photo is the scholarship entry.
#### Additional Information

For campaigns that have scholarships on them:
—— scholarship, no RB: “Add your photo here for a scholarship entry!”
—— scholarship, RB: “Add another photo and we’ll check it out!”

For campaigns that don't have a scholarship on them:
— — no scholarship, no RB “Add your photo here and we’ll check it out!”
—— no scholarship, RB “Add another photo and we’ll check it out!"
#### Why This Matters

We want to add clarity around the RB process -- a part of that is clarifying why the photo is important and that someone actually looks at them!
#### Examples

screen shot 2016-08-22 at 10 27 08 am

screen shot 2016-08-22 at 10 28 22 am

### LINKED ISSUE #6930: "Join Others" title styling is incorrect
The styling of the new title in Step 4 is incorrect. It's currently styled as an all caps version of the interior campaign title, but it should have the same weight as a Step X header. It should be slightly larger and bolder.

Current:
screen shot 2016-08-23 at 11 46 19 am

Desired:
example

angaither:
#### What's this PR do?

bow's down to @lkpttn's #fender demands.

🎱 

I can look at a mock up and everything will look exactly the same to me. 
#### Relevant tickets

Fixes #6930

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
@@ -343,7 +343,7 @@
               <p class="copy inline-alt-text-color"><?php print $reportback_copy; ?></p>
             <?php endif; ?>
             <?php //@TODO @fenders look into adding a .-yelling class that all caps things ?>
-            <h3 class="inline-alt-text-color"><?php print t('JOIN OTHERS MAKING A DIFFERENCE'); ?></h3>
+            <h2 class="heading -emphasized"><?php print t('JOIN OTHERS MAKING A DIFFERENCE'); ?></h2>
             <?php if(dosomething_reportback_exists($campaign->nid)): ?>
               <a href="#" data-modal-href="#modal-missing-photos">Is your photo not showing up?</a>
             <?php endif; ?>
--- a/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
@@ -342,8 +342,7 @@
             <?php if (isset($reportback_copy)): ?>
               <p class="copy inline-alt-text-color"><?php print $reportback_copy; ?></p>
             <?php endif; ?>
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -343,7 +343,7 @@
               <p class="copy inline-alt-text-color"><?php print $reportback_copy; ?></p>
             <?php endif; ?>
             <?php //@TODO @fenders look into adding a .-yelling class that all caps things ?>
-            <h3 class="inline-alt-text-color"><?php print t('JOIN OTHERS MAKING A DIFFERENCE'); ?></h3>
+            <h2 class="heading -emphasized"><?php print t('JOIN OTHERS MAKING A DIFFERENCE'); ?></h2>
```
DFurnes: The [`.heading.-emphasized`](http://forge.dosomething.org/#heading) style modifier makes things all-caps as part of its visual appearance, so you can remove the `@TODO` on the line above and make the text sentence case again. ✏️

DFurnes: One Q (perhaps for @lkpttn): do we want the emphasized heading's bottom-border to be visible here on the yellow Step 4 background? If so, we should add the `.-inverse` modifier to make it white.

DFurnes: If not, it feels like this is perhaps a whole new style variant? Nothing we need to necessarily address in this PR, but worth thinking about if we see this all-caps no-visible-underline heading style appear elsewhere.

angaither: I will leave that up to you and sir @lkpttn 🎨

PR #6934, opened 2016-08-23
Don't send "@mobile.import" fields to Northstar either.

DFurnes:
#### What's this PR do?

This PR removes the "email" field from the payload when syncing users with `@mobile.import` email addresses, like we do with `@mobile` emails (to reflect recent change in the "fake" email that Gambit uses when creating Drupal users for MobileCommons reportbacks).

Check out [that regex](https://jex.im/regulex/#!embed=false&flags=&re=%5E%5B0-9%5D%2B%40mobile%28%5C.import%29%3F%24). 

This wasn't breaking anything, but is fake data that we don't want to have in Northstar.
#### Relevant tickets

References DoSomething/gambit#558.
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -240,10 +240,10 @@ function dosomething_northstar_transform_user($user) {
     }
   }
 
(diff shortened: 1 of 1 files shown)


angaither:
dear @mobile.import, take a right!

PR #6940, opened 2016-08-24
Add even more text to the rb section

### LINKED ISSUE #6913: Changing the language in the upload photo section

angaither:
#### What's this PR do?
- Add text to the reportback image holder to say 'and we'll check it out'
- Refactor how that text is added so we know if it's a new reportback or an updated one.

Take a look at the ~new world~ 
I personally think this looks more crowded/worse than what we used to have. 😢 
_new reportback_
image

_updating reportback_
image

The scholarship block was actually quite a bit of refactoring and causing a lot of bugs, I'm wondering if there's a way we can test that in optimizely without writing code to see if it makes a difference? 
#### Relevant tickets

Fixes #6913
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -88,22 +88,23 @@ function dosomething_reportback_form($form, &$form_state, $entity = NULL) {
     ),
     '#prefix' => 'Reportback Image',
     '#suffix' => 'The image must be at least 480 by 480 pixels. The site will crop the image on its own, so if possible submit a square image with the subject in the center of the crop.',
+    '#children' => t("Add your photo here and we’ll check it out!"),
 
   );
+
+  $form['reportback_submissions']['reportback_item']['#attributes']['is_supersized'] = TRUE;
+
   // If rbid doesn't exist, this is a create form.
   if (!isset($entity->rbid)) {
     $entity->rbid = 0;
 
-    $form['reportback_submissions']['reportback_item']['#attributes']['is_supersized'] = TRUE;
-
     $submit_label = t("Submit your pic");
+
+
   }
 
   // Else, it's an update form.
   else {
-    // Output reportback images with date last updated.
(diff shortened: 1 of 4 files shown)


ngjo:
@angaither yes! just checked, we can test on the no RB, scholarship one the text "“Add your photo here for a scholarship entry!”" in optimizely!

I don't like the "and" hanging there like that...maybe we could do

add you photo here!
we'll check 'em out!

add you photo here
and
we'll check 'em out

probably terrible ideas, i'll let @lkpttn speak to the layout more. i'll set up the scholarship block test in optimizely for campaigns w/ scholarships

DFurnes:
Keep in mind the text is just wrapping to fit the container, so it's probably a losing battle to try to have exact control over where things break to a new line.

angaither:
^^^ yeah I tested with removing the `and` and it doesn't split in a sensible way there either.
image

lkpttn:
Is the arrow just an :after element applied to the text box? If the interior text box has a specified width, I think we can narrow it a bit and bring the arrow element closer. That way, even if we're breaking in an odd way, it doesn't look like there's a big separation for the arrow.

angaither:
@lkpttn dis?

image

DFurnes:
🌵⛵

angaither:
going to 🚀  this for now, and @lkpttn can revisit if he feels, when he feels bettah.

PR #6943, opened 2016-08-24
Check installed NPM dependencies for local task.

### LINKED ISSUE #6499: Prevent building assets with out-of-date dependencies.
#### Current Behavior

If someone has out-of-date modules in their `node_modules` directory, they'll get error messages and broken things when they try to run `ds assets`.
#### Desired Behavior

We should see if we can use something like [check-dependencies](https://www.npmjs.com/package/check-dependencies) or [installed-check](https://www.npmjs.com/package/installed-check) to provide a more helpful error message (or automatically re-install modules?) before trying to build with out of date dependencies.
#### Steps to Replicate

Idk, it keeps happening.
#### Why This Matters

You should never get an opaque ugly error message.

DFurnes:
#### What's this PR do?

This PR adds `check-dependencies` to the "start" task used when doing local front-end development. This will check that the locally installed dependencies match what is listed in the project's `package.json`, like so:

screen shot 2016-08-24 at 2 28 28 pm

Check that you can still build. It should holler if you have any out-of-date dependencies.

🔍 
#### Relevant tickets

Fixes #6499.
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/package.json
+++ b/lib/themes/dosomething/paraneue_dosomething/package.json
@@ -1,7 +1,7 @@
 {
   "private": true,
   "scripts": {
-    "start": "npm run clean && npm run modernizr && webpack",
+    "start": "check-dependencies & npm run clean && npm run modernizr && webpack",
     "build": "npm run clean && npm run modernizr && NODE_ENV=production webpack",
     "clean": "rm -rf dist",
     "modernizr": "modernizr -c modernizr.json -d dist/modernizr.js",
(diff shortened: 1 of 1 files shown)


sbsmith86:
nice!

lib/themes/dosomething/paraneue_dosomething/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "private": true,
   "scripts": {
-    "start": "npm run clean && npm run modernizr && webpack",
+    "start": "check-dependencies & npm run clean && npm run modernizr && webpack",
```
sbsmith86, line 4: does this need to be `&&`

DFurnes: D'oh, I was too quick merging! 🐰 

Switching it to `&&` would halt the following steps from running if a dependency was out of date (whereas `&` will let it continue through even if `check-dependencies` returns an error code).

DFurnes: I could see a good argument for switching it to `&&` to be stricter, but it's also nice to be able to just continue along and build if it's just a patch version difference for example...

PR #6947, opened 2016-08-25
Send the user lang prefix in magic link login.

### LINKED ISSUE #6942: Magic link on prod doesn't work because of global redirect
#### Current Behavior
- after you click the magic link (in the app) it redirects to the country specific page
- this means an additional redirect query `redirect=campaigns/the-best-campaign` would be ignored and you are sent to the home page.
#### Desired Behavior
- you're logged in and redirected properly 
#### Steps to Replicate
- blah it's a whole thing
- Request a magic link from an authenticated API request
- Append a redirect query to that link
- Go to that link in prod

Look at [this PR](https://github.com/DoSomething/LetsDoThis-iOS/pull/1049) for more info

angaither:
#### What's this PR do?

Adds global prefix based on the user language to the magic link. 

Make a request to `users/[uid]/magic_login_link` and see the b-e-a-utiful country code pre-appended to it! 

This is needed to avoid redirects happening when you use the link on prod, since a global redirect will override and `redirect=campaigns/whatever` appended onto the query string of the URL. 
#### Relevant tickets

Fixes #6942
#### Checklist
- [ ] Tested on staging.

cc @aaronschachter

### DIFF
--- a/documentation/endpoints/users.md
+++ b/documentation/endpoints/users.md
@@ -228,7 +228,7 @@ POST /api/v1/users/{uid}/magic_login_url
 // 200 Okay
 
 {
-  "url": "http://dev.dosomething.org:8888/user/magic/1700226/1425495228/P-f-5d6kHLrOXl0VrQfXavgmMjiNz042uihpxJW4jBc",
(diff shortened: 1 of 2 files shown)


DFurnes:
⛵ 🌵

Issue #6929, opened 2016-08-23
Theme oboarding as a modal


itsjoekent:
#### User Story

As a user, I should see onboarding as a modal instead of a dropdown.
#### Additional Information (optional)

This should only be used via a specific class flag, which we'll use Optimizely to apply
#### Why This Matters

We want to test if modals work better for the type of interaction we're looking for

angaither:
@deadlybutter can you add some info here about how this works with other post-signup modals?
- signup data form
- voter reg
- competitions

itsjoekent:
@angaither wont be used on any campaign with one of those modals, thats why its triggered by the flag class via optimizely

angaither:
@deadlybutter is this done? can we close this out?

itsjoekent:
yep sorry about that!

PR #6949, opened 2016-08-25
Test modifier for onboarding

itsjoekent:
#### What's this PR do?

This is a variation / modifier for onboarding to make it full screen.
We will apply the modifier via Optimizely for now on campaigns with no competition/voter reg/etc modal flow

Add -onboarding class to the onboarding div

This _was_ 90% mobile ready but we decided to not worry about that for now, we'll only target desktop in optimizely
#### Relevant tickets

Fixes 1 part of #6929 
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/Onboarding.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/Onboarding.js
@@ -31,7 +31,7 @@ class Onboarding extends React.Component {
     }
 
     return (
-      <div id="onboarding" className="onboarding">
+      <div id="onboarding" className="onboarding -fullscreen">
         <div className="wrapper">
           <Slideshow {...this.props} />
 
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/Onboarding.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/Onboarding.js
@@ -31,7 +31,7 @@ class Onboarding extends React.Component {
     }
 
     return (
-      <div id="onboarding" className="onboarding -fullscreen">
+      <div id="onboarding" className="onboarding">
         <div className="wrapper">
           <Slideshow {...this.props} />
 
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
@@ -8,18 +8,32 @@
     position: relative;
   }
 
-  .-fullscreen {
+  &.-fullscreen {
     position: fixed;
     z-index: 1000;
     top: 0;
     left: 0;
     width: 100vw;
-    height: 100vh;
-    overflow-y: scroll;
+    height: 100%;
+
+    &:before {
+      content: "";
+      display: block;
+      background: url('../logo.png');
+      background-size: 72px 60px;
+      width: 72px;
+      height: 60px;
+      position: absolute;
+      top: 0;
+      left: 0;
+      z-index: 1;
+      margin: 12px;
+    }
 
     > .wrapper {
       height: inherit;
       max-width: none;
+      padding-top: 64px;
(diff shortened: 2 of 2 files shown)


weerd:
> Add -onboarding class to the onboarding div

Do you mean `-fullscreen`?

lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
```diff
@@ -8,6 +8,45 @@
     position: relative;
   }
 
+  &.-fullscreen {
+    position: fixed;
+    z-index: 1000;
+    top: 0;
+    left: 0;
+    width: 100vw;
```
weerd: Did `100%` not work here?

weerd:
Add a screenshot of what it looks like in the issue description!

itsjoekent:
> Do you mean -fullscreen?

Uhhhhhhh :shifty: 👀

lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
itsjoekent: I think this is leftover from mobile, but works the same for this purpose. do you prefer just 100%?

itsjoekent:
screen shot 2016-08-25 at 12 50 32 pm

lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
weerd: Mostly curious if there was a specific reason. Did you test in IE and other browsers to make sure it works right? I think support for `vw` units is better nowadays, but best to confirm it works right across multiple browsers!

itsjoekent: ooof good call http://caniuse.com/#search=vw

_can you imagine a world without IE..._

ill flip it to 100%

itsjoekent:
@weerd okie dokes flipped it to percents

lib/themes/dosomething/paraneue_dosomething/js/components/Onboarding.js
```diff
@@ -31,7 +31,7 @@ class Onboarding extends React.Component {
     }
 
     return (
-      <div id="onboarding" className="onboarding">
```
itsjoekent, line 34: dammit i commited the wrong file

lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
weerd: Yep yep. I checked the caniuse.com info, and seems it's only partial support for `vmax` so I think we'd be in the clear with `vw` but if we're still targeting IE8 then defs not...

lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
```diff
@@ -8,6 +8,45 @@
...
+      max-width: none;
+      padding-top: 64px;
+    }
+
+    .slideshow {
+      position: relative;
+      top: 50%;
+      transform: translateY(-50%);
+
+      .slide {
```
weerd: we use `slideshow__slide` for the specific slide and it already should have a `display: block` declaration...

itsjoekent: just checked, yep! ill just get rid of this entire selector then

weerd:
¡OLÉ! :sheep: it!

itsjoekent:
🚂 🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋🚋 🚋

Issue #6910, opened 2016-08-19
Show reactions on Onboarding Slide 3 and allow interaction


weerd:
#### User Story

As a user, when I reach the end of the Onboarding experience and I see Reportback Items, I want to be able to see heart reactions and also add a reaction to the items I see.
#### Why This Matters

More interaction for the user with the community on the specified campaign.

weerd:
Duplicate! see #6879

PR #6951, opened 2016-08-25
Onboarding  enabling reactions and apiclient enhancements

### LINKED ISSUE #6879: Add hearts functionality to reportback photos on onboarding
#### User Story

As a user, I want to interact with photos by giving hearts. 

Pulled this out as a separate issue so we can ship slide 3 without hearts initially. 
#### Why This Matters

We want our members to start interacting with our site as early and often as possible.

weerd:
#### What's this PR do?

This PR updates the `ApiClient` class to fix the `get()` method and add `post()` and `delete()` methods. It also updates how these methods work with `fetch()` and javascript promises, by _correctly_ returning the promise to the location that instantiated the class and called it, so that it may use the `.then()` method to continue the chain when the promise returns a response and allow it to proceed to the next desired functionality.

It also does a bit of code cleanup and adds the `<ReactionsList />` and `<Reaction />` components which work together to output reactions on Reportback Items. The `ReactionList` component uses the updates to the `ApiClient` to both create or delete reactions by the current user on a Reportback Item.

Random, but also trying out leaving in a couple useful `console.log()` calls (and coloring them) with the `ApiClient` so that during local development, we can see what activity is taking place when making API calls. These logs get pulled out during builds on Production!

Pull down code, and checkout out slide 3. Try reacting to a few of the presented RB items. Maybe also doublecheck the database to see if reactions are being created/deleted in the `dosomething_kudos` table as you interact with them.

Slide 3 now with reactions:
image

_NOTE:_ One thing to consider... since we're not using React to create the RB items in the Reportback gallery, but we are to create the RB items on Onboarding Slide 3, I could potentially see an issue where a new user signs up for a campaign, sees the onboarding experience and then "likes" a few photos in Slide 3, **_but_** since we're not using React for the gallery, we can't dynamically update RB items in there, if in both areas, the same RB item is presented. This could happen if there are only a few items for a campaign. I've had this happen while testing and for now I don't see an immediate solution. <-- @jessleenyc the "caveat"
#### Relevant tickets

Fixes #6879
Refs #6910
#### Checklist
- [ ] Tested on staging.

---

@DFurnes @deadlybutter 

cc: @jessleenyc

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/ReactionsList.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/ReactionsList.js
@@ -2,7 +2,7 @@ const React = require('react');
 
 import Reaction from './Reaction';
 import ApiClient from '../utilities/ApiClient';
-import Setting from '../utilities/Setting';
+import setting from '../utilities/Setting';
 
 /**
  * ReactionsList Component
@@ -14,7 +14,7 @@ class ReactionsList extends React.Component {
 
     this.apiClient = new ApiClient('v1');
     this.approvedReactions = this.getApprovedReactions();
-    this.enabled = Setting('dsKudosReactions.enabled', false);
+    this.enabled = setting('dsKudosReactions.enabled', false);
 
     this.state = {
       reactions: this.setInitialReactions()
@@ -32,9 +32,11 @@ class ReactionsList extends React.Component {
    */
   getApprovedReactions() {
     const reactions = {};
-    const approvedReactions = Setting('dsKudosReactions.approved', []);
-    const availableReactions = Setting('dsKudosReactions.terms');
+    const approvedReactions = setting('dsKudosReactions.approved', []);
+    const availableReactions = setting('dsKudosReactions.terms');
 
+    // @TODO: consider possibly using lodash's keyBy() to simplify this:
+    // https://lodash.com/docs#keyBy
     approvedReactions.forEach((value) => {
       if (availableReactions.hasOwnProperty(value)) {
         reactions[availableReactions[value]] = value;
@@ -94,7 +96,7 @@ class ReactionsList extends React.Component {
    */
   toggleReaction(termId) {
     const reaction = this.state.reactions[termId];
-    let reactionId = reaction.id;
+    const reactionId = reaction.id;
 
     // Delete specified reaction record.
     if (reactionId) {
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/ReactionsList.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/ReactionsList.js
@@ -1,0 +1,158 @@
+const React = require('react');
+
+import Reaction from './Reaction';
+import ApiClient from '../utilities/ApiClient';
+import Setting from '../utilities/Setting';
+
+/**
+ * ReactionsList Component
+ * <ReactionsList />
+ */
+class ReactionsList extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.apiClient = new ApiClient('v1');
+    this.approvedReactions = this.getApprovedReactions();
+    this.enabled = Setting('dsKudosReactions.enabled', false);
+
+    this.state = {
+      reactions: this.setInitialReactions()
+    };
+
+    // Bindings
+    this.renderReaction = this.renderReaction.bind(this);
+    this.toggleReaction = this.toggleReaction.bind(this);
+  }
+
+  /**
+   * Get all approved reactions from list of available reactions.
+   *
+   * @return {Object}
+   */
+  getApprovedReactions() {
+    const reactions = {};
+    const approvedReactions = Setting('dsKudosReactions.approved', []);
+    const availableReactions = Setting('dsKudosReactions.terms');
+
+    approvedReactions.forEach((value) => {
+      if (availableReactions.hasOwnProperty(value)) {
+        reactions[availableReactions[value]] = value;
+      }
+    });
+
+    return reactions;
+  }
+
+  /**
+   * Render a Reaction component.
+   *
+   * @param  {String} key
+   * @param  {Number} index
+   * @return {Object}
+   */
+  renderReaction(key, index) {
+    return <li key={index}><Reaction termId={key} total={this.state.reactions[key].total} reacted={this.state.reactions[key].reacted} onClick={this.toggleReaction} /></li>
+  }
+
+  /**
+   * Collect and setup the initial reactions for the
+   * ReactionList component state.
+   *
+   * @return {Object}
+   */
+  setInitialReactions() {
+    const reactions = {};
+
+    Object.keys(this.approvedReactions).forEach((key) => {
+      reactions[key] = {
+        id: null,
+        reacted: false,
+        termId: key,
+        termName: this.approvedReactions[key],
+        total: 0
+      };
+
+      this.props.reactions.forEach((data) => {
+        if (data.term.id === key) {
+          reactions[key].id = data.current_user.kudos_id;
+          reactions[key].reacted = data.current_user.reacted;
+          reactions[key].total = data.term.total;
+        }
+      });
+    });
+
+    return reactions;
+  }
+
+  /**
+   * Toggle the specified reaction state by either creating
+   * a new reaction or deleting the existing one.
+   *
+   * @param  {String} termId
+   * @return {void}
+   */
+  toggleReaction(termId) {
+    const reaction = this.state.reactions[termId];
+    let reactionId = reaction.id;
+
+    // Delete specified reaction record.
+    if (reactionId) {
+      this.apiClient.delete(`kudos/${reactionId}`)
+        .then((response) => {
+          if (response.hasOwnProperty('success')) {
+            this.updateReaction(termId);
+          }
+        });
+    }
+
+    // Create new record for reaction.
+    if (!reactionId) {
+      this.apiClient.post('kudos', {
+        'reportback_item_id': this.props.reportbackItemId,
+        'user_id': this.props.user.info.drupal_id,
+        'term_ids':[termId]
+      }).then((response) => {
+        response.forEach((data) => {
+          if (data) {
+            this.updateReaction(termId, data);
+          }
+        });
+
+      });
+    }
+  }
+
+  /**
+   * Update the specified reaction in state.
+   *
+   * @param  {String} termId
+   * @param  {Object|null} data
+   * @return {void}
+   */
+  updateReaction(termId, data = null) {
+    const reaction = this.state.reactions[termId];
+
+    reaction.reacted = data ? true : false;
+    reaction.id = data ? data.kid : null;
+    reaction.total = data ? reaction.total + 1 : reaction.total - 1;
+
+    this.setState({
+      reactions: this.state.reactions
+    });
+  }
+
+  render () {
+    if (!this.enabled) {
+      return null;
+    }
+
+    return (
+      <ul className="form-actions -inline kudos">
+        {Object.keys(this.state.reactions).map(this.renderReaction)}
+      </ul>
+    );
+  }
+}
+
+export default ReactionsList;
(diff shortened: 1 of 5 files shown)


weerd:
@jessleenyc to resolve the "caveat", we could maybe have a function that gets called when reacting on an RB item in the React onboarding version, that jumps out to find a corresponding item in the gallery, and if so, then change the attributes on it?

lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
```diff
@@ -36,6 +43,37 @@ class ApiClient {
...
+   * Send a DELETE request to the given path URI.
+   *
+   * @param  {String} path
+   * @return {Object}
+   */
+  delete(path) {
+    const url = `${this.url}${path}`;
+
+    // @TODO: see todo in post() method!
+    console.log('%c ApiClient DELETE Request: \n' + url,
```
itsjoekent, line 55: whoops?

lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
```diff
@@ -36,6 +43,37 @@ class ApiClient {
...
+    // provide options & configurations, etc.
+    return window.fetch(url, {
+      method: 'DELETE',
+      headers: this.headers,
+      credentials: 'same-origin'
+    })
+      .then(this.checkStatus)
+      .then(this.parseJson)
+      .catch(function(data) {
+        console.log(data);
```
itsjoekent, line 71: Also should probably remove the .log(data) here & in the one below

itsjoekent, line 71: Oh I see in your TODO below

lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
weerd: Nope. On purpose!

> Random, but also trying out leaving in a couple useful console.log() calls (and coloring them) with the ApiClient so that during local development, we can see what activity is taking place when making API calls. These logs get pulled out during builds on Production!

weerd: This is what it looks like at the moment on local development:

image

Will help us while developing to know what is getting passed to API calls via the API client 😄

itsjoekent:
So im getting a bit lost on the ReactionList component.

We start using the dsKudos data in the settings object,

```
{
 "approved": ["heart"],
 "enabled": true,
 "terms": {"heart":"641"},
 "user":"1700226"}
```

We then go through each approved reaction (approved array) and check if its an available term (terms array) https://github.com/DoSomething/phoenix/pull/6951/files#diff-3d725c871d1c6f13bde530ede939aefeR38

and then we save in the new approved reactions array the value of the term (641)

after we gather this data we set the state with `setInitialReactions`. This goes through every approved reaction we made above & sets the reaction in the state, with the key being the term id

```
    Object.keys(this.approvedReactions).forEach((key) => {
       reactions[key] = {
```

But I get lost here, because right now `key` is always going to be the value of heart, `641`? How do you identify different users reacting...

apparently reaction list is used per reportback item. WHOOPS. 

this really threw me off because (AFAIK?) we're not doing multiple kudo terms, which led me to assume ReactionList was a container of all reactions like the name implies.

lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
DFurnes: Yep, I think this is a good idea to get in the habit of since there's no downside and can be massively useful for debugging (with useful messages, obvs, not like `console.log('💩')`).

DFurnes:
> @jessleenyc to resolve the "caveat", we could maybe have a function that gets called when reacting on an RB item in the React onboarding version, that jumps out to find a corresponding item in the gallery, and if so, then change the attributes on it?

I think this is going to be an edge case (especially looking at current kudos engagement). It'd be nice to consider a React refactoring of the gallery later on to tackle some of the technical debt there, rather than layering on a hacky fix for something that shouldn't affect most users.

lib/themes/dosomething/paraneue_dosomething/js/components/ReactionsList.js
```diff
@@ -0,0 +1,158 @@
+const React = require('react');
+
+import Reaction from './Reaction';
+import ApiClient from '../utilities/ApiClient';
+import Setting from '../utilities/Setting';
```
DFurnes: Since this is a function, let's keep it lowercase: `import setting from '../utilities/Setting'`. And maybe we should move it into that `helpers.js` file we made to reduce confusion with the capitalized filename.

lib/themes/dosomething/paraneue_dosomething/js/components/ReactionsList.js
```diff
@@ -0,0 +1,158 @@
...
+   * Get all approved reactions from list of available reactions.
+   *
+   * @return {Object}
+   */
+  getApprovedReactions() {
+    const reactions = {};
+    const approvedReactions = Setting('dsKudosReactions.approved', []);
+    const availableReactions = Setting('dsKudosReactions.terms');
+
+    approvedReactions.forEach((value) => {
```
DFurnes, line 40: Could this be simplified down like this? Although it would lose the custom keying, not sure if that's necessary.

``` js
return approvedReactions.filter(value => availableReactions.hasOwnProperty(value));
```

weerd:
@deadlybutter yep yep. It's a little confusing but something we were doing in the `Reportback.js` for the view more gallery too. We need to reconcile between the approved reactions and the available reactions, because there may be like more available but we only want to show one for now.

As for the `key`, yeah, it won't conflict. Say there's a reaction with key `641`, in the ReactionList there can only ever be 1 item with property name `641` that contains the totals and whether current user reacted, _per_ reportback item.

Here's a visual that helps explain. I find pictures do a better job than words:

reportbackitem reactionlist reactions

lib/themes/dosomething/paraneue_dosomething/js/components/ReactionsList.js
```diff
@@ -0,0 +1,158 @@
...
+  /**
+   * Toggle the specified reaction state by either creating
+   * a new reaction or deleting the existing one.
+   *
+   * @param  {String} termId
+   * @return {void}
+   */
+  toggleReaction(termId) {
+    const reaction = this.state.reactions[termId];
+    let reactionId = reaction.id;
```
DFurnes: This can be `const` right?

lib/themes/dosomething/paraneue_dosomething/js/components/ReactionsList.js
weerd: hmmm, would be cleaner but I'm relying on the custom `key`ing cause I'm relying on it later in `setInitialReactions()`.

itsjoekent:
@weerd yeah, originally i was thinking the ReactionList component applied to all the reportbacks, in which case it would have conflicted

lib/themes/dosomething/paraneue_dosomething/js/components/ReactionsList.js
weerd: ah, I was wondering why we lowercased it elsewhere. Good call!

lib/themes/dosomething/paraneue_dosomething/js/components/ReactionsList.js
```diff
@@ -0,0 +1,158 @@
...
+    // Create new record for reaction.
+    if (!reactionId) {
+      this.apiClient.post('kudos', {
+        'reportback_item_id': this.props.reportbackItemId,
+        'user_id': this.props.user.info.drupal_id,
+        'term_ids':[termId]
+      }).then((response) => {
+        response.forEach((data) => {
+          if (data) {
+            this.updateReaction(termId, data);
```
DFurnes, line 120: I'm not sure, but does this cause "n" re-renders because of the `.forEach` and the `this.setState()` in the update function? I forget whether React is smart about batching those up.

weerd:
@deadlybutter indeed! But good to call it out so we could clarify here! 💃

lib/themes/dosomething/paraneue_dosomething/js/components/ReactionsList.js
weerd: doh! yep; indeed it can.

lib/themes/dosomething/paraneue_dosomething/js/components/ReactionsList.js
DFurnes: Ah, yeah I see. Could maybe do something with Lodash's [keyBy](https://lodash.com/docs#keyBy), but defs not blocking!

lib/themes/dosomething/paraneue_dosomething/js/components/ReactionsList.js
itsjoekent: looking at the update function we'd only have 2 re-renders assuming react isnt being smart about it, not a _yuge_ deal

DFurnes:
That drawing is my favorite thing ever. 🎨

weerd:
@DFurnes iPad Pro + Apple Pencil + ProCreate. I really want all my PR descriptions to be drawings from now on, haha!

lib/themes/dosomething/paraneue_dosomething/js/components/ReactionsList.js
weerd: Hmmm, not sure. Good point tho... is there a good way to check? I didn't see anything crazy going on but also wasn't aware to look for it...

itsjoekent: @weerd I have the React developer tools, adds another tab in the dev console thing

https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi?hl=en

I think we can track React updates with this, let me check...

itsjoekent: hm, it has a trace updates feature but it just highlights the area which updated. 

I guess a console.log (or console.trace if youre in chrome) might work, who needs fancy shit!

weerd: I've been using the React dev tools too! It's kinda crucial, haha Wasn't sure if there's a better way to check in there is all.

DFurnes: Might be really helpful to include [something like this](https://github.com/ryanflorence/react-lumberjack) in dev builds. (Or as we build more complicated things, it's probably a good argument for using something like Redux)

itsjoekent: that looks neat

DFurnes: For conditionally including in dev build:

``` js
if (process.env.NODE_ENV !== 'production') {
  require('react-lumberjack')
}

```

lib/themes/dosomething/paraneue_dosomething/js/components/ReactionsList.js
weerd: Added it as a `@TODO` to look into using it :)

DFurnes:
🚢

weerd:
@DFurnes @deadlybutter new commit addresses avoiding the potential for "n" re-renders. Since it's a Promise I can just chain another [`.then()` method](https://github.com/DoSomething/phoenix/pull/6951/files#diff-3d725c871d1c6f13bde530ede939aefeR135) to update the state only _after_ processing through the `forEach()` loop 💃 Seems to work fine!

weerd:
Nevermind. Reverted back after IRL convo w/ @DFurnes!

itsjoekent:
@weerd why did you revert?

weerd:
Talked it over with @DFurnes and it's really just not a good idea to have the state mutations happening separate from the actual setting of the state via React API; it could lead to issues and bugs. We talked it over and how I had it before should be fine, but eventually we should consider making changes in a cloned object and then passing that into the state via React API. Ideally, we'd also move to using something like Redux...

PR #6953, opened 2016-08-26
add utms and fallback if no northstar id

### LINKED ISSUE #6807: Add UTM parameter to the custom social link
#### User Story

As an admin, I would like to easily see in GA all traffic coming in from custom social links. 

If this is doable, will get the UTM parameter we want to use
#### Why This Matters

So we can see who is coming in from the custom social link

There's a way to manually group all of the traffic from custom social links in GA, but it's not ideal.

katiecrane:
#### What's this PR do?

Added the UTM parameters that we want to include in the custom share link. Added a fallback of using the phoenix ID in case the Northstar ID is NONE.

Check for typos. 
#### Relevant tickets

Fixes #6807 
#### Checklist
- [ ] Tested on staging.

cc: @ngjo

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -213,9 +213,12 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
   // 1) Materials
 
   // Build the unique share link.
-  $utm_source = 'influencer';
-  $utm_medium = 'referral_web';
-  $utm_campaign = 'members';
+  $utm_params = [
+    'utm_source' => 'influencer',
+    'utm_medium' => 'referral_web',
+    'utm_campaign' => 'members',
+  ];
+  $utm_params = dosomething_helpers_utm_parameters(null, $utm_params);
 
   if (dosomething_user_get_field('field_northstar_id') !== 'NONE') {
     $share_path = 'user/' . dosomething_user_get_field('field_northstar_id');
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -213,8 +213,17 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
   // 1) Materials
 
   // Build the unique share link.
-  $share_path = 'user/' . dosomething_user_get_field('field_northstar_id');
-  $custom_social_share_link = dosomething_global_url(current_path(), array('absolute' => TRUE,'query' => array('source' => $share_path)));
+  $utm_source = 'influencer';
```
angaither: all this stuffs already can already be set with a [function](https://github.com/DoSomething/phoenix/blob/a810fa26d6fc9051d1b688e0f7c6bd448087abe9/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module#L577), ya?

katiecrane: @angaither I'm not sure if we can use that if we want to build the query all together [below](https://github.com/DoSomething/phoenix/pull/6953/files#diff-f3920561a5ca7c40913d6f8452fb9c0fR226), since there is more to the query than just the utms?

angaither: you should be able to append more onto the link after utms are added

katiecrane: on it!

katiecrane: @angaither I think it looks way cleaner now, good call!

angaither:
🚢

PR #6956, opened 2016-08-26
Make sure that "fallback" register call always uses full user object.

DFurnes:
#### What's this PR do?

When our data cleanup scripts get a 404 on updating a user profile by Drupal ID, they'll try to register that user via the `POST users/` endpoint. Unfortunately, sending the same payload along would attempt to make a "partial" user (just the changed fields)...

This PR refactors those Northstar sync methods to always operate on a full `$user` object rather than being passed a custom payload. This way, if the `dosomething_northstar_update_user` fails, it'll try to run `dosomething_northstar_register_user` based on the full user object.

I tried to do a little tidying up while I was working on this PR, so I'd recommend taking a look commit-by-commit so it doesn't just look like a bunch of noise:

👾 Just whitespace: 5c02291

🏭 Use the `$user` object in the register method rather than a `$payload` array: b47c4b1

🔒 Handle the `password` vs `drupal_password` hash business in the transform too: e4740cd

🍃 Shuffle things around (short array syntax, move private methods to bottom of file, rename "register" method to "create" so it's same terminology as we use in Northstar): 17ac4e1

Hopefully this is the last PR for this? 😁 
#### Relevant tickets

References #6409.
#### Checklist
- [ ] Tested on staging.


angaither:
let's goooo

PR #6962, opened 2016-08-29
Vcard social image

### LINKED ISSUE #6963: Adding image to the social share buttons
#### User Story

As we've been talking about at length, we wanna be able to use a vcard in the social share button.
#### Why This Matters

So when they click on the custom social share button (not problem share), it pulls in a vcard and not the problem share photo.

katiecrane:
#### What's this PR do?

We're trying to put a custom image in the social share for voter reg. If voter reg is enabled, use this image instead. The nid of the image is part of the custom share custom settings. We use that to pull in the image.

Try the 3 shares and make sure the image comes up.

v-cardz
#### Relevant tickets

Fixes #6963
#### Checklist
- [ ] Tested on staging.

cc: @ngjo

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -420,9 +420,11 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
   $base_url = url(base_path(), array('absolute'=> TRUE, 'language' => 'en-global'));
   $campaign_path = url(current_path(), array('absolute' => TRUE));
   $country_prefix_for_share = (dosomething_global_get_current_prefix()) ? strtoupper(dosomething_global_get_current_prefix()) : 'global';
-  $share_image = $base_url . 'sites/default/files/images/problem-' . $campaign->nid . '-' . $country_prefix_for_share . '.png';
   if (dosomething_campaign_feature_on($campaign, 'enable_voter_registration', TRUE)) {
-    $share_image = $base_url . 'sites/default/files/' . 'social-v-card' . '.png';
+      $vars['share_image'] = dosomething_image_get_themed_image_url($campaign->variables['share_image_nid'], 'landscape');
+  }
+  else {
+    $vars['share_image'] = $base_url . 'sites/default/files/images/problem-' . $campaign->nid . '-' . $country_prefix_for_share . '.png';
   }
 
   // Add social share bar.
@@ -430,7 +432,7 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
     'facebook' => array(
       'type' => 'feed_dialog',
       'parameters' => array(
-      'picture' => $share_image,
+      'picture' => $vars['share_image'],
     ),
   ),
     'twitter' => array(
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -421,6 +421,9 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
   $campaign_path = url(current_path(), array('absolute' => TRUE));
   $country_prefix_for_share = (dosomething_global_get_current_prefix()) ? strtoupper(dosomething_global_get_current_prefix()) : 'global';
   $share_image = $base_url . 'sites/default/files/images/problem-' . $campaign->nid . '-' . $country_prefix_for_share . '.png';
+  if (dosomething_campaign_feature_on($campaign, 'enable_voter_registration', TRUE)) {
+    $share_image = $base_url . 'sites/default/files/' . 'social-v-card' . '.png';
```
angaither: do you need to upload the actual file here?

katiecrane: it would need to live there yeah. I don't think there's a way to do it dynamically but I don't like this either

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -420,22 +420,27 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
   $base_url = url(base_path(), array('absolute'=> TRUE, 'language' => 'en-global'));
   $campaign_path = url(current_path(), array('absolute' => TRUE));
   $country_prefix_for_share = (dosomething_global_get_current_prefix()) ? strtoupper(dosomething_global_get_current_prefix()) : 'global';
-  $share_image = $base_url . 'sites/default/files/images/problem-' . $campaign->nid . '-' . $country_prefix_for_share . '.png';
+  if (dosomething_campaign_feature_on($campaign, 'enable_voter_registration', TRUE)) {
+      $vars['share_image'] = dosomething_image_get_themed_image_url($campaign->variables['share_image_nid'], 'landscape');
```
angaither, line 424: I think you also want to specifcy which image crop you need here

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.variable.inc
```diff
@@ -32,6 +32,15 @@ function dosomething_helpers_variable_form($form, &$form_state, $node) {
       'maxlength' => '140',
     ),
   ];
+  $form['custom_social_sharing']['share_image_nid'] = [
+    '#type' => 'textfield',
+    '#title' => t('Photo nid'),
```
angaither, line 37: put some help text in the field so people know to just put in the id

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
katiecrane: @angaither you can, but it's optional, so I thought we'd want to just stick with the original?

angaither: 💭  typically the images are pretty large, but we can give it a shot

angaither:
🚀

PR #6966, opened 2016-08-29
Clean up NPM dependencies.

### LINKED ISSUE #6498: Clean up NPM git dependencies.
#### Current Behavior

We had previously configured a couple of our NPM dependencies (`fixed-sticky`, `html5shiv`, `unveil`, and `jquery.iecors`) to install from my GitHub forks so we could apply patches.
#### Desired Behavior

We should revisit whether we still need these custom forks, and if not, use the versions in the registry.
#### Steps to Replicate

Look at the [`package.json`](https://github.com/DoSomething/phoenix/blob/b16a916bbe8b821890b8b3c96fe01111cc4e1230/lib/themes/dosomething/paraneue_dosomething/package.json) in the theme directory. You will see things.
#### Why This Matters

We are currently using an old "snapshot" I made of each of those dependencies. It'd be nice to resume using the versions in the registry so we can get bug fixes or improvements if any are made. Also, NPM is not particularly good at handling Git dependencies so it'd be nice to not have to make it do that.

DFurnes:
#### What's this PR do?

This pull request updates our NPM dependencies so they don't rely on my personal forks anymore:

🔪 For `fixed-sticky` and `html5shiv`, this just means updating to the latest published release.

🍸 I removed `jquery.iecors` because I don't believe we make cross-domain AJAX requests anymore (since switching to Fastly), and it's only included for IE8 support.

💤 Switches out Unveil for [Layzr](http://callmecavs.com/layzr.js/) for lazy loading because it's available on NPM, doesn't have a jQuery dependency, and has a nicer API (so we can remove some hacks and todos!)

👀

This shouldn't have any negative side effects, but it's probably worth doing a little once-over anywhere where we do lazy loading to make sure it all still works well. From testing on my local, all is a-okay.
#### Relevant tickets

Fixes #6498.
#### Checklist
- [ ] Tested on staging.


weerd:
> my personal forks

🍴 🍴 🍴 🍴

sbsmith86:
👍

lib/themes/dosomething/paraneue_dosomething/js/utilities/LazyLoader.js
```diff
@@ -0,0 +1,50 @@
...
+
+  loader = new Layzr({
+    normal: 'data-src',
+    retina: 'data-src-retina',
+  });
+
+  loader.on('src:before', image => {
+    // On load, set the images to be transparent so we can fade them in.
+    image.style.transition = 'opacity 0.5s ease-in';
+    image.style.opacity = 0;
```
weerd, line 36: Any chance we would want to use/remove classes here instead of adding properties to `style` object? Would it make it easier to modify the effect from CSS rather than JS? Just a thought.

weerd:
1 comment! Otherwise 
:sheep: it!

lib/themes/dosomething/paraneue_dosomething/js/utilities/LazyLoader.js
DFurnes: I originally did that with `.lazy-loaded` and `.is-loaded` classes, but then it seemed ridiculous to have to hunt down everywhere this applies to add the class. I don't have strong feelings, but it does feel kinda nice and tidy to have everything in this one little file.

weerd: That's the ReactJS in you talking :trollface:

PR #6967, opened 2016-08-29
Add a script to fix users with empty email.

DFurnes:
#### What's this PR do?

Next of the gotchas: there were 413 users with an empty-string `mail` field. All those users had a `field_mobile` so perhaps there was just a brief moment where we didn't require emails for API users? Anyways, this fixes that so we don't accidentally deactivate them all.

Is the script okay?

👊
#### Relevant tickets

References #6409.
#### Checklist
- [ ] Tested on staging.


angaither:
📱 ➡️  ✉️

PR #6968, opened 2016-08-29
DELETE all the code!! And replace with a bit of less, yet better code!

### LINKED ISSUE #6954: Add whatwg-fetch library to lib.js instead of webpack config.
The `whatwg-fetch` library is a polyfill we add to help browsers out that don't support the native `window.fetch` api. We currently include the library via the webpack config file, but typically we add our polyfills in the `lib.js` file. Let's move it in there and test to make sure the `ApiClient.js` is still working as it should!

weerd:
#### What's this PR do?

This PR DRYs up the request methods in the `ApiClient` class in JS, so that they all send their requests through the new `send()` method. This will help keeps things cleaner and easier to manage in the class. It also update the console output for development purposes when a request is made via the class.

Example in the console on development only:
screen shot 2016-08-29 at 4 18 59 pm

This PR also moves the `whatwg-fetch` polyfill library out of the webpack config and into **libs.js**.

👀 do API calls happening in the Onboarding experience still work?!?
#### Relevant tickets

Fixes #6954
Fixes: 💅 
#### Checklist
- [ ] Tested on staging.

---

@DFurnes @deadlybutter 

cc: @jessleenyc

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
@@ -130,9 +130,13 @@ class ApiClient {
       Object.assign(defaults, options);
     }
 
-    console.log('%c ApiClient Request! \n',
-      'background-color: rgba(105,157,215,0.5); color: rgba(33,70,112,1); line-height: 1.5;'
+    // Developer Output:
+    console.log('%c ApiClient %s Request:',
+      'background-color: rgba(105,157,215,0.5); color: rgba(33,70,112,1); display: block; font-weight: bold; line-height: 1.5;',
+      method
     );
+    console.log('URL: \n%s', url);
+    console.log('Options: \n%o', defaults);
 
     return window.fetch(url, defaults)
       .then(this.checkStatus)
--- a/lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
@@ -1,3 +1,5 @@
+const isEmpty = require('lodash/isEmpty');
+
 /**
  * ApiClient class to make calls to different services.
  * This could eventually be a class that other classes more
@@ -5,10 +7,6 @@
  * request structure on their own so it can accommodate
  * various endpoint URL structures... a boy can dream.
  * For example: class PhoenixApiClient extends ApiClient { .. }
- *
- * @todo when extending this ApiClient to create more specific
- * service clients, could use Object.assign or Lodash extend()
- * to merge header objects with default headers.
  */
 class ApiClient {
   constructor(version, host = window.location.hostname, port = window.location.port) {
@@ -51,26 +49,7 @@ class ApiClient {
   delete(path) {
     const url = `${this.url}${path}`;
 
-    // @TODO: see todo in post() method!
-    console.log('%c ApiClient DELETE Request: \n' + url,
-      'background-color: rgba(105,157,215,0.5); color: rgba(33,70,112,1); line-height: 1.5;'
-    );
-
-    // @TODO: lots of repeated code with get() and post().
-    // Update to use a send() method (created below) to handle
-    // all the fetch stuff, and these other methods simple
-    // provide options & configurations, etc.
-    return window.fetch(url, {
-      method: 'DELETE',
-      headers: this.headers,
-      credentials: 'same-origin'
-    })
-      .then(this.checkStatus)
-      .then(this.parseJson)
-      .catch(function(data) {
-        console.log(data);
-        console.log('um, there was an error! handle that shizzzzz!');
-      });
+    return this.send('DELETE', url);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
```diff
@@ -137,29 +106,45 @@ class ApiClient {
...
+   * @return {Promise}
+   */
+  send(method, url, options = {}) {
+    const defaults = {
+      method: method,
+      headers: this.headers,
+      credentials: 'same-origin'
+    };
+
+    if (! isEmpty(options)) {
```
DFurnes: Would Lodash's [`defaults`](https://lodash.com/docs#defaults) function work here?

lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
```diff
@@ -137,29 +106,45 @@ class ApiClient {
...
+    );
+    console.log('URL: \n%s', url);
+    console.log('Options: \n%o', defaults);
+
+    return window.fetch(url, defaults)
+      .then(this.checkStatus)
+      .then(this.parseJson)
+      .catch((data) => {
+        console.log(data);
+        console.log('um, there was an error! handle that shizzzzz!');
```
DFurnes: What do you think of making this a `console.error` so it stands out?

lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
weerd: hmm, looking into it. Didn't know it was a thing!

DFurnes:
Two little thoughts – otherwise looks great! 🚢

lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
weerd: ah, jes. I need to still flesh out the error handling stuff, but totally makes sense to use that instead!

lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
weerd: Nice. Yeah, that would be better so we could override the default ones if needed! Good call @DFurnes

lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
```diff
@@ -137,29 +106,45 @@ class ApiClient {
...
+   * @param  {String} method
+   * @param  {String} url
+   * @param  {Object} options
+   * @return {Promise}
+   */
+  send(method, url, options = {}) {
+    const defaults = {
+      method: method,
+      headers: this.headers,
+      credentials: 'same-origin'
```
itsjoekent, line 126: Since my mind is now thinking about Social Auth, at some point this should probably become one of the configurable options so we can use this for various services / api's beyond phoeix

weerd:
I was considering using: https://github.com/marak/colors.js/ for coloring the output in console, but also wasn't sure how best to add a package that doesn't get bundled for prod. I wasn't even sure I'd end up using it or not, but curious if you had suggestions on the non-prod bundling issue @DFurnes

lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
weerd: Yeah, with switching to Lodash's `defaults()`, it should allow overriding items you pass in. Working on it now as per [@DFurnes suggestion](https://github.com/DoSomething/phoenix/pull/6968#discussion_r76680252)

itsjoekent:
@weerd dev dependency? `npm install <thing> --save-dev` i think

lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
itsjoekent: OH thats neat

DFurnes:
You can conditionally include things with the `NODE_ENV` environment variable.

``` js
if (process.env.NODE_ENV !== 'production') {
  require('colors.js')
}
```

weerd:
@deadlybutter yeah, the problem is I'd be `require`-ing in the same `ApiClient` class, and thus would get bundled in. @DFurnes yeah, more something like that, cool. I'll play around with it at some point and see where it gets me and if it's worth it.

DFurnes:
@deadlybutter NPM's `dependencies` and `devDependencies` will both always get installed, since this isn't a library (where NPM will skip the `devDependencies` when a package is being installed as another project's dependency).

We've been adding "local dev" (build, linting, etc) dependencies in that group by convention, but it just so happens that they also aren't things that are included in the final production build. Nothing's really stopping us from doing that if we wanted to be _wild_.

itsjoekent:
@DFurnes Oh interesting, I always thought production builds excluded those

weerd:
@deadlybutter new lodash `merge()` will override any props on `default` object that you pass into via the `options` object. 😉

DFurnes:
🚢

PR #6971, opened 2016-08-30
Fixes security issue with Kudos creation

### LINKED ISSUE #6933: Any User Can Set Kudos Anonymously
#### Current Behavior

Full report here:

https://cobalt.io/dosomething/dosomething/reports/97
#### Desired Behavior

Provide authentication mechanism for Kudos to be restricted to individual user and not over-rideable.
#### Steps to Replicate

Check Cobalt report, or with @deadlybutter who's also re-created it.
#### Why This Matters

We want kudos to accurately reflect use trends and likes, and not be open to spammable/botnet behavior.

itsjoekent:
#### What's this PR do?

Prevents you from creating a Kudo for someone else 
1. Can you perform kudos?
2. Edit the `<meta name="drupal-user-id"` with a different user id. Can you still perform kudos? (Look at the JS console, errors blocking auth should show up)

Never trust the client!
#### Relevant tickets

Fixes #6933 
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
@@ -95,6 +95,7 @@ function _kudos_resource_definition() {
             ],
           ],
         ],
+        'access arguments append' => TRUE,
         'access callback' => '_kudos_resource_access',
         'access arguments' => ['create'],
       ],
@@ -135,7 +136,9 @@ function _kudos_resource_definition() {
  *
  * @return bool
  */
-function _kudos_resource_access($op) {
+function _kudos_resource_access($op, $args = []) {
+  global $user;
+
   if ($op === 'view') {
     return TRUE;
   }
@@ -145,9 +148,17 @@ function _kudos_resource_access($op) {
   }
 
   if ($op === 'create' || $op === 'delete') {
-    if (!user_is_anonymous()) {
-      return TRUE;
+    // Only auth'd users should be giving Kudos
+    if (user_is_anonymous()) {
+      return FALSE;
     }
+
+    // We need to make sure people are only creating Kudos for themselves
+    if ($op === 'create' && $user->uid !== $args[1]) {
+      return FALSE;
+    }
+
+    return TRUE;
   }
 
   return FALSE;
--- a/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
@@ -77,15 +77,6 @@ function _kudos_resource_definition() {
             ],
           ],
           [
-            'name' => 'user_id',
-            'description' => 'The ID of the user executing a kudos.',
-            'optional' => FALSE,
-            'type' => 'string',
-            'source' => [
-              'data' => 'user_id',
-            ],
-          ],
-          [
             'name' => 'term_ids',
             'description' => 'An array of kudos term IDs.',
             'optional' => FALSE,
@@ -95,7 +86,6 @@ function _kudos_resource_definition() {
             ],
           ],
         ],
-        'access arguments append' => TRUE,
         'access callback' => '_kudos_resource_access',
         'access arguments' => ['create'],
       ],
@@ -149,16 +139,9 @@ function _kudos_resource_access($op, $args = []) {
 
   if ($op === 'create' || $op === 'delete') {
     // Only auth'd users should be giving Kudos
-    if (user_is_anonymous()) {
-      return FALSE;
+    if (!user_is_anonymous()) {
+      return TRUE;
     }
-
-    // We need to make sure people are only creating Kudos for themselves
-    if ($op === 'create' && $user->uid !== $args[1]) {
-      return FALSE;
-    }
-
-    return TRUE;
   }
 
   return FALSE;
@@ -200,10 +183,9 @@ function _kudos_resource_index($reportbackitems, $count) {
  * @param $user_id
  * @param $term_ids
  */
-function _kudos_resource_create($reportback_item_id, $user_id, $term_ids) {
+function _kudos_resource_create($reportback_item_id, $term_ids) {
   $parameters = [
     'fid' => $reportback_item_id,
-    'uid' => $user_id,
     'tids' => $term_ids,
   ];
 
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
```diff
@@ -145,9 +148,17 @@ function _kudos_resource_access($op) {
...
   if ($op === 'create' || $op === 'delete') {
-    if (!user_is_anonymous()) {
-      return TRUE;
+    // Only auth'd users should be giving Kudos
+    if (user_is_anonymous()) {
+      return FALSE;
+    }
+
+    // We need to make sure people are only creating Kudos for themselves
+    if ($op === 'create' && $user->uid !== $args[1]) {
```
weerd: Is `$args[1]` always the user id? Seems like it could be a bit brittle if things change a bit? Not sure... been a while since I touched this stuff @_@

itsjoekent: yeah its always [1]. if someone adds more arguments out of order it would break this but there unfortunately isn't another way that i know of

DFurnes: Yeah, I couldn't find any better solutions either when doing this for the magic link endpoints. 😞 Would it make sense to remove the `uid` value from the URL altogether since it always has to be the same as the logged-in user?

itsjoekent: :thinking_face:

itsjoekent: yeah thats probably the optimal way of doing it

weerd: @deadlybutter maybe add a note up above in the arrays to then see the stuff in this function for the time being?

itsjoekent: @weerd looking at how hard it is to rip user id out. so far its going well... if i hit a major roadblock ill just add the note & merge

weerd: @DFurnes I think we need to keep it... I feel there's stuff that would break if removed. @_@

itsjoekent: @weerd what would break...?

weerd: Are there scenarios like the mobile app, where the "logged-in" user is a generic auth'd user for the sake of allowing access to certain things, but it's not _actually_ the current user... that data is just getting passed around? I feel that's the scenario it will break on.

itsjoekent: We should probably never have generic users like that, sounds like it would have a lot unintended consequences beyond Kudos....

weerd: @deadlybutter given we have no tests; I'm sure there's something... _Murphy's Law_

weerd: @deadlybutter mobile app does that, I believe...

DFurnes: I don't believe we're doing that anymore (neither mobile app nor Northstar has kudos). They'd also break regardless since the "API user" wouldn't have the same UID as the target kudos owner.

DFurnes: Definitely not necessary for this PR, but it'd be a nice clean-up eventually.

itsjoekent: Well, this patch would break there experience anyway

itsjoekent: whoops dave already wrote that, dammit github

weerd: Yeah, I was gonna mention that then it'd break regardless if that's the case.

itsjoekent:
to break the insanely long comment chain, i already ripped it out, just testing it all over the place to make sure things are good

itsjoekent:
@weerd @DFurnes updated

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
```diff
@@ -135,7 +126,9 @@ function _kudos_resource_definition() {
  *
  * @return bool
  */
-function _kudos_resource_access($op) {
+function _kudos_resource_access($op, $args = []) {
```
weerd: remove `$args` parameter if no longer using.

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
```diff
@@ -135,7 +126,9 @@ function _kudos_resource_definition() {
...
+  global $user;
```
DFurnes: This variable is no longer used, right? 👾

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
```diff
@@ -145,6 +138,7 @@ function _kudos_resource_access($op) {
   }
 
   if ($op === 'create' || $op === 'delete') {
+    // Only auth'd users should be giving Kudos
```
weerd: extraneous comment... kinda obvious from the code 😄

DFurnes:
One tiny thing, otherwise looks solid to me. 👍

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
itsjoekent: whoops, left over from before when it got broken out into different bits

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
itsjoekent: yep

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
itsjoekent: yap

weerd:
https://github.com/DoSomething/phoenix/blob/8d17e7947ac6a4cca77893886daab0ea8f0195e3/lib/themes/dosomething/paraneue_dosomething/js/components/ReactionsList.js#L115 This needs to update too right?

itsjoekent:
oh right we have two version of the JS now, good call @weerd !

itsjoekent:
okie dokes should be good to go

weerd:
¡OLÉ! :sheep: it! :crossingfingers:

PR #6973, opened 2016-08-30
add all the copy and put the share image in the share modal

### LINKED ISSUE #6965: Hard coded content for voter modals
#### User Story

content in slides 2, 3, 4 in [this](https://docs.google.com/presentation/d/1GTBZhBOlbPgff4WoLPY8WmRmq4VIT42G_2OlSDD7bCs/edit#slide=id.g16d059bc41_0_16) gdoc
#### Why This Matters

context!

katiecrane:
#### What's this PR do?
- Added the final copy in all the places we don't have fields for. 
- Added the share image into the share modal
- Require custom share AND voter reg to be on for the custom share modal since now we have text specific to voter reg in there

Make sure the text shows up in the right places (which can be found in the google doc linked from the issue)
#### Relevant tickets

Fixes #6965
#### Checklist
- [ ] Tested on staging.

cc: @ngjo

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
@@ -174,7 +174,7 @@
                   <?php endif; ?>
 
                   <?php if ($register_voters && $can_vote) : ?>
-                    <li><a href="#" data-modal-href="#modal-voter-registration">Register to Vote</a></li>
+                    <li><a href="#" data-modal-href="#modal-voter-registration"><?php print t('This online registration form (if you’re not registered already)') ?></a></li>
                   <?php endif; ?>
 
                 </ul>
@@ -235,7 +235,7 @@
             <div><?php print render($signup_data_form); ?></div>
             <?php if ($register_voters && dosomething_campaign_feature_on($campaign, 'social_share_unique_link')) : ?>
               <div class="modal__block">
-                <?php print t('Copy and paste your share link:') ?>
+                <?php print t('Use your unique registration link below for this competition. You must share this link to qualify for this competition.') ?>
                 <div class="padded-gray-box">
                   <?php print $custom_social_share_link ?>
                 </div>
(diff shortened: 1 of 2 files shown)


lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -235,7 +235,7 @@
             <div><?php print render($signup_data_form); ?></div>
             <?php if ($register_voters && dosomething_campaign_feature_on($campaign, 'social_share_unique_link')) : ?>
               <div class="modal__block">
-                <?php print t('Copy and paste your share link:') ?>
+                <?php print t('Use your unique registration link below for this competition. You must share this link to qualify for this competition.') ?>
```
angaither, line 238: 💭 shouldn't this pull from the custom variable social share language?

katiecrane: the custom social share language is currently being used as what the text to open the share modal says. The text before the link is also different depending on if it's in the competitions modal or the share modal, so unless we want to add more fields now I think it makes sense to use the share language field as it is?

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -430,12 +430,18 @@
 
 
       <?php // Custom Share Modal // ?>
-      <?php if ($custom_social_share_link) : ?>
+      <?php if ($custom_social_share_link && $register_voters) : ?>
         <div data-modal id="modal-share" role="dialog">
           <div class="modal__block">
-            <h3><?php print t('Share this Campaign') ?></h3><br>
+            <h3><?php print t('Share This Now!') ?></h3><br>
```
angaither: ugh I know this is annoying, but ideally we want to re-use this custom link across the site, and don't want it tied to votes.

Can we pre-process the text here and print it in here if `register_voters` is set? instead of requiring that var to display the custom share?

katiecrane: yeah I was kind of figuring that next time we use it there will be more fields added to deal with all of this text. but that's a really good idea for now, I'll do some shufflin'

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
angaither: @_@ okay

angaither:
lovely ✌️

PR #6984, opened 2016-08-31
Disable CSP stuffz

### LINKED ISSUE #6981: Prevent Facebook from violating the content security policy
screen shot 2016-08-31 at 10 26 36 am

angaither:
#### What's this PR do?

Unchecks some boxes so that console isn't filled up with bogus errors. 

here's what I unchecked 
image

since @sergii-tkachenko isn't able to focus on this, going to disable for now so we aren't blowing up console logs with these notifications. 
#### Relevant tickets

Refs #5783 Fixes #6981

### DIFF
--- a/lib/modules/dosomething/dosomething_settings/dosomething_settings.strongarm.inc
+++ b/lib/modules/dosomething/dosomething_settings/dosomething_settings.strongarm.inc
@@ -49,8 +49,8 @@ function dosomething_settings_strongarm() {
   $strongarm->name = 'seckit_xss';
   $strongarm->value = array(
     'csp' => array(
-      'checkbox' => 1,
-      'report-only' => 1,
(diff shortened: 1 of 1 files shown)


DFurnes:
👍 😌

sergiitk:
Wouldn't it better to just disable the module?

angaither:
@sergii-tkachenko yes, but AFAIK I don't think we have the ability to disable modules in our deploy, we only [enable modules](https://github.com/DoSomething/phoenix/blob/3759e46a598d71c0e72781be5dd3c17e47771ad4/bin/ds#L290)

DFurnes:
I believe we also use the Security Kit module for other things, don't we?

sergiitk:
@DFurnes Good catch. 👍

PR #6985, opened 2016-08-31
If user is left with no mobile or email, set a 'bad mobile' placeholder.

DFurnes:
#### What's this PR do?

When we sanitize or unset mobile numbers in our data cleanup scripts, sometimes we might end up with a user who has neither a real email or a mobile number. This happened to ~4000 users when testing on Thor. Of those users, only 255 had ever logged in.

These scripts should do the same thing, but if the user end up without a "real" email or a mobile number, then it should give them a "valid" placeholder email so they can just sync and be forgotten.

💃
#### Relevant tickets

References #6409.
#### Checklist
- [ ] Tested on staging.


angaither:
:ship:

PR #6992, opened 2016-09-01
Use a less strict regex to match placeholder emails.

DFurnes:
#### What's this PR do?

Switches out the regex for detecting placeholder "fake" mobile emails to be less strict (must end in `@mobile` or `@mobile.import`, but can have _anything_ before that). This is because we had some crazier ones before we started sanitizing mobiles properly, like `(123) 444-5555@mobile`.

👀 

📵 
#### Relevant tickets

References #6409.
#### Checklist
- [ ] Tested on staging.


angaither:
📵

PR #6994, opened 2016-09-01
Always send the mobile field so its unset if removed.

DFurnes:
#### What's this PR do?

This updates the Northstar modules "transform" method to always send the `mobile` field in the Northstar payload, so if a user ever removes their phone number from their account we send `null` and unset it on the corresponding Northstar profile.

👀

🔥
#### Relevant tickets

References #6409.
#### Checklist
- [ ] Tested on staging.


angaither:
🚢 🚢 🚢 🚢 📴

Issue #6990, opened 2016-09-01
One click sign-up research


ngjo:
#### User Story

One click sign-up from emails!
#### Why This Matters

More sign-ups to improve the RB rate (one click sign-up overview doc [here](https://docs.google.com/document/d/1D_tIeVH7EfF6eVDl1EA5gjsn1xtJ-nMKzH4cLg6QPQs/edit))

PR #7002, opened 2016-09-03
Merge multiple Signup classes into single ones.

weerd:
#### What's this PR do?

This PR does a bunch of cleanup for the **dosomething_signup** module, which contained duplicate entity classes and entity controller classes. When the signups endpoint was added, additional classes were added without cleaning up/consolidating the existing classes.

👀 

Also running some tests to check and see if Phoenix builds properly with these changes. I was having a few issues with my local, and needed to try a full `destroy` of the vagrant box and rebuild.

After the `destroy` and rebuild with these changes, site seemed to build fine and was able to signup to a few campaigns.
#### Relevant tickets

Refs
- #6990
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/includes/SignupsController.php
+++ b/lib/modules/dosomething/dosomething_signup/includes/SignupsController.php
@@ -13,47 +13,47 @@ class SignupsController extends EntityAPIController {
    *
    * Adds Signup properties into Signup entity's render.
    */
-  public function buildContent($entity, $view_mode = 'full', $langcode = NULL, $content = array()) {
+  public function buildContent($entity, $view_mode = 'full', $langcode = NULL, $content = []) {
     // Load user to get username.
     $account = user_load($entity->uid);
     $build = parent::buildContent($entity, $view_mode, $langcode, $content);
-    $build['nid'] = array(
+    $build['nid'] = [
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_signup/includes/SignupsController.php
```diff
@@ -1,73 +1,59 @@
 <?php
 
-/**
- * @file
- * Provides classes for the Signup Entity.
- */
+class SignupsController extends EntityAPIController {
```
DFurnes, line 3: Definitely not for this PR, but I wonder if it'd improve testability to have the controller classes completely independent of Drupal, and then just have a `EntityAPIRouter` class or something that acts as the "glue" to connect it. Then we could potentially run PHPUnit tests without bootstrapping the whole monstrosity.

DFurnes:
Looks good to me! 🚢

lib/modules/dosomething/dosomething_signup/includes/SignupsController.php
weerd: Ah, the Dream! I defs like the sound of that. Gonna make an issue from it for future improvements 💃

PR #7001, opened 2016-09-03
Fix bug with new members subscription DS Master Campaign on MobileCommons

sergiitk:
#### What's this PR do?
- Fixes [nasty bug](https://github.com/DoSomething/phoenix/issues/7000) with new members subscription DS Master Campaign on MobileCommons
#### How should this be reviewed?
- Go to https://www.dosomething.org/user/register
- Use phone number not registered with us
- Search for this number on https://secure.mcommons.com/profiles
- Ensure this number is subscribed to "Do Something Master Campaign"
#### Relevant tickets

Fixes #7000

cc @angaither

PR #7006, opened 2016-09-06
Fixes syntax error from #7001.

sergiitk:
Fixes syntax error introduced in #7001. Missed to comment out a bracket.


angaither:
💯

PR #7010, opened 2016-09-06
an index makes the world go round.

### LINKED ISSUE #7009: a query on the signup source is mega slow
#### Current Behavior

getting alphas and betas for competitions is 🐌 
#### Desired Behavior

let's make it fast 🐇 
#### Steps to Replicate

run this query and watch the spinner spin

```
select beta.mail as beta_email, alpha.uid, alpha.mail as alpha_email
from users beta
inner join dosomething_signup s on s.uid = beta.uid
inner join field_data_field_northstar_id ns on substring(s.source, 6, 30) = ns.field_northstar_id_value
inner join users alpha on ns.entity_id = alpha.uid
where s.source like "user/%"
and s.nid = 7059;
```

angaither:
#### What's this PR do?

Add an index to dosomething_signup.source

`drush updb -y`
see an index? watch it run.

since we are doing more and more with this col, it only makes sense to speed things up.
#### Relevant tickets

Fixes #7009
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.install
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.install
@@ -74,11 +74,12 @@ function dosomething_signup_schema() {
     'unique keys' => array(
       'uid-nid-run_nid' => array('uid', 'nid', 'run_nid'),
     ),
-    'indexes' => array(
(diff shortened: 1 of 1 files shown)


DFurnes:
🐎

PR #7018, opened 2016-09-07
Automagic signup

### LINKED ISSUE #7012: Setup new /magicks endpoint to handle the automagical signup
Create a new endpoint for `/magicks` in the DoSomething Phoenix API, that can handle a variety of URL params needed for creating the automagical signup link from emails.

### LINKED ISSUE #7013: Circumvent the signup controller save method to allow non-global user signups
The current approach with how the Signups controller handles saving a new record in the database requires a comparison between the user you're trying to signup and the logged in `global $user`. Since we're trying to do a magic signup, the `global $user` doesn't really exist and thus this always fails. Need to provide a way around this issue.

weerd:
#### What's this PR do?

This PR enables a new `/magicks` endpoint that will allow us to execute special action "spells" 🎩 ✨  via URL parameters that specify different things.

Example URI:

```
/api/v1/magicks?action=signup&user=1700013&campaign=1280&campaign_run=1832&key=somecr4zyl0ngstr1ng
```

Oh man... just 👀  it and see if anything wrong is staring back at you.

Right now the `key` parameter doesn't do anything other than being required. The idea is that eventually that can be a sort of token to help avoid spamming signups... upcoming PR once I chat w/ @DFurnes about tokens and shiz.

This is an experiment. I started adding a custom POST method to the `/signups` method, but since the main purpose of this is to provide buttons in emails with custom links that automagically signup a member for a campaign, it can't really be using the POST method.

Thus, I decided on creating the new `/magicks` endpoint to handle it. At first I tried adding this as a Services Module "action" to the `/magicks` method. After pulling all my hair out 👴  because every other HTTP method seemed to work, but the custom "action" wouldn't, I came to learn that "actions" in Services module can only be executed as POST methods.

So ultimately I settled on using an index method, and you pass specified URL pararms. The nice thing about it is it could leave open the possibility for adding more magic methods in the future and mixing and matching the URL params as needed, specifying which ones are required based on the `action=` that is requested.
#### Relevant tickets

Fixes #7012
Fixes #7013
Refs #7024
#### Checklist
- [x] Tested on staging.
- [x] Pinged a PM if this is a larger PR that would benefit from some additional testing love.


weerd:
Working on adding in the documentation at the moment!

angaither:
@weerd I thought we were going to keep with the `/user/magic` prefixs? does that not make sense?

lib/modules/dosomething/dosomething_api/resources/magick_resource.inc
```diff
@@ -0,0 +1,144 @@
...
+  return TRUE;
+}
+
+/**
+ * Mediate the request and pass along to specfied action if
+ * available.
+ *
+ * @param  string $action
+ * @param  string $campaign
+ * @param  string $user
```
angaither, line 105: can you rename the param to `$user_id` since `$user` normally refers to a fully loaded user object?

lib/modules/dosomething/dosomething_api/resources/magick_resource.inc
```diff
@@ -0,0 +1,144 @@
...
+
+  if (! $parameters['user']) {
+    return services_error('The request is missing the \'user\' parameter.', 401);
+  }
+
+  if (! $parameters['campaign']) {
+    return services_error('The request is missing the \'campaign\' parameter.', 401);
+  }
+
+  // @TODO: forcing requirement for including the campaign_run, but can likely remove this
```
angaither: yaaa! we should be able to do without this and always signup the user and get the current run with `dosomething_helpers_get_current_campaign_run_for_user`

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -269,6 +268,8 @@ function dosomething_signup_create($nid, $uid = NULL, $source = NULL, $timestamp
     'transactionals' => $transactionals,
   );
 
+  // return $values;
```
angaither: 🔥

lib/modules/dosomething/dosomething_signup/includes/SignupsController.php
```diff
@@ -2,6 +2,8 @@
 
 class SignupsController extends EntityAPIController {
 
+  protected $ignore_global_user = FALSE;
```
angaither, line 7: instead of doing this, you can pass the uid into `dosomething_signup_create` or that `signup::save` and only load the `global $user` if it isn't already passed in!

lib/modules/dosomething/dosomething_api/resources/magick_resource.inc
weerd: We don't do that elsewhere is resources which is why I went w/out the `_id`. In a resource it can never be an object, which I think why in other **something_resource.inc** files we don't use `_id`... I think we later adjust things in the transformers.

lib/modules/dosomething/dosomething_api/resources/magick_resource.inc
```diff
@@ -0,0 +1,144 @@
...
+ * @param  string $campaign
+ * @param  string $user
+ * @return array|ServicesException
+ */
+function _magick_resource_index($action, $campaign, $campaign_run, $user, $key, $source) {
+  if ($action === 'signup') {
+    return magick_signup(compact('campaign', 'campaign_run', 'user', 'key', 'source'));
+  }
+
+  return services_error('This endpoint is specifically for calling custom actions (spells). Please refer to the documentation.', 401);
```
DFurnes: Should this be a `400 Bad Request`, since it's a problem with the request (rather than an authentication thing)?

lib/modules/dosomething/dosomething_api/resources/magick_resource.inc
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * Complete the automatic campaign signup for a user via
+ * the SignupTransformer class.
+ *
+ * @param  array $parameters
+ * @return array|Exception
+ */
+function magick_signup($parameters) {
+  if (! $parameters['key']) {
```
DFurnes: Would it make sense to DRY this up with a loop? Also let's return a `422` since it's a validation error.

``` php
for (['key', 'user', 'campaign', 'campaign_run'] as $name) {
  if (! $parameters[$name]) {
    return services_error('The request is missing the \'user\' parameter.', 422);
  }
}
```

lib/modules/dosomething/dosomething_api/resources/magick_resource.inc
weerd: Yeah, I didn't add it in this first attempt because I'd likely need to do that in the transformer, and fully load the user, since `dosomething_helpers_get_current_campaign_run_for_user()` requires a user object and not just the id as well as the fully loaded campaign object. For this first run, I wanted to keep it simple, and eventually extend on it and not require the run id and just let the functions figure it all out.

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
weerd: 😱 didn't even remember adding that, haha

lib/modules/dosomething/dosomething_api/resources/magick_resource.inc
angaither: hmmm okie, can you just add some desc so it says it's the id

lib/modules/dosomething/dosomething_signup/includes/SignupTransformer.php
```diff
@@ -66,6 +66,46 @@ public function show($id) {
...
+
+      http_response_code('201');
+
+      $response = [
+        'success' => [
+          'message' => 'Signup successfully created! (But actually return the record created)',
+        ],
+      ];
+    } catch (Exception $error) {
+      http_response_code('409');
```
DFurnes: I wonder if it'd make sense to use a 422 Unprocessable Entity validation error here for consistency with how we enforce uniqueness elsewhere? Here's the [status codes that we use on Northstar](https://github.com/DoSomething/northstar/tree/dev/documentation#errors--status-codes).

lib/modules/dosomething/dosomething_signup/includes/SignupsController.php
weerd: If you look at the overridden `save()` method in the controller, you'll see that it compares passed in [user with global user](https://github.com/DoSomething/phoenix/pull/7018/files#diff-d06deaedbb4cf86687781d67ba7d5eddR120) regardless and will always if it's coming from the magic link and global user thus anon.

lib/modules/dosomething/dosomething_api/resources/magick_resource.inc
weerd: ah, jes. I'm so bad at determining best HTTP code for stuff 😞

lib/modules/dosomething/dosomething_signup/includes/SignupsController.php
```diff
@@ -59,6 +61,37 @@ public function buildContent($entity, $view_mode = 'full', $langcode = NULL, $co
...
+    // afformentioned conditional check on global $user does not happen.
+    $this->ignore_global_user = TRUE;
+
+    try {
+      $record = $this->save($signup);
+
+      // @TODO: eventually return the actual signup record. S'okay for now.
+      return TRUE;
+    } catch (Exception $error) {
+      throw new Exception('There is already an existing signup for this user on this campaign/campaign run combination.');
```
DFurnes: We should probably make a `ValidationError` exception class here that we can handle uniquely in the controller. Otherwise any sort of exception (expected or unexpected) would just get reported as a uniqueness thing.

lib/modules/dosomething/dosomething_api/resources/magick_resource.inc
weerd: Cool. Yeah I like that.

weerd:
> I thought we were going to keep with the /user/magic prefixs? does that not make sense?

@angaither nope. When @DFurnes and I chatted about it, there was stuff it didn't currently support and since this is an experiment, I think we opted for going this route.

angaither:
@weerd lawlz.

lib/modules/dosomething/dosomething_api/resources/magick_resource.inc
weerd: righty-o. I'll be in the docs too :)

I'm down for maybe changing that around... but didn't want to entirely switch unless we update the other endpoints, but those are already in use 😬

lib/modules/dosomething/dosomething_signup/includes/SignupsController.php
DFurnes: Yeah, I think it'd make sense to add a new parameter to the `save()` method for that rather than relying on `global $user` in there. Having to switch modes like this feels like it'd be easy to make a mistake with.

weerd: Yeah, I was just afraid to change the current `save()` method too much and potentially break signups which are kind of a crucial feature. I make notes to eventually find a better solution for this, since it's not ideal, but figured not worth modifying if the `/magicks` experiment doesn't work out.

angaither: @weerd hmm, I think I'd rather we update the method to accept the user as a param than have this work around

weerd: Oh do you mean updating the method to not require the `global $user` and only use the passed in one?

weerd: I added an issue and [referenced it here](https://github.com/DoSomething/phoenix/pull/7018/files#diff-d06deaedbb4cf86687781d67ba7d5eddR118) since it's something I'm kind of concerned to update just for an experiment. Drupal doesn't necessarily instantiate these classes how we'd expect... it uses functions to call them via names saved as properties on the entityInfo stuff which makes it really difficult to find all occurrences of classes use and methods referenced. There seemed to be a very particular scenario as to why we have overridden the `save()` method on the controller and check against the global user when this module was created. Since we have no testing in place I'm rather concerned to mess with this too much, fearing I'll entirely break signups. This felt like a decent enough temporary solution for something we're not even sure will stick around. So if `/magicks` does end up working out well, we can take some time to really do some cleanup of the signups module.

angaither: yeah fair, I am worried that this isn't really an experiment and we won't ever clean up and this introduces a lot of confusion/complexity to the code.

documentation/adding_and_updating_endpoints.md
```diff
@@ -0,0 +1,83 @@
...
+
+  return $resources;
+}
+ 
+?>
+```
+
+4) Clear the cache using `drush cc all`. Everytime you edit the array in the definitions function in the **example_resource.inc** file you need to be sure to clear the Drupal cache, otherwise your changes will not take effect!
+
+5) Profit! (err, Non-profit!)
```
DFurnes: 😂

DFurnes:
Looks good to me! 🚢 ⚠️ 🎉

documentation/adding_and_updating_endpoints.md
```diff
@@ -0,0 +1,83 @@
+# Adding And Updating New Phoenix Endpoints
+
+***
+
+
```
angaither: 💭 this maybe belongs in the wiki section, since it's not an endpoint, and we have other 'how tos' for devs in that area

weerd: Yeah, sounds good. I might reference a link to it in this endpoints doc to the wiki though.

angaither:
🔮

weerd:
akthanakos_teleporting ancient magicks

PR #7007, opened 2016-09-06
Extend quicksilver user registration payload with Northstar id.

sergiitk:
#### What's this PR do?

This PR extends `user_register` Quicksilver payload issued for new users with optional `northstar_id` field.
#### How should this be reviewed?
- Register on the site with new mobile phone.
- Check the corresponding message in RabbitMQ, queue `mobileCommonsQueue`. It should contain `northstar_id` field.

Example:

```
a:16:{s:8:"activity";s:13:"user_register";s:5:"email";s:35:"***CENSORED***";s:3:"uid";s:7:"1708309";s:10:"merge_vars";a:2:{s:12:"MEMBER_COUNT";s:11:"3.5 million";s:5:"FNAME";s:6:"Sergii";}s:13:"user_language";s:2:"en";s:12:"northstar_id";s:24:""***CENSORED***";s:12:"user_country";s:2:"US";s:14:"email_template";s:19:"mb-user-register-US";s:17:"mailchimp_list_id";s:10:""***CENSORED***";s:9:"birthdate";s:9:"852163200";s:10:"subscribed";i:1;s:6:"mobile";s:10:"3478227222";s:17:"mc_opt_in_path_id";s:6:"164905";s:10:"email_tags";a:1:{i:0;s:20:"drupal_user_register";}s:18:"activity_timestamp";i:1473185803;s:14:"application_id";s:2:"US";}
```
#### Relevant tickets

A part of Lose your v-card messaging campaign tasks.
https://trello.com/c/tZMqE6kc/110-lose-your-v-card-signup-email-sms-data-import

PR #7021, opened 2016-09-07
Extend quicksilver payload with optional Northstar ID.

sergiitk:
#### What's this PR do?

This PR extends `dosomething_signup` Quicksilver payload issued for campaign subscriptions with optional `northstar_id` field.
- Sign up for a campaign.
- Check the corresponding message in RabbitMQ, queue `mobileCommonsQueue`. It should contain `northstar_id` field.

Example:

```
a:18:{s:8:"activity";s:15:"campaign_signup";s:5:"email";s:35:"sergii+test-09316-4@dosomething.org";s:3:"uid";s:7:"1708317";s:10:"merge_vars";a:8:{s:12:"MEMBER_COUNT";s:11:"3.5 million";s:5:"FNAME";s:6:"Sergii";s:14:"CAMPAIGN_TITLE";s:33:"Hunt Day 2: Trash Scavenger Hunt!";s:13:"CAMPAIGN_LINK";s:93:"http://dev.dosomething.org:8888/us/campaigns/hunt-day-2-trash-scavenger-hunt?source=node/1286";s:14:"CALL_TO_ACTION";s:49:"Show your friends that trash cleanups can be fun.";s:8:"STEP_ONE";s:16:"Start Your Hunt!";s:8:"STEP_TWO";s:10:"Snap a Pic";s:10:"STEP_THREE";s:7:"Toss It";}s:14:"transactionals";b:1;s:13:"user_language";s:2:"en";s:17:"campaign_language";s:2:"en";s:16:"campaign_country";s:2:"US";s:12:"northstar_id";s:24:"57cf0eb4a59dbf593d8b4808";s:12:"user_country";s:2:"US";s:14:"email_template";s:21:"mb-campaign-signup-US";s:10:"subscribed";i:1;s:8:"event_id";s:4:"1286";s:10:"email_tags";a:2:{i:0;s:4:"1286";i:1;s:22:"drupal_campaign_signup";}s:6:"mobile";s:10:"3478227222";s:17:"mc_opt_in_path_id";s:6:"167401";s:18:"activity_timestamp";i:1473274141;s:14:"application_id";s:2:"US";}
```
#### Relevant tickets

A part of Lose your v-card messaging campaign tasks.
https://trello.com/c/tZMqE6kc/110-lose-your-v-card-signup-email-sms-data-import

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -248,6 +248,35 @@ function dosomething_northstar_save_id_field($uid, $northstar_response) {
 }
 
 /**
+ * Save the user's Northstar ID to their local profile.
+ *
+ * @param object $account
+ *   Fully loaded user object.
+ *
+ * @return string
+ *   The Northstar id or empty string on error.
+ */
+function dosomething_northstar_get_id($account = NULL) {
+  if (!$account) {
+    global $user;
+    $account = $user;
+  }
+
+  try {
+    $northstar_id = dosomething_user_get_field('field_northstar_id', $account);
+    // NONE is a special value representing that Northstar id is not present.
+    // @see dosomething_northstar_save_id_field();
+    if (!empty($northstar_id) && $northstar_id !== 'NONE') {
+      return $northstar_id;
+    }
+  } catch (EntityMetadataWrapperException $e) {
+    // Don't fall into error 500 when the field is not accessible.
+    watchdog('dosomething_northstar', 'Northsar user doesn\'t exist for: ' . $account->mail, NULL, WATCHDOG_NOTICE);
+  }
+  return '';
+}
+
+/**
  * If the log is enabled, log this request to the database.
  *
  * @param string $op - label for the operation being performed
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -248,35 +248,6 @@ function dosomething_northstar_save_id_field($uid, $northstar_response) {
 }
 
 /**
- * Save the user's Northstar ID to their local profile.
- *
- * @param object $account
- *   Fully loaded user object.
- *
- * @return string
- *   The Northstar id or empty string on error.
- */
-function dosomething_northstar_get_id($account = NULL) {
-  if (!$account) {
-    global $user;
-    $account = $user;
-  }
-
-  try {
-    $northstar_id = dosomething_user_get_field('field_northstar_id', $account);
-    // NONE is a special value representing that Northstar id is not present.
-    // @see dosomething_northstar_save_id_field();
-    if (!empty($northstar_id) && $northstar_id !== 'NONE') {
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -694,6 +694,21 @@ function dosomething_signup_get_mbp_params($account, $node, $opt_in, $options =
     'campaign_country' => $campaign_country_code,
   ];
 
+  // Northstar id.
+  if (module_exists('dosomething_northstar')) {
```
angaither: can we DRY this and [this](https://github.com/DoSomething/phoenix/pull/7007/files#diff-ad30a5d569ca2ae6d2e04223aa501104R644) up?

sergiitk: Yes, makes sense.

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -248,6 +248,35 @@ function dosomething_northstar_save_id_field($uid, $northstar_response) {
...
 /**
+ * Save the user's Northstar ID to their local profile.
+ *
+ * @param object $account
+ *   Fully loaded user object.
+ *
+ * @return string
+ *   The Northstar id or empty string on error.
+ */
+function dosomething_northstar_get_id($account = NULL) {
```
angaither: hmm instead of putting this here it might make more sense to put this in the `dosomething_user` module as a [hander](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_user/dosomething_user.module#L1264) for the `dosomething_user_get_field` function 
so that way you just have to call the function to get a field like normal and it returns this result alllways.

sergiitk: It won't work as a handler for `dosomething_user_get_field()`: it'd fail on systems where Northstar is turned off (staging, local) because of this: https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_user/dosomething_user.module#L1261.

angaither: hmm, the module isn't turned off on stage, only locally, and there should always be values there locally b/c you pull the database from stage, that shouldn't fail

sergiitk: I tested it — and it fails. That's why I added `try {} } catch (EntityMetadataWrapperException $e) {}` block.

angaither: hmm, then perhaps that too belongs in the `dosomething_user_get_field` block.
in an effort to simplify the code, it would be really nice to have one way to call field info on a user, everywhere instead of breaking out another one off solution here.

sergiitk: Yes, I see your point. Will do the refactoring then!

sergiitk:
@angaither ready!

angaither:
yais! :shipit:

PR #7022, opened 2016-09-07
put the gallery heading outside the gallery ul

### LINKED ISSUE #6945: Advisory Board: First member's label disappears on narrow views
#### Current Behavior

As you resize the window, the first item in each gallery loses its `figure__body`.

Seems to be something with the `overflow` rule and where that element gets pushed to.
#### Desired Behavior

All names/labels should stay visible for all viewport sizes
#### Steps to Replicate

Replicated easily on Chrome; haven't checked other pages with similar galleries.
#### Why This Matters

Because people have names, even in small windows
#### OS/Browser (optional)

Def latest Chrome / OS X

katiecrane:
#### What's this PR do?

Puts the `gallery__heading` div outside of the `gallery` ul.

Make sure that this looks right and doesn't mess up all galleries everywhere.

The problem in the issue goes away if you hide the heading altogether as well, so I think something to do with the non-li div inside the ul was funky once it got little.
#### Relevant tickets

Fixes #6945
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/system/patterns/gallery.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/system/patterns/gallery.tpl.php
@@ -14,12 +14,12 @@
  * @see paraneue_dosomething_get_gallery()
  */
 ?>
+<?php if (!empty($title)): ?>
+  <div class="gallery__heading">
+    <h1><?php print $title; ?></h1>
+  </div>
+<?php endif; ?>
 <ul class="gallery -<?php print $layout; ?> <?php print $classes; ?>" <?php if (isset($roles)) { print $roles; } ?>>
-  <?php if (!empty($title)): ?>
-    <div class="gallery__heading">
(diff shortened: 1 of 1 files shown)


DFurnes:
I tested this fix in the pattern library and it seems to mucks up the padding a bit:

screen shot 2016-09-09 at 10 54 02 am

I think you might be able to fix it without adjusting the markup by adding an `overflow: hidden` to the `li` elements so they don't overflow out of their containers and look crazy (check out the figure's elements in the web inspector and you can see they're just being pushed outside of the visible area).

katiecrane:
Oh hey look at that. That does work! Although tbh I can't tell that the padding is weird in that screenshot. 

Haha yeah I could see the hidden thing there but unhiding it was a whole thing! 

CSS. `¯\_(ツ)_/¯`

katiecrane:
closing this for [the PR in forge](https://github.com/DoSomething/forge/pull/562)

PR #7023, opened 2016-09-07
Adds most of the Facebook login functionality

### LINKED ISSUE #6995: Retrieve user profile info upon login
When a user logs in via Facebook we need to grab basic profile information from them,
- [x] First name
- [x] Last name
- [x] Email

Other details _could_ be grabbed, but I would rather not block us from continuing by waiting on the Facebook review process. If we pass review at some point we'll just make another issue / reopen this to get the rest of the profile data.

### LINKED ISSUE #6996: Create Northstar account after FB Login
## Blocked by: https://github.com/DoSomething/phoenix/issues/6995

Once we grab all of the users profile information via the SDK we'll need to tell Northstar. [This pull request](https://github.com/DoSomething/northstar/pull/411) added support for storing Facebook ID's. However creating a new account with Northstar requires admin scope, so this will probably need to route through the Drupal API (we'll need Drupal endpoints later on anyway to actually do the login).

From Drupal we'll attempt to create the new NS account. If it passes, hooray! (Eventually this will lead into the user being logged in).

If it fails however, we should check the error to see if the email / facebook ID was duplicated. If it was, we can do a PUT request for that specific user & merge the ID's all into one account.

itsjoekent:
#### What's this PR do?
- NS Helper to get user by different terms
- Updates API client to be a bit more helpful for my login form
- Adds a new endpoint under /users/ for people to login via Facebook
- Does a bunch of communication with NS to get Facebook auth working
- Puts the Facebook button inside a form which the Javascript can fake trigger / append user details too

Main test scenarios I've been running through
- Fresh new account
- Facebook email matches existing Northstar account
- Matching Facebook ID

For debugging I highly suggest making multiple tabs in Postman (or whatever rest client you use)
My first tab is setup to grab oauth credentials (they expire often enough to warrant this)
Second tab is setup to grab my NS user based on my Facebook email 
`northstar-qa.dosomething.org/v1/users/email/thedeadlybutter@gmail.com`
Third tab is setup to delete the entire user based on user ID (Get user ID from tab 2)
Fourth tab is setup to set the facebook ID to null based on user ID (Get user id from tab 2)
1. DO NOT MAKE THIS LIVE (duh)
   This works great for setting up users but DOES NOT AUTHENTICATE BEYOND THIS POINT. I need to add another endpoint in Northstar to quickly check your oauth token.
2. The function is pretty freaking big, I tried breaking it up where it was easy to do so. Unfortunately there is just a lot of steps involved & a specific sequence they must happen. 
#### Relevant tickets

Fixes #6995
Fixes #6996
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -158,12 +158,12 @@ function dosomething_northstar_get_northstar_user($drupal_id) {
 }
 
 /**
- * Checks if a Northstar user exists for a given index & type.
+ * Get a Northstar user for a given index & type. Useful for non Drupal ID lookups.
  * @param  string $index_type The type of index to query for (Email, Facebook, Mobile, _id)
  * @param  string $index The index to query for
- * @return bool|object
+ * @return object|null
  */
-function dosomething_northstar_user_exists_by_index($index_type, $index) {
+function dosomething_northstar_get_user_by_custom_index($index_type, $index) {
   $client = _dosomething_northstar_build_http_client();
   $response = drupal_http_request($client['base_url'] . '/users/' . $index_type . '/' . $index, [
     'headers' => $client['headers'],
@@ -174,7 +174,7 @@ function dosomething_northstar_user_exists_by_index($index_type, $index) {
     return json_decode($response->data);
   }
 
-  return FALSE;
+  return null;
 }
 
 /**
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -125,49 +125,39 @@ function dosomething_northstar_update_user($user, $password = null) {
 }
 
 /**
- * Get user profile data from Northstar for the given Drupal ID.
+ * Get user profile data from Northstar for the given ID.
  *
- * @param int $drupal_id Drupal user id.
+ * @param int $id Drupal user id.
+ * @param string $type The index type of the ID
  * @return object
  */
-function dosomething_northstar_get_northstar_user($drupal_id) {
-  $northstar_user_info = cache_get('northstar_user_info_' . $drupal_id, 'cache_dosomething_northstar');
+function dosomething_northstar_get_northstar_user($id, $type = 'drupal_id') {
+  $northstar_user_info = NULL;
+
+  if ($type === 'drupal_id') {
+    $northstar_user_info = cache_get('northstar_user_info_' . $id, 'cache_dosomething_northstar');
+  }
 
   if ($northstar_user_info) {
     $northstar_user_info = $northstar_user_info->data;
   }
   else {
-    $northstar_user_info = dosomething_northstar_get_user_by_custom_index('drupal_id', $drupal_id);
+    $client = _dosomething_northstar_build_http_client();
+    $response = drupal_http_request($client['base_url'] . '/users/' . $type . '/' . $id, [
+      'headers' => $client['headers'],
+      'method' => 'GET',
+    ]);
+    $northstar_user_info = $response->data;
 
     if (!empty($northstar_user_info->error)) {
-      $error = sprintf("Error fetching Northstar user data, uid=%d: '%s'", $drupal_id, $northstar_user_info->error);
+      $error = sprintf("Error fetching Northstar user data, uid=%d: '%s'", $id, $northstar_user_info->error);
       watchdog_exception('northstar', new Exception($error));
     } else {
-      cache_set('northstar_user_info_' . $drupal_id, $northstar_user_info, 'cache_dosomething_northstar', REQUEST_TIME + 60*60*24);
+      cache_set('northstar_user_info_' . json_decode($northstar_user_info)->data->drupal_id, $northstar_user_info, 'cache_dosomething_northstar', REQUEST_TIME + 60*60*24);
     }
   }
 
   return $northstar_user_info;
-}
-
-/**
- * Get a Northstar user for a given index & type. Useful for non Drupal ID lookups.
- * @param  string $index_type The type of index to query for (Email, Facebook, Mobile, _id)
- * @param  string $index The index to query for
- * @return object|null
- */
-function dosomething_northstar_get_user_by_custom_index($index_type, $index) {
-  $client = _dosomething_northstar_build_http_client();
-  $response = drupal_http_request($client['base_url'] . '/users/' . $index_type . '/' . $index, [
-    'headers' => $client['headers'],
-    'method' => 'GET',
-  ]);
-
-  if ($response->code === "200") {
-    return $response->data;
-  }
-
-  return NULL;
 }
 
 /**
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -174,7 +174,7 @@ function dosomething_northstar_get_user_by_custom_index($index_type, $index) {
     return json_decode($response->data);
   }
 
-  return null;
+  return NULL;
 }
 
 /**
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -157,9 +157,15 @@ function dosomething_northstar_get_northstar_user($drupal_id) {
   return $northstar_user_info;
 }
 
-function dosomething_northstar_get_user_by_term($term, $id) {
+/**
+ * Checks if a Northstar user exists for a given index & type.
+ * @param  string $index_type The type of index to query for (Email, Facebook, Mobile, _id)
+ * @param  string $index The index to query for
+ * @return bool Whether the user exists
+ */
+function dosomething_northstar_user_exists_by_index($index_type, $index) {
   $client = _dosomething_northstar_build_http_client();
-  $response = drupal_http_request($client['base_url'] . '\/users/' . $term . '/' . $id, [
+  $response = drupal_http_request($client['base_url'] . '\/users/' . $index_type . '/' . $index, [
     'headers' => $client['headers'],
     'method' => 'GET',
   ]);
@@ -179,6 +185,7 @@ function dosomething_northstar_transform_user($user, $password = null) {
     'birthdate' => 'field_birthdate',
     'first_name' => 'field_first_name',
     'last_name' => 'field_last_name',
+    'facebook_id' => 'field_facebook_id',
   ];
 
   // Address fields
@@ -237,7 +244,7 @@ function dosomething_northstar_transform_user($user, $password = null) {
   if(empty($northstar_user['source'])) {
     $northstar_user['source'] = 'phoenix';
   }
-
+  print_r($northstar_user);die();
   return $northstar_user;
 }
 
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -157,6 +157,15 @@ function dosomething_northstar_get_northstar_user($drupal_id) {
   return $northstar_user_info;
 }
 
+function dosomething_northstar_get_user_by_term($term, $id) {
+  $client = _dosomething_northstar_build_http_client();
+  $response = drupal_http_request($client['base_url'] . '\/users/' . $term . '/' . $id, [
+    'headers' => $client['headers'],
+    'method' => 'GET',
+  ]);
+  return $response->code !== "404";
+}
+
 /**
  * Transform Drupal user fields into the appropriate schema for Northstar.
  *
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -161,7 +161,7 @@ function dosomething_northstar_get_northstar_user($drupal_id) {
  * Checks if a Northstar user exists for a given index & type.
  * @param  string $index_type The type of index to query for (Email, Facebook, Mobile, _id)
  * @param  string $index The index to query for
- * @return bool Whether the user exists
+ * @return bool|object
  */
 function dosomething_northstar_user_exists_by_index($index_type, $index) {
   $client = _dosomething_northstar_build_http_client();
@@ -169,7 +169,12 @@ function dosomething_northstar_user_exists_by_index($index_type, $index) {
     'headers' => $client['headers'],
     'method' => 'GET',
   ]);
-  return $response->code !== "404";
+
+  if ($response->code === "200") {
+    return json_decode($response->body);
+  }
+
+  return FALSE;
 }
 
 /**
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -137,21 +137,14 @@ function dosomething_northstar_get_northstar_user($drupal_id) {
     $northstar_user_info = $northstar_user_info->data;
   }
   else {
-    $client = _dosomething_northstar_build_http_client();
+    $northstar_user_info = dosomething_northstar_get_user_by_custom_index('drupal_id', $drupal_id);
 
-    $northstar_user_info = drupal_http_request($client['base_url'] . '/users/drupal_id/' . $drupal_id, array(
-      'headers' => $client['headers'],
-      'method' => 'GET',
-      ));
-
-      $northstar_user_info = $northstar_user_info->data;
-
-      if (!empty($northstar_user_info->error)) {
-        $error = sprintf("Error fetching Northstar user data, uid=%d: '%s'", $drupal_id, $northstar_user_info->error);
-        watchdog_exception('northstar', new Exception($error));
-      } else {
-        cache_set('northstar_user_info_' . $drupal_id, $northstar_user_info, 'cache_dosomething_northstar', REQUEST_TIME + 60*60*24);
-      }
+    if (!empty($northstar_user_info->error)) {
+      $error = sprintf("Error fetching Northstar user data, uid=%d: '%s'", $drupal_id, $northstar_user_info->error);
+      watchdog_exception('northstar', new Exception($error));
+    } else {
+      cache_set('northstar_user_info_' . $drupal_id, $northstar_user_info, 'cache_dosomething_northstar', REQUEST_TIME + 60*60*24);
+    }
   }
 
   return $northstar_user_info;
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -216,9 +216,8 @@ function dosomething_northstar_transform_user($user, $password = null) {
       continue;
     }
 
-    $field = $user->$drupal_key;
-    if (isset($field[LANGUAGE_NONE][0]['value'])) {
-      $northstar_user[$ns_key] = $field[LANGUAGE_NONE][0]['value'];
+    if (isset($user->$drupal_key[LANGUAGE_NONE][0]['value'])) {
+      $northstar_user[$ns_key] = $user->$drupal_key[LANGUAGE_NONE][0]['value'];
     }
   }
   foreach ($address as $ns_key => $drupal_key) {
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -205,11 +205,7 @@ function dosomething_northstar_transform_user($user, $password = null) {
 
   // Set values in ns_user if they are set.
   foreach ($optional as $ns_key => $drupal_key) {
-    if (! isset($user->$drupal_key)) {
-      continue;
-    }
-
-    if (isset($user->$drupal_key[LANGUAGE_NONE][0]['value'])) {
+    if (isset($user->{$drupal_key}[LANGUAGE_NONE][0]['value'])) {
       $northstar_user[$ns_key] = $user->$drupal_key[LANGUAGE_NONE][0]['value'];
     }
   }
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -164,7 +164,7 @@ function dosomething_northstar_get_user_by_custom_index($index_type, $index) {
   ]);
 
   if ($response->code === "200") {
-    return json_decode($response->data);
+    return $response->data;
   }
 
   return NULL;
@@ -206,7 +206,7 @@ function dosomething_northstar_transform_user($user, $password = null) {
   // Set values in ns_user if they are set.
   foreach ($optional as $ns_key => $drupal_key) {
     if (isset($user->{$drupal_key}[LANGUAGE_NONE][0]['value'])) {
-      $northstar_user[$ns_key] = $user->$drupal_key[LANGUAGE_NONE][0]['value'];
+      $northstar_user[$ns_key] = $user->{$drupal_key}[LANGUAGE_NONE][0]['value'];
     }
   }
   foreach ($address as $ns_key => $drupal_key) {
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -165,13 +165,13 @@ function dosomething_northstar_get_northstar_user($drupal_id) {
  */
 function dosomething_northstar_user_exists_by_index($index_type, $index) {
   $client = _dosomething_northstar_build_http_client();
-  $response = drupal_http_request($client['base_url'] . '\/users/' . $index_type . '/' . $index, [
+  $response = drupal_http_request($client['base_url'] . '/users/' . $index_type . '/' . $index, [
     'headers' => $client['headers'],
     'method' => 'GET',
   ]);
 
   if ($response->code === "200") {
-    return json_decode($response->body);
+    return json_decode($response->data);
   }
 
   return FALSE;
@@ -190,7 +190,6 @@ function dosomething_northstar_transform_user($user, $password = null) {
     'birthdate' => 'field_birthdate',
     'first_name' => 'field_first_name',
     'last_name' => 'field_last_name',
-    'facebook_id' => 'field_facebook_id',
   ];
 
   // Address fields
@@ -213,6 +212,10 @@ function dosomething_northstar_transform_user($user, $password = null) {
 
   // Set values in ns_user if they are set.
   foreach ($optional as $ns_key => $drupal_key) {
+    if (! isset($user->$drupal_key)) {
+      continue;
+    }
+
     $field = $user->$drupal_key;
     if (isset($field[LANGUAGE_NONE][0]['value'])) {
       $northstar_user[$ns_key] = $field[LANGUAGE_NONE][0]['value'];
@@ -249,7 +252,12 @@ function dosomething_northstar_transform_user($user, $password = null) {
   if(empty($northstar_user['source'])) {
     $northstar_user['source'] = 'phoenix';
   }
-  print_r($northstar_user);die();
+
+  // Facebook ID is not in the field_ format
+  if (isset($user->facebook_id)) {
+    $northstar_user['facebook_id'] = $user->facebook_id;
+  }
+
   return $northstar_user;
 }
 
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -114,7 +114,17 @@ function dosomething_user_menu() {
     'page arguments' => array(2, 3, 4),
     'access callback' => TRUE,
   );
+  $items['user/social/facebook'] = [
+    'title' => 'Facebook Login',
+    'page callback' => 'dosomething_user_facebook_login',
+    'access callback' => 'dosomething_user_facebook_login_enabled',
+    'file' => 'dosomething_user_social_auth.inc',
+  ];
   return $items;
+}
+
+function dosomething_user_facebook_login_enabled() {
+  return variable_get('dosomething_user_social_auth_enabled', FALSE) && module_exists('dosomething_northstar');
 }
 
 function dosomething_user_get_view($id) {
@@ -1345,7 +1355,6 @@ function dosomething_user_set_fields(&$edit, $fields) {
     }
 
     switch ($key) {
-      case 'country':
       case 'postal_code':
         $edit['field_address'][LANGUAGE_NONE][0][$key] = $value;
         break;
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -123,6 +123,10 @@ function dosomething_user_menu() {
   return $items;
 }
 
+/**
+ * Is the Facebook login functionality enabled
+ * @return bool
+ */
 function dosomething_user_facebook_login_enabled() {
   return variable_get('dosomething_user_social_auth_enabled', FALSE) && module_exists('dosomething_northstar');
 }
@@ -1355,6 +1359,7 @@ function dosomething_user_set_fields(&$edit, $fields) {
     }
 
     switch ($key) {
+      case 'country':
       case 'postal_code':
         $edit['field_address'][LANGUAGE_NONE][0][$key] = $value;
         break;
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -598,6 +598,23 @@ function dosomething_user_update_user($form, &$form_state) {
 }
 
 /**
+ * Takes care of misc. tasks required for a new user.
+ *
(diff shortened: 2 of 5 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    if ($create_new_account) {
+      // Create the Drupal user.
+      // Unfortunately there doesn't seem to be anyway around this because of how Drupal is architected.
+      $user = dosomething_user_make_facebook_user($account_details);
+      $account_details['drupal_id'] = $user->uid;
+
+      // Create Northstar user w/ Facebook ID
+      $response = drupal_http_request($client['base_url'] . '/users', [
```
itsjoekent: I know we have Northstar functions for this and the PUT below, but I was worried that they wouldnt work right because they use the global $user object & we're not fully logged in / built yet

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -157,6 +157,15 @@ function dosomething_northstar_get_northstar_user($drupal_id) {
   return $northstar_user_info;
 }
 
+function dosomething_northstar_get_user_by_term($term, $id) {
```
DFurnes: This needs a docblock.

DFurnes: If this isn't returning a user, perhaps `get_user_by_term` is poorly phrased? Maybe `user_exists_by_term`?

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
```diff
@@ -0,0 +1,98 @@
+<?php
+
+function dosomething_user_make_facebook_user($account_details) {
+  $edit = [
+    'mail' => $account_details['email'],
+    'name' => $account_details['email'],
```
DFurnes, line 11: We usually set the name to the user's `uid`. Should we do that here too?

DFurnes, line 11: Oh never mind, I see what's happening below. Mind adding a inline comment here for clarity?

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
itsjoekent: ah yes that makes sense

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
itsjoekent: sure thing

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
```diff
@@ -0,0 +1,98 @@
...
+  $user = user_save(null, $edit);
+  user_save($user, ['name' => $user->uid]);
+
+  return $user;
+}
+
+function dosomething_user_facebook_login() {
+  global $user;
+
+  if (!isset($_POST['facebook_id']) || !isset($_POST['access_token']) || !isset($_POST['email'])) {
```
DFurnes, line 92: Doesn't services have a way of setting and passing arguments through the resource definition?

itsjoekent: this isn't services, it used to be! I moved it out of there & into the user module, which as far as I can tell is why it started working.

given that, from what I researched / saw in our own code (eg: dosomething_user_valid_address.inc), using $_POST was the way to go

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
DFurnes: I don't see where the global `$user` is used in the [`dosomething_northstar_create_user`](https://github.com/DoSomething/phoenix/blob/3662081c9e7c8f2b89448ad585d2f176b08f1c7f/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module#L66) or [`dosomething_user_update_user`](https://github.com/DoSomething/phoenix/blob/3662081c9e7c8f2b89448ad585d2f176b08f1c7f/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module#L99) functions. Can you explain the problems you were running into?

itsjoekent: Oof your right its not! :facepalm: 

I think For the `update_user` function tho I might have to make a small modification to supply a custom payload vs a transformed user object, and I'll also have to update the transformer anyway to account for Facebook ID when we create.

itsjoekent: Actually I guess if I load the user first, and then add the facebook ID, i can throw it to update_user without a problem? in theory...

DFurnes: Yeah, that's the right way to do this.

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
```diff
@@ -0,0 +1,98 @@
...
+        'headers' => $client['headers'],
+        'method' => 'POST',
+        'data' => json_encode($account_details),
+      ]);
+
+      // No need to run a verify on the initial signup.
+      $verify_token = FALSE;
+      $logged_in = TRUE;
+    }
+    else {
```
DFurnes: There's way too much branching logic going on here. Can you extract some of these into their own methods so it's easier to follow what's going on?

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -114,9 +114,19 @@ function dosomething_user_menu() {
...
+  $items['user/social/facebook'] = [
+    'title' => 'Facebook Login',
+    'page callback' => 'dosomething_user_facebook_login',
+    'access callback' => 'dosomething_user_facebook_login_enabled',
+    'file' => 'dosomething_user_social_auth.inc',
+  ];
   return $items;
 }
 
+function dosomething_user_facebook_login_enabled() {
```
angaither, line 130: docblock

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -1286,7 +1296,6 @@ function dosomething_user_set_fields(&$edit, $fields) {
     }
 
     switch ($key) {
-      case 'country':
```
angaither, line 1348: should this line really be removed?

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
```diff
@@ -0,0 +1,98 @@
...
+    'mail' => $account_details['email'],
+    'name' => $account_details['email'],
+    'pass' => user_password(),
+    'status' => 1,
+    'created' => REQUEST_TIME,
+  ];
+
+  dosomething_user_set_fields($edit, $account_details);
+
+  $user = user_save(null, $edit);
```
angaither: shouldn't the first arg of this be `$user`? instead of `null`?

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
```diff
@@ -0,0 +1,98 @@
...
+
+  $account_details = [
+    'email' => data_get($_POST, 'email'),
+    'first_name' => data_get($_POST, 'first_name'),
+    'last_name' => data_get($_POST, 'last_name'),
+    'facebook_id' => $facebook_id,
+    'user_registration_source' => 'facebook_auth',
+  ];
+
+  if (dosomething_northstar_get_user_by_term('facebook_id', $facebook_id)) {
```
angaither: I don't think this term `facebook_id` is a supported lookup by NS

angaither:
all new methods need docblocks

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
itsjoekent: in this case $user is an anonymous user, so in order to make it work in this function,

> If you want to create a new user account, you can set $account->is_new to TRUE or omit the $account->uid field.
> https://api.drupal.org/api/drupal/modules%21user%21user.module/function/user_save/7.x

I saw we used something similar to this in the member resource API https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_api/resources/member_resource.inc#L218

but I figured NULL made more sense then `''`

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
itsjoekent: I added that!

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
itsjoekent: sure thing

lib/modules/dosomething/dosomething_user/dosomething_user.module
itsjoekent: nope! that was from debugging a different issue I was running into

(tho it does beg the question why this was here to begin with, there isnt even a break statement)

angaither: it does below postal code, it's handling multiple things

itsjoekent: oh interesting, i didnt know switch statements worked that way

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
```diff
@@ -0,0 +1,130 @@
...
+
+  if ($create_new_account) {
+    // Create the Drupal user.
+    // Unfortunately there doesn't seem to be anyway around this because of how Drupal is architected.
+    $user = dosomething_user_make_facebook_user($account_details);
+    dosomething_northstar_create_user($user);
+  }
+  else {
+    // Update existing Northstar user to link Facebook ID
+    $user->facebook_id = $account_details['facebook_id'];
```
itsjoekent: @DFurnes I was toying with adding this here, but then i realized this user is anonymous, so NS wont know what to do with it?

lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
```diff
@@ -18,7 +18,8 @@ class ApiClient {
     this.host = host;
     this.port = port === '' ? null : port;
     this.version = `/api/${version}/`;
-    this.url = `//${this.host}${this.port ? ':' + this.port : ''}${this.version}`;
+    this.hostname = `//${this.host}${this.port ? ':' + this.port : ''}`;
```
weerd: would it be better to call this `this.domain`? :shrug: I feel like `host` and `hostname` seem like the same thing.

itsjoekent: yeah that works 👍

itsjoekent:
Have not run through all of the tests yet, going to do so now

itsjoekent:
All of my testing seems to work good 👍

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -192,6 +212,10 @@ function dosomething_northstar_transform_user($user, $password = null) {
 
   // Set values in ns_user if they are set.
   foreach ($optional as $ns_key => $drupal_key) {
+    if (! isset($user->$drupal_key)) {
```
itsjoekent: Added this because _currently_ Facebook users dont have birthdays being added, was throwing warnings

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -158,6 +158,26 @@ function dosomething_northstar_get_northstar_user($drupal_id) {
 }
 
 /**
+ * Checks if a Northstar user exists for a given index & type.
+ * @param  string $index_type The type of index to query for (Email, Facebook, Mobile, _id)
+ * @param  string $index The index to query for
+ * @return bool|object
```
DFurnes: Oh hmm, this feels iffy to me. Can we return either `object|null` or just `bool`? Not sure if the method changed from my first read-through, but it seems like it does different things rather unexpectedly (and so it isn't really a `exists_by_index` _or_ a `get_user_by_index`... it's a bit of both!)

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
DFurnes: Can we combine this with the `isset` on [line 220](https://github.com/DoSomething/phoenix/pull/7023/files#diff-6fba28391e7b23bdda9b664d3a71c59fR220) in that case (by inlining the `$user->{$drupal_key}` part)?

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
```diff
@@ -0,0 +1,122 @@
...
+  $user = user_save(null, $edit);
+  user_save($user, ['name' => $user->uid]);
+
+  return $user;
+}
+
+/**
+ * Registers a new user with Northstar (or upserts with existing) & returns a Drupal account
+ * @param  array $account_details Account details for the user to register
+ * @return $user Created (or retrieved) Drupal user
```
DFurnes: The [`@return` annotation](https://www.phpdoc.org/docs/latest/references/phpdoc/tags/return.html) should have a type, not a name.

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
itsjoekent: sure

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
itsjoekent: Yeah I ended up changing it to a bit of both, I can rename it & return object|null

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
itsjoekent: actually @DFurnes no i cant, `$field = $user->$drupal_key;` was the code triggering the error. If i put this after that its pointless & the errors would show

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
```diff
@@ -25,7 +25,7 @@ function dosomething_user_make_facebook_user($account_details) {
 /**
  * Registers a new user with Northstar (or upserts with existing) & returns a Drupal account
  * @param  array $account_details Account details for the user to register
- * @return $user Created (or retrieved) Drupal user
+ * @return $user
```
DFurnes: This is still wrong. It can be either `@return {type}` or `@return {type} {description}`. Saying it returns a variable name doesn't really provide any value. Check out the link from my previous comment if you're still confused.

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
DFurnes: Right, you can inline the `$user->{$drupal_key}` part in the `isset` so it gets caught alongside the other potentially nonexistent values:

``` php
      if (isset($user->{$drupal_key}[LANGUAGE_NONE][0]['value'])) {
        $northstar_user[$ns_key] = $user->{$drupal_key}[LANGUAGE_NONE][0]['value'];
     }
   }
```

itsjoekent: Ohhh :facepalm:

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
angaither: let's keep it the same as other parts of the code base and send in `''`

itsjoekent: okie dokes

itsjoekent:
@angaither Is this good to merge?

lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
```diff
@@ -18,7 +18,8 @@ class ApiClient {
     this.host = host;
     this.port = port === '' ? null : port;
     this.version = `/api/${version}/`;
-    this.url = `//${this.host}${this.port ? ':' + this.port : ''}${this.version}`;
+    this.domain = `//${this.host}${this.port ? ':' + this.port : ''}`;
```
angaither, line 21: do you need this line?

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
```diff
@@ -0,0 +1,122 @@
...
+    $token_verified = dosomething_user_verify_facebook_user($facebook_id, $access_token);
+  }
+
+  // If they passed the login verification, create a session
+  if ($token_verified) {
+    dosomething_user_create_facebook_session($user);
+  }
+
+  // Always return them to the previous page
+  drupal_goto($_SERVER['HTTP_REFERER']);
```
angaither, line 133: after login we normally send the user to the profile page

angaither, line 133: not sure if we want to keep/remove/change that here, but that's what currently happens

itsjoekent: Just talked with Luke, he said lets keep this functionality

Tho I'd rather not block this PR with this, as I'm about to endeavor into doing the NS <--> Phoenix verification. Ill make an issue if you're okie with that!

lib/themes/dosomething/paraneue_dosomething/js/utilities/ApiClient.js
itsjoekent: Yes, its used in the social-auth.js

angaither: oh it has `//`?

itsjoekent: Yeah thats a trick Diego taught me, lets you skip defining http or https (Which caused problems while developing onboarding locally i think)

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -158,6 +158,26 @@ function dosomething_northstar_get_northstar_user($drupal_id) {
...
+  $response = drupal_http_request($client['base_url'] . '/users/' . $index_type . '/' . $index, [
+    'headers' => $client['headers'],
+    'method' => 'GET',
+  ]);
+
+  if ($response->code === "200") {
+    return json_decode($response->data);
+  }
+
+  return null;
```
angaither: `NULL` not `null` in droops

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -174,7 +174,7 @@ function dosomething_northstar_get_user_by_custom_index($index_type, $index) {
     return json_decode($response->data);
   }
 
-  return null;
+  return NULL;
```
angaither: this wasn't the only case of this in the php code

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -158,6 +158,26 @@ function dosomething_northstar_get_northstar_user($drupal_id) {
 }
 
 /**
+ * Get a Northstar user for a given index & type. Useful for non Drupal ID lookups.
+ * @param  string $index_type The type of index to query for (Email, Facebook, Mobile, _id)
+ * @param  string $index The index to query for
+ * @return object|null
+ */
+function dosomething_northstar_get_user_by_custom_index($index_type, $index) {
```
angaither: ah, sorry I didn't catch this before, but we already have a `dosomething_northstar_get_northstar_user` which grabs by `drupal_id` but I'd love to take the approach of having _less_ methods and refactor that one a bit so it can take a `$type` and a `$key` (which is I think what they are called in NS) instead of creating a new method to do essentially the same thing just for a different type of `get`

itsjoekent: Yeah I didn't think it would be a problem, I'll refactor the other one to use this

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
```diff
@@ -0,0 +1,122 @@
...
+
+  return $user;
+}
+
+/**
+ * Registers a new user with Northstar (or upserts with existing) & returns a Drupal account
+ * @param  array $account_details Account details for the user to register
+ * @return user
+ */
+function dosomething_user_register_facebook_user($account_details) {
```
angaither, line 30: so if I understand this function correctly, the logic could create a duplicate drupal user, of X email if there's no matching NS user. shouldn't this also check if a drupal user of X email exists before creating one?

itsjoekent: Why wouldn't there be a matching NS user?

itsjoekent: I can add a check sure, but isn't that reinforcing the idea of Drupal being the gate keeper of all things users? (and eventually *tm this is just gonna be a replica of northstar anyway)

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
```diff
@@ -0,0 +1,122 @@
...
+}
+
+/**
+ * Creates a Drupal session for a given Facebook user
+ * @param $user User to create session for
+ */
+function dosomething_user_create_facebook_session($user) {
+  $form_state['uid'] = $user->uid;
+  user_login_submit([], $form_state);
+  user_login_finalize($form_state);
```
angaither, line 82: do all of the hooks to send emails and create new drupal users run if you manually invoke these methods?

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
DFurnes: > shouldn't this also check if a drupal user of X email exists before creating one?

Yeah, I think that's a good idea. I think there shouldn't be too many discrepancies after running those user data cleanup scripts last week, but we do still have some users (with wonky emails, etc.) that only exist in Drupal.

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
itsjoekent: not sure, do you know off the top of your head which function this is done? if not ill go searching for it

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
```diff
@@ -0,0 +1,122 @@
...
+    $token_verified = TRUE;
+  }
+
+  // We now have Northstar verify the given oAuth token is actually for this person
+  if ($verify_token) {
+    $token_verified = dosomething_user_verify_facebook_user($facebook_id, $access_token);
+  }
+
+  // If they passed the login verification, create a session
+  if ($token_verified) {
```
angaither, line 128: what is going on here with the `$verify_token` and `$token_verified` variables?
they are manually set, in [106](https://github.com/DoSomething/phoenix/pull/7023/files#diff-c9c442ab4dc3ef53ab18da58542f4ef0R106)-[107](https://github.com/DoSomething/phoenix/pull/7023/files#diff-c9c442ab4dc3ef53ab18da58542f4ef0R107) and then [tested?](https://github.com/DoSomething/phoenix/pull/7023/files#diff-c9c442ab4dc3ef53ab18da58542f4ef0R111)

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
angaither: that's fine, though I know we have some issues with `HTTP_REFERER` on prod, because of fastly reasons, I'm not sure it's the most reliable method...

itsjoekent: wont be live on prod anytime soon, just needs to work on thor/staging for now

theres another issue in the milestone (configure app id's i think) that also needs to be done before this would get turned on prod, along with the design of the button itself

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
angaither: this is for a [new user](https://github.com/DoSomething/phoenix/blob/aac08a551d112eba8b21e5ef896d5044b99ad2f5/lib/modules/dosomething/dosomething_user/dosomething_user.module#L592)

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
itsjoekent: They are manually set in 106/07 to make sure new users aren't authenticated twice. See the comment above!

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
itsjoekent: @DFurnes noticed this has been resulting in errors 
screen shot 2016-09-09 at 10 34 53 am

Mind if I go back to the way I had it? I know it was an extra conditional but it prevented the warnings from popping up

DFurnes: Using brackets to make the variable-variable not ambiguous fixes that:

``` php
if (isset($user->{$drupal_key}[LANGUAGE_NONE][0]['value'])) { 
   // ...
}
```

itsjoekent: Oh interesting

itsjoekent:
Need to run through all of the tests again, and rebase apparently, but I think I addressed all of the comments

itsjoekent:
🆗  all my testing passed. In addition to whats in the PR description, I also added another step which is signing up for a campaign before running through everything. This way we can check if the right Drupal account is being loaded

itsjoekent:
also made sure that the get northstar user function still worked in other places on the site where its being used

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
itsjoekent: (note the comment is still showing in github but the refactoring did happen, see above)

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
itsjoekent: (also note here, comment is showing but the work was completed, few lines down)

lib/modules/dosomething/dosomething_user/dosomething_user_social_auth.inc
itsjoekent: fixed with the new function i made `dosomething_user_new_user_tasks`

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
angaither: doesn't look like the refactoring happened? you still created `dosomething_northstar_get_user_by_custom_index` which is essentially the same as `dosomething_northstar_get_northstar_user`

itsjoekent: Oh you wanted to merge the two, 

> the approach of having less methods and refactor that one a bit 

I must have forgotten this bit while doing everything else. Should be an easy fix, ill just make the index = drupal_id by default so everything in theory continues to work

itsjoekent: hm on second thought, this might be bigger lift then that since a good chunk of the function uses drupal id, specifically the cache / logging.

itsjoekent: actually one sec, i think i can make it work

itsjoekent: @angaither didnt test locally, but what I did was

```
function dosomething_northstar_get_northstar_user($id, $type = 'drupal_id') {
  $northstar_user_info = NULL;

  if ($type === 'drupal_id') {
    $northstar_user_info = cache_get('northstar_user_info_' . $id, 'cache_dosomething_northstar');
  }

 ...
```

and then for setting the cache 

```
cache_set('northstar_user_info_' . $northstar_user_info->drupal_id, $northstar_user_info, 
```

gonna try it out now

itsjoekent: I think this is what you were getting at?

itsjoekent: @angaither great, merged the two function & tested to make sure its still caching alright. Going to run through a bit more testing but let me know if this looks good!

(might need to refresh page)

weerd: @deadlybutter please test ALL api endpoints that use `dosomething_northstar_get_northstar_user()`... Kudos, Reportbacks, ReportbackItems...

angaither:
there's a lot of changes to NS/new user signup here
so we will want to **fully** test this on staging/thor to make sure users are still getting added to MB and MC and NS. 

also, here is the alot monster.

itsjoekent:
hmf, so i have two options

1) change every single place dosomething_northstar_get_northstar_user is called (like the places you mentioned @weerd ), to accept an actual object instead of a JSON String
2) make two variables in dosomething_northstar_get_northstar_user, one that is a JSON string & one that isnt

why is this an issue? `$northstar_user_info = $response->data;` is a JSON string, yet the code below it mistakenly tries to access it like an object (Im not sure how this has been working...)

example: 

```
if (!empty($northstar_user_info->error)) {
```

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -125,33 +125,36 @@ function dosomething_northstar_update_user($user, $password = null) {
...
  *
- * @param int $drupal_id Drupal user id.
+ * @param int $id Drupal user id.
+ * @param string $type The index type of the ID
  * @return object
  */
-function dosomething_northstar_get_northstar_user($drupal_id) {
-  $northstar_user_info = cache_get('northstar_user_info_' . $drupal_id, 'cache_dosomething_northstar');
+function dosomething_northstar_get_northstar_user($id, $type = 'drupal_id') {
+  $northstar_user_info = NULL;
```
weerd, line 135: You might not need this at all...

if `cache_get()` doesn't find what you ask for, it'll return `NULL`. Which means even if an email gets passed into the string id you generate, it's not going to find it, and thus return `NULL`.

itsjoekent: yeah i always forget PHP & JS scoping are different (which is why I normally why I define variables at the top). ill nix it!

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -125,33 +125,36 @@ function dosomething_northstar_update_user($user, $password = null) {
...
-        cache_set('northstar_user_info_' . $drupal_id, $northstar_user_info, 'cache_dosomething_northstar', REQUEST_TIME + 60*60*24);
-      }
+    ]);
+    $northstar_user_info = $response->data;
+
+    if (!empty($northstar_user_info->error)) {
+      $error = sprintf("Error fetching Northstar user data, uid=%d: '%s'", $id, $northstar_user_info->error);
+      watchdog_exception('northstar', new Exception($error));
+    } else {
+      cache_set('northstar_user_info_' . json_decode($northstar_user_info)->data->drupal_id, $northstar_user_info, 'cache_dosomething_northstar', REQUEST_TIME + 60*60*24);
```
weerd, line 156: we could also maybe set 2 caches... `northstar_user_info_` with drupal id, and one with email, so eventually if this function is called again `cache_get()` with an email id does get found and returns info.

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
weerd, line 135: In general this all just feels a little convoluted...

itsjoekent: actually no, i think it does need to be here. `cache_get` only runs if that conditional is equal to drupal id. so if i fetch by facebook id, and then do `if $northstar_info)` that seems funky?

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
itsjoekent: we have a lot of ID's (4 now?) and if we do more social stuff it'll grow, I think we're better off finding a more dynamic solution later on

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
weerd: What's a facebook ID look like?

itsjoekent: This is mine! `10207186494360722`

PR #7033, opened 2016-09-12
align the join others text

### LINKED ISSUE #7011: Misaligned text in PROVE IT Section
#### Current Behavior

"Join others making a difference" is misaligned
#### Desired Behavior

Alignment!
#### Steps to Replicate
1. Go to a campaign
2. Scroll to Prove it
3. See the misalignment
#### Why This Matters

So that it's aligned and pretty!

screen shot 2016-09-06 at 4 34 57 pm

cc @lkpttn

katiecrane:
#### What's this PR do?

Removes the `container__block` div that was around the misaligned text so that now it is aligned.
1. Go to a campaign
2. Scroll to Prove it
3. See the alignment

image
#### Relevant tickets

Fixes #7011
#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
@@ -343,28 +343,35 @@
       </div>
 
       <div class="container__row">
-          <div class="container__block">
-            <h3 class="inline-alt-text-color"><?php print t('Pics or It Didn&rsquo;t Happen'); ?></h3>
-            <?php if (isset($reportback_copy)): ?>
-              <p class="copy inline-alt-text-color"><?php print $reportback_copy; ?></p>
-            <?php endif; ?>
+        <div class="container__block">
+          <h3 class="inline-alt-text-color"><?php print t('Pics or It Didn&rsquo;t Happen'); ?></h3>
+          <?php if (isset($reportback_copy)): ?>
+            <p class="copy inline-alt-text-color"><?php print $reportback_copy; ?></p>
+          <?php endif; ?>
 
-            <?php if (isset($helpful_tips)): ?>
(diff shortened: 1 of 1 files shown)


weerd:
Hmmmm, the `container__block` is kind of a required item in Forge...

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -357,9 +357,7 @@
             <?php endif; ?>
 
             <div class="container__row">
```
DFurnes: There shouldn't be a `container__row` nested inside a `container__block` (see line 346 above). I think that's what's messing up alignment here.

weerd:
Might be better if you change it to an `h3` tag, but with the `-beta` class so it looks the size of an `h2`...

weerd:
Oye! Yeah @DFurnes is right. That `container__row` should be a sibling of the other one and not inside it.

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -343,31 +343,36 @@
...
+        </div>
       </div>
 
+      <?php if(dosomething_reportback_exists($campaign->nid)): ?>
+        <div class="container__row">
+          <div class="container__block">
+            <a href="#" data-modal-href="#modal-missing-photos">Is your photo not showing up?</a>
+          </div>
+        </div>
+      <?php endif; ?>
```
weerd: I don't think you need a separate `container__row` for the "Is your photo not showing up?" content. It relates to the reportback gallery below which the "Join others making a difference" is kind of a heading for. You should be able to put this inside the `container__row` above.

weerd:
¡OLÉ! :sheep: it!

Issue #6972, opened 2016-08-30
Update reportback item status in rogue


angaither:
#### User Story

As an admin when I review reportbacks, the updated `status` should be reflected in rogue's tables
#### Additional Information (optional)

Send requests on review to update the status in rogue

angaither:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;NkTxuhoU&#x2F;10-as-a-developer-i-want-reportback-status-updates-to-be-sent-to-rogue"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> As a developer, I want reportback status updates to be sent to rogue.</a>

Issue #6960, opened 2016-08-29
Send reportback to rogue


ngjo:
#### User Story

On submission of reportback form, send it to rogue. Deal with good/failures

sbsmith86:
@angaither I was talking to @chloealee yesterday and I think there are some things we need to consider here. 

I was thinking through the flow of events and in the case of reviewing a reportback. We need to be able to specify which reportback item needs to be updated but phoenix doesn't have any idea about rogue identifiers. Because we are going to be storing the reportbacks in two different places drupal will still reference a reportback by it's own `rbid` and the item's`fid` but those won't correlate with rogue ids.

So two options: 

1) we store rogue ids and which `rbids` they correspond to in drupal. So the flow would be: 
-> user submits reportback form or review form
-> phoenix sends reportback to rogue 
-> rogue stores or updates reportback 
-> rogue sends response back to phoenix with rogue reportback `id` and the item `id`
-> **Drupal stores those ids in a new table that has the drupal `rbid` and the rogue ids.** (we might need two separate tables for reportbacks and items)
-> anytime drupal makes a request to rogue, it looks up the corresponding `rogue_reportback_id` to the `rbid` that it knows about and sends the request with the `rogue_reportback_id` 

2) Flip it so the flow would be: user submits reportback form 
-> user submits reportback form or review form
-> phoenix sends reportback to rogue 
-> rogue stores or updates reportback 
-> **phoenix sends response back with the `rbid` and `fid` it stored in phoenix** 
-> rogue stores that in it's own tables
-> anytime drupal makes a request to rogue, it sends the drupal version of the ids over and rogue looks them up that way.

Both of these options would take a little bit more work and probably the same amount of time, but I am partial to the first option. One, I don't think we want to spend too much time on having rogue be aware of anything drupal is doing since if we really separate this out as a service it shouldn't matter what an `rbid` is. Also, if we did #2, we would have to update the response that drupal sends back when you POST to the reportback endpoint, because it only returns `rbid` right now and we would also need `fid`. Not sure if doing that would mess with anything else using the endpoint at the moment. 

Sorry for the essay, but wanted to document this here in case you had thoughts.

angaither:
@sbsmith86 hm
1. this idea makes sense, however I wonder if we could do secret idea 3 


3. enable rogue across campaigns and set the index in rogue/phoenix to something to match so that `id` in rogue == `rbid` in phoenix? 
that was our original goal I think?

[...]

PR #7044, opened 2016-09-14
Drupal Rogue client

sbsmith86:
#### What's this PR do?

Adds functions to the new `dosomething_rogue` module that connects to the rogue api and sets up the wrapper for the guzzle client that makes the actual request.

You should be able to call `ddosomething_rogue_send_reportback_to_rogue()` and receive a response from rogue. 

Unblocks work on sending reportbacks and updates to reportbacks.
#### Relevant tickets

Ref #6960 #6972

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
@@ -6,7 +6,7 @@
 
 include_once('dosomething_rogue.admin.inc');
 
-define('ROGUE_API_URL', variable_get('dosomething_rogue_url', 'http://rogue.app/'));
+define('ROGUE_API_URL', variable_get('dosomething_rogue_url', 'http://rogue.app/api'));
 define('ROGUE_API_VERSION', variable_get('dosomething_rogue_api_version', 'v1'));
 define('ROGUE_API_KEY', variable_get('dosomething_rogue_api_key', ''));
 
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -29,3 +29,58 @@ function dosomething_rogue_menu() {
...
+ * @param array $values
+ *   Values to send to Rogue.
+ * @todo - Do we need to pass in user here? Do we need to pass in the item separately?
+ */
+function dosomething_rogue_send_reportback_to_rogue($values, $user = NULL) {
+  if (!isset($user)) {
+    global $user;
+  }
+
+  $campaign = node_load($values['nid']);
```
angaither: do you need to load the node here? looks like you just use the `nid` below

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -29,3 +29,58 @@ function dosomething_rogue_menu() {
...
+          'quantity' => $values['quantity'],
+          'why_participated' => $values['why_participated'],
+          'file' => $values['file'],
+          'caption' => $values['caption'],
+          // @TODO - review status stuff can be passed here also
+        ]),
+  ];
+
+  $response = drupal_http_request($client['base_url'] . '/reportbacks', $options);
+
```
angaither, line 83: maybe add a todo here to add some response/error handling

angaither (changes requested):

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
sbsmith86: yea my bad, that was a relic of some previous version of this code.

angaither (approved):

angaither:
🚢

PR #7045, opened 2016-09-14
Prototype  automagic signup  redirects

### LINKED ISSUE #7027: Adding redirect option to take the user to the campaign page
#### User Story

As a user, when I click the sign-up magick link, I want to be taken to the appropriate campaign page.
#### Why This Matters

So the user can see the campaign that they just signed up for.

weerd:
#### What's this PR do?

This PR adds a redirect after the automagic signup action takes place. Depending on whether the action created a new signup or determined the user is already signed up for the specified campaign, the redirect can be handled differently.

Take a gander 👀  and see if the code make sense.

Currently, for a user already signed up for the campaign, the redirect will send them to the campaign pitch page and prompt them to login by popping up the login modal.

For a user who is signing up for the first time, it also redirects them to the campaign pitch page but passes along some specific url parameters so that in the next step we can do some automagicks with the info and override the conditionals for showing just the pitch page to an anon user and allowing them to see the action page instead.

Sample redirection uri:

```
http://dev.dosomething.org:8888/us/node/1280?magicks=1&user_id=1700013&campaign_id=1280
```
#### Relevant tickets

Fixes #7027
#### Checklist
- [ ] Tested on staging.

---

@angaither @DFurnes 

cc: @ngjo

### DIFF
--- a/lib/modules/dosomething/dosomething_api/dosomething_api.info
+++ b/lib/modules/dosomething/dosomething_api/dosomething_api.info
@@ -2,6 +2,7 @@ name = DoSomething API
 core = 7.x
 package = DoSomething
 dependencies[] = ctools
+dependencies[] = dosomething_global
 dependencies[] = dosomething_image
 dependencies[] = services
 dependencies[] = taxonomy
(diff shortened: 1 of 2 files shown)


weerd:
Chatted with @angaither about moving the next steps for overriding the pitch page into a different issue & PR. Not sure if we should merge this one which is why I renamed the branch w/ the prototype prefix. It shouldn't cause any issues if we merge it in though. The PR which overrides the pitch page stuff would be more of a concern to keep in a separate prototype branch to test on Thor. This only really touches the `magicks_resource` which is about to go into some suspended animation anyway...

DFurnes (approved):

PR #7046, opened 2016-09-15
Experiment: OpenID Connect

DFurnes:
#### What's this PR do?

This PR adds the [OpenID Connect module]() to our installation profile, creates a Northstar OpenID provider, and configures it to pull user information from Northstar on login.

This is still all experimental, and shouldn't affect anything when the feature flag is off… although we do have the `/user/openid` route for _secret_ testing!

References DoSomething/northstar#428.

💃 
#### Relevant tickets

Fixes DoSomething/northstar#423.
Fixes DoSomething/northstar#424.
Fixes DoSomething/northstar#425.
#### Checklist
- [ ] Tested on staging.


DFurnes:
Things we should continue with:
- [x] Add a feature flag for swapping out the login form. (@weerd)
- [x] Add a `/user/openid` route so we can test OpenID login even if toggled off. (@weerd)
- [ ] ~~Troubleshoot email conflict when authorizing as an existing Drupal account.~~ Migrate the custom `field_northstar_id` field to Drupal's `authmap` table so this module can find it.
- [ ] Add a registration page in Northstar. (@dfurnes)
- [ ] Persist the access token & refresh token in Phoenix between requests. (@dfurnes)
- [ ] Destroy the Northstar session when the user logs out of Phoenix.

DFurnes:
So for the "email conflict when authorizing as an existing Drupal account" task above, it looks like it checks for existing Drupal users by comparing the `sub` claim (Northstar ID) against Drupal's `authmap` table – this makes total sense, but isn't how we're currently storing that value (it's in a custom field).

We'll need to make a little script to copy those values over, and then potentially drop the custom field.

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
```diff
@@ -0,0 +1,43 @@
...
+
+include_once('plugins/openid_connect_client/northstar/OpenIDConnectClientNorthstar.class.php');
+
+/**
+ * Register our custom OpenID Connect provider.
+ *
+ * Implements hook_ctools_plugin_directory().
+ */
+function dosomething_northstar_ctools_plugin_directory($module, $plugin) {
+  if ($module == 'openid_connect') {
```
weerd: `===` triple equals?

weerd:
¡OLÉ! :sheep: it!

weerd (commented):
A minor comment on a rebellious ==

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
DFurnes: 😱

weerd:
¡OLÉ! :sheep: it!

Issue #7042, opened 2016-09-14
Display Three Campaign Signup Cards


jessleenyc:
_TASK_
Show three campaigns on last slide of onboarding. Once the mockup is ready, I'll add it to this ticket.

jessleenyc:
@deadlybutter - mockups, still waiting on updated copy. https://cloud.githubusercontent.com/assets/1865372/18595537/2acfda54-7c13-11e6-9570-49a9a22567d4.png

itsjoekent:
Also @jessleenyc in the future can we link to the issue Luke was working in rather than the png itself?

jessleenyc:
@deadlybutter absolutely! do you want the png link also or just the link to the issue?

#7039

itsjoekent:
Just the issue is fine! :D

itsjoekent:
Note: Only remaining task here is styling. Markup is on the page!

Issue #7041, opened 2016-09-14
Experiment with new slide type


jessleenyc:
_TASK_
Let's see if it makes sense to create a new slide type that shows three featurecards, instead of changing our current slide type.

itsjoekent:
Looking through the code,

I'd agree a new slide is the simple way to go. The real issue is everything else!
- Need to extract the functionality out of the feature card so multiple card types can share this (Or other things? eg: Galleries)
- In doing that we would also fix the logic problem of fetching more, as this new layer wouldn't care about auto updating or not.

Issue #7040, opened 2016-09-14
Remove dynamic updating of additional campaigns


jessleenyc:
_TASK_
- [x] Update logic so that new campaigns aren't pulled into featurecards when users sign up. These campaigns should be prioritized the same way we're currently filtering them. 
- [x] Make sure check mark remains shown after a user signs up for a campaign

Issue #7039, opened 2016-09-14
Mockup: Three Cards 


jessleenyc:
_TASK_
Create mockups to show three campaigns that users can sign up for and how copy should be placed. Please provide mocks for mobile, tablet, and desktop.

lkpttn:
![campaigncards](https://cloud.githubusercontent.com/assets/1865372/18595423/ab7a60c6-7c12-11e6-9cef-7e4eea1424b4.png)

lkpttn:
^ blue text is an accident!

lkpttn:
<img width="1107" alt="screen shot 2016-09-16 at 1 40 16 pm" src="https://cloud.githubusercontent.com/assets/1865372/18595537/2acfda54-7c13-11e6-9570-49a9a22567d4.png">

PR #7047, opened 2016-09-15
Feature card refactoring

itsjoekent:
#### What's this PR do?

This PR moves a chunk of the feature card functionality into a generic controller class. We're going to need this because I'll be making a new variant of the context slide, which requires grabbing campaigns but with different logic. By making the controller the way I did, its implementation agnostic so it can support both static & dynamic slides (which we will have).

The only key area of concern is making sure the feature card is still properly wired up.
- Does it populate the page on load with campaigns? Sorted correctly?
- Does "view more" work?
- Does signup work?
- Does it show duplicates?

There isn't too much code changes to the Feature Card itself. Mostly code removal & renaming variables to match the controller.
#### Relevant tickets

I don't think this fixes any issue actually, but relevant too / assists / unblocks 
#7041
#7040
#7042

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
@@ -1,0 +1,75 @@
+import setting from '../utilities/Setting';
+import ApiClient from '../utilities/ApiClient';
+import { sortCampaignsByRelevance } from '../campaign/helpers';
+
+class GalleryController {
+  constructor() {
+    this.campaignsToExclude = setting('dsUser.activity.signups', []);
+    this.campaignsApiPage = 1;
+    this.apiClient = new ApiClient('v1');
+    this.retrievedCampaigns = [];
+  }
+
+  /**
+   * Grab campaigns & alter based on given args.
+   * @param {Object} args containing the following optional parameters
+   *   - {boolean} isStaffPick - Defaults to false.
+   *   - {int} page - Defaults to 1.
+   *   - {boolean} removeDuplicateResults
+   *   - {boolean} sortByRelevane
+   *   - {boolean} removeSignups
+   * @return {Promise}
+   */
+  getCampaigns(args) {
+    const query = {
+      staff_pick: args['isStaffPick'] ? +args['isStaffPick'] : 0, // API requires 1/0 instead of true/false
+      page: args['page'] || 1,
+    };
+
+    return new Promise((resolve, reject) => {
+      this.apiClient.get('campaigns', query).then(response => {
+        let campaigns = response.data;
+
+        // Remove any campaigns we've already fetched for this gallery.
+        if (args['removeDuplicateResults']) {
+          campaigns = campaigns.filter(campaign => this.retrievedCampaigns.indexOf(parseInt(campaign.id)) == -1);
+        }
+        // Keep track of the campaigns we've recieved.
+        this.retrievedCampaigns.concat(campaigns.map(campaign => parseInt(campaign.id)));
+
+        // Use our secret sauce to sort them in order.
+        if (args['sortByRelevane']) {
+          campaigns = sortCampaignsByRelevance(campaigns);
+        }
+
+        if (args['removeSignups']) {
+          const signups = setting('dsUser.activity.signups', []);
+          campaigns = campaigns.filter(campaign => signups.indexOf(parseInt(campaign.id)) == -1);
+        }
+
+        resolve(campaigns);
+      });
+    });
+  }
+
+  /**
+   * Sign the current user up for the specified campaign.
+   *
+   * @param {int} campaignId
+   * @param {string} source
+   * @return {Promise}
+   */
+  signup(campaignId, source) {
+    const url = `campaigns/${campaignId}/signup`;
+    const data = {source: source};
+
+    return new Promise((resolve, reject) => {
+      this.apiClient.post(url, data).then(response => {
+        resolve();
+      });
+    });
+  }
+
+}
+
+export {GalleryController as default}
--- a/lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
@@ -72,4 +72,4 @@ class GalleryController {
 
 }
 
-export {GalleryController as default}
+export default GalleryController;
--- a/lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
@@ -50,7 +50,7 @@ class GalleryController {
         campaigns = this.filter(campaigns, signups);
       }
 
-      resolve(campaigns);
+      return campaigns;
     });
   }
 
@@ -66,7 +66,7 @@ class GalleryController {
     const data = {source: source};
 
     return this.apiClient.post(url, data).then(response => {
-      resolve();
+      return true;
     });
   }
 
--- a/lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
@@ -30,29 +30,27 @@ class GalleryController {
       page: args['page'] || 1,
     };
 
-    return new Promise((resolve, reject) => {
-      this.apiClient.get('campaigns', query).then(response => {
-        let campaigns = response.data;
+    return this.apiClient.get('campaigns', query).then(response => {
+      let campaigns = response.data;
 
-        // Remove any campaigns we've already fetched for this gallery.
-        if (args['removeDuplicateResults']) {
-          campaigns = this.filter(campaigns, this.retrievedCampaigns);
-        }
-        // Keep track of the campaigns we've recieved.
-        this.retrievedCampaigns.concat(campaigns.map(campaign => parseInt(campaign.id)));
+      // Remove any campaigns we've already fetched for this gallery.
+      if (args['removeDuplicateResults']) {
+        campaigns = this.filter(campaigns, this.retrievedCampaigns);
+      }
+      // Keep track of the campaigns we've recieved.
+      this.retrievedCampaigns.concat(campaigns.map(campaign => parseInt(campaign.id)));
 
-        // Use our secret sauce to sort them in order.
-        if (args['sortByRelevance']) {
-          campaigns = sortCampaignsByRelevance(campaigns);
-        }
+      // Use our secret sauce to sort them in order.
+      if (args['sortByRelevance']) {
+        campaigns = sortCampaignsByRelevance(campaigns);
+      }
 
-        if (args['removeSignups']) {
-          const signups = setting('dsUser.activity.signups', []);
-          campaigns = this.filter(campaigns, signups);
-        }
+      if (args['removeSignups']) {
+        const signups = setting('dsUser.activity.signups', []);
+        campaigns = this.filter(campaigns, signups);
+      }
 
-        resolve(campaigns);
-      });
+      resolve(campaigns);
     });
   }
 
@@ -67,10 +65,8 @@ class GalleryController {
     const url = `campaigns/${campaignId}/signup`;
     const data = {source: source};
 
-    return new Promise((resolve, reject) => {
-      this.apiClient.post(url, data).then(response => {
-        resolve();
-      });
+    return this.apiClient.post(url, data).then(response => {
+      resolve();
     });
   }
 
--- a/lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
@@ -20,7 +20,7 @@ class GalleryController {
    *   - {boolean} isStaffPick - Defaults to false.
    *   - {int} page - Defaults to 1.
    *   - {boolean} removeDuplicateResults
-   *   - {boolean} sortByRelevane
+   *   - {boolean} sortByRelevance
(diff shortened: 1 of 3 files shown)


lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
```diff
@@ -0,0 +1,79 @@
...
+   *   - {boolean} isStaffPick - Defaults to false.
+   *   - {int} page - Defaults to 1.
+   *   - {boolean} removeDuplicateResults
+   *   - {boolean} sortByRelevane
+   *   - {boolean} removeSignups
+   * @return {Promise}
+   */
+  getCampaigns(args) {
+    const query = {
+      staff_pick: args['isStaffPick'] ? +args['isStaffPick'] : 0, // API requires 1/0 instead of true/false
```
weerd: I don't believe this is true. You should be able to pass `true` or `false` instead of `1` or `0`.

```
http://staging.beta.dosomething.org/api/v1/campaigns?staff_pick=true
```

works just fine.

weerd: also, what's the `+` in `+args['isStaffPick']`?

itsjoekent: Uhhhh 
screen shot 2016-09-15 at 2 58 44 pm

itsjoekent: converts the boolean to 1/0

lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
```diff
@@ -0,0 +1,79 @@
...
+      page: args['page'] || 1,
+    };
+
+    return new Promise((resolve, reject) => {
```
weerd: Not sure why you're wrapping this in a Promise...

lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
```diff
@@ -0,0 +1,79 @@
...
+   *
+   * @param {int} campaignId
+   * @param {string} source
+   * @return {Promise}
+   */
+  signup(campaignId, source) {
+    const url = `campaigns/${campaignId}/signup`;
+    const data = {source: source};
+
+    return new Promise((resolve, reject) => {
```
weerd: Again, why a Promise? The `apiClient` returns a promise. You just need a `.then()` and a `.catch()` or whatever.

lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
itsjoekent: and here it is with 0 
screen shot 2016-09-15 at 3 01 58 pm

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -15,67 +13,42 @@ class FeatureCard extends React.Component {
...
-
     this.state = {
       actionDisabled: false,
       campaignIndex: -1,
       isSignedUp: false,
       isLoading: false,
     };
 
+    this.campaigns = [];
+    this.controller = new GalleryController();
```
weerd: So wait, the implication w/ `GalleryController` is for helping to manage a gallery of feature cards? Your're doing this for the scenario when we have like 3 feature cards presented on one slide right?

lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
```diff
@@ -0,0 +1,79 @@
...
+    return campaigns.filter(campaign => comparison.indexOf(parseInt(campaign.id)) == -1);
+  }
+
+  /**
+   * Grab campaigns & alter based on given args.
+   * @param {Object} args containing the following optional parameters
+   *   - {boolean} isStaffPick - Defaults to false.
+   *   - {int} page - Defaults to 1.
+   *   - {boolean} removeDuplicateResults
+   *   - {boolean} sortByRelevane
```
weerd: typo "sortByRelevance"

lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
```diff
@@ -0,0 +1,79 @@
...
+
+        // Remove any campaigns we've already fetched for this gallery.
+        if (args['removeDuplicateResults']) {
+          campaigns = this.filter(campaigns, this.retrievedCampaigns);
+        }
+        // Keep track of the campaigns we've recieved.
+        this.retrievedCampaigns.concat(campaigns.map(campaign => parseInt(campaign.id)));
+
+        // Use our secret sauce to sort them in order.
+        if (args['sortByRelevane']) {
```
weerd: typo "sortByRelevance"

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
```diff
@@ -15,67 +13,42 @@ class FeatureCard extends React.Component {
...
-      // Get campaigns ordered by relevance & handle exclusions.
-      const campaigns = sortCampaignsByRelevance(result.data).filter(campaign => this.campaignsToExclude.indexOf(parseInt(campaign.id)) == -1);
-
-      // Add it to the master list
+  getCampaigns(staffPick, page, onRecieve) {
+    this.controller.getCampaigns({
+      isStaffPick: staffPick,
+      page: page,
+      removeDuplicateResults: true,
+      sortByRelevane: true,
```
weerd: typo "sortByRelevance"

weerd (changes requested):
It's coming along but I have a few questions.

lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
itsjoekent: Oh thats a good point, I didn't think about that

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
itsjoekent: GalleryController is used by components to fetch recommended campaigns & let users signup for them. It doesn't care what the interface is so yep I can use it for both the current slide & the 3 on 1.

lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
weerd: Ok, I think that broke then and it's a bug in the API. Can you add a @TODO comment to eventually change what you pass in the JS to `true` / `false`, once the bug is fixed in the API? I'd like to be consistent w/ the URL params and we usually use `true` & `false` but if it's broken then nothing you can really do at the moment.

itsjoekent: Sure thing!

weerd: I think you can probs just do the following and make it less cryptic:

``` javascript
staff_pick: args['isStaffPick'] ? 1 : 0
```

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
weerd: So I find it weird though that it's the Feature Card that is implementing the `GalleryController` then... particularly for the slide that will have 3 `FeatureCard`s, then each will have an implementation of the `GalleryController`? That seems incorrect. Ideally, something else (the slide?) is instantiating the `GalleryController` (which sounds more like it's a `CampaignsController`), and then if it wants 1 `FeatureCard` or 3 `FeatureCard`s, then it creates as many as needed and passes the info the `FeatureCard` would need to make each of them. Does this make sense?

itsjoekent: This is just a generic controller that the feature card or the slide could use. For the new slide I'm building which has 3 static campaigns, the controller is bound to the slide which instantiates the cards.

lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
DFurnes: Another nice option would be `Number(args['isStaffPick'])`:

screen shot 2016-09-15 at 3 29 37 pm

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
weerd: Gotcha. Still feels like it's gotten to the point where this should just always be a controller on the slide then, and regardless of whether its 1 FeatureCard or 3 or more, is just a value somewhere for it to know how many FeatureCards to create. Also, I think it should be renamed to `CampaignsController`... a gallery just happens to be the format it take in on instance for how it shows.

weerd: It would also help limit what the FeatureCard has to do... all it really cares about is receiving some info about what campaign to display and allowing a passed in signup action (along with a viewmore action when it's a single one) instead of dealing with gathering all the campaigns and sorting, etc.

lib/themes/dosomething/paraneue_dosomething/js/campaign/GalleryController.js
```diff
@@ -0,0 +1,79 @@
...
+    return new Promise((resolve, reject) => {
+      this.apiClient.post(url, data).then(response => {
+        resolve();
+      });
+    });
+  }
+
+}
+
+export {GalleryController as default}
```
DFurnes: I know these are functionally the same, but can you use the same syntax we use elsewhere?

``` js
export default GalleryController;
```

itsjoekent: ah of course

lib/themes/dosomething/paraneue_dosomething/js/components/FeatureCard.js
itsjoekent: @weerd this sounds like a much larger rewrite then, should I be doing that in this sprint?

weerd: I don't think so... you already did the major chunk of pulling most of the methods, etc into the controller.

itsjoekent: I must be confused as to what you're looking for then, the only thing left in the FeatureCard is state handling as "dealing with gathering all the campaigns and sorting, etc." -- was moved to the controller

DFurnes: It does feel strange that each card needs it's own GalleryController, since that means they'll need to make three independent HTTP requests too. I don't want to over-engineer this, but I think it'd make more sense to inject the same "controller" into each feature card so that they can all shared the same API request & data source.

weerd: @deadlybutter it's a controller you're adding to _each_ feature card. As @DFurnes and I indicated, for the slide with 3 FeatureCards, each one is instantiating a gallery controller separately, which doesn't make sense and isn't ideal.

itsjoekent: ahhh, i guess im not being very clear. Each card _does not_ need its own controller. In this case, the feature card has its own controller so it can maintain its own state. But for the new slide we're building which has 2/4 static "cards" which get inserted once, the slide would own the controller.

itsjoekent: @DFurnes @weerd Wait, what slide are you guys referring too with multiple feature cards?

DFurnes: Ideally, this campaign class is a "store" that's responsible for loading in data from the API as necessary, and then "popping" out individual campaigns to the view so they can be displayed by one of the feature cards. We're not really seeing a lot of the benefit of extracting this if they all still naively fetch & maintain their own independent sets of data.

DFurnes: > @DFurnes @weerd Wait, what slide are you guys referring too with multiple feature cards?

Maybe I'm misunderstanding the context of this pull request then. Isn't this in preparation for [including three campaigns](https://github.com/DoSomething/phoenix/issues/7039) on the "signup" slide?

itsjoekent: @DFurnes Correct! Just making sure we're on the same page. If we went the "store" route, are you saying that would dispatch campaigns to some event handler essentially? And the store would also handle reloading when its out of campaigns

itsjoekent: NEVERMIND TO ALL OF THIS. I misunderstood what the new Feature card / Slide was...

itsjoekent:
From Slack

> @joe lets start with this. Can you create a new PR that just adds what you did so far for just the GalleryController and rename to `Controller` inside `/js/campaign/`? Along with that add a new slide called something that indicates it’s one that has a “gallery of campaigns` and start from there.
> 
> @diego just to double check, this new PR should not touch the feature card
> 
> Yeah, lets do that. This PR is just to add a new `Controller` class in campaigns that will be used in upcoming PR for helping with FeatureCards on different types of slides. But lets get this step in there and it’ll be fine to merge cause it’s standalone new class that won’t affect anything for the time being. I feel it’ll help think through this in an easier way :slightly_smiling_face: that cool?

Issue #7049, opened 2016-09-15
API staff_pick URL param bug when passing string boolean


weerd:
#### Current Behavior
- The API should accept passing `true` or `false` for the `staff_pick` url parameter, but it does not seem to work, and only works if you pass an integer of `0` or `1`.
#### Desired Behavior
- The API should accept passing `true` or `false` for the `staff_pick` url parameter.
#### Relevant Screenshots + Links

Where it was discovered: https://github.com/DoSomething/phoenix/pull/7047#pullrequestreview-231573

PR #7050, opened 2016-09-15
Adds campaigns controller 

itsjoekent:
#### What's this PR do?

Adds the campaigns controller standalone

Call the functions!

See the previous PR https://github.com/DoSomething/phoenix/pull/7047#issuecomment-247443454
#### Relevant tickets

I don't think this fixes any issue actually, but relevant too / assists / unblocks 
#7041
#7040
#7042

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/campaign/Controller.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/campaign/Controller.js
@@ -1,0 +1,75 @@
+import setting from '../utilities/Setting';
+import ApiClient from '../utilities/ApiClient';
+import { sortCampaignsByRelevance } from '../campaign/helpers';
+
+class CampaignController {
+  constructor() {
+    this.campaignsToExclude = setting('dsUser.activity.signups', []);
+    this.campaignsApiPage = 1;
+    this.apiClient = new ApiClient('v1');
+    this.retrievedCampaigns = [];
+  }
+
+  filter(campaigns, comparison) {
+    return campaigns.filter(campaign => comparison.indexOf(parseInt(campaign.id)) == -1);
+  }
+
+  /**
+   * Grab campaigns & alter based on given args.
+   * @param {Object} args containing the following optional parameters
+   *   - {boolean} isStaffPick - Defaults to false.
+   *   - {int} page - Defaults to 1.
+   *   - {boolean} removeDuplicateResults
+   *   - {boolean} sortByRelevance
+   *   - {boolean} removeSignups
+   * @return {Promise}
+   */
+  getCampaigns(args) {
+    const query = {
+      staff_pick: args['isStaffPick'] ? +args['isStaffPick'] : 0, // API requires 1/0 instead of true/false
+      page: args['page'] || 1,
+    };
+
+    return this.apiClient.get('campaigns', query).then(response => {
+      let campaigns = response.data;
+
+      // Remove any campaigns we've already fetched for this gallery.
+      if (args['removeDuplicateResults']) {
+        campaigns = this.filter(campaigns, this.retrievedCampaigns);
+      }
+      // Keep track of the campaigns we've recieved.
+      this.retrievedCampaigns.concat(campaigns.map(campaign => parseInt(campaign.id)));
+
+      // Use our secret sauce to sort them in order.
+      if (args['sortByRelevance']) {
+        campaigns = sortCampaignsByRelevance(campaigns);
+      }
+
+      if (args['removeSignups']) {
+        const signups = setting('dsUser.activity.signups', []);
+        campaigns = this.filter(campaigns, signups);
+      }
+
+      return campaigns;
+    });
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


lib/themes/dosomething/paraneue_dosomething/js/campaign/Controller.js
```diff
@@ -0,0 +1,75 @@
...
+   *   - {boolean} isStaffPick - Defaults to false.
+   *   - {int} page - Defaults to 1.
+   *   - {boolean} removeDuplicateResults
+   *   - {boolean} sortByRelevance
+   *   - {boolean} removeSignups
+   * @return {Promise}
+   */
+  getCampaigns(args) {
+    const query = {
+      staff_pick: args['isStaffPick'] ? +args['isStaffPick'] : 0, // API requires 1/0 instead of true/false
```
weerd: Use the approach @DFurnes [mentioned](https://github.com/DoSomething/phoenix/pull/7047#r79043289)`Number(args['isStaffPick'])`

and add a `@todo` referencing this [bug](https://github.com/DoSomething/phoenix/issues/7049) 😄

lib/themes/dosomething/paraneue_dosomething/js/campaign/Controller.js
```diff
@@ -0,0 +1,75 @@
...
+      page: args['page'] || 1,
+    };
+
+    return this.apiClient.get('campaigns', query).then(response => {
```
weerd, line 33: Maybe you don't need this `return` if also using a `return` inside the anonymous function dealing with the response?

lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSlide.js
```diff
@@ -0,0 +1,19 @@
...
+  constructor(props) {
+    super(props);
+
+    this.controller = new CampaignController();
+  }
+
+  render() {
+
+  }
+}
```
weerd, line 19: 💃 yay!

weerd (changes requested):
Just a couple comments on some small things :)

lib/themes/dosomething/paraneue_dosomething/js/campaign/Controller.js
itsjoekent: Still getting used to promises, this worked!

itsjoekent:
@weerd Addressed your review!

lib/themes/dosomething/paraneue_dosomething/js/campaign/Controller.js
DFurnes: Wait... sorry to roll in after the fact, but this definitely should have a `return` statement. I'm not even sure how it works without it (the outer `return` returns the Promise from `getCampaigns`, and the inner `return` returns the value to the next step in the Promise chain).

weerd: No worries, so am I :)

weerd:
¡OLÉ! :sheep: it!

weerd (approved):

lib/themes/dosomething/paraneue_dosomething/js/campaign/Controller.js
itsjoekent: 👀  well this is odd then

DFurnes: How did you test this to verify that removing the `return` still worked? As far as I can tell, block functions don't have implicit returns (only bracketless one-liners do that).

itsjoekent: @DFurnes apparently when i called the function it was still the old file, you're def right :facepalm:

itsjoekent: i often get this, where i refresh quickly enough the build didnt finish, so i still have the old file

weerd: count to 10? 😉 :trollface:

itsjoekent: 👊

weerd: I deserve that. haha

weerd: Yeah, I forgot we're in the `getCampaigns()` method which should return the data. Alternatively you could put all the logic inside that anonymous function inside a separate method like `processCampaigns()` or something, and then just call that from within the anonymous function? Not a huge deal though.

itsjoekent:
Looks good to me! 
screen shot 2016-09-16 at 10 12 12 am

weerd:
Looking good 👍

DFurnes (approved):

PR #7051, opened 2016-09-16
Adding custom route and template file for experimenting with openid login

weerd:
#### What's this PR do?

This PR adds a new `/user/openid` route to Phoenix to allow us to experiment with adding OpenID login that connects to Northstar for our members.

The **dosomething_northstar** module sets up the route since it's related to this functionality. There is a new template file in the `/theme` directory in the module, but it is recommended to use a template file in the activated theme to override it (and says so in the message if the northstar template loads 😉 ). I've provided a template in the **Paraneue** theme to override it and get the ball rolling.

Make sure the **dosomething_northstar** module is enabled, then do a good ole `drush cc all` and head to `/user/openid` and see if you get the correct page loading.
#### Relevant tickets

Refs #7046
#### Checklist
- [ ] Tested on staging.

---

@DFurnes 

cc: @jessleenyc

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -283,8 +283,12 @@ function dosomething_northstar_log_request($op, $user, $request_body, $response)
     ->execute();
 }
 
+/**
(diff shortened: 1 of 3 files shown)


DFurnes (approved):
⛵

PR #7054, opened 2016-09-19
Northstar field authmap

DFurnes:
#### What's this PR do?

This updates the `dosomething_northstar_save_id_field` function to also save the Northstar ID to the `authmap` table so it's available to the OpenID Connect module. It also adds a primary key to the `dosomething_northstar_request_log` table because that's a reasonable thing to do.

👀

🍃 
#### Relevant tickets

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -241,7 +241,7 @@ function dosomething_northstar_transform_user($user, $password = null) {
 }
 
 /**
- * Save the user's Northstar ID to their local profile.
+ * Save the user's Northstar ID to their local profile field and the authmap.
  *
  * @param $uid - Drupal user ID
  * @param $northstar_response - Northstar user JSON response
@@ -250,6 +250,15 @@ function dosomething_northstar_save_id_field($uid, $northstar_response) {
   $northstar_id = !empty($northstar_response->data->id) ? $northstar_response->data->id : null;
   $user = user_load($uid);
 
+  // Save a record to the `authmap` table if Northstar user exists:
+  if (! is_null($northstar_id)) {
+    db_merge('authmap')
+      ->key(['uid' => $user->uid])
+      ->fields(['authname' => $northstar_id, 'module' => 'openid_connect_northstar'])
+      ->execute();
+  }
+
+  // Save the `field_northstar_id`:
   $edit = [];
   dosomething_user_set_fields($edit, ['northstar_id' => $northstar_id]);
   user_save($user, $edit);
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -241,15 +241,24 @@ function dosomething_northstar_transform_user($user, $password = null) {
...
  * @param $northstar_response - Northstar user JSON response
  */
 function dosomething_northstar_save_id_field($uid, $northstar_response) {
-  $northstar_id = !empty($northstar_response->data->id) ? $northstar_response->data->id : 'NONE';
+  $northstar_id = !empty($northstar_response->data->id) ? $northstar_response->data->id : null;
   $user = user_load($uid);
 
+  // Save a record to the `authmap` table if Northstar user exists:
+  if (! is_null($northstar_id)) {
+    db_merge('authmap')
```
angaither: 💭  can you call [`user_set_authmaps`](https://api.drupal.org/api/drupal/modules%21user%21user.module/function/user_set_authmaps/7.x) instead of doing a `db_merge`

DFurnes: Oh, nice! Updated. 🔨

angaither:
🚢 

the other thing was I _know_ that @sergii-tkachenko just wrote some special function for getting NS ids... using that `NONE` logic. but this shouldn't break that. just speaking out into the void here.

DFurnes:
Yep, it should [handle both cases safely](https://github.com/DoSomething/phoenix/blob/6693b2c5fda3cd84f84f6d17cb91fe7ccec8e840/lib/modules/dosomething/dosomething_user/dosomething_user.module#L1318-L1322) in the accessor function that he added! I did just do a grep and found a theme function that relied on the `NONE` string though, so pushed up a fix for that in 3d19fc5.

weerd:
¡OLÉ! :sheep: it!

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
sergii-tkachenko: @angaither @DFurnes [Here](https://github.com/DoSomething/phoenix/pull/7021/files#diff-ad30a5d569ca2ae6d2e04223aa501104R1320).

sergiitk:
@angaither @DFurnes Regarding my handler of 'NONE' value: it's introduced [here](https://github.com/DoSomething/phoenix/pull/7021/files#diff-ad30a5d569ca2ae6d2e04223aa501104R1320).

DFurnes:
> @angaither @DFurnes Regarding my handler of 'NONE' value: it's introduced [here](https://github.com/DoSomething/phoenix/pull/7021/files#diff-ad30a5d569ca2ae6d2e04223aa501104R1320).

Yep, I linked to that same line in my comment above. 😉

PR #7055, opened 2016-09-19
Add a unique database constraint on email and mobile fields.

### LINKED ISSUE #6409: Fix issue where multiple Drupal users share the same email.

DFurnes:
#### What's this PR do?

This adds unique indexes to the user's email ~~and mobile fields~~, finally making a happy ending for the long and terrifying tale of the [duplicates](https://en.wikipedia.org/wiki/Auton). Phew.

🚧 🚧 🚧 See comment below.

👀

We'll want to do a spot-check to make sure these fields can be safely made unique before deploying (as conflicts can still be made through a race condition). I just checked when making this PR, and there are 3 dupe emails and 3 dupe mobiles on production.
#### Relevant tickets

Fixes #6409. 😌 ‼️ 🔨
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.install
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.install
@@ -58,6 +58,14 @@ function dosomething_user_update_7006() {
 }
 
 /**
+ * Add a unique constraint to the `mail` field on the users table.
+ */
+function dosomething_user_update_7008() {
+  db_drop_index('users', 'mail');
+  db_add_unique_key('users', 'mail', ['mail']);
+}
+
+/**
  * Implements hook_uninstall().
  */
 function dosomething_user_uninstall() {
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.install
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.install
@@ -66,6 +66,14 @@ function dosomething_user_update_7008() {
 }
 
 /**
+ * Add a unique constraint to the `field_mobile_value` field for the
+ * mobile field (since fields don't support unique database constraints).
+ */
+function dosomething_user_update_7009() {
+  db_add_unique_key('field_data_field_mobile', 'field_mobile_value', ['field_mobile_value']);
+}
+
+/**
  * Implements hook_uninstall().
  */
 function dosomething_user_uninstall() {


DFurnes:
A frustrating complication is that it doesn't seem to be easy to add a unique constraint on a Drupal field. Everything I see online points to this [Unique Field module](https://www.drupal.org/project/unique_field) but that just adds validation checks in the `node_validate` hook, not a real database constraint.

Do we think it's safe to add that constraint with an update hook...? I'm worried that reverting features on the field in the future might just unset that change.

/cc @angaither @mshmsh5000 for Drupal smarts

angaither:
looks reasonable. 
I have a sinking feeling that this may take a bit to run on thor with all the real data. 
we can give it a shot in one deploy and see what happens. after all what could go wrong?

mshmsh5000:
I added an index (since removed) to `field_mobile_value` on Thor for some data work. It took in the seconds range, not minutes. Theoretically could take longer in production, since `CREATE INDEX` [waits for current transactions to complete](http://stackoverflow.com/a/36064200), but doubtful we'd notice this in real time.

DFurnes:
That's good to know! 😌

DFurnes:
So these ran super fast but _not_ super successful:

```
SQLSTATE[42000]: Syntax error or access violation: 1071 Specified key[error]
was too long; max key length is 767 bytes
```

According to [this StackOverflow question](http://stackoverflow.com/questions/1814532), you can't set a database index on a `VARCHAR` field with a length greater than 255, and our `mail` field is set to `VARCHAR(320)`. Probably safe to shorten that, since I don't think emails can be that long?

DFurnes:
Surprise! We have a different column length for `mail` on Production/Thor than on Staging/Dev! Production's `users` table has it set to 320, but staging is set to 254.

It looks like `filter_var`'s `FILTER_VALIDATE_EMAIL` [restricts emails to max length of 254](http://stackoverflow.com/a/21631648/811624), so we shouldn't ever need more than that. And the longest value in that column on production is 141 so this should be a safe change to make without any data changes.

/cc @angaither @mshmsh5000

angaither:
Surprise!

I don't remember changing this value, but it seems like somehow that happened?!?!
but if that's the case, seems like a safe change to make!

DFurnes:
Woo! Updating the column length on Thor took a little while but fixed it! 😄

DFurnes:
The mobile table update hook seems to be blocked by [this other failing update hook](https://github.com/DoSomething/phoenix/issues/7062). Not an emergency Blocker™, but something that'd be nice to clean up sometime soon.

DFurnes:
The email index has now been successfully added on both Thor and production. The index on `field_data_field_mobile`'s `field_mobile_value` column is still missing on production, though.

PR #7056, opened 2016-09-19
Fix typo when unsetting the app key variable.

DFurnes:
#### What's this PR do?

When users run `ds pull stage --db`, we remove the Staging server's Northstar credentials so we don't accidentally pollute data when re-enabling that module on local. It turns out I made a typo when adding this `drush variable-delete` call, so it wasn't removing the "app key".

If you pull staging, you should see `secret1` for the "Northstar App Key". This is the default value for that field, and also what Northstar seeds for the `trusted-test-client` in development.

We should update these to be `dosomething_northstar_client_id` and `dosomething_northstar_client_secret` when we start using Northstar-PHP because the current naming is rather confusing.
#### Relevant tickets

Fixes `¯\_(ツ)_/¯`.
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -468,7 +468,7 @@ function pull_db {
       drush variable-delete dosomething_northstar_url -y
       drush variable-delete dosomething_northstar_port -y
       drush variable-delete dosomething_northstar_app_id -y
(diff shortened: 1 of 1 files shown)


angaither:
🚢

PR #7057, opened 2016-09-19
Save NULL rather than empty string when unsetting mobiles.

DFurnes:
#### What's this PR do?

One last gotcha! When a user sets the mobile number on their profile, we run it through `dosomething_user_clean_mobile_number`, which returns false if it's not valid. That ends up storing a empty string though, which would fail our new uniqueness constraint. This adds a quick fix to provide null in that case instead, which unsets the field.

👀

It might make sense to refactor `dosomething_user_clean_mobile_number` to always return `string|null` instead of `string|false` so things work more intuitively?
#### Relevant tickets

Fixes 🔥.
#### Checklist
- [ ] Tested on staging.


angaither:
🚢

PR #7058, opened 2016-09-19
Adds campaign signup card styling

### LINKED ISSUE #7042: Display Three Campaign Signup Cards

itsjoekent:
#### What's this PR do?

Styles the campaign card
screen shot 2016-09-19 at 4 57 33 pm
screen shot 2016-09-19 at 4 57 44 pm

Do the styles look like the mock?
Mobile / desktop?

no
#### Relevant tickets

Fixes #7042 
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSignupCard.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSignupCard.js
@@ -68,27 +68,25 @@ class CampaignSignupCard extends React.Component {
   render() {
     const header = this.renderHeader();
     return (
-      <li>
-        <article className={`beta-card ${header === null ? '-headerless' : ''}`}>
-          <div className="beta-card__body">
+      <article className={`beta-card ${header === null ? '-headerless' : ''}`}>
+        <div className="beta-card__body">
 
-            {header}
+          {header}
 
-            <div className="beta-card__row">
-              <img src={this.props.image_uri} />
-            </div>
-
-            <div className="beta-card__row">
-              <div className="beta-card__block -centered">
-                <h2 className="heading -delta">{this.props.title}</h2>
-                <p className="tagline">{this.props.tagline}</p>
-                {this.renderForm()}
-              </div>
-            </div>
-
+          <div className="beta-card__row">
+            <img src={this.props.image_uri} />
           </div>
-        </article>
-      </li>
+
+          <div className="beta-card__row">
+            <div className="beta-card__block -centered">
+              <h2 className="heading -delta">{this.props.title}</h2>
+              <p className="tagline">{this.props.tagline}</p>
+              {this.renderForm()}
+            </div>
+          </div>
+
+        </div>
+      </article>
     )
   }
 }
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSignupCard.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSignupCard.js
@@ -66,26 +66,29 @@ class CampaignSignupCard extends React.Component {
   }
 
   render() {
+    const header = this.renderHeader();
     return (
-      <article className="beta-card -signup">
-        <div className="beta-card__body">
+      <li>
+        <article className={`beta-card ${header === null ? '-headerless' : ''}`}>
+          <div className="beta-card__body">
 
-          {this.renderHeader()}
+            {header}
 
-          <div className="beta-card__row">
-            <img src={this.props.image_uri} />
-          </div>
-
-          <div className="beta-card__row">
-            <div className="beta-card__block -centered">
-              <h2 className="heading -delta">{this.props.title}</h2>
-              <p className="tagline">{this.props.tagline}</p>
-              {this.renderForm()}
+            <div className="beta-card__row">
+              <img src={this.props.image_uri} />
             </div>
-          </div>
 
-        </div>
-      </article>
+            <div className="beta-card__row">
+              <div className="beta-card__block -centered">
+                <h2 className="heading -delta">{this.props.title}</h2>
+                <p className="tagline">{this.props.tagline}</p>
+                {this.renderForm()}
+              </div>
+            </div>
+
+          </div>
+        </article>
+      </li>
     )
   }
 }
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSignupCard.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSignupCard.js
@@ -67,7 +67,7 @@ class CampaignSignupCard extends React.Component {
 
   render() {
     return (
-      <article className="beta-card">
+      <article className="beta-card -signup">
         <div className="beta-card__body">
 
           {this.renderHeader()}
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSlide.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSlide.js
@@ -25,6 +25,7 @@ class CampaignSlide extends React.Component {
 
     this.controller.getCampaigns(args).then(campaigns => {
       this.campaigns = campaigns.splice(0, 3); // Grab the first 3 campaigns returned
+      this.forceUpdate();
     });
   }
 
@@ -35,20 +36,22 @@ class CampaignSlide extends React.Component {
           <div className="wrapper">
             {this.renderHeader()}
 
-            {this.campaigns.map(campaign => {
-              // Fix for local files / data missing by preventing null error
-              const image_uri = campaign.cover_image.default != null ? campaign.cover_image.default.sizes.landscape.uri : '';
+            <div className="container__cards">
+              {this.campaigns.map(campaign => {
+                // Fix for local files / data missing by preventing null error
+                const image_uri = campaign.cover_image.default != null ? campaign.cover_image.default.sizes.landscape.uri : '';
 
-              const props = {
-                id: campaign.id,
-                image_uri: image_uri,
-                tagline: campaign.tagline,
-                title: campaign.title,
-                signup: this.controller.signup.bind(this.controller)
-              }
+                const props = {
+                  id: campaign.id,
+                  image_uri: image_uri,
+                  tagline: campaign.tagline,
+                  title: campaign.title,
+                  signup: this.controller.signup.bind(this.controller)
+                }
 
-              return <CampaignSignupCard {...props} key={campaign.id} />
-            })}
+                return <CampaignSignupCard {...props} key={campaign.id} />
+              })}
+            </div>
           </div>
         </div>
       </div>
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSlide.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSlide.js
@@ -11,6 +11,10 @@ class CampaignSlide extends React.Component {
   constructor(props) {
     super(props);
 
+    this.setState({
+      campaigns: []
+    });
+
     this.controller = new CampaignController();
     this.campaigns = [];
   }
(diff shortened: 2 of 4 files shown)


lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSlide.js
```diff
@@ -25,6 +25,7 @@ class CampaignSlide extends React.Component {
 
     this.controller.getCampaigns(args).then(campaigns => {
       this.campaigns = campaigns.splice(0, 3); // Grab the first 3 campaigns returned
+      this.forceUpdate();
```
itsjoekent: Added this because while testing the campaign slide wasn't updating with campaigns if it was the only slide in the array. I think its safe to keep it

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_card.scss
```diff
@@ -42,6 +42,7 @@
 }
 
 
+// Soon to be deprecated in favor of signup-card.
```
itsjoekent: This comment can be removed

itsjoekent: its 6 minutes later & github still shows this outdated comment. @github please fix.

lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSlide.js
DFurnes: You should be storing the campaigns in state if you want React to diff them and re-render.

DFurnes (changes requested):

lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSlide.js
weerd: Can you elaborate on this? I don't think I understand why?

lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSlide.js
```diff
@@ -35,20 +36,22 @@ class CampaignSlide extends React.Component {
           <div className="wrapper">
             {this.renderHeader()}
 
-            {this.campaigns.map(campaign => {
-              // Fix for local files / data missing by preventing null error
-              const image_uri = campaign.cover_image.default != null ? campaign.cover_image.default.sizes.landscape.uri : '';
+            <div className="container__cards">
```
weerd: isn't this just a triad gallery? http://forge.dosomething.org/#gallery-triad

If you use the triad gallery pattern, you probs won't need the `-signup` class to define the widths etc, and it's already set w/ specifics for responsively working on all device sizes.

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_card.scss
```diff
@@ -66,6 +66,17 @@
...
+  &.-signup {
+    width: 30%;
+    min-width: 256px;
+    margin: 1.5%;
+    display: inline-block;
+
+    .beta-card__row:first-of-type {
+      margin-top: 0;
+    }
+  }
```
weerd, line 74: see comment regarding triad gallery

weerd (changes requested):
A couple questions and a recommendation to use the triad gallery.

lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSlide.js
itsjoekent: Switched to triad gallery in this commit: https://github.com/DoSomething/phoenix/pull/7058/commits/ab7261183167a782c8247b35455c04a67265a3c0

Heres what it looks like
screen shot 2016-09-20 at 10 22 53 am
screen shot 2016-09-20 at 10 23 06 am

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_card.scss
itsjoekent: This has nothing to do with the gallery / triad format. When the campaign signup card has no header, it leaves a margin at the top which creates an empty row of white rather than being flush with the border. Heres an example:

screen shot 2016-09-20 at 10 24 26 am

I renamed this class from `-signup` to `-headerless` to better reflect what the style was doing

itsjoekent:
@DFurnes @weerd Changed it so the campaigns array now lives on the slide state

lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSlide.js
```diff
@@ -11,6 +11,10 @@ class CampaignSlide extends React.Component {
   constructor(props) {
     super(props);
 
+    this.setState({
+      campaigns: []
+    });
+
     this.controller = new CampaignController();
     this.campaigns = [];
```
DFurnes: This can be removed now.

lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSlide.js
```diff
@@ -35,20 +41,22 @@ class CampaignSlide extends React.Component {
           <div className="wrapper">
             {this.renderHeader()}
 
-            {this.campaigns.map(campaign => {
-              // Fix for local files / data missing by preventing null error
-              const image_uri = campaign.cover_image.default != null ? campaign.cover_image.default.sizes.landscape.uri : '';
+            <div className="gallery -triad">
+              {this.state.campaigns.map(campaign => {
+                // Fix for local files / data missing by preventing null error
+                const image_uri = campaign.cover_image.default != null ? campaign.cover_image.default.sizes.landscape.uri : '';
```
DFurnes, line 46: Can we use [Lodash's `get`](https://lodash.com/docs/4.16.0#get) for this? The ternary is kinda ugly to read and potentially brittle if other keys in the chain are missing.

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_card.scss
weerd: Hmmm, not keen on this class. Maybe we change the base card to add that top margin/padding on the heading and take it off the card. This way, if there is a heading then the right spacing is added, otherwise none is.

itsjoekent: beta_card_row is used by more than just the header though

lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSlide.js
```diff
@@ -35,20 +41,22 @@ class CampaignSlide extends React.Component {
...
+                  id: campaign.id,
+                  image_uri: image_uri,
+                  tagline: campaign.tagline,
+                  title: campaign.title,
+                  signup: this.controller.signup.bind(this.controller)
+                }
 
-              return <CampaignSignupCard {...props} key={campaign.id} />
-            })}
+                return <CampaignSignupCard {...props} key={campaign.id} />
```
DFurnes: I think it reads a little better to just specify the props here, rather than forming a temporary object above and applying it with a spread. It's okay if the element wraps to multiple lines, for example:

``` jsx
return (
  <CampaignSignupCard id={campaign.id} image_uri={image_uri} tagline={campaign.tagline}
    key={campaign.id} title={campaign.title} signup={this.controller.signup} />
);
```

lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSignupCard.js
```diff
@@ -66,26 +66,29 @@ class CampaignSignupCard extends React.Component {
...
+    const header = this.renderHeader();
     return (
-      <article className="beta-card">
-        <div className="beta-card__body">
+      <li>
+        <article className={`beta-card ${header === null ? '-headerless' : ''}`}>
+          <div className="beta-card__body">
 
-          {this.renderHeader()}
+            {header}
```
weerd: I think it was fine before w/out saving it to a `const` and just calling `{this.renderHeader()}`. You don't seem to be using it elsewhere, where it would warrant reuse?

lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSlide.js
DFurnes: I'd also recommend always binding the methods in the constructor of their class. It removes some extra cruft from  references elsewhere, avoids a likely place for bugs if we happen to forget, and has a nice little micro-optimization in performance (only binds once, rather than re-binding on every single render). For example:

``` js
class CampaignController {
  constructor() {
    this.signup = this.signup.bind(this);
  }

  signup() {
    // ... `this` will be CampaignController no matter what now
  }
}
```

DFurnes (commented):
Left a few suggestions.

lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSlide.js
itsjoekent: @DFurnes Re: spread, is this just a stylistic thing we prefer?

DFurnes: Yeah, I don't have _strooooong_ feels on it, but I think it looks better stylistically & then we're being explicit about exactly what items we're passing to a component.

lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSignupCard.js
```diff
@@ -66,26 +66,29 @@ class CampaignSignupCard extends React.Component {
   }
 
   render() {
+    const header = this.renderHeader();
     return (
-      <article className="beta-card">
-        <div className="beta-card__body">
+      <li>
```
weerd: Can you follow sort of what I did for the slide that renders ReportbackItems, with not having it output the `<li>`s in the component. For the slide with the single campaign card, this will output an `<li>` tag which we don't want.

weerd (changes requested):
Updates regarding some markup stuffs.

lib/themes/dosomething/paraneue_dosomething/js/components/CampaignSignupCard.js
itsjoekent: @weerd Lookup a bit, was using it for the -headerless class

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_card.scss
weerd: I'm confused... what does this have to do with the row? It's dealing w/ just the card...

weerd: Oh, ok. Sorry, I confused myself. I think eventually we should clean this up cause not ideal, but ok for merge.

itsjoekent: the only style `.-headerless` applies is to tell the first `.beta-card__row` in `.card` to have 0 margin-top. When you said,

> Maybe we change the base card to add that top margin/padding on the heading and take it off the card

I was trying to point out that the margin is being applied on `.beta_card__row`, and the beta card row is used in multiple places so it won't be easy to remove the margin from this entire class & shift it to the header.

itsjoekent: For example, this is what it looks like if we removed the margin from .beta_card__row

screen shot 2016-09-20 at 12 30 12 pm

If I remove the `.-headerless`, you can see where `.beta-card__row` is applying the margin

screen shot 2016-09-20 at 12 31 17 pm

and this is what happens when you add `.-headerless`
screen shot 2016-09-20 at 12 35 26 pm

DFurnes (approved):

weerd (approved):

itsjoekent:
🚂 🚋 🚋

PR #7059, opened 2016-09-19
OpenID Connect: Store access & refresh tokens.

DFurnes:
#### What's this PR do?

This pull request adds access and refresh token fields to the user's profile, and updates them whenever they authenticate using OpenID Connect.

Logging in via OpenID Connect should result in those three fields being filled out on your user profile. The fields should not be editable. Copying and pasting the access token into the [JWT Debugger](https://jwt.io) should show the same expiration as we store in the timestamp field.

We don't use these at the moment, but the idea is that in the future these can be used to make requests to any other internal services using the authed user's credentials.
#### Relevant tickets

#### Checklist
- [ ] Tested on staging.


lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
```diff
@@ -42,5 +42,12 @@ function dosomething_northstar_openid_connect_claims_alter(&$claims) {
  * @param string $provider - The provider used to authenticate
  */
 function dosomething_northstar_openid_connect_post_authorize($tokens, $account, $userinfo, $provider) {
-  // @TODO: Persist these in the database.
+  $edit = [];
+  dosomething_user_set_fields($edit, [
```
angaither, line 46: do we need an if (module_exists) around this stuff? not sure if the fields always exist or if they're only created if the module is enabled

DFurnes: I added the fields on `dosomething_user` since they're part of the profile, so should be safe.

weerd:
¡OLÉ! :sheep: it!

PR #7066, opened 2016-09-20
Adding logic to avoid global redirect logic craziness for API requests.

weerd:
#### What's this PR do?

This PR just adds a tiny bit of logic to help avoid running through all the crazy **dosomething_global.module** `init()` function logic for API requests where we don't need to figure out any country prefixes to add to the request path! 

👓 

💅 
#### Relevant tickets

Fixes 🍰 
#### Checklist
- [ ] Tested on staging.

---

@DFurnes @angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -355,7 +355,7 @@ function dosomething_helpers_ip_address() {
 function dosomething_helpers_is_api_request() {
   $path = explode('/', request_path());
 
-  return in_array('api', $path);
+  return $path[0] === 'api';
 }
 
 /**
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -348,6 +348,17 @@ function dosomething_helpers_ip_address() {
...
 
 /**
+ * Utility function to check if current request is an api request.
+ *
+ * @return boolean
+ */
+function dosomething_helpers_is_api_request() {
+  $path = explode('/', request_path());
+
+  return in_array('api', $path);
```
DFurnes: Should we compare against `$path[0]` here? Wondering if we'd ever want this to trigger for a page where the first part isn't `api/`... e.g. `www.dosomething.org/documentation/api` or something like that.

DFurnes (commented):
One question, otherwise looks good!

DFurnes (approved):

PR #7070, opened 2016-09-21
Add redirect to logout of Northstar as well.

DFurnes:
#### What's this PR do?

This PR adds a logout hook to redirect to Northstar and kill a single sign-on as well.

~~I don't know if we can only do this for OpenID Connect sessions... any ideas?~~ We added a session variable to track whether a user's session was started via OpenID Connect, so we only redirect to Northstar to kill the single sign-on session then.

I think this seems okay, but would love to hear any thoughts/counter-arguments!

➡️ 🚪 👋 
#### Relevant tickets

References DoSomething/northstar#434.
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
@@ -51,3 +51,22 @@ function dosomething_northstar_openid_connect_post_authorize($tokens, $account, 
 
   user_save($account, $edit);
 }
+
+/**
+ * Redirect to destroy a Northstar single-sign-on session if one exists.
+ *
+ * Implements hook_user_logout().
+ *
+ * @param $account
+ */
+function dosomething_northstar_user_logout($account) {
+  // @TODO: Conditionally apply this only for OpenID Connect logins...?
+  $logout_url = url(NORTHSTAR_URL . '/logout', [
+    'query' => ['redirect' => url('/', ['absolute' => TRUE])],
+  ]);
+
+  // Destroy the current session, and reset $user to the anonymous user.
+  session_destroy();
+
+  drupal_goto($logout_url);
+}
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
@@ -49,6 +49,9 @@ function dosomething_northstar_openid_connect_post_authorize($tokens, $account, 
     'access_token_expiration' => $tokens['expire'],
   ]);
 
+  // Let's remember that this is a OpenID Connect session.
+  $_SESSION['DOSOMETHING_NORTHSTAR_OPENID_CONNECT'] = true;
+
   user_save($account, $edit);
 }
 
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
```diff
@@ -51,3 +51,22 @@ function dosomething_northstar_openid_connect_post_authorize($tokens, $account,
...
 }
+
+/**
+ * Redirect to destroy a Northstar single-sign-on session if one exists.
+ *
+ * Implements hook_user_logout().
+ *
+ * @param $account
+ */
+function dosomething_northstar_user_logout($account) {
```
DFurnes, line 65: One gotcha: because of the `drupal_goto` part of this, this will prevent any other hooks from firing afterwards (and hence why we need to `session_destroy` in here rather that using Drupal's default in `user_logout`).

I don't see a relevant "post-logout" hook but open to suggestions!

DFurnes:
I think this is probably safe to merge in so we can test it out on staging/Thor. It'll only redirect to Northstar for the SSO logout when the `DOSOMETHING_NORTHSTAR_OPENID_CONNECT` session key has been set, so we won't be unnecessarily redirecting traditional logins.

/cc @angaither for thoughts on Monday.

weerd:
¡OLÉ! :sheep: it!

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
```diff
@@ -49,5 +49,27 @@ function dosomething_northstar_openid_connect_post_authorize($tokens, $account,
...
+  $is_openid_login = ! empty($_SESSION['DOSOMETHING_NORTHSTAR_OPENID_CONNECT']);
```
angaither, line 66: these two lines should probably only happen if the modules exist 😺

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
```diff
@@ -49,5 +49,27 @@ function dosomething_northstar_openid_connect_post_authorize($tokens, $account,
...
+  $northstar_logout_url = url(NORTHSTAR_URL . '/logout', [
+    'query' => ['redirect' => url('/', ['absolute' => TRUE])],
+  ]);
+
+  // Destroy the current session, and reset $user to the anonymous user.
+  session_destroy();
```
angaither, line 72: do you need to call this manually? shouldn't that happen as a part of [`user_logout`](https://api.drupal.org/api/drupal/modules!user!user.pages.inc/function/user_logout/7.x) already?

angaither (changes requested):

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
DFurnes: Yep yep, mentioned on a line comment above:

> One gotcha: because of the `drupal_goto` part of this, this will prevent any other hooks from firing afterwards (and hence why we need to `session_destroy` in here rather that using Drupal's default in `user_logout`).

Luckily `dosomething_northstar` is the only module we have installed that uses the `hook_user_logout` hook, but still perhaps something to think carefully about. 💭

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
DFurnes: Which module? The `openid_connect` module is always enabled since it's in our installation profile, and `dosomething_northstar` should be enabled since it's a hook in that module.

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
angaither: If I only had a brain. 📖

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
angaither: NS, it's on my brain b/c I just ran into this error #7073

DFurnes: Oh, yeah so I don't think that should be an issue right?

angaither: oy yeah, I think you're right.

angaither (approved):

PR #7074, opened 2016-09-22
Fix fatal error on review locally.

### LINKED ISSUE #7073: Fatal error on RB review on local
#### Current Behavior

locally, if you review reportbacks you get a PHP fatal
`) Fatal error: Call to undefined function dosomething_northstar_get_northstar_user() in /var/www/dev.dosomething.org/lib/modules/dosomething/dosomething_reportback/includes/ReportbackItem.php on line 113` 
because the NS module is turned on a `ds pull`
#### Desired Behavior
- that doesn't happen
#### Steps to Replicate
- review a reportback locally

angaither:
#### What's this PR do?

check if NS module exists before calling a function in that module 

👀 

this module gets disabled on local, so it will always break
also always check to see if a module exists before calling a function in there. 
#### Relevant tickets

Fixes #7073

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/includes/ReportbackItem.php
+++ b/lib/modules/dosomething/dosomething_reportback/includes/ReportbackItem.php
@@ -109,7 +109,7 @@ class ReportbackItem extends Entity {
     ];
     $this->campaign = Campaign::get($data->nid);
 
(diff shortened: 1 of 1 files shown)


sbsmith86:
👍

PR #7078, opened 2016-09-22
Use the configured Northstar URL.

DFurnes:
#### What's this PR do?

This is what I get for only testing this on local! This PR updates the Northstar OpenID Connect provider to use the actual Northstar URL.

We should be safe using this constant because the provider is provided as a plugin bundled in the Northstar module, so it'll never not be defined. Anyways, 👀.

🍃
#### Relevant tickets

Fixes 🐛.
#### Checklist
- [ ] Tested on staging.


weerd (approved):
😱 🙀

PR #7080, opened 2016-09-22
OpenID Connect: Implements custom authorize button.

DFurnes:
#### What's this PR do?

After merging this, we'll have a nice and stable prototype of OpenID Connect authentication on staging that we can demo & anyone can play around with! This PR implements our own custom markup for the "authorize" button, for two reasons:
1. The included Form API array in the `openid_connect` module had PHP warnings.
2. We want to eventually try showing this form inside an `<iframe>`.

👀

It's a bummer that we have to re-implement all the logic for building the authorization link, but unfortunately the `openid_connect` module doesn't give a handy way to _only_ override that one part of the parent method without also setting it as a `drupal_goto` redirect. Oh well.
#### Relevant tickets

Closes #426. 🎉
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -295,9 +295,6 @@ function dosomething_northstar_log_request($op, $user, $request_body, $response)
  * @return array
  */
 function dosomething_northstar_openid_login_view() {
-  // Save our intended destination so we'll be redirected afterwards.
-  openid_connect_save_destination();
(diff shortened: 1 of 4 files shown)


weerd (approved):
Looks good!

DFurnes:
@weerd Added two more fixes: it will now always redirect to the profile page like we do with normal logins (4d96c78), and fixes an issue verifying the state token when modal was enabled on the `/user/openid` login page (ece1b7a).

weerd:
Updates look good.

PR #7082, opened 2016-09-22
Remove the default client info from features.

DFurnes:
#### What's this PR do?

This removes the `client_id` and `client_secret` settings from features so it doesn't get reverted on every deploy. It also removes the features for the "generic" provider since that isn't used (outside of my initial testing).

👀

🔥
#### Relevant tickets

Fixes 🐛.
#### Checklist
- [ ] Tested on staging.


weerd (approved):
👍

PR #7086, opened 2016-09-26
Northstar php  enablealizationing

weerd:
#### What's this PR do?

This PR incorporates Northstar-PHP into Phoenix. It sets up the package, including the repository for the client credentials tokens. It does not yet begin the process of switching out the old code in **dosomething_northstar** module from using `_dosomething_northstar_build_http_client()` to build the client and `drupal_http_request()` to make the HTTP calls. This will happen in an upcoming PR to help keep the focus of this one to just enabling the use of Northstar-PHP from within Phoenix.

👀 
#### Relevant tickets

Fixes https://github.com/DoSomething/TeamRocket/issues/12
#### Checklist
- [ ] Tested on staging.

---

@DFurnes @angaither 

@jessleenyc @ngjo


lib/app/Auth/PhoenixOAuthRepository.php
```diff
@@ -0,0 +1,85 @@
...
+    return NULL;
+  }
+
+  /**
+   * Get the OAuth client's token.
+   *
+   * @return \League\OAuth2\Client\Token\AccessToken|null
+   */
+  public function getClientToken() {
+    $clientId = variable_get('dosomething_northstar_oauth_client_token_clientid', NULL);
```
DFurnes: This variable looks unused.

lib/app/Auth/PhoenixOAuthRepository.php
```diff
@@ -0,0 +1,85 @@
...
+   * Save the access token for an authorized client.
+   *
+   * @param  string $clientId
+   * @param  string $accessToken
+   * @param  integer $expiration
+   * @param  string $role
+   * @return void
+   */
+  public function persistClientToken($clientId, $accessToken, $expiration, $role) {
+    variable_set('dosomething_northstar_oauth_client_token_clientid', $clientId);
```
DFurnes: See above.

DFurnes (changes requested):
Two Qs, otherwise looks good!

lib/app/Auth/PhoenixOAuthRepository.php
weerd: Yeah, I was referencing the example from the `LaravelOAuthRepository` in the NS-PHP package, and kinda realized that's only really used as an id to find the record in the DB... wasn't sure if we'd _want_/_need_ it here in Phoenix land.

DFurnes (approved):
🚢 🌠 🚀

PR #7095, opened 2016-09-29
Northstar php switcheroo

weerd:
#### What's this PR do?

This PR implements the use of the newly added [Northstar-PHP](https://github.com/DoSomething/northstar-php) package in Phoenix. It specifically targets the `dosomething_northstar_get_northstar_user()` function, which has been renamed to `dosomething_northstar_get_user()` and updates any locations that use this function to grab a single NS user.

It also updates composer to reference the latest release candidate version of NS-PHP.

This PR also switches out the use of `dosomething_helpers_isset()` function which is slowly being deprecated in favor of `data_get()`.

👀 You can also feel free to spin up Phoenix locally and test a few of the locations where the `dosomething_northstar_get_user()` is used and make sure you get the expected results.
#### Relevant tickets

Refs https://github.com/DoSomething/TeamRocket/issues/11
#### Checklist
- [ ] Tested on staging.

---

@DFurnes @angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -746,31 +746,28 @@ function dosomething_helpers_get_user_info($drupal_id) {
     return NULL;
   }
 
-  $northstar_response = dosomething_northstar_get_northstar_user($drupal_id);
-  $northstar_response = json_decode($northstar_response);
+  $northstar_user = dosomething_northstar_get_user($drupal_id);
 
-  if ($northstar_response && isset($northstar_response->error)) {
-    return NULL;
+  if ($northstar_user) {
+    return [
+      'id' => data_get($northstar_user, 'id'),
+      'email' => data_get($northstar_user, 'email'),
+      'first_name' => data_get($northstar_user, 'first_name'),
+      'last_name' => data_get($northstar_user, 'last_name'),
+      'last_initial' => data_get($northstar_user, 'last_initial'),
+      'photo' => data_get($northstar_user, 'photo'),
+      'birthdate' => data_get($northstar_user, 'birthdate'),
+      'addr_street1' => data_get($northstar_user, 'addr_street1'),
+      'addr_street2' => data_get($northstar_user, 'addr_street2'),
+      'addr_city' => data_get($northstar_user, 'addr_city'),
+      'addr_state' => data_get($northstar_user, 'addr_state'),
+      'addr_zip' => data_get($northstar_user, 'addr_zip'),
+      'country' => data_get($northstar_user, 'country'),
+      'drupal_id' => data_get($northstar_user, 'drupal_id'),
+    ];
   }
 
-  $northstar_user = $northstar_response->data;
-
-  return [
-    'id' => data_get($northstar_user, 'id'),
-    'email' => data_get($northstar_user, 'email'),
-    'first_name' => data_get($northstar_user, 'first_name'),
-    'last_name' => data_get($northstar_user, 'last_name'),
-    'last_initial' => data_get($northstar_user, 'last_initial'),
-    'photo' => data_get($northstar_user, 'photo'),
-    'birthdate' => data_get($northstar_user, 'birthdate'),
-    'addr_street1' => data_get($northstar_user, 'addr_street1'),
-    'addr_street2' => data_get($northstar_user, 'addr_street2'),
-    'addr_city' => data_get($northstar_user, 'addr_city'),
-    'addr_state' => data_get($northstar_user, 'addr_state'),
-    'addr_zip' => data_get($northstar_user, 'addr_zip'),
-    'country' => data_get($northstar_user, 'country'),
-    'drupal_id' => $drupal_id,
-  ];
+  return NULL;
 }
 
 /**
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -24,14 +24,27 @@ define('NORTHSTAR_APP_KEY', variable_get('dosomething_northstar_app_key', 'secre
  * @return DoSomething\Northstar\Northstar
  */
 function dosomething_northstar_client() {
-  return new Northstar([
+  // If record is set in the cache, grab it...
+
+  // If record not in cache, retrieve it from Northstar...
+  $client = new Northstar([
     'grant' => 'client_credentials',
     'url' => NORTHSTAR_URL,
-    'client_id' => NORTHSTAR_APP_ID,
-    'client_secret' => NORTHSTAR_APP_KEY,
-    'scope' => ['admin'],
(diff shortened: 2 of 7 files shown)


lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -746,31 +746,28 @@ function dosomething_helpers_get_user_info($drupal_id) {
...
-  $northstar_response = dosomething_northstar_get_northstar_user($drupal_id);
-  $northstar_response = json_decode($northstar_response);
-
-  if ($northstar_response && isset($northstar_response->error)) {
-    return NULL;
+  $northstar_user = dosomething_northstar_get_user($drupal_id);
+
+  if ($northstar_user) {
+    return [
+      'id' => data_get($northstar_user, 'id'),
```
DFurnes, line 753: Would it make more sense to just return the `NorthstarUser` here, rather than re-serializing as an array?

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -153,36 +155,35 @@ function dosomething_northstar_update_user($user, $password = null) {
...
+function dosomething_northstar_get_user($id, $type = 'drupal_id') {
+  $cache_bin = 'cache_dosomething_northstar';
+  $cache_key = 'northstar_user_'.$type.'='.$id;
+
+  $northstar_user = cache_get($cache_key, $cache_bin);
 
-  if ($northstar_user_info) {
-    $northstar_user_info = $northstar_user_info->data;
+  if ($northstar_user) {
+    return $northstar_user->data;
```
DFurnes, line 171: Does this work? I don't believe `data` is accessible unless you're somehow extracting the raw JSON.

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -153,36 +155,35 @@ function dosomething_northstar_update_user($user, $password = null) {
...
 
-    $northstar_user_info = drupal_http_request($client['base_url'] . '/users/drupal_id/' . $drupal_id, array(
-      'headers' => $client['headers'],
-      'method' => 'GET',
-      ));
+  $northstar_user = dosomething_northstar_client()->getUser($type, $id);
 
-      $northstar_user_info = $northstar_user_info->data;
+  if ($northstar_user) {
+    cache_set($cache_key, $northstar_user, $cache_bin, REQUEST_TIME + 60*60*24);
```
DFurnes, line 177: Since this will be a `NorthstarUser` rather than an array, we may need to add a `->toArray()` method to the response class to serialize before caching here.

DFurnes (commented):
Left a few questions/suggestions.

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
weerd: Yeah, it's been working for me while testing all this...

weerd: Oh sorry, so this is only for when it's getting pulled from Redis cache. Drupal pops it into an object with a `data` property, some timestamps and the `cid` key used to cache it. So if it's found in redis cache, I need to grab it from this data property.

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
weerd: Yeah, I think I can do that, but need to test. There's other things I still need to update, but this first PR was just meant to tackle the `dosomething_northstar_get_northstar_user()` function and replace it w/ NS-PHP call. This is just the first step.

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
weerd: Here's an example of how the record is stored in cache _after_ Drupal adds a few things to it.

```
stdClass Object
(
    [created] => 1475173802
    [expire] => 1475260201
    [data] => DoSomething\Northstar\Resources\NorthstarUser Object
        (
            [attributes:protected] => Array
                (
                    [id] => 57e112947hdy4799278b4b7c
                    [_id] => 57e112947hdy4799278b4b7c
                    [first_name] => Jen
                    [last_name] => a
                    [last_initial] => A
                    [photo] => 
                    [email] => bob@dosomething.org
                    [mobile] => 
                    [facebook_id] => 
                    [interests] => 
                    [birthdate] => 1986-12-31 00:00:00
                    [addr_street1] => 
                    [addr_street2] => 
                    [addr_city] => 
                    [addr_state] => 
                    [addr_zip] => 
                    [source] => phoenix
                    [mobilecommons_id] => 
                    [parse_installation_ids] => 
                    [mobilecommons_status] => 
                    [language] => en
                    [country] => 
                    [drupal_id] => 1700000
                    [role] => user
                    [updated_at] => 2016-09-20T16:21:19+0000
                    [created_at] => 2016-06-01T14:39:44+0000
                )

            [dates:protected] => Array
                (
                    [0] => created_at
                    [1] => updated_at
                )

        )

    [cid] => northstar_user_drupal_id=1700000
    [serialized] => 1
)
```

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
weerd: Hmmm, I don't think so... I checked Redis cache and the data is saved as a serialized hash. When it's returned I can still access it as an object.

weerd: You can see in the [example](https://github.com/DoSomething/phoenix/pull/7095/files#r81214754) retrieved cache record, it's still a `NorthstarUser` class object.

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
DFurnes: That makes sense! 👍

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
DFurnes: Oh, got it! I was confused with the `data` wrapper in the JSON response.

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
DFurnes: Awesome!

DFurnes (approved):
⛵

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
weerd: 💃

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
weerd: Yep yep! Defs... :drupal: always causes confusion :trollface:

PR #7096, opened 2016-09-29
Reworking some old functions to better support returned values after using NS-PHP package

weerd:
#### What's this PR do?

This PR updates the `dosomething_helpers_get_current_user()` function so that it returns the user as `NorthstarUser` object thanks to the newly added NS-PHP package. It also adds an additional parameter to the function so that the user information can be returned as an array. This is used for outputting user information into the `Drupal.settings.dsUser` object for use by JavaScript on the front-end.

👓 👀  all looks good?
#### Relevant tickets

Refs https://github.com/DoSomething/TeamRocket/issues/11
#### Checklist
- [ ] Tested on staging.

---

@DFurnes @angaither

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -699,7 +699,7 @@ function dosomething_helpers_get_current_language_content_code() {
 /**
  * Get the current, logged in user as a NorthstarUser object or array.
  *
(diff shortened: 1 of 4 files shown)


DFurnes (approved):
👍

weerd:
@DFurnes found a small bug while testing. See latest commit for fix and also did a little extra cleanup!

DFurnes (approved):
✅!

PR #7098, opened 2016-09-30
Adds signup button tracking on pitch page

itsjoekent:
#### What's this PR do?

Tracks everytime a user clicks signup on the pitch page

Enable GA debugging
Click signup on a pitch page
Verify event fired
#### Relevant tickets

https://trello.com/c/I6qIdHUq/65-as-an-admin-i-want-to-track-number-of-sign-up-clicks-on-all-pitch-pages-so-that-i-can-see-how-many-users-click-sign-up-and-make-?menu=filter&filter=label:orange,label:Onboarding,label:User,label:purple,@joekent3

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/utilities/Analytics.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/utilities/Analytics.js
@@ -52,7 +52,6 @@ function subscribeEvents() {
 }
 
 function exists() {
-  return true;
   return typeof ga !== 'undefined' && ga !== null;
 }
 
--- a/lib/themes/dosomething/paraneue_dosomething/js/utilities/Analytics.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/utilities/Analytics.js
@@ -51,7 +51,6 @@ function subscribeEvents() {
 }
 
 function exists() {
-  return true;
   return typeof ga !== 'undefined' && ga !== null;
 }
 
--- a/lib/themes/dosomething/paraneue_dosomething/js/utilities/Analytics.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/utilities/Analytics.js
@@ -40,9 +40,18 @@ function subscribeEvents() {
     const label = args !== null ? '#' + args.attr('id') : '';
     ga('send', 'event', 'Modal', 'Close', label);
   });
+
+  // Signup button on pitch page
+  const urlParts = window.location.pathname.split('/');
+  if (urlParts[urlParts.length - 1] === 'pitch') {
+    $('#dosomething-signup-form-primary').on('submit', () => {
+      ga('send', 'event', 'Signup button', 'Pitch page');
+    });
+  }
 }
 
 function exists() {
+  return true;
   return typeof ga !== 'undefined' && ga !== null;
 }
 
--- a/lib/themes/dosomething/paraneue_dosomething/js/utilities/Analytics.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/utilities/Analytics.js
@@ -45,8 +45,14 @@ function subscribeEvents() {
   const isPitchPage = $('.pitch').length > 0;
   if (isPitchPage) {
     // This applies to both authenticated & anonymous users. It also covers both modal register & login.
-    $('.header__signup form, #user-login-form, #user-register-form').on('submit', () => {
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/js/utilities/Analytics.js
```diff
@@ -40,6 +40,14 @@ function subscribeEvents() {
     const label = args !== null ? '#' + args.attr('id') : '';
     ga('send', 'event', 'Modal', 'Close', label);
   });
+
+  // Signup button on pitch page
+  const urlParts = window.location.pathname.split('/');
+  if (urlParts[urlParts.length - 1] === 'pitch') {
```
DFurnes: Have you tested this with a normal account? As far as I  know, `/pitch` is just an override URL for admins.

For example, this is the pitch page URL for Lose Your V-Card (for a logged-out or not signed up user):

[`https://www.dosomething.org/us/campaigns/lose-your-v-card`](https://www.dosomething.org/us/campaigns/lose-your-v-card)

lib/themes/dosomething/paraneue_dosomething/js/utilities/Analytics.js
```diff
@@ -40,6 +40,14 @@ function subscribeEvents() {
...
+    $('#dosomething-signup-form-primary').on('submit', () => {
```
DFurnes: Is this the ID for both the logged-in "sign up" button form and the logged-out "register and sign up" form? If not, will this still give us the results we're expecting for this test?

DFurnes (changes requested):

lib/themes/dosomething/paraneue_dosomething/js/utilities/Analytics.js
itsjoekent: Hmm thats a good point, maybe I could go by classname?

`class="campaign campaign--pitch pitch"`

DFurnes: That seems like a better idea. :thumbsup:

lib/themes/dosomething/paraneue_dosomething/js/utilities/Analytics.js
itsjoekent: Good catch! `link--campaign-signup-login` is the other ID.

Potentially better option, listen for the button submit under `.header__signup`

lib/themes/dosomething/paraneue_dosomething/js/utilities/Analytics.js
```diff
@@ -40,6 +40,15 @@ function subscribeEvents() {
     const label = args !== null ? '#' + args.attr('id') : '';
     ga('send', 'event', 'Modal', 'Close', label);
   });
+
+  // Signup button on pitch page
+  const isPitchPage = $('.pitch').length > 0;
+  if (isPitchPage) {
+    // This applies to both authenticated & anonymous users. It also covers both modal register & login.
+    $('.header__signup form, #user-login-form, #user-register-form').on('submit', () => {
```
DFurnes: Have you tested that `#user-register-form` doesn't trigger this analytics call if there are form validation errors? That may have to be a second event listener on the `Validation:Submitted` event so we don't get false positives from that form (see example usage above, `args` is the form ID).

DFurnes (changes requested):

lib/themes/dosomething/paraneue_dosomething/js/utilities/Analytics.js
itsjoekent: Looks like it doesn't trigger this event listener, the only GA event I see firing is for the validation error

DFurnes: So all working as expected (including actually submitting on successful validation)? If so, merge away!

DFurnes (approved):

lib/themes/dosomething/paraneue_dosomething/js/utilities/Analytics.js
itsjoekent: @DFurnes Actually looks like I didn't catch it the first time, I think I'll need to do the Validation Submitted approach

itsjoekent:
@DFurnes okie, _now_ it works  with 
- new accounts
- logging in
- currently logged in

DFurnes:
🚢

PR #7099, opened 2016-10-03
Appends source param to homepage links

itsjoekent:
#### What's this PR do?

Appends `source=homepage_campaign_tile` to homepage campaign links

Click the campaigns on the homepage, is the source param in the url?
#### Relevant tickets

Fixes https://github.com/DoSomething/TeamRocket/issues/26

### DIFF
--- a/lib/modules/dosomething/dosomething_home/dosomething_home.module
+++ b/lib/modules/dosomething/dosomething_home/dosomething_home.module
@@ -88,7 +88,7 @@ function dosomething_home_preprocess_node(&$vars) {
       $tiles['image'] = '<img src="data:image/gif;base64,R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs=" data-src="' . $tiles['image_url'] . '">';
 
       $link_language = dosomething_global_get_language($user, $node, FALSE);
-      $tiles['link'] = dosomething_global_node_url($nid, $link_language);
+      $tiles['link'] = dosomething_global_node_url($nid, $link_language) . '?source=homepage_campaign_tile';
       $thumbnails[$nid] = paraneue_dosomething_get_gallery_item($tiles, 'tile');
     }
   }
--- a/lib/modules/dosomething/dosomething_home/dosomething_home.module
+++ b/lib/modules/dosomething/dosomething_home/dosomething_home.module
@@ -19,6 +19,8 @@ function dosomething_home_preprocess_node(&$vars) {
   // Adds $subtitle variable.
   $vars['subtitle'] = $vars['field_subtitle'][0]['safe_value'];
 
+  $source = '?source=node/' . $vars['nid'];
+
   // Prepare campaign thumbnails for homepage
   $campaigns = array();
   if(isset($vars['field_campaigns'])) {
@@ -88,7 +90,7 @@ function dosomething_home_preprocess_node(&$vars) {
       $tiles['image'] = '<img src="data:image/gif;base64,R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs=" data-src="' . $tiles['image_url'] . '">';
 
       $link_language = dosomething_global_get_language($user, $node, FALSE);
-      $tiles['link'] = dosomething_global_node_url($nid, $link_language) . '?source=node/1141';
+      $tiles['link'] = dosomething_global_node_url($nid, $link_language) . $source;
       $thumbnails[$nid] = paraneue_dosomething_get_gallery_item($tiles, 'tile');
     }
   }
--- a/lib/modules/dosomething/dosomething_home/dosomething_home.module
+++ b/lib/modules/dosomething/dosomething_home/dosomething_home.module
@@ -88,7 +88,7 @@ function dosomething_home_preprocess_node(&$vars) {
       $tiles['image'] = '<img src="data:image/gif;base64,R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs=" data-src="' . $tiles['image_url'] . '">';
 
       $link_language = dosomething_global_get_language($user, $node, FALSE);
-      $tiles['link'] = dosomething_global_node_url($nid, $link_language) . '?source=homepage_campaign_tile';
+      $tiles['link'] = dosomething_global_node_url($nid, $link_language) . '?source=node/1141';
       $thumbnails[$nid] = paraneue_dosomething_get_gallery_item($tiles, 'tile');
     }
   }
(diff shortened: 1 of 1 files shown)


angaither:
two things
1. this should already be stored as `node/1141`
2. query strings should be appended with [`url`](https://api.drupal.org/api/drupal/includes!common.inc/function/url/7.x)

but I don't think we need this pr

itsjoekent:
cc: @jessleenyc ^ Andrea said we already track this

jessleenyc:
@angaither does that mean i should see 
https://www.dosomething.org/us/campaigns/treats-troops?source=node/1141 when i click signup from a tile on the homepage? if so, i don't think me or @deadlybutter are seeing this happen.

angaither:
:thinking: I wonder if that broke when we launched #global with all the redirects

ngjo:
Separately and less priority than this global issue... iono what the difficulty level would be, but could be useful to rename node/1141 to homepage_campaign_tile just so it's more clear what it is for someone looking at the database sources. Not even sure if we can retroactively change all node/1141 instances or not either. 

Anyways, that's my inclination, but if most people looking at the database know what node/1141 is or there's some sort of glossary that explains that, then ignore me.

jessleenyc:
@deadlybutter 
1) can you see if this broke when we launched global? and if so, find a fix for it?
2) once we're sure tracking is working can you see if it's possible to rename the parameter from node/1141 to 'homepage_campaign_tile'? cc @ngjo

jessleenyc:
[(3) As an engineer, I want all homepage campaigns to be tagged with a source so it's tracked in our tables](https://trello.com/c/zM5OjkwK/60-3-as-an-engineer-i-want-all-homepage-campaigns-to-be-tagged-with-a-source-so-it-s-tracked-in-our-tables)

itsjoekent:
@jessleenyc Regardless of when/why it broke, this is the fix for both of those things

jessleenyc:
@deadlybutter but that means the data is separate right? in order to see which signups came from the homepage tiles, we would have to query for both '/node/1141' and 'homepage_campaign_tile'?

itsjoekent:
Oh you want to retroactively go back as well

jessleenyc:
ideally, i would think so! @ngjo?

ngjo:
personal preference, id say that it'd be clearer to change it to homepage_campaign_tile for people to understand what the source is...node/1141 isn't very telling). but we might want to ask jeff or mike (or anyone on the data team) what their preference is -- if they're fine just remember that it's node/1141, then we can keep it that way.

@deadlybutter would be useful to know how difficult/time consuming/problematic it could be to retroactively change node/1141 to homepage_campaign_tile tho cc @jessleenyc

itsjoekent:
its not super problematic, we'd have to make a script in Drupal which we would run on prod

angaither:
I'd say let's _not_ do a script. There isn't an index on that row, and I think the script would probably be pretty slow. IMO I don't think it's worth it. 

We've been working under that data assumption for ~2 years and haven't had issues with it

itsjoekent:
`node/1141` it is then!

itsjoekent:
@angaither ready for another look!

angaither:
@deadlybutter [#2 from my comment](https://github.com/DoSomething/phoenix/pull/7099#issuecomment-251207130) still applies

And as I mentioned in my original comment, this should already be working. If it's not we should fix that, not append a new source.

itsjoekent:
Looks like it was first implemented here, https://github.com/DoSomething/phoenix/commit/7d1ef3a8baaa46f077d88feafe5678a8eef19656

But it got removed during 🌐 (Andrea wins! 🎉 )
https://github.com/DoSomething/phoenix/commit/78239361dbc7425b9a5763edc714285a00fa57a1#diff-8faaf5002b546e41219b874defb2239e

In _theory_ `dosomething_global_node_url` should accept a 3rd argument (in this case source) & should properly append it to the URL... Gonna try it out

itsjoekent:
hm no it looks like the 3rd argument was designed to accept a url fragment, so the url looks like this `http://dev.dosomething.org:8888/us/node/1485#?source=node/1165`

I think I should probably stick with appending the string at the end, otherwise we'll have to modify global functions :D

itsjoekent:
alrighty its now updated to more closely resemble the original implementation

angaither:
@deadlybutter does it work and actually save to the DB when the source is a fragment with a query string? 
You shouldn't have to modify `dosomething_global_url` since the last argument is an array of options, so you can include a query string as an option.

itsjoekent:
@angaither problem with `dosomething_global_url` is I'd have to duplicate the logic for making the node URL. I'll try adding it via the fragment & see what happens

itsjoekent:
@angaither welp, I don't think I'll be able to give you an answer on that. The URL causes jquery to freak out `#?source=node/1165` which prevents the JS from working (so i cant open the signup modal)

```
Uncaught Error: Syntax error, unrecognized expression: #?source=node/1165
```

angaither:
@deadlybutter you don't have to duplicate logic to use `dosomething_global_url`

should be something like 

```
$path = `node/ . $vars['nid']`
dosomething_global_url($path, ['language' => $link_language, 'query' => 'source=' . $path])`
```

itsjoekent:
:thinking: ill give that a shot one sec

DFurnes:
I might be missing context so sorry if this is intentional, but you can't have a query-string after a URL fragment. For it to be parsed by the server, it'd have to go `?source=node/1165#...`.

itsjoekent:
@DFurnes yep, was just confirming that because andrea mentioned it here https://github.com/DoSomething/phoenix/pull/7099#issuecomment-251781831

DFurnes:
I think the JavaScript syntax error is probably because our modal package is trying to find a modal on the page to open with that ID (and then getting upset because IDs can't have `?`s). We should probably make an issue over there so we can fix that up sometime later.

itsjoekent:
code updated, generates the proper URL

```
http://dev.dosomething.org:8888/us/campaigns/test-campaign?source=node/1165
```

lib/modules/dosomething/dosomething_home/dosomething_home.module
```diff
@@ -88,7 +90,7 @@ function dosomething_home_preprocess_node(&$vars) {
       $tiles['image'] = '<img src="data:image/gif;base64,R0lGODlhAQABAIAAAAUEBAAAACwAAAAAAQABAAACAkQBADs=" data-src="' . $tiles['image_url'] . '">';
 
       $link_language = dosomething_global_get_language($user, $node, FALSE);
-      $tiles['link'] = dosomething_global_node_url($nid, $link_language);
+      $tiles['link'] = dosomething_global_url('node/' . $nid, ['language' => $link_language, 'query' => ['source' => $source]]);
```
angaither, line 93: one final nit picky thing, but if you're going to use a variable for `node/ . $nid` use it for both, otherwise scrap the temp var

itsjoekent: These are two different NID's. The first param here is for the actual node we're linking too. The source param is the NID of the homepage.

angaither: awwwyis!

angaither:
🚢 🔗 :trackball:

PR #7102, opened 2016-10-03
Send to rogue

### LINKED ISSUE #6960: Send reportback to rogue

sbsmith86:
#### What's this PR do?
- Sends Reportbacks to rogue before saving them in the DB. 
  - Gets the base64 encoded url of the uploaded image and sends that to rogue to be stored in s3
- Rogue then sends back the reportback to phoenix, then. 
- We store references to where the RB and RB item is stored in Rogue so we can send rogue any updates to a reportback or item.
- Send the user to the permalink page for the recently uploaded RB item.
- Also started sending over the cropping information as we will need to handle this in rogue 

Pull down this branch and the most recent rogue master. 

Submit a reportback and confirm that: 
- You are brought to the permalink page and the correct image/caption shows.
- That the reportback and reportback item are stored in the Rogue DB in the `reportback` and `reportback_item` tables.
- That the reportback and reportback item are stored in the Phoenix DB in the `dosomething_reportback` and `dosomething_reportback_file` tables.
- That the `dosomething_rogue_reportback` table has a record for the recently uploaded reportback item that stores the drupal `fid` and `rbid` and the corresponding Rogue `reportback.id` and `reportback_item.id`
#### Relevant tickets

Fixes #6960 
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -932,11 +932,11 @@ function dosomething_helpers_get_current_campaign_run_for_user($nid, $user = NUL
  * @return string
 */
 function dosomething_helpers_get_data_uri_from_image($image) {
-  $imagePath = drupal_realpath($image->source);
+  $image_path = drupal_realpath($image->source);
 
   // Read image path, convert to base64 encoding
-  $imageData = base64_encode(file_get_contents($imagePath));
+  $image_data = base64_encode(file_get_contents($image_path));
 
   // Return the formated data uri: data:{mime};base64,{data};
-  return 'data: ' . mime_content_type($imagePath) . ';base64,' . $imageData;
+  return 'data: ' . mime_content_type($image_path) . ';base64,' . $image_data;
 }
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -925,6 +925,12 @@ function dosomething_helpers_get_current_campaign_run_for_user($nid, $user = NUL
   return node_load($run['id']);
 }
 
+/**
+ * Returns a base64 encoded data uri for an Image object
+ *
+ * @param  Image object $image
+ * @return string
+*/
 function dosomething_helpers_get_data_uri_from_image($image) {
   $imagePath = drupal_realpath($image->source);
 
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -924,3 +924,13 @@ function dosomething_helpers_get_current_campaign_run_for_user($nid, $user = NUL
   // @TODO: may want to investigate how it loads the proper translation...
   return node_load($run['id']);
 }
+
+function dosomething_helpers_get_data_uri_from_image($image) {
+  $imagePath = drupal_realpath($image->source);
+
+  // Read image path, convert to base64 encoding
+  $imageData = base64_encode(file_get_contents($imagePath));
+
+  // Return the formated data uri: data:{mime};base64,{data};
+  return 'data: ' . mime_content_type($imagePath) . ';base64,' . $imageData;
+}
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -371,7 +371,8 @@ function dosomething_reportback_form_submit($form, &$form_state) {
     $values['uid'] = $user->uid;
   }
 
-  if (module_exists('dosomething_rogue')) {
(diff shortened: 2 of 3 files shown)


lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -924,3 +924,19 @@ function dosomething_helpers_get_current_campaign_run_for_user($nid, $user = NUL
...
 }
+
+/**
+ * Returns a base64 encoded data uri for an Image object
+ *
+ * @param  Image object $image
+ * @return string
+*/
+function dosomething_helpers_get_data_uri_from_image($image) {
+  $imagePath = drupal_realpath($image->source);
```
angaither: php vars are 🐍 case!

🐘

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -123,3 +128,29 @@ function dosomething_rogue_get_by_file_id($fid)
...
+
+/**
+ * Insert record that stores reference to the most recent uploaded reportback item in
+ * phoenix and it's corresponding id's in Rogue
+ *
+ * @param string $rbid
+ * @param string $fid
+ * @param object $rogue_reportback
+ *
+ * @return InsertQuery object
```
angaither, line 140: does this actually return an insert query if it's executed?

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
sbsmith86: ack wrong framework.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
sbsmith86: That is what the [documentation](https://api.drupal.org/api/drupal/includes%21database%21database.inc/function/db_insert/7.x) for `db_insert` says and I believe if you were to `dpm()` the return value that is what you get. Do we usually document this differently?

angaither: I don't think we have any other instances of an insert living alone in a function. 😂

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -924,3 +924,19 @@ function dosomething_helpers_get_current_campaign_run_for_user($nid, $user = NUL
...
 }
+
+/**
+ * Returns a base64 encoded data uri for an Image object
+ *
+ * @param  Image object $image
+ * @return string
+*/
+function dosomething_helpers_get_data_uri_from_image($image) {
+  $image_path = drupal_realpath($image->source);
```
sbsmith86, line 935: @angaither 🐍 💯

katiecrane:
@sbsmith86 your comments make this so easy to follow!! 🙌  I pulled down this branch and when I submit a reportback I'm not ending up on the permalink page, but if it's working for you I'm guessing I have some sort of configuration problem on my end

sbsmith86:
@katiecrane hmm. Is the reportback and item making it into the Rogue and Phoenix DB's?

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
sbsmith86: @angaither lol ah. Bad practice? I can not have it in it's own function.

katiecrane:
@sbsmith86 it's not making it into either DB 😞 when I hit submit it seems like it just refreshes the campaign page?

sbsmith86:
@katiecrane yea it seems something is up with your local phoenix posting to your local rogue. Has that been generally working for you?  Make sure all the correct settings are set here: `/admin/config/services/rogue`

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -371,24 +371,59 @@ function dosomething_reportback_form_submit($form, &$form_state) {
     $values['uid'] = $user->uid;
   }
 
-  // Save uploaded file.
-  if ($file = dosomething_reportback_form_save_file($form_state)) {
-    // Store new file's fid into values.
-    $values['fid'] = $file->fid;
-  }
+  if (module_exists('dosomething_rogue')) {
```
angaither: ooo, so this module does exist & is enabled in prod currently. so I think we want to add an additional check here to not have this stuff happen in prod!

sbsmith86: true! Updated to check for the flag we added in the custom settings form. This is per node for now. If we move forward with wanting this on all campaigns we should probably switch this to a more global feature flag.

angaither:
🚢

angaither:
hm, actually I think we should make this a global checkbox (since I think that's what we want to do)
So things will automatically work on staging for everything and we don't have to turn it on for each campaign.

sbsmith86:
@angaither cool. I can merge this and do a separate branch for that work?

PR #7103, opened 2016-10-04
Adds tracking for the onboarding finish button

itsjoekent:
#### What's this PR do?

We want to track how many people are clicking "finish" at the end of onboarding

Enable GA debugging
- Test the event is fired
- Test onboarding still closes out
#### Relevant tickets

Fixes https://github.com/DoSomething/TeamRocket/issues/30

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/js/components/SlideshowController.js
+++ b/lib/themes/dosomething/paraneue_dosomething/js/components/SlideshowController.js
@@ -1,3 +1,5 @@
+import { sendEvent } from '../utilities/Analytics';
+
 const React = require('react');
 
 /**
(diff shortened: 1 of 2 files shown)


lib/themes/dosomething/paraneue_dosomething/js/components/SlideshowController.js
```diff
@@ -1,14 +1,26 @@
...
  * <SlideshowController />
  */
 class SlideshowController extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.finish = this.finish.bind(this);
+  }
+
+  finish() {
```
DFurnes, line 16: Not something we need to do for this PR, but let's start thinking about naming conventions for methods like this. Maybe `onClose` for the prop (to indicate it's a function prop), and `handleClose` for event handler. Otherwise it feels like a bit of a hunt to figure out how things are wired up.

/cc @weerd

DFurnes (approved):
👍

lib/themes/dosomething/paraneue_dosomething/js/components/SlideshowController.js
itsjoekent: agreed! Would be nice to make it standard on all components I think so they're more extensible

Thought I just had -- Should these functions even be on the component itself, or passed in as a property? If the goal is to extend functionality, I'm leaning on passing the function in.

weerd: @DFurnes maybe we should start a further "React" subsection under the Javascript section on the code-style repo?

DFurnes: Yeah! 👍

ngjo: jumping in here cuz ive been eavesdropping! i agree that using more standard components moving forwarad, but if it's at all possible to sit down with whoever might be the creator of the existing events/know the most about them and document what they're tracking. I know it was a struggle when digging into event data for lose your v card

PR #7104, opened 2016-10-04
Fixes onboarding card layout to be more responsive

itsjoekent:
#### What's this PR do?
- Removes old testing 🎨 
- Adjusts margin around slide to match Luke's mock
- Sets height for the signup card tagline so they always match
- Adjusts beta card padding to be more responsive

Test in small, medium & large screens to verify the layout is correct
#### Relevant tickets

Fixes https://github.com/DoSomething/TeamRocket/issues/29

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
@@ -25,8 +25,16 @@
   }
 
   .slideshow__slide .container > .wrapper {
-    margin: 0 33px;
+    margin: 12px;
     width: initial;
+
+    @include media($medium) {
+      margin: 0 33px;
+    }
+
+    @include media($large) {
+      margin: 0 12.5%;
+    }
   }
 }
 
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
@@ -24,7 +24,7 @@
     color: #444;
   }
 
-  .wrapper.-compact {
+  .container.-compact > .wrapper {
     margin: 0 $base-spacing / 2;
     width: initial;
 
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
@@ -24,7 +24,7 @@
     color: #444;
   }
 
-  .slideshow__slide .container > .wrapper {
+  .wrapper.-compact {
     margin: 0 $base-spacing / 2;
     width: initial;
 
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
@@ -25,11 +25,11 @@
   }
 
   .slideshow__slide .container > .wrapper {
-    margin: 12px;
+    margin: 0 $base-spacing / 2;
     width: initial;
 
     @include media($medium) {
-      margin: 0 33px;
+      margin: 0 $base-spacing;
     }
 
     @include media($large) {
--- a/lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
+++ b/lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
@@ -8,100 +8,6 @@
     position: relative;
   }
 
-  &.-fullscreen {
-    position: fixed;
-    z-index: 1000;
-    top: 0;
-    left: 0;
-    width: 100%;
-    height: 100%;
-
-    &:before {
-      content: "";
-      display: block;
-      background: url('../logo.png');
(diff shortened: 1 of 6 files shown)


itsjoekent:
screen shot 2016-10-04 at 2 51 35 pm
screen shot 2016-10-04 at 2 51 47 pm
screen shot 2016-10-04 at 2 51 59 pm

lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
```diff
@@ -117,6 +23,19 @@
     border-radius: 4px;
     color: #444;
   }
+
+  .slideshow__slide .container > .wrapper {
+    margin: 12px;
```
DFurnes: Can we avoid using pixel widths here? It looks like the first `12px` is just `$base-spacing / 2`, but I'm pretty lost on the other ones. If they're based on our grid/spacing, use the appropriate mix-ins or variables. Otherwise, can you add a comment explaining the choice?

This should also be applied as a modifier on the `.container` pattern (like `.container.-compact` or something) rather than reaching in and mucking with it based on what element it's being held in).

lib/themes/dosomething/paraneue_dosomething/scss/patterns/_card.scss
```diff
@@ -63,6 +63,7 @@
     font-size: $font-small;
     color: $dark-gray;
     font-style: italic;
+    height: 60px;
```
DFurnes: Can you add a comment explaining that this is so all the cards have the same height?

lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
```diff
@@ -8,100 +8,6 @@
     position: relative;
   }
 
-  &.-fullscreen {
```
DFurnes, line 11: Yay for deleting old code! 🔥 👏

DFurnes (changes requested):
Left some comments.

lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
itsjoekent: The numbers for mobile/tablet are based on what Luke told me yesterday, I'll ask him in a sec if he wants to stick with this or switch to something more standard. (the 12.5% is what we currently have on prod)

lkpttn: We might be able to solve this with smaller margins on Medium sized devices than the 12.5% it currently takes. 33px is what it came out to but really we just want almost full width for tablets.

I don't know how much that goofs up the rest of tablet sized content, might be worth me taking a big old look at Forge for medium sizes.

itsjoekent: @lkpttn 24px looks fine on my computer, mind if we just switcheroo to that for now?

lkpttn: 👍

lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
```diff
@@ -24,12 +24,12 @@
     color: #444;
   }
 
-  .slideshow__slide .container > .wrapper {
-    margin: 12px;
+  .wrapper.-compact {
```
DFurnes: Can you switch this to `.container.-compact > .wrapper` like suggested in my original comment? We can't target modifiers on the wrapper itself because that's a common classname used in multiple components.

lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
```diff
@@ -117,6 +23,19 @@
     border-radius: 4px;
     color: #444;
   }
+
+  .wrapper.-compact {
```
itsjoekent: Renamed this to -compact (:thumbsup:) but unfortunately had to change the wrapper in every slide, in another PR I'd really like to pull,

```
      <div className={`slideshow__slide ${this.props.isActive ? '-active fade-in': ''}`}>
        <div className="container">
          <div className="wrapper -compact">
```

Out of every slide because it's just repeated for all of them & makes changing something like this quite annoying

Regardless ready for a quick 2nd look @DFurnes

DFurnes (changes requested):

lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
itsjoekent: Oh gotcha!

itsjoekent: I think this is now what you had in mind?

DFurnes (approved):

lib/themes/dosomething/paraneue_dosomething/scss/content/_onboarding.scss
DFurnes: Looks good! ✅

PR #7106, opened 2016-10-05
Adding updated text strings to a POT file for #6686

### LINKED ISSUE #6686: POEditor Refresh for Phoenix Language Updates
#### User Story

We have a couple of things we need to update across all global experiences. To do this, we need to add the string to the PO file and import into POEditor, then translate, export and import back into production.
#### Additional Information

Text Strings: 
###### Problem Share Prompt

Original US EN:
"F that, spread the word"

Updated US EN:
- [x] "Share this stat with friends"

We are going to leave the Portuguese and Spanish translations the same, even though the US EN language is changing.
###### Member Count Token

Original US EN: 

Updated US EN:

Instead of the member count on the site ticking up using a dynamic "member count" token, we'd like to set it at 5.3 million members.
###### Mic Drop

Original US EN: 
DoSomething.org makes the world suck less. One of the largest orgs for young people and social change, our millions of members tackle campaigns that impact every cause, from poverty to violence to the environment to literally everything else. Any cause, anytime, anywhere. _mic drop_ 

Updated US EN: 
- [x] DoSomething.org makes the world suck less. One of the largest orgs for young people and social change, our millions of members tackle campaigns that impact every cause, from poverty to violence to the environment to literally everything else. Any cause, anytime, anywhere.

We are removing "mic drop" from branding per the request from Ben here: https://trello.com/c/tN8pKBnB/1560-please-remove-mic-drop-from-mobile-and-my-entire-life

In POEditor, there is also simply the "*mic drop" string -- this should be removed as well.
#### Why This Matters

The original problem share prompt has been flagged a number of times by sponsors for language. We would like to change that language on the US EN site, as well as make some other brand updates.

weerd:
#### What's this PR do?

This PR addresses some updates to a couple strings. I added them as additional items into the POT file instead of modifying the existing ones so when imported we don't lose the translations on POEditor. 

👁‍🗨 

I'll add an additional issue to remove the old strings once the new ones are imported into POEditor, translated, exported from POEditor and imported into Prod CMS.
#### Relevant tickets

Fixes #6686
Refs https://github.com/DoSomething/TeamRocket/issues/2
#### Checklist
- [x] Upload POTs to POEditor

---

@DFurnes @angaither 

cc @ngjo

### DIFF
--- a/translations/pots/dosomething_configuration_variables.pot
+++ b/translations/pots/dosomething_configuration_variables.pot
@@ -105,6 +105,10 @@ msgstr ""
 msgid "F that, spread the word"
 msgstr ""
 
+#: dosomething_campaign_problem_share_prompt
+msgid "Share this stat with friends"
+msgstr ""
+
 #: dosomething_campaign_run_no_total_copy
 msgid "We’re still tallying the [label]. The bigger the number, the longer it takes...so stay tuned!"
 msgstr ""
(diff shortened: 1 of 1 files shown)


angaither:
🌐

PR #7113, opened 2016-10-07
Updating phoenix by adding more gateways

weerd:
#### What's this PR do?

This PR updates a few of the methods in the **dosomething_northstar.module** to replace the repeated `drupal_http_request()` with the new `dosomething_northstar_client()` which uses the **Gateway** package.

👀 
#### Relevant tickets

Fixes https://github.com/DoSomething/TeamRocket/issues/11
#### Checklist
- [ ] Tested on staging.

---

@DFurnes @deadlybutter 

cc: @ngjo @jessleenyc

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -61,7 +61,6 @@ function dosomething_northstar_menu() {
 /**
  * Verify the given credentials against Northstar, returning the authorized
  * user account if one exists.
- * @todo Needs to be deprecated.
  *
  * @param  array $credentials
  * @return array|null
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -261,7 +261,7 @@ function dosomething_northstar_transform_user($user, $password = null) {
  * Save the user's Northstar ID to their local profile field and the authmap.
  *
  * @param string $uid - Drupal user ID
- * @param object $northstar_response - Northstar user JSON response
+ * @param array $northstar_response - Northstar user JSON response
  */
 function dosomething_northstar_save_id_field($uid, $northstar_response) {
   $northstar_id = !empty($northstar_response['data']['id']) ? $northstar_response['data']['id'] : null;
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -117,8 +117,8 @@ function dosomething_northstar_create_user($user, $password = null) {
 /**
  * Update a user's existing Northstar profile based on their Drupal ID.
  *
- * @param $user - Drupal user object
- * @param $password - Unhashed password provided during registration.
+ * @param object $user - Drupal user object
+ * @param string $password - Unhashed password provided during registration.
  */
 function dosomething_northstar_update_user($user, $password = null) {
   $payload = dosomething_northstar_transform_user($user, $password);
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -126,22 +126,40 @@ function dosomething_northstar_update_user($user, $password = null) {
   $payload = dosomething_northstar_transform_user($user, $password);
 
   try {
-    $response = dosomething_northstar_client()->put('v1/users/drupal_id/' . $user->uid, $payload);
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -304,8 +292,8 @@ function dosomething_northstar_log_request($op, $user, $request_body, $response)
       'user_country' => $user->field_address[LANGUAGE_NONE][0]['country'],
       'fastly_country' => dosomething_settings_get_geo_country_code(),
       'request_values' => json_encode($request_body),
-      'response_code' => $response->code,
-      'response_values' => $response->data,
+      'response_code' => (isset($response['code']) ? $response['code'] : '200'),
```
DFurnes: This bit won't work anymore – this was using a Drupal HTTP response object-array-thing which had a `code` property for the status code and a `data` property with the actual response.

We should probably save this when we catch an exception (not sure we do this all that elegantly in Gateway at the moment, so might need a tiny update there), since we want to be able to see what kind of error it is.

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -62,96 +62,84 @@ function dosomething_northstar_menu() {
...
-  if (_dosomething_northstar_is_successful_response($response)) {
-    if (module_exists('stathat')) {
-      stathat_send_ez_count('drupal - Northstar - user updated - count', 1);
-    }
-  }
-  elseif ($response->code === '404') {
-    // If the given Drupal ID 404s, try to register them.
-    $password_or_null = ! empty($form_state['values']['pass']) ? $form_state['values']['pass'] : null;
-    dosomething_northstar_create_user($user, $password_or_null);
-  }
```
weerd, line 151: @DFurnes I removed this part. I could do something where if it catches an `Exception` then it `json_decodes` the message and grabs the error code and checks to see if it's a 404... but when I tried feeding the `put` request a bad ID, it didn't `404` it just returned `null`. So alternatively could try do a conditional w/ if `$response` is `null` then try to create the user. HOWEVER, there's something weird in the old original code above... `$form_state` doesn't exist in this function, nor is it passed in... it exists in the caller function that called this function `dosomething_user_update_user()`... so technically `$password_or_null` would always be `null`?? Seemed weird to me.

DFurnes (changes requested):
One comment, but the rest looks great!

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
DFurnes: Oh yep, that certainly seems like a mistake. Should have been using the `$password` parameter.

404s will return null (see [`send()`](https://github.com/DoSomething/gateway/blob/2de6ade1b724fc93353545996426ecee49b7cd7e/src/Common/RestApiClient.php#L216-L240) in RestApiClient) so checking for null seems fine to me. Other types of error codes should trigger their respective exceptions.

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
weerd: yeah, this was a bit weird. I had done this while I still had  some of the functions using the old approach, but now that I've updated the lot, can probs do something else. I'll see what I can come up with based on your suggestion. The error message has a code, I can try json_decoding it and then grabbing it for this log?

DFurnes: I don't believe we store it in a particularly useful manner at the moment (like a `$e->getCode()`), but it'd make sense to pull it from the exception and then pass along to the log function.

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -294,7 +313,7 @@ function dosomething_northstar_log_request($op, $user, $request_body, $response)
       'user_country' => $user->field_address[LANGUAGE_NONE][0]['country'],
       'fastly_country' => dosomething_settings_get_geo_country_code(),
       'request_values' => json_encode($request_body),
-      'response_code' => (isset($response['code']) ? $response['code'] : '200'),
+      'response_code' => $code,
       'response_values' => json_encode($response['data']),
```
DFurnes: This should probably just be `json_encode($response)` now.

weerd: ya, especially if we end up passing any meta-data in the future. Okie dokie

DFurnes (approved):

weerd:
@jessleenyc @ngjo @DFurnes 

Ok! I double-checked and ran a bunch of user-related tests w/ registration, logging in and user profile updates as well as double-checking the Phoenix-to-Northstar user export scripts to make sure stuff seemed cool and worked right (I didn't run the script, but double-checked the params passed should be working).

PR #7114, opened 2016-10-07
posting to rogue via gateway

katiecrane:
#### What's this PR do?

Instead of using `drupal_http_request` to send to Rogue, we now use a call to `post` on Gateway's `RestApiClient`.

Pull it down and make sure your local phoenix is posting to Rogue still! There shouldn't be any noticeable changes.
#### Checklist
- [ ] Tested on staging.


lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -86,9 +90,7 @@ function dosomething_rogue_send_reportback_to_rogue($values, $user = NULL) {
           'crop_rotate' => $values['crop_rotate'],
         ]),
   ];
-
-  $response = drupal_http_request($client['base_url'] . '/reportbacks', $options);
-
+  $response = post($client['base_url'] . '/reportbacks', $options);
```
weerd: so `post()` doesn't exist standalone like this. You would need to call it off the instantiated `$client` so something more like `$client->post(...)`

weerd: It might help if you reference what we're doing with Northstar in the **dosomethng_northstar.module** file here: https://github.com/DoSomething/phoenix/blob/db40d34f09c1856d3eb5193ff6ed5e8b71eb5e5b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module#L26

weerd: Actually, nevermind on the second comment... forgot you're only using the `RestApiClient`...

weerd: To coincide better with what we're doing in **dosomething_northstar.module** mind renaming this function `_dosomething_rogue_build_http_client()` to `dosomething_rogue_client()` and ignore the stupid underscore `_` prefixing that's a dumb Drupal standard 😉

weerd: So part of the issue too is that you're not actually instantiating the `RestApiClient` when building the client in https://github.com/DoSomething/phoenix/blob/db40d34f09c1856d3eb5193ff6ed5e8b71eb5e5b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module#L38 So you'll pretty much never have access to the `post()` method without it!

weerd: So you would need something more like this in the function:

``` php
function dosomething_rogue_client() {
  return new RestApiClient(ROGUE_API_URL . '/' . ROGUE_API_VERSION);
}
```

weerd: and don't forget to pop a `use` statement up top for including the class

``` php
use DoSomething\Gateway\Common\RestApiClient;
```

katiecrane: @weerd so helpful! 🎉  we have access to the post method now! still have some other problems, but that one is solved! and we got the use statement up top already 😉

weerd: Yay!! 💃

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -1,4 +1,8 @@
 <?php
+
+use DoSomething\Gateway\Contracts\TransactionBridgeContract;
```
chloealee: @katiecrane did we decide we didn't need this here and only needed 

```
use DoSomething\Gateway\Common\RestApiClient;
use DoSomething\Gateway\ForwardsTransactionIds;
```

per our [slack convo](https://dosomething.slack.com/archives/D0F16UE2D/p1475871108000074)?

katiecrane: @chloealee yeah I think you're right!

chloealee: @katiecrane cool, otherwise everything else seems right to me!

lib/app/Auth/PhoenixTransactionBridge.php
```diff
@@ -0,0 +1,28 @@
...
+      return drupal_get_http_header($name);
+  }
+
+  /**
+   * Write a log message.
+   *
+   * @return void
+   */
+  public function log($message, array $details) {
+      watchdog('Transaction', $details, [], WATCHDOG_INFO);
```
angaither: does this work as expected? I think the [args](https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/watchdog/7.x) are reversed? 

also, it might be useful to provide more info in the message, I think the standard we've been using is `(module_name, thing that happened)`

katiecrane: @angaither I actually should probably take this file out of this PR. This won't be used until we get the Transaction IDs to get added.

angaither: awwwyip!

katiecrane: which is coming next. @chloealee if you get to that before I hop back over, FYI ^

angaither:
you gots conflicts.

katiecrane:
@angaither that cat is SO CUTE and the conflicts are gone!

angaither:
:shield:

katiecrane:
@chloealee this is merged in FYI 😸

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -63,32 +59,27 @@ function dosomething_rogue_send_reportback_to_rogue($values, $user = NULL) {
...
+        'status' => isset($values['status']) ? $values['status'] : 'pending',
+        'crop_x' => $values['crop_x'],
+        'crop_y' => $values['crop_y'],
+        'crop_width' => $values['crop_width'],
+        'crop_height' => $values['crop_height'],
+        'crop_rotate' => $values['crop_rotate'],
   ];
 
-  $response = drupal_http_request($client['base_url'] . '/reportbacks', $options);
+  $response = $client->post($client->getBaseUri() . '/reportbacks', $data);
```
DFurnes, line 82: Sorry to comment late but just wanted to mention that you can omit the base URL here 😄

``` php
$response = $client->post('reportbacks', $data);
```

katiecrane: @DFurnes hmm what you have written there doesn't seem to work :/ but also my local is being screwy anyway so that could be why?

DFurnes: Hmm, strange! 😮 Happy take a look at some point.

katiecrane: @DFurnes thanks! I'll try it again once my local is working. Is the base URL just not necessary or should it not work at all with that in there?

DFurnes: It will still work if you include it each time, but providing a relative URL should automatically prepend the base URL specified when you created the client above (in theory! 😆)

katiecrane: @DFurnes cool! just wanted to make sure it wasn't full broken. thanks!

chloealee: @DFurnes @katiecrane worked for me too on my local but I will incorporate this change into the changes I'm making now (created a `Rogue` class in Phoenix in order to pull in the transaction id trait :))

PR #7117, opened 2016-10-07
Fixes taxonomy page showing duplicate campaigns

itsjoekent:
#### What's this PR do?
- Removes old method of retrieving campaigns for a given TID in favor of a single line drupal function

Do you see duplicate campaigns on the taxonomy pages?
screen shot 2016-10-07 at 12 15 36 pm

Also spot check the campaigns to make sure they have "donate something" applied (or whatever taxonomy you're testing on)

I wrote up the debug process in detail on this trello card: https://trello.com/c/zfLHlD6F/82-research-as-a-developer-i-want-the-know-why-the-volunteer-pages-image-section-is-broken#comment-57f7c76c34aca52c139159f5
#### Relevant tickets

Fixes N/A ?
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
+++ b/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
@@ -110,6 +110,7 @@ function dosomething_taxonomy_preprocess_taxonomy_term(&$vars) {
 
 /**
  * Returns array of node nid's with the given Term $tid.
+ * @deprecated Use taxonomy_select_nodes instead.
  *
  * @param int $tid
  *   A taxonomy term tid.
--- a/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
+++ b/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
@@ -117,14 +117,7 @@ function dosomething_taxonomy_preprocess_taxonomy_term(&$vars) {
  * @return array
  */
 function dosomething_taxonomy_get_campaigns($tid) {
-  $nids = array();
-  $view = views_get_view('campaigns_by_term');
-  $view->set_arguments(array($tid));
-  $view->execute();
-  foreach ($view->result as $record) {
-    $nids[] = $record->nid;
-  }
-  return $nids;
+  return taxonomy_select_nodes($tid, FALSE);
 }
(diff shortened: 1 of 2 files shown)


itsjoekent:
[RESEARCH: As a developer, I want the know why the /volunteer pages image section is broken](https://trello.com/c/zfLHlD6F/82-research-as-a-developer-i-want-the-know-why-the-volunteer-pages-image-section-is-broken)

lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
```diff
@@ -117,14 +117,7 @@ function dosomething_taxonomy_preprocess_taxonomy_term(&$vars) {
...
 function dosomething_taxonomy_get_campaigns($tid) {
-  $nids = array();
-  $view = views_get_view('campaigns_by_term');
-  $view->set_arguments(array($tid));
-  $view->execute();
-  foreach ($view->result as $record) {
-    $nids[] = $record->nid;
-  }
-  return $nids;
+  return taxonomy_select_nodes($tid, FALSE);
```
DFurnes, line 121: Will this return non-campaign nodes assigned to that taxonomy? Is that a problem if so?

DFurnes (approved):
One question, but looks good to me! 🔥⚓️

lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
itsjoekent: Good point! Personally I'm not sure but I never noticed it when testing. Even if it does I think its a better tradeoff in comparison to what we have now

weerd (commented):
Looks cool. Maybe add a `@deprecated` and indicate in the docblock for `dosomething_taxonomy_get_campaigns()` that devs should favor the use of `taxonomy_select_nodes()` instead and that this is just to keep compatibility. So future code (hopefully not any since moving off of this codebase) should just default to using `taxonomy_select_nodes()` instead.

itsjoekent:
@weerd 👍  one sec & ill have it up!

itsjoekent:
@weerd okie dokes that should do the trick

weerd (approved):
👍

PR #7118, opened 2016-10-07
Failures

### LINKED ISSUE #6959: Post reportback errors to a failed task log
#### User Story

if a reportback fails posting to rogue, log that, and then re-try to send that request

this could either be a sql table (a-la gladiator) or a queue (a la message broker requests)
#### Why This Matters

safety first

sbsmith86:
#### What's this PR do?

Builds on @chloealee's work on handling failures when sending reportbacks to rogue. 
- creates a `dosomething_rogue_failed_task_log` table that holds failed reportbacks
- Stores failed reportbacks in this table
- Creates a cron job that will retry sending the failed reportbacks to rogue. 

Try to send a reportback without rogue running and it should get stored in the failed table. 

Turn rogue back on and run cron, the reportback should be sent to rogue successfully (or restored in the table if it fails again.)
#### Relevant tickets

Fixes #6959 
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install
@@ -88,6 +88,9 @@ function dosomething_rogue_schema() {
         'not null' => FALSE,
       ],
       'file' => [
+        'description' => 'The file string to save of the reportback image.',
+        'type' => 'text',
+        'not null' => TRUE,
       ],
       'caption' => [
         'description' => 'File caption.',
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install
@@ -113,7 +113,7 @@ function dosomething_rogue_schema() {
         'length' => '500000',
         'not null' => FALSE,
       ],
-    ]
+    ],
     'primary key' => ['id'],
   ];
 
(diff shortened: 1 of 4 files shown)


chloealee:
@sbsmith86 My local env is still really messed up so unfortunately, I can't actually test, but from looking at the code, all looks good to me!

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install
```diff
@@ -47,6 +47,72 @@ function dosomething_rogue_schema() {
...
+        'description' => 'Why the user participated in the campaign.',
+        'type' => 'text',
+        'length' => '2048',
+        'not null' => FALSE,
+      ],
+      'file' => [
+        'description' => 'The file string to save of the reportback image.',
+        'type' => 'text',
+        'length' => '500000',
+        'not null' => TRUE,
```
katiecrane: I think this should be nullable in case it's an update

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install
```diff
@@ -47,6 +47,72 @@ function dosomething_rogue_schema() {
...
+      'quantity' => [
+        'description' => 'The quantity of reportback_nouns reportback_verbed.',
+        'type' => 'int',
+        'not null' => TRUE,
+      ],
+      'why_participated' => [
+        'description' => 'Why the user participated in the campaign.',
+        'type' => 'text',
+        'length' => '2048',
+        'not null' => FALSE,
```
katiecrane: I believe the why is required so should have to be not null

sbsmith86:
@katiecrane updated

katiecrane:
@sbsmith86 I think it looks good then, but I don't think I've ever touched cron stuff before, so might want some other 👀 on it too

PR #7119, opened 2016-10-12
Only pull in campaigns from taxonomy get function

itsjoekent:
#### What's this PR do?

Only pull campaigns in the taxonomy get function, not all content types

Visit a taxonomy page, check the only content is campaigns

Wasnt happening locally, Jen saw it happening on thor (so blocking deploy now). Wasnt able to reproduce locally until i saved an image, very strange...
#### Relevant tickets

Fixes #no

### DIFF
--- a/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
+++ b/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
@@ -119,7 +119,7 @@ function dosomething_taxonomy_preprocess_taxonomy_term(&$vars) {
 function dosomething_taxonomy_get_campaigns($tid) {
   $nids = taxonomy_select_nodes($tid, FALSE);
   $campaigns = [];
-  foreach ($nids as $delta => $nid) {
+  foreach ($nids as $nid) {
     if (node_load($nid)->type === 'campaign') {
       $campaigns[] = $nid;
     }
--- a/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
+++ b/lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
@@ -117,13 +117,9 @@ function dosomething_taxonomy_preprocess_taxonomy_term(&$vars) {
  * @return array
  */
 function dosomething_taxonomy_get_campaigns($tid) {
-  $nodes = node_load_multiple(taxonomy_select_nodes($tid, FALSE));
-  $campaigns = [];
-  foreach ($nodes as $node) {
-    if ($node->type === 'campaign') {
-      $campaigns[] = $node->nid;
-    }
-  }
+  $nodes = node_load_multiple(taxonomy_select_nodes($tid, FALSE), ['type' => 'campaign']);
+  $campaigns = array_map(function($node) { return $node->nid; }, $nodes);
+
   return $campaigns;
 }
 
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
```diff
@@ -110,15 +110,21 @@ function dosomething_taxonomy_preprocess_taxonomy_term(&$vars) {
...
  * @param int $tid
  *   A taxonomy term tid.
  *
  * @return array
  */
 function dosomething_taxonomy_get_campaigns($tid) {
-  return taxonomy_select_nodes($tid, FALSE);
+  $nids = taxonomy_select_nodes($tid, FALSE);
+  $campaigns = [];
+  foreach ($nids as $delta => $nid) {
```
weerd: You can probably remove the `$delta` here since you're not using it. Just leave it as `foreach ($nids as $nid) { ... }`

weerd (changes requested):

weerd (approved):
👍

lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
```diff
@@ -110,15 +110,21 @@ function dosomething_taxonomy_preprocess_taxonomy_term(&$vars) {
...
+    if (node_load($nid)->type === 'campaign') {
```
DFurnes: This will make a ton of database queries since it needs to first run the query for `taxonomy_select_nodes` and then independently load each node from the DB. A quick improvement would be to use [`node_load_multiple`](https://api.drupal.org/api/drupal/modules%21node%21node.module/function/node_load_multiple/7.x).

DFurnes (changes requested):

lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
itsjoekent: ah good point, I'll swap that out

weerd: oh jes! good call @DFurnes Drupal does do some caching of nodes on it's own for speed, but yeah, `node_load_multiple` would be more ideal.

itsjoekent:
@DFurnes updated, I think this is what you had in mind

lib/modules/dosomething/dosomething_taxonomy/dosomething_taxonomy.module
```diff
@@ -110,15 +110,21 @@ function dosomething_taxonomy_preprocess_taxonomy_term(&$vars) {
...
- * @deprecated Use taxonomy_select_nodes instead.
  *
  * @param int $tid
  *   A taxonomy term tid.
  *
  * @return array
  */
 function dosomething_taxonomy_get_campaigns($tid) {
-  return taxonomy_select_nodes($tid, FALSE);
+  $nodes = node_load_multiple(taxonomy_select_nodes($tid, FALSE));
```
angaither: if you're looking for a specific type of node (it seems we are here) you can add a `type` to `node_load_multiple` to only pull campaign nodes

itsjoekent: @angaither where do we add a type? The docs list the conditions as deprecated https://api.drupal.org/api/drupal/modules%21node%21node.module/function/node_load_multiple/7.x

angaither: @deadlybutter that note is for d8

itsjoekent:
@angaither got it down to two lines, still had to extract the NID's from the query

itsjoekent:
hm, i should probably flip those variable names now

angaither:
🚀

PR #7121, opened 2016-10-12
OpenID Connect: Add ability to queue post-authorization actions.

### LINKED ISSUE #7122: Allow campaign signups as a result of OpenID Connect logins.
#### User Story

As an engineer, I want campaign sign ups to work when authenticating through northstar so that I can deploy northstar auth to prod.

Currently, OpenID Connect login always redirects you to your user profile. This is usually what we want, but for campaign signups we "bundle" the signup action alongside logging in. So once the user is redirected back from Northstar, we want to automatically sign them up for the right campaign.
#### Why This Matters

We need this to ship OpenID Connect authentication on prod.

No.

DFurnes:
#### What's this PR do?

This pull request adds the ability to queue up "post-authorization" actions (like signing up for a campaign) when using the OpenID Connect auth flow. This allows us to keep the same behavior for the "Sign Up" button on the pitch page.

With yer eyes. 👀 This refactors a fair amount of login/register code so definitely worth giving a careful look to everything. I tried to test as many flows as possible, but always good to get a second set of eyes.

I renamed the `dosomething_user_get_login_modal_link_markup` function because it's more about signup links than modal links (especially since signups don't necessarily make a modal now). ~~I also removed a bunch of code that was used for theming anonymous confirmation pages because it didn't seem necessary...?~~
#### Relevant tickets

Fixes #7122.
#### Checklist
- [ ] Tested on staging.
- [x] Pinged a PM if this is a larger PR that would benefit from some additional testing love

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.pages.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.pages.inc
@@ -33,7 +33,7 @@ function dosomething_campaign_reportback_confirmation_page($node) {
   $image_size = '740x480';
 
 
-  // If SMS player is anonymous, he button should open the login/register form.
+  // If SMS player is anonymous, the button should open the login/register form.
   if (user_is_anonymous()) {
     $button_text = t(variable_get('dosomething_campaign_confirmation_page_anon_button_text'));
     $vars['call_to_action'] = t(variable_get('dosomething_campaign_confirmation_page_anon_cta_text'));
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.pages.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.pages.inc
@@ -32,15 +32,29 @@ function dosomething_campaign_reportback_confirmation_page($node) {
   // Image size to pass to campaign block.
   $image_size = '740x480';
 
-  // Set button to link to /campaigns.
-  $button_text = t(variable_get('dosomething_campaign_confirmation_page_button_text'));
-  $vars['more_campaigns_link'] = l(t($button_text), 'campaigns', [
-    'attributes' => [
-      'class' => [
-        'button'
+
+  // If SMS player is anonymous, he button should open the login/register form.
+  if (user_is_anonymous()) {
+    $button_text = t(variable_get('dosomething_campaign_confirmation_page_anon_button_text'));
+    $vars['call_to_action'] = t(variable_get('dosomething_campaign_confirmation_page_anon_cta_text'));
+
+    $vars['more_campaigns_link'] = l(t($button_text), '#', [
+      'attributes' => [
+        'class' => ['button'],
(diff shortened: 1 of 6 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.pages.inc
```diff
@@ -31,37 +31,27 @@ function dosomething_campaign_reportback_confirmation_page($node) {
   ];
   // Image size to pass to campaign block.
   $image_size = '740x480';
-  // Signup source to pass to campaign block.
-  $source = dosomething_campaign_get_confirmation_path($node->nid);
 
-  // Set variables for anonymous user:
-  if (!user_is_logged_in()) {
```
weerd, line 38: @DFurnes I found this [gem](https://github.com/DoSomething/phoenix/blob/ec89a8484f65841358bccd50042b449ffaa34597/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module#L404). Seems there _is_ (?) a scenario when an anonymous user would hit the confirmation page...

weerd (changes requested):
⚖

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.pages.inc
DFurnes: Ah, that'd explain it. 😩

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.pages.inc
```diff
@@ -31,37 +31,41 @@ function dosomething_campaign_reportback_confirmation_page($node) {
...
+
+  // If SMS player is anonymous, he button should open the login/register form.
```
weerd: typo `the button...`

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
```diff
@@ -49,13 +90,52 @@ function dosomething_northstar_openid_connect_post_authorize($tokens, $account,
...
+  global $user;
+
+  if (! isset($_SESSION['openid_connect_action'])) {
+    return;
+  }
+
+  $actions = $_SESSION['openid_connect_action'];
+  foreach ($actions as $action => $target) {
+    if($action === 'signup') {
+      dosomething_signup_user_signup((int) $target, $user, 'lolol'); // @TODO lololol
```
weerd: set a `source`?

weerd: although could be a fun easter egg for the data team :trollface:

DFurnes: 😈

weerd (approved):
Our work is awesome. 💃

DFurnes:
🎉

Issue #7123, opened 2016-10-14
checking on status of posting to Rogue with Gateway


katiecrane:
#### Current Behavior

When submitted a reportback to Rogue, we currently check the code of the response that we get back from Gateway. However, what gets returned back to us is not the full response but is just the reportback. So, when we try to get the response code, we get nothing, which makes us think that the task failed. The reportback ends up in the normal table AND the failed task table, and the user does not end up on the permalink page, even though their reportback was successfully posted.
#### Desired Behavior

We don't want reportbacks that were posted successfully in the failed task log, and we want users to hit the permalink page if their reportback was successfully posted.
#### Steps to Replicate

Turn on Rogue and submit a reportback.

ngjo:
![](https://github.trello.services/images/mini-trello-icon.png) [As a developer, I want to check for Rogue failures in a way that works with the Gateway response.](https://trello.com/c/Mq96cp5X/108-as-a-developer-i-want-to-check-for-rogue-failures-in-a-way-that-works-with-the-gateway-response)

PR #7127, opened 2016-10-17
Transaction id trait

### LINKED ISSUE #7126: Use ForwardsTransactionId Trait when sending RBs/RB Items from Phoenix->Rogue via Gateway
#### Current Behavior

-Currently, Phoenix is using Gateway to make requests from Phoenix to Rogue to send reportbacks. It also has the foundation to use the `ForwardsTransactionIds` trait however is not using it because it needs a new class to bring in this trait. 
- Phoenix is not logging transaction Ids properly because it is not using trait correctly.
#### Desired Behavior
- Phoenix to use `ForwardsTransactionIds` trait to log transaction ids and keep track of requests. 
#### Steps to Replicate
- Create a reportback.
- Check dblogs and see there is no log of `Request made.` for transaction ids.
#### Why This Matters
- In order to properly track transaction ids, we want Phoenix to create this in the header and start the trace in Phoenix.

chloealee:
#### What's this PR do?

Creates a new `Rogue` class to pull in the `ForwardsTransactionIds` trait to successfully log transaction Id requests from Phoenix -> Rogue. 
1. Clear your dblogs. 
2. Submit a reportback. 
3. Refresh your dblogs and you should see the following:
   screen shot 2016-10-17 at 2 24 00 pm

Question - how can we access the `$details` array that we send to watchdog as the third param? 

By looking in SequelPro, it is confirmed that the reportback and reportback items are successfully being sent to Rogue however I noticed the "reportback not migrated to Rogue" error in the dblogs. I believe this is something that will be fixed with @katiecrane's work with [this issue](https://github.com/DoSomething/phoenix/issues/7123) since it looks like it [logs](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module#L108) it if the [status code is not `200` or `201`](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module#L84).
#### Relevant tickets

Fixes #7126
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/includes/Rogue.php
+++ b/lib/modules/dosomething/dosomething_rogue/includes/Rogue.php
@@ -14,16 +14,6 @@ class Rogue extends RestApiClient {
   protected $transactionBridge = Phoenix\Auth\PhoenixTransactionBridge::class;
 
   /**
-   * Create a new Rogue API client.
-   */
-  // public function __construct() {
-  //   $url = ROGUE_API_URL . '/' . ROGUE_API_VERSION . '/';
-
-  //   parent::__construct($url);
-  // }
-
-
-  /**
    * Send a POST request of the reportback to be saved in Rogue.
    *
    * @param string $baseurl
@@ -44,7 +34,7 @@ class Rogue extends RestApiClient {
    * @return object|false
    */
   public function updateReportback($baseurl, $data) {
-    $response = $this->put($url . $baseurl, $data);
+    $response = $this->put(ROGUE_API_URL . '/' . ROGUE_API_VERSION . '/' . $baseurl, $data);
 
     return is_null($response) ? null : $response;
   }
--- a/lib/modules/dosomething/dosomething_rogue/includes/Rogue.php
+++ b/lib/modules/dosomething/dosomething_rogue/includes/Rogue.php
@@ -6,5 +6,7 @@ use DoSomething\Gateway\ForwardsTransactionIds;
 class Rogue extends RestApiClient {
   use ForwardsTransactionIds;
 
-
+  public function postReportback($url, $data) {
+    return $this->post($url, $data);
+  }
 }
--- a/lib/modules/dosomething/dosomething_rogue/includes/Rogue.php
+++ b/lib/modules/dosomething/dosomething_rogue/includes/Rogue.php
@@ -16,11 +16,11 @@ class Rogue extends RestApiClient {
   /**
    * Create a new Rogue API client.
    */
-  public function __construct() {
-    $url = ROGUE_API_URL . '/' . ROGUE_API_VERSION . '/';
+  // public function __construct() {
+  //   $url = ROGUE_API_URL . '/' . ROGUE_API_VERSION . '/';
 
-    parent::__construct($url);
-  }
+  //   parent::__construct($url);
+  // }
 
 
   /**
@@ -31,7 +31,8 @@ class Rogue extends RestApiClient {
    * @return object|false
    */
   public function postReportback($baseurl, $data) {
-    $response = $this->post($baseurl, $data);
+    // die(ROGUE_API_URL . '/' . ROGUE_API_VERSION . '/' . $baseurl);
+    $response = $this->post(ROGUE_API_URL . '/' . ROGUE_API_VERSION . '/' . $baseurl, $data);
 
     return is_null($response) ? null : $response;
   }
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_rogue/includes/Rogue.php
```diff
@@ -0,0 +1,51 @@
...
+   * Send a POST request of the reportback to be saved in Rogue.
+   *
+   * @param string $baseurl
+   * @param array $data
+   * @return object|false
+   */
+  public function postReportback($baseurl, $data) {
+    $response = $this->post(ROGUE_API_URL . '/' . ROGUE_API_VERSION . '/' . $baseurl, $data);
+
+    return is_null($response) ? null : $response;
```
DFurnes: What is this ternary used for?

lib/modules/dosomething/dosomething_rogue/includes/Rogue.php
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * Send a POST request of the reportback to be saved in Rogue.
+   *
+   * @param string $baseurl
+   * @param array $data
+   * @return object|false
+   */
+  public function postReportback($baseurl, $data) {
+    $response = $this->post(ROGUE_API_URL . '/' . ROGUE_API_VERSION . '/' . $baseurl, $data);
```
DFurnes: You shouldn't need to specify `ROGUE_API_URL . '/' . ROGUE_API_VERSION` here because those are set as the base URL when you create the client. I'm also a bit confused by the `$baseurl` parameter here... is there ever a case where we'd want `postReportback` to post to a different path than `reportbacks`?

DFurnes (changes requested):

lib/modules/dosomething/dosomething_rogue/includes/Rogue.php
chloealee: @DFurnes I remember you saying that on one of Katie's old PRs but couldn't get it to work taking that out and just sending `'reportbacks'` along :-/ That is a good point though that we can just hardcode `/reportbacks` instead of sending the param here!

lib/modules/dosomething/dosomething_rogue/includes/Rogue.php
chloealee: @DFurnes You're right, I can just `return $response!`

lib/modules/dosomething/dosomething_rogue/includes/Rogue.php
DFurnes: That's so weird. Can you try doing a `var_dump($uri); die();` below line 115 in `vendor/guzzlehttp/guzzle/src/Client.php` when you omit the base URL here? Maybe that would give us a clue as to why it's not working. 🕵️

chloealee: @DFurnes weird! It seems to know what it is: 

```
/var/www/dev.dosomething.org/vendor/guzzlehttp/guzzle/src/Client.php:116:
object(GuzzleHttp\Psr7\Uri)[327]
  private 'scheme' => string 'http' (length=4)
  private 'userInfo' => string '' (length=0)
  private 'host' => string 'rogue.dev' (length=9)
  private 'port' => int 8000
  private 'path' => string '/api/reportbacks' (length=16)
  private 'query' => string '' (length=0)
  private 'fragment' => string '' (length=0)
```

chloealee: Also, when I `var_dump($url); die():` [above this line](https://github.com/DoSomething/gateway/blob/master/src/Common/RestApiClient.php#L44), it also is away of the `$url`: 

```
/var/www/dev.dosomething.org/vendor/dosomething/gateway/src/Common/RestApiClient.php:44:string 'http://rogue.dev:8000/api/v1' (length=28)
```

DFurnes: Oh, it's discarding the `v1/` part of the URL in the first example you posted.

It looks like you need a trailing slash on the base URI (provided in the constructor) or else it'll just discard any included path in the base URI... which is really confusing! Check out the [third example for `base_uri`](http://docs.guzzlephp.org/en/latest/quickstart.html#creating-a-client) from the docs. :rage1:

chloealee: ah, brilliant! thank you, @DFurnes ! It is updated now :)

DFurnes (approved):
Looks good to me! 🚀

PR #7130, opened 2016-10-18
don't require caption on RB update

### LINKED ISSUE #7128: 'Submit' button broken when updating a reportback
#### Current Behavior
- When updating a reportback (either the quantity or why participated fields), the 'Submit' button is broken (does not save changes, does not update in Rogue, stays on page). 
#### Desired Behavior
- This should redirect to the permalink page, update in the Phoenix database, and send the update to Rogue. 
#### Steps to Replicate
- Create a reportback for a campaign if not already created.
- Update the quantity. 
- Click the 'Submit' button.
- Page will not do anything (except the page might jump a bit), database will not update, and will not be sent to Rogue.
#### Why This Matters
- Broken reportback process.

katiecrane:
#### What's this PR do?

Unset `['#attributes']['data-validate']` and `['#required']` on the reportback caption if it is an update to a reportback. Members may submit reportback updates without images, so it doesn't make sense to always require captions.

Submit a reportback. Make sure it won't let you submit a reportback without an image OR an image without a caption. Make sure you hit the permalink page.
Update your reportback, either the why or the quantity or both! Make sure you hit the permalink page.
Try to update your reportback with an image but no caption. It shouldn't let you.

The caption validation doesn't seem to be anything new, weird that this ever worked. Shoutout to @weerd and @DFurnes for spelunking with me at the end of the day yesterday!
#### Relevant tickets

Fixes #7128
#### Checklist
- [x] Tested on staging.
- [x] Pinged a PM if this is a larger PR that would benefit from some additional testing love (cc: @ngjo)


katiecrane:
The fix here doesn't seem to work with rogue. I'm actually getting a `422` trying to update a RB with no image when posting to rogue.

katiecrane:
This plays nice with rogue now!

sbsmith86:
looks fine to me. Did it work before because we were pre populating the form with values from the last submission?

katiecrane:
@sbsmith86  It never should have worked because we were never unsetting the stuff that made the `caption` be required and it doesn't look like anything relating to the caption has changed since the initial fix. So it's hard to say why it was working before 😬

katiecrane:
@sbsmith86 does that mean I should merge or should I get more 👀 ?

sbsmith86:
Merge it!

PR #7132, opened 2016-10-18
Fix issues with export scripts.

DFurnes:
#### What's this PR do?

This fixes a bug with the export script that I noticed when restarting it.

The Northstar module now expects responses to be associate arrays (which is what we return from Gateway clients, rather than `stdClass` objects) so the "save ID field" was erring out when it got a stdClass instead.

This just updates those scripts to follow the rules (while still using old HTTP client since they're not going to be around that much longer).

👀

I tested that these fixes work correctly when running this script on `drupal.admin`.
#### Relevant tickets

Fixes 🐛.
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/scripts/delete-bad-users-in-northstar.php
+++ b/scripts/delete-bad-users-in-northstar.php
@@ -38,14 +38,15 @@ foreach ($records as $record) {
     'data' => json_encode($ns_user),
   ]);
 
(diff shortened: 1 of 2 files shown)


angaither:
🚢 
love those arrays

PR #7138, opened 2016-10-20
#7137: Passes class parameter as array

mshmsh5000:
#### What's this PR do?

Passing `$class` as an array when invoking the [new login helper](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_user/dosomething_user.theme.inc#L35).
#### How should this be reviewed?

Calls to the login helper should no longer spawn errors like

```
Error: Uncaught exception 'Error' with message '[] operator not supported for strings' in /var/www/www.dosomething.org/releases/20161019170210/html/includes/common.inc:2494
in l called at /var/www/www.dosomething.org/releases/20161019170210/lib/modules/dosomething/dosomething_user/dosomething_user.theme.inc (40)
[...]
```
#### Relevant tickets

Fixes #7137

PR #7139, opened 2016-10-20
Update default params and typehint for other link function.

DFurnes:
#### What's this PR do?

This makes some more updates based on the changes in #7138: making sure the class parameter to the register link helper is also an array, and updating the default parameters if omitted.

👀

🍃 
#### Relevant tickets

References #7138.
#### Checklist
- [ ] Tested on staging.


weerd (approved):
👍

PR #7141, opened 2016-10-20
Prefer Northstar ID whenever possible and cover some bases, huzzah!

weerd:
#### What's this PR do?

This PR does a few updates to fix up an issue occurring on Staging. It defaults to referring to the Northstar ID for a user when working with their record in Phoenix and when querying Northstar for more information on them. (see background context below!)

When using OpenID to register as a new user via Northstar, Northstar creates a new record in its system and then passes the information over to Phoenix to then create a new Drupal user in its system and associating the NS ID created on the Northstar side. However, we don't then go back and tell Northstar what the `drupal_id` for this new user is, which means that all the functions that later go grab Northstar info for this user, and ask to find the user using their `drupal_id` will return `null` because that id was never passed into the NS system.
#### Relevant tickets

Fixes # https://github.com/DoSomething/northstar/issues/458
#### Checklist
- [ ] Tested on staging.

---

@DFurnes @angaither 

cc: @jessleenyc

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
@@ -718,7 +718,7 @@ function dosomething_helpers_get_current_user($user = NULL, $asArray = FALSE) {
   }
 
   // Authenticated User
-  $northstar_user = dosomething_northstar_get_user($user->uid);
+  $northstar_user = dosomething_northstar_get_user($user->uid, 'drupal_id');
 
   if ($northstar_user && $asArray) {
     return dosomething_helpers_get_user_as_array($northstar_user);
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
@@ -84,6 +84,7 @@ function dosomething_northstar_openid_connect_claims_alter(&$claims) {
  */
 function dosomething_northstar_openid_connect_post_authorize($tokens, $account, $userinfo, $provider) {
   $edit = [];
+
   dosomething_user_set_fields($edit, [
     'access_token' => $tokens['access_token'],
     'refresh_token' => $tokens['access_token'],
@@ -95,11 +96,22 @@ function dosomething_northstar_openid_connect_post_authorize($tokens, $account, 
     _dosomething_user_send_to_message_broker();
   }
 
+  // Make sure the user's Northstar ID is set in the field_data_field_northstar_id table.
+  if (!dosomething_user_get_northstar_id($account->uid)) {
+    dosomething_northstar_save_id_field($account->uid, ['data' => $userinfo]);
+  }
+
   // If we saved a "post-authorization" action, do it now.
   dosomething_northstar_perform_authorization_actions();
 
   // Let's remember that this is a OpenID Connect session.
   $_SESSION['DOSOMETHING_NORTHSTAR_OPENID_CONNECT'] = true;
+
+  // Determine the user's language code if possible.
+  $language_code = dosomething_settings_get_geo_country_code();
+
+  $edit['language'] = $language_code ? $language_code : 'en-global';
+  $edit['timezone'] = 'UTC';
 
   user_save($account, $edit);
 }
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
@@ -109,7 +109,7 @@ function dosomething_northstar_openid_connect_post_authorize($tokens, $account, 
 
   if (empty($account->language)) {
     // Determine the user's language code if possible.
-    $language_code = dosomething_settings_get_geo_country_code();
+    $language_code = dosomething_global_get_language($account);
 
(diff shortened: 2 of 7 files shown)


lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
```diff
@@ -84,18 +84,35 @@ function dosomething_northstar_openid_connect_claims_alter(&$claims) {
...
+
   // If we saved a "post-authorization" action, do it now.
   dosomething_northstar_perform_authorization_actions();
 
   // Let's remember that this is a OpenID Connect session.
   $_SESSION['DOSOMETHING_NORTHSTAR_OPENID_CONNECT'] = true;
 
+  if (empty($account->language)) {
+    // Determine the user's language code if possible.
+    $language_code = dosomething_settings_get_geo_country_code();
```
angaither: I think you should be using `dosomething_global_get_language` here

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -718,7 +718,7 @@ function dosomething_helpers_get_current_user($user = NULL, $asArray = FALSE) {
   }
 
   // Authenticated User
-  $northstar_user = dosomething_northstar_get_user($user->uid);
+  $northstar_user = dosomething_northstar_get_user($user->uid, 'drupal_id');
```
angaither, line 721: `drupal_id` is already the default in that function ;) unless there's an update to come which changes this...

weerd: @angaither [Update coming!](https://github.com/DoSomething/phoenix/pull/7141/files#diff-6fba28391e7b23bdda9b664d3a71c59fR183) You're too quick! (it's why I had the "work in progress" in description header... just had a bunch of stuff changing and wanted to push up commits bit by bit so I don't lose track 😬 )

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
weerd: I can give that a whirl... wasn't sure since the user isn't anonymous, just doesn't have a `language` set and it's coming from a redirect from Northstar.

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
angaither: SORRY I CAN'T READ "WIP"

weerd: :trollface:

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
```diff
@@ -95,12 +96,28 @@ function dosomething_northstar_openid_connect_post_authorize($tokens, $account,
...
+
   // If we saved a "post-authorization" action, do it now.
   dosomething_northstar_perform_authorization_actions();
 
   // Let's remember that this is a OpenID Connect session.
   $_SESSION['DOSOMETHING_NORTHSTAR_OPENID_CONNECT'] = true;
 
+  if (empty($account->language)) {
+    // Determine the user's language code if possible.
+    $language_code = dosomething_global_get_language($account);
```
weerd, line 112: @angaither yep, your suggestion did work better 😄

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -180,12 +180,23 @@ function dosomething_northstar_get_error_code($errorMessage) {
  * @param  string $type
  * @return DoSomething\Gateway\Resources\NorthstarUser|null
  */
-function dosomething_northstar_get_user($id, $type = 'drupal_id') {
+function dosomething_northstar_get_user($id, $type = 'id') {
+  // Prefer using user's Northstar ID.
```
DFurnes, line 184: I like this approach. 👍

DFurnes (approved):
Looks good to me! 🚢

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -1337,13 +1337,28 @@ function dosomething_user_get_field_northstar_id($field_value) {
   return $field_value;
 }
 
+/**
+ * Return the specified Drupal user/global user's Northstar ID.
+ *
+ * @param  string $drupal_id
+ * @return mixed
+ */
+function dosomething_user_get_northstar_id($drupal_id = NULL) {
```
angaither, line 1346: Hmm post merge comment
But this function should be a handler or 'dosomething_user_get_field'
Not a stand alone function

weerd: It is using `dosomething_user_get_field()`. Based on where this function is being called/used, we either have the user object, or just the user's drupal id, so this is a small helper to deal with then properly calling the `dosomething_user_get_field()` with the required params.

jessleenyc:
[Users registered via OpenID Connect flow do not have `drupal_id` set on their Northstar profile.](https://trello.com/c/Mcrg7nCU/173-users-registered-via-openid-connect-flow-do-not-have-drupal-id-set-on-their-northstar-profile)

PR #7143, opened 2016-10-20
Send transactionals if user registered via OpenID Connect flow.

DFurnes:
#### What's this PR do?

This PR has a wonderfully hacky solution for sending transactional emails for OpenID Connect registrations. Since we're not told whether the user is newly registered or not, comparing the `created_at` timestamp seems to work nicely.

👀

This is a short-term fix, since we want these transactional messages to be triggered from Northstar once Quicksilver API is ready for production use.
#### Relevant tickets

Fixes ⁉️.
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
@@ -90,6 +90,11 @@ function dosomething_northstar_openid_connect_post_authorize($tokens, $account, 
     'access_token_expiration' => $tokens['expire'],
   ]);
 
+  // HACK: Is this account newly created (last 15s)? If so, send email.
+  if ($userinfo['created_at'] > time() - 15) {
+    _dosomething_user_send_to_message_broker();
+  }
+
   // If we saved a "post-authorization" action, do it now.
   dosomething_northstar_perform_authorization_actions();
 


DFurnes:
[(8) As a developer, I want all transactional e-mails triggered when authenticating via northstar so that the user gets all proper messaging.](https://trello.com/c/gzZD5guM/125-8-as-a-developer-i-want-all-transactional-e-mails-triggered-when-authenticating-via-northstar-so-that-the-user-gets-all-proper-m)

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
```diff
@@ -90,6 +90,11 @@ function dosomething_northstar_openid_connect_post_authorize($tokens, $account,
     'access_token_expiration' => $tokens['expire'],
   ]);
 
+  // HACK: Is this account newly created (last 15s)? If so, send email.
+  if ($userinfo['created_at'] > time() - 15) {
```
angaither, line 94: I like what you've done with the place

angaither:
I feel very shocked that this isn't a thing that exists already.

angaither (approved):

weerd:
:trollface: 👍

PR #7144, opened 2016-10-20
handle rogue failures when using gateway

### LINKED ISSUE #7123: checking on status of posting to Rogue with Gateway

katiecrane:
#### What's this PR do?

Catch all the errors that we can get back when posting to Rogue via Gateway. Gateway returns the response in a different way than we got before (already `json_decode`d). Gateway also doesn't handle exceptions of type `GuzzleHttp\Exception\ServerException` for us, so we are looking out for those.

Make sure that no matter what error you get back, the campaign page just refreshes and you see a fresh entry in the `dosomething_rogue_failed_task_log` table.

@DFurnes I believe had reservations about using `catch (Exception $e)` but I wasn't able to catch the exceptions in any other way. Maybe I wasn't throwing them right though.
#### Relevant tickets

Fixes #7123
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
@@ -69,7 +69,6 @@ function dosomething_rogue_send_reportback_to_rogue($values, $user = NULL) {
         'quantity' => $values['quantity'],
         'why_participated' => $values['why_participated'],
         'file' => $values['file'],
-        'file_id' => '', // @TODO - currently require by rogue, should not be.
         'caption' => $values['caption'],
         'status' => isset($values['status']) ? $values['status'] : 'pending',
         'crop_x' => $values['crop_x'],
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -77,36 +77,29 @@ function dosomething_rogue_send_reportback_to_rogue($values, $user = NULL) {
...
+    if (!$response) {
+      // This is a 404
+      dosomething_rogue_handle_failure($user, $values, $run, $response, $e);
+    }
+  }
+  catch (GuzzleHttp\Exception\ServerException $e) {
+    // These aren't yet caught by Gateway
+    dosomething_rogue_handle_failure($user, $values, $run, $response, $e);
+  }
+  catch (Exception $e) {
```
DFurnes: Can you try adding a `dump(get_class($e));die();` below this line and see what it says?

DFurnes: Or actually `dump(get_class($e), get_parent_class($e));die();` would be pretty helpful too.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -77,36 +77,29 @@ function dosomething_rogue_send_reportback_to_rogue($values, $user = NULL) {
         'crop_rotate' => $values['crop_rotate'],
   ];
 
-  $response = $client->postReportback($data);
```
sbsmith86, line 80: Is this `postReportback()` function not being used anymore for any reason? If so, we should probably remove the function definition if it is not useful.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -164,3 +157,39 @@ function dosomething_rogue_store_rogue_references($rbid, $fid, $rogue_reportback
...
+        'file' => $values['file'],
+        'caption' => $values['caption'],
+        'timestamp' => REQUEST_TIME,
+        'response_code' => $response->code, // @TODO: this is currently null until there a better way to get it from Gateway
+        'response_values' => (isset($e)) ? $e->getMessage() : NULL,
+      ])
+      ->execute();
+
+      watchdog('dosomething_rogue', 'reportback not migrated to Rogue', ['user' => $user->uid, 'campaign_id' => $values['nid'], 'campaign_run_id' => $run->nid], WATCHDOG_ERROR);
+}
```
sbsmith86: newline needed 😬

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -164,3 +157,39 @@ function dosomething_rogue_store_rogue_references($rbid, $fid, $rogue_reportback
...
+   if (module_exists('stathat')) {
+      stathat_send_ez_count('drupal - Rogue - reportback failed - count', 1);
+    }
+
+    // Save fail to a db log so we can easily export.
+    db_insert('dosomething_rogue_failed_task_log')
+      ->fields([
+        'drupal_id' => $user->uid,
+        'campaign_id' => $values['nid'],
+        'campaign_run_id' => $run->nid,
```
sbsmith86, line 183: Since this is in it's own function now, instead of passing `$run` into this function you can just grab it based on the values array with `$run = dosomething_helpers_get_current_campaign_run_for_user($values['nid']);`

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
katiecrane: good catch!

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
katiecrane: oh whoops, that should be used

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
katiecrane: `"DoSomething\Gateway\Exceptions\BadRequestException" 
 "DoSomething\Gateway\Exceptions\ApiException"`

the second one prints out that when it throws `new BadRequestException($endpoint, json_encode($response))`

katiecrane: well now it looks like it actually _is_ catching `DoSomething\Gateway\Exceptions\ApiException`s...everything is showing up in the failed task log it looks like!

katiecrane:
@sbsmith86 Do you think this is good to go? The fixes from your comments should all be in and @DFurnes 's concern was I believe about catching type `Exception` but now it's catching type `DoSomething\Gateway\Exceptions\ApiException`

sbsmith86:
Looks good to me 👍

PR #7146, opened 2016-10-26
Updating onboarding member count

itsjoekent:
#### What's this PR do?

Updates the member count in the onboarding copy
~(We should probably pull this from the API like everything else, @jessleenyc should we make a card for this? Should be a quick one)~

Does it display, get ready for it. FIVE POINT FOUR MILLION?
image

Also searched here to make sure we had nothing else to update https://github.com/DoSomething/phoenix/search?utf8=%E2%9C%93&q=5.3+million&type=Code
#### Relevant tickets

Fixes https://trello.com/c/orhCm3ka/154-as-an-admin-i-want-onboarding-to-show-that-we-have-5-4-million-members-so-that-users-see-how-big-our-community-is
#### Checklist
- [ ] Tested on staging.


weerd (approved):
Excellent GIF choice. 👍

PR #7152, opened 2016-10-27
Rogue fail permalink page

### LINKED ISSUE #7116: Rogue failure experience for end users.
#### User Story

As a user, if my reportback doesn't successfully get sent to rogue or get sent back to phoenix, I would like to see some sort of messaging or experience that indicates my reportback was submitted even though I don't get taken to the permalink page. 

When rogue is turned on we attempt to send the reportback to rogue and the the reportback gets sent back and stored in phoenix. We then direct the user to the permalink page that confirms the submission was received. If something errors out on the backend with sending the RB to rogue or sending it back to Phoenix we should have some sort of sensible experience that lets the user know what's going on. We are logging stuff on our back end, but don't surface things to user. 

Currently, we just reload the campaign page.

katiecrane:
#### What's this PR do?
- if a user's reportback fails to post to Rogue, they hit this (tongue cat is not hard coded, and neither is the copy!):
  image
- the image (grabbed by image `nid`) and copy are editable by admins at `admin/config/dosomething/dosomething_campaign` under Permalink Page > Owners Permalink (this won't be shared so only an "owner" will hit it) > Owners Permalink Failure Page

image

- Should anything happen and we want to change the copy on this page, it is a quick change not needing any code (for example, if Rogue goes down and it will take longer than we say for the RB to show up and we want to change it). 

Make sure the page looks okay and that we have access to all the variables and info we need/want for the failure page. Also if anyone has any other thoughts on if the path should be something other than `reportback/failure` I'm very open to that but didn't know what else to make it.

`drush cc all` often when doing stuff like this or it'll take you longer than it should to track down your mistake!
#### Relevant tickets

Fixes #7116
#### Checklist
- [x] Tested on staging.
- [ ] Pinged a PM if this is a larger PR that would benefit from some additional testing love (cc: @lkpttn @ngjo)

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
@@ -271,12 +271,12 @@ function dosomething_campaign_admin_config_form($form, &$form_state) {
     '#required' => TRUE,
     '#default_value' => t(variable_get('dosomething_campaign_permalink_failure_copy')),
   );
-  $form['permalink']['owners']['failure']['dosomething_campaign_permalink_failure_image_fid'] = array(
+  $form['permalink']['owners']['failure']['dosomething_campaign_permalink_failure_image_nid'] = array(
     '#type' => 'textfield',
     '#title' => t('Permalink failure page image'),
-    '#description' => t('FID of the image that appears on the failure page in place of the reportback image.'),
+    '#description' => t('NID of the image that appears on the failure page in place of the reportback image.'),
     '#required' => TRUE,
-    '#default_value' => t(variable_get('dosomething_campaign_permalink_failure_image_fid')),
+    '#default_value' => t(variable_get('dosomething_campaign_permalink_failure_image_nid')),
   );
   $form['permalink']['owners']['dosomething_campaign_permalink_owners_page_title'] = array(
     '#type' => 'textfield',
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
@@ -258,6 +258,26 @@ function dosomething_campaign_admin_config_form($form, &$form_state) {
     '#required' => TRUE,
     '#default_value' => t(variable_get('dosomething_campaign_permalink_confirmation_owners_social_network_copy')),
   );
+  $form['permalink']['owners']['failure'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Owners Permalink Failure Page'),
+    '#collapsible' => TRUE,
+    '#collapsed' => TRUE,
+  );
+  $form['permalink']['owners']['failure']['dosomething_campaign_permalink_failure_copy'] = array(
+    '#type' => 'textarea',
+    '#title' => t('Permalink failure page text'),
+    '#description' => t('Appears on the failure page to the right side of the image.'),
+    '#required' => TRUE,
+    '#default_value' => t(variable_get('dosomething_campaign_permalink_failure_copy')),
+  );
+  $form['permalink']['owners']['failure']['dosomething_campaign_permalink_failure_image_fid'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Permalink failure page image'),
+    '#description' => t('FID of the image that appears on the failure page in place of the reportback image.'),
+    '#required' => TRUE,
+    '#default_value' => t(variable_get('dosomething_campaign_permalink_failure_image_fid')),
+  );
   $form['permalink']['owners']['dosomething_campaign_permalink_owners_page_title'] = array(
     '#type' => 'textfield',
     '#title' => t('Permalink owners title'),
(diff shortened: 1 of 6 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -410,7 +410,14 @@ function dosomething_reportback_form_submit($form, &$form_state) {
     } else {
       // @TODO - if the reportback doesn't successfully get sent back to phoenix, how should
```
angaither: this isn't a @todo anymore ;)

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -242,6 +242,13 @@ function dosomething_reportback_menu() {
       'file' => 'dosomething_reportback.admin.inc',
     );
   }
+  $items['reportback/failure'] = array(
+    'title' => 'Reportback',
+    'page callback' => 'dosomething_reportback_view_failure',
+    // 'page arguments' => array(1),
```
angaither: remove commented out code

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -616,6 +623,42 @@ function dosomething_reportback_view_entity($entity) {
...
+ * We hit this when a reportback sent to Rogue fails to send back to Phoenix
+ */
+function dosomething_reportback_view_failure() {
+  // Get the user in all the forms that we need
+  global $user;
+  $rb_simple_user = dosomething_user_get_simple_user($rb_user = user_load($user));
+
+
+  // Load the campaign node so we can find out its name, if there's a scholarship, etc.
+  $nid = $_GET['nid'];
```
angaither: we can go back to passing this as an arg, so you don't need to `$_GET`

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -616,6 +623,42 @@ function dosomething_reportback_view_entity($entity) {
...
+  $node = dosomething_campaign_load(node_load($nid));
+
+  // Get all the copy and allow admins to build a link to the user's profile
+  $copy_vars = _dosomething_reportback_get_permalink_copy_vars($rb_user);
```
angaither: does this work? looks like the variable above is `$rb_simple_user` not `$rb_user`

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -616,6 +623,42 @@ function dosomething_reportback_view_entity($entity) {
...
+  $failure_copy = token_replace(t(variable_get('dosomething_campaign_permalink_failure_copy', '')), ['user' => $user]);
+  $incentive = ($node->scholarship) ? $copy_vars['owners_rb_scholarship'] : "";
+
+  // Get the image
+  $file = reportback_file_load(variable_get('dosomething_campaign_permalink_failure_image_fid'));
```
angaither: I think this probably throws a lot of errors because it's trying to load an actual reportback item, but we are just using an image node to use for this, right? 
should be a call to `dosomething_image_get_themed_image`

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.admin.inc
```diff
@@ -258,6 +258,26 @@ function dosomething_campaign_admin_config_form($form, &$form_state) {
...
+    '#type' => 'textarea',
+    '#title' => t('Permalink failure page text'),
+    '#description' => t('Appears on the failure page to the right side of the image.'),
+    '#required' => TRUE,
+    '#default_value' => t(variable_get('dosomething_campaign_permalink_failure_copy')),
+  );
+  $form['permalink']['owners']['failure']['dosomething_campaign_permalink_failure_image_fid'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Permalink failure page image'),
+    '#description' => t('FID of the image that appears on the failure page in place of the reportback image.'),
```
angaither: this should probably be an image node

angaither (changes requested):

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
katiecrane: yeah it pulls the "You rocked this campaign!" and stuff. I went off of [how it's done here](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module#L544) for the regular permalink page, should it be different?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
katiecrane: so that part broke and I'm not sure if maybe it never should have worked? Every time I tried it after that initial commit it went to `dosomething_reportback_view_entity` as the callback function

angaither:
🚢

angaither (approved):

PR #7154, opened 2016-10-27
Sends most recently created reportback item and caption if RB exists in Phoenix but not Rogue and quantity/why is updated

### LINKED ISSUE #7153: If a RB doesn't exist in Rogue and a member updates their quantity/why, the request to Rogue fails because as far as Rogue knows there is no file.
#### Current Behavior
- If a reportback doesn't exist in Rogue and a member wants to update their quantity/why for their reportback, the request fails because Rogue doesn't have a file for the reportback. 
#### Desired Behavior
- If the reportback doesn't yet exist in Rogue, Phoenix will send the most recently created reportback item and caption along with the updated quantity/why. 
#### Steps to Replicate
- In Sequel Pro, go to the `dosomething_rogue_reportbacks` table in the Phoenix database and find a reportback that exists both in Phoenix and Rogue. 
- Note the `rogue_reportback_id`. 
- Go to the `reportbacks` table in the Rogue database and find this reportback by `rogue_reportback_id`. Delete this (it will also delete associated reportback items). 
- In your local Phoenix browser, go to the campaign that the reportback belongs to and update the why participated or the quantity. 
- The campaign page should refresh. 
- Check the Phoenix database - this reportback item should not have updated with new quantity/why. 
- Check the Rogue database - there should be no new reportbacks or reportback items from this update. 
#### Why This Matters
- We want to make sure Rogue does not break if a reportback exists in Phoenix but not yet in Rogue and a user updates their quantity/why.

chloealee:
#### What's this PR do?

If a reportback item does not yet exist in Rogue but exists in Phoenix and a user updates the quantity/why or the reportback, instead of breaking, it just sends Rogue the most recently created reportback item and caption. 
- In Sequel Pro, go to the `dosomething_rogue_reportbacks` table in the Phoenix database and find a reportback that exists both in Phoenix and Rogue. 
- Note the `rogue_reportback_id`. 
- Go to the `reportbacks` table in the Rogue database and find this reportback by `rogue_reportback_id`. Delete this (it will also delete associated reportback items). 
- In your local Phoenix browser, go to the campaign that the reportback belongs to and update the why participated or the quantity. 
- The next page should be the permalink page. 
- Check the database - the record should show up in the `dosomething_rogue_reportbacks` table. Additionally, you should see this new record in both Rogue's `reportbacks` and `reportback_items` tables. 
#### Relevant tickets

Fixes #7153
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
@@ -193,6 +193,13 @@ function dosomething_rogue_handle_failure($user, $values, $response = NULL, $e =
       watchdog('dosomething_rogue', 'reportback not migrated to Rogue', ['user' => $user->uid, 'campaign_id' => $values['nid'], 'campaign_run_id' => $run->nid], WATCHDOG_ERROR);
 }
 
+/**
+ * Query to see if a reportback exists in Rogue by Phoenix $rbid.
+ *
+ * @param string $rbid
+ * Phoenix $rbid of reportback.
+ *
+ */
 function dosomething_rogue_rb_exists_in_rogue($rbid)
 {
   return db_query("SELECT * FROM {dosomething_rogue_reportbacks} rogue_rbs WHERE rbid = :rbid", array(':rbid' => $rbid))->fetchAll();
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
@@ -65,7 +65,7 @@ function dosomething_rogue_send_reportback_to_rogue($values, $user = NULL) {
     $rbid = dosomething_reportback_exists($values['nid'], $run->nid, $user->uid);
     // If there is a RB in Phoenix, check the dosomething_rogue_reportback table to see if the RB already exists in Rogue.
     if ($rbid) {
-      $exists_in_rogue = db_query("SELECT * FROM {dosomething_rogue_reportbacks} rogue_rbs WHERE rbid = :rbid", array(':rbid' => $rbid))->fetchAll();
+      $exists_in_rogue = dosomething_rogue_rb_exists_in_rogue($rbid);
       // If the RB does not yet exist in Rogue, send along the most recently created reportback_item so the RB can be created in Rogue.
       if (! $exists_in_rogue) {
         $reportback = reportback_load($rbid);
@@ -209,4 +209,9 @@ function dosomething_rogue_handle_failure($user, $values, $response = NULL, $e =
       ->execute();
 
       watchdog('dosomething_rogue', 'reportback not migrated to Rogue', ['user' => $user->uid, 'campaign_id' => $values['nid'], 'campaign_run_id' => $run->nid], WATCHDOG_ERROR);
+}
+
+function dosomething_rogue_rb_exists_in_rogue($rbid)
+{
+  return db_query("SELECT * FROM {dosomething_rogue_reportbacks} rogue_rbs WHERE rbid = :rbid", array(':rbid' => $rbid))->fetchAll();
 }
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
@@ -60,6 +60,20 @@ function dosomething_rogue_send_reportback_to_rogue($values, $user = NULL) {
 
   $client = dosomething_rogue_client();
 
+  // If there is no file, check to see if the user has already reported back for this campaign in Phoenix.
+  if (! isset($values['file'])) {
+    // $rbid = '2351';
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -60,21 +60,39 @@ function dosomething_rogue_send_reportback_to_rogue($values, $user = NULL) {
 
   $client = dosomething_rogue_client();
 
+  // If there is no file, check to see if the user has already reported back for this campaign in Phoenix.
+  if (! isset($values['file'])) {
+    $rbid = dosomething_reportback_exists($values['nid'], $run->nid, $user->uid);
+    // If there is a RB in Phoenix, check the dosomething_rogue_reportback table to see if the RB already exists in Rogue.
+    if ($rbid) {
+      $exists_in_rogue = db_query("SELECT * FROM {dosomething_rogue_reportbacks} rogue_rbs WHERE rbid = :rbid", array(':rbid' => $rbid))->fetchAll();
```
sbsmith86: I feel like this could be a useful funcition

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -60,21 +60,39 @@ function dosomething_rogue_send_reportback_to_rogue($values, $user = NULL) {
...
+      // If the RB does not yet exist in Rogue, send along the most recently created reportback_item so the RB can be created in Rogue.
+      if (! $exists_in_rogue) {
+        $reportback = reportback_load($rbid);
+        $most_recent_rb_item_id = array_pop($reportback->fids);
+        $reportback_file = entity_load('reportback_item', array($most_recent_rb_item_id));
+        $image = file_load($most_recent_rb_item_id);
+        $values['file'] = 'data: ' . mime_content_type($image->filemime) . ';base64,' . base64_encode(file_get_contents($image->uri));
```
sbsmith86: There is a function to do this `dosomething_helpers_get_data_uri_from_image($image);`

sbsmith86:
I have a more general comment that the logic for this should be done in the form submit hook. It feels like the `dosomething_rogue_send_reportback_to_rogue()` function should only be responsible for sending the request and we are starting to duplicate logic. 

For example [starting here](https://github.com/DoSomething/phoenix/blob/b3cf80366c63b2ab3d60b93255e5d53bb524fb99/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc#L383) we are already doing a lot of the same work (checking if a file was uploaded, checking if the rb exists, etc)

And then the `dosomething_rogue` module could define functions that help us talk to rogue or know if it exists there, or anything else.

However, I like having to just call one rogue function and having it deal with what it needs itself. So open to thoughts on this.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
chloealee: I saw that but when just sending the `$image` to that function, the object didn't exactly match up exactly :-/ e.g. in the helpers function, `$image->source` doesn't exist and is used to get the same thing as what `$image->filemime` is via this line:`$image_path = drupal_realpath($image->source);` 

I used the same logic but just used the info. that was already available in the `$image` object.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
chloealee: Good call! I will create a helpfer function for this :)

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
sbsmith86: ah ok well then this is why I made that other comment that we should be doing this logic in the submit hook. It seems inefficient to duplicate logic.

chloealee: yes agreed - I forgot about that logic there, thanks for pointing out! I'll work on refactoring this now.

chloealee:
@sbsmith86 ready for another 👀 !

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -193,3 +192,8 @@ function dosomething_rogue_handle_failure($user, $values, $response = NULL, $e =
 
       watchdog('dosomething_rogue', 'reportback not migrated to Rogue', ['user' => $user->uid, 'campaign_id' => $values['nid'], 'campaign_run_id' => $run->nid], WATCHDOG_ERROR);
 }
+
+function dosomething_rogue_rb_exists_in_rogue($rbid)
```
sbsmith86, line 203: Needs a docblock

chloealee: @sbsmith86 ah yes, thanks! Updated :)

sbsmith86:
👍

PR #7155, opened 2016-10-28
Fix login/profile when we don't have a locally hashed password.

### LINKED ISSUE #7150: Hide password field if user doesn't have a local Drupal password.
#### User Story

"As a dev, I want to keep track of how someone signed up"
#### Additional Information

We need to be able to know if an account has a "local" password in the Drupal database – if not, it's ridiculous to ask a user to enter their "current password" (which OpenID Connect module sets to a random string) in order to change it.

If a user has registered via the OpenID Connect flow (or changed their password that way??), we should set their account to _always_ update that info in Northstar from then on.
#### Why This Matters

Without this, we'll have a poor user experience for people who are A/B tested and register/change their password via the OpenID Connect flow. 
#### Current Workaround w/o Feature

Suffer.

DFurnes:
#### What's this PR do?

This pull request fixes a couple of _fun_ edge cases that would've made A/B testing Northstar auth difficult. In order to deal with the potential cases where Drupal doesn't know a user's password, we set a boolean field on their local profile if they were created on Northstar. ❇️

This means that we can conditionally remove the password field on these users's profiles (they'll be redirected to Northstar once that page is shipped), and we can hand off login requests for these accounts to Northstar's `v1/auth/verify` endpoint since we have no way of checking their password locally. Any accounts that were made locally continue to function normally.

Give it a good look. 👀

I'm going to try to document all the possible ways a user could go through the login/register/profile flows once we toggle this on so we don't get surprised by any edge cases, but hopefully this is a good start to a sane gradual rollout.
#### Relevant tickets

Fixes #7150.
#### Checklist
- [ ] Tested on staging.


weerd:
> fun edge cases

What other kind are there!?! :trollface:

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
```diff
@@ -160,3 +166,33 @@ function dosomething_northstar_user_logout($account) {
...
 }
+
+/**
+ * Submit handler for handling "classic" logins of accounts that do not have
+ * local passwords, using Northstar's `v1/auth/verify` endpoint.
+ *
+ * @param $form
+ * @param $form_state
+ */
+function dosomething_northstar_fallback_classic_login(&$form, &$form_state) {
```
weerd, line 177: classic 🎩

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
```diff
@@ -160,3 +166,33 @@ function dosomething_northstar_user_logout($account) {
...
+  // If the account does not have a local password, verify using Northstar
+  $northstar_user = dosomething_northstar_verify_user([
+    'username' => $form_state['input']['name'],
+    'password' => $form_state['input']['pass'],
+  ]);
+
+  // If successful response & ID field is given, the credentials worked.
+  if (! empty($northstar_user['data']['id'])) {
+    $sub = $northstar_user['data']['id'];
```
weerd, line 186: What does `$sub` stand for?

DFurnes: Named for the [`sub` JWT claim](https://github.com/DoSomething/northstar/blob/dev/documentation/authentication.md#access-tokens), since the OpenID Connect method below refers to it that way.

weerd: Cool cool. I was just curious :)

lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -1091,6 +1089,11 @@ function dosomething_user_login_validate($form, &$form_state) {
       }
     }
   }
+
+  // If Northstar is enabled & account is externally created, try to auth with Northstar.
```
weerd, line 1093: Huzzah!

weerd (approved):
Looks awesome! Another nail for the ⚰ ! 👍

PR #7156, opened 2016-10-28
remove nid but leave %

katiecrane:
#### What's this PR do?

This seems to work if you leave the wildcard as just `%`.

Can you hit `/reportback/failure/1283`?
#### Checklist
- [x] Tested on staging.


angaither:
👍

PR #7161, opened 2016-10-28
Multiple rb items status updates

### LINKED ISSUE #7159: Only last rb item's status update gets sent to Rogue from Phoenix when reviewed
#### Current Behavior
- Only the last rb item's status update gets sent to Rogue from Phoenix when reviewed.
#### Desired Behavior
- If multiple rb items are reviewed in Phoenix, all of the rb item's updated statuses should be sent to Rogue. 
#### Steps to Replicate
- Pick a campaign that has two or more reportback items to review and note them in the rogue database. 
- Review the items and change to `approved`. 
- Check rogue database - only the last rb item's status is updated to `approved` - the other is still `pending`. 
#### Why This Matters

We want the same data to be in the Phoenix and Rogue databases.

chloealee:
#### What's this PR do?
- Fixes bug that it only sends one rb item's updated status to Rogue. 
- Adds logic to send the new Rogue rb item `status` terms instead of `Approved`, `Promoted`, `Excluded`, and `Flagged`  
- Review a campaign's reportback items that has two or more reportback items. 
- Make sure the items `status` is updated in the Rogue database with the following: 
  - `Approved`, `Promoted`, `Excluded` should be `accepted` in the Rogue database
  - `Flagged` should be `rejected` in the Rogue database

done with @angaither !
#### Relevant tickets

Fixes #7159 
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
@@ -291,14 +291,14 @@ function dosomething_reportback_files_form_submit($form, &$form_state) {
       // If there is a rogue_item_id in the dosomething_rogue_reportbacks table, send update to Rogue.
       if ($rogue_item_id[0]->rogue_item_id) {
         if ($values['status'] === 'promoted' || $values['status'] === 'approved' || $values['status'] === 'excluded') {
-          $values['status'] = 'accepted';
+          $rogue_status = 'accepted';
         } else {
-          $values['status'] = 'rejected';
+          $rogue_status = 'rejected';
         }
 
         $data = [
           'rogue_reportback_item_id' => $rogue_item_id[0]->rogue_item_id,
-          'status' => $values['status'],
+          'status' => $rogue_status,
         ];
 
         $updates_to_send_to_rogue[] = $data;
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
@@ -240,7 +240,6 @@ function dosomething_reportback_files_form_submit($form, &$form_state) {
   }
 
   $updates_to_send_to_rogue = [];
-
   foreach ($rb_files as $fid => $values) {
     $kudos = [
       'fid' => $fid,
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
```diff
@@ -290,13 +289,19 @@ function dosomething_reportback_files_form_submit($form, &$form_state) {
       $rogue_item_id = dosomething_rogue_get_by_file_id($fid);
 
       // If there is a rogue_item_id in the dosomething_rogue_reportbacks table, send update to Rogue.
-      if (!is_null($rogue_item_id[0]->rogue_item_id)) {
+      if ($rogue_item_id[0]->rogue_item_id) {
+        if ($values['status'] === 'promoted' || $values['status'] === 'approved' || $values['status'] === 'excluded') {
+          $values['status'] = 'accepted';
```
angaither: I'm not sure if naming this will cause issues with how it's saved or something, might be nice to save it as `rogue_value` or something so there isn't a collision

angaither (changes requested):

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
chloealee: @angaither ah good call! Just updated it to `rogue_status`

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
```diff
@@ -268,7 +267,7 @@ function dosomething_reportback_files_form_submit($form, &$form_state) {
 
     // Save kudos reaction from admin on RB item.
     if ($values['status'] === 'promoted' || $values['status'] === 'approved' || $values['status'] === 'excluded') {
-      $record = (new KudosController)->create($kudos);
+      // $record = (new KudosController)->create($kudos);
```
angaither: oh an this should be uncommented!

chloealee: ah yes forgot that again! thanks! updated :)

angaither (approved):

angaither:
:shinto_shrine:

PR #7162, opened 2016-10-28
Fixes issues with SSO logout.

DFurnes:
#### What's this PR do?
Fixing more issues with the OpenID Connect flow!

🚪 Fixes an issue where we were losing the session "marker" that was determining whether to use SSO logout, by instead making that decision based on whether the user has a refresh token.

🏗 Clears out the token fields when a user logs out, so we can track whether to use SSO logout for them or not. In the future, we should also revoke the refresh token when doing this but I ran into some Gateway issues so putting a hold on that.

🍃 Cleans up conditionals to make `dosomething_northstar_fallback_classic_login` easier to follow. The diff looks kinda messy but the end result looks nicer!

👀 

🇱🇦

#### Relevant tickets
Fixes DoSomething/northstar#457.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
@@ -180,10 +180,6 @@ function dosomething_northstar_user_logout($account) {
 
   $is_openid_login = ! empty($user->field_refresh_token[LANGUAGE_NONE][0]['value']);
   if ($is_openid_login) {
-    $northstar_logout_url = url(NORTHSTAR_URL . '/logout', [
-      'query' => ['redirect' => url('/', ['absolute' => TRUE])],
-    ]);
-
     // Clear out token information to track that SSO session will be revoked.
     // @TODO: Actually revoke the refresh token in Northstar on logout.
     $edit = [];
@@ -192,9 +188,13 @@ function dosomething_northstar_user_logout($account) {
       'access_token_expiration' => NULL,
       'refresh_token' => NULL,
     ]);
-
     user_save($user, $edit);
 
+    // And redirect to SSO logout to prevent users from automatically
+    // logging in to the same account again.
+    $northstar_logout_url = url(NORTHSTAR_URL . '/logout', [
+      'query' => ['redirect' => url('/', ['absolute' => TRUE])],
+    ]);
     drupal_goto($northstar_logout_url);
   }
 
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
@@ -173,11 +173,12 @@ function dosomething_northstar_perform_authorization_actions() {
  * @param $account
  */
 function dosomething_northstar_user_logout($account) {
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
```diff
@@ -173,15 +173,32 @@ function dosomething_northstar_perform_authorization_actions() {
...
-  $northstar_logout_url = url(NORTHSTAR_URL . '/logout', [
-    'query' => ['redirect' => url('/', ['absolute' => TRUE])],
-  ]);
+  $user = user_load($account->uid);
 
   // Destroy the current session, and reset $user to the anonymous user.
   session_destroy();
 
-  drupal_goto($is_openid_login ? $northstar_logout_url : '/');
+  $is_openid_login = ! empty($user->field_refresh_token[LANGUAGE_NONE][0]['value']);
```
weerd, line 181: just curious, will this _always_ be `LANGUAGE_NONE` or could it vary for users from Brazil, Mexico?

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
```diff
@@ -173,15 +173,32 @@ function dosomething_northstar_perform_authorization_actions() {
...
+      'refresh_token' => NULL,
+    ]);
+    user_save($user, $edit);
+
+    // And redirect to SSO logout to prevent users from automatically
+    // logging in to the same account again.
+    $northstar_logout_url = url(NORTHSTAR_URL . '/logout', [
+      'query' => ['redirect' => url('/', ['absolute' => TRUE])],
+    ]);
+    drupal_goto($northstar_logout_url);
```
weerd, line 198: ▶️ ◀️ 🔁

weerd: hehe :)

weerd (approved):
One quick question, but otherwise looks good! 👍

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
DFurnes: I'm fairly certain it will always be saved under `LANGUAGE_NONE` because that's [how the `dosomething_user_set_fields` function saves it](https://github.com/DoSomething/phoenix/blob/085eed871725ce89425bf19ff99515f3d3ef1c76/lib/modules/dosomething/dosomething_user/dosomething_user.module#L1383), but I'm not sure if there's anything else that could hook in and cause trouble. @angaither?

DFurnes: I decided against using `dosomething_user_get_field` here because it logs to Watchdog if the field isn't set and that seemed like it would create a lot of noise.

angaither: yup it'll be `LANGUAGE_NONE` unless we decide to translate those fields, which I don't think we ever will!

weerd: 👍

PR #7164, opened 2016-11-01
hit correct permalink page on reportback creation with Rogue

### LINKED ISSUE #7160: not making the right checks to see if reportback made it back from Rogue
#### Current Behavior

Currently, we are saying that the post to Rogue was successful if there is a reportback ID before it even tries to post AKA it only thinks the post was successful IFF you are updating your reportback.
#### Desired Behavior

Don't tell users their image is processing if it is processed already.
#### Steps to Replicate
1. Turn on Rogue
2. Make a new reportback
3. You will hit the failed permalink page
4. Go back to the campaign
5. Well well well, see that your reportback is in fact there
6. Post an update
7. You will hit the normal permalink page
8. 👻 
#### Why This Matters

We don't want to tell users their stuff isn't processing when it is.

### LINKED ISSUE #7167: Incorrect values getting stored in the `dosomething_rogue_reportbacks` table.
_BUG_
-------------------
#### Current Behavior
- The incorrect `fid` in getting stored in the `dosomething_rogue_reportbacks` table. 
  - E.g. I submitted a reportback on Phoenix and then submitted two more reportback items for that reportback (total of 3 reportback items for this reportback). 
  - I then checked the Phoenix database and found their fids in the `dosomething_reportback_file` table: `9307`, `9309`, and `9311`. 
  - I checked the Rogue database and found their rogue ids in the `reportback_items` table: `340`, `341`, and `342`.
  - I went back to the Phoenix database to check that the values were correct in the `dosomething_rogue_reportbacks` table and they were not. Also, there were only two reportback items in the table. The table said the reportback item with `rogue_item_id` = `341` is associated with `fid` = `9313` (it is supposed to be associated with `fid` = `9309`) and the reportback item with `rogue_item_id` = `342` is associated with `fid` = `9315` (it is supposed to be associated with `fid` = `9311`). 

#### Desired Behavior
- All 3 reportback items should have made it in this table. 
- All of their values should reflect those in the `dosomethign_reportback_file` and `reportback_items` table.

#### Steps to Replicate
- See above.

#### Why This Matters
This table is used to make sure information of reportbacks and reportback items is consistent in Phoenix and Rogue.

katiecrane:
#### What's this PR do?
We check if a reportback exists:
- before the update/new one is sent to Rogue (and grab the most recent file in case we need to send it to Rogue as well) <- this is what we already did
- after the reportback should have come back from Rogue so we know if it made it back successfully (if we know this we can send the user to the right permalink page)

This also fixes a bug where the permalink page is 1 image behind (meaning it shows the second most recently uploaded image).

Make sure you are hitting the right permalink page when creating a new reportback.

I think this was working right at some point, but got muddled in another fix.

#### Relevant tickets
Fixes #7160
Fixes #7167 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -379,9 +379,12 @@ function dosomething_reportback_form_submit($form, &$form_state) {
 
   // Collect reportbacks in Rogue if rogue collection is turned on for this campaign.
   if (variable_get('rogue_collection', FALSE)) {
+    // If this is a reportback update, grab the file most recent file in case we need it later
     $rbid = dosomething_reportback_exists($form_state['build_info']['args'][0]->nid, $form_state['build_info']['args'][0]->run_nid, $form_state['build_info']['args'][0]->uid);
-    $reportback = reportback_load($rbid);
-    $fid = array_pop($reportback->fids);
+    if ($rbid) {
+      $reportback = reportback_load($rbid);
+      $fid = array_pop($reportback->fids);
+    }
 
     // Store the base64 encoded data uri of the file in the values array to send to rogue.
     if (array_key_exists('storage', $form_state)) {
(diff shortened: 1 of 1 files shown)


sbsmith86:
Looks ok to me!

PR #7165, opened 2016-11-01
Refactor data uri generation and remove space in generated string

sbsmith86:
#### What's this PR do?

Removes a space in the data uri that is generated for uploaded image. This was causing errors in rogue when trying to edit the image because the library we are using to do cropping and rotating in rogue could not read the malformed uri. 

Also refactors how we generate these URIs. Before, there was one function that took in an image object but we also needed it to work for file objects. This generalizes the helper function so that we can use one function to generate a base64 encoded data uri instead of doing it two places.

#### Relevant tickets
Fixes https://github.com/DoSomething/rogue/issues/61

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -58,16 +58,16 @@
         },
         {
             "name": "giggsey/libphonenumber-for-php",
-            "version": "7.7.2",
+            "version": "7.7.3",
             "source": {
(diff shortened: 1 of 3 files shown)


angaither:
🚢

PR #7169, opened 2016-11-02
Handles failures when making PUT request to Rogue's /items endpoint 

### LINKED ISSUE #7157: Handle errors when making a `PUT` request to Rogue's `/items` endpoint
#### User Story

If Phoenix makes a request to Rogue's `/items` endpoint and it fails, we want to save this information in the `dosomething_rogue_failed_task_log`
#### Why This Matters

We want to record all the information that doesn't make it from Phoenix->Rogue

chloealee:
#### What's this PR do?
- Updates the  `dosomething_rogue_failed_task_log`: 
  - Adds the `rogue_item_id`, `status`, and `type` columns 
  - If a reportback fails to send to Rogue, the `type` will be `reportback`.
  - If a reportback item's status fails to send to Rogue, the `type` will be `reportback item`. 
  - `drupal_id`, `campaign_id`, `campaign_run_id`, `quantity`, and `why_participated` are now all set to `'not_null' => FALSE`.
- Enters the `rogue_item_id` and updated `status` in the `dosomething_rogue_failed_task_log` if the `PUT` request to `/items` fails and sends stathat alerts. 
- Updates the cron job to retry the `PUT` request to `/items` for everything in the `dosomething_rogue_failed_task_log`

- In `lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module`: 
  - Comment out line 81 and replace with `$response = NULL;`
  - Comment out line 87 and replace with `if(is_null($response){`
- Submit a new reportback. 
- Check the `dosomething_rogue_failed_task_log` and all expected columns should be filled out. The `type` should say `reportback`.

- Submit a new reportback item (or many!). 
- In `lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module`: 
  - Comment out line 115 and replace with `$response = NULL;`
  - Comment out line 120 and replace with `if(is_null($response){`
- Review the reportback items. 
- Check the `dosomething_rogue_failed_task_log` table and `rogue_item_id`, `status`, and `type` should be filled out. `type` should be `reportback item`. 

#### Relevant tickets
Fixes #7157 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
@@ -15,17 +15,28 @@
     db_truncate('dosomething_rogue_failed_task_log')->execute();
 
     foreach ($task_log as $task) {
-      $values = [
-        'nid' => $task->campaign_id,
-        'campaign_run_id' => $task->campaign_run_id,
-        'quantity' => $task->quantity,
-        'why_participated' => $task->why_participated,
-        'file' => $task->file,
-        'caption' => $task->caption,
-      ];
+      if ($task->type === 'reportback') {
+        $values = [
+          'nid' => $task->campaign_id,
+          'campaign_run_id' => $task->campaign_run_id,
+          'quantity' => $task->quantity,
+          'why_participated' => $task->why_participated,
+          'file' => $task->file,
+          'caption' => $task->caption,
+        ];
 
-      $user = user_load($task->drupal_id);
+        $user = user_load($task->drupal_id);
 
-      dosomething_rogue_send_reportback_to_rogue($values, $user);
+        dosomething_rogue_send_reportback_to_rogue($values, $user);
+      } else {
+        $values = [
+          [
+            'rogue_reportback_item_id' => $task->rogue_item_id,
+            'status' => $task->status,
+          ]
+        ];
+
+        dosomething_rogue_update_rogue_reportback_items($values);
+      }
    }
  }
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install
@@ -181,3 +181,107 @@ function dosomething_rogue_update_7005(&$sandbox) {
   }
 }
 
+/**
+ * Change schema so drupal_id is not required.
+ */
+function dosomething_rogue_update_7006(&$sandbox) {
+  $table_name = 'dosomething_rogue_failed_task_log';
+  if (db_table_exists($table_name)) {
+    $spec = [
+      'description' => 'The drupal_id of the user.',
+      'type' => 'int',
+      'not null' => FALSE,
+    ];
+    db_change_field($table_name, 'drupal_id', 'drupal_id', $spec);
+  }
+}
+
+/**
+ * Change schema so campaign_id is not required.
+ */
+function dosomething_rogue_update_7007(&$sandbox) {
+  $table_name = 'dosomething_rogue_failed_task_log';
+  if (db_table_exists($table_name)) {
+    $spec = [
+      'description' => 'Campaign nid for the reportback.',
+      'type' => 'int',
+      'not null' => FALSE,
+    ];
+    db_change_field($table_name, 'campaign_id', 'campaign_id', $spec);
+  }
+}
+
+/**
+ * Change schema so campaign_run_id is not required.
+ */
+function dosomething_rogue_update_7008(&$sandbox) {
+  $table_name = 'dosomething_rogue_failed_task_log';
+  if (db_table_exists($table_name)) {
+    $spec = [
+      'description' => 'Campaign run nid for the reportback.',
+      'type' => 'int',
+      'not null' => FALSE,
+    ];
+    db_change_field($table_name, 'campaign_run_id', 'campaign_run_id', $spec);
+  }
+}
+
+/**
+ * Change schema so quantity is not required.
+ */
+function dosomething_rogue_update_7009(&$sandbox) {
+  $table_name = 'dosomething_rogue_failed_task_log';
+  if (db_table_exists($table_name)) {
+    $spec = [
+      'description' => 'The quantity of reportback_nouns reportback_verbed.',
+      'type' => 'int',
+      'not null' => FALSE,
+    ];
+    db_change_field($table_name, 'quantity', 'quantity', $spec);
+  }
+}
+
+/**
+ * Change schema so why_participated is not required.
+ */
+function dosomething_rogue_update_7010(&$sandbox) {
+  $table_name = 'dosomething_rogue_failed_task_log';
+  if (db_table_exists($table_name)) {
+    $spec = [
+      'description' => 'Why the user participated in the campaign.',
+      'type' => 'text',
+      'length' => '2048',
+      'not null' => FALSE,
+    ];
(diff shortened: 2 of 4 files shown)


katiecrane:
@chloealee when you say "Turn Rogue off" do you mean uncheck the box?

chloealee:
@katiecrane sorry for not clarifying! The way I was testing was actually pretty scrappy - I'll update with better instructions now!

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install
```diff
@@ -137,3 +137,151 @@ function dosomething_rogue_update_7002(&$sandbox) {
...
+      'description' => 'The updated status of the reportback item',
+    ];
+    db_change_field($table_name, 'status', 'status', $spec);
+  }
+}
+
+/**
+ * Changes the type to be text in the type column.
+ */
+function dosomething_rogue_update_7012(&$sandbox) {
```
sbsmith86: Couldn't you do all these schema updates in one update function?

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -79,22 +79,23 @@ function dosomething_rogue_send_reportback_to_rogue($values, $user = NULL) {
 
   try {
     $response = $client->postReportback($data);
+    $values['type'] = 'reportback';
 
     if (module_exists('stathat')) {
       stathat_send_ez_count('drupal - Rogue - reportback sent - count', 1);
     }
-    if (!$response) {
+    // if (!$response) {
```
sbsmith86: think you can remove this :)

katiecrane: you can get rid of this commented out line (but this is a great edit!)

sbsmith86:
So we also have a [cron job](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc) that runs to re-send the failed stuff so that we eventually get it (hopefully) I think you will need to update that to re-send the items.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install
chloealee: @sbsmith86 I tried to do this before for another PR and for some reason couldn't get it to work but I might have just done something incorrectly! It would look a lot better if we just did it in one though! To change, since it's already run these, would it be best to delete these columns in my local Sequel Pro, then try to update these all in the same update function but do `7013` just for testing purposes and if it works, then change it to be `7006`?

sbsmith86: yea i think you delete the column ins sequel pro and then go into the `systems` table and search for the module and then update the `schema_version` to be the update right before the new one you want to run again.

sbsmith86: for example: http://drupal.stackexchange.com/questions/69840/resetting-reverting-a-custom-modules-update-n-value

sbsmith86: also real pseudo-code like would something like this work? 

``` php
function dosomething_rogue_update_7003(&$sandbox) {
  $table_name = 'dosomething_rogue_failed_task_log';

  $newfields = [
    'rogue_item_id' => 'description text', 
    'status' => 'description text'
  ];

  if (db_table_exists($table_name)) {
    foreach ($newfields as $field => $description) {
      $spec = [
        'type' => int,
        'description' => $description,
      ];
      db_add_field($table_name, $field, $spec);    
    }

    // similar thing for the fields you just want to change
  }
}
```

chloealee:
awesome, thanks for pointing that out @sbsmith86 ! Is there anyway to test to make sure the code in the cron job works?

sbsmith86:
@chloealee yea. After you update the cron job you run `drush cron` and the reportbacks and items in there should be sent to rogue and then removed from the table or if they fail again, they should get re-stored in failed table.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -109,11 +109,35 @@ function dosomething_rogue_send_reportback_to_rogue($values, $user = NULL) {
  */
 function dosomething_rogue_update_rogue_reportback_items($data)
 {
+  print_r($data);
+  die();
```
katiecrane: don't forget to remove these

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install
chloealee: this is awesome, thank you so much @sbsmith86 !

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -184,13 +210,28 @@ function dosomething_rogue_handle_failure($user, $values, $response = NULL, $e =
...
-      watchdog('dosomething_rogue', 'reportback not migrated to Rogue', ['user' => $user->uid, 'campaign_id' => $values['nid'], 'campaign_run_id' => $run->nid], WATCHDOG_ERROR);
+    watchdog('dosomething_rogue', 'reportback not migrated to Rogue', ['user' => $user->uid, 'campaign_id' => $values['nid'], 'campaign_run_id' => $run->nid], WATCHDOG_ERROR);
+  } else {
+    db_insert('dosomething_rogue_failed_task_log')
+      ->fields([
+        'rogue_item_id' => $values['rogue_reportback_item_id'],
+        'status' => $values['status'],
+        'type' => $values['type'],
+        'timestamp' => REQUEST_TIME,
+        'response_code' => (isset($response->code)) ? $response->code : NULL, // @TODO: this is currently null until there a better way to get it from Gateway
```
katiecrane, line 228: 🙌  would you mind changing the one above to be like this as well while you're at it?

chloealee:
@sbsmith86 just updated everything and ran the cron job and everything worked! Ready for another review :) 

Also - if testing, you'll need to do some steps you helped me with above! 
- delete the `dosomething_rogue_failed_task_log`
- go to the `system` table and search for `name` `dosomething_rogue`
- change the `schema_version` to `7001`
- close your tab and re-open another connection to your local phoenix database

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
```diff
@@ -15,17 +15,28 @@
...
+          'nid' => $task->campaign_id,
+          'campaign_run_id' => $task->campaign_run_id,
+          'quantity' => $task->quantity,
+          'why_participated' => $task->why_participated,
+          'file' => $task->file,
+          'caption' => $task->caption,
+        ];
 
-      $user = user_load($task->drupal_id);
+        $user = user_load($task->drupal_id);
```
sbsmith86, line 28: identation looks off here. Sorry this is the most nitpicky i get.

chloealee: nono totally - just saw that too and saw it looked weird! I think it is ok though because I wrapped that in an `if` statement so now should be indented to line up with the `$values` array above?

sbsmith86: ah i see! yes yes

sbsmith86:
Looks good!

PR #7170, opened 2016-11-03
Extract analytics to helper module, fix lost events on page redirects, and add a few new events.

DFurnes:
#### What's this PR do?
This PR extracts our little Google Analytics helper code into a [standalone npm package](https://github.com/DoSomething/analytics).

It also updates `dosomething_helpers_add_analytics_event` to flash analytics events to the session, and inject the analytics script itself when rendering a page, which fixes an issue where any events that were followed by `drupal_goto`s would be lost. (This will fix the `Authentication:Login` and `Authentication:Register` events that weren't firing.)

Finally, it removes some confusing events and adds a couple of new ones for tracking events that happen in the authentication flow: `Link:Click:Login`, `Link:Click:Register`, `Link:Click:Signup (Authenticated)`, and `Link:Click:Signup (Unauthenticated)`.

#### How should this be reviewed?
The interesting stuff happens in `dosomething_helpers_add_analytics_event` and `preprocess_html.inc`. Most of the rest is just plopping attributes into arrays.

Definitely check out [the source](https://github.com/DoSomething/analytics/blob/56a0e52a32cec86a001ac172f2f6d721000ad13c/analytics.js) for the extracted Analytics package and let me know how it looks!

#### Any background context you want to provide?
💃

#### Relevant tickets
🍃

#### Checklist
- [ ] Documentation added for new features/changed endpoints.
- [ ] Tested on staging.
- [ ] Pinged a PM if this is a larger PR that would benefit from some additional testing love
- [ ] Post a message in #phoenix if this includes something that causes a rebuild!

PR #7171, opened 2016-11-03
Fix errors when trying to output analytics for anonymous users.

DFurnes:
#### What's this PR do?
This fixes an error introduced in #7170 if the user doesn't have the session key set:

screen shot 2016-11-03 at 2 45 53 pm

I also fixed a bunch of calls to the old `sendEvent` function, double-whoops!

👀

💣

#### Relevant tickets
Fixes 🐛.

#### Checklist
- [ ] Tested on staging.


weerd (approved):
😱 👍

PR #7175, opened 2016-11-04
Pass campaign title to Northstar on login-signup.

DFurnes:
#### What's this PR do?
Pass campaign title to Northstar when making a authorization link for the campaign signup flow.

screen recording 2016-11-04 at 11 18 am

👀

Here's the documentation for [Northstar authorization endpoint](https://github.com/DoSomething/northstar/blob/dev/documentation/endpoints/auth.md#create-token-authorization-code-grant).

I'd like to rename the `destination` query parameter to `title`, because Drupal has some crazy overrides if you have a `?destination` on a URL (since it thinks it's a URL and _runs with it_). `?title=` makes more sense anyways!

#### Relevant tickets
Fixes 🚯.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.theme.inc
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.theme.inc
@@ -7,15 +7,16 @@
 /**
  * Returns markup for link to sign up for a campaign.
  *
- * @param string $label
- *   The text to display on the button.
+ * @param string $label - The text to display on the button.
+ * @param string $nid - The campaign node ID
+ * @param string $title - The campaign title
  * @return string
  */
-function dosomething_user_get_sign_up_link_markup($label, $nid) {
+function dosomething_user_get_sign_up_link_markup($label, $nid, $title) {
   // If the OpenID Connect feature flag is enabled, link to login page rather than showing classic modal.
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.theme.inc
```diff
@@ -7,15 +7,16 @@
...
  *
- * @param string $label
- *   The text to display on the button.
+ * @param string $label - The text to display on the button.
+ * @param string $nid - The campaign node ID
+ * @param string $title - The campaign title
  * @return string
  */
-function dosomething_user_get_sign_up_link_markup($label, $nid) {
+function dosomething_user_get_sign_up_link_markup($label, $nid, $title) {
```
weerd, line 15: Should we default to something for `$title`? Or are we requiring always passing the node title when calling the function?

weerd (commented):
quick question

lib/modules/dosomething/dosomething_user/dosomething_user.theme.inc
DFurnes: Yep, that function is only ever used in one place (which is also updated in this PR).

weerd: Righty-o! Just wanted to make sure 💃

weerd (approved):
👍

PR #7176, opened 2016-11-04
Users who registered via Northstar should edit profile there.

DFurnes:
#### What's this PR do?
If a user was _registered_ via Northstar, they should be redirected to Northstar to edit their profile. This will allow them to change their password (since Northstar knows and can check against these users' hashed passwords, and Drupal cannot). This also allows us to start testing the new profile form with a small percentage of users before we launch it as the real (and only) deal.

With 'yer eyes. 👀 

We currently don't redirect the user back to Phoenix after completing their profile form. That's a change we'll have to make on the ✨ Northstar side ✨.

#### Relevant tickets
Fixes 🔄.

#### Checklist
- [ ] Tested on staging.


DFurnes:
[(11) As a developer, I want to update user profile through northstar so that we're moving everything out of drupal.](https://trello.com/c/GqAw4Rb8/110-11-as-a-developer-i-want-to-update-user-profile-through-northstar-so-that-we-re-moving-everything-out-of-drupal)

weerd (approved):
Snazzy! 👍

PR #7178, opened 2016-11-04
Have rogue configured on by default

### LINKED ISSUE #7177: Rogue on by default
_FEATURE OVERVIEW_
-------------------
#### User Story

Rogue is currently turned off by default. But it is live now and we should turn it on by default so we don't have to turn it on manually after every deploy.

sbsmith86:
#### What's this PR do?

Turns rogue on by default so that we do not need to manually do it after every deploy. 

#### Relevant tickets
Fixes #7177 

#### Checklist
- [ ] Tested on staging.


katiecrane:
I think this looks good 🎉

DFurnes:
These look like sane defaults. Do we need to update `bin/ds pull stage` to disable this on local dev?

PR #7184, opened 2016-11-08
Include messagingGroupsQueue.

sergiitk:
#### What's this PR do?
Includes new `messagingGroupsQueue` into Quicksilver (ex. Message Broker) queue mapping.

Run the update.
Tested locally:
image

#### Relevant tickets
A part of DoSomething/Quicksilver-PHP#64.

#### Checklist
- [x] Tested on thor
- [x] Post a message in #phoenix if this includes something that causes a rebuild!

### DIFF
--- a/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
+++ b/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
@@ -144,6 +144,14 @@ function dosomething_mbp_update_7008() {
 }
 
 /**
+ * Updates mbp productions to include messagingGroupsQueue.
+ */
+function dosomething_mbp_update_7009() {
+  db_drop_table('message_broker_producer_productions');
+  dosomething_mbp_install_productions();
+}
+
+/**
  * Common function for install and update of related DoSomething specific
  * message_broker_producer produciton settings.
  */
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
```diff
@@ -144,6 +144,14 @@ function dosomething_mbp_update_7008() {
 }
 
 /**
+ * Updates mbp productions to include messagingGroupsQueue.
+ */
+function dosomething_mbp_update_7009() {
+  db_drop_table('message_broker_producer_productions');
```
aaronschachter, line 150: What does this table store? Want to double check this won't affect the resurrection of mailchimp groups that @justkika was looking to implement for Truth

aaronschachter (approved):

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
sergiitk: Unfortunately, that's normal way to update the mapping. See updates 7004, 7005, 7007, 7008.

That's another one example of a legacy code. It supposed to be refactored to features, but, I believe, it'll be deprecated before the refactoring.

PR #7187, opened 2016-11-10
only pull mime from path

katiecrane:
#### What's this PR do?
There are times when the mimetype ends up being `null` which shouldn't happen. It seems like that was happening because we were calling `mime_content_type($mime)` when `$mime` may already be set as the type if the `if` statement is true. `mime_content_type` accepts a path.

Now that is just called in the `else` statement where it gives `mime_content_type` a path.

Is this going to break anything else that you know of?

#### Checklist
- [x] Tested on staging.

cc: @chloealee you wanted to know about this one!


sbsmith86:
👍

PR #7188, opened 2016-11-11
get content of path (not of mimetype...)

katiecrane:
#### What's this PR do?
Fixes a silly mistake. Makes sure we call `file_get_contents` on a path. 

Do all the functions get called with parameters that make sense?

#### Checklist
- [x] Tested on staging.


sbsmith86:
👍

PR #7190, opened 2016-11-11
pass reportback file source to rogue

### LINKED ISSUE #7189: reportback file source not sent to Rogue
_BUG_
-------------------
#### Current Behavior
If a reportback file has a `source` and is sent to Rogue, the `source` is not.

#### Desired Behavior
Send the `source` to Rogue.

#### Steps to Replicate
Seems that `source` is only set on files that were created via calls to the Phoenix api, so need to check for it when sending items that already exist.

katiecrane:
#### What's this PR do?
Include the item source when sending a file to Rogue. As far as I know this value is only set when a reportback is created via a call to the Phoenix api, which means that it will only ever be sent to Rogue if it is sending a previously uploaded file, so that is the only place where it tries to pull the file.

No work on the Rogue side because it already has this column and tries to pull this value.

Does it grab it okay? Does it need to try anywhere else?

#### Relevant tickets
Fixes #7189

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
@@ -75,6 +75,7 @@ function dosomething_rogue_send_reportback_to_rogue($values, $user = NULL) {
     'crop_width' => $values['crop_width'],
     'crop_height' => $values['crop_height'],
     'crop_rotate' => $values['crop_rotate'],
+    'source' => isset($values['source']) ? $values['source'] : NULL,
   ];
 
   $values['type'] = 'reportback';
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -392,6 +392,7 @@ function dosomething_reportback_form_submit($form, &$form_state) {
       $image = image_load($file->uri);
 
       $values['file'] = dosomething_helpers_get_data_uri_from_image($image);
+      $values['source'] = 'phoenix-web';
     } else {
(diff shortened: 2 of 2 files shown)


lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -75,6 +75,7 @@ function dosomething_rogue_send_reportback_to_rogue($values, $user = NULL) {
     'crop_width' => $values['crop_width'],
     'crop_height' => $values['crop_height'],
     'crop_rotate' => $values['crop_rotate'],
+    'source' => isset($values['source']) ? $values['source'] : NULL,
```
sbsmith86, line 78: Should we start sending a source value to rogue like `phoenix_web` or something for when a reportback is sent from the reportback form to rogue? I know we don't usually do it for the web form but when rogue starts collecting reportbacks from different sources that could be a nice to have. Thoughts?

katiecrane: I like that a lot better than just having `null` mean "from the web"

sbsmith86:
👍

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
katiecrane: @sbsmith86 As of now I left this part but pass source from elsewhere in phoenix. I figured that once the data is in this function we might not know for sure that it came from phoenix web because I'm guessing we will be using this function when sending over data that was sent to the phoenix api. What are your thoughts?

sbsmith86: yea i think this is great. Grab the source if it is there, otherwise assume it is from the web form and set a default

PR #7197, opened 2016-11-17
Send reactions to rogue

### LINKED ISSUE #7196: push reactions to Rogue
_FEATURE OVERVIEW_
-------------------
#### User Story
Gathering more activity in Rogue! Reactions are one type of activity that we would like to live in Rogue.

The endpoint is already setup in Rogue to receive reactions.

katiecrane:
#### What's this PR do?
Sends reactions to Rogue as they are created and deleted! Checks to see if Rogue is on first. This will send reactions created via the Phoenix api to Rogue as well. It only sends the reaction to Rogue if the item exists in Rogue. Logs failures in StatHat.

Give some reactions to some reportbacks. Try a mix of ones in Rogue and not in Rogue. Then check the `reactions` table in Rogue and `dosomething_kudos` in Phoenix to make sure the correct data is there.

#### Relevant tickets
Fixes #7196

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
@@ -255,3 +255,58 @@ function dosomething_rogue_rb_exists_in_rogue($rbid)
 {
   return db_query("SELECT * FROM {dosomething_rogue_reportbacks} rogue_rbs WHERE rbid = :rbid", array(':rbid' => $rbid))->fetchAll();
 }
+
+/**
+ * Sends a reaction to Rogue.
+ *
+ * @param array $values
+ *   Values to send to Rogue. (just the FID)
+ */
+function dosomething_rogue_send_reaction_to_rogue($fid, $user = NULL) 
+{
+  // Get rogue item id by fid
+  $rogue_item_id = dosomething_rogue_get_by_file_id($fid)[0]->rogue_item_id;
+  
+  // Only try to send to Rogue if 
+  if (!is_null($rogue_item_id))
+  {
+    // Get the user who left the reaction
+    if (!isset($user)) {
+      global $user;
+    }
+
+    $northstar_id = dosomething_user_get_field('field_northstar_id', $user);
+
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -255,3 +255,60 @@ function dosomething_rogue_rb_exists_in_rogue($rbid)
...
+        stathat_send_ez_count('drupal - Rogue - reaction sent - count', 1);
+      }
+      if (!$response) {
+        // This is a 404
+        if (module_exists('stathat')) {
+          stathat_send_ez_count('drupal - Rogue - reaction failed - count', 1);
+        }
+      }
+    }
+    catch (GuzzleHttp\Exception\ServerException $e) {
```
angaither, line 302: since we don't do anything with the different types of exceptions, we could just catch all exceptions and send a failed count.

angaither: also, I'm wondering if it's useful to watchdog errors here as well

katiecrane: I think watchdog would be nice to have for local development, I think I should add that in

katiecrane: @angaither wait I thought all the types of exceptions that get caught have to have their own `catch`?

angaither: @katiecrane should be able to `catch (Exception $e)` to pokemon

katiecrane: @angaither THAT WAS SO CLEVER!! I thought I wasn't supposed to do that and was only supposed to catch the specific ones here because these are the ones that can be handled like this?

katiecrane:
we've decided to hold off on this probably until Rogue powers the gallery 🎨

PR #7198, opened 2016-11-17
Sends Reportback from Phoenix API -> Rogue 

### LINKED ISSUE #7183: Rogue to accept rbs that are sent to Phoenix via API.
_FEATURE OVERVIEW_
-------------------
#### User Story
As a developer, I want Rogue to accept reportbacks that are sent to phoenix via API so that we're collecting all reportbacks.

#### Why This Matters
We want to collect ALL reportbacks (some things, like Gambot and the DoSomething Mobile App use Phoenix API to submit reportbacks).

chloealee:
#### What's this PR do?
- When an app hits Phoenix's `/api/v1/campaigns/[nid]/reportback` endpoint, instead of saving directly to Phoenix, it will send along to Rogue first and then Rogue will send back to Phoenix to be saved.
- If crop values aren't set, don't send to Rogue (both on first send and on cron job).

- Hit the `/api/v1/campaigns/[nid]/reportback` endpoint.
screen shot 2016-11-17 at 3 51 11 pm
- Check the Rogue `reportback_items` table for the reportback item.
- Check the Phoenix `dosomething_reportback_file` table for the reportback item. 
- Check that Phoenix's `dosoemthing_rogue_reportbacks` table has the reportback item with all correct data.

#### Relevant tickets
Fixes #7183 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -387,7 +387,7 @@ function _campaign_resource_reportback($nid, $values) {
     // Make sure the reportback exists (meaning it got back from Rogue).
     $rbid = dosomething_reportback_exists($rogue_reportback['data']['campaign_id'], $rogue_reportback['data']['campaign_run_id'], $rogue_reportback['data']['drupal_id']);
 
-    // Store reference to the rb in rogue, redirect user to permalink page.
+    // Store reference to the rb in rogue.
     if ($rbid) {
       $reportback = entity_load_unchanged('reportback', [$rbid]);
       $fid = array_pop($reportback->fids);
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -343,11 +343,12 @@ function _campaign_resource_reportback($nid, $values) {
     }
   }
 
-  $values['fid'] = $file->fid;
   $values['file'] = dosomething_helpers_get_data_uri_from_fid($file->fid);
 
   // @todo: Move this logic into dosomething_reportback_save().
-  $rbid = dosomething_reportback_exists($nid, NULL, $uid);
+  if ($file) {
+    $values['fid'] = $file->fid;
+  }
 
   if (!$rbid) {
     $rbid = 0;
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -364,7 +364,7 @@ function _campaign_resource_reportback($nid, $values) {
   $values['fid'] = $file->fid;
   $values['file'] = dosomething_helpers_get_data_uri_from_fid($file->fid);
 
-  // // @todo: Move this logic into dosomething_reportback_save().
+  // @todo: Move this logic into dosomething_reportback_save().
   $rbid = dosomething_reportback_exists($nid, NULL, $uid);
 
   if (!$rbid) {
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -344,6 +344,7 @@ function _campaign_resource_reportback($nid, $values) {
       }
     }
   }
+
   $values['file'] = dosomething_helpers_get_data_uri_from_fid($file->fid);
 
   if (!$rbid) {
@@ -368,9 +369,7 @@ function _campaign_resource_reportback($nid, $values) {
 
     // Store reference to the rb in rogue.
     if ($rbid) {
-      $reportback = entity_load_unchanged('reportback', [$rbid]);
-      $fid = array_pop($reportback->fids);
-      dosomething_rogue_store_rogue_references($rbid, $fid, $rogue_reportback);
+      dosomething_rogue_store_rogue_references($rbid, $rogue_reportback);
     }
   }
   else {
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -346,9 +346,7 @@ function _campaign_resource_reportback($nid, $values) {
   $values['file'] = dosomething_helpers_get_data_uri_from_fid($file->fid);
 
   // @todo: Move this logic into dosomething_reportback_save().
-  if ($file) {
-    $values['fid'] = $file->fid;
-  }
+  $rbid = dosomething_reportback_exists($nid, NULL, $uid);
 
   if (!$rbid) {
     $rbid = 0;
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -380,7 +380,17 @@ function _campaign_resource_reportback($nid, $values) {
   $transaction_id = $_SERVER['HTTP_X_REQUEST_ID'];
 
   if (! $transaction_id) {
-    dosomething_rogue_send_reportback_to_rogue($values, $user);
+    $rogue_reportback = dosomething_rogue_send_reportback_to_rogue($values, $user);
+
+    // Make sure the reportback exists (meaning it got back from Rogue).
+    $rbid = dosomething_reportback_exists($rogue_reportback['data']['campaign_id'], $rogue_reportback['data']['campaign_run_id'], $rogue_reportback['data']['drupal_id']);
+
+    // Store reference to the rb in rogue, redirect user to permalink page.
+    if ($rbid) {
+      $reportback = entity_load_unchanged('reportback', [$rbid]);
+      $fid = array_pop($reportback->fids);
+      dosomething_rogue_store_rogue_references($rbid, $fid, $rogue_reportback);
+    }
   }
   else {
     dosomething_reportback_save($values, $user);
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -318,6 +318,7 @@ function _campaign_resource_reportback($nid, $values) {
 
   $file = NULL;
 
+
   // @todo: Move this logic into dosomething_reportback_save().
   $rbid = dosomething_reportback_exists($nid, NULL, $uid);
 
@@ -343,11 +344,7 @@ function _campaign_resource_reportback($nid, $values) {
       }
     }
   }
-
   $values['file'] = dosomething_helpers_get_data_uri_from_fid($file->fid);
-
-  // @todo: Move this logic into dosomething_reportback_save().
-  $rbid = dosomething_reportback_exists($nid, NULL, $uid);
 
   if (!$rbid) {
     $rbid = 0;
@@ -371,7 +368,9 @@ function _campaign_resource_reportback($nid, $values) {
 
     // Store reference to the rb in rogue.
     if ($rbid) {
-      dosomething_rogue_store_rogue_references($rbid, $rogue_reportback);
+      $reportback = entity_load_unchanged('reportback', [$rbid]);
+      $fid = array_pop($reportback->fids);
+      dosomething_rogue_store_rogue_references($rbid, $fid, $rogue_reportback);
     }
   }
   else {
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -353,11 +353,9 @@ function _campaign_resource_reportback($nid, $values) {
     $rbid = 0;
   }
 
-  $values['rbid'] = $rbid;
-
-  if (DOSOMETHING_REPORTBACK_LOG) {
-    watchdog('dosomething_api', '_campaign_resource_reportback:' . json_encode($values));
-  }
+  // if (DOSOMETHING_REPORTBACK_LOG) {
+  //   watchdog('dosomething_api', '_campaign_resource_reportback:' . json_encode($values));
+  // }
   return dosomething_rogue_send_reportback_to_rogue($values, $user);
   // return dosomething_reportback_save($values, $user);
 }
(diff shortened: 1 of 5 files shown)


chloealee:
Dropping this here before I forget! Before I merge, should I notify anyone about this? I think Aaron is out so not sure who else to mention it to?

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -352,6 +352,25 @@ function _campaign_resource_reportback($nid, $values) {
     watchdog('dosomething_api', '_campaign_resource_reportback:' . json_encode($values));
   }
 
-  return dosomething_reportback_save($values, $user);
+  $transaction_id = $_SERVER['HTTP_X_REQUEST_ID'];
+
+  if (! $transaction_id) {
```
angaither, line 361: can you add some comments here to explain why we are only sending the reportback if there isn't a transaction id

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -71,14 +70,17 @@ function dosomething_rogue_send_reportback_to_rogue($values, $user = NULL) {
...
     'status' => isset($values['status']) ? $values['status'] : 'pending',
-    'crop_x' => $values['crop_x'],
-    'crop_y' => $values['crop_y'],
-    'crop_width' => $values['crop_width'],
-    'crop_height' => $values['crop_height'],
-    'crop_rotate' => $values['crop_rotate'],
     'source' => isset($values['source']) ? $values['source'] : NULL,
   ];
 
+  if ($values['crop_x']) {
```
angaither: since we are changing this code in multiple places it might make sense to pull this out into a helper function that prepares the item before sending to rouge, but you can do a follow up PR after merge if you want

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -352,6 +352,25 @@ function _campaign_resource_reportback($nid, $values) {
...
+    $rogue_reportback = dosomething_rogue_send_reportback_to_rogue($values, $user);
+
+    // Make sure the reportback exists (meaning it got back from Rogue).
+    $rbid = dosomething_reportback_exists($rogue_reportback['data']['campaign_id'], $rogue_reportback['data']['campaign_run_id'], $rogue_reportback['data']['drupal_id']);
+
+    // Store reference to the rb in rogue, redirect user to permalink page.
+    if ($rbid) {
```
angaither, line 368: what happens if there's no rbid?

angaither (changes requested):

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
chloealee: yes good call!

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
chloealee: would it make sense to do a watchdog error here? This would mean that Rogue's `POST` wouldn't have been successful.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
chloealee: ah yes, great call! Since the two are structured a bit differently (ie `$data['crop_x']` vs `$values['crop_x']` and are set to an array vs object `$values['crop_x']` vs `$task->crop_x`) would it be cleanest to write a helper function with conditionals checking if it's in the `$data` format, set to array, else set to the `$task` object?

angaither: or cast the array as an object when you pass it in so you always are dealing with the same format

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
angaither: doesn't `rbid` not exist if it's not already saved to the phoenix DB?

I meant more what's the case where the transaction ID doesn't exist & the reportback id doesn't exist, doesn't that need to save the reportback?

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
chloealee: ah i see, thank you!

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
chloealee: oh yes I see what you mean now, @angaither ! So if there is no transaction ID, then it sends it to Rogue and Rogue sends back to this function and should now have a transaction id. Since it now has a transaction id, it'll move on to the `else` statement and then save.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
chloealee: @angaither this isn't actually working - I will tag you when it is!

chloealee: @angaither I'm having trouble wrapping my head around this for some reason, can I do this in a separate PR like you suggested above in order to not hold up this PR from being merged?

chloealee: actually I think I got something to work!

chloealee: @angaither I'm playing around with this and think it might make it a little confusing to pull out to a helper function to prepare `$data` to send since here, we'd have to send many different variables to the function (`$northstar_id`, `$user`, `$values`, and `$run`) while in `dosomethign_rogue.cron.inc`, it gets these values from the table/from the `$task` object. I think it makes more sense to keep as is unless I'm not understanding correctly?

angaither: @chloealee the reason I'd advocate for pulling it out is we are doing it in multiple places, and I will also be creating a similar object for the migration script. 
We can pass in an array or an object with all the variables and a common function can prepare it 

Similar to[ `dosomething_northstar_create_user`](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module#L226)

chloealee: @angaither ah okay makes sense - thank you for the example!

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -338,8 +337,9 @@ function _campaign_resource_reportback($nid, $values) {
   }
 
   $values['fid'] = $file->fid;
+  $values['file'] = dosomething_helpers_get_data_uri_from_fid($file->fid);
 
-  // @todo: Move this logic into dosomething_reportback_save().
+  // // @todo: Move this logic into dosomething_reportback_save().
```
angaither: double commented here

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -352,6 +352,27 @@ function _campaign_resource_reportback($nid, $values) {
     watchdog('dosomething_api', '_campaign_resource_reportback:' . json_encode($values));
   }
 
-  return dosomething_reportback_save($values, $user);
+  $transaction_id = $_SERVER['HTTP_X_REQUEST_ID'];
+
+  // If there is no transaction id, the request is coming from an app other than Rogue.
```
angaither: I'd add (mobile app, or sms reportbacks) here to be super clear in the comment

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -263,3 +246,59 @@ function dosomething_rogue_rb_exists_in_rogue($rbid)
...
+ * Transform reportback into the appropriate scheme to send to Rogue.
+ *
+ * @param $northstar_id - User's Northstar user id.
+ * @param $user - Drupal user object.
+ * @param $values - Reportback values to send to Rogue.
+ * @param $run - Campaign run id.
+ * @return array
+ */
+function dosomething_rogue_transform_reportback($northstar_id, $user, $values, $run)
+{
```
angaither: this bracket belongs on the row above in drupal, there's a few other times in this PR it's wrong

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -263,3 +246,59 @@ function dosomething_rogue_rb_exists_in_rogue($rbid)
 {
   return db_query("SELECT * FROM {dosomething_rogue_reportbacks} rogue_rbs WHERE rbid = :rbid", array(':rbid' => $rbid))->fetchAll();
 }
+
+/**
+ * Transform reportback into the appropriate scheme to send to Rogue.
+ *
+ * @param $northstar_id - User's Northstar user id.
```
angaither: you should probably only send one user value in, either ns id, or drupal user. I know we are starting to prefer NS id, so it's probably safe to only send that.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -263,3 +246,59 @@ function dosomething_rogue_rb_exists_in_rogue($rbid)
...
+ * @param $user - Drupal user object.
+ * @param $values - Reportback values to send to Rogue.
+ * @param $run - Campaign run id.
```
angaither: the run should be a part of the values array like the other vars

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
```diff
@@ -29,33 +29,37 @@
           $mime_split = explode(':', $task->file);
           $task->file = $mime_split[0] . ':' . $mimetype . $mime_split[1];
         }
-        $values = [
+
+        $data = [
           'nid' => $task->campaign_id,
```
angaither: should this section be updated to use the transform function as well?

angaither (changes requested):

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
chloealee: since there is a small percentage of people that do not have NS Ids, should I do a check before this helper function to send `$northstar_id` if there is one and if not, send the `$user`?

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
chloealee: I was getting a little caught up on this because in order to send it to `dosomething_rogue_transform_reportback`, we'd be sending in only the `$task` object to fill out `$data` array and seemed to not match up exactly so thought it was cleaner just to keep as is?

angaither: hmmm, I think it makes it a little less clear if there's multiple ways this data can be formatted though

chloealee: ohhh okay I get what you are saying now - will change!

chloealee: @angaither sorry - confusion again! Looking at this, I think it's okay to leave as is since it eventually will get to the transformer anyway? The flow for the cron job is that it gets sent to `dosomething_rogue_send_reportback_to_rogue` [here](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc#L49), and in this function, it calls `dosomething_rogue_transform_reportback` so that data will just get formatted eventually anyway. LMK what you think!

chloealee: @angaither i did do some tweaking that should make it better though so I'll push that up as well!

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
```diff
@@ -29,33 +29,38 @@
...
           'file' => $task->file,
           'caption' => $task->caption,
           'type' => $task->type,
-          'crop_x' => $task->crop_x,
-          'crop_y' => $task->crop_y,
-          'crop_width' => $task->crop_width,
-          'crop_height' => $task->crop_height,
-          'crop_rotate' => $task->crop_rotate,
         ];
+        // $data = dosomething_rogue_transform_reportback($task);
```
angaither: this line is still commented out?

chloealee: oh right - I will delete that! Just had as a placeholder if we wanted to send to the helper function.

angaither: maybe we are just going in circles here, but shouldn't that whole chunk be using the helper function?

chloealee: @angaither omg YES sorry this took me so long to wrap my head around! All fixed :)

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -410,13 +410,17 @@ function dosomething_reportback_form_submit($form, &$form_state) {
...
     $rbid = dosomething_reportback_exists($rogue_reportback['data']['campaign_id'], $rogue_reportback['data']['campaign_run_id'], $rogue_reportback['data']['drupal_id']);
 
     // Store reference to the rb in rogue, redirect user to permalink page.
     if ($rbid) {
       $reportback = entity_load_unchanged('reportback', [$rbid]);
-      $fid = array_pop($reportback->fids);
-      dosomething_rogue_store_rogue_references($rbid, $fid, $rogue_reportback);
+      // If a file was added, store references to the relevant Rogue data
+      if (isset($file)) {
+        // $fid = array_pop($reportback->fids);
```
angaither: is this supposed to be commented out still?

chloealee:
closing with fixes from #7227

PR #7202, opened 2016-11-18
If phoenix does not have ns id, grab it directly from ns

### LINKED ISSUE #7201: Get NS user info from NS if it doesn't exist in Phoenix
_BUG_
-------------------
#### Current Behavior

Some users are not making it into the `authmap` table when they are created and so phoenix is unaware of their northstar id. This causes issues in Rogue when the user reports back because we try to post a `null` northstar id to Rogue, which is not allowed.

#### Desired Behavior

If phoenix does not return a NS id then we should go directly to NS to get the Id. 

#### Why This Matters

- Rogue does not get the reportback

- We keep trying and trying it causing many errors.

sbsmith86:
#### What's this PR do?

If phoenix is unaware of a user's Northstar ID, then we grab the user's NS id directly from northstar. 

Fixes new relic errors we are seeing in Rogue. There is an API error that is causing Phoenix to not store a user's Northstar Id. When the user reports back we try to send a `null` Northstar id to rogue, the request fails, the request goes in the failed log table with a `null` NS id and we keep trying it and it keeps failing. Rogue does not accept `null `values for Northstar id.

The "Bandaid" fix to handle the rogue error alerts is to grab the northstar id directly from northstar if it doesn't exists in phoenix.

#### Relevant tickets
Fixes #7201

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
@@ -57,12 +57,20 @@ function dosomething_rogue_send_reportback_to_rogue($values, $user = NULL) {
 
   $northstar_id = dosomething_user_get_field('field_northstar_id', $user);
 
+  // Band-aid fix for an issue we are seeing with phoenix not being
+  // aware of a user's northstar id. If it doesn't find one, we just grab it
+  // from northstar directly.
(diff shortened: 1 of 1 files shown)


angaither:
🚢

PR #7204, opened 2016-11-18
Pass user to rogue

### LINKED ISSUE #7203: Explicitly pass user to dosomething_rogue_send_reportback_to_rogue().
_BUG_
-------------------
#### Current Behavior

When an admin tries to update a users reportback, it tries to update the admins reportback for the campaign instead of the correct user.

More explanation: https://trello.com/c/RPIuhfnp/202-explicitly-pass-the-user-we-are-updating-the-reportback-for-to-dosomething-rogue-send-reportback-to-rogue

#### Desired Behavior

Admin can update a users reportback 

#### Steps to Replicate

Try to edit a users reportback and it will either fail because you have never reported back for the campaign, or it will actually update your reportback instead.

sbsmith86:
#### What's this PR do?
 
Passes reportback user to rogue otherwise it defaults to the `global $user`

As an admin, find a reportback and try to update something, the quantity for example. It should update the quantity for that user.

More info is in the ticket and trello card, referenced below.

#### Relevant tickets
Fixes #7203 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -407,7 +407,7 @@ function dosomething_reportback_form_submit($form, &$form_state) {
     }
 
     // Send the reportback to rogue.
-    $rogue_reportback = dosomething_rogue_send_reportback_to_rogue($values);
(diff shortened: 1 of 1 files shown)


angaither:
💯

PR #7205, opened 2016-11-18
check if failed rbs are missing mimetype and add

### LINKED ISSUE #7200: files missing mimetype in Rogue failed logs will never make it
_BUG_
-------------------
#### Current Behavior
Some items in the `dosomething_rogue_failed_task_log` table are doomed to fail forever unless we rescue them! These items have `file`s that start with `data:;base64`. 

#### Desired Behavior
We need to pop in their mimetype if they ever want to make it to Rogue! We want to add their mimetype in during the cron job so that they will make it to Rogue.

#### Why This Matters
We don't want to leave these dudes in this table forever!

katiecrane:
#### What's this PR do?
Before trying to send a failed reportback to Rogue, check to see if the file has the MIME type listed. If not, find out what it is and rebuild the file string to contain the MIME type.

Make sure that failed reportbacks with and without MIME types all make it to Rogue when cron runs.

#### Relevant tickets
Fixes #7200 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
@@ -16,6 +16,19 @@
 
     foreach ($task_log as $task) {
       if ($task->type === 'reportback') {
+        // Check to see if the MIME type is missing
+        if (strpos($task->file, 'data:;') !== false) {
+          // Split file string to access the data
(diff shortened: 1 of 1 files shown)


angaither:
🎱

mshmsh5000:
> _We don't want to leave these dudes in this table forever!_

Love this PR & the issue that made it possible, thank you to the world

PR #7206, opened 2016-11-18
When attempting to get NS id, add type=drupal id.

angaither:
#### What's this PR do?
sends type when calling the function

👀 

this function was updated to prefer NS id, and wasn't working without this

#### Relevant tickets
Fixes #

#### Checklist
- [ ] Tested on staging.


sbsmith86:
yes. sweet. grand. 💯

PR #7208, opened 2016-11-21
Restrict Phoenix registration endpoint & save Northstar ID for API users.

DFurnes:
#### What's this PR do?
This PR makes some updates to fix issues with users created via the API:

🔒 It locks down the `POST /api/users` to only allow admin accounts to create users. This shouldn't be a problem because all user registrations _shoooould_ be happening via Northstar by now.

➡️ It requires the `_id` parameter to be included, and saves it in `field_northstar_id` and the `authmap` table so the Drupal account can be linked to it's corresponding Northstar user.

🚳 Removes code to forward users created via Drupal endpoint to Northstar, because we only want this endpoint to be used _by Northstar_ from now on. This will help simplify things and make it easier to think about what's going on.

👀 and 🤔. I've tested this on my local, but would appreciate an extra sanity check. I'll also post in #phoenix to make sure that nobody else can think of anything that is relying on this endpoint.

There's a corresponding [Northstar pull request](https://github.com/DoSomething/northstar/pull/499) that needs to be deployed before this can safely be merged.

#### Relevant tickets
See attached Trello card.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/documentation/endpoints/users.md
+++ b/documentation/endpoints/users.md
@@ -2,11 +2,10 @@
 
 ## Create a user
 
-Create a new Phoenix user with the given profile information. Consider using Northstar's [Create User](https://github.com/DoSomething/northstar/blob/dev/documentation/endpoints/users.md#create-a-user)
-or [Register](https://github.com/DoSomething/northstar/blob/dev/documentation/endpoints/auth.md#register-user) endpoints
-instead, as this endpoint will likely be removed in the future. 
+Create a new Phoenix user with the given profile information.
 
-This endpoint is currently available to anonymous users.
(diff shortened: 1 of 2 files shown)


DFurnes:
[(4) As a user, I want my Phoenix & NS accounts connected](https://trello.com/c/GaB9z96k/177-4-as-a-user-i-want-my-phoenix-ns-accounts-connected)

DFurnes:
Thanks to @mshmsh5000 for thinking to check logs.dosomething.org – confirmed that no applications besides Northstar are using this Phoenix endpoint so should be safe to merge! 🚗

angaither:
:shipit:

PR #7209, opened 2016-11-21
Make sure to set Drupal ID field on NS profile for OpenID Connect users.

DFurnes:
#### What's this PR do?
When users register as part of the OpenID Connect flow, their Northstar account is created via Northstar's web interface (rather than by a Phoenix API call). This means they will never get the `drupal_id` field set on their profile.

This pull request adds a step to the "post-authorize" hook which updates the given Northstar ID with the user's Drupal UID, so we can continue to rely on that being available.

👀

We obviously want to minimize use of the `drupal_id` field, but it's unfortunately still required for a lot of core experiences (like signing up or reporting back for a campaign via the mobile app).

#### Relevant tickets
See linked Trello card.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
@@ -122,6 +122,13 @@ function dosomething_northstar_openid_connect_post_authorize($tokens, $account, 
     dosomething_northstar_save_id_field($account->uid, ['data' => $userinfo]);
   }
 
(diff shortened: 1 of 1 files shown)


DFurnes:
[As a dev, I want all NS users to receive its associated drupal_id so that the accounts are connected.](https://trello.com/c/9BqYGlmw/185-as-a-dev-i-want-all-ns-users-to-receive-its-associated-drupal-id-so-that-the-accounts-are-connected)

PR #7219, opened 2016-11-29
allow reportback updates with no file via Phoenix api

### LINKED ISSUE #7218: campaigns/[nid]/reportback endpoint doesn't allow for updates with no image
_BUG_
-------------------
#### Current Behavior
If you try to update an existing reportback's `quantity` or `why participated` via the API (aka with no image) you receive a `500` with `Reportback file_url not found.` 

#### Desired Behavior
While it's true that Reportback file_url is not found, it shouldn't necessarily need to be. A file should only be required when creating a new reportback.

#### Steps to Replicate
Create a reportback via the API. Then try to update just the quantity or the why, also via the API.

#### Why This Matters
Rogue (and I'm guessing other services) need to be able to send these kinds of updates to Phoenix.

katiecrane:
#### What's this PR do?
Allows reportbacks to update just their `quantity` or `why participated` via the API. It still checks if a reportback already exists and does still require an image on a new reportback. Otherwise, it just checks to see if there is a file it needs to include.

Make sure the following tasks work as expected via phoenix web and the API:
1. Attempting to submit a new reportback without an image (API should return a `500` with `Reportback file_url not found.`)
2. Submitting a valid new reportback
3. Submitting an update to just the `quantity` or `why participated` of an existing reportback
4. Submitting an update to a reportback that includes a new image

I want to make extra sure this doesn't break anything else using the API. I thiiiink in theory it should all be okay because it relaxes the requirements.

This will allow https://github.com/DoSomething/rogue/issues/76 to be fixed since Rogue currently is sending the most recent image because it has to send something.

#### Relevant tickets
Fixes #7218

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/documentation/endpoints/campaigns.md
+++ b/documentation/endpoints/campaigns.md
@@ -312,9 +312,9 @@ Must be authenticated to post.
     The number of reportback nouns verbed.
   - **why_participated**: (string) required.
     The reason why the user participated.
-  - **file_url**: (string) required if `file` is not provided.
+  - **file_url**: (string) required if `file` is not provided and a new reportback is being created.
     An image URL for the reportback.
-  - **file**: (string) required if `file_url` is not provided.
+  - **file**: (string) required if `file_url` is not provided and a new reportback is being created.
     Base64 encoded file string to save.
   - **filename**: (string) required if `file` is provided.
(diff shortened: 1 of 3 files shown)


chloealee:
this looks great to me!

angaither:
🎱

angaither (approved):

PR #7224, opened 2016-11-30
store references to failed Rogue rbs

### LINKED ISSUE #7223: store references to reportbacks send to Rogue via cron
_BUG_
-------------------
#### Current Behavior
If a reportback from the `dosomething_rogue_failed_task_log` is successfully sent to Rogue via the cron job, its info does not get stored in the `dosomething_rogue_reportbacks` table.

#### Desired Behavior
If a reportback from the `dosomething_rogue_failed_task_log` is successfully sent to Rogue via the cron job, its info gets stored in the `dosomething_rogue_reportbacks` table.

#### Steps to Replicate
Run cron and see the failed reportbacks show up in Rogue but not in `dosomething_rogue_reportbacks`.

#### Why This Matters
We want to keep track of these!

katiecrane:
#### What's this PR do?
Adds reportbacks that were successfully send to Rogue via cron to the `dosomething_rogue_reportbacks` table. Includes some helper methods to expedite that both in cron and when we send reportbacks to Rogue via the web form.

Make sure you can submit reportbacks normally. 
Submit a reportback that ends up in the failed table. Run cron. Make sure that reportback ends up both in Rogue and in the `dosomething_rogue_reportbacks` table.

We need all these little dudes to live in this table so they know that they made it! 

#### Relevant tickets
Fixes #7223

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
@@ -46,7 +46,9 @@
 
         $user = user_load($task->drupal_id);
 
-        dosomething_rogue_send_reportback_to_rogue($values, $user);
+        $rogue_reportback = dosomething_rogue_send_reportback_to_rogue($values, $user);
+
+        dosomething_rogue_check_rbid_and_store_ref($rogue_reportback);
       } else {
         $values = [
           [
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
@@ -267,17 +267,17 @@ function dosomething_rogue_rb_exists_in_rogue($rbid) {
  *
  */
 function dosomething_rogue_check_rbid_and_store_ref($rogue_reportback) {
+  // @TODO: this check is really only good for a reportback that has just been created, it doesn't tell us much regarding updates
   $rbid = dosomething_reportback_exists($rogue_reportback['data']['campaign_id'], $rogue_reportback['data']['campaign_run_id'], $rogue_reportback['data']['drupal_id']);
 
   // Make sure the reportback exists (meaning it got back from Rogue)
   if ($rbid) {
     // Store reference to the rb in rogue
    $reportback = entity_load_unchanged('reportback', [$rbid]);
-    // If a file was added, store references to the relevant Rogue data
-    if (isset($file)) {
-      $fid = array_pop($reportback->fids);
-      dosomething_rogue_store_rogue_references($rbid, $fid, $rogue_reportback);
-    }
+
+    // Store references to the relevant Rogue data
+    $fid = array_pop($reportback->fids);
+    dosomething_rogue_store_rogue_references($rbid, $fid, $rogue_reportback);
 
     $data = [
       'rbid' => $rbid,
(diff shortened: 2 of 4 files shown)


lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -263,3 +263,35 @@ function dosomething_rogue_rb_exists_in_rogue($rbid)
...
+
+/**
+ * Checks to see if the given RB data exists in Phoenix and if so stores in Rogue reference table.
+ *
+ * @param array $rogue_reportback
+ * The response from dosomething_rogue_send_reportback_to_rogue().
+ *
+ */
+function dosomething_rogue_check_rbid_and_store_ref($rogue_reportback)
+{
```
angaither: ⬆️  brackets go up in drooops

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
```diff
@@ -46,7 +46,9 @@
 
         $user = user_load($task->drupal_id);
 
-        dosomething_rogue_send_reportback_to_rogue($values, $user);
+        $rogue_reportback = dosomething_rogue_send_reportback_to_rogue($values, $user);
+
+        dosomething_rogue_check_rbid_and_store_ref($rogue_reportback);
```
angaither, line 41: does it matter if this returns false here?

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
katiecrane: oh boy they sure do

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -263,3 +263,35 @@ function dosomething_rogue_rb_exists_in_rogue($rbid)
...
+  $rbid = dosomething_reportback_exists($rogue_reportback['data']['campaign_id'], $rogue_reportback['data']['campaign_run_id'], $rogue_reportback['data']['drupal_id']);
+
+  // Make sure the reportback exists (meaning it got back from Rogue)
+  if ($rbid) {
+    // Store reference to the rb in rogue
+   $reportback = entity_load_unchanged('reportback', [$rbid]);
+    // If a file was added, store references to the relevant Rogue data
+    if (isset($file)) {
```
angaither: I'm trying to imagine how/if this function will work with a migration script, where the reportback and file already exist in drupal, so maybe all I need to do is call `dosomething_rogue_store_rogue_references` ?

katiecrane: Yeah I think that once it exists in both places and you have all the needed info that is what you would want to call

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
katiecrane: @angaither hmm good question. I guess in this particular case we don't care what it returns because `dosomething_rogue_send_reportback_to_rogue` takes care of stuff ending up in the failed table. But elsewhere we want to know if it is false so we can hit the permalink failure page.

angaither (approved):

angaither:
@katiecrane can you synccccc with @chloealee on her PR, I think this one will cause conflicts with hers

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
katiecrane: I just noticed that `$file` doesn't seem to actually get set. fixing now!

katiecrane: actually, I think this check for a new file can go altogether, because each reference is for an item so we don't want to call this function if there is no new file

angaither:
📷

angaither:
🚥 🍏

Issue #7225, opened 2016-11-30
Move reportbacks


angaither:
All legacy reportbacks live in phoenix.
They should also live in rogue, to eventually turn phoenix reportbacks off

mirie:
This has been handled / completed in other issues.

PR #7226, opened 2016-11-30
Migrate reportback data

angaither:
#### What's this PR do?
- added a function to transform what a phoenix status is to a rogue status
- added script to migrate reportback data to phoenix

👀 

I still have a few todos/checks
- Once in rogue, does this retry sending back to phoenix? Can we disable that?
- If the script fails and we need to re-migrate all info, the call to `dosomething_rogue_store_rogue_references` will fail because it doesn't check for duplicate rows

#### Relevant tickets
Refs #7225

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/scripts/export-reportbacks-to-rogue.php
+++ b/scripts/export-reportbacks-to-rogue.php
@@ -12,27 +12,29 @@ include_once('../lib/modules/dosomething/dosomething_northstar/dosomething_north
 
 $last_saved = variable_get('dosomething_rogue_last_rbi_migrated', NULL);
 if ($last_saved) {
-  $rbis = db_query("SELECT *
+  $rbis = db_query("SELECT rb.uid
             FROM dosomething_reportback_file rbf
-            LEFT JOIN dosomething_reportback rb on rbf.rbid = rb.rbid
+            INNER JOIN dosomething_reportback rb on rbf.rbid = rb.rbid
             WHERE rbf.fid > $last_saved
-            ORDER BY rbf.fid");
+            ORDER BY rbf.fid
+            LIMIT 1");
 }
 else {
-  // Get all the users!
+  // Get all the things!
   $rbis = db_query('SELECT *
                    FROM dosomething_reportback_file rbf
-                   LEFT JOIN dosomething_reportback rb on rbf.rbid = rb.rbid
-                   ORDER BY rbf.fid');
+                   INNER JOIN dosomething_reportback rb on rbf.rbid = rb.rbid
+                   ORDER BY rbf.fid
+                   LIMIT 15');
 }
 
 $client = dosomething_rogue_client();
 
 
-foreach ($rbis as $rbi) {
+foreach ($rbis as $rb) {
 
   $data = [
-    'northstar_id' => dosomething_northstar_get_user($rb->uid, 'drupal_id'),
+    'northstar_id' => dosomething_northstar_get_user($rb->uid, 'drupal_id')->id,
     'campaign_id' => $rb->nid,
     'campaign_run_id' => $rb->run_nid,
     'quantity' => $rb->quantity,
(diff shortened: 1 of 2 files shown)


scripts/export-reportbacks-to-rogue.php
```diff
@@ -0,0 +1,61 @@
...
+
+
+$last_saved = variable_get('dosomething_rogue_last_rbi_migrated', NULL);
+if ($last_saved) {
+  $rbis = db_query("SELECT rb.uid
+            FROM dosomething_reportback_file rbf
+            INNER JOIN dosomething_reportback rb on rbf.rbid = rb.rbid
+            WHERE rbf.fid > $last_saved
+            ORDER BY rbf.fid
+            LIMIT 1");
```
sbsmith86: Does this mean you only get 1 record back?

scripts/export-reportbacks-to-rogue.php
```diff
@@ -0,0 +1,61 @@
...
+    'status' => dosomething_rogue_transform_status($rb->status),
+  ];
+
+  $response = $client->postReportback($data);
+
+  // Output progress to see what's going on.
+  echo 'Migrated reportback ' . $rb->fid . ' to Rogue [' . $rb->caption . ']' . PHP_EOL;
+
+  // Store the reference in dosomething_rogue_reportbacks.
+  dosomething_rogue_store_rogue_references($rb->rbid, $rb->fid, $response);
```
katiecrane: if the post fails I think this is going to break because `$response` won't be anything

scripts/export-reportbacks-to-rogue.php
angaither: oh yeah look at that, I left in the test code

scripts/export-reportbacks-to-rogue.php
angaither: @katiecrane is it nothing, or an is there an error code I should check for?

scripts/export-reportbacks-to-rogue.php
```diff
@@ -0,0 +1,62 @@
...
+    'file' => dosomething_helpers_get_data_uri_from_fid($rb->fid),
+    'caption' => $rb->caption,
+    'source' => $rb->source,
+    'remote_addr' => $rb->remote_addr,
+    'status' => dosomething_rogue_transform_status($rb->status),
+  ];
+
+  $response = $client->postReportback($data);
+
+  if ($response) {
```
katiecrane: @angaither I think I was actually a little wrong before. My understanding is `$response` is only `null` for a `404` and other errors have to be caught otherwise it will crash if it gets an error back when posting. [This is how we do it](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module#L93-L112) for the web form

katiecrane: so it wouldn't crash because it's trying to access stuff that isn't there, but because an error was thrown

sbsmith86 (approved):
looks good to me!

sbsmith86:
🎉

scripts/export-reportbacks-to-rogue.php
```diff
@@ -0,0 +1,72 @@
...
+      $response = $client->postReportback($data);
+      // Output progress to see what's going on.
+      echo 'Migrated reportback ' . $rb->fid . ' to Rogue [' . $rb->caption . ']' . PHP_EOL;
+
+      // Store the reference in dosomething_rogue_reportbacks.
+      dosomething_rogue_store_rogue_references($rb->rbid, $rb->fid, $response);
+
+      // If the script fails, we can use this to start the script from the last reportback migrated.
+      variable_set('dosomething_rogue_last_rbi_migrated', $rb->fid);
+    } catch (GuzzleHttp\Exception\ServerException $e) {
```
katiecrane, line 64: I think you need to catch this kind too `DoSomething\Gateway\Exceptions\ApiException`

PR #7227, opened 2016-12-01
Sends Reportback from Phoenix API -> Rogue Take 2

### LINKED ISSUE #7183: Rogue to accept rbs that are sent to Phoenix via API.

chloealee:
#### What's this PR do?
- When an app hits Phoenix's `/api/v1/campaigns/[nid]/reportback` endpoint, instead of saving directly to Phoenix, it will send along to Rogue first and then Rogue will send back to Phoenix to be saved.
- If crop values aren't set, don't send to Rogue (both on first send and on cron job).
- Adds helper functions to prepare rb item to send to Rogue. 
- When cron job runs, insert the rogue rb reference in the `dosomething_rogue_reportbacks` table.

- Hit the `/api/v1/campaigns/[nid]/reportback` endpoint.
- Check the Rogue `reportback_items` table for the reportback item.
- Check the Phoenix `dosomething_reportback_file` table for the reportback item. 
- Check that Phoenix's `dosoemthing_rogue_reportbacks` table has the reportback item with all correct data.
- Test cron job.
- Test that reportingback from phoenix web works.

#### Relevant tickets
#7198 got borked so starting over in this PR.
Fixes #7183 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -344,19 +344,28 @@ function _campaign_resource_reportback($nid, $values) {
     }
   }
 
+  $values['fid'] = $file->fid;
+  $values['file'] = dosomething_helpers_get_data_uri_from_image(image_load($file->uri));
+
   // @todo: Move this logic into dosomething_reportback_save().
-  if ($file) {
-    $values['fid'] = $file->fid;
-  }
+  $rbid = dosomething_reportback_exists($nid, NULL, $uid);
 
   if (!$rbid) {
     $rbid = 0;
   }
 
-  // if (DOSOMETHING_REPORTBACK_LOG) {
-  //   watchdog('dosomething_api', '_campaign_resource_reportback:' . json_encode($values));
-  // }
-  return dosomething_rogue_send_reportback_to_rogue($values, $user);
+  $values['rbid'] = $rbid;
+
+  if (DOSOMETHING_REPORTBACK_LOG) {
+    watchdog('dosomething_api', '_campaign_resource_reportback:' . json_encode($values));
+  }
+
+  $transaction_id = drupal_get_http_header('X-Request-ID');
+  print_r($transaction_id);
+  die();
+  // if the flag is less than 2, send to rogue
+  // dosomething_rogue_send_reportback_to_rogue($values, $user);
+  return $rbid;
   // return dosomething_reportback_save($values, $user);
 }
 
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -343,8 +343,10 @@ function _campaign_resource_reportback($nid, $values) {
     }
   }
 
-  $values['fid'] = $file->fid;
-  $values['file'] = dosomething_helpers_get_data_uri_from_fid($file->fid);
+  if ($file) {
+    $values['fid'] = $file->fid;
+    $values['file'] = dosomething_helpers_get_data_uri_from_fid($file->fid);
+  }
 
   // @todo: Move this logic into dosomething_reportback_save().
   $rbid = dosomething_reportback_exists($nid, NULL, $uid);
@@ -368,7 +370,7 @@ function _campaign_resource_reportback($nid, $values) {
     $rbid = dosomething_reportback_exists($rogue_reportback['data']['campaign_id'], $rogue_reportback['data']['campaign_run_id'], $rogue_reportback['data']['drupal_id']);
 
     // Store reference to the rb in rogue.
-    if ($rbid) {
+    if ($rbid && $file) {
       dosomething_rogue_store_rogue_references($rbid, $rogue_reportback);
     }
   }
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -371,7 +371,9 @@ function _campaign_resource_reportback($nid, $values) {
 
     // Store reference to the rb in rogue.
     if ($rbid && $file) {
-      dosomething_rogue_store_rogue_references($rbid, $rogue_reportback);
+      $reportback = entity_load_unchanged('reportback', [$rbid]);
+      $fid = array_pop($reportback->fids);
+      dosomething_rogue_store_rogue_references($rbid, $fid, $rogue_reportback);
     }
   }
   else {
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -303,7 +303,6 @@ function _campaign_resource_signup($nid, $values) {
  */
 function _campaign_resource_reportback($nid, $values) {
   // @todo: Return error if signup doesn't exist.
-
   $values['nid'] = $nid;
 
   if (!isset($values['uid'])) {
@@ -345,7 +344,7 @@ function _campaign_resource_reportback($nid, $values) {
   }
 
   $values['fid'] = $file->fid;
-  $values['file'] = dosomething_helpers_get_data_uri_from_image(image_load($file->uri));
+  $values['file'] = dosomething_helpers_get_data_uri_from_fid($file->fid);
 
   // @todo: Move this logic into dosomething_reportback_save().
   $rbid = dosomething_reportback_exists($nid, NULL, $uid);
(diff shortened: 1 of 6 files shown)


lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -344,10 +343,11 @@ function _campaign_resource_reportback($nid, $values) {
     }
   }
 
+  $values['fid'] = $file->fid;
+  $values['file'] = dosomething_helpers_get_data_uri_from_fid($file->fid);
+
   // @todo: Move this logic into dosomething_reportback_save().
-  if ($file) {
```
katiecrane, line 348: You need this conditional before you do anything with `$file` to make sure that a file was added

chloealee: ahh I see - added back in!

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -359,6 +362,23 @@ function _campaign_resource_reportback($nid, $values) {
...
+
+  if (! $transaction_id) {
+    $rogue_reportback = dosomething_rogue_send_reportback_to_rogue($values, $user);
+
+    // Make sure the reportback exists (meaning it got back from Rogue).
+    $rbid = dosomething_reportback_exists($rogue_reportback['data']['campaign_id'], $rogue_reportback['data']['campaign_run_id'], $rogue_reportback['data']['drupal_id']);
+
+    // Store reference to the rb in rogue.
+    if ($rbid) {
+      dosomething_rogue_store_rogue_references($rbid, $rogue_reportback);
```
katiecrane: I think we only want to call this function if there is a new image as well, since one of the columns is `fid` (and we also pop the most recent `fid` so if there was no new file we'd end up with one that should already be in there). I wonder if this is what was causing those `fid` related errors before?

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -359,6 +362,23 @@ function _campaign_resource_reportback($nid, $values) {
     watchdog('dosomething_api', '_campaign_resource_reportback:' . json_encode($values));
   }
 
-  return dosomething_reportback_save($values, $user);
+  $transaction_id = $_SERVER['HTTP_X_REQUEST_ID'];
+
+  if (! $transaction_id) {
```
sbsmith86, line 366: I'm not sure I understand this logic. If you have a request _without_ a transaction ID you send to rogue? If you do have one you don't? 

What happens when like the Mobile app sends over RBs when it hasn't implemented transaction ids yet?

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
chloealee: ahh very smart @katiecrane ! updated this!

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -169,13 +150,15 @@ function dosomething_rogue_get_by_file_id($fid)
...
- * @param string $fid
  * @param object $rogue_reportback
  *
  * @return InsertQuery object
  */
-function dosomething_rogue_store_rogue_references($rbid, $fid, $rogue_reportback)
+function dosomething_rogue_store_rogue_references($rbid, $rogue_reportback)
 {
+  $reportback = entity_load_unchanged('reportback', [$rbid]);
+  $fid = array_pop($reportback->fids);
```
katiecrane: hmm I'm wondering if this part should get held off on and then there will be fewer conflicts with my PR? I handled this part differently with `dosomething_rogue_check_rbid_and_store_ref` https://github.com/DoSomething/phoenix/pull/7224/files

what do you think would be the best way to proceed?

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
chloealee: Yes! This is to prevent the infinite loop. 

So when the mobile app or sms sends a reportback, it won't have a transaction id. Since it doesn't have one, we know it comes from those two sources and we want to send to Rogue. 

Then, Rogue receives it and sends back to Phoenix with this same API endpoint. It will now have a transaction id so we don't want it to send to Rogue again and get stuck in an infinite loop. So instead of sending back to Rogue, we just save it in Phoenix.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
chloealee: @katiecrane would it be helpful if I changed it back so that this logic lived outside of `dosomething_rogue_store_rogue_references` and that function still has `$fid` as a param?

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
sbsmith86: hmm ok gotcha...seems like a good short term solution!

angaither: yeah we did something similar, but perhaps more apparent with NS/Phoenix there was a query param `?create_drupal_user` on NS and [`forward`] (https://github.com/DoSomething/phoenix/commit/184063973871bc4d4296c73db19c2bbbb3f79215#diff-1492ad29ad6cef475c92f6ed9d58a7ebL34) on phoenix (which was just removed) to allow this.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
katiecrane: @chloealee this helped me notice an error in my PR oh boy! but I thiiiiink that would be pretty helpful? not if it would be a huge pain though

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -263,3 +246,60 @@ function dosomething_rogue_rb_exists_in_rogue($rbid)
...
+     'quantity' => $values['quantity'],
+     'why_participated' => $values['why_participated'],
+     'file' => isset($values['file']) ? $values['file'] : NULL,
+     'caption' => isset($values['caption']) ? $values['caption'] : NULL,
+     'status' => isset($values['status']) ? $values['status'] : 'pending',
+     'source' => isset($values['source']) ? $values['source'] : NULL,
+     'type' => isset($values['type']) ? $values['type'] : NULL,
+   ];
+
+   if ($values['crop_x']) {
```
sbsmith86, line 278: I created a function once that just extracted properties from an array. It was a little cleaner. So you can do like 

```
/**
 * Given an array of key/value pairs, create a new array with only the
 * keys given.
 *
 * @param  associative array  $data
 * @param  array $desiredKeys
 *
 * @return array
 */
function extract_values_by_keys($array, $desiredKeys)
{
    $desiredValues = array_intersect_key($array, array_flip($desiredKeys));
    return (count($array) > 0) ? $desiredValues : null;
}
```

and then use it like:

```
$cropped_values = extract_values_by_keys($values,['crop_x', 'crop_y'....]); //gives you an array

$data = array_merge($data, $cropped_values);
```

allows you to remove some of the duplicated lines and the extra loop at the end.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
chloealee: @katiecrane oh good! Not a pain at all - all set! Let me know if that is what you were thinking.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
chloealee: @sbsmith86 this is brilliant and a lot more helpful in a general sense - thank you! I popped this in `dosomething_helpers.module`

sbsmith86:
Looks good to me!

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -969,3 +969,18 @@ function dosomething_helpers_get_data_uri_from_fid($fid) {
...
+ * keys given.
+ *
+ * @param  associative array  $data
+ * @param  array $desiredKeys
+ *
+ * @return array
+ */
+function dosomething_helpers_extract_values_by_keys($array, $desiredKeys)
+{
+    $desiredValues = array_intersect_key($array, array_flip($desiredKeys));
```
angaither: 🐍 case vars!

chloealee: ah shoot - thanks! All fixed :)

angaither (approved):
⛵

PR #7229, opened 2016-12-05
Allow Northstar to create users without providing a password.

DFurnes:
#### What's this PR do?
This pull request updates Drupal's "create user" endpoint to set the `created_by_openid_connect` boolean on a user profile if a password was not provided. This will cause logins for that user account to  always be routed to Northstar, and will allow us to sync any Northstar accounts to Phoenix regardless of whether we have the password on hand.

It also fixes a typo when saving refresh token to a user's Phoenix account.

👀

🍃

#### Relevant tickets
References DoSomething/northstar#406.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
@@ -207,6 +207,7 @@ function _member_resource_create($request) {
     'mobile' => !empty($request['mobile']) ? dosomething_user_clean_mobile_number($request['mobile']) : NULL,
     'user_registration_source' => !empty($request['user_registration_source']) ? $request['user_registration_source'] : NULL,
     'northstar_id' => $request['_id'],
+    'created_by_openid_connect' => empty($request['password']) ? 1 : 0, // If no password given, specify that this user *must* login via Northstar.
   ];
 
   dosomething_user_set_fields($edit, $fields);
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_api/resources/member_resource.inc
```diff
@@ -207,6 +207,7 @@ function _member_resource_create($request) {
     'mobile' => !empty($request['mobile']) ? dosomething_user_clean_mobile_number($request['mobile']) : NULL,
     'user_registration_source' => !empty($request['user_registration_source']) ? $request['user_registration_source'] : NULL,
     'northstar_id' => $request['_id'],
+    'created_by_openid_connect' => empty($request['password']) ? 1 : 0, // If no password given, specify that this user *must* login via Northstar.
```
angaither, line 210: does this value `created_by_openid_connect` force a login via NS?

DFurnes: [Yes](https://github.com/DoSomething/phoenix/blob/463f43ec8e37009c57a1d41312f0d2d9954d2c39/lib/modules/dosomething/dosomething_user/dosomething_user.module#L1091-L1094)! That's what allows people to have registered via the OpenID Connect flow but still log in if they return to a different computer later and aren't opted-in.

angaither (approved):
lovely 🔏  🌟

DFurnes:
[Always create a Phoenix account when registering users via Northstar API endpoints.](https://trello.com/c/7zCUSYma/145-always-create-a-phoenix-account-when-registering-users-via-northstar-api-endpoints)

DFurnes:
[Unplanned Work: Fix issues caused by Celebs Gone Good users not having Drupal IDs.](https://trello.com/c/kIvMHGsr/202-unplanned-work-fix-issues-caused-by-celebs-gone-good-users-not-having-drupal-ids)

PR #7234, opened 2016-12-06
Changes check for $transaction_id

chloealee:
#### What's this PR do?
Checks to see if `$transaction_id` is `NULL` instead of `if (! $transaction_id)` check.

👀 

Right now, testing from the API works on local and staging but not on Thor. 

It seems like when reporting back from Thor via API, it seems to think it has a `$transaction_id` so just skips the code in lines 365-370 and just saves to Phoenix directly instead in line 373. 

This `is_null` check could be a more deliberate check to see if there is a `$transaction_id`?

#### Checklist
- [ ] Tested on staging.


chloealee:
Deployed to thor and tested and didn't fix the issue. Closing this PR.

PR #7233, opened 2016-12-06
Fix to api mess

chloealee:
#### What's this PR do?
- Fixes error and gets new reportbacks and new reportback items to send to Rogue from both Phoenix web and Phoenix API. 
  - Problem was if this was a new reportback, the reportback was being saved twice. 
- Cleans up values in `dosomething_reportback.forms.inc` so it's easier to read and removes typo. 

#### How should this be reviewed?
Test that the following work: 

1. Creating a new reportback from Phoenix web.
2. Creating a new reportback item from Phoenix web.
3. Hitting `/api/v1/campaigns/[nid]/reportback` with a new reportback.
4. Hitting `/api/v1/campaigns/[nid]/reportback` with a new reportback item. 

And all of the above items make it into the following tables: 
1. Check the Rogue reportback_items table for the reportback item.
2. Check the Phoenix dosomething_reportback_file table for the reportback item.
3. Check that Phoenix's dosoemthing_rogue_reportbacks table has the reportback item with all correct data.

Lastly, make sure the cron job works. 

#### Relevant tickets
Fixes #7232

#### Checklist
- [ ] Documentation added for new features/changed endpoints.
- [ ] Tested on staging.
- [ ] Pinged a PM if this is a larger PR that would benefit from some additional testing love
- [ ] Post a message in #phoenix if this includes something that causes a rebuild!

PR #7235, opened 2016-12-06
Fix around getting transaction id when sending RB from API -> Rogue 

chloealee:
#### What's this PR do?
- Changes the way we are grabbing the transaction id (`X-Request-Id`) from the header.
- Changes a check to `isset()` to get rid of log messages.

Same as testing from [this PR](https://github.com/DoSomething/phoenix/pull/7233).

- When testing on Thor, it thought that it had a transaction id so skipped lines 365-370 and went straight to 373 to save to Phoenix. 
- I saw this error in the logs and realized that it wasn't grabbing the header correctly: 
screen shot 2016-12-06 at 4 29 48 pm

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -359,9 +359,9 @@ function _campaign_resource_reportback($nid, $values) {
     watchdog('dosomething_api', '_campaign_resource_reportback:' . json_encode($values));
   }
 
-  $transaction_id = $_SERVER['HTTP_X_REQUEST_ID'];
+  $headers = getallheaders();
 
-  if (is_null($transaction_id)) {
+  if (! isset($headers['X-Request-Id'])) {
     $rogue_reportback = dosomething_rogue_send_reportback_to_rogue($values, $user);
 
     // Make sure reportback made it back to Phoenix and store reference to the Rogue item if one was added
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -359,9 +359,9 @@ function _campaign_resource_reportback($nid, $values) {
     watchdog('dosomething_api', '_campaign_resource_reportback:' . json_encode($values));
   }
 
-  $transaction_id = $_SERVER['HTTP_X_REQUEST_ID'];
+  $headers = getallheaders();
 
-  if (! $transaction_id) {
+  if (! isset($headers['X-Request-Id'])) {
     $rogue_reportback = dosomething_rogue_send_reportback_to_rogue($values, $user);
```
chloealee, line 365: I know this looks weird but the [`getallheaders()`](http://php.net/manual/en/function.apache-request-headers.php) returns the headers that are all caps into only first letter capitalized and the rest lowercase e.g. `X-CSRF-Token` is returned as `X-Csrf-Token` and `X-Request-ID` is returned as `X-Request-Id`

sbsmith86: weird

sbsmith86:
👍

sbsmith86 (approved):

PR #7243, opened 2016-12-08
send Rogue remote addr from Phoenix web

### LINKED ISSUE #7241: let Rogue send Phoenix the right `remote_addr`
_BUG_
-------------------
#### Current Behavior
When a reportback item is posted to Phoenix from Rogue via the API, the `remote_addr` is from Rogue, not the user.

#### Desired Behavior
When a member reports back via Phoenix web, we want to grab their remote address and pass it along down the chain with everything else associated with the reportback.

#### Steps to Replicate
With Rogue on, reportback from Phoenix web and watch it overwrite the remote address.

#### Why This Matters
Data.

katiecrane:
#### What's this PR do?
Before sending a reportback to Rogue, grab the remote address and send that too. Then, allow requests coming into the Phoenix API to specify a remote address instead of grabbing the address when the reportback item is created. If none is specified, grab it like we were before. 

This allows reportbacks submitted via Phoenix web to get created with the address of where they were actually created, instead of the address of Rogue.

I tested by hardcoding the address that got sent to Rogue and making sure that that is the address that showed up in the Phoenix database as well. Also make sure everything still works when hitting the Phoenix API directly and _not_ providing a `remote_addr`.

This was maybe the most confused I've ever been by Drupal. 

#### Relevant tickets
Fixes #7241

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/documentation/endpoints/campaigns.md
+++ b/documentation/endpoints/campaigns.md
@@ -322,6 +322,8 @@ Must be authenticated to post.
     Corresponding caption for the reportback image.
   - **source**: (string).
     Where the reportback file was submitted from.
+  - **remote_addr**: (string). optional
+    Remote address the reportback file was submitted from.
 
 Response: The reportback rbid if success, `FALSE` if not.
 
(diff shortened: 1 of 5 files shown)


sbsmith86 (approved):
Looks great!

PR #7247, opened 2016-12-09
Edit failed logs table

### LINKED ISSUE #7246: Add new columns for more information in the `dosomething_rogue_failed_task_log` table
_FEATURE OVERVIEW_
-------------------
#### User Story
As a developer, I want to add columns to the `dosomething_rogue_failed_task_log` for future debugging.

Specifically, we'd like to add  `source`, `tries`, `failed_at`, and `remote_addr` to give more information about the failed task. 

#### Why This Matters
With more information, we can better debug why certain tasks are stuck in the failed task log.

chloealee:
#### What's this PR do?
- Adds the `source`, `remote_addr`, `failed_at`, and `tries` columns to the `dosomething_rogue_failed_task_log` table. 
- Adds logic to save `failed_at` timestamp for each item in the log.
- Adds logic to increment `tries` every time the cron job runs and fails. 

Updated table
1. Run `drush updb`.
2. Make sure all new columns are added to the `dosomething_rogue_failed_task_log` table.

Reportbacks
1. Change Rogue URL to `rogue.com` in Rogue API Settings.
2. Reportback for a campaign via postman using the `/campaigns/[nid]/reportback` url.
3. Check the `dosomething_rogue_failed_task_log` and make sure the new columns are filled in. 
  - `tries` should be `null`
4. Hit the same url with all the same info.
5. Check the table again that:
  - `tries` increments to `1`
  - `failed_at` doesn't change from original value and is different from `timestamp`.

Reportback Item Status Update
1. Change Rogue URL back to correct URL. 
2. Reportback for a campaign.
3. Change Rogue URL back to `rogue.com`
4. Approve the reportback item.
5. Check that this made it to the `dosomething_rogue_failed_task_log` table. 
  - The `tries` column should be `null`.
6. In Sequel PRO, change that reportback item's `status` to `pending` so it goes back into the review page. Approve the reportback item. 
5. Check the table again that:
  - `tries` increments to `1`
  - `failed_at` doesn't change from original value and is different from `timestamp`.

This will help in the future to give more information to debug any items that are stuck in this table, like the ones in [this card](https://trello.com/c/RfzM3IIN/218-6-as-a-developer-i-want-to-look-into-some-errors-that-are-showing-up-possibly-related-to-some-new-relic-errors).

#### Relevant tickets
Fixes #7246

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install
@@ -259,32 +259,27 @@ function dosomething_rogue_update_7006(&$sandbox) {
 function dosomething_rogue_update_7007(&$sandbox) {
   $table = 'dosomething_rogue_failed_task_log';
   if (db_table_exists($table)) {
-    $source_spec = [
-      'type' => 'text',
-      'description' => 'Where the user reported back from.',
-    ];
-
-    db_add_field($table, 'source', $source_spec);
-
-    $remote_addr_spec = [
-      'type' => 'varchar',
-      'length' => '45',
-      'description' => 'The IP address of the user that submitted the file.',
-    ];
-
-    db_add_field($table, 'remote_addr', $remote_addr_spec);
-
     $new_fields = [
-      'failed_at' => 'The Unix timestamp at which the reportback first fails to send to Rogue.',
-      'tries' => 'The number of attempts the reportback is sent to Rogue.',
+      'source' => [
+        'type' => 'text',
+        'description' => 'Where the user reported back from.',
+      ],
+      'remote_addr' => [
+        'type' => 'varchar',
+        'length' => '45',
+        'description' => 'The IP address of the user that submitted the file.',
+      ],
+      'failed_at' => [
+        'type' => 'int',
+        'description' => 'The Unix timestamp at which the reportback first fails to send to Rogue.',
+      ],
+      'tries' => [
+        'type' => 'int',
+        'description' => 'The number of attempts the reportback is sent to Rogue.',
+      ],
     ];
 
-    foreach ($new_fields as $field => $description) {
-      $spec = [
-        'type' => 'int',
-        'description' => $description,
-      ];
-
+    foreach ($new_fields as $field => $spec) {
       db_add_field($table, $field, $spec);
     }
   }
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install
@@ -252,3 +252,28 @@ function dosomething_rogue_update_7006(&$sandbox) {
     }
   }
 }
+
+/**
+ * Adds text
+ */
+function dosomething_rogue_update_7007(&$sandbox) {
+  $table = 'dosomething_rogue_failed_task_log';
+
+  db_add_field($table, 'source', ['type' => 'text', 'description' => 'Where the user reported back from.']);
+
+  db_add_field($table, 'remote_addr', ['type' => 'varchar', 'description' => 'The IP address of the user that submitted the file.']);
+
+  $new_fields = [
+    'failed_at' => 'The Unix timestamp at which the reportback first fails to send to Rogue.',
+    'tries' => 'The number of attempts the reportback is sent to Rogue.',
+  ];
+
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -208,41 +208,82 @@ function dosomething_rogue_handle_failure($values, $response = NULL, $e = NULL,
...
+      db_update('dosomething_rogue_failed_task_log')
+        ->fields([
+          'tries' => $incremented_tries,
+          'timestamp' => REQUEST_TIME,
+        ])
+        ->condition ('northstar_id', $values['northstar_id'])
+        ->condition ('campaign_run_id', $values['campaign_run_id'])
+        ->condition ('caption', $values['caption'])
+        ->condition ('file', $values['file'])
+        ->execute();
```
chloealee: @sbsmith86 do you think it'd be easier to read if I broke all of the update and insert code into their own functions or is this ok as is?

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install
```diff
@@ -252,3 +252,40 @@ function dosomething_rogue_update_7006(&$sandbox) {
     }
   }
 }
+
+/**
+ * Adds source, remote_addr, failed_at, and tries columns to dosomething_rogue_failed_task_log.
+ */
+function dosomething_rogue_update_7007(&$sandbox) {
```
sbsmith86, line 259: I think you should be able to add these fields in a loop. Like use an array with each field's settings and then loop over the array to execute the repeated parts.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -335,3 +376,26 @@ function dosomething_rogue_check_rbid_and_store_ref($rogue_reportback) {
...
+/**
+ * Query to check if a reportback item exists in the dosomething_rogue_failed_task_log table.
+ *
+ * @param string $northstar_id
+ * @param int $campaign_run_id
+ * @param string $caption
+ * @param string $file
+ *
+ */
+function dosomething_rogue_rb_previously_failed($northstar_id, $campaign_run_id, $caption, $file) {
```
sbsmith86: I think you can just pass an array into this function and then when you call it, pass the values array in. Then you can grab the values you need to run the query from there.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -335,3 +376,26 @@ function dosomething_rogue_check_rbid_and_store_ref($rogue_reportback) {
...
+  return db_query("SELECT * FROM {dosomething_rogue_failed_task_log} failed_rbs WHERE northstar_id = :northstar_id AND campaign_run_id = :campaign_run_id AND caption = :caption AND file = :file", array(':northstar_id' => $northstar_id, ':campaign_run_id' => $campaign_run_id, ':caption' => $caption, ':file' => $file))->fetchAll();
+}
+
+/**
+ * Query to check if a reportback item failed updated status exists in the dosomething_rogue_failed_task_log table.
+ *
+ * @param int $rogue_item_id
+ *
+ */
+function dosomething_rogue_rb_item_previously_failed($rogue_item_id) {
```
sbsmith86: Is there a way you can combine `dosomething_rogue_rb_previously_failed()` and `dosomething_rogue_rb_item_previously_failed($rogue_item_id)` If you generalize the params you pass in and also pass in like the `type` maybe you could make a more general query that just checks if something is already in the table. ?

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
sbsmith86: That seems like a good idea! also maybe if you broke them into their own functions you can use it when you do the inserts and updates below.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install
chloealee: ah yes! Updated below :)

sbsmith86 (approved):
👍

PR #7248, opened 2016-12-12
reportback migration script updates

katiecrane:
#### What's this PR do?
1. Select only what we use later in the query
2. Update some comments
3. Prints out that there's no northstar ID only if there is no northstar ID
4. Does not pull reportbacks whose images were lost

Run it and see if it all breaks.

For getting these reportbacks into Rogueyboy or however we are spelling that.

#### Checklist
- [ ] Tested on staging.


angaither (approved):
📷

PR #7250, opened 2016-12-13
Fixes `failed_at` and `tries` logic error with cron job and limits cron job tries to 5.

### LINKED ISSUE #7249: failed_at and tries logic breaks with cron job
_BUG_
-------------------
#### Current Behavior
- When the cron job runs, the `failed_at` time and number of `tries` gets re-set.

#### Desired Behavior
- The `failed_at` time should always remain the same and the `tries` column should always increment when a new try is made and fails.

chloealee:
#### What's this PR do?
- Refactors `$previously_failed` and makes it an object so code is more readable. 
- Deletes `db_truncate` of `dosomething_rogue_failed_task_log` so logic works for `failed_at` and `tries` columns. 
- Limits cron job to 5 tries when re-sending Rogue failed tasks in `dosomething_rogue_failed_task_log`.
- Fixes formatting.

- Change the Rogue API to something incorrect so failed jobs will continue failing when running the cron job. 
- Run the cron job. Original `failed_at` times should be the same (not the same as `timestamp`) and `tries` should be incremented by one. 
- Reset `tries` column to equal `4`. 
- Change the Rogue API url back to what is it supposed to be.
- Run the cron job. 
- All of the reportbacks and items should make it to Rogue and Phoenix and the `dosomething_rogue_failed_task_log` should be empty. 

#### Relevant tickets
Fixes #7249 and #7251 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
@@ -12,7 +12,6 @@
       ->fields('t')
       ->execute()
       ->fetchAll();
-    db_truncate('dosomething_rogue_failed_task_log')->execute();
 
     foreach ($task_log as $task) {
       if ($task->type === 'reportback') {
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
@@ -3,17 +3,18 @@
 /**
  * Implements hook_cron()
  */
- function dosomething_rogue_cron() {
-    dosomething_rogue_retry_failed_reportbacks();
- }
+function dosomething_rogue_cron() {
+  dosomething_rogue_retry_failed_reportbacks();
+}
 
- function dosomething_rogue_retry_failed_reportbacks() {
-    $task_log = db_select('dosomething_rogue_failed_task_log', 't')
-      ->fields('t')
-      ->execute()
-      ->fetchAll();
+function dosomething_rogue_retry_failed_reportbacks() {
+  $task_log = db_select('dosomething_rogue_failed_task_log', 't')
+    ->fields('t')
+    ->execute()
+    ->fetchAll();
 
-    foreach ($task_log as $task) {
+  foreach ($task_log as $task) {
+    if ($task->tries < 5) {
       if ($task->type === 'reportback') {
         // Check to see if the MIME type is missing
         if (strpos($task->file, 'data:;') !== false) {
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
```diff
@@ -12,7 +12,6 @@
       ->fields('t')
       ->execute()
       ->fetchAll();
-    db_truncate('dosomething_rogue_failed_task_log')->execute();
```
katiecrane, line 15: do items still get removed from the failed log if they make it to Rogue successfully?

chloealee: ah good call @katiecrane thank you! I don't think they do - will figure this out now!

katiecrane:
this looks good to me!

chloealee:
cool thanks @katiecrane ! Would you mind approving the review? :)

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -330,31 +330,31 @@ function dosomething_rogue_previously_failed($values) {
...
  *
  * @param array $values
- * @param array $previously_failed
+ * @param object $previously_failed
  *
  */
 function update_failed_task_log($values, $previously_failed) {
   // Increment the number of tries the cron job attempted to send failed record to Rogue.
-  $incremented_tries = $previously_failed[0]->tries + 1;
+  $incremented_tries = $previously_failed->tries + 1;
```
katiecrane, line 338: Just making sure that this defaults `$previously_failed->tries` to 0 if it has never been set before? With no warnings? I can't find it getting set initially anywhere

katiecrane: Or 1 I guess, not 0

chloealee: ah great question! So [`$previously_failed`](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module#L209) come from [this query](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module#L321) that checks to see if the reportback item is in the `dosomething_rogue_failed_task_log` already. 

When [inserted into this table initially](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module#L371), it doesn't add anything to the `tries` and is `NULL`. 

Once it tries again to send, it uses the `dosomething_rogue_previously_failed` query again to look for that. If it is there, it returns that row and in this PR, I just turn it into an object to get the `tries` in a more readable fashion. Then it increments. Let me know if that is still confusing or if I totally missed what you meant in your Q!

katiecrane: This is exactly what I meant! So on the 2nd try overall, it grabs `tries` which is `NULL`. Then it adds 1 to that (I guess you can do that if it's not crashing) to get 1?

chloealee: Yup exactly!

katiecrane (approved):
looks good to me!

PR #7252, opened 2016-12-13
only tries to re-send Rogue failed tasks if tries is less than 5

chloealee:
#### What's this PR do?
- Limits cron job to 5 tries when re-sending Rogue failed tasks in `dosomething_rogue_failed_task_log`.
- Fixes formatting.

- Change the Rogue API to something incorrect so failed jobs will continue failing when running the cron job.
- Run the cron job. The `tries` column should increment by one. 


chloealee:
Closing this PR - adding this to #7250

PR #7254, opened 2016-12-14
Do not redirect to Northstar profile page when resetting password.

DFurnes:
#### What's this PR do?
Users whose accounts have been created by Northstar (either via the API or the OpenID Connect flow) do not have hashed passwords stored in the Phoenix database. Because of this, we redirect them to Northstar when editing their profile since Phoenix could never verify their current password.

This causes a problem for users who are resetting their passwords though, because that happens on the profile page – and because of the above logic, they're always just redirected to log in on Northstar. 💩! This pull request adds a conditional to skip that redirect when resetting password.

I'm going to deploy this branch to staging to test.

🤕

#### Relevant tickets
Fixes 🐛.

#### Checklist
- [ ] Tested on staging.


DFurnes:
Tested on staging and this is working now!

weerd:
> 🤕

weerd (approved):
nice quick fix! 🛠

PR #7255, opened 2016-12-14
Fixes Rogue cron job bug that didn't send rb items updated status 

chloealee:
#### What's this PR do?
- Fixes bug in Rogue cron job that doesn't successfully send updated rb item's status.
  - Adds `reviewer` column to `dosomething_rogue_failed_task_log`.
  - Updates `insert_failed_task_into_failed_task_log()` to insert `reviewer` to table.
  - Updates cron job to send `reviewer` in `$data` to Rogue. 

#### How should this be reviewed?
1. Make sure there are reportback items in `dosomething_rogue_failed_task_log`.
2. Run the cron job.
3. Reportback items should clear out of the table and updates should be reflected in Rogue. 

#### Any background context you want to provide?
Cron job was breaking because Rogue [saves the `reviewer` to the reportback item](https://github.com/DoSomething/rogue/pull/73) but the cron job was not updated so [this line](https://github.com/DoSomething/rogue/blob/master/app/Repositories/ReportbackRepository.php#L179) broke.  

#### Relevant tickets
Fixes #7253

PR #7257, opened 2016-12-15
Fixes cron job bug with incorrect query 

chloealee:
#### What's this PR do?
- Fixes the cron job bug we saw on 12/14/16 that continued to add the same failed items into the log instead of incrementing one row. 
- @sbsmith86 and I discovered that the query wasn't working - some entries had `NULL` values and sequel doesn't query for `NULL` as a value (e.g. the `northstar_id` was `NULL` and in the query, it didn't find anything with the same `campaign_run_id`, `file`, `caption`, and `null` for the `northstar_id` value). Because it couldn't match this, it didn't know it was already in the `dosomething_rogue_failed_task_log` and continued to add a new record each time. 
  - To combat this, @sbsmith86 had a great idea to send the index of the failed task when it tries to re-send to Rogue as a param to indicate that if it is not `NULL`, this is already in the table and query with the index instead. 

-  Delete a record's `northstar_id` in the `dosomething_rogue_failed_task_log` (or ask Chloe for a csv to import of a failed task that will work for this test). 
-  Change the Rogue API to something incorrect so failed jobs will continue failing when running the cron job. 
- Run the cron job. Original `failed_at` times should be the same (not the same as `timestamp`) and `tries` should be incremented by one. 
- Reset `tries` column to equal `4`. 
- Change the Rogue API url back to what is it supposed to be.
- Run the cron job. 
- All of the reportbacks and items should make it to Rogue and Phoenix and the `dosomething_rogue_failed_task_log` should be empty. 

*NOTE* - the reportback item's with updated statuses will still fail and will be in the table. There's a fix for that in #7255. Once this PR is merged and tested, I will rebase and rebase/update #7255 and will ping for another review!

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
@@ -14,52 +14,49 @@ function dosomething_rogue_retry_failed_reportbacks() {
     ->fetchAll();
 
   foreach ($task_log as $task) {
-    if ($task->tries < 5) {
-      if ($task->type === 'reportback') {
-        // Check to see if the MIME type is missing
-        if (strpos($task->file, 'data:;') !== false) {
-          // Split file string to access the data
-          $data = explode(',', $task->file)[1];
+      if ($task->tries < 5) {
+        if ($task->type === 'reportback') {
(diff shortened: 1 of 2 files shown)


sbsmith86 (approved):
Looks sweet!

PR #7259, opened 2016-12-15
handle sending missing images

katiecrane:
#### What's this PR do?
Reworked `dosomething_helpers_get_data_uri_from_fid` to return `null` if there is no data in the file. If there is no image, pass a flag to Rogue so we can set it as the default. Through many iterations I ended up pretty close to where we started.

Run on some reportbacks with missing images and make sure everything shows up as expected in Rogue.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/scripts/export-reportbacks-to-rogue.php
+++ b/scripts/export-reportbacks-to-rogue.php
@@ -53,13 +53,12 @@ foreach ($rbis as $rb) {
       'status' => dosomething_rogue_transform_status($rb->status),
       // Tell rogue not to try to forward the reportback back to phoenix.
       'do_not_forward' => TRUE,
-      'default_image' => isset($file) ? NULL : TRUE,
     ];
 
     try {
       $response = $client->postReportback($data);
       // Output progress to see what's going on.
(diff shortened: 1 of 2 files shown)


scripts/export-reportbacks-to-rogue.php
```diff
@@ -38,20 +36,24 @@
...
       'why_participated' => $rb->why_participated,
-      'file' => dosomething_helpers_get_data_uri_from_fid($rb->fid),
+      'file' => $file,
       'caption' => $rb->caption,
       'source' => $rb->source,
       'remote_addr' => $rb->remote_addr,
       'status' => dosomething_rogue_transform_status($rb->status),
       // Tell rogue not to try to forward the reportback back to phoenix.
       'do_not_forward' => TRUE,
+      'default_image' => isset($file) ? NULL : TRUE,
```
angaither: do we need to send a new param if `file` is null?

katiecrane: oh oh oh yes I think because if it is an update with no image I'm not sure Rogue could tell the difference between that and a missing image. I guess if the file helper returns `false` instead of `null` then Rogue could see if `file` isset and is false?

angaither:
🚢

angaither (approved):

PR #7260, opened 2016-12-19
add statment so we know which file we fail on

katiecrane:
#### What's this PR do?
Prints out what we are about to attempt to migrate so if it fails, we know which item to look at.

#### Checklist
- [ ] Tested on staging.


angaither (approved):
✉

PR #7261, opened 2016-12-19
catch api exceptions when migrating rbs to rogue

katiecrane:
#### What's this PR do?
Adds a catch for `DoSomething\Gateway\Exceptions\ApiException` like we do in the Rogue module. Hopefully this will catch the `422`s we are getting (if there are more after that first one) when testing on Thor. I'm thinking it might be helpful to get everything not causing trouble migrated over so we can see how many are actually causing trouble and try to find a pattern.

#### Checklist
- [ ] Tested on staging.


angaither:
should we add these into a log?

katiecrane:
@angaither like throw the ones that failed this way into their own table so we can keep better track of them?

angaither:
@katiecrane maybz, or a log file that's probably easier. I think @DFurnes did something like this for a NS script?

katiecrane:
@angaither oooo yeah that'd probably be easier and just as good for what we need. I'll dig around

PR #7262, opened 2016-12-20
Add a script to sanitize invalid email addresses.

DFurnes:
#### What's this PR do?
This pull request includes a quick little script to iterate over the 5,183 users that exist in Phoenix but not Northstar and either sanitize their email if it's invalid or attempt to re-sync them if not. (c62d63d)

This account also removes the code which would sync a user's hashed password to Northstar, since we'll now have a significant segment of users who do not have a valid password stored in the Phoenix database and we'd never want to overwrite the real one in Northstar with that. (17cc4e3)

👀

I'm investigating an issue reported by @mikefantini where his Phoenix account had disappeared from the `authmap` table and therefore couldn't log in. While this doesn't fix whatever _caused_ that to happen, it's a good short-term fix for other potentially affected accounts.

#### Relevant tickets
Fixes 🐛.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/scripts/sanitize-email-addresses.php
+++ b/scripts/sanitize-email-addresses.php
@@ -1,0 +1,27 @@
+<?php
+/*
+ * Sanitize email addresses on user accounts from ages ago.
+ *
+ * drush --script-path=../scripts php-script sanitize-email-addresses.php
+ */
+
+// Any invalid email addresses weren't synced to Northstar, so we can start there.
+$wild_typers = db_query('SELECT uid, mail FROM users WHERE uid NOT IN (SELECT uid FROM authmap);');
+
+foreach($wild_typers as $wilder) {
+  if ($wilder->uid == 0) {
+    continue;
+  }
+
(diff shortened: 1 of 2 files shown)


DFurnes:
I'll make an issue, but I think we should take some time in one of the early sprints next year to audit and remove any syncing code that's no longer necessary now that all users go through Northstar.

scripts/sanitize-email-addresses.php
```diff
@@ -0,0 +1,34 @@
+<?php
+/*
+ * Sanitize email addresses on user accounts from ages ago.
+ *
+ * drush --script-path=../scripts php-script sanitize-email-addresses.php
+ */
+
+// Any invalid email addresses weren't synced to Northstar, so we can start there.
+$wild_typers = db_query('SELECT uid, mail FROM users WHERE uid NOT IN (SELECT uid FROM authmap);');
```
angaither, line 9: WILD TYPERS! yaiiiis!

angaither (approved):

PR #7270, opened 2017-01-05
Cron job bug

### LINKED ISSUE #7269: Cron job that re-sends failed tasks to Rogue breaks if record doesn't have a drupal_id
_BUG_
-------------------
#### Current Behavior
- If a record in the `dosomething_failed_task_log` doesn't have a `drupal_id`, the cron job breaks and never makes it to remaining records to re-try sending to Rogue.

#### Desired Behavior
- We should handle errors properly to continue with the job even if a record doesn't have all the information needed. 

#### Steps to Replicate
- - See https://jenkins.dosomething.org/job/Cron-Production/14523/console as an example of this error:
```
WD cron: DoSomething\Gateway\Exceptions\InternalException: Exception     [error]
in Northstar "GET v1/users/drupal_id/" endpoint: [405] Client error:
`GET https://profile.dosomething.org/v1/users/drupal_id/` resulted in
a `405 Method Not Allowed` response:
{"error":{"code":405,"message":""}}
 in DoSomething\Gateway\Common\RestApiClient->send() (line 239 of
/var/www/www.dosomething.org/releases/20170103183811/vendor/dosomething/gateway/src/Common/RestApiClient.php).
```

#### Why This Matters
- Not all records are being reached in the cron job.

chloealee:
#### What's this PR do?
- If a reportback record in the `dosomething_rogue_failed_task_log` does not have a `drupal_id`, it currently breaks the job and never makes it to the remaining records to try to re-send. 
- This PR fixes this by getting the `drupal_id` with the user's `northstar_id`

- Delete a record's `drupal_id` in the `dosomething_rogue_failed_task_log` table.
- Run the cron job. 
- The record should be removed from the table and you should see this successfully migrated to Rogue.

#### Relevant tickets
Fixes #7269 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
@@ -30,14 +30,19 @@ function dosomething_rogue_retry_failed_reportbacks() {
           $task->file = $mime_split[0] . ':' . $mimetype . $mime_split[1];
         }
 
+        $data = (array)$task;
+
         if ($task->drupal_id) {
           $user = user_load($task->drupal_id);
-        } else {
+        }
+        elseif ($task->northstar_id) {
           $northstar_user = dosomething_northstar_get_user($task->northstar_id);
           $user = user_load($northstar_user->drupal_id);
         }
-
-        $data = (array)$task;
+        else {
+          dosomething_rogue_handle_failure($data, NULL, $task->id);
+          continue;
+        }
 
         $rogue_reportback = dosomething_rogue_send_reportback_to_rogue($data, $user, $task->id);
 
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
@@ -29,6 +29,7 @@ function dosomething_rogue_retry_failed_reportbacks() {
           $mime_split = explode(':', $task->file);
           $task->file = $mime_split[0] . ':' . $mimetype . $mime_split[1];
         }
+
         $user = user_load($task->drupal_id);
         $data = (array)$task;
 
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
@@ -30,7 +30,13 @@ function dosomething_rogue_retry_failed_reportbacks() {
           $task->file = $mime_split[0] . ':' . $mimetype . $mime_split[1];
         }
 
-        $user = user_load($task->drupal_id);
+        if ($task->drupal_id) {
+          $user = user_load($task->drupal_id);
+        } else {
+          $northstar_user = dosomething_northstar_get_user($task->northstar_id);
+          $user = user_load($northstar_user->drupal_id);
+        }
+
         $data = (array)$task;
 
         $rogue_reportback = dosomething_rogue_send_reportback_to_rogue($data, $user, $task->id);
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
@@ -65,12 +65,7 @@ function dosomething_rogue_send_reportback_to_rogue($values, $user = NULL, $fail
   // aware of a user's northstar id. If it doesn't find one, we just grab it
   // from northstar directly.
   if (!$northstar_id) {
-    if ($user->uid) {
-      $northstar_user = dosomething_northstar_get_user($user->uid, 'drupal_id');
-    } else {
-      $northstar_user = dosomething_northstar_get_user($values['northstar_id']);
-      $user = user_load($northstar_user->drupal_id);
-    }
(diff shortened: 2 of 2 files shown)


lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -65,7 +65,12 @@ function dosomething_rogue_send_reportback_to_rogue($values, $user = NULL, $fail
   // aware of a user's northstar id. If it doesn't find one, we just grab it
   // from northstar directly.
   if (!$northstar_id) {
-    $northstar_user = dosomething_northstar_get_user($user->uid, 'drupal_id');
+    if ($user->uid) {
+      $northstar_user = dosomething_northstar_get_user($user->uid, 'drupal_id');
+    } else {
+      $northstar_user = dosomething_northstar_get_user($values['northstar_id']);
+      $user = user_load($northstar_user->drupal_id);
```
angaither: the code here looks fine, but above we are using the global user, so I'm not sure what case we are in where we don't have a user from that call. 
I don't think there would ever be a case where we would have a northstar id, but not a drupal id

angaither (changes requested):

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
chloealee: Doesn't the global user refer to the current user that is logged in? If it is running the cron job, I don't think it'll be able to grab the global user in that case? So if the failed task doesn't have a `drupal_id` but does have a `northstar_id`, we can still grab the `$user`.

chloealee: I think that's why we were getting the below error when the cron job ran too because there was no `$user->uid` (or `$user`) in line 69.
```
WD cron: DoSomething\Gateway\Exceptions\InternalException: Exception     [error]
in Northstar "GET v1/users/drupal_id/" endpoint: [405] Client error:
`GET https://profile.dosomething.org/v1/users/drupal_id/` resulted in
a `405 Method Not Allowed` response:
{"error":{"code":405,"message":""}}
 in DoSomething\Gateway\Common\RestApiClient->send() (line 239 of
/var/www/www.dosomething.org/releases/20170103183811/vendor/dosomething/gateway/src/Common/RestApiClient.php).
```

angaither: maybe i'm missing something, but shouldn't the cron job be getting the drupal id [from the table?](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc#L32)

chloealee: that's where we're running into the issue! So some of the records in the `dosomething_rogue_failed_task_log` do not have a `drupal_id`, so `$user` in the line you linked to is `NULL`.

angaither: ah, tricky. seems like there's a few without a ns id as well. so we might want to make this an else if to make sure there's a ns id!

angaither: ORRR, another (maybe better) idea is to set the user if it still doesn't exist after the `global $user` line

chloealee: do you mean getting rid of line 62 and the `if (!northstar_id)` on line 67?

angaither: upon even closer inspection, the code you added is actually in a chunk that is commented as a bandaid fix. maybe the check should be in the [cron job](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc#L32) instead?

chloealee:
@angaither updated!

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
```diff
@@ -29,7 +29,14 @@ function dosomething_rogue_retry_failed_reportbacks() {
           $mime_split = explode(':', $task->file);
           $task->file = $mime_split[0] . ':' . $mimetype . $mime_split[1];
         }
-        $user = user_load($task->drupal_id);
+
+        if ($task->drupal_id) {
+          $user = user_load($task->drupal_id);
+        } else {
```
angaither: else is below in drooops

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
```diff
@@ -29,7 +29,14 @@ function dosomething_rogue_retry_failed_reportbacks() {
...
+          $northstar_user = dosomething_northstar_get_user($task->northstar_id);
```
angaither, line 39: what if northstar id doesn't exist? there still won't be a user then

chloealee:
@angaither fixed the styling, added logic if there is no drupal or ns id,  and also got rid of a param we were sending to `update_failed_task_log` that wasn't being used

angaither (approved):

angaither:
👍

PR #7273, opened 2017-01-09
Custom share on all campaigns (other than LYCV) 

### LINKED ISSUE #7272: Custom social share on all campaigns
_FEATURE OVERVIEW_
-------------------
#### User Story
As an admin, I want to be able to use the custom social share on campaigns other than LYVC

#### Why This Matters
We want to make it easier for user's to share the campaigns they are doing and encourage friends to join as part of our OKRs.

chloealee:
#### What's this PR do?
- Adds a photo by photo's `nid` if `social_share_unique_link` is enabled (even if voter reg is not).
- Adds field to customize description under "Share This Campaign" in share modal (by campaign basis). 
- Removes all text below image. 

- On any campaign, go to the "Custom Settings" tab.
- Check off `Create unique social share links for each user.` box.
- Fill out all fields and save.
- On campaign page, make sure the custom share setting is there with all information you put in the form. 

#### Relevant tickets
Fixes #7272

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -443,9 +443,6 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
       $vars['social_share_custom_description'] = $campaign->variables['social_share_custom_description'];
       $vars['share_image'] = dosomething_image_get_themed_image_url($campaign->variables['share_image_nid'], 'landscape');
   }
-  else {
-    $vars['share_image'] = $base_url . 'sites/default/files/images/problem-' . $campaign->nid . '-' . $country_prefix_for_share . '.png';
-  }
 
   // Add social share bar.
   $social_share_types = array(
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -439,7 +439,7 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
   $base_url = url(base_path(), array('absolute'=> TRUE, 'language' => 'en-global'));
   $campaign_path = url(current_path(), array('absolute' => TRUE));
   $country_prefix_for_share = (dosomething_global_get_current_prefix()) ? strtoupper(dosomething_global_get_current_prefix()) : 'global';
-  if (dosomething_campaign_feature_on($campaign, 'enable_voter_registration', TRUE)) {
+  if (dosomething_campaign_feature_on($campaign, 'social_share_unique_link')) {
       $vars['share_image'] = dosomething_image_get_themed_image_url($campaign->variables['share_image_nid'], 'landscape');
   }
   else {
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -440,6 +440,7 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
   $campaign_path = url(current_path(), array('absolute' => TRUE));
   $country_prefix_for_share = (dosomething_global_get_current_prefix()) ? strtoupper(dosomething_global_get_current_prefix()) : 'global';
   if (dosomething_campaign_feature_on($campaign, 'social_share_unique_link')) {
+      $vars['social_share_custom_description'] = $campaign->variables['social_share_custom_description'];
       $vars['share_image'] = dosomething_image_get_themed_image_url($campaign->variables['share_image_nid'], 'landscape');
   }
   else {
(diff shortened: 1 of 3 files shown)


lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -451,9 +454,6 @@
             <?php if ($register_voters) : ?>
               <p><?php print $voter_reg_share_modal_text_2 ?></p>
             <?php endif; ?>
-            <div class="padded-gray-box">
-              <?php print $custom_social_share_link ?>
```
angaither, line 455: this should probably be broken out it to it's own checkbox at some point, unless it's a feature we no longer want.

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -443,6 +443,9 @@
               <p><?php print $voter_reg_share_modal_text_1 ?></p>
```
angaither, line 443: I think we could take this as a chance to clean up this code and remove the voter modal

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.variable.inc
```diff
@@ -32,6 +32,15 @@ function dosomething_helpers_variable_form($form, &$form_state, $node) {
       'maxlength' => '140',
     ),
   ];
+  $form['custom_social_sharing']['social_share_custom_description'] = [
+    '#type' => 'textfield',
+    '#title' => t('Custom social share description'),
+    '#description' => t("Custom description to share with social links."),
+    '#default_value' => $vars['social_share_custom_description'],
+    '#attributes' => array(
```
angaither: array with `[]`

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.variable.inc
```diff
@@ -32,6 +32,15 @@ function dosomething_helpers_variable_form($form, &$form_state, $node) {
       'maxlength' => '140',
     ),
   ];
+  $form['custom_social_sharing']['social_share_custom_description'] = [
```
angaither, line 35: what is this part for?!

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -439,7 +439,8 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
   $base_url = url(base_path(), array('absolute'=> TRUE, 'language' => 'en-global'));
   $campaign_path = url(current_path(), array('absolute' => TRUE));
   $country_prefix_for_share = (dosomething_global_get_current_prefix()) ? strtoupper(dosomething_global_get_current_prefix()) : 'global';
-  if (dosomething_campaign_feature_on($campaign, 'enable_voter_registration', TRUE)) {
+  if (dosomething_campaign_feature_on($campaign, 'social_share_unique_link')) {
+      $vars['social_share_custom_description'] = $campaign->variables['social_share_custom_description'];
       $vars['share_image'] = dosomething_image_get_themed_image_url($campaign->variables['share_image_nid'], 'landscape');
   }
   else {
```
angaither: I think we can safely remove the else here, since it seems like we don't want problem shares at all anymore.

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.variable.inc
```diff
@@ -32,6 +32,15 @@ function dosomething_helpers_variable_form($form, &$form_state, $node) {
...
+    '#type' => 'textfield',
```
katiecrane: I feel like this should be a `textarea` so it can be longer, seems like a place someone might want some more copy potentially

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
chloealee: ok! should I check with Jen about this or just go ahead and make it a checkbox?

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.variable.inc
chloealee: this allows for campaign leads to put custom text above the share image like the below: 
screen shot 2017-01-09 at 3 56 38 pm

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.variable.inc
chloealee: ah great call! will change now :)

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
angaither: might want to double check with @ngjo before doing that. because with this update it I think it  would become an unusable feature

ngjo: @chloealee great question -- i think there's a chance we might want that in the future, but for New Year, New Us, it's not something we're going to be needing. Does that help?

chloealee: yes that makes sense! Should I do a separate PR for that or should I combine here? @ngjo @angaither

angaither (approved):

PR #7277, opened 2017-01-11
Custom gray box

### LINKED ISSUE #7274: add checkbox for custom social share link below share image
_FEATURE OVERVIEW_
-------------------
#### User Story
- As an admin, I want the option to add the custom social share link (in the gray box) below the share image).

chloealee:
#### What's this PR do?
- Adds a checkbox to add the option to show the user's custom share url in a gray box below the share image.

- In custom settings, make sure `Create unique social share links for each user.` is checked off. 
- Also check the `Visible custom social share link` box. 
- Run `drush cc all`. 
- On the campaign page, the custom social modal should look like the below:
screen shot 2017-01-11 at 10 35 13 am

Refs #7273

#### Relevant tickets
Fixes #7274 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
@@ -433,6 +433,11 @@
             <?php if ($register_voters) : ?>
               <p><?php print $voter_reg_share_modal_text_2 ?></p>
             <?php endif; ?>
+            <?php if ($show_social_share_link) : ?>
+              <div class="padded-gray-box">
+                <?php print $custom_social_share_link ?>
+              </div>
+            <?php endif; ?>
           </div>
         </div>
        <?php endif; ?>
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -442,9 +442,7 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
   if (dosomething_campaign_feature_on($campaign, 'social_share_unique_link')) {
       $vars['social_share_custom_description'] = $campaign->variables['social_share_custom_description'];
       $vars['share_image'] = dosomething_image_get_themed_image_url($campaign->variables['share_image_nid'], 'landscape');
-      if ($campaign->variables['show_social_share_link']) {
-        $vars['show_social_share_link'] = $campaign->variables['show_social_share_link'];
-      }
+      $vars['show_social_share_link'] = $campaign->variables['show_social_share_link'];
   }
 
   // Add social share bar.
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -442,6 +442,9 @@ function dosomething_campaign_preprocess_action_page(&$vars, &$wrapper) {
   if (dosomething_campaign_feature_on($campaign, 'social_share_unique_link')) {
       $vars['social_share_custom_description'] = $campaign->variables['social_share_custom_description'];
       $vars['share_image'] = dosomething_image_get_themed_image_url($campaign->variables['share_image_nid'], 'landscape');
+      if ($campaign->variables['show_social_share_link']) {
+        $vars['show_social_share_link'] = $campaign->variables['show_social_share_link'];
+      }
   }
 
   // Add social share bar.
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.variable.inc
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.variable.inc
@@ -70,6 +70,11 @@ function dosomething_helpers_variable_form($form, &$form_state, $node) {
     '#description' => t("Custom Tumblr copy to populate Tumblr share. If blank, copy will default to the problem statement."),
     '#default_value' => $vars['custom_tumblr_copy'],
     '#size' => 20,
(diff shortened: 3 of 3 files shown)


lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -454,6 +454,11 @@
             <?php if ($register_voters) : ?>
               <p><?php print $voter_reg_share_modal_text_2 ?></p>
             <?php endif; ?>
+            <?php if ($show_social_share_link) : ?>
```
katiecrane, line 436: if `$campaign->variables['show_social_share_link']` is false does this give any kind of error, maybe in the console? it looks like `$show_social_share_link` only gets set if the other one is `true`, but I might be missing something

chloealee: @katiecrane what do you mean by "the other one"?

chloealee: ohhh @katiecrane , do you mean `$custom_social_share_link`?

katiecrane: sorry I meant the other value I mentioned: `$campaign->variables['show_social_share_link']`
referring to here: https://github.com/DoSomething/phoenix/blob/c41162ae39e2e7c562fe9674b44bd44fcc00d0e5/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc#L445-L447

chloealee: @katiecrane ah I see what you mean! Will take that out since you're right, totally unneeded :)

sbsmith86: it might be a little more clean to do the check here, like you have it and just set `$show_social_share_link` to whatever the campaign variable is. It would be more readable in the template that it is an optional thing.

chloealee: @sbsmith86 not sure I quite understand. Do you mean to take out the if statement [here](https://github.com/DoSomething/phoenix/blob/c41162ae39e2e7c562fe9674b44bd44fcc00d0e5/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc#L445) and `$vars['show_social_share_link']` will be either `0` or `1`. Then in the `campaign.tpl.php` on line 457, write `<?php if ($show_social_share_link === 1) : ?>`?

sbsmith86: yep cause then when you are in the template you know that it is something that isn't always there and depends on a flag.

chloealee: ahh yes this is a lot better - I'll change now!

chloealee: @sbsmith86 updated!

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -433,7 +433,7 @@
             <?php if ($register_voters) : ?>
               <p><?php print $voter_reg_share_modal_text_2 ?></p>
             <?php endif; ?>
-            <?php if ($show_social_share_link) : ?>
+            <?php if ($show_social_share_link == 1) : ?>
```
katiecrane: do you need the `== 1`? I thought `$show_social_share_link` itself would be `true` or `false` so you could just have that?

chloealee: @katiecrane I think the idea here that @sbsmith86 mentioned is that it is more readable as an optional feature instead of it just not existing.

katiecrane: 🙈 🙈 🙈  I did not read that all carefully enough sorry!!

chloealee: no worries at all! Blame :octocat: for hiding it :)

katiecrane: wait I think I'm still missing something. Doesn't `<?php if ($show_social_share_link) : ?>` already show that it's optional in the template?

chloealee: hmm I guess it is a bit telling by the name of the variable. I think just more obvious with this unless @sbsmith86 I'm missing something else?

sbsmith86: I think the if statement is all you need to express what i thought was a little more readable. how you do the check (i.e checking it as a boolean or if  it is `==1`) is based on what the value of that flag is. So doing `<?php if ($show_social_share_link)` should work.

chloealee: cool @sbsmith86 @katiecrane updated!

sbsmith86 (approved):
🎉

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
katiecrane: woo thanks for explaining everyone!

PR #7285, opened 2017-01-24
Correctly parse ISO-8601 dates provided for birthdate.

DFurnes:
#### What's this PR do?
This fixes a fatal error that occurred when trying to parse a date string as a user's birthdate from their user info. Drupal expects all dates to be UNIX timestamps, but the spec says that field in the JSON response should be a ISO-8601 string  (e.g. `1990-10-25`) so we'll handle that here.

👀

🍃

#### Relevant tickets
References DoSomething/northstar#531.

#### Checklist
- [ ] Tested on staging.


weerd (approved):
🕐  👍

PR #7286, opened 2017-01-24
Adding new affiliate configurations form in CMS specific to each campaign.

weerd:
#### What's this PR do?
This PR adds a custom form for specifying affiliate information on a per campaign basis and updates the pitch page to render the form with the custom settings as needed, when a user is signed in. It also adds a new per campaign export form where it will allow export of the users that opted in (opting out for this first one ;), but could be used for exporting other campaign specific data in the future.

👀 

This is Part 1 of work related to https://trello.com/c/YmS6rixd/216-5-as-a-biz-dev-i-want-to-build-an-opt-out-option-for-3rd-party-messaging-during-sign-up

Part 2 will expand on this work and enable saving the opt-outs (ins) and exporting them. It will also handle the logged out pitch page and the redirect over to Northstar and back with the info.

#### Relevant tickets
Refs #7296

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -275,14 +275,10 @@ function dosomething_campaign_load($node, $public = FALSE, $cache = TRUE) {
       $campaign->image_header = $campaign->image_cover;
     }
 
-    // dump($node);
-    // die();
-
     // Load Creator property.
     dosomething_campaign_load_creator($campaign, $wrapper);
 
-    // @TODO: Uncomment following line.
-    // cache_set('ds_campaign_' . $campaign->nid .'_'. $current_lang, $campaign, 'cache_dosomething_campaign', REQUEST_TIME + 60*60*24);
+    cache_set('ds_campaign_' . $campaign->nid .'_'. $current_lang, $campaign, 'cache_dosomething_campaign', REQUEST_TIME + 60*60*24);
 
     // If this is accessed via API:
     if ($public) {
@@ -293,9 +289,6 @@ function dosomething_campaign_load($node, $public = FALSE, $cache = TRUE) {
     }
   }
 
-  $campaign->poop = TRUE;
-  // dump($campaign);
-  // die();
   return $campaign;
 }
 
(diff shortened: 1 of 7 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
```diff
@@ -35,11 +35,15 @@ function dosomething_signup_forms($form_id, $args) {
  *   The label to display on the form submit button.
  */
 function dosomething_signup_form($form, &$form_state, $nid, $label = "Sign Up") {
+
```
DFurnes: 👾

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
```diff
@@ -636,3 +636,39 @@ function dosomething_campaign_get_runs($nid, $language) {
...
+
+/**
+ * Get all the affiliate configurations data for a campaign.
+ *
+ * @param  string|array $campaign_info Either a node ID or an array of campaign variables.
+ * @return array
+ */
+function dosomething_campaign_get_affiliate_data($campaign_info) {
+  if (!$campaign_info) {
+    return FALSE;
```
DFurnes: Let's return `NULL` rather than `FALSE` on failure, and update the docblock to `@return array|null`.

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
```diff
@@ -636,3 +636,39 @@ function dosomething_campaign_get_runs($nid, $language) {
...
+  }
+
+  // Is provided info a string node ID?
+  if (is_string($campaign_info)) {
+    $variables = dosomething_helpers_get_variables('node', $campaign_info);
+  }
+
+  $data = [
+    'messaging_opt_in' => [
+      'enabled' => $variables['affiliate_opt_in_toggle'] ? TRUE : FALSE,
```
DFurnes: Can we cast this as a boolean instead of using a ternary?

```php
'enabled' => (bool) $variables['affiliate_opt_in_toggle'],
```

DFurnes (approved):
Looks good! Left a few little comments.

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
weerd: oh jes!

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
weerd: yeah, defs. seems cleaner.

DFurnes (approved):

PR #7292, opened 2017-01-26
Affiliate opt ins saving to db

### LINKED ISSUE #7291: Record affiliate messaging opt-ins in the database.
https://trello.com/c/YmS6rixd/216-5-as-a-biz-dev-i-want-to-build-an-opt-out-option-for-3rd-party-messaging-during-sign-up

weerd:
#### What's this PR do?
This PR updates the affiliate messaging opt-in(out) feature to respond to whether the opt-in has been enabled for the campaign and then when the form is submitted account for this and create a record in the database if the user has decided _not_ to uncheck the toggle to receive messaging. Currently this only works for the auth'd user. Next up will be allowing this to also happen for users that haven't logged in yet and get redirected to Northstar, but this PR sets up a lot of the ground work to ideally just call a function with some info and have the same process happen 😄 

🕵️ 

Exporting the records from the database is coming in a near-future update.

#### Relevant tickets
Fixes #7291
Refs #7296

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc
@@ -372,6 +372,30 @@ function dosomething_campaign_load_stats(&$campaign) {
 }
 
 /**
+ * Create a new record to store the user's messaging opt-in preference.
+ *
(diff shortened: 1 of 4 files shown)


DFurnes (approved):
Looks good to me! 👌

PR #7294, opened 2017-01-30
Send signups to Rogue

### LINKED ISSUE #7293: Send signups to Rogue
_FEATURE OVERVIEW_
-------------------
#### User Story
When a user signs up, the signup is send to Rogue and then back to Phoenix. This is the same idea of what we currently do with reportbacks.

#### Why This Matters
So Rogue has all the data!

katiecrane:
#### What's this PR do?
Send Phoenix signups to Rogue!

Admin:
- Adds feature flag in the Rogue settings for admins to turn on/off sending signups to Rogue

Failures:
- Retry failed signups with cron
- Send count to stathat for failed signups
- Includes failed signups in `dosomething_rogue_failed_task_log`

Sending:
- Ability to send signups to Rogue!
- Creates `dosomething_rogue_signups` table and stores references to Rogue signups in there
- Follows same logic as reportbacks for sending Phoenix API signups to Rogue (checks for transaction ID and if there isn't one and Rogue signups are on, sends it to Rogue)

In my tests the functionality that I was expecting was all working, so 👀  on the logic is probably the most helpful. Make sure that everything would work normally if the feature flag is off please!

This won't be turned on until there is an experience planned for the user for the case that their signup fails to post.

The Rogue endpoint needs functionality added to deal with the `transactionals` value that can come to the signup endpoint. This was not included because it is not in the Phoenix docs, but should be easy enough to add.

#### Relevant tickets
Fixes #7293

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -284,23 +284,21 @@ function _campaign_resource_signup($nid, $values) {
 
   $headers = getallheaders();
 
+  // Check to see if the request came from Rogue and if we should be sending it there
   // $headers['X-Request-Id'] will only be set if the request is coming from Rogue
-  if ( !isset($headers['X-Request-Id'])) {
-    // Send to Rogue since the signup is not there yet
-    if (variable_get('rogue_signup_collection', FALSE)) {
-      // Pass the campaign_id
-      $values['campaign_id'] = $nid;
+  if ( !isset($headers['X-Request-Id']) && variable_get('rogue_signup_collection', FALSE)) {
+    // Pass the campaign_id
+    $values['campaign_id'] = $nid;
 
-      // Get the run to pass
-      $user = user_load($values['uid']);
-      $run = dosomething_helpers_get_current_campaign_run_for_user($nid, $user);
+    // Get the run to pass
+    $user = user_load($values['uid']);
+    $run = dosomething_helpers_get_current_campaign_run_for_user($nid, $user);
 
-      // Post the signup to Rogue
-      $rogue_signup = dosomething_rogue_send_signup_to_rogue($values);
+    // Post the signup to Rogue
+    $rogue_signup = dosomething_rogue_send_signup_to_rogue($values);
 
-      // Store the reference to the Rogue signup
-      return dosomething_rogue_check_sid_and_store_ref($rogue_signup);
-    }
+    // Store the reference to the Rogue signup
+    return dosomething_rogue_check_sid_and_store_ref($rogue_signup);
   } else {
     // @TODO: make helper function to do this?
     // Increment transaction id and log that the request has been received with new transaction id.
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -286,20 +286,28 @@ function _campaign_resource_signup($nid, $values) {
 
   // Check to see if the request came from Rogue and if we should be sending it there
   // $headers['X-Request-Id'] will only be set if the request is coming from Rogue
-  if ( !isset($headers['X-Request-Id']) && variable_get('rogue_signup_collection', FALSE)) {
-    // Pass the campaign_id
-    $values['campaign_id'] = $nid;
+  if ( !isset($headers['X-Request-Id'])) {
+    if (variable_get('rogue_signup_collection', FALSE)) {
+      // Pass the campaign_id
+      $values['campaign_id'] = $nid;
 
-    // Get the run to pass
-    $user = user_load($values['uid']);
-    $run = dosomething_helpers_get_current_campaign_run_for_user($nid, $user);
+      // Get the run to pass
+      $user = user_load($values['uid']);
+      $run = dosomething_helpers_get_current_campaign_run_for_user($nid, $user);
 
-    // Post the signup to Rogue
-    $rogue_signup = dosomething_rogue_send_signup_to_rogue($values);
+      // Post the signup to Rogue
+      $rogue_signup = dosomething_rogue_send_signup_to_rogue($values);
 
-    // Store the reference to the Rogue signup
-    return dosomething_rogue_check_sid_and_store_ref($rogue_signup);
+      // Store the reference to the Rogue signup
+      return dosomething_rogue_check_sid_and_store_ref($rogue_signup);
+    }
+    // The request is not from Rogue and we don't want to send it there, so just save to Phoenix
+    else {
+      // transactionals = FALSE will result in neither email or SMS transactional messages being sent.
+      return dosomething_signup_create($nid, $values['uid'], $values['source'], NULL, $values['transactionals']);
+    }
   } else {
+    // Only try to increment transaction IDs if they were already set
     // Increment transaction id and log that the request has been received with new transaction id.
     $transaction_id_parts = explode('-', $headers['X-Request-Id']);
     $transaction_id_base = $transaction_id_parts[0];
@@ -307,6 +315,7 @@ function _campaign_resource_signup($nid, $values) {
     $incremented_transaction_id = $transaction_id_base . '-' . $incremented_step;
 
     watchdog('PhoenixTransactionBridge', 'Request received. Transaction ID: !incremented_transaction_id', ['!incremented_transaction_id' => $incremented_transaction_id], WATCHDOG_INFO);
+
     // transactionals = FALSE will result in neither email or SMS transactional messages being sent.
     return dosomething_signup_create($nid, $values['uid'], $values['source'], NULL, $values['transactionals']);
   }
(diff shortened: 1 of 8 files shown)


lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -274,14 +274,42 @@ function _campaign_resource_signup($nid, $values) {
   if (!isset($values['source'])) {
     $values['source'] = NULL;
   }
+  // @TODO: handle passing this to Rogue so it is not lost (and add to Phoenix documentation which is missing this)
   if (!isset($values['transactionals'])) {
     $values['transactionals'] = TRUE;
```
sbsmith86, line 279: What is this used for?

katiecrane: I'm not sure, it's not in the Phoenix docs so I didn't realize it existed until I moved from Rogue to the Phoenix side

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -274,14 +274,42 @@ function _campaign_resource_signup($nid, $values) {
...
+    $user = user_load($values['uid']);
+    $run = dosomething_helpers_get_current_campaign_run_for_user($nid, $user);
+
+    // Post the signup to Rogue
+    $rogue_signup = dosomething_rogue_send_signup_to_rogue($values);
+
+    // Store the reference to the Rogue signup
+    return dosomething_rogue_check_sid_and_store_ref($rogue_signup);
+  } else {
+    // Increment transaction id and log that the request has been received with new transaction id.
```
sbsmith86: Do we do this transactional id logic in multiple places? Should we create a helper for it?

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -274,14 +274,42 @@ function _campaign_resource_signup($nid, $values) {
...
+    $transaction_id_parts = explode('-', $headers['X-Request-Id']);
+    $transaction_id_base = $transaction_id_parts[0];
+    $incremented_step = $transaction_id_parts[1] + 1;
+    $incremented_transaction_id = $transaction_id_base . '-' . $incremented_step;
+
+    watchdog('PhoenixTransactionBridge', 'Request received. Transaction ID: !incremented_transaction_id', ['!incremented_transaction_id' => $incremented_transaction_id], WATCHDOG_INFO);
+    // transactionals = FALSE will result in neither email or SMS transactional messages being sent.
```
sbsmith86: Oh it looks like `transactionals` deals with whether to send emails or not. We should look in how we are gonna handle transactional messages in rogue.

katiecrane: Yeah it seems like that's one of the bigger things we need to tackle in order to turn off Phoenix. I guess Rogue will still have to handle ferrying this value around in the meantime though

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
katiecrane: I think we should! I'll add that

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install
```diff
@@ -118,6 +118,31 @@ function dosomething_rogue_schema() {
...
+        'not null' => TRUE,
+      ],
+      'rogue_signup_id' => [
+        'description' => 'The signup id as it is stored in rogue.',
+        'type' => 'int',
+        'not null' => TRUE,
+      ],
+      'created_at' => [
+        'description' => 'The Unix timestamp of when the reportback was stored',
+        'type' => 'int',
```
chloealee, line 136: I think this `type` should be `TIMESTAMP`?

katiecrane: hmm maybe it should be but we don't have that elsewhere:
[reportbacks table](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install#L36-L38)
[failed table 1](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install#L97-L99)
[failed table 2](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install#L272-L274)

chloealee: ah you're totally right! Sorry - I was looking at the Rogue tables that have all those as `TIMESTAMP`. I think it's ok as is :)

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
```diff
@@ -53,6 +53,23 @@ function dosomething_rogue_retry_failed_reportbacks() {
             ->condition('id', $task->id)
             ->execute();
         }
+      } elseif ($task->type === 'signup') {
+        $data = (array)$task;
```
sbsmith86: I think if you cast this to an array here, you should use that array throughout the logic. It is a little hard to understand why you are using the `$task` object some places and the `$data` array in others.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -226,7 +239,8 @@ function dosomething_rogue_handle_failure($values, $response = NULL, $failed_tas
   // Save fail to a db log so we can easily export.
   if ($values['type'] === 'reportback') {
     watchdog('dosomething_rogue', 'Reportback not migrated to Rogue: northstar_id: !northstar_id, campaign_id: !campaign_id, campaign run_nid: !campaign_run_id.', ['!northstar_id' => $values['northstar_id'], '!campaign_id' => $values['campaign_id'], '!campaign_run_id' => $values['campaign_run_id']], WATCHDOG_ERROR);
-
+  } elseif ($values['type'] === 'signup') {
```
sbsmith86: I sorta feel like it would clean up this handle failures function if we separated it out the signup logic and the reportback logic. So like maybe make a `dosomething_rogue_handle_signup_failures()` function?

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -399,3 +426,132 @@ function insert_failed_task_into_failed_task_log($values, $response = NULL, $e =
...
+*/
+function dosomething_rogue_transform_signup($values, $northstar_id, $user) {
+  // Make sure we have the data in the right form (failed table stores campaign_id, not nid)
+  if (isset($values['nid'])) {
+    $values['campaign_id'] = $values['nid'];
+  }
+
+  $run = dosomething_helpers_get_current_campaign_run_for_user($values['campaign_id'], $user);
+
+
```
sbsmith86: Nitpicky extra line here

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -399,3 +426,132 @@ function insert_failed_task_into_failed_task_log($values, $response = NULL, $e =
...
+    // Store references to rogue IDs.
+    db_insert('dosomething_rogue_signups')
+    ->fields([
+      'sid' => $sid,
+      'rogue_signup_id' => $rogue_signup['data']['id'],
+      'created_at' => REQUEST_TIME,
+      ])
+    ->execute();
+  }
+}
```
sbsmith86: Extra line 😬

lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
```diff
@@ -106,8 +106,14 @@ function dosomething_signup_form_submit($form, &$form_state) {
     $params['affiliate_messaging_opt_in'] = (bool) $form_state['values']['affiliate_messeging_opt_in'];
   }
 
-  // Create signup for logged in user.
-  dosomething_signup_user_signup($nid, NULL, $source, $params);
+  if (variable_get('rogue_signup_collection', FALSE)) {
+    $rogue_signup = dosomething_rogue_send_signup_to_rogue($form_state['values']);
+
+    dosomething_rogue_check_sid_and_store_ref($rogue_signup);
```
sbsmith86: Should you do any check here if `$rogue_signup` is not `NULL` or `FALSE` before trying to use it?

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -274,14 +274,50 @@ function _campaign_resource_signup($nid, $values) {
...
+    // Only try to increment transaction IDs if they were already set
+    // Increment transaction id and log that the request has been received with new transaction id.
+    $incremented_transaction_id = dosomething_api_increment_transaction_id($headers['X-Request-Id']);
+
+    watchdog('PhoenixTransactionBridge', 'Request received. Transaction ID: !incremented_transaction_id', ['!incremented_transaction_id' => $incremented_transaction_id], WATCHDOG_INFO);
+
+    // transactionals = FALSE will result in neither email or SMS transactional messages being sent.
+    return dosomething_signup_create($nid, $values['uid'], $values['source'], NULL, $values['transactionals']);
+  }
 }
```
chloealee, line 321: @katiecrane this logic looks solid to me!

sbsmith86 (approved):
Looking good to me :)

PR #7295, opened 2017-01-30
Sends reportbacks/items and updates to Rogue's /posts and /reviews endpoints 

### LINKED ISSUE #7280: Reportback to use new /posts endpoint in Rogue
_FEATURE OVERVIEW_
-------------------
#### User Story
 As a user, when I reportback for a campaign I want my reportback to be stored in the posts table instead of the rogue reportback tables. This needs to happen when you reportback both on phoenix-web and the Phoenix API.

We'll also need to update the `dosomething_rogue_reportbacks` table to replace `rogue_reportback_id` to `rogue_signup_id`

#### Why This Matters
Updating Phoenix to new changes and data schema in Rogue.

chloealee:
#### What's this PR do?
This PR updates Phoenix code to send reportbacks, reportback items (new and reviewed/updated statuses) to Rogue's `/posts` and `/reviews` endpoints. 

In order to update the above, it also: 
- updates columns in the `dosomething_rogue_reportbacks` and `dosomething_rogue_failed_task_log` tables
- updates Rogue cron jobs 

This should be tested with https://github.com/DoSomething/rogue/pull/112
1. Change Rogue API version number field to `v2` in the Rogue API settings.
2. Run `drush updb` to reflect new database changes.  

Make sure all information matches up between Phoenix and Rogue databases for below scenarios:
- Create a new signup (RB) and photo (reportback item) from Phoenix web. 
- Add another photo (rb item).
- Add another photo (rb item) and update the signup (reportback's `quantity` and `why_participated`).
- Only update the signup (reportback's `quantity` and `why_participated`).
- Review a photo (reportback item's status as an admin). 
- Cron job for reportbacks and reportback items. 

**NOTE** once we merge this and update the table columns, we should make a note of the timing - anything after merge will have `rogue_event_id` instead of `rogue_item_id` in `dosomething_rogue_reportbacks` and `dosomething_rogue_failed_task_log` (this is less important than the first table). 

#### Relevant tickets
Fixes #7280 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
@@ -289,7 +289,7 @@ function dosomething_reportback_files_form_submit($form, &$form_state) {
       $rogue_item_id = dosomething_rogue_get_by_file_id($fid);
 
       // If there is a rogue_item_id in the dosomething_rogue_reportbacks table, send update to Rogue.
-      if ($rogue_item_id[0]->rogue_item_id) {
+      if ($rogue_item_id[0]->rogue_event_id) {
         if ($values['status'] === 'promoted' || $values['status'] === 'approved' || $values['status'] === 'excluded') {
           $rogue_status = 'accepted';
         } else {
@@ -297,7 +297,7 @@ function dosomething_reportback_files_form_submit($form, &$form_state) {
         }
 
         $data = [
-          'rogue_reportback_item_id' => $rogue_item_id[0]->rogue_item_id,
+          'rogue_event_id' => $rogue_item_id[0]->rogue_event_id,
           'status' => $rogue_status,
           'reviewer' => dosomething_user_get_field('field_northstar_id'),
         ];
@@ -309,7 +309,7 @@ function dosomething_reportback_files_form_submit($form, &$form_state) {
 
   // Send updates to Rogue.
   if (variable_get('rogue_collection', FALSE)) {
-    $response = dosomething_rogue_update_rogue_reportback_items($updates_to_send_to_rogue);
+    $response = dosomething_rogue_send_reportback_to_rogue($updates_to_send_to_rogue);
     if ($response->code != 201) {
       drupal_set_message(print_r($response->code, TRUE) . ' ' . print_r($response->status_message, TRUE), 'warning');
     } else {
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
@@ -310,7 +310,7 @@ function dosomething_reportback_files_form_submit($form, &$form_state) {
 
   // Send updates to Rogue.
   if (variable_get('rogue_collection', FALSE)) {
-    $response = dosomething_rogue_send_reportback_to_rogue($updates_to_send_to_rogue);
+    $response = dosomething_rogue_update_rogue_reportback_items($updates_to_send_to_rogue);
     if ($response->code != 201) {
       drupal_set_message(print_r($response->code, TRUE) . ' ' . print_r($response->status_message, TRUE), 'warning');
     } else {
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
@@ -299,7 +299,8 @@ function dosomething_reportback_files_form_submit($form, &$form_state) {
         $data = [
           'rogue_event_id' => $rogue_item_id[0]->rogue_event_id,
           'status' => $rogue_status,
-          'reviewer' => dosomething_user_get_field('field_northstar_id'),
+          'event_type' => 'post_photo',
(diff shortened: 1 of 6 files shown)


katiecrane:
I'm confused why this says `Create a new signup (RB) and photo (reportback item) from Phoenix web.` Why is it `signup (RB)`?

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
```diff
@@ -56,9 +56,9 @@ function dosomething_rogue_retry_failed_reportbacks() {
       } else {
         $data = [
           [
-            'rogue_reportback_item_id' => $task->rogue_item_id,
+            'rogue_event_id' => $task->rogue_event_id,
             'status' => $task->status,
-            'reviewer' => $task->reviewer,
+            'event_type' => 'post_photo',
```
katiecrane, line 61: Are reviews a `post_photo` type in Rogue?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
```diff
@@ -287,19 +287,18 @@ function dosomething_reportback_files_form_submit($form, &$form_state) {
...
           $rogue_status = 'accepted';
         } else {
           $rogue_status = 'rejected';
         }
 
         $data = [
-          'rogue_reportback_item_id' => $rogue_item_id[0]->rogue_item_id,
+          'rogue_event_id' => $rogue_item_id[0]->rogue_event_id,
           'status' => $rogue_status,
-          'reviewer' => dosomething_user_get_field('field_northstar_id'),
```
katiecrane, line 302: Do we not want to store this anymore?

chloealee:
@katiecrane sorry about that confusion! I said signup (RB) because in the new world/Rogue terms, it is a signup but in the old world/Phoenix, it is a reportback. Same with reportback item/photo.

katiecrane:
@chloealee ahhh got it got it, thanks!

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
chloealee: @katiecrane for now, yes! Kept this as `post_photo` since I remember when we were talking about the signup business that we shouldn't add `update_signup` as the `event_type` yet and not get too wrapped up in the naming quiet yet. Happy to change though if we think this is a good time to discuss `event_type`s!

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
chloealee: In the `photos` table, we don't have this column anymore like we do in the `reportback_items` table. There is a bit more explanation about this in [this PR](https://github.com/DoSomething/rogue/pull/112) (again, sorry for the back and forth!)

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
katiecrane: `source` lives on the photo though so I was expecting `something_photo`, I probably missed the decision on this but just wondering because it's not really a post. I totally agree with the signup situation though that part is confusing.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.admin.inc
katiecrane: @chloealee got it! I think we will want something like that for that change log stuff on the reportbacks but we'll see what everyone else says too!

sbsmith86: @chloealee @katiecrane we do have a `reviews` table in rogue that was supposed to store stuff like this but I don't know if we have fully implemented that yet. I think we can talk more about how we want to handle this because we haven't really thought through it all yet.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
sbsmith86: I think we should have a `review` type if we can. That can be in a separate PR if it keeps things smaller.

chloealee: I can update that here! I think you both are right because on the Rogue side, for these updates, I've changed the `submission_type` to `admin review` instead of `user`. This will make more sense with both these changes.

sbsmith86 (approved):
Looks ok to me. 

Two things we should probably discuss sooner than later based on questions in this PR:

* codifying a list of event types based on what we know now
* Talking about the `reviews` table and what we should or shouldn't use it for

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.cron.inc
chloealee: wait sorry I spoke too soon! This actually has to be `post_photo`  because that is a validation rule on Posts in Rogue. To fix this though, I just [hardcoded this](https://github.com/DoSomething/rogue/pull/112/commits/27cdefcfa9f6288215e0989f4c23f0b004ed009b) in on the Rogue side to save this as a `review` `event_type`.

Issue #7296, opened 2017-01-30
[Feature] Affiliate Messaging Opt-ins


weerd:
_FEATURE OVERVIEW_
-------------------
#### User Story
As a user I would like to be able to opt-in (out) to receive email messaging from DoSomething affiliates when I sign up for a campaign, regardless of whether I'm signed in.

#### Additional Information (optional)
Add a toggle and label underneath the campaign signup button on the pitch page (for both unauthenticated and authenticated users) to allow them to decide whether they want messaging from affiliates or not. Also allow campaign leads to add more information for the user, as well as an ability to export from the database a CSV list of users that _did_ sign up.

#### Why This Matters
We want to give our affiliates an opportunity to contact users that have signed up for a campaign they are specifically sponsoring or are a partner with.

#### Tasks
This feature includes the following tasks:
- [x] add form for specifying affiliate information on a per campaign basis [#7286](https://github.com/DoSomething/phoenix/pull/7286)
- [x] update the pitch page to output the opt-in data [#7286](https://github.com/DoSomething/phoenix/pull/7286)
- [x] style the pitch page opt-in data
- [x] save record of users that opt-in(out) to affiliate messaging per campaign [#7292](https://github.com/DoSomething/phoenix/pull/7292)
- [x] add export form and allow it to export CSV of opt-in users for a campaign [#7306](https://github.com/DoSomething/phoenix/pull/7306)
- [x] enable affiliate opt-in(out) on pitch page for authenticated users [#7292](https://github.com/DoSomething/phoenix/pull/7292)
- [x] enable affiliate opt-in(out) on pitch page for unauthenticated users [#7301](https://github.com/DoSomething/phoenix/pull/7301)
- [x] Fix styling of message callout when opt-in/out is present

#### Checklist
- [x] Tested on staging.
- [x] Pinged a PM if this is a larger PR that would benefit from some additional testing love
- [x] Post a message in #phoenix if this includes something that causes a rebuild!

ngjo:
![](https://github.trello.services/images/mini-trello-icon.png) [(5) As a biz dev, I want to build an Opt-out Option for 3rd party messaging during sign up.](https://trello.com/c/YmS6rixd/216-5-as-a-biz-dev-i-want-to-build-an-opt-out-option-for-3rd-party-messaging-during-sign-up)

PR #7301, opened 2017-02-01
Messaging opt in for unauth users

weerd:
#### What's this PR do?
This PR adds some styles to the messaging opt-in feature and also enables unauthenticated users to be able to Signup and opt-in _after_ logging in via Northstar.

👁 

There are some small CSS related quirks that result from this messaging opt-in but seems to mostly apply to member created campaigns.

#### Relevant tickets
Refs #7296

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
@@ -152,9 +152,11 @@ function dosomething_northstar_set_authorization_action($action, $nid, $params =
   }
 
   $_SESSION['openid_connect_action'] = [
-    'signup' => $nid,
-    'params' => [
-      'affiliate_messaging_opt_in' => isset($params['affiliate_messaging_opt_in']) ? $params['affiliate_messaging_opt_in'] : FALSE,
+    'signup' => [
+      'nid' => $nid,
+      'params' => [
+        'affiliate_messaging_opt_in' => isset($params['affiliate_messaging_opt_in']) ? $params['affiliate_messaging_opt_in'] : FALSE,
+      ],
     ],
   ];
 }
@@ -176,9 +178,9 @@ function dosomething_northstar_perform_authorization_actions() {
   foreach ($actions as $action => $target) {
     if ($action === 'signup') {
       $source = dosomething_signup_get_query_source();
-      dosomething_signup_user_signup((int) $target, $user, $source, $actions['params']);
+      dosomething_signup_user_signup((int) $target['nid'], $user, $source, $target['params']);
 
-      drupal_goto('node/' . $target);
+      drupal_goto('node/' . $target['nid']);
     }
   }
 }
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
```diff
@@ -135,15 +142,17 @@ function dosomething_northstar_openid_connect_post_authorize($tokens, $account,
...
+ * @param array $param
  */
-function dosomething_northstar_set_authorization_action($action, $nid) {
+function dosomething_northstar_set_authorization_action($action, $nid, $params = []) {
   if ($action !== 'signup') {
     return;
   }
 
   $_SESSION['openid_connect_action'] = ['signup' => $nid];
+  $_SESSION['affiliate_messaging_opt_in'] = isset($params['affiliate_messaging_opt_in']) ? $params['affiliate_messaging_opt_in'] : FALSE;
```
DFurnes: I think this makes sense to minimize the number of unique keys we store on the session and include as part of the existing `signup` post-auth action:

```php
$_SESSION['openid_connect_action'] = [
    'signup' => [
        'nid' => $nid,
        'params' => [
                'messaging_opt_in' => ! empty($params['affiliate_messaging_opt_in'])
        ],
    ]
];
```

Then we can grab all of the details of the signup from one place:

```php
 function dosomething_northstar_perform_authorization_actions() {
    global $user;
 
    if (! isset($_SESSION['openid_connect_action'])) {
     return;
    }
  
    $actions = $_SESSION['openid_connect_action'];

    foreach ($actions as $action => $target) {
        if ($action === 'signup') {
            $source = dosomething_signup_get_query_source();
            dosomething_signup_user_signup((int) $target['nid'], $user, $source, $target['params']);
        }
    }
}
```

DFurnes (commented):
One suggestion for queueing up the "opt-in", but otherwise looks good!

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
weerd: Ah, good call. For some reason I had it in my head that `$_SESSION['openid_connect_action'] = ['signup' => $nid];` needed to be separate. Ok, will update!

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.auth.inc
```diff
@@ -151,8 +151,12 @@ function dosomething_northstar_set_authorization_action($action, $nid, $params =
     return;
   }
 
-  $_SESSION['openid_connect_action'] = ['signup' => $nid];
-  $_SESSION['affiliate_messaging_opt_in'] = isset($params['affiliate_messaging_opt_in']) ? $params['affiliate_messaging_opt_in'] : FALSE;
+  $_SESSION['openid_connect_action'] = [
+    'signup' => $nid,
+    'params' => [
```
DFurnes: Should this be part of the `signup` "action", since it doesn't do anything on it's own? It feels awkward to loop over these but then "grab" this one on the `signup` iteration.

DFurnes (approved):
Ship it! ⛵️

Issue #7304, opened 2017-02-02
Some accounts are unable to log in due to inability to read auth code response.


DFurnes:
_BUG_
-------------------
#### Current Behavior
Some (hopefully small) subset of users are seeing a 404 page when Northstar redirects them back to Phoenix as part of the authorization code flow. Here's an example of a URL that 404s:

```
https://www.dosomething.org/us/openid-connect/northstar?code=U91hsBpSrlLVvr4afvUl1pJu98K8xhUwSrun3Jz5tj4GhW2ESODPR6Pnp2skIBeixd0QRwv8aTAWPIEn8a2hGD12wvE%2FqJdUENdhr1L%2B6Mbwhjb7HNepHSMrFmrsA6ISt5fHwMIK7hcjVrS42ZiNZOVarpO1g9Lz7ObZOfNLKnCfQF08ICsSPZG%2FyRgjkbsMKiArVvX0xzwvc61tjSQ1lVChjaKxc7yNSVM6ASwnVrx9rTf1Q7%2FosW75ei4L%2F98w8sPA0i2SbqnnDJJFEhdj8JftEF3do2Qp1rPq4fg%2BpxYN%2Fcg5uq1fze%2FEUmVv9k16np%2BFmc0bTrK0raMF0fXvwnLVWKGMImdmqLliIaBn%2FCN0CsMZcYeNHofC8jG5w10NDc%2FqHE71JvB24vYQJXwJt8kwLFf%2B4V2D35%2BVnhWQPcNLYG1hhNvfTKz1BmGPb59Ik5qE53HOmCDWMhKFEvrazACNcd05GFGeGFpfAe%2FFalRtbsFKigWqWafmJKcREAAcXxnOkzawhWGyOtN5%2BDdVtHuSBBD5Z6KwbD3E472JtgrPaF%2Bd%2F2laboxUbfrRyVTWEx0LRQGxAjn%2BZjxJyk2%2BX%2B1r5OdjshbIG5ycAkZJwDQFZ5Ghnn1RQqBenxu%2Bso8vo0vwOYWdfiaDLy4yUgkOEnLtbSRve8LDS2MZmeVa76s%3D&state=f0d97e3c28ee80e6bbc2940d4d3fc0af
```

For comparison, a working callback URL from my last login looks like this:
```
[...]

Issue #7302, opened 2017-02-01
Some accounts are unable to login due to no authmap record.


DFurnes:
_BUG_
-------------------
#### Current Behavior
Some accounts are unable to log in because they receive an "email address was already taken" error. From the few accounts I've manually looked at, this is because the account's record in the `authmap` table was deleted somehow.

![screen shot 2017-02-01 at 3 36 29 pm](https://cloud.githubusercontent.com/assets/583202/22525182/3c593fde-e894-11e6-8cb5-a4b68cc3af34.png)

#### Desired Behavior
I had [made a little script to fix affected accounts](https://github.com/DoSomething/phoenix/pull/7262) before ⛄️ but forgot to run it when we got back in the new year. We should run that to fix the 5,208 affected accounts (many of which are because of invalid email addresses).

It also makes sense to remove the ability for the `dosomething_northstar_save_id_field` function to unset the `authmap` record, since we do not ever delete accounts in Northstar.

#### Steps to Replicate
…

#### Why This Matters
We want to make sure all users can log in!

🐛

PR #7309, opened 2017-02-06
Fixes email script & a bug where a user's Northstar ID field could be wiped.

DFurnes:
#### What's this PR do?
This pull request updates the `sanitize-email-addresses.php` to work correctly, since I had removed the "sync" functions when refactoring Northstar code last week. This will fix up ~5k users with missing authmap records so we can more easily see if the problem in #7302 is still occurring.

This also includes an update to the `dosomething_northstar_save_id_field` so that it will not automatically remove a user's `authmap` record if it's given a bad `$response` body – there's a chance that a bad response (like a 503) from Northstar could have caused issues for these users.

👀

This also removes some noise to help with debugging #7304.

#### Relevant tickets
References #7302.

#### Checklist
- [ ] Tested on staging.


weerd (approved):
Looks good! 👍

PR #7310, opened 2017-02-06
Custom social share modal 

### LINKED ISSUE #7303: Custom share modal after sign up
_FEATURE OVERVIEW_
-------------------
#### User Story
As an admin, I want the custom share modal to pop-up immediately after someone signs up for a campaign that doesn't have competitions on it. 

#### Why This Matters
We want to add this in effort to increase shares (OKR). 

Right now, a user can share a campaign with their own custom share link however, it is a little more hidden than we'd like. We want to make the custom share modal pop out to the use to encourage more shares.

chloealee:
#### What's this PR do?
- Adds a checkbox under `SIGNUP DATA FORM` to enable custom share modal after signup. 
- Doesn't make `Confirmation message` field required any longer. 

- Run `drush updb`. 
- Make sure `Collect additional signup data` and `Enable Custom Share Modal` checkboxes are checked. 
- Under `CUSTOM SOCIAL SHARING` in `Custom Settings`, customize social sharing. If these are not filled out, no image will appear and the problem fact will populate custom social message. 

Wanted to get this up as quickly as possible. If we want, I can play around with having both competitions and custom social sharing possible next! @ngjo can we sync about this? 

#### Relevant tickets
Fixes #7303 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.share.inc
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.share.inc
@@ -164,3 +164,4 @@ function dosomething_helpers_share_bar($path, $share_types, $share_description, 
 
   return $share_bar . "</ul>";
 }
+
--- a/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.share.inc
+++ b/lib/modules/dosomething/dosomething_helpers/dosomething_helpers.share.inc
@@ -165,3 +165,52 @@ function dosomething_helpers_share_bar($path, $share_types, $share_description, 
   return $share_bar . "</ul>";
 }
 
+/**
+ * Return social share types and custom content.
+ *
+ * @param object $campaign
+ *   The campaign object to get custom content from.
+ * @return array $social share types
+ *   An array of social share types and custom content.
+ */
+
+function dosomething_helpers_campaign_page_social_share_types($campaign) {
+  $custom_social_copy_variables = [
+    'custom_facebook_copy' => dosomething_helpers_get_variable('node', $campaign->nid, 'custom_facebook_copy'),
+    'custom_twitter_copy' => dosomething_helpers_get_variable('node', $campaign->nid, 'custom_twitter_copy'),
+    'custom_tumblr_copy' => dosomething_helpers_get_variable('node', $campaign->nid, 'custom_tumblr_copy'),
+  ];
+
+  foreach ($custom_social_copy_variables as $key => $value) {
+    if ($value) {
+      $$key = $value;
+    }
+    elseif (isset($campaign->fact_problem['fact'])) {
+      $$key = $campaign->fact_problem['fact'];
+    }
+    else {
+      $$key = NULL;
+    }
+  }
+
+  $social_share_types = array(
+    'facebook' => array(
+      'type' => 'feed_dialog',
+      'parameters' => array(
+      'picture' => dosomething_image_get_themed_image_url(dosomething_helpers_get_variable('node', $campaign->nid, 'share_image_nid'), 'landscape'),
+      'caption' => $custom_facebook_copy,
+    ),
+    ),
+    'twitter' => array(
+      'tweet' => $custom_twitter_copy,
+    ),
+    'tumblr' => array(
+      'posttype' => 'photo',
+      'content' => dosomething_image_get_themed_image_url(dosomething_helpers_get_variable('node', $campaign->nid, 'share_image_nid'), 'landscape'),
+      'caption' => $custom_tumblr_copy,
+    ),
+  );
+
+  return $social_share_types;
+
+}
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
@@ -316,11 +316,21 @@ function dosomething_signup_user_signup_data_form($form, &$form_state, $signup) 
     '#value' => $signup->run_nid,
     '#access' => FALSE,
   );
-  $form['header'] = array(
-    '#prefix' => '<div class="modal__block"><h2 class="heading">',
-    '#markup' => $config['form_header'],
-    '#suffix' => '</h2></div>',
-  );
+
+  // If custom social share signup is turned on, use the $config['form_header'] from signup data form.
+  if ($config['custom_social_share_signup']) {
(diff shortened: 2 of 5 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -309,11 +316,94 @@ function dosomething_signup_user_signup_data_form($form, &$form_state, $signup)
...
+    ];
+
+    $campaign = node_load($signup->nid);
+
+    foreach ($custom_social_copy_variables as $key => $value) {
+      if ($value) {
+        $$key = $value;
+      }
+      elseif (isset($campaign->field_fact_problem['en'][0]['entity']->title)) {
+        $$key = $campaign->field_fact_problem['en'][0]['entity']->title;
```
chloealee: I was having trouble getting the problem fact for some reason and just realized that this won't work for all campaigns because of the translation :-/ Is there a better way to get this?

chloealee: In some tpl files at the tope, it says `$campaign: A campaign object. @see dosomething_campaign_load()` - I searched for `dosomething_campaign_load()` but wasn't able to find this function. Does anyone know how this is used?

DFurnes: @chloealee Here's [the function](https://github.com/DoSomething/phoenix/blob/4bd401efb9d40440e7d169ae00b27a7f22d2b89b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.helpers.inc#L7-L293) that grabs that variable. It looks like it gets the country code by URL prefix and loads translated fields using that, so hopefully it works out of the box?

/cc @weerd for Phoenix API knowledge

chloealee: ah thank you @DFurnes ! Facepalm - I was searching with `()` at the end so couldn't that's why I couldn't find it :(

weerd: gah! @chloealee see if you can use the new `Campaign` class I added during the API stuff. You can use `Campaign::get()` and pass a node id. It'll return the campaign and should have the problem fields on it :)

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -55,6 +55,7 @@ function dosomething_signup_node_signup_data_form($form, &$form_state, $node) {
     '#weight' => 60,
     '#collapsible' => TRUE,
     '#collapsed' => FALSE,
+    '#description' => t('Note: you must either chose Competition Signup or Enable Custom Share Modal, cannot have both.'),
     // Config fieldset should only be visible if the status field is checked.
```
chloealee: Note: chose not to make Competition Signup and Enable Custom Sharing Modal as radio buttons and keep as checkboxes in case we want functionality to have them both to be enabled in the future.

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
chloealee: cool that works, thanks @weerd !

chloealee: ah @weerd but the `field_fact_problem` is still under by language type. I think I'll stick with using `dosomething_campaign_load()` as it seems to work better for this scenario!

weerd: image

hmmm, this is how it returns from the API which does a bit of transforming but should still be easy to extract. It should already account for the language. If it works with the other format that's cool too. Just thought it would be more straightforward since the Campaign class was easier to use in other scenarios but might not apply to this one! No worries :)

lib/modules/dosomething/dosomething_signup/dosomething_signup.signup_data_form.inc
```diff
@@ -309,11 +301,94 @@ function dosomething_signup_user_signup_data_form($form, &$form_state, $signup)
...
+      }
+      elseif (isset($campaign->fact_problem['fact'])) {
+        $$key = $campaign->fact_problem['fact'];
+      }
+      else {
+        $$key = NULL;
+      }
+    }
+
+    $social_share_types = array(
```
sbsmith86: Is this the same configuration from the other sharing functionality on the page? If it is, do we have to do it again here?

chloealee: yes sorry about that! Just pulled out into a helper function.

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.share.inc
```diff
@@ -164,3 +164,53 @@ function dosomething_helpers_share_bar($path, $share_types, $share_description,
...
+/**
+ * Return social share types and custom content.
+ *
+ * @param object $campaign
+ *   The campaign object to get custom content from.
+ * @return array $social share types
+ *   An array of social share types and custom content.
+ */
+
+function dosomething_helpers_social_share_types($campaign) {
```
sbsmith86: Sorry this is super nitpicky. But i feel like this should be a helper function in the `dosomething_campaign.theme.inc` because it is specific to campaign page functionality, it's just being used in two places on the page. So then you should be able to call the helper once and then have that array set to use where ever you need it.

chloealee: @sbsmith86 that makes sense to move it there! Just updated it. However, I'm a little confused about the second part. I'm calling it two files - `dosomething_campaign.theme.inc` and `dosomething_signup-.signup_data_form.inc`. Is this wrong to do?

sbsmith86:
@chloealee Oh I see, I guess that is fine, maybe leave the helper in `dosomething_helpers` but rename it to something specific indicating it is just for custom sharing on campaign pages?

chloealee:
@sbsmith86 ok cool! Just updated it to add more specificity.

sbsmith86 (approved):
:shipit:

PR #7314, opened 2017-02-21
adds check to see if rogue collection is on

### LINKED ISSUE #7313: Don't send reportbacks from the API to Rogue if feature flag isn't checked
_BUG_
-------------------
#### Current Behavior
- If a reportback comes in from the API, it always gets sent to Rogue, even if the feature flag isn't checked (it only checks if there is a `X-Request-ID` before sending to Rogue). 

#### Desired Behavior
- The reportback should only be sent to Rogue if the Rogue feature flag is checked.

chloealee:
#### What's this PR do?
- Adds a check to see if Rogue collection is turned on before sending RBs to Rogue from API.

- Turn off Rogue collection. 
- Submit reportback via `/api/v1/campaigns/:nid/reportback` endpoint with no `X-Request-Id` header.
- Reportback should be saved in Phoenix database and should return `rbid` as response.

- @aaronschachter reportbacks from the API _might_ have been saved to the database even though the response was `null`. When testing, I saw that the rbs actually were saved. 

#### Relevant tickets
Fixes #7313

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -406,7 +406,7 @@ function _campaign_resource_reportback($nid, $values) {
 
   $headers = getallheaders();
 
-  if (! isset($headers['X-Request-Id'])) {
+  if (variable_get('rogue_collection', FALSE) && ! isset($headers['X-Request-Id'])) {
     $rogue_reportback = dosomething_rogue_send_reportback_to_rogue($values, $user);
 
     // Make sure reportback made it back to Phoenix and store reference to the Rogue item if one was added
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -415,10 +415,12 @@ function _campaign_resource_reportback($nid, $values) {
     return $rb_info['rbid'];
   }
   else {
-    // Increment transaction id and log that the request has been received with new transaction id.
-    $incremented_transaction_id = dosomething_api_increment_transaction_id($headers['X-Request-Id']);
+    if ($headers['X-Request-Id']) {
```
sbsmith86, line 418: What is this check for? Above you check if it is set, but if it is not, you check if it is set again?

chloealee: Originally, if it comes to the `else` portion, it would have meant that it was already sent to Rogue and this is when Rogue sends back to Phoenix. Rogue uses Gateway's API client which sets up the `X-Request-Id`. However, if it is coming from the API, we know that it doesn't have the `X-Request-Id` yet. So if the Rogue Collection flag is turned off but it is still coming from the API, the `X-Request-Id` header won't be set and the below lines will break.

sbsmith86 (approved):

PR #7316, opened 2017-02-23
Adds Reaction User param to Reportback endpoints

itsjoekent:
#### What's this PR do?
This pull request adds a `reaction_user` parameter to the RB + RB Item index endpoints. This parameter lets you override what the `current_user` is in the Kudos response.

*Why do we need this?*
When making requests from a generic admin api account, the current user is not properly reflected in the Kudos response. Additionally, we have no other way to query whether a given user has reacted to a reportback item.

*Why didn't you just edit the `dosomething_kudos_group_by_taxonomy_term`?*
That was my initial idea, less hacky, simpler, etc. Unfortunately, there is several layers of functions between this & the API response, and it would look quite ugly to pass a "user override" through that many functions (I think)

Hit these two endpoints (Change the reaction_user with other users as well)

`http://dev.dosomething.org:8888/api/v1/reportback-items?reaction_user=1700225`
`http://dev.dosomething.org:8888/api/v1/reportbacks?reaction_user=1700229`

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -467,4 +467,19 @@ abstract class Transformer {
     ];
   }
 
+  /**
+   * Set the global user to a different account based on Northstar id.
+   *
+   * @param  String $northstarUserId
+   */
+  protected function asNorthstarUser($northstarUserId) {
+    if (!user_access('view any reportback')) return;
+
+    $userOverride = dosomething_user_get_user_by_northstar_id($northstarUserId);
+    if ($userOverride) {
+      global $user;
+      $user = $userOverride;
+    }
+  }
+
 }
--- a/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
@@ -92,11 +92,11 @@ function _reportback_item_resource_definition() {
             'default value' => FALSE,
           ],
           [
-            'name' => 'reaction_user',
-            'description' => 'Override the current user in respect to Reaction data',
+            'name' => 'as_user',
+            'description' => 'Override the current global user',
             'optional' => TRUE,
-            'type' => 'int',
-            'source' => ['param' => 'reaction_user'],
+            'type' => 'string',
+            'source' => ['param' => 'as_user'],
             'default_value' => NULL,
           ],
         ],
@@ -141,19 +141,10 @@ function _reportback_item_resource_access($op) {
  * @param  bool    $random
  * @param  int     $page
  * @param  bool    $load_user
- * @param  int     $reaction_user
+ * @param  string  $as_user
  * @return array
  */
-function _reportback_item_resource_index($campaigns, $exclude, $status, $count, $random, $page, $load_user, $reaction_user) {
-  if ($reaction_user) {
-    $reaction_user = user_load($reaction_user);
-
-    if ($reaction_user) {
-      global $user;
-      $user = $reaction_user;
-    }
-  }
-
+function _reportback_item_resource_index($campaigns, $exclude, $status, $count, $random, $page, $load_user, $as_user) {
   $parameters = array(
     'campaigns' => $campaigns,
     'exclude' => $exclude,
@@ -162,6 +153,7 @@ function _reportback_item_resource_index($campaigns, $exclude, $status, $count, 
     'random' => $random,
     'page' => $page,
     'load_user' => $load_user,
+    'as_user' => $as_user,
   );
 
   $reportbackItems = new ReportbackItemTransformer;
--- a/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
@@ -91,6 +91,14 @@ function _reportback_item_resource_definition() {
             'source' => array('param' => 'load_user'),
             'default value' => FALSE,
           ],
+          [
+            'name' => 'reaction_user',
+            'description' => 'Override the current user in respect to Reaction data',
+            'optional' => TRUE,
+            'type' => 'int',
+            'source' => ['param' => 'reaction_user'],
+            'default_value' => NULL,
+          ],
         ],
         'access callback' => '_reportback_item_resource_access',
         'access arguments' => array('index'),
@@ -133,9 +141,19 @@ function _reportback_item_resource_access($op) {
  * @param  bool    $random
  * @param  int     $page
  * @param  bool    $load_user
+ * @param  int     $reaction_user
  * @return array
  */
-function _reportback_item_resource_index($campaigns, $exclude, $status, $count, $random, $page, $load_user) {
+function _reportback_item_resource_index($campaigns, $exclude, $status, $count, $random, $page, $load_user, $reaction_user) {
+  if ($reaction_user) {
+    $reaction_user = user_load($reaction_user);
+
+    if ($reaction_user) {
+      global $user;
+      $user = $reaction_user;
+    }
+  }
+
   $parameters = array(
     'campaigns' => $campaigns,
     'exclude' => $exclude,
--- a/lib/modules/dosomething/dosomething_api/resources/reportback_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/reportback_resource.inc
@@ -157,7 +157,7 @@ function _reportback_resource_access($op) {
  * @param  bool    $random
  * @param  int     $page
  * @param  bool    $load_user
- * @param  int     $reaction_user
+ * @param  string  $reaction_user
  * @param  bool    $flagged
  * @return array
  */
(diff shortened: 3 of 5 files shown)


lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
```diff
@@ -133,9 +141,19 @@ function _reportback_item_resource_access($op) {
  * @param  bool    $random
  * @param  int     $page
  * @param  bool    $load_user
+ * @param  int     $reaction_user
  * @return array
  */
-function _reportback_item_resource_index($campaigns, $exclude, $status, $count, $random, $page, $load_user) {
+function _reportback_item_resource_index($campaigns, $exclude, $status, $count, $random, $page, $load_user, $reaction_user) {
+  if ($reaction_user) {
+    $reaction_user = user_load($reaction_user);
```
DFurnes: We should load the user by their Northstar ID here (since ideally we don't want to have to keep track of `drupal_id` in Phoenix Next)!

lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
```diff
@@ -91,6 +91,14 @@ function _reportback_item_resource_definition() {
             'source' => array('param' => 'load_user'),
             'default value' => FALSE,
           ],
+          [
+            'name' => 'reaction_user',
```
DFurnes: Do you think this makes more sense as `user` or `as_user` since it really sets the user for the entire request?

lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
```diff
@@ -133,9 +141,19 @@ function _reportback_item_resource_access($op) {
...
+function _reportback_item_resource_index($campaigns, $exclude, $status, $count, $random, $page, $load_user, $reaction_user) {
+  if ($reaction_user) {
+    $reaction_user = user_load($reaction_user);
+
+    if ($reaction_user) {
+      global $user;
+      $user = $reaction_user;
+    }
+  }
+
```
weerd: The above logic shouldn't happen in the resource file. It's supposed to just pass the params into the `$parameters` array and then this logic should happen elsewhere... Maybe in the transformer as a separate function called in the `index()` method.

lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
```diff
@@ -133,9 +141,19 @@ function _reportback_item_resource_access($op) {
...
  * @param  bool    $load_user
+ * @param  int     $reaction_user
  * @return array
  */
-function _reportback_item_resource_index($campaigns, $exclude, $status, $count, $random, $page, $load_user) {
+function _reportback_item_resource_index($campaigns, $exclude, $status, $count, $random, $page, $load_user, $reaction_user) {
+  if ($reaction_user) {
+    $reaction_user = user_load($reaction_user);
+
+    if ($reaction_user) {
```
DFurnes: Should we only allow overriding the global `$user` when the API request is made as an admin? Otherwise I think this could be used by a "malicious" user (well, sorta… 😜) to gain the "view all" privileges of an admin and see flagged or unapproved reportbacks.

DFurnes (commented):
Left some thoughts!

lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
itsjoekent: 🤔  Good point, `as_user` might make more sense

lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
itsjoekent: Well it's only on the `index` resource, and all of the data on that is public, so I don't think there is anything malicious to exploit?

lib/modules/dosomething/dosomething_api/resources/reportback_resource.inc
```diff
@@ -149,10 +157,20 @@ function _reportback_resource_access($op) {
...
-function _reportback_resource_index($ids, $campaigns, $status, $count, $random, $page, $load_user, $flagged, $runs) {
+function _reportback_resource_index($ids, $campaigns, $status, $count, $random, $page, $load_user, $reaction_user, $flagged, $runs) {
+  if ($reaction_user) {
+    $reaction_user = user_load($reaction_user);
+
+    if ($reaction_user) {
+      global $user;
+      $user = $reaction_user;
+    }
+  }
```
weerd: Same same. Maybe put this as a function in the reportback module file, and then it's called in the respective `index()` methods.

weerd (changes requested):
Some comments to help DRY things up and cleanup the resource files.

lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
DFurnes: Don't we use the current user to limit whether people can see flagged or unapproved reportbacks? I used the word "malicious" sorta in jest, but seems like something to lock down.

itsjoekent: Right but this piece of code doesn't touch the review endpoints or anything else, it's only effecting the `get`

DFurnes: I'm sure I'm missing something, but doesn't the `status` parameter on these endpoints get passed to [`removeUnauthorizedStatuses`](https://github.com/DoSomething/phoenix/blob/a05bd197bea695eed6cd0ff58d796d94a886f3a0/lib/modules/dosomething/dosomething_reportback/includes/ReportbackItemTransformer.php#L86-L111) which relies on the global `$user` for the `user_access` check?

weerd: > Otherwise I think this could be used by a "malicious" user

Yeah, I was kinda wondering about that as well...

itsjoekent: Oh hm, you're right about that one @DFurnes

itsjoekent:
I think I might scrap this PR & see if I can fix up the Kudos index instead, I don't think this approach will work

DFurnes:
@deadlybutter I don't think you should scrap this PR... we still want to be able to load in kudos when getting reportbacks for the gallery.

itsjoekent:
@DFurnes Any ideas on a better way of doing this then? As I briefly outlined in the PR description, I could change the root function which is handling this (`dosomething_kudos_group_by_taxonomy_term`) but getting the `as_user` parameter down there doesn't seem like fun / would look good

Which is why I'm leaning on fixing the Kudos endpoint itself, as maybe that'll be a bit easier then refactoring a bunch of functions

DFurnes:
I think your current approach seems fine! No need to restart. I'd just suggest putting a check to `user_access('view all reportbacks')` before you set the global `$user` by the provided ID.

itsjoekent:
OH, I see what you're saying 😅

itsjoekent:
It's friday right?

itsjoekent:
So in the latest round of changes...

- Added `user_access` check ✅ 
- Pulled the code out of the resource file & into a single function on the base transformer 🏖 
- Switched the name to `as_user` and treating it as a Northstar ID 🌠 

To test it...

1) Hit /reportbacks?as_user=fdsfdsfsdfsd (& /reportback-items). Repeat inside an incognito window.

lib/modules/dosomething/dosomething_api/resources/reportback_resource.inc
```diff
@@ -149,10 +157,11 @@ function _reportback_resource_access($op) {
  * @param  bool    $random
  * @param  int     $page
  * @param  bool    $load_user
+ * @param  int     $reaction_user
```
weerd: `string` right?

itsjoekent: good catch!

lib/modules/dosomething/dosomething_api/includes/Transformer.php
```diff
@@ -467,4 +467,19 @@ protected function transformUser($data) {
...
+   *
+   * @param  String $northstarUserId
+   */
+  protected function asNorthstarUser($northstarUserId) {
+    if (!user_access('view any reportback')) return;
+
+    $userOverride = dosomething_user_get_user_by_northstar_id($northstarUserId);
+    if ($userOverride) {
+      global $user;
+      $user = $userOverride;
```
weerd, line 481: Does this actually override the `global` user? Or do you need to `return` this user back as the thing to actually use moving forward?

itsjoekent: Yep this overrides the actual global user

DFurnes (approved):
Woo! Looks good to me! 🚀

lib/modules/dosomething/dosomething_api/includes/Transformer.php
weerd: Cool cool 👍

weerd (approved):
Look good! 👍

PR #7317, opened 2017-02-23
Rogue signup, reportback, and reportback file migration

katiecrane:
#### What's this PR do?
Migrates all of the things to Rogue!
- Builds a table to keep track of failed requests for debugging
- Sends the following to Rogue:
    - For signups with no reportbacks:
        - `northstar_id`
        - `campaign_id`
        - `campaign_run_id`
        - `do_not_forward` (so the data is not sent back to Phoenix)
        - `source`
        - `created_at`
        - `updated_at`
    - Signups with reportbacks also include the following:
        - `northstar_id` (repeated so Rogue can create the Post more easily)
        - `why_participated`
        - `quantity_pending`
        - `quantity` (`quantity` will be `NULL` and `quantity_pending` will hold the value if any of the files are `flagged` or `pending`)
        - `source`
        - `remote_addr`
        - `caption`
        - `'event_type' => 'post_photo'`
        - `northstar_id`
        - `status` (Rogue status)
        - `'do_not_forward' => TRUE` (so the data is not sent back to Phoenix)
        - `submission_type => 'user` (needed by Rogue)
        - `file`
        - `created_at`

Make sure I didn't forget any data! See if this is okay to run on Thor (aka won't mess up any existing Phoenix data). We will be clearing Rogue and Rogue references as far as I know.

I ran the whole thing through on my local and it a) didn't stop and b) looks like all the data appears in Rogue as expected and c) looks like all the Phoenix Rogue reference data is correct 🎉 

#### Checklist
- [ ] Tested on staging.


scripts/export-activity-to-rogue.php
```diff
@@ -0,0 +1,177 @@
...
+        $data['photo'][$key] = [
+          'source' => $photo->source,
+          'remote_addr' => $photo->remote_addr,
+          'caption' => $photo->caption,
+          'event_type' => 'post_photo',
+          'northstar_id' => $northstar_id->id,
+          'status' => $rogue_status,
+          'do_not_forward' => TRUE,
+          'submission_type' => 'user',
+          'file' => dosomething_helpers_get_data_uri_from_fid($photo->fid),
```
sbsmith86: has this module been imported anywhere? or is it globally available?

katiecrane: @sbsmith86 hmm good question. I think it's globally available because this line was behaving as expected

DFurnes: Yeah, functions like `dosomething_helpers_get_data_uri_from_fid` are accessible everywhere (which is crazy, but it's the life we live). The only thing we'd potentially have to worry about is if the Drupal module wasn't enabled, but that shouldn't be the case here.

sbsmith86: ah interesting.

scripts/export-activity-to-rogue.php
```diff
@@ -0,0 +1,177 @@
+<?php
+/**
+ * Script to export Signups, Reportbacks, and whatever else from drupal into Rogue.
+ *
+ * to run:
+ * drush --script-path=../scripts/ php-script export-activity-to-rogue.php
+ */
+
+include_once('../lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module');
+include_once('../lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module');
```
DFurnes: You shouldn't have to do this since Drupal will load the modules when it bootstraps.

scripts/export-activity-to-rogue.php
```diff
@@ -0,0 +1,177 @@
...
+                      WHERE signups.sid > $last_saved AND signups.sid NOT IN (SELECT sid FROM dosomething_rogue_signups)
+                      ORDER BY signups.sid");
+}
+else {
+  // Get ALL the things!
+  $signups = db_query('SELECT signups.sid, signups.uid, signups.nid, signups.run_nid, signups.source, signups.timestamp, rb.quantity, rb.why_participated, rb.rbid, rb.flagged, rb.updated
+                      FROM dosomething_signup signups
+                      LEFT JOIN dosomething_reportback rb ON signups.uid = rb.uid AND signups.nid = rb.nid AND signups.run_nid = rb.run_nid
+                      WHERE signups.sid NOT IN (SELECT sid FROM dosomething_rogue_signups)
+                      ORDER BY signups.sid')->fetchAll();
```
DFurnes: This miiiight cause trouble on production, since the `->fetchAll()` will try to load all of the signup records into an array in memory before continuing on past this line. If you just save the `db_query('...')` to `$signups` here, Drupal will use a cursor to iterate over the results in the foreach loop (so it should only load a little bit at a time).

scripts/export-activity-to-rogue.php
```diff
@@ -0,0 +1,177 @@
...
+}
+
+$client = dosomething_rogue_client();
+
+foreach ($signups as $signup) {
+  $data = [];
+  echo 'Trying sid ' . $signup->sid . '...' . PHP_EOL;
+
+  $northstar_id = dosomething_northstar_get_user($signup->uid, 'drupal_id');
```
DFurnes: This might make more sense as `$northstar_user`, since otherwise it seems like it'd be a string ID.

DFurnes (commented):
This is one heck of a script. 📚 Left a few comments!

scripts/export-activity-to-rogue.php
katiecrane: @DFurnes that's what I tried to do originally but my local machine was getting mad at me. Let me see what error it was giving. Regardless, definitely seems like a good idea to try it the way you suggested on Thor.

scripts/export-activity-to-rogue.php
katiecrane: @DFurnes good call!

DFurnes (approved):
👍 from me!

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.install
```diff
@@ -376,3 +500,14 @@ function dosomething_rogue_update_7011(&$sandbox) {
...
+/**
+ * Adds dosomething_rogue_failed_migrations table.
+ */
+function dosomething_rogue_update_7012(&$sandbox) {
+  $table_name = 'dosomething_rogue_failed_migrations';
+  if (!db_table_exists($table_name)) {
+    $schema = dosomething_rogue_schema();
+    db_create_table($table_name, $schema[$table_name]);
+  }
+}
```
sbsmith86: new line 😬

sbsmith86:
looks good to me :)

PR #7324, opened 2017-03-14
Adding partner embed to settings and template

itsjoekent:
#### What's this PR do?
Adds the partner embed option to the settings (per campaign)
Renders the markup on the template

screen shot 2017-03-14 at 11 35 40 am

screen shot 2017-03-14 at 11 35 59 am

does the buttons do what they say they do

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
+++ b/lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
@@ -241,7 +241,7 @@
             <h2 class="heading -emphasized"><span><?php print t('Step 3: Do It'); ?></span></h2>
           </div>
         </div>
-        <?php if (!empty($partner_embed_markup)): ?>
+        <?php if (isset($partner_embed_markup)): ?>
           <div class="container__row">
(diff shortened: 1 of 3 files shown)


itsjoekent:
@weerd Ready for review!

lib/themes/dosomething/paraneue_dosomething/templates/campaign/node--campaign.tpl.php
```diff
@@ -241,6 +241,11 @@
             <h2 class="heading -emphasized"><span><?php print t('Step 3: Do It'); ?></span></h2>
           </div>
         </div>
+        <?php if (!empty($partner_embed_markup)): ?>
```
weerd: Should this be `!isset()`? If the partner embed isn't enabled, then the variable is never set.

itsjoekent: 👍

itsjoekent:
@weerd Had to be `isset` not `!is..` otherwise all good on my end!

weerd:
oh yeah, duh. tee hee

itsjoekent:
@weerd Thanks for closing my PR! 😆

PR #7331, opened 2017-03-27
Update to fix some bugs with /signups endpoint.

### LINKED ISSUE #7329: Error when providing Northstar ID to signups endpoint.
_BUG_
-------------------
#### Current Behavior
@aaronschachter reported a bug with the Phoenix signups endpoint [in Slack](https://dosomething.slack.com/archives/C02BBRN5A/p1490378610150414):
> hey @weerd -- i remember you mentioned a change made to the Phoenix signups to accept NS User ID. I'm wondering if it could maybe be causing a really strange, and new -- i believe -- issue I'm seeing in Gambit https://github.com/DoSomething/gambit/issues/827#issuecomment-289100468

Likely the specific error from the logs:
```
Notice: Trying to get property of non-object in dosomething_user_convert_to_legacy_id():
line 1164 of var/www/www.dosomething.org/releases/20170323151344/lib/modules/
dosomething/dosomething_user/dosomething_user.module.
```

#### Desired Behavior
We should be able to make requests to this endpoint with either Drupal IDs or Northstar IDs and receive a successful response.

#### Steps to Replicate
See linked issue comment above.

#### Why This Matters
We need to be able to fetch a user's signups without errors! 💣

📶

weerd:
#### What's this PR do?
This PR provides an update to the `/signups` endpoint to allow passing through user data properly for each signup item, including the Northstar ID for the user. It also makes an update to fix a bug which was causing the endpoint to break at times if no `users=...` url parameter was added to the endpoint API request.

The `/signups` endpoints should work better now. As a bonus aside, I think this also fixed up the index `/signups` endpoint so now it actually returns something (on my local anyhoo 😬)

🔍 

The [Trying to get property of non-object](https://github.com/DoSomething/phoenix/issues/7329) issue was related to the problem with the `users` url parameter.

The issue with the [missing NS ID](https://github.com/DoSomething/gambit/issues/827#issuecomment-289496978) in the response was because when switching to use the Phoenix `/signups` endpoint, we weren't passing along the user's NS ID. Prior to switching to using this endpoint, the endpoint was proxied via Northstar and NS was doing its due-diligence and likely auto-filling it out for us.

#### Relevant tickets
Fixes #7329 
Refs https://github.com/DoSomething/gambit/issues/827

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/includes/Signup.php
+++ b/lib/modules/dosomething/dosomething_signup/includes/Signup.php
@@ -86,11 +86,22 @@ class Signup extends Entity {
    * @param $data
    */
   private function build($data) {
+    $northstar_user = (object) [];
+
     $this->id = $data->sid;
     $this->created_at = $data->timestamp;
 
+    if (module_exists('dosomething_northstar')) {
+      $northstar_user = dosomething_northstar_get_user($data->uid, 'drupal_id');
+    }
+
     $this->user = [
-      'drupal_id' => $data->uid,
+      'drupal_id' => data_get($northstar_user, 'drupal_id') ?: $data->uid,
+      'id' => data_get($northstar_user, 'id') ?: $data->field_northstar_id_value,
+      'first_name' => data_get($northstar_user, 'first_name'),
+      'last_initial' => data_get($northstar_user, 'last_initial'),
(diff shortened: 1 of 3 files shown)


aaronschachter (approved):
Looks ok here 👀

PR #7332, opened 2017-03-27
One feature flag to turn Rogue on/off 

### LINKED ISSUE #7330: Create one feature flag to turn Rogue on/off
_FEATURE OVERVIEW_
-------------------
#### User Story
As a developer, I want to make one feature flag to turn Rogue on/off so that it isn't in two feature flags.

Right now, sending signups and RBs are different feature flags - let's combine them. 

Note: This won't be flipped until after we have done the gallery and migration work.

#### Why This Matters
So that turning Rogue on/off is in one flag. When flipped on, this will send Signups, Reportbacks. 

Two separate feature flags. 

#### Definition of Done
One feature flag - when flipped, it sends signups and RBs to Rogue.

chloealee:
#### What's this PR do?
- Combines signup and reportback Rogue feature flags into one. 
  - Updates signup to check if `rogue_collection` is on instead of `rogue_signup_collection`

- Go to `/admin/config/services/rogue`. You should now only see one checkbox as follows:
screen shot 2017-03-27 at 3 12 14 pm
- When signing up and reporting back on Phoenix (when the box is checked), all data should be reflected in Rogue as well. Right now (3/27/17), we can't test this though because Rogue needs to be updated. However, can confirm that able to see that the flag itself works - when checked, the reportback item/signup gets sent to the failed task log and when not checked, all works as expected. 

#### Relevant tickets
Fixes #7330 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -292,7 +292,7 @@ function _campaign_resource_signup($nid, $values) {
   // Check to see if the request came from Rogue and if we should be sending it there
   // $headers['X-Request-Id'] will only be set if the request is coming from Rogue
   if ( !isset($headers['X-Request-Id'])) {
-    if (variable_get('rogue_signup_collection', FALSE)) {
+    if (variable_get('rogue_collection', FALSE)) {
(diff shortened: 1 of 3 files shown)


katiecrane (approved):
Lookin good to me!

PR #7335, opened 2017-03-30
Switch to authorization_code for Phoenix <-> Northstar requests

crittermike:
#### What's this PR do?

This PR makes Phoenix talk to Northstar using the `authorization_code` grant as opposed to the `client_credentials` grant, which in turns means telling Northstar about individual user accounts.

Besides the visual code review, testing would be very helpful. Check out the branch, then log into your local and do operations that require communicating with Northstar.

I am not sure that this is REALLY working as expected. I can't seem to encounter any errors when using the site, but I'm also not sure how to ensure that methods like `PhoenixOAuthBridge::getUser()` even run, to test them. My testing so far has mostly comprised of doing everything I can find on the site to do, and fixing any errors that I see when doing it.

**Important: also see the 2 places that have `@TODO` statements - those are open questions that I have!**

#### Relevant tickets

I'm not sure of the ticket number(s).

### DIFF
--- a/lib/app/Auth/PhoenixOAuthBridge.php
+++ b/lib/app/Auth/PhoenixOAuthBridge.php
@@ -24,16 +24,11 @@ class PhoenixOAuthBridge implements OAuthBridgeContract {
    * @return NorthstarUserContract|null
    */
   public function getUser($id) {
-    $query = new EntityFieldQuery();
-    $query->entityCondition('entity_type', 'user')
-      ->fieldCondition('field_northstar_id', 'value', $id);
-    $results = $query->execute();
-    if (empty($results)) {
-      return null;
+    $user = dosomething_user_get_user_by_northstar_id($id);
+    if (empty($user)) {
+      return FALSE;
     }
-
-    $uid = key($results['user']);
-    $account = new PhoenixOAuthUser($uid);
+    $account = new PhoenixOAuthUser($user->uid);
     return $account;
   }
 
@@ -44,12 +39,7 @@ class PhoenixOAuthBridge implements OAuthBridgeContract {
    * @return NorthstarUserContract
    */
   public function getOrCreateUser($id) {
-    $account = $this->getUser($id);
-    if (!empty($account)) {
-      return $account;
-    }
-
-    // @TODO: Create user here, but with what info for username/email?
+    // This method can stay unimplemented for now because it isn't called.
   }
 
   /**
--- a/lib/app/Auth/PhoenixOAuthBridge.php
+++ b/lib/app/Auth/PhoenixOAuthBridge.php
@@ -94,7 +94,7 @@ class PhoenixOAuthBridge implements OAuthBridgeContract {
    * @return void
    */
   public function requestUserCredentials() {
-    user_logout();
+    $this->logout();
   }
 
   /**
@@ -134,6 +134,7 @@ class PhoenixOAuthBridge implements OAuthBridgeContract {
    * @return mixed
    */
   public function logout() {
+    module_load_include('pages.inc', 'user');
     user_logout();
   }
 
--- a/lib/app/Auth/PhoenixOAuthBridge.php
+++ b/lib/app/Auth/PhoenixOAuthBridge.php
@@ -13,7 +13,9 @@ class PhoenixOAuthBridge implements OAuthBridgeContract {
    * @return NorthstarUserContract|null
    */
   public function getCurrentUser() {
-    //
+    $user = new PhoenixOAuthUser();
+    // Grab Northstar ID and assign it here.
+    return $user;
   }
 
   /**
--- a/lib/app/Auth/PhoenixOAuthBridge.php
+++ b/lib/app/Auth/PhoenixOAuthBridge.php
@@ -49,7 +49,7 @@ class PhoenixOAuthBridge implements OAuthBridgeContract {
       return $account;
     }
 
-    // @TODO: Create user here.
+    // @TODO: Create user here, but with what info for username/email?
   }
 
   /**
--- a/lib/app/Auth/PhoenixOAuthBridge.php
+++ b/lib/app/Auth/PhoenixOAuthBridge.php
@@ -13,9 +13,9 @@ class PhoenixOAuthBridge implements OAuthBridgeContract {
    * @return NorthstarUserContract|null
    */
   public function getCurrentUser() {
-    $user = new PhoenixOAuthUser();
-    // Grab Northstar ID and assign it here.
-    return $user;
+    global $user;
+    $account = new PhoenixOAuthUser($user->uid);
+    return $account;
   }
 
   /**
@@ -24,7 +24,17 @@ class PhoenixOAuthBridge implements OAuthBridgeContract {
    * @return NorthstarUserContract|null
    */
   public function getUser($id) {
-    //
+    $query = new EntityFieldQuery();
+    $query->entityCondition('entity_type', 'user')
+      ->fieldCondition('field_northstar_id', 'value', $id);
+    $results = $query->execute();
+    if (empty($results)) {
+      return null;
+    }
+
+    $uid = key($results['user']);
+    $account = new PhoenixOAuthUser($uid);
+    return $account;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


lib/app/Auth/PhoenixOAuthBridge.php
```diff
@@ -22,7 +24,17 @@ public function getCurrentUser() {
    * @return NorthstarUserContract|null
    */
   public function getUser($id) {
-    //
+    $query = new EntityFieldQuery();
```
DFurnes: I think it would make sense to use [`dosomething_user_get_user_by_northstar_id`](https://github.com/DoSomething/phoenix/blob/2c0ee83bd7ed57e9b4afb72fe06467c06f2ff3dd/lib/modules/dosomething/dosomething_user/dosomething_user.module#L1053-L1071) here, and then pass the returned `$user` object into the `PhoenixOAuthUser` constructor.

lib/app/Auth/PhoenixOAuthBridge.php
```diff
@@ -22,7 +24,17 @@ public function getCurrentUser() {
...
+    $query->entityCondition('entity_type', 'user')
+      ->fieldCondition('field_northstar_id', 'value', $id);
+    $results = $query->execute();
+    if (empty($results)) {
+      return null;
+    }
+
+    $uid = key($results['user']);
+    $account = new PhoenixOAuthUser($uid);
```
DFurnes: I don't think it's likely to happen, but this should probably check if the user exists first:

```php
if (! $account) {
    return null;
}

return new PhoenixOAuthUser($account);
```

lib/app/Auth/PhoenixOAuthBridge.php
```diff
@@ -32,7 +44,12 @@ public function getUser($id) {
    * @return NorthstarUserContract
    */
   public function getOrCreateUser($id) {
-    //
+    $account = $this->getUser($id);
+    if (!empty($account)) {
+      return $account;
+    }
+
+    // @TODO: Create user here, but with what info for username/email?
```
DFurnes: This is probably safe to leave unimplemented since this method is only used during the authorization code grant. At least for now, [we let Drupal's `openid_connect` module handle this](https://github.com/sanduhrs/drupal-openid_connect/blob/82ba5ed175a6cc8ae72124112c975a8b35c2e151/openid_connect.module#L344-L369) part of the auth flow.

lib/app/Auth/PhoenixOAuthUser.php
```diff
@@ -0,0 +1,107 @@
...
+   *
+   * @return AccessToken|null
+   */
+  public function getOAuthToken() {
+    if (empty($this->user)) {
+      return null;
+    }
+    return new AccessToken([
+      'resource_owner_id' => $this->getNorthstarIdentifier(),
+      'access_token' => $this->user->field_access_token[LANGUAGE_NONE][0]['value'],
```
DFurnes: We made [a little helper method](https://github.com/DoSomething/phoenix/blob/3a6bce989ce69896daa099930e91631c2c90948c/lib/modules/dosomething/dosomething_user/dosomething_user.module#L1171-L1186) for grabbing fields like this:

```diff
-'access_token' => $this->user->field_access_token[LANGUAGE_NONE][0]['value'],
+'access_token' => dosomething_user_get_field('field_access_token', $this->user),,
```

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -21,14 +21,10 @@ define('NORTHSTAR_URL', variable_get('dosomething_northstar_url', 'http://norths
...
-    'grant' => 'client_credentials',
+    'grant' => 'authorization_code',
     'url' => NORTHSTAR_URL,
     'bridge' => PhoenixOAuthBridge::class,
-    'client_credentials' => [
-      'client_id' => variable_get('dosomething_northstar_app_id', 'trusted-test-client'),
-      'client_secret' => variable_get('dosomething_northstar_app_key', 'secret1'),
-      'scope' => ['admin', 'user'],
-    ],
+    // @TODO: Does other stuff need to go here?
```
DFurnes: The client ID and secret for the auth code grant should be set here:

```php
    'authorization_code' => [
        'client_id' => 'example',
        'client_secret' => 'xxxxxxxxxxxxx',
        'scope' => ['user'],
        'redirect_uri' => 'user/login',
    ],
```

You should be able to grab the client ID & secret from the [`openid_connect` module's variables](https://github.com/sanduhrs/drupal-openid_connect/blob/82ba5ed175a6cc8ae72124112c975a8b35c2e151/openid_connect.module#L98).

DFurnes (commented):
Looking good! Left some comments.

I think you should be able to see the result of a couple requests when viewing the gallery at the bottom of the "action page" for a campaign. For example, on [Test Campaign](http://dev.dosomething.org:8888/us/campaigns/test-campaign) I'm pretty sure we should make a handful of requests to get user info for the people in the gallery. If you're not seeing anything happen, make sure to clear cache since we do cache those responses.

DFurnes (approved):
This seems good to me! I'll pull this down and do some more thorough testing on Monday.

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -20,14 +20,17 @@ define('NORTHSTAR_URL', variable_get('dosomething_northstar_url', 'http://norths
...
  */
 function dosomething_northstar_client() {
+  $oauth_client = variable_get('openid_connect_client_northstar',
+    array('client_id' => NULL, 'client_secret' => NULL));
   return new Northstar([
-    'grant' => 'client_credentials',
+    'grant' => 'authorization_code',
     'url' => NORTHSTAR_URL,
     'bridge' => PhoenixOAuthBridge::class,
-    'client_credentials' => [
```
DFurnes: We should keep this config array in case we do want to make requests on behalf of the client later on.

```diff
     'grant' => 'authorization_code',
     'url' => NORTHSTAR_URL,
     'bridge' => PhoenixOAuthBridge::class,
+    'client_credentials' => [
+      'client_id' => variable_get('dosomething_northstar_app_id', 'trusted-test-client'),
+      'client_secret' => variable_get('dosomething_northstar_app_key', 'secret1'),
+      'scope' => ['admin', 'user'],
+    ],
     'authorization_code' => [
       'client_id' => $oauth_client['client_id'],
       'client_secret' => $oauth_client['client_secret'],
```

lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
```diff
@@ -20,14 +20,17 @@ define('NORTHSTAR_URL', variable_get('dosomething_northstar_url', 'http://norths
...
-      'client_id' => variable_get('dosomething_northstar_app_id', 'trusted-test-client'),
-      'client_secret' => variable_get('dosomething_northstar_app_key', 'secret1'),
-      'scope' => ['admin', 'user'],
+    'authorization_code' => [
+      'client_id' => $oauth_client['client_id'],
+      'client_secret' => $oauth_client['client_secret'],
+      'scope' => ['user'],
```
DFurnes: Let's pull in the same scopes we request when users log in here.

```diff
- 'scope' => ['user'],
+ 'scope' => openid_connect_get_scopes(),
```

lib/app/Auth/PhoenixOAuthBridge.php
```diff
@@ -22,7 +24,12 @@ public function getCurrentUser() {
    * @return NorthstarUserContract|null
    */
   public function getUser($id) {
-    //
+    $user = dosomething_user_get_user_by_northstar_id($id);
+    if (empty($user)) {
+      return FALSE;
```
DFurnes: Nitpicky, but this should be `null` to match the interface for this method.

DFurnes (changes requested):
I pulled this down to test on my local & noticed a few little things. After that, looks ready to ship!

DFurnes:
Let's also force [requests in `dosomething_northstar_get_user `](https://github.com/DoSomething/phoenix/blob/caa4158e89f13dcb9f8aa79c3e413e0d0c790e70/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module#L102) to be made by the client for now, since we cache those responses locally in the Phoenix database:

```diff
-  $northstar_user = dosomething_northstar_client()->getUser($type, $id);
+  $northstar_user = dosomething_northstar_client()->asClient()->getUser($type, $id);
```

crittermike:
@DFurnes Thanks! Those updates have been made now - all ready for final review.

DFurnes (approved):

DFurnes:
@mikecrittenden Looks great! Merging! ⛵️

PR #7338, opened 2017-04-03
Updates to logic for post RBs via the API.

weerd:
#### What's this PR do?
This PR updates the `/api/v1/campaigns/:nid/reportback` endpoint to update the logic so that submitted Reportbacks are not submitted as the DS API User, instead the DS API User just grants access for posting, but then the RB is posted as the actual user. We were running into scenarios where depending on how the logic ran, the system would default to using the `global $user` for posting RBs which was causing problems for the submission gallery in Phoenix-Next to accurately showcase member submissions.

Successful insert and update as the specified `uid` user:
image

One breaking change to note, is that now the `uid` is a required URL parameter to successfully post a RB as the user, otherwise an error response is returned.

The endpoint is also updated to provide some useful HTTP error responses depending on issues with a request.

🔎 👁 

This was a necessary update to get things working correctly for the new campaign experience provided by Phoenix-Next.

#### Relevant tickets
Refs: https://trello.com/c/X1X46JMq/424-8-as-as-user-i-want-to-be-able-to-upload-photos-that-aren-t-sized-for-ants

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -333,10 +333,10 @@ function _campaign_resource_signup($nid, $values) {
  *   The Node nid to post the reportback to.
  * @param array $values
  *   The reportback data to post. Expected keys:
- *   - uid: The user uid (int).  Optional, uses global $user if not set.
- *   - file: Base64 encoded file string to save.
- *   - filename: The filename of the file provided as file.
- *   - file_url: The URL of the reportback file to save (used if no file/filename exist).
+ *   - uid: The user uid (int).
+ *   - file: Base64 encoded file string to save (required if using 'filename').
+ *   - filename: The filename of the file provided as file (required if using 'file').
+ *   - file_url: The URL of the reportback file to save (used if no 'file'/'filename' exist).
  *   - caption: The caption for the Reportback File if provided.
  *   - quantity (int).
  *   - why_participated (string).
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -333,73 +333,73 @@ function _campaign_resource_signup($nid, $values) {
...
     $result = _file_resource_create($values);
+
     $file = file_load($result['fid']);
   }
+  elseif (!empty($values['file_url'])) {
+    $file = dosomething_reportback_save_file_from_url($nid, $uid, $values['file_url']);
 
-  if (!$file) {
-    // If this is to create a new reportback, there must be an image
-    if (!array_key_exists('file_url', $values) && !$rbid) {
```
weerd, line 380: Not really sure what the aim of this check was...

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
```diff
@@ -1023,3 +1023,21 @@ function dosomething_helpers_extract_values_by_keys($array, $desired_keys)
...
 }
+
+/**
+ * Return a basic HTTP response.
+ *
+ * @param  Integer $code
+ * @param  String $message
+ * @return Array
+ */
+function dosomething_helpers_basic_http_response($code, $message) {
```
DFurnes, line 1034: 👏

DFurnes (approved):
This looks great to me! ⛵️

lib/modules/dosomething/dosomething_helpers/dosomething_helpers.module
weerd: tee hee ☺️

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
katiecrane: @weerd I believe this was to make sure that new reportbacks always came with images

weerd: Oh ok. The logic felt kinda weird since it's separated from the above alternative file related params as well... I'm guessing that should also be a validation error in func that actually saves RB if it's the first RB... but ya, Drupal ;)

aaronschachter:
Tested Gambit Thor 2.8.26 and was able to create a reportback, update it (via going through the flow a 2nd time to add a new item/caption/quantity), and verified I could create a different reportback for a 2nd campaign. 👍 

I did notice this[ message in the Phoenix logs](https://thor.dosomething.org/admin/reports/event/4692133), not sure if it's an issue?

````
Notice: Undefined index: X-Request-Id in _campaign_resource_reportback() (line 418 of /var/www/www.dosomething.org/releases/20170403220238/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc).
````

PR #7339, opened 2017-04-03
Adding HTTP error response if first time post and missing file.

weerd:
#### What's this PR do?
This PR make s quick fix for the `/api/v1/campaigns/:nid/reportbacks` endpoint after @katiecrane [clarified something in the old logic](https://github.com/DoSomething/phoenix/pull/7338#discussion_r109524502) that was confusing to me based on how it was written. 

This makes sure that if this is the first Reportback submission for the specified campaign by the specified user, than a `$file` is required, otherwise the endpoint responds with an HTTP error response.

👁 

#### Relevant tickets
Refs #7338


katiecrane (approved):
I think this covers it!

PR #7342, opened 2017-04-04
add unique keys to northstar id field

sbsmith86:
#### What's this PR do?

This PR adds indexes to the `northstar_id` and the `entity_id` fields in `field_data_northstar_id` table.

In new relic, it was showing that this query was taking a lot of time, because there were no indexes on this table. 

```  
SELECT f.entity_id AS entity_id
FROM 
field_data_field_northstar_id f
WHERE  (f.field_northstar_id_value = :db_condition_placeholder_?)
```

one commit to review, make sure keys are being set correctly. 

**QUESTION:** Looking at the query above are there any other keys that should be set on this table to make it run faster?

I did work in gladiator to use the Phoenix API `/signups` endpoint to grab user activity for competitions. 

This was done to solve an issue in gladiator where no activity was showing across all campaigns.

This change then broke functionality in gladiator that would cause timeouts when grabing user activity for bulk users. 

#### Relevant tickets
Fixes #

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.install
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.install
@@ -102,3 +102,13 @@ function dosomething_user_uninstall() {
     variable_del($var);
   }
 }
+
+/**
+ * Add a unique constraint to the `field_data_field_northstar_id` field for the
+ * northstar_id and entitiy_id fields.
+ */
+function dosomething_user_update_7010() {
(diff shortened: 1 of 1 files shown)


DFurnes:
> __QUESTION:__ Looking at the query above are there any other keys that should be set on this table to make it run faster?

Nothing comes to mind... let's try merging this in and see how things perform on Thor!

DFurnes (approved):

PR #7343, opened 2017-04-04
Update to use Rogue `/reportbacks` endpoint to power gallery if Rogue is turned on 

### LINKED ISSUE #7334: Update Phoenix to hit Rogue's GET /reportbacks endpoint that lives behind the Rogue feature flag
_FEATURE OVERVIEW_
-------------------
#### User Story
As a developer, I want Phoenix to be updated so that it's hitting Rogue's `GET /reportbacks` endpoint so that we can have Rogue powering the gallery. 

We need to update the gallery in Phoenix to point to Rogue. 
We also want this to only be turned on if Rogue's feature flag is turned on.

#### Why This Matters
Preliminary work for Rogue to power the gallery.

chloealee:
#### What's this PR do?
Updates client and server side in Phoenix to check if Rogue flag is enabled. If so, change the logic so Rogue is powering the gallery instead of Phoenix. 

- Turn Rogue feature flag on.
- Visit a campaign page with Rogue reportbacks. 
- You should see the gallery with reportbacks from Rogue. 
- When you click `View more`, the reportbacks should also be from Rogue. 

#### Relevant tickets
Fixes #7334 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -681,9 +681,8 @@ function dosomething_reportback_preprocess_entity(&$variables) {
   }
 
   if (variable_get('rogue_collection', FALSE)) {
-    //@TODO: make sure this is the correct Rogue URL
     $rogue_base_url = variable_get('dosomething_rogue_url', 'https://rogue.dosomething.org/api');
-    $apiUrl = $rogue_base_url . '/v2/reportbacks';
+    $apiUrl = $rogue_base_url . '/v1/reportbacks';
   }
   else {
     $apiUrl = '/api/v1/reportback-items';
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -690,11 +690,11 @@ function dosomething_reportback_preprocess_entity(&$variables) {
   }
 
   drupal_add_js(
-    array('dosomethingReportback' =>
-      array(
+    [
+      'dosomethingReportback' => [
         'apiUrl' => $apiUrl,
-      )
-    ),
+      ]
+    ],
     'setting'
   );
 }
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -679,6 +679,24 @@ function dosomething_reportback_preprocess_entity(&$variables) {
     $variables['theme_hook_suggestions'][] = 'reportback';
     $entity = $variables['elements']['#entity'];
   }
+
+  if (variable_get('rogue_collection', FALSE)) {
+    //@TODO: make sure this is the correct Rogue URL
+    $rogue_base_url = variable_get('dosomething_rogue_url', 'https://rogue.dosomething.org/api');
+    $apiUrl = $rogue_base_url . '/v2/reportbacks';
+  }
+  else {
+    $apiUrl = '/api/v1/reportback-items';
+  }
+
+  drupal_add_js(
+    array('dosomethingReportback' =>
+      array(
+        'apiUrl' => $apiUrl,
+      )
+    ),
+    'setting'
+  );
 }
 
 
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -21,7 +21,6 @@ include_once 'includes/ReportbackItem.php';
 include_once 'includes/ReportbackItemController.php';
 include_once 'includes/ReportbackItemTransformer.php';
 
-
 /**
  * Implements hook_entity_info().
  *
@@ -679,23 +678,6 @@ function dosomething_reportback_preprocess_entity(&$variables) {
     $variables['theme_hook_suggestions'][] = 'reportback';
     $entity = $variables['elements']['#entity'];
   }
-
-  if (variable_get('rogue_collection', FALSE)) {
-    $rogue_base_url = variable_get('dosomething_rogue_url', 'https://rogue.dosomething.org/api');
-    $apiUrl = $rogue_base_url . '/v1/reportbacks';
-  }
-  else {
-    $apiUrl = '/api/v1/reportback-items';
-  }
-
-  drupal_add_js(
-    [
-      'dosomethingReportback' => [
-        'apiUrl' => $apiUrl,
-      ]
-    ],
-    'setting'
-  );
 }
 
 
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
@@ -38,8 +38,8 @@ function dosomething_rogue_menu() {
  *
  * https://github.com/DoSomething/rogue
  */
-function dosomething_rogue_client() {
-  return new Rogue(ROGUE_API_URL . '/' . ROGUE_API_VERSION . '/',
+function dosomething_rogue_client($version = ROGUE_API_VERSION) {
+  return new Rogue(ROGUE_API_URL . '/' . $version . '/',
     ['headers' => ['X-DS-Rogue-API-Key' => ROGUE_API_KEY]]
   );
 }
@@ -52,9 +52,7 @@ function dosomething_rogue_client() {
   */
 function dosomething_rogue_get_reportback_items($inputs)
 {
-  $client = new Rogue(ROGUE_API_URL . '/v1/',
-    ['headers' => ['X-DS-Rogue-API-Key' => ROGUE_API_KEY]]
-  );
+  $client = dosomething_rogue_client('v1');
 
   $response = $client->getAllReportbacks($inputs);
 
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
@@ -52,7 +52,10 @@ function dosomething_rogue_client() {
   */
 function dosomething_rogue_get_reportback_items($inputs)
 {
-  $client = dosomething_rogue_client();
+  $client = new Rogue(ROGUE_API_URL . '/v1/',
+    ['headers' => ['X-DS-Rogue-API-Key' => ROGUE_API_KEY]]
+  );
+
   $response = $client->getAllReportbacks($inputs);
 
   return $response;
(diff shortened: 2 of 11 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -353,17 +353,20 @@ function paraneue_dosomething_get_themed_reportbacks_by_status($campaign_id, $st
...
     return NULL;
   }
 
-  if ($data->id && !dosomething_kudos_disable_reactions($data->campaign->id)) {
+  if ($data->id && !dosomething_kudos_disable_reactions($data->campaign->id) && !$rogue_rb_item) {
     $data->kudos = dosomething_kudos_get_kudos_data_for_fid($data->id);
   }
 
+  // @TODO: do we want to add kudos from Rogue here?
+
```
chloealee, line 369: Do we want to add reaction stuff from Rogue here or should I just keep this as a `TODO` until we make more concrete decisions about this? I think we'd have to make another call to the `/activity` endpoint in Rogue but might be a little tricky to get the information we need at the moment.

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
```diff
@@ -236,23 +236,41 @@ function paraneue_dosomething_preprocess_reportback_gallery(&$vars, $count = 6)
...
-  // # Approved Reportbacks:
-  $gallery['total_approved'] = dosomething_reportback_get_reportback_items_total_by_filters(['nid' => $gallery['campaign_id'], 'status' => 'approved']);
+    // # Approved Reportbacks:
+    $gallery['total_approved'] = dosomething_reportback_get_reportback_items_total_by_filters(['nid' => $gallery['campaign_id'], 'status' => 'approved']);
+  }
 
   // # Build the full initial gallery:
 
   // Collect Reportbacks to fulfill desired amount.
+  // @TODO: delete this comment but the below should be fine if we're okay with grabbing the placeholder still from Phoenix?
```
chloealee: Per comment, the below should be fine if we're okay with grabbing placeholders still from Phoenix, correct?

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
```diff
@@ -262,24 +280,40 @@ function paraneue_dosomething_preprocess_reportback_gallery(&$vars, $count = 6)
...
+    // Get total of Promoted & Approved RB Items for Campaign.
+    $gallery['total_sum'] = $gallery['total_promoted'] + $gallery['total_approved'];
+  }
 
   // Get total remaining after subtracting the RB Items already retrieved.
   $gallery['remaining'] = $gallery['total_sum'] - $gallery['prefetched'];
 
   global $user;
 
+  // @TODO: how do we deal with the below?
```
chloealee, line 313: Per the comment, how do we want to deal with the below?

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
```diff
@@ -236,23 +236,41 @@ function paraneue_dosomething_preprocess_reportback_gallery(&$vars, $count = 6)
     'random' => TRUE,
   ];
 
+  // If the Rogue feature flag is on, grab reportbacks from Rogue.
+  if (variable_get('rogue_collection', FALSE)) {
+    // Hit the GET /reportbacks endpoint in Rogue with campaign_id filter.
+    // On the Rogue side, it will only return 'approved', which will be shown in the gallery.
+    $promotedReportbackItems = dosomething_rogue_get_reportback_items(['campaign_id' => $gallery['campaign_id']])['data'];
```
chloealee: This only returns 20 at a time since this endpoint is paginated. I was thinking to check in the response if ``links`['next']` is not null in the `meta` tag, hit the next page and add to `$promotedReportbackItems` but I don't think this is the best way to handle this logic. Any suggestions @DFurnes @sbsmith86 @katiecrane ?

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -45,6 +45,23 @@ function dosomething_rogue_client() {
...
 
 /**
+  * Send a GET request to return all reportback items matching a given
+  * query from Rogue.
+  *
+  * @param array $inputs
+  */
+function dosomething_rogue_get_reportback_items($inputs)
+{
+  $client = new Rogue(ROGUE_API_URL . '/v1/',
```
DFurnes: You can use the `dosomething_rogue_client()` here so we don't have to duplicate this setup.

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
chloealee: or maybe I could set the `limit` to something ridiculously large?

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
DFurnes: Ideally Rogue should be able to send back reaction information within that first API call to `/reportbacks` and then pass down as part of `$data` so we don't need to make individual requests for each specific reaction.

That said, I also think it'd be _totally_ reasonable to just disable reactions on "legacy" campaigns since it doesn't get that much user interaction... worth considering as a team. 🤷‍♂️

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
DFurnes: Yeah, definitely okay to still grab placeholders from Phoenix in this case!

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
chloealee: @DFurnes I set up a new one because all this endpoint `GET /reportbacks` is `v1` in Rogue and all others are `v2`!

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
DFurnes: I think it's safe to only grab a maximum of 6 here since everything else is fetched when hitting the "view more" button in the user's browser (which will go directly to Rogue). If you need a count of total accepted (below), you can probably grab that from the `meta` on this response.

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
DFurnes: Hmm, good catch. The most responsible thing is probably setting up feature flags so that those calls can be made to Rogue (to both fetch reactions & save new ones). I also wonder if we can just remove reactions from old Phoenix, and _just_ have that as a feature on the new campaign template...?

DFurnes:
This is so exciting! Can't wait until we can flip the switch. 🔜

itsjoekent:
giphy 12

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
chloealee: ooo interesting! Should I flag to discuss in grooming tomorrow @DFurnes ?

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
chloealee: oh smart! Thank you, @DFurnes !

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -679,6 +679,23 @@ function dosomething_reportback_preprocess_entity(&$variables) {
     $variables['theme_hook_suggestions'][] = 'reportback';
     $entity = $variables['elements']['#entity'];
   }
+
+  if (variable_get('rogue_collection', FALSE)) {
+    $rogue_base_url = variable_get('dosomething_rogue_url', 'https://rogue.dosomething.org/api');
+    $apiUrl = $rogue_base_url . '/v1/reportbacks';
```
chloealee: @DFurnes I should probably be passing the `campaign_id` filter here, correct?

DFurnes: That will be appended to the URL when fetching each page [in the user's browser](https://github.com/DoSomething/phoenix/blob/caa4158e89f13dcb9f8aa79c3e413e0d0c790e70/lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js#L117-L124)!

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
DFurnes: Oh, hmm okay. I don't think we should be including versions in base URL in that case, but can handle that in a different pull request later on.

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
DFurnes: Yeah, let's talk over as a group then! 💬

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
chloealee: ah brilliant, thank you! It seems like when I click more, the reportbacks are being populated but doesn't seem like from Rogue. I don't know what I'm missing though?

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
chloealee: I agree! Made an issue [here](https://github.com/DoSomething/phoenix/issues/7345).

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
DFurnes: Hmm, strange! I can swing by and help investigate tomorrow. 🕵️ 🕵️‍♀️

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
chloealee: For documentation, we discussed in grooming and I'll tackle this in a separate PR!

lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_node.inc
chloealee: Also will tackle this in a separate PR!

DFurnes (approved):
Woooo! Ship it!! ⛵️ 📸 🎉

PR #7346, opened 2017-04-05
Fix rogue update hooks

sbsmith86:
#### What's this PR do?

Adds checks to update hooks in `dosomething_rogue` and `dosomething_user` so that the hooks don't fail and block other hooks from running .

#### Checklist
- [ ] Tested on staging.


weerd (approved):
Looks good to me! 👍

PR #7349, opened 2017-04-05
Fix issue with new users' campaign signups.

### LINKED ISSUE #7348: Campaign signups are giving access token error.
_BUG_
-------------------
#### Current Behavior
Campaign signups for new accounts are showing the following error on Thor, yikes!

screen shot 2017-04-05 at 1 44 29 pm

#### Desired Behavior
Campaign signups should happen without an error.

#### Steps to Replicate
Click the "Sign Up" button on a campaign page when unauthenticated. Then, register a new account as part of the signup flow –– you'll see that nasty error instead of the action page.

#### Why This Matters
We'd lose signups if we deployed this to production. No good!

#### Countries Affected (optional)
…

#### OS/Browser (optional)
…

- [Ghost Inspector failure](https://app.ghostinspector.com/tests/577fd31444d2fa94053ab2e0)

DFurnes:
#### What's this PR do?
This fixes #7348, and adds a little error-checking so this doesn't fail as bad in the future:

🔃 Shuffles around the post-auth actions so we don't try to perform them before finishing everything else – otherwise part of this function was getting skipped because campaign signups include a `drupal_goto`.

👮 Adds some checks in the `PhoenixOAuthUser` class so we don't try to create [an `AccessToken` without the required fields](https://github.com/thephpleague/oauth2-client/blob/56f6860e8773a2fc2bb4524f344d04bc28bb1c09/src/Token/AccessToken.php#L62-L64). This means we'd return `null` (and gracefully logout or try to refresh token) rather than giving the user a big ol' error message.

👀

#### Relevant tickets
Fixes #

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/app/Auth/PhoenixOAuthUser.php
+++ b/lib/app/Auth/PhoenixOAuthUser.php
@@ -71,11 +71,19 @@ class PhoenixOAuthUser implements NorthstarUserContract {
     if (empty($this->user)) {
       return null;
     }
+
+    $access_token = dosomething_user_get_field('field_access_token', $this->user);
+    $refresh_token = dosomething_user_get_field('field_refresh_token', $this->user);
+    $expires = dosomething_user_get_field('field_access_token_expiration', $this->user);
+    if (empty($access_token) || empty($refresh_token) || empty($expires)) {
+      return null;
+    }
+
     return new AccessToken([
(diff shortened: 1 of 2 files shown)


sbsmith86 (approved):

weerd (approved):
🔢  💯

PR #7350, opened 2017-04-05
Make sure get_user request is made as client.

DFurnes:
#### What's this PR do?
This updates the `dosomething_northstar_get_user` to always use the Client Credentials grant, so that it'll fetch the full profile before caching it. This also fixes this method so that it works in Drush scripts (where we won't have a logged in user).

👀

When running the Rogue export script, @katiecrane noticed this error:

```
PHP Fatal error:  Call to a member function getRefreshToken() on a non-object in /var/www/dev.dosomething.org/vendor/dosomething/gateway/src/AuthorizesWithNorthstar.php on line 191
```

This was happening because it couldn't refresh the token of a non-existant user.

#### Relevant tickets
References #7335.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -107,7 +107,7 @@ function dosomething_northstar_get_user($id, $type = 'id') {
     return $northstar_user->data;
   }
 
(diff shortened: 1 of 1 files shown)


katiecrane (approved):
Matches up with this comment https://github.com/DoSomething/phoenix/pull/7335#issuecomment-291635622
🚀

PR #7351, opened 2017-04-05
Ensure access token is always current.

DFurnes:
#### What's this PR do?
This pull request will check a user's access token on page load, and refresh it if it's expired. This is important because it lets Northstar know accurate activity information (to within 1 hour), which is necessary for tracking our MAU goal.

We should make sure this doesn't have any unexpected cases where a user will get logged out unexpectedly, and that it doesn't have any performance impact for normal page loads.

💹 

#### Relevant tickets
…

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
+++ b/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module
@@ -2,6 +2,7 @@
 
 use DoSomething\Gateway\Northstar;
 use Phoenix\Auth\PhoenixOAuthBridge;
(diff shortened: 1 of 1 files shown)


itsjoekent (approved):
Yeah, I agree I'm a tad worried it could impact site performance, but we can deal with that when the time comes I guess?

DFurnes:
Yeah I'm hoping not, since the `$user` is already loaded in memory and at that point it's just a quick integer comparison on the timestamps but we'll see! Definitely something to monitor in New Relic.

Issue #7353, opened 2017-04-12
Signups sometimes don't work.


DFurnes:
_BUG_
-------------------
#### Current Behavior
Sometimes Phoenix Next is receiving a "false" failure response when trying to create a signup:

```json
[false]
```

#### Desired Behavior
We should handle errors more gracefully.

#### Steps to Replicate
We can't! Oy.

#### Why This Matters
Users need to be able to sign up for a campaign!

#### Countries Affected (optional)
…

#### OS/Browser (optional)
…

#### Relevant Screenshots + Links
…

DFurnes:
There are a [couple branches](https://github.com/DoSomething/phoenix/blob/4bd401efb9d40440e7d169ae00b27a7f22d2b89b/lib/modules/dosomething/dosomething_signup/dosomething_signup.module#L271-L296) where the signup endpoint could return false...

PR #7354, opened 2017-04-12
Some cleanup in the signup API...?

DFurnes:
#### What's this PR do?
Was diving in to try to figure out why Phoenix Next signups were returning `[false]` sometimes, and did some cleanup along the way: added a few comments, and removed some code that was used when Rogue used to "echo" signups back to Phoenix.

I'd also like to try [returning `["<signup_id>"]` here if a signup exists](https://github.com/DoSomething/phoenix/commit/3a8bb42f986c858b29577c28134bedc473b4199d), rather than `[false]`. I don't think that's a super crazy thing to do, but would love a sanity check from a few folks...

👀

Ehhhh

#### Relevant tickets
References #7353.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -267,7 +267,6 @@ function _campaign_resource_index($ids, $staff_pick, $term_ids, $count, $random,
  *       Defaults to sending messages.
  */
 function _campaign_resource_signup($nid, $values) {
-
   if (isset($values['northstar_id'])) {
     $values['uid'] = dosomething_user_get_user_by_northstar_id($values['northstar_id'])->uid;
   }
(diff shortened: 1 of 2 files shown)


lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -287,42 +286,19 @@ function _campaign_resource_signup($nid, $values) {
     watchdog('dosomething_api', '_campaign_resource_signup values:' . json_encode($values));
   }
 
-  $headers = getallheaders();
```
weerd, line 290: So no longer need the `$headers` since not doing the `X-Request-Id` check?

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -254,9 +254,9 @@ function dosomething_signup_create($nid, $uid = NULL, $source = NULL, $timestamp
   $campaign = Campaign::get($nid);
   $run = dosomething_helpers_get_current_campaign_run_for_user($nid, $user, $campaign);
 
-  // If a signup already exists, exit.
-  if (dosomething_signup_exists($nid, $run->nid, $uid)) {
-    return FALSE;
+  // If a signup already exists, return the signup ID.
+  if ($existing_id = dosomething_signup_exists($nid, $run->nid, $uid)) {
+    return $existing_id;
```
weerd, line 259: Do we know if this will break existing usage of API expected responses?

DFurnes: Sure don't! 😈

weerd: Wears the paddle? :canoe:

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
DFurnes: Yeah, this part of the code block was never getting called.

weerd: hokay!

weerd (approved):
Cautious 👍

PR #7356, opened 2017-04-13
Reactions for current user in gallery powered by Rogue

### LINKED ISSUE #7358: Reactions for logged in user in gallery powered by Rogue
_FEATURE OVERVIEW_
-------------------
#### User Story
As a developer, if the current user has reacted to a reportback item, I want the reaction button to be active and this information to be pulled from Rogue. 

#### Why This Matters
Brings us closer to Rogue completely powering the gallery!

chloealee:
#### What's this PR do?
Pulls reaction data from Rogue into the gallery for the logged in user. 
- If the logged in user has reacted to any reportback items, the reaction buttons will be active for these reportback items in the gallery. 

Server Side Gallery Reaction Loading 
1. `preprocess_node.inc` lines `243-249`
2. `paraneue_dosomething_get_themed_reportback_item` gets called in ``preprocess_node.inc` on line `293`
3. `helpers.inc` lines `364-385`
4. `dosomething_kudos.module` lines `323-337` 
5. `photo.tpl.php` 

Client Side Gallery Reaction Loading 
`reportback_item_resource.inc` and `Reportback.js` files.

I know this code is pretty hacky so would love to get suggestions on how to do things better!

#### Relevant tickets
Fixes #7358

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
@@ -158,22 +158,24 @@ function _reportback_item_resource_index($campaigns, $exclude, $status, $count, 
 
   if (variable_get('rogue_collection', FALSE)) {
     $client = dosomething_rogue_client('v1');
-    // is there anyway to get the url string here so we can grab the page number param?
-    // $query_parts = explode("&", $_SERVER['REQUEST_URI']);
-    // $page = end($query_parts);
-    // // // page=2
-    // $page = explode("=", $page);
-    // $page = end($page);
-    // if ($page) {
-    //   $response = $client->getAllReportbacks([
-    //     'filter' => [
-    //       'campaign_id' => $campaigns,
-    //     ],
-    //     'as_user' => dosomething_user_get_field('field_northstar_id'),
-    //     'limit' => 12,
-    //     'page' => 3,
-    //   ]);
-    // } else {
+    $query_parts = explode("&", $_SERVER['REQUEST_URI']);
+    $last_query_param = end($query_parts);
+
+    if ($last_query_param === 'page') {
+      $page_parts = explode("=", $$last_query_param);
+      $page_number = end($page_parts);
+
+      if ($page_number) {
+        $response = $client->getAllReportbacks([
+          'filter' => [
+            'campaign_id' => $campaigns,
+          ],
+          'as_user' => dosomething_user_get_field('field_northstar_id'),
+          'limit' => 12,
+          'page' => $page_number,
+        ]);
+      }
+    } else {
       $response = $client->getAllReportbacks([
         'filter' => [
           'campaign_id' => $campaigns,
@@ -181,7 +183,7 @@ function _reportback_item_resource_index($campaigns, $exclude, $status, $count, 
         'as_user' => dosomething_user_get_field('field_northstar_id'),
         'limit' => 12,
       ]);
-    // }
+    }
 
     return $response;
   }
--- a/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
@@ -163,8 +163,8 @@ function _reportback_item_resource_index($campaigns, $exclude, $status, $count, 
         'campaign_id' => $campaigns,
       ],
       'as_user' => dosomething_user_get_field('field_northstar_id'),
-      'limit' => 6
     ]);
+
     return $response;
   }
 
--- a/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
@@ -158,13 +158,30 @@ function _reportback_item_resource_index($campaigns, $exclude, $status, $count, 
 
   if (variable_get('rogue_collection', FALSE)) {
     $client = dosomething_rogue_client('v1');
-    $response = $client->getAllReportbacks([
-      'filter' => [
-        'campaign_id' => $campaigns,
-      ],
-      'as_user' => dosomething_user_get_field('field_northstar_id'),
-      'limit' => 12,
-    ]);
+    // is there anyway to get the url string here so we can grab the page number param?
+    // $query_parts = explode("&", $_SERVER['REQUEST_URI']);
+    // $page = end($query_parts);
+    // // // page=2
+    // $page = explode("=", $page);
+    // $page = end($page);
+    // if ($page) {
+    //   $response = $client->getAllReportbacks([
+    //     'filter' => [
+    //       'campaign_id' => $campaigns,
+    //     ],
+    //     'as_user' => dosomething_user_get_field('field_northstar_id'),
+    //     'limit' => 12,
+    //     'page' => 3,
+    //   ]);
+    // } else {
+      $response = $client->getAllReportbacks([
+        'filter' => [
+          'campaign_id' => $campaigns,
+        ],
+        'as_user' => dosomething_user_get_field('field_northstar_id'),
+        'limit' => 12,
+      ]);
+    // }
 
     return $response;
   }
--- a/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
@@ -158,7 +158,13 @@ function _reportback_item_resource_index($campaigns, $exclude, $status, $count, 
 
   if (variable_get('rogue_collection', FALSE)) {
     $client = dosomething_rogue_client('v1');
-    $response = $client->getAllReportbacks($inputs);
+    $response = $client->getAllReportbacks([
+      'filter' => [
+        'campaign_id' => $campaigns,
+      ],
+      'as_user' => dosomething_user_get_field('field_northstar_id'),
+      'limit' => 6
+    ]);
     return $response;
   }
 
--- a/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
@@ -158,32 +158,16 @@ function _reportback_item_resource_index($campaigns, $exclude, $status, $count, 
 
   if (variable_get('rogue_collection', FALSE)) {
     $client = dosomething_rogue_client('v1');
-    $query_parts = explode("&", $_SERVER['REQUEST_URI']);
-    $last_query_param = end($query_parts);
-    $last_query_param_parts = explode("=", $last_query_param);
 
-    if ($last_query_param_parts[0] === 'page') {
-      $page_number = end($last_query_param_parts);
-
-      if ($page_number) {
-        $response = $client->getAllReportbacks([
-          'filter' => [
-            'campaign_id' => $campaigns,
-          ],
-          'as_user' => dosomething_user_get_field('field_northstar_id'),
-          'limit' => 12,
-          'page' => $page_number,
-        ]);
-      }
-    } else {
-      $response = $client->getAllReportbacks([
-        'filter' => [
-          'campaign_id' => $campaigns,
-        ],
-        'as_user' => dosomething_user_get_field('field_northstar_id'),
-        'limit' => 12,
-      ]);
-    }
+    $response = $client->getAllReportbacks([
+      'filter' => [
+        'campaign_id' => $campaigns,
+        'exclude' => $exclude,
+      ],
+      'as_user' => dosomething_user_get_field('field_northstar_id'),
+      'limit' => 12,
+      'page' => $page,
+    ]);
 
     return $response;
   }
--- a/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
@@ -160,10 +160,10 @@ function _reportback_item_resource_index($campaigns, $exclude, $status, $count, 
     $client = dosomething_rogue_client('v1');
     $query_parts = explode("&", $_SERVER['REQUEST_URI']);
     $last_query_param = end($query_parts);
+    $last_query_param_parts = explode("=", $last_query_param);
 
-    if ($last_query_param === 'page') {
-      $page_parts = explode("=", $$last_query_param);
-      $page_number = end($page_parts);
+    if ($last_query_param_parts[0] === 'page') {
+      $page_number = end($last_query_param_parts);
 
       if ($page_number) {
         $response = $client->getAllReportbacks([
--- a/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
@@ -157,7 +157,7 @@ function _reportback_item_resource_index($campaigns, $exclude, $status, $count, 
   );
 
   if (variable_get('rogue_collection', FALSE)) {
-    $client = dosomething_rogue_client('v1');
+    $client = dosomething_rogue_client();
 
     $response = $client->getAllReportbacks([
       'filter' => [
--- a/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
@@ -163,7 +163,7 @@ function _reportback_item_resource_index($campaigns, $exclude, $status, $count, 
         'campaign_id' => $campaigns,
       ],
       'as_user' => dosomething_user_get_field('field_northstar_id'),
-      'limit' => 6,
+      'limit' => 12,
     ]);
 
     return $response;
--- a/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
@@ -163,6 +163,7 @@ function _reportback_item_resource_index($campaigns, $exclude, $status, $count, 
         'campaign_id' => $campaigns,
       ],
       'as_user' => dosomething_user_get_field('field_northstar_id'),
+      'limit' => 6,
     ]);
 
     return $response;
--- a/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
+++ b/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
@@ -326,6 +326,7 @@ function dosomething_kudos_get_kudos_data_for_fid($fid = NULL, $total_rogue_kudo
   }
   $kudos = [];
   $kudos['fid'] = $fid;
+  // @TODO: update the below to grab from the Rogue response when we start sending reaction data to Rogue when a user newly reacts to a post.
   $kudos['existing_kids'] = dosomething_kudos_get_kudos_for_user_by_file($fid);
   $kudos['reaction_totals'] = [
     'heart' =>
@@ -333,7 +334,8 @@ function dosomething_kudos_get_kudos_data_for_fid($fid = NULL, $total_rogue_kudo
   ];
   $kudos['allowed_reactions'] = [dosomething_kudos_get_term_id_by_name('heart')];
 
-  $kudos['from_rogue'] = $liked_by_current_user ? $liked_by_current_user : FALSE;
+  // Use this in photo.tpl.php for the logged in user's reaction. If the user has reacted to the post, make the reaction button active.
+  $kudos['from_rogue'] = $liked_by_current_user;
 
   return $kudos;
 }
--- a/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
+++ b/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
@@ -297,6 +297,10 @@ function dosomething_kudos_term_is_selected($kudos, $term) {
     return false;
   }
 
+  if ($kudos['from_rogue']) {
+    return true;
+  }
+
   return true;
 }
 
@@ -312,14 +316,14 @@ function dosomething_kudos_disable_reactions($nid) {
 /**
  * Grab all Kudos data required to render the markup for a given reportback fid
  * @param  int $fid Reportback file id
- * @param  int $total_rogue_kudos Total kudos from Rogue response (If Rogue is powering the gallery)
+ * @param  int $total_rogue_kudos Total kudos from Rogue response (If Rogue is powering the gallery).
+ * @param  bool $liked_by_current_user Response from Rogue if the current user has reacted to the post or not.
  * @return array Values needed to render Kudos markup
  */
-function dosomething_kudos_get_kudos_data_for_fid($fid = NULL, $total_rogue_kudos = NULL) {
+function dosomething_kudos_get_kudos_data_for_fid($fid = NULL, $total_rogue_kudos = NULL, $liked_by_current_user = NULL) {
   if (! $fid) {
     return [];
   }
-
   $kudos = [];
   $kudos['fid'] = $fid;
   $kudos['existing_kids'] = dosomething_kudos_get_kudos_for_user_by_file($fid);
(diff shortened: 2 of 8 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -361,11 +361,28 @@ function paraneue_dosomething_get_themed_reportback_item($data, $rogue_rb_item =
...
+
+        $total_rogue_kudos = $data->kudos['total'];
+        $liked_by_current_user = $data->kudos['liked_by_current_user'] ? true : false;
+
+        if ($liked_by_current_user) {
+          drupal_add_js(
+            [
+              'dsKudosReactions' => [
+                'likedByCurrentUser' => $liked_by_current_user,
+              ]
```
chloealee: @DFurnes adding this setting to pass data to the front end if a user has liked the current reportback item. Do I need to add another level per item so it doesn't make this value be the same for each reportback item when it sends to the front end? Maybe this isn't the best way to do this...

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
```diff
@@ -308,7 +327,7 @@ const Reportback = {
         'isListItem': true,  // @TODO: probably a better way to address this... :insert shruggy emoji here:
         'reactions': {
           enabled: setting('dsKudosReactions.enabled'),
-          data: this.getReactions(items[i].kudos.data),
+          data: setting('dsKudosReactions.likedByCurrentUser') ? this.getReactions(items[i].kudos, true) : this.getReactions(items[i].kudos.data, false),
         }
```
chloealee: @DFurnes START HERE! if it is not coming from Rogue, each item has `kudos.data`. However, this is not how it is returned in the Rogue response. So I manipulated it here but again, I don't think this is the correct way to go about this? I didn't change the response in the Rogue data because it won't ever send multiple reactions, just one.

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
```diff
@@ -183,34 +183,53 @@ const Reportback = {
    * @param  {array} data  Array of Kudos objects from API.
    * @return {array}
    */
-  getReactions: function(data) {
-    let reactions = [];
-
-    forEach(this.reactions.approved, (term) => {
-      const termId = this.reactions.terms[term];
+  getReactions: function(data, reactionFromRogue) {
+    if (reactionFromRogue) {
```
chloealee: @DFurnes SECOND COMMENT! Since we're only passing one reaction here, I wrote some code that will give the same response but with Rogue data.

lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
```diff
@@ -158,7 +158,13 @@ function _reportback_item_resource_index($campaigns, $exclude, $status, $count,
...
     $client = dosomething_rogue_client('v1');
-    $response = $client->getAllReportbacks($inputs);
+    $response = $client->getAllReportbacks([
+      'filter' => [
+        'campaign_id' => $campaigns,
+      ],
+      'as_user' => dosomething_user_get_field('field_northstar_id'),
+    ]);
+
     return $response;
```
chloealee: @DFurnes this query works however, it is still returning the first 6 posts that we already have shown and gives all the remaining posts, not just the next 6. Not sure how to write to show only next 6?

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
chloealee: shoot now this is broken if rogue is turned off with new code :/ Here's the error: 
```
Uncaught TypeError: Cannot read property 'length' of undefined
    at eval (eval at <anonymous> (app.js:14415), <anonymous>:23:36)
    at Object.templatize (app.js:17455)
    at Object.<anonymous> (app.js:17415)
    at fire (lib.js:3282)
    at Object.fireWith [as resolveWith] (lib.js:3394)
    at done (lib.js:9448)
    at XMLHttpRequest.callback (lib.js:9852)
```

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
DFurnes: Ideally, we'd want to avoid doing any sort of conditional logic here and just return an "identical" response from Rogue. Is it possible to manipulate Rogue's reportback endpoint so it has `kudos.data` structure too?

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
```diff
@@ -183,34 +183,53 @@ const Reportback = {
...
       reactions.push(reaction);
-    });
+    } else {
+      let reactions = [];
+
+      forEach(this.reactions.approved, (term) => {
+        const termId = this.reactions.terms[term];
+
+        let reaction = {
+          kudosId: null,
```
DFurnes: I'd recommend trying to do most of this work (formatting the data into `kudosId`, `term`, `termId`, `total` and `userReacted`) within Rogue, so you don't have to then transform it again within the Phoenix client-side code. That way Phoenix doesn't have to be aware of any of the implementation details of the new setup.

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
chloealee: @DFurnes I was thinking that but then I _think_ we'd run into trouble [here](https://github.com/DoSomething/phoenix/blob/dev/lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js#L189), since we don't have the Kudos object?

chloealee: ohh you mean instead of returning a boolean for `liked_by_current_user`, return the action `reaction` object in the Rogue response?

DFurnes: Yeah! Wherever possible, I think it'll be easier to try to make the Rogue endpoint match the old Phoenix endpoint's structure as close as we can.

chloealee: ah got it! I started to go down that route but then noticed that there isn't any `kudos` data in the old Phoenix endpoint's structure. But I can just add that info. that we need to try to match the code in Phoenix ashes for what it wants!

lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
```diff
@@ -158,7 +158,33 @@ function _reportback_item_resource_index($campaigns, $exclude, $status, $count,
 
   if (variable_get('rogue_collection', FALSE)) {
     $client = dosomething_rogue_client('v1');
-    $response = $client->getAllReportbacks($inputs);
+    $query_parts = explode("&", $_SERVER['REQUEST_URI']);
```
chloealee: @DFurnes I'm trying to get the page number that is set in `Reportback.js` on line 290 of this PR. Is this the best way to do it? I'm having a bit of trouble grabbing the page number. If it's easier to walk you through IRL what I was thinking, LMK!

chloealee: @DFurnes oh wait I found my mistake - got it nearly there!

lib/themes/dosomething/paraneue_dosomething/js/campaign/prototype.js
```diff
@@ -25,6 +27,17 @@ function makeKudosRequest($el, id)
 }
 
 /**
+ * Make an API request to Rogue to create/delete the Reaction for that item.
+ *
+ * @param string User's northstar ID.
+ */
+ function makeKudosRequestToRogue($northstarId, $postId)
```
chloealee: this is a placeholder for the next chunk of work to send new reactions to Rogue. Is it ok to keep here?

lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
```diff
@@ -158,7 +158,58 @@ function _reportback_item_resource_index($campaigns, $exclude, $status, $count,
 
   if (variable_get('rogue_collection', FALSE)) {
     $client = dosomething_rogue_client('v1');
-    $response = $client->getAllReportbacks($inputs);
+
+    // Get the query params to see if we are populating client or server side gallery portion.
```
DFurnes: Phoenix's server-rendered galleries can call `$client->getAllReportbacks` wherever necessary in the codebase, rather than making a HTTP request to it's own API endpoints.

lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
```diff
@@ -158,7 +158,58 @@ function _reportback_item_resource_index($campaigns, $exclude, $status, $count,
...
+    // If page param is present, we are populating "view more" section of gallery (client side).
+    $query_parts = explode("&", $_SERVER['REQUEST_URI']);
+    $query_params = [];
```
DFurnes: You should be able to use the [same `$parameters`](https://github.com/DoSomething/phoenix/blob/665e49bf4097be42ac74cd299e25e3c8a5fcee01/lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc#L148-L157) here as we use for the "legacy" response to this endpoint.

lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
```diff
@@ -158,7 +158,58 @@ function _reportback_item_resource_index($campaigns, $exclude, $status, $count,
...
+    if ($last_query_param_parts[0] === 'page') {
+      $page_number = end($last_query_param_parts);
+
+      if ($page_number) {
+        $response = $client->getAllReportbacks([
+          'filter' => [
+            'campaign_id' => $campaigns,
+            'exclude' => $exclude,
+          ],
+          'as_user' => dosomething_user_get_field('field_northstar_id'),
```
DFurnes: For the API endpoint, we want `as_user` to be something specified in the query string of the request (since a lot of our API requests actually come in with the `ds-api@dosomething.org` user and we then specify an override user ID with this option).

lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
```diff
@@ -158,7 +158,58 @@ function _reportback_item_resource_index($campaigns, $exclude, $status, $count,
...
+          'limit' => 12,
+          'page' => $page_number,
+        ]);
+      }
+    } else {
```
DFurnes: The services-defined `$page` variable above should already default to `1` if omitted, so you shouldn't need to do any special logic here!

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -311,21 +316,24 @@ function dosomething_kudos_disable_reactions($nid) {
...
  */
-function dosomething_kudos_get_kudos_data_for_fid($fid = NULL) {
+function dosomething_kudos_get_kudos_data_for_fid($fid = NULL, $total_rogue_kudos = NULL, $liked_by_current_user = NULL) {
   if (! $fid) {
     return [];
   }
-
   $kudos = [];
   $kudos['fid'] = $fid;
+  $kudos['existing_kids'] = dosomething_kudos_get_kudos_for_user_by_file($fid);
```
DFurnes: This needs to come from the Rogue response as well, right?

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -311,21 +316,24 @@ function dosomething_kudos_disable_reactions($nid) {
...
   $kudos['reaction_totals'] = [
-    'heart' => dosomething_kudos_get_total_for_file_by_term_name($fid, 'heart'),
+    'heart' =>
+        $total_rogue_kudos ? $total_rogue_kudos : dosomething_kudos_get_total_for_file_by_term_name($fid, 'heart')
   ];
-  $kudos['existing_kids'] = dosomething_kudos_get_kudos_for_user_by_file($fid);
-  $kudos['reaction_totals'] = ['heart' => dosomething_kudos_get_total_for_file_by_term_name($fid, 'heart')];
   $kudos['allowed_reactions'] = [dosomething_kudos_get_term_id_by_name('heart')];
 
+  $kudos['from_rogue'] = $liked_by_current_user ? $liked_by_current_user : FALSE;
```
DFurnes: Can this line be removed?

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -361,11 +361,28 @@ function paraneue_dosomething_get_themed_reportback_item($data, $rogue_rb_item =
...
+  if ($data->id && !dosomething_kudos_disable_reactions($data->campaign->id)) {
+      if ($rogue_rb_item) {
+        // Find the fid from the rogue reference table
+        // $data->id = db_query("SELECT rogue_rbs.fid FROM {dosomething_rogue_reportbacks} rogue_rbs WHERE rogue_event_id = :rogue_event_id", array(':rogue_event_id' => $data->event_id))->fetchField();
+
+        $total_rogue_kudos = $data->kudos['total'];
+        $liked_by_current_user = $data->kudos['data']['current_user']['reacted'] ? true : false;
+
+        if ($liked_by_current_user) {
+          drupal_add_js(
```
DFurnes: I think this `drupal_add_js` can be removed now, right?

lib/themes/dosomething/paraneue_dosomething/js/campaign/prototype.js
DFurnes: Let's use the same Phoenix reactions endpoint for now so the front-end doesn't need to change!

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
```diff
@@ -273,7 +278,28 @@ const Reportback = {
         }
       },
     }).done(function(response) {
-      _this.apiUrl = response.meta.pagination.links.next_uri || null;
+      // If response.meta.pagination.links.next, this is from Rogue.
+      if (response.meta.pagination.links.next) {
+        var splitResponseUrl = response.meta.pagination.links.next.split("&");
```
DFurnes: What's happening in this code block? Is there something wrong with the "next" link returned from Rogue?

lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
chloealee: I'm not quiet sure what you mean here?

lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
chloealee: omg I am a dummy! I swear I tried to do that and it didn't work but just tried again and it did. So much better!!

lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
chloealee: Not sure if I'm quite understanding what you're saying here either! Isn't this in the query string of the request?

lib/modules/dosomething/dosomething_api/resources/reportback_item_resource.inc
chloealee: ah !

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
chloealee: @DFurnes yes, I haven't done this logic yet! I was going to tackle this with a new PR that will be responsible for sending reaction data to Rogue for when a user newly reacts to a post. Does that make sense to do?

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
chloealee: We need this for the server side reactions for the first 6 posts in the gallery! This is involved in the logic to have the button be active if the current user has liked the post.

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
chloealee: yes! Thanks for catching this!

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
chloealee: So in Rogue, the pagination is structured as 
```
"meta": {
    "pagination": {
      "total": 24,
      "count": 20,
      "per_page": 20,
      "current_page": 1,
      "total_pages": 2,
      "links": {
        "next": "http://rogue.dev:8000/api/v1/reportbacks?page=2"
      }
```
but in Phoenix, it is 
```
"meta": {
    "pagination": {
      "total": 24,
      "count": 20,
      "per_page": 20,
      "current_page": 1,
      "total_pages": 2,
      "links": {
        "next_uri": "http://rogue.dev:8000/api/v1/reportbacks?page=2"
      }
```
I don't think we can change this since we're using `IlluminatePaginatorAdapter`?

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
DFurnes: That makes good sense to me! In the meantime, it might be nice to add a little `@TODO` comment.

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
DFurnes: Oh, I see now. Can you add a comment with more context for where this gets used so it's a bit more clear when first reading through? It also seems like this ternary may be unnecessary.

lib/themes/dosomething/paraneue_dosomething/js/campaign/prototype.js
```diff
@@ -56,19 +58,23 @@ $(document).on('click', '.js-kudos-button', function(event) {
   const dataKid = $el.attr('data-kid');
   const method = dataKid ? 'DELETE' : 'POST';
 
-  makeKudosRequest($el, dataKid)
-    .done((response) => {
-      const kid = get(response, '0.kid', '');
+  // if (setting('dsKudosReactions.sendToRogue')) {
+  //   makeKudosRequestToRogue(setting('dsKudosReactions.northstarId'), setting('dsKudosReactions.postId'))
+  // } else {
+    makeKudosRequest($el, dataKid)
```
DFurnes: It looks like all these lines got accidentally indented.

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
DFurnes: Ahh gotcha. We could update that in Rogue by extending `IlluminatePaginatorAdaptor` and changing the little bit that we need, but we'd also need to find a way to pass along the right base URL so the request continues to go through Phoenix as we paginate... could be nice to look into if you want.

chloealee: Ooh that makes sense! Totally down to try that but just thinking out loud here - is there any benefit to having just `next` over `next_uri` (is that a good standard that apps will you across the board?). Wondering because I know we chatted about not worrying about how code is written since we'll trash Phoenix-ashes soon and wondering if it's better code for Rogue to have `next` instead of `next_uri`?

DFurnes: Yeah, if we change it in Rogue let's definitely only change it for this one endpoint!

DFurnes (approved):

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
chloealee: oh okay that makes sense! I'm having a difficult time seeing where this would be changed in Rogue. Does it make sense to create a new file `PaginatorForPhoenixAshesGallery` that extends `IlluminatePaginatorAdapter` and call that [here](https://github.com/DoSomething/rogue/blame/master/app/Http/Controllers/Api/ApiController.php#L119)? Then in `PaginatorForPhoenixAshesGallery` rewrite `paginator` function to add `next_uri` instead of `next`?

DFurnes: Yes! 👍

chloealee: Ah very cool! Thanks for walking me through that, Dave!

chloealee: @DFurnes It actually seems like this happens in the way deeper in the files in [Fractal's `ArraySerializer.php`](https://github.com/thephpleague/fractal/blob/master/src/Serializer/ArraySerializer.php#L112). I'm not sure how now how to create a custom file to extend this since it's way deeper than `IlluminatePaginatorAdaptor`?

chloealee: oh wait I think I am close - disregard this and I'll holler if I get stuck again! Sorry!

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -361,11 +361,27 @@ function paraneue_dosomething_get_themed_reportback_item($data, $rogue_rb_item =
...
+        // Let the front end know that Rogue is powering the gallery.
+        drupal_add_js(
+          [
+            'dsRogue' => [
+              'poweringGallery' => true,
+            ]
+          ],
+          'setting'
+        );
+      }
```
chloealee, line 381: @DFurnes had to pass this back in because we have still have to be able to tell if Rogue is powering the gallery on the front end! If it is, we need to append the page number from the response to the proxy to `/reportbacks` to then make the correct Request to Rogue based on the pagination.

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
```diff
@@ -273,7 +278,33 @@ const Reportback = {
         }
       },
     }).done(function(response) {
-      _this.apiUrl = response.meta.pagination.links.next_uri || null;
+      // Check to see if the response if from Rogue.
+      if (Drupal.settings.dsRogue.poweringGallery) {
```
chloealee, line 282: @DFurnes still need this block of code again to pull out page number to proxy via Phoenix's `/reportbacks` endpoint and then send the page number in the request to Rogue.

chloealee:
@DFurnes I know that you already approved this but made changes based on your comment to update Rogue! Would you mind taking another look? 

The Rogue PR will need to be merged first in order for this PR to work: https://github.com/DoSomething/rogue/pull/230

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
DFurnes: Can you make sure this is only adding one `dsRogue.poweringGallery` onto the page? I'm worried that this method might be getting called more than once and injecting a ton of script tags.

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
```diff
@@ -183,34 +182,40 @@ const Reportback = {
...
-      const termId = this.reactions.terms[term];
-
-      let reaction = {
-        kudosId: null,
-        term: term,
-        termId: termId,
-        total: 0,
-        userReacted: false,
-      };
+    getReactions: function(data) {
```
DFurnes: Oh sorry, I also realized this looks like it got accidentally indented too! Can you fix this before we merge?

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
DFurnes: Hmm, yeah I'm curious if we can pass along the proper base URL so Rogue can add that to the original request (maybe via a custom header?) but this is definitely good to merge in the meantime!

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
chloealee: ah yes! good eyes!

lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
chloealee: @DFurnes ah I see! I did this in the console and I _think_ this means it is only adding one onto the page? 
screen shot 2017-04-21 at 3 14 55 pm

DFurnes (approved):

PR #7360, opened 2017-04-18
gets rid of Rogue API version and hard codes in url

### LINKED ISSUE #7345: Add hardcoded versioning when making requests to Rogue API
_BUG_
-------------------
#### Current Behavior
- Currently, we are hardcoding the API version in the [`dosomething_rogue_client`](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module#L42) function. 
- As a developer, I don't want the API versions to be retrieved from Rogue feature flag when we build the client to make requests to Rogue.

#### Desired Behavior
- Since we're hitting endpoints in Rogue in both `v1` and `v2`, we don't want to grab the version from Rogue settings when we build the client. 

#### Why This Matters
- So we can use `dosomething_rogue_client` whenever we are making calls to Rogue. 

Refs https://github.com/DoSomething/phoenix/pull/7343#discussion_r109768985

chloealee:
#### What's this PR do?
- Gets rid of `Rogue API version number` field. 
- Hardcodes version in URL. 

👀 

#### Relevant tickets
Fixes #7345 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.admin.inc
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.admin.inc
@@ -23,13 +23,6 @@ function dosomething_rogue_config_form($form, &$form_state) {
     '#default_value' => variable_get('dosomething_rogue_url', 'https://rogue.dosomething.org/api'),
   ];
 
-  $form['rogue']['dosomething_rogue_api_version'] = [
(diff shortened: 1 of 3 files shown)


DFurnes (approved):
1️⃣ ✅

PR #7362, opened 2017-04-26
Add endpoint to trigger transactional message.

DFurnes:
#### What's this PR do?
This pull request adds a `/api/v1/transactionals` endpoint to send a "welcome" email to a newly registered user and sign them up for the appropriate MailChimp and MobileCommons lists.

I tested that the right parameters get filled briefly on my local, but planning on doing more extensive testing on this on Thor since Message Broker is configured there.

This was added as a new endpoint (rather than part of `/api/v1/users`) because it simplifies [implementation in Northstar](https://github.com/DoSomething/northstar/pull/565) (no need to pass a parameter down through the Registrar to the Phoenix API client), and should also be easier to tear out when we want this to live in Blink/Customer.io instead.

:v:

#### Relevant tickets
[Trello Card](https://trello.com/c/rPFyZcsu/571-5-as-a-developer-i-want-northstar-to-tell-phoenix-ashes-to-send-transactional-messaging-to-new-users)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/transactional_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/transactional_resource.inc
@@ -1,0 +1,76 @@
+<?php
+
+function _transactional_resource_definition() {
+  return [
+    'transactionals' => [
+      'operations' => [
+        'create' => [
+          'help' => 'Create a transactional message.',
+          'file' => [
+            'type' => 'inc',
+            'module' => 'dosomething_api',
+            'name' => 'resources/transactional_resource',
+          ],
+          'callback' => '_transactional_resource_create',
+          'args' => [
+            [
+              'name' => 'id',
+              'type' => 'string',
+              'description' => 'The user\'s Northstar ID.',
+              'source' => 'data',
+              'optional' => FALSE,
+            ],
+            [
+              'name' => 'template',
+              'type' => 'string',
+              'description' => 'The type of transactional to send.',
+              'source' => 'data',
+              'optional' => FALSE,
+            ],
+          ],
+          'access callback' => '_transactional_resource_access',
+          'access arguments' => ['create'],
+        ],
+      ],
+    ]
+  ];
+}
+
+/**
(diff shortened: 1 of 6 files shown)


DFurnes:
[(5) As a developer, I want Northstar to tell Phoenix-Ashes to send transactional messaging to new users.](https://trello.com/c/rPFyZcsu/571-5-as-a-developer-i-want-northstar-to-tell-phoenix-ashes-to-send-transactional-messaging-to-new-users)

itsjoekent (approved):
Seems fine to me!

lib/modules/dosomething/dosomething_api/resources/transactional_resource.inc
```diff
@@ -0,0 +1,76 @@
...
+ *   Array passed to the endpoint. Possible keys:
+ *   - $id (string). The user's Northstar ID
+ *   - $template (string). The type of transactional to send.
+ *
+ * @return mixed
+ */
+function _transactional_resource_create($request) {
+  global $user;
+
+  $account = user_load(dosomething_user_convert_to_legacy_id($request['id']));
```
itsjoekent, line 61: Actually wait, didn't we have a load user by northstar ID function?

itsjoekent (approved):

lib/modules/dosomething/dosomething_api/resources/transactional_resource.inc
DFurnes: Oh yep, would ya look at that! This is what we've used in the other API endpoints, and is nice since it supports both Northstar IDs and Drupal IDs... it could be nice to combine the two, but I'm always wary of refactoring in Drupal-land.

itsjoekent: Yeah at this point I wouldn't bother...

DFurnes: Then it's decided! 😬

PR #7366, opened 2017-04-27
Send reactions to rogue

### LINKED ISSUE #7365: Send gallery reactions to Rogue
_FEATURE OVERVIEW_
-------------------
#### User Story
As a developer, if a user reacts to a post in Phoenix-ashes, I want to send a new reaction to Rogue so that Rogue knows this has happened.

Final step to having Rogue power the Phoenix-ashes gallery! 

#### Why This Matters
So Rogue can completely power the Phoenix-ashes gallery!

chloealee:
#### What's this PR do?
When a user reacts to a post, Phoenix sends the request to Rogue's `/reactions` endpoint to log in the database. 

- When logged in, react to a post you haven't reacted to before. 
  - This reaction should be added in the Rogue `reactions` table. 
  - The reaction button should be active and the counter should be incremented.
- "Un-react" to the same post. 
  - The reaction should be soft deleted in the Rogue `reactions` table. 
  The reaction button should not be active and the counter should decrease by one. 

Final step for Rogue to power the Phoenix-ashes gallery! 

#### Relevant tickets
Fixes #7365

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
@@ -189,13 +189,15 @@ function _kudos_resource_index($reportbackitems, $count) {
  */
 function _kudos_resource_create($reportback_item_id, $term_ids, $northstar_id) {
   if (variable_get('rogue_collection', FALSE)) {
+    global $user;
+    $northstar_user = dosomething_northstar_get_user($user->uid, 'drupal_id');
+
     $data = [
-      'northstar_id' => $northstar_id,
+      'northstar_id' => $northstar_user->id,
       'post_id' => $reportback_item_id,
     ];
 
-    $client = dosomething_rogue_client();
-    $response = $client->postReaction($data);
+    $response = dosomething_rogue_client()->postReaction($data);
 
     return [
       [
@@ -213,7 +215,6 @@ function _kudos_resource_create($reportback_item_id, $term_ids, $northstar_id) {
   return (new KudosTransformer)->create($parameters);
 }
 
-
 /**
  * Delete the specified kudos record.
  *
@@ -222,23 +223,21 @@ function _kudos_resource_create($reportback_item_id, $term_ids, $northstar_id) {
  * @return string
  */
 function _kudos_resource_delete($kudos_id) {
-  // if (variable_get('rogue_collection', FALSE)) {
-  //   // @TODO: we need to get the below values
-  //   $data = [
-  //     'northstar_id' => $northstar_id,
-  //     'post_id' => $reportback_item_id,
-  //   ];
+  if (variable_get('rogue_collection', FALSE)) {
+    global $user;
+    $northstar_user = dosomething_northstar_get_user($user->uid, 'drupal_id');
 
-  //   $response = dosomething_rogue_client()->postReaction($data);
+    $data = [
+      'northstar_id' => $northstar_user->id,
+      'post_id' => $kudos_id,
+    ];
 
-  //   if ($response['meta']['action'] === 'unliked') {
-  //     return [
-  //       [
-  //         'kid' => 0,
-  //       ]
-  //     ];
-  //   }
-  // }
+    $response = dosomething_rogue_client()->postReaction($data);
+
+    if ($response['meta']['action'] === 'unliked') {
+      return;
+    }
+  }
 
   return (new KudosTransformer)->delete($kudos_id);
 }
--- a/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
@@ -180,7 +180,6 @@ function _kudos_resource_index($reportbackitems, $count) {
   return (new KudosTransformer)->index($parameters);
 }
 
-
 /**
  * Create a new kudos record.
  *
@@ -194,6 +193,22 @@ function _kudos_resource_create($reportback_item_id, $term_ids, $northstar_id) {
     'tids' => $term_ids,
     'northstar_id' => $northstar_id,
   ];
+
+  if (variable_get('rogue_collection', FALSE)) {
+    $data = [
+      'northstar_id' => $northstar_id,
+      'post_id' => $reportback_item_id,
+    ];
+    $response = dosomething_rogue_client()->postReaction($data);
+    print_r($response);
+    die();
+
+    return [
+      [
+        'kid' => $response['post_id'],
+      ]
+    ];
+  }
 
   return (new KudosTransformer)->create($parameters);
 }
--- a/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
@@ -211,9 +211,6 @@ function _kudos_resource_create($reportback_item_id, $term_ids, $northstar_id) {
     }
     else {
       return;
-      // return [
-      //     'kid' => 'unliked',
-      // ];
     }
   }
 
--- a/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
@@ -189,8 +189,12 @@ function _kudos_resource_index($reportbackitems, $count) {
  */
 function _kudos_resource_create($reportback_item_id, $term_ids, $northstar_id) {
   if (variable_get('rogue_collection', FALSE)) {
-    global $user;
-    $northstar_user = dosomething_northstar_get_user($user->uid, 'drupal_id');
+    if (isset($northstar_id)) {
+      $northstar_user->id = $northstar_id;
+    } else {
+      global $user;
+      $northstar_user = dosomething_northstar_get_user($user->uid, 'drupal_id');
+    }
 
     $data = [
       'northstar_id' => $northstar_user->id,
--- a/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
@@ -188,27 +188,27 @@ function _kudos_resource_index($reportbackitems, $count) {
  * @param $term_ids
  */
 function _kudos_resource_create($reportback_item_id, $term_ids, $northstar_id) {
-  $parameters = [
-    'fid' => $reportback_item_id,
-    'tids' => $term_ids,
-    'northstar_id' => $northstar_id,
-  ];
-
   if (variable_get('rogue_collection', FALSE)) {
     $data = [
       'northstar_id' => $northstar_id,
       'post_id' => $reportback_item_id,
     ];
-    $response = dosomething_rogue_client()->postReaction($data);
-    print_r($response);
-    die();
+
+    $client = dosomething_rogue_client();
+    $response = $client->postReaction($data);
 
     return [
       [
-        'kid' => $response['post_id'],
+        'kid' => $response['data']['post_id'],
       ]
     ];
   }
+
+  $parameters = [
+    'fid' => $reportback_item_id,
+    'tids' => $term_ids,
+    'northstar_id' => $northstar_id,
+  ];
 
   return (new KudosTransformer)->create($parameters);
 }
@@ -222,5 +222,23 @@ function _kudos_resource_create($reportback_item_id, $term_ids, $northstar_id) {
  * @return string
  */
 function _kudos_resource_delete($kudos_id) {
+  // if (variable_get('rogue_collection', FALSE)) {
+  //   // @TODO: we need to get the below values
+  //   $data = [
+  //     'northstar_id' => $northstar_id,
+  //     'post_id' => $reportback_item_id,
+  //   ];
+
+  //   $response = dosomething_rogue_client()->postReaction($data);
+
+  //   if ($response['meta']['action'] === 'unliked') {
+  //     return [
+  //       [
+  //         'kid' => 0,
+  //       ]
+  //     ];
+  //   }
+  // }
+
   return (new KudosTransformer)->delete($kudos_id);
 }
--- a/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
@@ -189,15 +189,13 @@ function _kudos_resource_index($reportbackitems, $count) {
  */
 function _kudos_resource_create($reportback_item_id, $term_ids, $northstar_id) {
   if (variable_get('rogue_collection', FALSE)) {
-    if (isset($northstar_id)) {
-      $northstar_user->id = $northstar_id;
-    } else {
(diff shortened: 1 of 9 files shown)


lib/themes/dosomething/paraneue_dosomething/js/campaign/prototype.js
```diff
@@ -58,23 +57,16 @@ $(document).on('click', '.js-kudos-button', function(event) {
...
     .done((response) => {
       const kid = get(response, '0.kid', '');
-
-      if (method === 'POST' && !kid) {
-        return;
-      }
-
+      // if (method === 'POST' && !kid) {
+      //   return;
+      // }
```
chloealee: Do we need this block here? I had to comment this out to get the logic correct in this case: 
- Current user has already reacted to a post (reaction button is active). 
- When the current user clicks on the reaction again, the counter increments and the button remains active (looks like it adds another reaction when the user has already reacted instead of undoing the reaction).
- Everything is accurately reflected in the database, it's just the front end that isn't displaying correctly. 
- When you refresh the page, front end is updated and is accurate. 

Why this is happening? 
- I believe it's because the method is `POST` since on a new page, it doesn't know about a `kid` from Rogue. Really, the method should be `DELETE` however, the logic still works since the `/reactions` endpoint in Rogue both creates and soft deletes. 
- Since it deletes, it doesn't send back a `kid` so when this block is not commented out, it just returns before updating the Kudos Interface.

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
```diff
@@ -189,6 +188,29 @@ function _kudos_resource_index($reportbackitems, $count) {
...
+    $response = dosomething_rogue_client()->postReaction($data);
+
+    if ($response['meta']['action'] === 'liked') {
+      return [
+        [
+          'kid' => $response['data']['post_id'],
+        ]
+      ];
+    }
+    else {
```
chloealee, line 212: If this actually soft deletes or `un-likes` a post, don't return the `kid` so front end knows what to correctly display.

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
```diff
@@ -207,5 +228,21 @@ function _kudos_resource_create($reportback_item_id, $term_ids, $northstar_id) {
...
+  if (variable_get('rogue_collection', FALSE)) {
+    global $user;
+    $northstar_user = dosomething_northstar_get_user($user->uid, 'drupal_id');
+
+    $data = [
+      'northstar_id' => $northstar_user->id,
+      'post_id' => $kudos_id,
+    ];
+
+    $response = dosomething_rogue_client()->postReaction($data);
```
chloealee, line 244: do we need to write logic to handle errors if a request doesn't make it to Rogue?

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
```diff
@@ -189,6 +188,29 @@ function _kudos_resource_index($reportbackitems, $count) {
...
+  if (variable_get('rogue_collection', FALSE)) {
+    global $user;
+    $northstar_user = dosomething_northstar_get_user($user->uid, 'drupal_id');
+
+    $data = [
+      'northstar_id' => $northstar_user->id,
+      'post_id' => $reportback_item_id,
+    ];
+
+    $response = dosomething_rogue_client()->postReaction($data);
```
chloealee: same here - do we need to handle errors?

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
```diff
@@ -189,6 +188,29 @@ function _kudos_resource_index($reportbackitems, $count) {
  * @param $term_ids
  */
 function _kudos_resource_create($reportback_item_id, $term_ids, $northstar_id) {
+  if (variable_get('rogue_collection', FALSE)) {
+    global $user;
+    $northstar_user = dosomething_northstar_get_user($user->uid, 'drupal_id');
```
DFurnes: We'll want "admin" accounts (like `ds-api@dosomething.org`) to be able to set an override ID using that `$northstar_id` parameter above. That's how we allow Phoenix Next to create reactions on a user's behalf.

DFurnes: Something like [this](https://github.com/DoSomething/phoenix/blob/ca42f51d11d54f86fc6244f68b94a693247fca98/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc#L270-L276).

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
DFurnes: I think it's okay letting this 500 if there are any problems – we can always revisit and add better error handling then.

lib/themes/dosomething/paraneue_dosomething/js/campaign/prototype.js
DFurnes: Can you try logging out what the `kid` variable is when this happens? I think this should work without changes on the front end, unless perhaps we're not providing that response the way it's expected.

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
DFurnes: I think it's safe to leave this as is for now!

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
chloealee: ok!

lib/themes/dosomething/paraneue_dosomething/js/campaign/prototype.js
chloealee: It thinks it's a `POST` request (since there is no `kid`) in line 68 and when I `console.log(kid);` after line 61, `kid` is nothing so then it just returns without updating the interface. LMK if you want me to show you IRL!

DFurnes: Right! The `POST` endpoint [return a `kid`](https://github.com/DoSomething/phoenix/blob/b43bc156041a59dcb08894df4132ad93065be3e1/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc#L203-L207) if it's successful (which we're using the `post_id` as a stand-in for). The idea behind this conditional is that if we're posting a new kudo but a value isn't returned in `0.kid` then we assume an error.

Can you try logging out the `response` variable above? Maybe it's not structured how we expect it to be here?

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
chloealee: ooo very cool! Updated :)

lib/themes/dosomething/paraneue_dosomething/js/campaign/prototype.js
chloealee: ahh I see - thanks for explaining more! The `response` comes back as `null` - instead of `return` [here](https://github.com/DoSomething/phoenix/blob/b43bc156041a59dcb08894df4132ad93065be3e1/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc#L210), should I set the `kid` to something like `0`?

chloealee: wait that doesn't quite work actually...

chloealee: The problem with this conditional is that if the user has liked this already and tries to unlike it, since it doesn't know about the `kid` in [this line](https://github.com/DoSomething/phoenix/blob/dev/lib/themes/dosomething/paraneue_dosomething/js/campaign/prototype.js#L58), it thinks it's a `POST` request instead of a `DELETE`. (sorry if I'm not understanding something and just repeating myself at this point!)

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
DFurnes: This still doesn't let us specify an arbitrary user ID. Ideally we want to use the current "session" when this API is hit via the web on legacy campaigns, but allow the `ds-api@dosomething.org` admin user to specify a custom ID (via a string passed to that `$northstar_id` parameter)  for other internal API requests.

Let's talk tomorrow and we can figure it out! 😄

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
DFurnes: Oh right! Good catch! 👍

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
```diff
@@ -207,5 +228,21 @@ function _kudos_resource_create($reportback_item_id, $term_ids, $northstar_id) {
...
+
+    $data = [
+      'northstar_id' => $northstar_user->id,
+      'post_id' => $kudos_id,
+    ];
+
+    $response = dosomething_rogue_client()->postReaction($data);
+
+    if ($response['meta']['action'] === 'unliked') {
+      return;
```
DFurnes, line 247: Same as above, we should figure out what sort of response to send if the action isn't `unliked` here.

lib/themes/dosomething/paraneue_dosomething/js/campaign/prototype.js
DFurnes: Now that you've added the "deleted" response above, is it safe to re-enable this conditional?

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
chloealee: ah sorry @DFurnes - it looks like my update didn't get pushed up! Is this ok now?

lib/themes/dosomething/paraneue_dosomething/js/campaign/prototype.js
chloealee: @DFurnes sorry a little lost! Where is the "deleted" response you're referring to?

DFurnes: Thinking of [this line](https://github.com/DoSomething/phoenix/pull/7366/files#diff-7a7e10f73c023858ee16feab2fbbc093R214) where we return a response without a `kid` in it!

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
```diff
@@ -189,6 +188,33 @@ function _kudos_resource_index($reportbackitems, $count) {
  * @param $term_ids
  */
 function _kudos_resource_create($reportback_item_id, $term_ids, $northstar_id) {
+  if (variable_get('rogue_collection', FALSE)) {
+    if (isset($northstar_id)) {
+      $northstar_user->id = $northstar_id;
```
DFurnes: Hmm, is `$northstar_user` defined in this case?

chloealee: ah great call! Updated - does this make more sense?

lib/themes/dosomething/paraneue_dosomething/js/campaign/untitled
```diff
@@ -0,0 +1,7 @@
+/*
+Theme Name: Azalea Child
```
DFurnes: :hibiscus: 👧

lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
```diff
@@ -189,15 +189,13 @@ function _kudos_resource_index($reportbackitems, $count) {
  */
 function _kudos_resource_create($reportback_item_id, $term_ids, $northstar_id) {
   if (variable_get('rogue_collection', FALSE)) {
-    if (isset($northstar_id)) {
-      $northstar_user->id = $northstar_id;
-    } else {
+    if (!isset($northstar_id)) {
```
DFurnes: Can we set `$northstar_id` within this conditional so we don't have to do the ternary below?

lib/themes/dosomething/paraneue_dosomething/js/campaign/prototype.js
chloealee: It's still not I don't think :-/ Sorry for all the back and forth on this one! I can explain IRL maybe if you have a sec?

lib/themes/dosomething/paraneue_dosomething/js/campaign/prototype.js
```diff
@@ -58,23 +57,18 @@ $(document).on('click', '.js-kudos-button', function(event) {
...
 
-  // if (setting('dsKudosReactions.sendToRogue')) {
-  //   makeKudosRequestToRogue(setting('dsKudosReactions.northstarId'), setting('dsKudosReactions.postId'))
-  // } else {
   makeKudosRequest($el, dataKid)
     .done((response) => {
       const kid = get(response, '0.kid', '');
 
-      if (method === 'POST' && !kid) {
+      if (method === 'POST' && (!kid || kid === 'unliked')) {
```
chloealee: @DFurnes actually found a hacky way to add this conditional back in! Instead of just doing a `return`, set the `kid` to `unliked` and check that here. Does this work?

lib/themes/dosomething/paraneue_dosomething/js/campaign/untitled
chloealee: oh shoot! Will delete that !

lib/themes/dosomething/paraneue_dosomething/js/campaign/prototype.js
DFurnes: Hmm, if we just `return` up above, then `kid` should be falsy I think. Can we try stepping through this IRL?

chloealee: yes would love that! I am free all day until 3 - swing by whenever works for you!

lib/themes/dosomething/paraneue_dosomething/js/reportback/Reportback.js
```diff
@@ -180,9 +180,10 @@ const Reportback = {
    * Get Kudos Reactions for a Reportback item.
    *
    * @param  {array} data  Array of Kudos objects from API.
+   * @param  {int} kudosId if from Rogue.
    * @return {array}
    */
-  getReactions: function(data) {
+  getReactions: function(data, kudosId = null) {
```
chloealee, line 186: @DFurnes had to update this a bit for the same logic we worked on together this am! When a user has already liked a post, the front end needs to know about the `kudosId` so passing it here. This fixes the logic to make sure if they `unlike` it, it makes a `DELETE` request instead of a `POST` now knowing it's `kudosId`

DFurnes (approved):
This looks good! Ship it!! ⛵️ 🎉

PR #7368, opened 2017-05-01
Keeper upper script

### LINKED ISSUE #7367: Keeper Upper script for Rogue
_FEATURE OVERVIEW_
-------------------
#### User Story
We want a keeper upper script so that we can keep Rogue up to date with the latest and greatest user activity.

#### Why This Matters
So Rogue can be a source of truth!

Nothing!

katiecrane:
#### What's this PR do?
This is the keeper upper script. This script will be run regularly to keep Rogue up to date with the latest Phoenix data. There are 3 types of updates sent:
1. New signup
2. New reportback/signup update
    - at least one new post
    - updating the `quantity` on the signup
    - updating the `why_participated` on the signup
3. Quantity or why update
    - In Phoenix, this would be an update on just the reportback without adding a new image
    - In Rogue, this would be a signup update

There are also three parts to the migration that do those 3 things in that order. We can see if signups or posts are in Rogue by using the reference tables, but for type 3 we have to check on timestamps. This is done by logging the timestamp of the most recent item (in this category) we sent. This is the timestamp on the item, not when we sent it. Then we use that information to see if there are any new quantity or why updates that we need to send over.

I suggest looking through each piece of the script as its own unit because they don't really interact. 

Do it for the data 📈 

#### Relevant tickets
Fixes #7367

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/scripts/rogue-keeper-upper.php
+++ b/scripts/rogue-keeper-upper.php
@@ -1,0 +1,302 @@
+<?php
+/**
+ * Keeper-Upper Script to export Signups, Reportbacks, and whatever else from drupal into Rogue.
+ * This is to be run on a recurring basis after the big migration has completed. 
+ *
+ * to run:
+ * drush --script-path=../scripts/ php-script rogue-keeper-upper.php
+ */
+
+// Pick up from where we left off, if we want to
+$last_saved = variable_get('dosomething_rogue_last_signup_kept_up', NULL);
+
+// Get ALL the unmigrated signups first
+if ($last_saved) {
+  $signups = db_query("SELECT signups.sid, signups.uid, signups.nid, signups.run_nid, signups.source, signups.timestamp, rb.quantity, rb.why_participated, rb.rbid, rb.flagged, rb.updated
+                      FROM dosomething_signup signups
+                      LEFT JOIN dosomething_reportback rb ON signups.uid = rb.uid AND signups.nid = rb.nid AND signups.run_nid = rb.run_nid
+                      WHERE signups.sid > $last_saved AND signups.sid NOT IN (SELECT sid FROM dosomething_rogue_signups)
+                      ORDER BY signups.sid");
+}
+else {
+  $signups = db_query('SELECT signups.sid, signups.uid, signups.nid, signups.run_nid, signups.source, signups.timestamp, rb.quantity, rb.why_participated, rb.rbid, rb.flagged, rb.updated
+                      FROM dosomething_signup signups
+                      LEFT JOIN dosomething_reportback rb ON signups.uid = rb.uid AND signups.nid = rb.nid AND signups.run_nid = rb.run_nid
+                      WHERE signups.sid NOT IN (SELECT sid FROM dosomething_rogue_signups)
+                      ORDER BY signups.sid');
+}
+
+$client = dosomething_rogue_client();
+
+foreach ($signups as $signup) {
+  $data = [];
+
+  echo 'Trying sid ' . $signup->sid . '...' . PHP_EOL;
+
+  $northstar_user = dosomething_northstar_get_user($signup->uid, 'drupal_id');
+
+  // Only try to send to Rogue if we have a Northstar ID
+  if (isset($northstar_user)) {
+    // Match Rogue's timestamp format
+    $created_at = date('Y-m-d H:i:s', $signup->timestamp);
+
+    // Format signup data
+    $data = [
+      'northstar_id' => $northstar_user->id,
+      'campaign_id' => $signup->nid,
+      'campaign_run_id' => $signup->run_nid,
+      'do_not_forward' => TRUE,
+      'source' => $signup->source,
+      'created_at' => $created_at,
+      'updated_at' => $created_at, // Set updated same as created, will be overwritten if there was a RB submitted
+    ];
+
+    // Include all the Reportback data and files if they exist
+    if ($signup->rbid) {
+      // Pull the quantity and why
+      $data['quantity'] = $signup->quantity;
+      $data['why_participated'] = $signup->why_participated;
+
+      // Match Rogue's timestamp format
+      $updated_at = date('Y-m-d H:i:s', $signup->updated);
+      $data['updated_at'] = $updated_at;
+    }
+
+    // Send the request to Rogue
+    try {
+      $response = $client->postSignup($data);
+
+      // Make sure we get a successful response
+      if ($response) {
+        // Store signup reference
+        dosomething_rogue_store_rogue_signup_references($signup->sid, $response);
+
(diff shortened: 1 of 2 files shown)


scripts/rogue-keeper-upper.php
```diff
@@ -0,0 +1,303 @@
...
+        variable_set('dosomething_rogue_last_signup_kept_up', $signup->sid);
+
+        // Send to StatHat
+        if (module_exists('stathat')) {
+          stathat_send_ez_count('drupal - Rogue - signup migrated - count', 1);
+        }
+      }
+
+      // Handle getting a 404
+      if (!$response) {
```
chloealee: super nit-picky (and out of curiosity!) but is there any reason why this is better than just doing an `else` statement in line 83?

scripts/rogue-keeper-upper.php
```diff
@@ -0,0 +1,303 @@
...
+    catch (DoSomething\Gateway\Exceptions\ApiException $e) {
+      // Put request in failed table for future investigation
+      dosomething_rogue_handle_migration_failure($data, $signup->sid, $signup->rbid, $fids);
+
+      // Set where we left off so we don't keep trying this one forever
+      variable_set('dosomething_rogue_last_signup_kept_up', $signup->sid);
+    }
+  }
+  else {
+    echo 'No northstar id, that is terrible ' . $signup->sid . PHP_EOL;
```
chloealee: love this message 😂

scripts/rogue-keeper-upper.php
katiecrane: @chloealee great question! I think that an `else` would work the same way! This is the way the logic is setup in the main script, but that's not a very good reason for keeping it like this. We usually will have a response, so I think changing it to an `else` actually makes a lot of sense so we don't have to do this check every time.

scripts/rogue-keeper-upper.php
katiecrane: all credit to @angaither hahaha

itsjoekent: giphy

scripts/rogue-keeper-upper.php
chloealee: ah okay cool, just wanted to make sure there wasn't anything I was missing!

chloealee (approved):
🎉

scripts/rogue-keeper-upper.php
```diff
@@ -0,0 +1,300 @@
...
+
+foreach ($signups as $signup) {
+  $data = [];
+
+  echo 'Trying sid ' . $signup->sid . '...' . PHP_EOL;
+
+  $northstar_user = dosomething_northstar_get_user($signup->uid, 'drupal_id');
+
+  // Only try to send to Rogue if we have a Northstar ID
+  if (isset($northstar_user)) {
```
DFurnes: Totally subjective, but I love using [guard clauses](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html) in cases like this, since it keeps the error handling closer to the check & helps avoid super-nested conditionals.

scripts/rogue-keeper-upper.php
```diff
@@ -0,0 +1,300 @@
...
+      $updated_at = date('Y-m-d H:i:s', $signup->updated);
+      $data['updated_at'] = $updated_at;
+    }
+
+    // Send the request to Rogue
+    try {
+      $response = $client->postSignup($data);
+
+      // Make sure we get a successful response
+      if ($response) {
```
DFurnes: Again, 100% subjective but everything besides the API call can be removed from the try/catch if you want.

scripts/rogue-keeper-upper.php
```diff
@@ -0,0 +1,300 @@
...
+        // Put request in failed table for future investigation
+        dosomething_rogue_handle_migration_failure($data, $signup->sid, $signup->rbid, $fids);
+
+        // Set where we left off so we don't keep trying this one forever
+        variable_set('dosomething_rogue_last_signup_kept_up', $signup->sid);
+
+      }
+    }
+    catch (GuzzleHttp\Exception\ServerException $e) {
+      // These aren't yet caught by Gateway
```
DFurnes: Gah, yeah we need to fix that! 😩

scripts/rogue-keeper-upper.php
```diff
@@ -0,0 +1,300 @@
...
+    $data = [
+      // Post
+      'northstar_id' => $northstar_user->id,
+      'campaign_id' => $post->nid,
+      'campaign_run_id' => $post->run_nid,
+      'caption' => $post->caption,
+      'source' => $post->source,
+      'remote_addr' => $post->remote_addr,
+      'status' => $rogue_status, //can you send this? not in docs
+      'do_not_forward' => TRUE,
```
DFurnes: Do we still do forwarding from Rogue to Phoenix?

DFurnes (approved):
Nice! This is one hell of a script! 🔥👏

scripts/rogue-keeper-upper.php
katiecrane: @DFurnes we don't use it, but in some cases I was getting errors without this because Rogue was still trying to. Removing that from the Rogue side is for sure the better solution since we don't ever want to do it now, I just made an issue for it so it doesn't get forgotten https://github.com/DoSomething/rogue/issues/241

scripts/rogue-keeper-upper.php
katiecrane: @DFurnes ahh and put all the rest after the try and catches? I think that would make it hard to check for a `404` because I'm not sure how to check if `$response` is truly not set because an error happened or if it's `null` because a `404` was returned.

scripts/rogue-keeper-upper.php
katiecrane: @DFurnes I'm confused about how this works if you're not returning something like from a function... 🆘

scripts/rogue-keeper-upper.php
DFurnes: Oh, yeah you'd have to drop in a `continue` for that I guess. I'm very okay leaving it as-is. 😄

scripts/rogue-keeper-upper.php
DFurnes: Cool cool, yeah that makes sense to clean up!

scripts/rogue-keeper-upper.php
DFurnes: You can use `continue` to skip to the next item in the foreach loop. ⏩

katiecrane: @DFurnes I updated to try to do this! it's already wayyy clearer I think

PR #7369, opened 2017-05-02
Allowing anonymous users to hit /northstar

itsjoekent:
#### What's this PR do?
This PR allows anonymous users to use the /northstar endpoint if they login, or continues redirecting authenticated users

Replicate the gifs!

Needed for phoenix next so users can access profile. We know this flow isn't perfect but we are going to overhaul session management in the future

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.module
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.module
@@ -106,8 +106,7 @@ function dosomething_user_menu() {
   $items['northstar/%'] = array(
     'page callback' => 'dosomething_user_get_view',
     'page arguments' => array(1),
-    'access callback' => 'user_view_access',
-    'access arguments' => array(1),
+    'access callback' => TRUE,
   );
   $items['user/magic/%'] = array(
     'page callback' => 'dosomething_user_magic_session',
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.module
```diff
@@ -106,8 +106,7 @@ function dosomething_user_menu() {
   $items['northstar/%'] = array(
     'page callback' => 'dosomething_user_get_view',
     'page arguments' => array(1),
-    'access callback' => 'user_view_access',
-    'access arguments' => array(1),
+    'access callback' => TRUE,
```
weerd, line 109: So we always want this to `TRUE`? Just double-checking...

itsjoekent: Sorry was at 1:1 w/ matt -- Yeah we always want this to be true, cause anonymous users shouldn't be blocked by it. I don't see any danger in this either, northstar id's are public

weerd: Cool. Just making sure.

weerd (approved):
👍

lib/modules/dosomething/dosomething_user/dosomething_user.module
DFurnes: Yeah, this makes good sense to me! 👍

PR #7371, opened 2017-05-04
A few keeper upper fixes

katiecrane:
#### What's this PR do?
1. Phoenix didn't like when the default timestamp variable was `null`. Making it `0` fixes the complaints. Also, setting it at 0 means that the first time the keeper upper is run, it will run through a lot of data that is already in Rogue. However, these updates only send the `why_participated` and `quantity` to Rogue and do not cause anything to be duplicated or any timestamps to be altered. Additionally, we are sure that we are capturing the latest `quantity` and `why_participated` for each signup. 

2. A `continue` was missing!

3. Added more helpful outputs.

👀  is probably sufficient!

I think everything is explained above, ping me about any context I didn't cover 🏓 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/scripts/rogue-keeper-upper.php
+++ b/scripts/rogue-keeper-upper.php
@@ -121,7 +121,7 @@ foreach ($signups as $signup) {
 }
 
 // 2. Quantity and why participated updates with no new file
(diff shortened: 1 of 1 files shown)


sbsmith86 (approved):
💯

PR #7375, opened 2017-05-05
Properly handle failing to find northstar user in Rogue keeper upper

### LINKED ISSUE #7374: Some errors not logging when attempting to send posts with no northstar id to Rogue
_BUG_
-------------------
#### Current Behavior
[These lines](https://github.com/DoSomething/phoenix/blob/dev/scripts/rogue-keeper-upper.php#L230-L231) are seemingly not executing. The Post is not sent to Rogue, but it also doesn't show up in the `dosomething_rogue_failed_migrations` table, so it keeps getting re-tried every time we run the keeper upper.

#### Desired Behavior
Failed posts are logged in the `dosomething_rogue_failed_migrations` table so that we know not to keep trying them.

#### Steps to Replicate
Run the Rogue keeper upper on Thor and you'll see a bunch of stuff like this:
```
Trying fid 374172 sid 3166626...
WD northstar: Exception: Error fetching Northstar user data for user with drupal_id = 58. in dosomething_northstar_get_user() (line 120 of /var/www/www.dosomething.org/releases/20170505143614/lib/modules/dosomething/dosomething_northstar/dosomething_northstar.module).    [error]
No northstar id, that is terrible 374172
```
Run the script again, and you'll see the same ones failing 😱 

#### Why This Matters
It doesn't do any harm to try to send these posts everytime, but it is a waste of time and makes it hard to see what actually does get sent through.

katiecrane:
#### What's this PR do?
When we got to a Post with no Northstar user, we weren't logging the information in the failed table, and now we are.

Compare what we send to `dosomething_rogue_handle_migration_failure` if we are dealing with a Post with no Northstar ID [here](https://github.com/DoSomething/phoenix/blob/dev/scripts/rogue-keeper-upper.php#L231) to the updates made to that function in this PR.

Check out #7374 for more details about this bug.

#### Relevant tickets
Fixes #7374

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
@@ -607,7 +607,24 @@ function dosomething_rogue_store_rogue_signup_references($sid, $rogue_signup) {
  */
 function dosomething_rogue_handle_migration_failure($values, $sid, $rbid = NULL, $fids = NULL, $response = NULL, $e = NULL) {
   // one row for each photo, just the sid if no northstar id, or just the signup info
-  if (!isset($values)) {
+  // When trying to send a file but there is no northstar id
+  if (empty($values) && isset($fids)) {
+      db_insert('dosomething_rogue_failed_migrations')
+      ->fields([
+        // Signup
+        'sid' => $sid,
+
+        // Reportback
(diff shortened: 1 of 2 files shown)


sbsmith86 (approved):
looks good to me!

PR #7378, opened 2017-05-08
Fixing path alias for reportback prove it button

itsjoekent:
#### What's this PR do?
Phoenix Next campaigns aren't redirected on the `node/*` urls, so users are getting the old template when they click `prove it` on their profile. This forces that button to always use the proper alias.

Click the prove it button

This path_alias was already [_globalized_](https://github.com/DoSomething/phoenix/blob/4495792c1d7f300bbd6de9e7d73e5947cc51dd73/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module#L579) so it doesn't need the global url function anymore.


itsjoekent:
@weerd Tagging you on this one but no rush!

weerd (approved):
👍

PR #7380, opened 2017-05-09
Reportback statuses count fix

weerd:
#### What's this PR do?
This PR adds a fix to help more accurately display the correct counts of Reportbacks of different statuses per campaign on the `/admin/users/rb` page. It updates the `ReportbackItem` class so that after an item is reviewed then the cached count of items based on status is updated so that it will accurately display the correct count back on the index list page.

However, this will only fix the counts _after_ a new Reportback is reviewed. If there are no RBs waiting for review for a campaign, then the counts will not be fixed. So to remedy this, this PR also adds a new page tab called "Refresh Counts" on the `/admin/users/rb` page. This new page provides a form to select a campaign and force refresh the Reportback statuses counts:

image

This PR also does some general code cleanup like switching to literal array syntax and adding some carriage returns, cause _why_ did we insist on mushing all our code together without some breathing room! It was hurting my poor 👀  😖 

👁 

#### Relevant tickets
Refs https://trello.com/c/52cHXmq3/566-2-drupal-inbox-overview-showing-zero-even-though-there-are-rbs-with-that-status-time-box-1-2-hours

#### Checklist
- [x] Tested on staging.
- [x] Pinged a PM if this is a larger PR that would benefit from some additional testing love

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,10 +12,10 @@
     "preferred-install": "dist"
   },
   "require": {
+    "dosomething/gateway": "^1.0.0-rc14",
     "stripe/stripe-php": "^1.18.0",
-    "zendesk/zendesk_api_client_php": "^1.2.0",
-    "tightenco/collect": "^5.2",
-    "dosomething/gateway": "^1.0.0-rc14"
+    "tightenco/collect": "^5.4",
+    "zendesk/zendesk_api_client_php": "^1.2.0"
   },
   "require-dev": {
     "symfony/var-dumper": "^3.1"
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -12,10 +12,10 @@
     "preferred-install": "dist"
   },
   "require": {
+    "dosomething/gateway": "^1.0.0-rc14",
     "stripe/stripe-php": "^1.18.0",
-    "zendesk/zendesk_api_client_php": "^1.2.0",
-    "tightenco/collect": "^5.2",
-    "dosomething/gateway": "^1.0.0-rc14"
+    "tightenco/collect": "^5.4",
```
weerd, line 17: One of the methods I used for the collection was giving back really odd results. Turns out the method was updated and [fixed in v5.4](https://github.com/tightenco/collect/pull/32/files#diff-1f18a78060929ba4f93e96a4d21d022dL628).

DFurnes (approved):
Nice! 💧

itsjoekent:
🎇

PR #7383, opened 2017-05-09
Ghost Inspector Bug

chloealee:
#### What's this PR do?
Fixes ghost inspector bug:
screen shot 2017-05-09 at 4 42 36 pm
 

👀 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
@@ -378,7 +378,7 @@ function paraneue_dosomething_get_themed_reportback_item($data, $rogue_rb_item =
         );
       }
 
-      $data->kudos = dosomething_kudos_get_kudos_data_for_fid($data->id, $total_rogue_kudos ? $total_rogue_kudos : 0, $liked_by_current_user ? $liked_by_current_user : null);
+      $data->kudos = dosomething_kudos_get_kudos_data_for_fid($data->id, $rogue_rb_item ? $total_rogue_kudos : 0, $liked_by_current_user ? $liked_by_current_user : null);
   }
 
   if ($data instanceof ReportbackItem && user_access('view any reportback')) {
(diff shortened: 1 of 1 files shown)


lib/themes/dosomething/paraneue_dosomething/includes/helpers.inc
```diff
@@ -378,7 +378,7 @@ function paraneue_dosomething_get_themed_reportback_item($data, $rogue_rb_item =
         );
       }
 
-      $data->kudos = dosomething_kudos_get_kudos_data_for_fid($data->id, $total_rogue_kudos ? $total_rogue_kudos : 0, $liked_by_current_user ? $liked_by_current_user : null);
+      $data->kudos = dosomething_kudos_get_kudos_data_for_fid($data->id, $rogue_rb_item ? $total_rogue_kudos : 0, $liked_by_current_user ? $liked_by_current_user : null);
```
chloealee: @DFurnes I _think_ this should fix it so if there is a `$rogue_rb_item` it'll know about `$total_rogue_kudos` and if not, it will set to `0`? Or will it still complain and say it is undefined if it doesn't go through that `if` statement?

DFurnes: Yeah, this should be okay since `$rogue_rb_item` is in the outer function scope! 👌 If we're not using them anymore, can we remove `$total_rogue_kudos` and `$liked_by_current_user`?

DFurnes: If you still want to use them, you can also just define them as `0` or `false` in the outer scope!

chloealee: oh good call! I think we still want to use `$liked_by_current_user` but I deleted `$total_rogue_kudos`!

DFurnes (approved):

PR #7384, opened 2017-05-11
Only prompt if user isn't in a competition.

DFurnes:
#### What's this PR do?
This pull request updates Phoenix Ashes to check with Gladiator before displaying the "competition prompt" to a user. This will prevent users who originally joined a competition on Phoenix Next from being prompted again.

👀! Any error-checking I'm missing?

This gives a better user experience for users who "start" in Phoenix Next but then return to the campaign in Phoenix Ashes.

#### Relevant tickets
Fixes 🏆.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -760,7 +760,15 @@ function dosomething_campaign_add_signup_data_form_vars(&$node, $langcode) {
   // Pass to the user signup data form.
   $node->content['signup_data_form'] = drupal_get_form('dosomething_signup_user_signup_data_form', $signup);
   // If the signup_data_form is required and user has not submitted form yet:
-  if ($config['required'] && !$signup->signup_data_form_timestamp) {
+  $hasCompetitionRecordedLocally = $signup->signup_data_form_timestamp;
+  $shouldShowCompetitionPrompt = $config['required'] && !$hasCompetitionRecordedLocally;
+
+  // If Gladiator is enabled, check there too!
+  if (module_exists('dosomething_gladiator')) {
+    $shouldShowCompetitionPrompt = $shouldShowCompetitionPrompt && ! dosomething_gladiator_in_a_competition($signup);
+  }
+
+  if ($shouldShowCompetitionPrompt) {
     // Store flag to indicate we need to prompt user (handled in theme).
     $node->required_signup_data_form = 1;
     // If form is configured to include a skip button:
--- a/lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.module
+++ b/lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.module
@@ -38,6 +38,41 @@ function _dosomething_gladiator_build_http_client() {
 }
 
 /**
+ * Check if the given user is already in a competition for this campaign.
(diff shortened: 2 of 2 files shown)


lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.module
```diff
@@ -38,6 +38,41 @@ function _dosomething_gladiator_build_http_client() {
...
+    'campaign_run_id' => $signup->run_nid,
+  ]);
+
+  $response = drupal_http_request($client['base_url'] . '/users?'.$query, [
+    'method' => 'GET',
+    'headers' => $client['headers'],
+  ]);
+
+  $json = json_decode($response->data, TRUE);
+  $already_in_gladiator = !empty($json['data']['waitingRoom']) || !empty($json['data']['competition']);
```
itsjoekent, line 64: Will this fail if `$json['data']` does not exist? I remember accounting for that in p-next

itsjoekent (approved):
Seems good, just a quick suggestion

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -760,7 +760,15 @@ function dosomething_campaign_add_signup_data_form_vars(&$node, $langcode) {
   // Pass to the user signup data form.
   $node->content['signup_data_form'] = drupal_get_form('dosomething_signup_user_signup_data_form', $signup);
   // If the signup_data_form is required and user has not submitted form yet:
-  if ($config['required'] && !$signup->signup_data_form_timestamp) {
+  $hasCompetitionRecordedLocally = $signup->signup_data_form_timestamp;
```
weerd, line 763: if timestamp doesn't exist, this returns `null` right?

lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.module
DFurnes: Nope, `empty` won't give a warning if the variable or key doesn't exist. Same deal as with `isset`.

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
DFurnes: Yep.

lib/modules/dosomething/dosomething_gladiator/dosomething_gladiator.module
weerd: I didn't know you can pass an `||` statement to `empty()`.  Thought it had to be `||` with two separate `empty()` calls on each side 🤔  
cool!

weerd: Nevermind, misread :man_facepalming: ☕️ is going to kick in reaaaal soon.

weerd (approved):
Looks good to me 👍

DFurnes:
[PRIORITY: As a developer, I want to add better error handling for "duplicate" gladiator signups for a campaign so that the site/experience doesn't BREAK.](https://trello.com/c/Hj3kcfz9/641-priority-as-a-developer-i-want-to-add-better-error-handling-for-duplicate-gladiator-signups-for-a-campaign-so-that-the-site-expe)

PR #7386, opened 2017-05-11
Counter bug when rogue is off

### LINKED ISSUE #7385: Counter for gallery is incorrect when Rogue is turned off.
_BUG_
-------------------
#### Current Behavior
- When Rogue is turned off, the counter in the gallery is incorrect. 

#### Desired Behavior
- We want the counter in the gallery to be correct when Rogue is both on and off! 

#### Why This Matters
- It is confusing to the user if the counter isn't correct or displays a `0` after reacting to a photo.

chloealee:
#### What's this PR do?
Changes conditional logic to make sure the counter in the gallery works when Rogue is both on and off. 

👀 

#### Relevant tickets
Fixes #7385 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
+++ b/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
@@ -325,13 +325,18 @@ function dosomething_kudos_get_kudos_data_for_fid($fid = NULL, $total_rogue_kudo
     return [];
   }
 
-  if ($total_rogue_kudos === 0) {
-    $total_kudos = 0;
-  } elseif ($total_rogue_kudos) {
-    $total_kudos = $total_rogue_kudos;
-  } else {
+  if(!isset($total_rogue_kudos)) {
     $total_kudos = dosomething_kudos_get_total_for_file_by_term_name($fid, 'heart');
+  } else {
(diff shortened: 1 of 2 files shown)


DFurnes (approved):

lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module
```diff
@@ -325,12 +325,10 @@ function dosomething_kudos_get_kudos_data_for_fid($fid = NULL, $total_rogue_kudo
     return [];
   }
 
-  if ($total_rogue_kudos === 0) {
-    $total_kudos = 0;
-  } elseif ($total_rogue_kudos) {
-    $total_kudos = $total_rogue_kudos;
-  } else {
+  if(!isset($total_rogue_kudos)) {
```
DFurnes: Not a merge blocker, but I think we'd have an easier time thinking about this if we always split code paths based on the feature flag, rather than based on whether a param exists or not.

chloealee: oh @DFurnes I like that a lot better!! I just updated it - do you mind taking another look and giving comments or a 👍 ?

DFurnes (approved):

PR #7393, opened 2017-05-25
Send new reviews to Rogue with the keeper upper script

katiecrane:
#### What's this PR do?
1. Adds a function to `Rogue.php` to hit the [`/reviews` endpoint](https://github.com/DoSomething/rogue/blob/master/documentation/endpoints/reviews.md#reviews---api) in Rogue.
2. Adds Part 4 to the keeper upper! Now if a post that lives in Rogue gets reviewed in Phoenix, we send that review on over. We look for all new reviews after the timestamp of the last one sent.
3. Handles failure of reviews. We have to store the ones that failed so we don't lose them forever, which is what would happen if one failed and then the next one succeeded.

Maybe test out the query and make sure it is catching all new reviews. Otherwise walk through the new logic in the script and see if it makes sense.

#### Relevant tickets
[Trello card
](https://trello.com/c/YrAFkbGX/434-5-as-an-admin-i-don-t-want-to-have-to-re-review-posts-in-rogue-once-it-is-turned-on)
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
@@ -697,6 +697,23 @@ function dosomething_rogue_handle_migration_failure($values, $sid, $rbid = NULL,
         ])
         ->execute();
       }
+    } elseif (array_key_exists('admin_northstar_id', $values)) {
+       // Store review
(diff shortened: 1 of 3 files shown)


chloealee (approved):
this looks perfect to me, @katiecrane ! Great work!

PR #7394, opened 2017-05-25
Keeper Upper - use the right variable

katiecrane:
#### What's this PR do?
Makes a fix to use the right variable when handling a failure. It's in the update section, so it should be using `$update` instead of `$post`.

👀  should do it!

#### Checklist
- [ ] Tested on staging.


chloealee (approved):
👍

PR #7397, opened 2017-05-30
Don't get whole northstar user in Rogue Keeper Upper

katiecrane:
#### What's this PR do?
We only need the Northstar ID and not the whole Northstar user, so let's only get that so we don't have to slow down the script with a possible ping to Northstar. Plus, we were seeing some 👻  in getting the whole Northstar user.

My biggest concern is that I am using all of the right variables in all of the right places. Also handling if for some reason there is no Northstar ID, and of course getting the Northstar ID in the first place.

At some point we started using `dosomething_northstar_get_user` instead of `dosomething_user_get_northstar_id` in the Rogue scripts, but now it seems like `dosomething_user_get_northstar_id` is for sure the way to go.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/scripts/rogue-keeper-upper.php
+++ b/scripts/rogue-keeper-upper.php
@@ -133,7 +133,7 @@ foreach ($postless_updates as $update) {
   echo 'Trying to update rbid ' . $update->rbid . '...' . PHP_EOL;
 
   // Get the Northstar ID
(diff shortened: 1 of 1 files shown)


DFurnes (approved):
👍 from me!

PR #7399, opened 2017-06-04
#7390: Upgrade Views from 7.x-3.14 to 7.x-3.16

### LINKED ISSUE #7390: Upgrade Views from 7.x-3.14 to 7.x-3.16
_FEATURE OVERVIEW_
-------------------
#### User Story
We are a few versions behind the recommended version of the Views contrib module. One of the releases includes a security release.

We are currently on version 7.x-3.14. 

**7.x-3.15:**
Maintenance and Security release. ([Release notes](https://www.drupal.org/project/views/releases/7.x-3.15))

Security Fix for [Moderately Critical - Access Bypass - SA-CONTRIB-2017-022](https://www.drupal.org/node/2854980)

_The Views module fails to add the required query tags to listings of Taxonomy Terms, which could cause private data stored on Taxonomy Terms to be leaked to users without permision to view it._

_This is mitigated by the fact that a View must exist that lists Taxonomy Terms which contain private data. If all the data on Taxonomy Terms is public or there are no applicable Views, then your site is unaffected._

Fixes:
- Add additional query tags
- Complete entity_uri support for taxonomy_term
- "In none of" (= Off/0) filter operator stopped working after Views update to 7.x-3.14
- Time-based output cache key no longer changes when database results change, results cache no longer depends on _GET parametrs
- AJAXified glossary view arguments fail when Language URL detection enabled
- Prevent jump menu warning message by checking for an empty value.
- Removed form_process_select from the process array. 
- Getting drupal_trigger_error_with_delayed_logging errors on Drupal 7.x dev (7.50)

**7.x-3.16-rc1:**
Maintenance release. ([Release notes](https://www.drupal.org/project/views/releases/7.x-3.16-rc1))

This release is focused primarily on fixing bugs and regressions, plus some minor tweaks that should not affect sites. Check release notes as the list of changes is large.

**7.x-3.16:**
Maintenance release. ([Release notes](https://www.drupal.org/project/views/releases/7.x-3.16))

**This is a rather major bug fix release.** 

It is strongly recommended to thoroughly test this release before uploading to a production site in case there are any regressions. There are two changes since 7.x-3.15 which affect how CSS classes are generated, so please pay particular attention to the themed output. Please report regressions immediately so that we can rectify them in a future release.

#### Why This Matters

crittermike:
#### What's this PR do?
Upgrades Views from 7.x-3.14 to 7.x-3.16.

The code itself is as simple as could be. For testing, apply it then run `ds build` then browse the site and look at any views and make sure they still appear/work correctly. 

The 3.16 release especially changed the way that CSS classes were generated so that is the highest risk factor, that a class would have changed, changing the display of a view. I couldn't find anything like that, and Views itself should have fixed any bugs like that since the release has been out for 2 months now, but it's still a good place to focus testing.

#### Relevant tickets
Fixes #7390

cc @mirie

### DIFF
--- a/lib/profiles/dosomething/drupal-org.make
+++ b/lib/profiles/dosomething/drupal-org.make
@@ -240,7 +240,7 @@ projects[view_unpublished][version] = "1.2"
 projects[view_unpublished][subdir] = "contrib"
 
 ; Views
-projects[views][version] = "3.14"
+projects[views][version] = "3.16"
 projects[views][subdir] = "contrib"
 
 ; Views Data Export


mirie (approved):

mirie:
@DFurnes, this change seems fine to me, but I'm unsure of any potential side effects from the css class change @mikecrittenden pointed out. Is this something the ghost inspector tests would catch on Thor?

DFurnes:
@mirie Maybe! We do screenshot regression testing with Ghost Inspector but I don't know if they always cause a test failure. I'm pretty sure the only place we use Views for user-facing elements is the Explore Campaigns page so probably safe to just do a quick visual once-over.

PR #7400, opened 2017-06-05
#7389: Upgrade Views Data Export from 7.x-3.0-beta8 to 7.x-3.2

### LINKED ISSUE #7389: Upgrade 	Views data export from 7.x-3.0-beta8 to 7.x-3.2
_FEATURE OVERVIEW_
-------------------
#### User Story
We are a few versions behind the recommended version of the Views data export contrib module. One of the releases includes a security release.

We are currently on version 7.x-3.0-beta8. 

**7.x-3.0-beta9:**
Maintenance release. ([Release notes](https://www.drupal.org/project/views_data_export/releases/7.x-3.0-beta9))

Fixes:
- Added allow carriage return replacement config
- Labels get translated
- Array to String error when exporting date exposed options

**7.x-3.0:**

Maintenance and Security release. ([Release notes](https://www.drupal.org/project/views_data_export/releases/7.x-3.0))

Fixes:
- Ensure access control is performed on batched export downloads
- Leading Zeros and exporting "Excel" files
- Fixed MySQL error for 'input_required' exposed filters, if no filter value is present in the URL
- Fixed non-batched exports unable to specify arguments in Drush
- Fixed Drush unknown option: arguments

**7.x-3.1:**

Maintenance release. ([Release notes](https://www.drupal.org/project/views_data_export/releases/7.x-3.1))

Fixes:
- Access denied when performing batched export as admin

**7.x-3.2:**

Maintenance release. ([Release notes](https://www.drupal.org/project/views_data_export/releases/7.x-3.2))

Fixes:
- Regression for non-batch export with no arguments.
- Current Images for Buttons Fail 508 Color Contrast
- Add support for batched exports of Search API views
- SHOW VARIABLES query for max_allowed_packet causes error
- Reduce garbage collection times

#### Why This Matters

crittermike:
#### What's this PR do?

Upgrades the Views Data Export module from 7.x-3.0-beta8 to 7.x-3.2.

Clone it and run `ds build` basically. As far as I can tell this isn't actually being used anywhere though...? The `views_data_export` DB table is empty (at least on my DB dump) and I don't see any exported views using that type of display in the codebase. 

If that's the case, maybe this should just be disabled? I feel like I must be missing something.

#### Relevant tickets
Fixes #7389.

### DIFF
--- a/lib/profiles/dosomething/drupal-org.make
+++ b/lib/profiles/dosomething/drupal-org.make
@@ -244,7 +244,7 @@ projects[views][version] = "3.14"
 projects[views][subdir] = "contrib"
 
 ; Views Data Export
-projects[views_data_export][version] = "3.0-beta8"
+projects[views_data_export][version] = "3.2"
 projects[views_data_export][subdir] = "contrib"
 
 ; Views Grouping Row Limit


mirie:
@aaronschachter / @DFurnes: Are we using the views data export module ? If we're no longer using this, I agree we should uninstall it and remove it from the codebase.

aaronschachter:
An [oldie but a goodie](https://github.com/DoSomething/phoenix/pull/3120), the initial issues for this feature were defined in Jira. I think we added them with the intent to disable on production, but use on other instances like QA/Thor for data ingestion. I didn't think we ever enabled this on production, but sounds like we did. If anyone still uses it, my guess would that it'd be the Campaigns teams...

mirie:
Ah, right good point (I keep forgetting that Staging is just completely dummy data). I see on Production that a single view, users_permissions, has a data export display (/users-all-perms-data). However, this view display is not exported to features (is it typical to have view configuration only in the db vs exported as part of a feature?).

aaronschachter:
Not typical, protocol should have been to export it as a feature (here's an [example](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.views_default.inc)). I don't know why or when the `users_permissions` view was set up.. it sorta looks a lot like the kind of export we didn't want to be running on production in the first place (creating a csv of user info on prod server)

mirie:
yeah, upon further investigation this view is definitely weird. I see that the csv download = most recently created 10 users who are active, don't have "dosomething" in their email address, and has a user role set. One page display returns 11 records, and the other one returns nothing. Asking about this in #product-team.

mirie:
Still trying to track down an answer on this one. Based on information gathered so far, feeling like this view is not actually used, and a module uninstall / removal is actually required as opposed to an upgrade. Will do some more asking around and will report back with next steps to take.

mirie:
@mikecrittenden Ok it's not being used. I've exported the view config on prod just in case someone pipes up later about it, and deleted it on Production. 

Instead of upgrading this module, can you refactor this PR to uninstall and remove this module. While you're addressing that problem, could you also handle the logging warning work I previously deprioritized? (This seems like a good time to clean up the complaints from Drupal about missing modules). I will add a comment to the original issue attached to this PR.

crittermike:
@mirie For uninstalling and removing a module, that'd have to be 2 separate PRs and deploys, right? One to uninstall and the next to remove? Otherwise the module would be removed from the codebase before it was uninstalled assuming the build runs drush make before it runs updb like the local `ds build` does. 

Actually, that may be the reason we're seeing all those warnings during updb - modules that were removed and uninstalled in the same PR that couldn't ever actually get uninstalled.

mirie:
yeah, that's more accurate. I should have said issue instead of PR. Thanks for clarifying :)

PR #7404, opened 2017-06-07
Fix for Signup Bug when Rogue is Turned On

chloealee:
#### What's this PR do?
- Fixes the signup bug when Rogue is turned on (where you can't get to regular campaign page after signing up) 

- Turn Rogue on.
- Try to sign up for a campaign. 
- Make sure you get re-directed to the campaign page instead of re-directed to the Signup page. 

Reminder: when merged and deployed to Thor, make sure that competitions prompt/custom social share prompt still works if on.

#### Relevant tickets
Fixes https://trello.com/c/vsSBl0Xf/462-8-%F0%9F%90%9B-when-rogue-is-turned-on-youre-not-able-to-sign-up-for-a-campaign-%F0%9F%90%9B

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -49,9 +49,9 @@ function dosomething_campaign_preprocess_node(&$vars) {
       ],
     ]);
 
-    $response['data'] ? $signup_exists = TRUE : $signup_exists = FALSE;
+    $signup_exists = ! empty($response['data']);
   } else {
-    dosomething_campaign_is_pitch_page($node) ? $signup_exists = FALSE : $signup_exists = TRUE;
+    $signup_exists = ! dosomething_campaign_is_pitch_page($node);
   }
 
   if ($pitch_path == $current_path || ! $signup_exists) {
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -40,9 +40,7 @@ function dosomething_campaign_preprocess_node(&$vars) {
     global $user;
     $northstar_user = dosomething_user_get_northstar_id($user->uid);
 
-    $rogue = new Rogue(ROGUE_API_URL . '/',
-      ['headers' => ['X-DS-Rogue-API-Key' => ROGUE_API_KEY]]
-    );
+    $rogue = dosomething_rogue_client();
 
     $response = $rogue->getActivity([
       'filter' => [
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
@@ -38,7 +38,7 @@ function dosomething_campaign_preprocess_node(&$vars) {
 
   if (variable_get('rogue_collection', FALSE)) {
     global $user;
-    $northstar_user = dosomething_northstar_get_user($user->uid, 'drupal_id');
+    $northstar_user = dosomething_user_get_northstar_id($user->uid);
 
     $rogue = new Rogue(ROGUE_API_URL . '/',
       ['headers' => ['X-DS-Rogue-API-Key' => ROGUE_API_KEY]]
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -36,7 +35,28 @@ function dosomething_campaign_preprocess_node(&$vars) {
...
+    );
+
+    $response = $rogue->getActivity([
+      'filter' => [
+        'northstar_id' => $northstar_user->id,
+        'campaign_id' => $vars['nid'],
+      ],
+    ]);
+
+    $response['data'] ? $signup_exists = TRUE : $signup_exists = FALSE;
```
DFurnes: Can you reformat this so the assignment is on the left?

```php
$signup_exists = ! empty($response['data']);
```

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -36,7 +35,28 @@ function dosomething_campaign_preprocess_node(&$vars) {
...
+  } else {
+    dosomething_campaign_is_pitch_page($node) ? $signup_exists = FALSE : $signup_exists = TRUE;
```
DFurnes: And same here:

```php
$signup_exists = dosomething_campaign_is_pitch_page($node);
```

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -36,7 +35,28 @@ function dosomething_campaign_preprocess_node(&$vars) {
   dosomething_campaign_preprocess_common_vars($vars, $wrapper);
 
   $pitch_path = 'node/' . $node->nid . '/pitch';
-  if ($pitch_path == $current_path || dosomething_campaign_is_pitch_page($node)) {
+
+  if (variable_get('rogue_collection', FALSE)) {
+    global $user;
+    $northstar_user = dosomething_northstar_get_user($user->uid, 'drupal_id');
+
+    $rogue = new Rogue(ROGUE_API_URL . '/',
```
DFurnes: You can use the `dosomething_rogue_client` method to get a pre-made client:

```php
$rogue = dosomething_rogue_client();
```

lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
```diff
@@ -98,11 +98,14 @@ function dosomething_signup_form_submit($form, &$form_state) {
   }
 
   if (variable_get('rogue_collection', FALSE)) {
+    $form_state['values']['source'] = 'phoenix-web';
```
DFurnes: Does this normally have more specific source information? Like `fastweb` or `node/1141`? If so, we should probably send it along so we make sure we don't lose any data.

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.theme.inc
```diff
@@ -36,7 +35,28 @@ function dosomething_campaign_preprocess_node(&$vars) {
   dosomething_campaign_preprocess_common_vars($vars, $wrapper);
 
   $pitch_path = 'node/' . $node->nid . '/pitch';
-  if ($pitch_path == $current_path || dosomething_campaign_is_pitch_page($node)) {
+
+  if (variable_get('rogue_collection', FALSE)) {
+    global $user;
+    $northstar_user = dosomething_northstar_get_user($user->uid, 'drupal_id');
```
DFurnes: Let's use [`dosomething_user_get_northstar_id`](https://github.com/DoSomething/phoenix/blob/66f013d8df75117d8e7f392fe59036d752451799/lib/modules/dosomething/dosomething_user/dosomething_user.module#L1270-L1284) to save a network request.

lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
chloealee: oh that is a good question! I just noticed that it was saving in the database here at `null` when being forwarded to Rogue so wanted to make that more specific. I can delete this line though @DFurnes

DFurnes: Oh hmm, are we still saving records in Phoenix when forwarding signups to Rogue? 😲

chloealee: oh sorry! I meant storing source as `null` in the Rogue database, not the phoenix!

chloealee: @DFurnes  I deleted that line because I just noticed that the `source` was actually coming through to the Rogue database - thanks for catching that! 
screen shot 2017-06-07 at 4 27 45 pm

DFurnes (approved):
This looks good to me! 👍

PR #7408, opened 2017-06-08
Drush command for backfilling age data for users that want affiliate messaging.

weerd:
#### What's this PR do?
This PR adds a Drush command that can be run to update all the records in the `dosomething_campaign_affiliate_opt_ins` to add each user's age if we don't already have that information.

```
$ drush affiliate-opt-ins-age-backfill
```

👁 

#### Relevant tickets
https://trello.com/c/akkAiFZq/678-3-as-a-campaign-lead-for-who-has-their-eye-on-you-i-want-age-added-to-the-opt-out-export-file

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.drush.inc
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.drush.inc
@@ -4,9 +4,10 @@
  * Implementation of hook_drush_command().
  */
 function dosomething_campaign_drush_command() {
-  $items = array();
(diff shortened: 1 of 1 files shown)


DFurnes (approved):
Nice! 💧I had no idea you could configure scripts this way.

weerd:
Me neither! I saw a tutorial that did it that way and then saw we have some other scripts within modules as well. Defs feels cleaner if the script applies more specifically to a module 👏

PR #7412, opened 2017-06-13
Update user

itsjoekent:
#### What's this PR do?
This PR adds a new API endpoint so users can update their profile on Northstar (Specifically location fields for now)

Hit the endpoint with some of the address fields 

We're going to be using this for shipment form stuffs

#### Checklist

### DIFF
--- a/documentation/endpoints/users.md
+++ b/documentation/endpoints/users.md
@@ -229,3 +229,27 @@ POST /api/v1/users/{uid}/magic_login_url
 }
 ```
 
+## Update User Profile
+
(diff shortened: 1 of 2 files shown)


DFurnes:
Why do we need an API endpoint for this? If possible, I'd rather just handle updating the user's Northstar profile in [`dosomething_shipment_form_submit`](https://github.com/DoSomething/phoenix/blob/a05bd197bea695eed6cd0ff58d796d94a886f3a0/lib/modules/dosomething/dosomething_shipment/dosomething_shipment.forms.inc#L102-L129).

Issue #7287, opened 2017-01-25
Address forms are broken.


DFurnes:
_BUG_
-------------------
#### Current Behavior
Any address forms are broken after the changes made in #7283. This means that campaigns relying on the "Signup Data Form" or "Shipment Form" will not be able to collect user addresses.

#### Desired Behavior
If we need to send users things, we should be able to. Specifically: we need to fix these forms to render correctly (fields are currently out of order and "State" is missing) and to correctly persist data to the user's Northstar profile (rather than local Phoenix database).

#### Steps to Replicate
Like try it, it's broken.

#### Why This Matters
It doesn't, unless we're running a campaign that uses this feature.

#### Countries Affected (optional)
Every single one.

#### OS/Browser (optional)
The whole lot of 'em.

#### Relevant Screenshots + Links
🍇

DFurnes:
@mikefantini Can you help prioritize this? If we're not likely to be running any campaigns with address forms on Phoenix, it's probably safe to hold off on a fix.

DFurnes:
It sounds like we're not using these anytime soon – according to the campaigns team, [we'll be using a Google Form](https://dosomething.slack.com/archives/campaign-template/p1485377299000193) for both the upcoming Give A Spit and Garnier campaigns.

Let's hope that we're safely in the _new world_ by the time we need a feature like this again. ✨

DFurnes:
![](https://github.trello.services/images/mini-trello-icon.png) [Thumb socks on Phoenix-ashes not working](https://trello.com/c/FLQgiNgc/597-thumb-socks-on-phoenix-ashes-not-working)

PR #7415, opened 2017-06-14
Send user address to northstar

itsjoekent:
#### What's this PR do?
This PR updates the signup data form to push the user address data to Northstar. I re-used some of my logic from #7412 and also updated the helper text of the shipment module so nobody ever uses that again... I also fixed some spacing in the places I was touching the code because I was losing my mind.

Here is a profile I updated via the form

screen shot 2017-06-14 at 3 50 55 pm

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.address.inc
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.address.inc
@@ -154,7 +154,6 @@ function dosomething_user_save_address_values($values) {
 }
 
 function dosomething_user_save_address_values_to_northstar($user, $values) {
-  $northstar_id = dosomething_user_get_northstar_id($user->uid);
   $profile = [];
   $submission = $values['user_address'];
 
@@ -175,7 +174,7 @@ function dosomething_user_save_address_values_to_northstar($user, $values) {
     }
   }
 
-  dosomething_northstar_client()->asClient()->put('v1/users/id/' . $northstar_id, $profile);
+  dosomething_northstar_client()->asUser()->post('v1/profile', $profile);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.address.inc
```diff
@@ -145,9 +149,35 @@ function dosomething_user_save_address_values($values) {
...
+
+  foreach ($address_fields as $drupal_key => $northstar_key) {
+    $field = $submission[$drupal_key];
+
+    if ($field) {
+      $profile[$northstar_key] = $field;
+    }
+  }
+
+  dosomething_northstar_client()->asClient()->put('v1/users/id/' . $northstar_id, $profile);
```
DFurnes: Let's do this with the logged-in user's access token and the `/profile` endpoint.

```php
dosomething_northstar_client()->post('/profile', $profile);
```

itsjoekent: @DFurnes Hm, this doesn't appear to work. I added the missing `v1/` from the example but I get the following error,

```
DoSomething\Gateway\Exceptions\UnauthorizedException: Exception in Northstar "POST v1/profile" endpoint: [401] {"error":"access_denied","message":"The resource owner or authorization server denied the request.","hint":"Requires the `user` scope."} in DoSomething\Gateway\Northstar->handleUnauthorizedException() (line 379 of /var/www/dev.dosomething.org/vendor/dosomething/gateway/src/AuthorizesWithNorthstar.php).
```

itsjoekent: I also just gave this a try but doesn't seem to make a difference 

```
dosomething_northstar_client()->asUser()->post('v1/profile', $profile);
```

DFurnes: Oh, we probably don't need to be requiring the `user` scope there. Let's fix that in Northstar.

itsjoekent:
switched back to profile endpoint, works now

screen shot 2017-06-19 at 10 03 47 am

DFurnes (approved):

itsjoekent:
fixes #7287

PR #7416, opened 2017-06-14
7389 uninstall modules

### LINKED ISSUE #7389: Upgrade 	Views data export from 7.x-3.0-beta8 to 7.x-3.2

crittermike:
#### What's this PR do?

Removes views_data_export as a dependency so it won't re-enable itself at build time.

Run a full `ds build` to ensure that the module is disabled and stays disabled after the build is complete. 

Nope.

#### Relevant tickets
Fixes #7389

### DIFF
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.install
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.install
@@ -65,6 +65,12 @@ function dosomething_global_update_7003(&$sandbox) {
 }
 
 /**
+ * Empty update function.
+ */
+function dosomething_global_update_7004() {
+}
+
+/**
  * Uninstall some unused modules.
  */
 function dosomething_global_update_7005() {
--- a/lib/modules/dosomething/dosomething_global/dosomething_global.install
+++ b/lib/modules/dosomething/dosomething_global/dosomething_global.install
@@ -67,7 +67,7 @@ function dosomething_global_update_7003(&$sandbox) {
 /**
  * Uninstall some unused modules.
  */
-function dosomething_global_update_7004() {
+function dosomething_global_update_7005() {
   // Disable and uninstall Views Data Export
   if (module_exists('views_data_export')) {
     module_disable(array('views_data_export'));
--- a/lib/profiles/dosomething/dosomething.info
+++ b/lib/profiles/dosomething/dosomething.info
@@ -68,7 +68,6 @@ dependencies[] = variable
 dependencies[] = varnish
 dependencies[] = view_unpublished
 dependencies[] = views
-dependencies[] = views_data_export
 dependencies[] = views_limit_grouping
 dependencies[] = views_ui
 dependencies[] = xmlsitemap


lib/modules/dosomething/dosomething_global/dosomething_global.install
```diff
@@ -67,7 +67,7 @@ function dosomething_global_update_7003(&$sandbox) {
 /**
  * Uninstall some unused modules.
  */
-function dosomething_global_update_7004() {
+function dosomething_global_update_7005() {
```
mirie: I'm wondering since the update numbers skip from 7002 to 7005, if it's good practice to add commented, empty update functions for 7003 and 7004. What do you think?

crittermike: There is one for `7003` already, which is why I originally did `7004`, but now that `7004` has already run on Staging, I had to bump it to `7005` to get it to re-run now that the dependency has been removed. Make sense? Thoughts? @mirie

mirie: @mikecrittenden (Also sorry, I was scanning the file and didn't see that the long update for 7003 wasn't part of 7002.) I guess I mean something like this for 7004 (since it has already run on staging): 

```
/**
 * Empty update function.
 */
function file_entity_update_7103() {
}
```

crittermike: @mirie Gotcha - can't hurt. Just pushed that.

mirie (approved):

Issue #7392, opened 2017-05-18
Upgrade Metatag module from 7.x-1.4 to 7.x-1.21 (!!!)


mirie:
_FEATURE OVERVIEW_
-------------------
#### User Story
We are a **several, several (!!!!)** versions behind the recommended version of the Metatag contrib module. One of the releases includes a security release.

#### Additional Information (optional)
We are currently on version 7.x-1.4. The latest recommend version is **_several_** releases later: **1.21** 

The following lists the newer versions/release notes:

- **7.x-1.5-beta1:** Maintenance release. ([Release notes](https://www.drupal.org/project/metatag/releases/7.x-1.5-beta1)) - Bug fixes and a few improvements.
- **7.x-1.5:** Maintenance release. ([Release notes](https://www.drupal.org/project/metatag/releases/7.x-1.5))
  - This release includes a huge number of improvements. The primary changes are:
    - All entities (that are not configuration entities and have view modes) can have meta tags added to them, it's no longer necessary to change anything in hook_entity_info().
    - The settings page (admin/config/search/metatags/settings) provides full control over which entities may have meta tags added to them.
    - It's now possible to change the theme region used to trigger meta tag output, which will allow sites to work around problems with meta tags not being output.
    - Using tokens to replicate one meta tag into another works correctly now.
    - An import program is provided for sites upgrading from Nodewords and Drupal 6.
    - Almost 250 meta tags are now supported, including App Links, 29 more Dublin Core tags, Open Graph product tags, mobile tags, favicon tags, etc.
    - There are two change notices for this release:
      - [API change for how Metatag identifies compatible entities](https://www.drupal.org/node/2495819)
      - [Change in how Metatag loads entity info](https://www.drupal.org/node/2415215)
- **7.x-1.6:** Maintenance release. ([Release notes](https://www.drupal.org/project/metatag/releases/7.x-1.6))
  - This release includes a few important improvements:
    - The entity selection logic added to v1.5 has been improved. All content types, vocabularies and the user entity will now automatically be enabled when the module is installed, while other entities are disabled; an update script is included that will enable meta tags for all content types, vocabularies and the user entity, other entity types are left as-is.
    - An API improvement will automatically search values for image URLs that might happen to be embedded in formatted image links. This removes the need for other modules to get the URL for an image field, it can now be automatically extracted.
    - The Open Graph Product meta tags were moved into a separate sub module, as were the Dublin Core Advanced meta tags.
[...]

PR #7423, opened 2017-06-26
No-Ticket: Remove user entities from metatag table prior to running metatag updates

mirie:
#### What's this PR do?
This is related to work done for issue #7392  - Upgrade Metatag module from 7.x-1.4 to 7.x-1.21

On Thor, with the larger dataset, the upgrade process took 1 day and 21 hours. The reason for this appears to be that metatags is updating all of the user entities (there's a lot of them). This is completely unnecessary because user profiles are not exposed to the public and thus having metatags on them is also unnecessary.

This PR does:

- Removes User from metatag configuration. 
- Also includes a re-export (minor change) of the twitter image metatag.
- Implements update hook for dosomething_metatag.
- Implements hook_update_dependencies() to ensure that the dosomething_metatag hook runs prior to the time-consuming metatag updates.

1. Refresh Thor with Prod db prior to deploying to Thor
2. We should have less than the original 974,944 records to update (for update # 7041 / the update for 7040 ran quicker and covered a smaller subset of records).

If reviewing locally, you will have to update the systems table for the metatag entry to roll back to the schema version of 7035 to match the state of version 1.4 (reference =  http://cgit.drupalcode.org/metatag/tree/metatag.install?h=7.x-1.4&id=4ea7c8dabbc2e4eca0d83774e35caeb08c03eba7)

@mikecrittenden: Couldn't add you as a reviewer, but could you take a look?

N/A

#### Relevant tickets
Related to problem reported when run on Thor: #7392 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_metatag/dosomething_metatag.features.metatag.inc
+++ b/lib/modules/dosomething/dosomething_metatag/dosomething_metatag.features.metatag.inc
@@ -101,9 +101,6 @@ function dosomething_metatag_metatag_export_default() {
       'twitter:description' => array(
         'value' => '',
       ),
-      'twitter:image:src' => array(
-        'value' => '',
-      ),
       'twitter:image:width' => array(
         'value' => '',
       ),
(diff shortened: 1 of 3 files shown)


crittermike:
This looks great to me. I can't find anything to complain about @mirie.

DFurnes (approved):
Looks good to me!

PR #7424, opened 2017-06-27
Clean up unused database tables.

DFurnes:
#### What's this PR do?
I _love_ deleting things. This pull request removes unused auth/registration logic & unused fields:

↪️ Removes old password reset functionality & redirects any users that stumble on `/user/password` to go to Northstar (since resetting your Phoenix password doesn't do anything for you). (b937f9b)

🛂 Removes method that would trigger password resets in Quicksilver. (fa830fb)

🍃 Removes settings for international-only registration fields, and form alter hooks and after-builds for the login, registration, and password reset forms since they no longer exist here! (685106e)

🌐 Removes logic for international-specific fields on the registration form (last name & zip), since we no longer support those international sites (and don't register users in Phoenix anyways). (07dd5bc)

📧 Removes `field_opt_in_email` and `field_opt_in_sms`. These were added for international sites and are no longer being used in production. (d1ae837)

🚸 Removes `field_under_thirteen`. As far as I can tell, this was never used... we write to it after checking against the user's birthdate in `dosomething_user_is_under_thirteen()` but never read the value stored there anywhere. (b2eefce)

🔏 Removes `field_created_by_openid_connect`. This was used when we were doing a partial rollout of Northstar authentication so we could track who had a "real" hashed password in the Phoenix database. Since all users are created via Northstar now, this is unnecessary. (c5e36a7)

🐷 Fixes an old docblock for the heck of it. (74ebb9e)

Gimme a sanity check that I'm not deleting anything important here, and then we can test more thoroughly on Thor before we deploy to production.

This will save us 10GB of space in production, which means we can save 💸 on that AWS instance.

#### Relevant tickets
[Trello Card](https://trello.com/c/hOFuLh0Z/628-3-as-a-database-server-i-dont-want-to-be-storing-tables-i-dont-need-turn-into-checklist-jen)

#### Checklist
- [ ] Tested on staging.


DFurnes:
[(3) As a database server, I don't want to be storing tables I don't need.](https://trello.com/c/hOFuLh0Z/628-3-as-a-database-server-i-dont-want-to-be-storing-tables-i-dont-need)

lib/themes/dosomething/paraneue_dosomething/template.php
```diff
@@ -17,7 +17,7 @@
 require_once PARANEUE_PATH . '/includes/auth/login.inc';
 
 /**
- * We use our own asset pipeline (through Grunt), so we want to bypass most
+ * We use our own asset pipeline (through Webpack), so we want to bypass most
```
weerd: 👌

weerd:
> ... dosomething_user_is_under_thirteen() but never read the value stored there anywhere.

LOL. 😬

weerd:
> This will save us 10GB of space in production, which means we can save 💸 on that AWS instance.

💰 💰 💰

weerd:
@DFurnes do we need to also add an install function to remove the items from the DB or best to keep them in there, just not adding anymore data to those columns/tables?

weerd (approved):
Quick question but looks good to me! 👍

DFurnes:
> @DFurnes do we need to also add an install function to remove the items from the DB or best to keep them in there, just not adding anymore data to those columns/tables?

Nope! Since they're made with features, the tables will be removed when we `drush fra` on each environment. 🥞

weerd:
Huzzah! Sounds good 💯

DFurnes:
Oh hmm, I might've lied. They're still there on staging after deploying this. 🤔

weerd:
Yeah, I thought Features only deals with database-based (say that 3 times fast!) settings configuration changes, but you need the install scripts to add/remove columns/tables .

PR #7428, opened 2017-06-28
Simple Permalink when Rogue is on

chloealee:
#### What's this PR do?
- When Rogue is on, creates a simple permalink page that is standard for all users (doesn't include post photo, specific quantity/why, etc). 
  - Note that this can be done and most of the work to do so is in a commit in the branch but we first need to add another endpoint in Rogue to retrieve all information needed. 

Simple standard permalink page when Rogue is on: 
screen shot 2017-06-28 at 3 58 50 pm

Failed page still works when RB doesn't make it to Rogue:
screen shot 2017-06-28 at 2 15 52 pm

- Turn Rogue on.
- Reportback and/or update a reportback. 
- You should see the same page as the screen shot above.
- Turn Rogue off. 
- Everything should work as it did before. 
- Change Rogue URL to incorrect URL. 
- You should hit failed page (like screen shot above). 

The photo used in this simple permalink page is the same one that is set for the [failed permalink page](https://github.com/DoSomething/phoenix/pull/7152). cc @ngjo @lkpttn  

#### Relevant tickets
Fixes https://trello.com/c/ahKepTKN/476-5-as-a-phoenix-ashes-user-when-i-reportback-i-want-the-normal-permalink-page-to-be-shown-to-me-w-generic-text-that-indicates-the

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
+++ b/lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
@@ -81,7 +81,10 @@ function dosomething_metatag_preprocess_page(&$vars) {
   }
 
   // Add special share info to the permalink page.
-  if (in_array('page__reportback', $vars['theme_hook_suggestions']) && !in_array('page__reportback__failure', $vars['theme_hook_suggestions'])) {
+  $isReportbackFailure = in_array('page__reportback__failure', $vars['theme_hook_suggestions']);
+  $isRogueReportback = variable_get('rogue_collection', FALSE);
+
+  if (in_array('page__reportback', $vars['theme_hook_suggestions']) && !$isReportbackFailure && !$isRogueReportback) {
     dosomething_metatag_get_permalink_vars();
   }
 }
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -427,12 +427,13 @@ function dosomething_reportback_form_submit($form, &$form_state) {
     $rogue_reportback = dosomething_rogue_send_reportback_to_rogue($values, $user);
 
     // Make sure reportback made it back to Phoenix and store reference to the Rogue item if one was added
-    $rb_info = dosomething_rogue_check_rbid_and_store_ref($rogue_reportback);
+    // $rb_info = dosomething_rogue_check_rbid_and_store_ref($rogue_reportback);
 
-    if ($rb_info) {
+    // if ($rb_info) {
       // Redirect to permalink page.
-      dosomething_reportback_permalink_page($rb_info['rbid']);
-    } elseif ($rogue_reportback['data']['id']) {
+      // dosomething_reportback_permalink_page($rb_info['rbid']);
+    // } elseif ($rogue_reportback['data']['id']) {
+    if ($rogue_reportback['data']['id']) {
       dosomething_reportback_permalink_page($rogue_reportback['data']['signup_id'], $rogue_reportback, $campaign, $values['uid'], $values['why_participated'], $values['quantity']);
     } else {
       // Send user to the failed permalink page
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -431,17 +431,9 @@ function dosomething_reportback_form_submit($form, &$form_state) {
 
     if ($rb_info) {
       // Redirect to permalink page.
-      $form_state['redirect'] = array(
-        'reportback/' . $rb_info['rbid'],
-        array(
-          'query' => array(
-            'fid' => $rb_info['fid'],
-          ),
-        ),
-      );
+      dosomething_reportback_permalink_page($rb_info['rbid']);
     } elseif ($rogue_reportback['data']['id']) {
-      // Redirect to permalink page, populated as /reportback/{rogue_signup_id}?fid={rogue_post_id}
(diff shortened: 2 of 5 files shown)


lib/modules/dosomething/dosomething_metatag/dosomething_metatag.module
```diff
@@ -81,7 +81,10 @@ function dosomething_metatag_preprocess_page(&$vars) {
   }
 
   // Add special share info to the permalink page.
-  if (in_array('page__reportback', $vars['theme_hook_suggestions']) && !in_array('page__reportback__failure', $vars['theme_hook_suggestions'])) {
+  $isReportbackFailure = in_array('page__reportback__failure', $vars['theme_hook_suggestions']);
+  $isRogueReportback = variable_get('rogue_collection', FALSE);
+
+  if (in_array('page__reportback', $vars['theme_hook_suggestions']) && !$isReportbackFailure && !$isRogueReportback) {
```
sbsmith86, line 87: Nice! Def more readable to me

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -426,23 +426,12 @@ function dosomething_reportback_form_submit($form, &$form_state) {
...
-      );
-    } else {
-      // Send user to the failed permalink page
-      $form_state['redirect'] = 'reportback/failure/' . $values['nid'];
+    if (! $rogue_reportback) {
+      return $form_state['redirect'] = 'reportback/failure/' . $values['nid'];
     }
+
+    $rid = $rogue_reportback['data']['signup_id'];
+    $fid = $rogue_reportback['data']['id'];
```
sbsmith86, line 434: this would be the posts `id`, right?

chloealee: yup!

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -522,6 +520,32 @@ function dosomething_reportback_form_alter(&$form, &$form_state, $form_id) {
...
+
+  return dosomething_reportback_view_entity(entity_load('reportback', $rbid));
+}
+
+/**
+ * Callback for /rbf/FID page, when Rogue is enabled.
+ */
+function dosomething_reportback_view_rogue($rbid, $fid) {
+    // @TODO: Request signup from Rogue so we can display it on permalink page.
+    // $rogue_signup = dosomething_rogue_client()->get('v2/activity')
```
sbsmith86, line 542: Just to understand, is the issue you can't get a specific signup because this returns a whole bunch of signups for the campaign?

lib/themes/dosomething/paraneue_dosomething/templates/reportback/reportback-simple-permalink.tpl.php
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * @file
+ * Reportback confirmation/permalink page for Rogue.
+ */
+?>
+<article class="reportback__permalink">
+  <header role="banner" class="header ">
+    <div class="wrapper">
+      <h1 class="header__title">Awesome! We got your photo.</h1>
+      <p class="header__subtitle">You've entered into the scholarship and your photo will show up in the gallery soon!</p>
```
sbsmith86, line 11: A campaign might not have a scholarship. I think in the other permalink template we only conditionally  show the scholarship language.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
chloealee: yes @sbsmith86! Right now, there are no endpoints in the Rogue API that return information for a specific signup. We'll have to build this out (and all the work to build out the Rogue permalink page that better mimics what is going on is a commit in this PR) and once it's finished, should be pretty easy to do full permalink page!

sbsmith86: Ah cool. I thought the filter functionality was pretty much baked into the `/activity` endpoint with the `filtersTrait` that you were using on that

chloealee: We can easily add that in as a filter for sure! It just kinda kept snowballing into something bigger and adding more functionality to Rogue and going back and forth so thought it was better to do this now and then add later. 

Also, @DFurnes mentioned that it might be better to work on building out the Rogue API post launch with more RESTful API endpoints and this could be one of them `/signup/{signup_id}`.

lib/themes/dosomething/paraneue_dosomething/templates/reportback/reportback-simple-permalink.tpl.php
chloealee: ah ok! I was just going off the copy from the wire but good catch! 

@ngjo is there alternative copy to have here if there is no scholarship? Maybe just "Your photo will show up in the gallery soon!"?

chloealee: Actually, we might just want to scrap the scholarship lang here? 

I'm not sure if we can tell if there is a scholarship or not at this point. The only information we have on this page is the Rogue signup id and the Rogue post id so we can't tell what campaign it belongs to (and thus, can't check to see if the campaign has a scholarship or not).

sbsmith86: Whatever is easiest!

chloealee: awesome, updated! Thanks for flagging that Shae! Also updated the screen show above.

sbsmith86 (approved):
💯

DFurnes (approved):

PR #7434, opened 2017-06-30
Address form

itsjoekent:
#### What's this PR do?
- Adds the `package-lock` to the .gitignore till we decide to use npm 5 here
- Adds a new endpoint to let users update Northstar profiles 
- Adds a new address form that inlines on the campaign template, toggled with custom setting
- Adds a new Northstar function to clear the user cache

STILL TO-DO in future PR
- Final copy
- Adding some form validation 

(I totally broke our smaller PR rule :man_facepalming: )

1. Enable the address form
2. Signup for the campaign
3. Fill in your deets
4. Submit (You should see a success message)
5. Check your Northstar profile, the fields should be there
6. Refresh the campaign page, open up the form. It should prefill those inputs.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/dosomething_api.services.inc
+++ b/lib/modules/dosomething/dosomething_api/dosomething_api.services.inc
@@ -222,6 +222,9 @@ function dosomething_api_default_services_endpoint() {
         ),
       ),
       'targeted_actions' => array(
+        'update' => array(
+          'enabled' => '1',
+        ),
         'password_reset_url' => array(
           'enabled' => '1',
         ),
--- a/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
@@ -177,7 +177,7 @@ function _member_resource_access($op) {
 /**
  * Access calback for the user update resource.
  *
- * Only allow users to edit their own profile.
+ * Only allow users to edit their own profile for now.
  */
 function _member_update_access($id) {
   global $user;
--- a/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
@@ -287,7 +287,7 @@ function _member_resource_create($request) {
  * @return mixed Object of the newly created user if successful. String if errors.
  */
 function _member_resource_update($id, $request) {
-  $keys = ['addr_street1', 'addr_street2', 'addr_city', 'addr_state'];
+  $keys = ['addr_street1', 'addr_street2', 'addr_city', 'addr_state', 'addr_zip'];
   $fields = [];
   $account = $request['account'];
 
--- a/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
@@ -300,6 +300,7 @@ function _member_resource_update($id, $request) {
     }
   }
 
+  dosomething_northstar_clear_user_cache($id);
   $response = dosomething_northstar_client()->asClient()->put('v1/users/id/' . $id, $fields);
 
   return $response;
--- a/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
@@ -53,7 +53,7 @@ function _member_resource_definition() {
           'module' => 'dosomething_api',
           'name' => 'resources/member_resource',
         ),
-        'help'   => 'Returns Current User Activity. GET from users/current/activity, users/23423/activity',
+        'help' => 'Returns Current User Activity. GET from users/current/activity, users/23423/activity',
         'access callback' => '_member_resource_access',
         'access arguments' => array('activity'),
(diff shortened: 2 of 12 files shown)


itsjoekent:
form

lib/modules/dosomething/dosomething_api/resources/member_resource.inc
```diff
@@ -236,6 +275,38 @@ function _member_resource_create($request) {
...
+      $value = $account[$key];
+
+      if (isset($value)) {
+        $fields[$key] = $value;
+      }
+    }
+  }
+
+  dosomething_northstar_clear_user_cache($id);
+  $response = dosomething_northstar_client()->asClient()->put('v1/users/id/' . $id, $fields);
```
DFurnes: Let's use the profile endpoint & not do this as the client.

lib/modules/dosomething/dosomething_api/resources/member_resource.inc
```diff
@@ -236,6 +275,38 @@ function _member_resource_create($request) {
...
+ *
+ * @param $request Array passed to the endpoint.
+ *   Possible keys:
+ *   - addr_street1
+ *   - addr_street2
+ *   - addr_city
+ *   - addr_state
+ * @return mixed Object of the newly created user if successful. String if errors.
+ */
+function _member_resource_update($id, $request) {
```
DFurnes, line 288: Are we sending thumb socks to everyone who signs up for a campaign? If not, we need to create a shipment entity here to keep track of who's submitted this form.

lib/modules/dosomething/dosomething_api/resources/member_resource.inc
```diff
@@ -90,6 +90,33 @@ function _member_resource_definition() {
...
+          'name' => 'resources/member_resource',
+        ],
+        'help' => 'Update the users profile',
+        'args' => [
+          [
+            'name' => 'id',
+            'optional' => FALSE,
+            'source' => array('path' => 0),
+            'type' => 'string',
+            'description' => 'The Northstar ID of the user we\'re updating.',
```
DFurnes, line 106: Do we need this in the URL, since the user is only allowed to edit their own profile?

lib/modules/dosomething/dosomething_api/resources/member_resource.inc
itsjoekent: just confirmed again w/ Adam & Aki all signups that have the address filled out should get Thumb Socks. No shipment entities needed!

lib/modules/dosomething/dosomething_api/resources/member_resource.inc
itsjoekent: I originally tried doing a generic PUT / Update in the CRUD operations but I wasn't able to get it working properly for some reason, it would only return 404. The targetted action worked so I stuck with that ¯\_(ツ)_/¯

The upside to this way is we could technically let admins now update other users as well if we made some small changes in the permission hook, not that I expect us to actually do that tho

lib/modules/dosomething/dosomething_api/resources/member_resource.inc
itsjoekent: Sure thing!

lib/modules/dosomething/dosomething_api/dosomething_api.services.inc
```diff
@@ -222,6 +222,9 @@ function dosomething_api_default_services_endpoint() {
         ),
       ),
       'targeted_actions' => array(
+        'update' => array(
```
weerd, line 225: Not super keen that we're using targeted actions instead of just REST routes, but also aware that the `/users` endpoint is using old approach and not the newer RESTful one. Explanation [here](https://github.com/DoSomething/phoenix/pull/7434/files#r125085472)

lib/modules/dosomething/dosomething_api/resources/member_resource.inc
weerd: So the `/users` endpoint wasn't one I got a chance to update when I did the refactoring of the API using transformers, etc. I was only able to update what was immediately needed at the time for the mobile app. So the CRUD style operations might not be working because it's using the old approach and not fully set up to be RESTful.

itsjoekent: Ah gotcha. For the sake of time, I don't think we should concern ourselves with updating the CRUD operations to be RESTful, as we need this live for Wednesday

weerd: Yeah, not suggesting that. Just noting for help w/ possible explanation as to why initial attempt might not have worked :)

weerd (approved):
I think it looks good 👍

lib/modules/dosomething/dosomething_api/resources/member_resource.inc
itsjoekent: Yeah I suspected it was something funky but my original assumption was Drupal. I'm going to continue blaming Drupal because its easier :P

weerd: > When in doubt, blame Drupal!

:trollface:

PR #7436, opened 2017-07-06
Checking if true

itsjoekent:
Bug of address form, its showing up on all campaigns, because im only checking if its not null, but its going to be false on all campaigns if its not set! :man_facepalming:


weerd (approved):
😱

PR #7441, opened 2017-07-11
GET /signups/:sid proxy 

chloealee:
#### What's this PR do?
- Creates a Rogue proxy for the `GET /signups/:sid` endpoint 

- Turn Rogue on.
- Hit the `GET /signups/:sid` endpoint and the response should look exactly the same as if data was coming from Phoenix (compared results below) 
- Make sure this doesn't break the `GET /signups` endpoint that @katiecrane worked on! 
  - @katiecrane I separated the signup formatting to a separate helper function. I don't think that this broke anything but would your eyes just to double check! 

An example when Rogue is OFF: 
```
{
    "data": {
        "id": "234",
        "created_at": "1400010537",
        "campaign_run": {
            "id": "1807",
            "current": true,
            "timing": {
                "start": "2015-08-26 00:00:00",
                "end": null
            }
        },
        "uri": "http://dev.dosomething.org:8888/api/v1/signups/234",
        "campaign": {
            "id": "886",
            "title": "Kickball for All",
            "campaign_runs": {
                "current": {
                    "en": {
                        "id": "1807"
                    }
                },
                "past": []
            },
            "language": {
                "language_code": "en",
                "prefix": "us"
            },
            "translations": {
                "original": "en",
                "en": {
                    "language_code": "en",
                    "prefix": "us"
                }
            },
            "tagline": "Start an inclusive kickball team in your neighborhood.",
            "status": "active",
            "type": "campaign",
            "reportback_info": {
                "copy": "Upload yours by August 31.",
                "confirmation_message": "Back, back, back, GONE! You knocked this one out of the park.",
                "noun": "Games",
                "verb": "Played"
            },
            "magic_link_copy": null
        },
        "reportback": null,
        "user": {
            "id": "***",
            "drupal_id": "***",
            "first_name": "Mike",
            "last_initial": "",
            "photo": null,
            "country": "US"
        }
    }
}
```

An example when Rogue is ON: 
```
{
    "data": {
        "id": 122,
        "created_at": 1498676480,
        "campaign_run": {
            "id": "1842",
            "current": false,
            "timing": {
                "start": "2014-07-30 00:00:00",
                "end": null
            }
        },
        "uri": null,
        "campaign": {
            "id": "1334",
            "title": "Can-Tribute",
            "campaign_runs": {
                "current": {
                    "en": {
                        "id": "1842"
                    },
                    "en-global": {
                        "id": "1842"
                    }
                },
                "past": []
            },
            "language": {
                "language_code": "en",
                "prefix": "us"
            },
            "translations": {
                "original": "en",
                "en": {
                    "language_code": "en",
                    "prefix": "us"
                },
                "en-global": {
                    "language_code": "en-global",
                    "prefix": null
                }
            },
            "tagline": "May the odds be ever in your favor in this competitive food drive.",
            "status": "active",
            "type": "campaign",
            "reportback_info": {
                "copy": "Show us your photo to win tickets to a local pre-screening of The Hunger Games: Mockingjay - Part 1 for you and your friends! Upload your photo by November 6.",
                "confirmation_message": "Thanks for helping stock food bank shelves. Now go parade around in a fiery dress.",
                "noun": "items",
                "verb": "donated"
            },
            "magic_link_copy": null
        },
        "user": {
            "id": "****",
            "drupal_id": "****",
            "first_name": "Chloe",
            "last_initial": "L",
            "photo": null,
            "country": "US"
        },
        "reportback": {
            "id": 122,
            "created_at": 1498676480,
            "updated_at": 1498679922,
            "quantity": 343,
            "uri": null,
            "why_participated": "test!",
            "flagged": null,
            "reportback_items": {
                "total": 3,
                "data": [
                    {
                        "id": 194,
                        "status": "pending",
                        "caption": "go to rouge",
                        "uri": null,
                        "media": {
                            "uri": "https://s3.amazonaws.com/ds-rogue-prod/uploads/reportback-items/122-482cab927f6529c7f5e5c4bfd2594186-1498676480.jpeg",
                            "type": "image"
                        },
                        "created_at": 1498676481,
                        "kudos": {
                            "data": []
                        }
                    },
                    {
                        "id": 195,
                        "status": "pending",
                        "caption": "go to rouge",
                        "uri": null,
                        "media": {
                            "uri": "https://s3.amazonaws.com/ds-rogue-prod/uploads/reportback-items/122-482cab927f6529c7f5e5c4bfd2594186-1498676518.jpeg",
                            "type": "image"
                        },
                        "created_at": 1498676520,
                        "kudos": {
                            "data": []
                        }
                    },
                    {
                        "id": 196,
                        "status": "pending",
                        "caption": "get out",
                        "uri": null,
                        "media": {
                            "uri": "https://s3.amazonaws.com/ds-rogue-prod/uploads/reportback-items/122-482cab927f6529c7f5e5c4bfd2594186-1498679921.jpeg",
                            "type": "image"
                        },
                        "created_at": 1498679922,
                        "kudos": {
                            "data": []
                        }
                    }
                ]
            }
        }
    }
}
```

#### Relevant tickets
Needs https://github.com/DoSomething/rogue/pull/300 to be merged in order for this to work. 

Fixes https://trello.com/c/QYdM5vVN/485-as-a-developer-i-want-to-make-sure-phoenixs-signups-sid-endpoint-has-a-rogue-proxy-for-gladiator-to-still-work-when-rogue-is-tur

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
@@ -180,7 +180,7 @@ function dosomething_rogue_format_signup($signup) {
     $current_run_end = NULL;
   }
 
-  $campaign = node_load($signup['campaign_id']);
+  $campaign = Campaign::get($signup['campaign_id']);
 
   $campaign_data = (new SignupTransformer)->transformCampaign((object) $campaign);
 
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
@@ -137,124 +137,132 @@ function dosomething_rogue_format_activity($response)
 
     // Format each signup
     foreach ($response['data'] as $signup) {
-      // Get campaign and run information for the signup data response
-        $northstar_user = dosomething_northstar_get_user($signup['northstar_id']);
-
-        // Get campaign information
-        $drupal_user = dosomething_user_get_user_by_northstar_id($signup['northstar_id']);
-
-        try {
-          $campaign_run = dosomething_helpers_get_current_campaign_run_for_user($signup['campaign_id'], $drupal_user); // get errors when calling this sometimes (500 - no campaign data found)
-        } catch(Exception $e) {
-          // the regular Phoenix endpoint doesn't seem to handle exceptions like these so we just set the run to null if there is an exception
-          $campaign_run = null;
-        }
-
-        $campaign_run_nid = $campaign_run->nid;
-
-        $current = ($signups['campaign_run_id'] == $campaign_run_nid) ? true : false;
-
-        if (isset($campaign_run->field_run_date['en'][0])) {
-          $current_run_start = $campaign_run->field_run_date['en'][0]['value'];
-          $current_run_end = $campaign_run->field_run_date['en'][0]['value2'];
-        }
-        else {
-          $current_run_start = NULL;
-          $current_run_end = NULL;
-        }
-
-        $campaign = node_load($signup['campaign_id']);
-
-        $campaign_data = (new SignupTransformer)->transformCampaign((object) $campaign);
-
-        $user_data = (new SignupTransformer)->transformUser((object) $northstar_user);
-
-      // Format the data for each signup
-        $signup_data = [
-          // SIGNUP
-          'id' => $signup['signup_id'],
-          'created_at' => strtotime($signup['created_at']),
-
-          // CAMPAIGN RUN
-          'campaign_run' => [
-            'id' => $campaign_run_nid,
-            'current' => $current,
-            'timing' => [
-              'start' => $current_run_start,
-              'end' => $current_run_end,
-            ],
-          ],
-          'uri' => null,
-
-          // CAMPAIGN - no changes
-          'campaign' => $campaign_data,
-
-          // USER - no changes
-          'user' => $user_data,
-        ];
-
-        // REPORTBACK IS CONDITIONAL
-        if (array_key_exists('posts', $signup)) {
-          $reportback_data = [
-              'id' => $signup['signup_id'],
-              'created_at' => strtotime($signup['created_at']),
-              'updated_at' => strtotime($signup['updated_at']),
-              'quantity' => $signup['quantity'],
-              'uri' => null,
-              'why_participated' => $signup['why_participated'],
-              'flagged' => null,
-          ];
-
-          // Add data for each reportback item
-          $reportback_item_array = [];
-          foreach ($signup['posts']['data'] as $post) {
-            $post_data = [
-              'id' => $post['id'],
-              'status' => dosomething_rogue_transform_rogue_status_to_phoenix_status($post['status']),
-              'caption' => $post['media']['caption'],
-              'uri' => null,
-              'media' => [
-                'uri' => $post['media']['url'],
-                'type' => 'image',
-              ],
-              'created_at' => strtotime($post['created_at']),
-              'kudos' => [
-                'data' => [],
-              ],
-            ];
-
-            // Include reaction data if there is any
-            if (count($post['reactions']) > 0) {
-              $post_data['kudos']['data'] = [
-                '641' => [
-                  'term' => [
-                    'id' => '641',
-                    'name' => 'heart',
-                    'total' => count($post['reactions']),
-                  ],
-                  'current_user' => null,
-                ]
-              ];
-            }
-
-            array_push($reportback_item_array, $post_data);
-          }
-
-          // Include total number of items in the response
-          $reportback_data['reportback_items']['total'] = count($reportback_item_array);
-
-          // Add reportback item data to reportback portion of response
-          $reportback_data['reportback_items']['data'] = $reportback_item_array;
-
-          // Add reportback data to the response for this signup
-          $signup_data['reportback'] = $reportback_data;
-        }
+      $signup_data = dosomething_rogue_format_signup($signup);
 
       // Add data to the overall response
       array_push($data['data'], $signup_data);
     }
 
     return json_encode($data);
+  }
+}
+
+/**
+*/
+function dosomething_rogue_format_signup($signup) {
+  // Get campaign and run information for the signup data response
+  $northstar_user = dosomething_northstar_get_user($signup['northstar_id']);
+
+  // Get campaign information
+  $drupal_user = dosomething_user_get_user_by_northstar_id($signup['northstar_id']);
+
+  try {
+    $campaign_run = dosomething_helpers_get_current_campaign_run_for_user($signup['campaign_id'], $drupal_user); // get errors when calling this sometimes (500 - no campaign data found)
+  } catch(Exception $e) {
+    // the regular Phoenix endpoint doesn't seem to handle exceptions like these so we just set the run to null if there is an exception
+    $campaign_run = null;
+  }
+
+  $campaign_run_nid = $campaign_run->nid;
+
+  $current = ($signups['campaign_run_id'] == $campaign_run_nid) ? true : false;
+
+  if (isset($campaign_run->field_run_date['en'][0])) {
+    $current_run_start = $campaign_run->field_run_date['en'][0]['value'];
+    $current_run_end = $campaign_run->field_run_date['en'][0]['value2'];
+  }
+  else {
+    $current_run_start = NULL;
+    $current_run_end = NULL;
+  }
+
+  $campaign = node_load($signup['campaign_id']);
+
+  $campaign_data = (new SignupTransformer)->transformCampaign((object) $campaign);
+
+  $user_data = (new SignupTransformer)->transformUser((object) $northstar_user);
+
+  // Format the data for each signup
+  $signup_data = [
+    // SIGNUP
+    'id' => $signup['signup_id'],
+    'created_at' => strtotime($signup['created_at']),
+
+    // CAMPAIGN RUN
+    'campaign_run' => [
+      'id' => $campaign_run_nid,
+      'current' => $current,
+      'timing' => [
+        'start' => $current_run_start,
+        'end' => $current_run_end,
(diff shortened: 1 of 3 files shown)


katiecrane:
This is looking great! Are the ids actually `***` in the responses? Also (I think we talked about this part in person earlier) just looks like you response just needs to be within `data`. You might want to pull [this line](https://github.com/DoSomething/phoenix/pull/7441/files#diff-75aae572ad7d94de41778845b7683db0R135) into the helper you created too

chloealee:
@katiecrane no - the `***` aren't actually in the responses but just starred those out for double security! Also, updated so the response is wrapped in `data`!

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -258,6 +147,129 @@ function dosomething_rogue_format_activity($response)
...
+  if (isset($campaign_run->field_run_date['en'][0])) {
+    $current_run_start = $campaign_run->field_run_date['en'][0]['value'];
+    $current_run_end = $campaign_run->field_run_date['en'][0]['value2'];
+  }
+  else {
+    $current_run_start = NULL;
+    $current_run_end = NULL;
+  }
+
+  $campaign = Campaign::get($signup['campaign_id']);
```
chloealee: hmm this works as expected for `GET /signups/:sid` but breaks `GET /signups` (when Rogue is on). Here's the error from the logs: 
```
Exception: No campaign data found. in Campaign::get() (line 65 of /var/www/dev.dosomething.org/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php).
```
but when I `dump($campaign);`, the correct campaign object is returned. 
not sure why this might be? cc @katiecrane @sbsmith86 @DFurnes

chloealee: this is very strange because if I do `dump($results);` [here](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php#L63) and then `dump('hi');` right above this [line](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_campaign/includes/Campaign.php#L65), it seems to work sometimes but not other times and getting inconsistent results 🤔

chloealee: update that I'm very confused by! If I `dump($results);` I get the expected response. If I then move on and `dump('hi');`, under `if (! $results)` it seems to also make it there but not sure why since there is a result?

also, if I `dump($results)`  under `if (! $results)`, it is an empty array ... am I doing something really dumb?

chloealee: alternatively, if we don't care about getting [all of the campaign data](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_api/includes/Transformer.php#L205-L273) and [this suffices](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_api/includes/Transformer.php#L192-L201), I don't think anything will break and I can change back to `$campaign = node_load($signup['campaign_id']);` here which works for both `/signups` and `/signups/:sid` proxy

chloealee: just chatted with @aaronschachter and turns out we don't need the full campaign fields from the transformer for Gambit ([here's](https://github.com/DoSomething/phoenix-js/blob/master/lib/phoenix-signup.js) the library that is parsing the Phoenix response) @katiecrane @sbsmith86 @DFurnes would we be okay not returning the [full campaign response](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_api/includes/Transformer.php#L205-L273) and just [this](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_api/includes/Transformer.php#L192-L201) instead? 

The major difference is that we won't be returning some properties nested in the campaign object (`reportback_info`, `tagline`, and `magic_link_copy` and `language`, `translations` and `reportback_info` will be returned in a bit of a different format)

katiecrane: @chloealee what is different about the formatting of those fields?

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -258,6 +147,129 @@ function dosomething_rogue_format_activity($response)
...
+  try {
+    $campaign_run = dosomething_helpers_get_current_campaign_run_for_user($signup['campaign_id'], $drupal_user); // get errors when calling this sometimes (500 - no campaign data found)
+  } catch(Exception $e) {
+    // the regular Phoenix endpoint doesn't seem to handle exceptions like these so we just set the run to null if there is an exception
+    $campaign_run = null;
+  }
+
+  $campaign_run_nid = $campaign_run->nid;
+
+  $current = ($signups['campaign_run_id'] == $campaign_run_nid) ? true : false;
```
sbsmith86: @chloealee I don't think this `$signups` (plural) array is defined in this function.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
sbsmith86: Is `$signup` in fact, just one signup array? Seems like there might be an issue where when you are grabbing just one signup via `GET /signups/:sid` it works because `$signup` is just one signup? but when you are trying to grab a bunch of signups, maybe you are not passing the right thing.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
chloealee: ah thank you for catching that @sbsmith86 ! Updating that now.

lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
chloealee: @sbsmith86 yup - it looks like `$signup` is the same thing when passed via the `dosomething_rogue_format_activity($response);` function and going directly through the `dosomething_rogue_format_signup($signup);` function :-/ 

@katiecrane here's the different formatting:
what we'd have with using `$campaign = node_load($signup['campaign_id']);`
```
            "campaign": {
                "id": "1283",
                "title": "Patient Playbooks GL",
                "campaign_runs": null,
                "language": "en",
                "translations": {
                    "original": "en",
                    "data": {
                        "en": {
                            "entity_type": "node",
                            "entity_id": "1283",
                            "revision_id": "5094",
                            "language": "en",
                            "source": "",
                            "uid": "11",
                            "status": "1",
                            "translate": "0",
                            "created": "1405003569",
                            "changed": "1493835518"
                        },
                        "en-global": {
                            "entity_type": "node",
                            "entity_id": "1283",
                            "revision_id": "5094",
                            "language": "en-global",
                            "source": "en",
                            "uid": "1702694",
                            "status": "0",
                            "translate": "0",
                            "created": "1451926478",
                            "changed": "1485546626"
                        },
                        "es-mx": {
                            "entity_type": "node",
                            "entity_id": "1283",
                            "revision_id": "5094",
                            "language": "es-mx",
                            "source": "en",
                            "uid": "10",
                            "status": "1",
                            "translate": "0",
                            "created": "1456512429",
                            "changed": "1457731544"
                        },
                        "pt-br": {
                            "entity_type": "node",
                            "entity_id": "1283",
                            "revision_id": "5094",
                            "language": "pt-br",
                            "source": "en",
                            "uid": "10",
                            "status": "1",
                            "translate": "0",
                            "created": "1456512404",
                            "changed": "1457731564"
                        }
                    }
                },
                "tagline": null,
                "status": "1",
                "type": "campaign"
            },
```
compared to current Phoenix Ashes response (long version in `transformCampaign($data)`: 
```
"campaign": {
            "id": "886",
            "title": "Kickball for All",
            "campaign_runs": {
                "current": {
                    "en": {
                        "id": "1807"
                    }
                },
                "past": []
            },
            "language": {
                "language_code": "en",
                "prefix": "us"
            },
            "translations": {
                "original": "en",
                "en": {
                    "language_code": "en",
                    "prefix": "us"
                }
            },
            "tagline": "Start an inclusive kickball team in your neighborhood.",
            "status": "active",
            "type": "campaign",
            "reportback_info": {
                "copy": "Upload yours by August 31.",
                "confirmation_message": "Back, back, back, GONE! You knocked this one out of the park.",
                "noun": "Games",
                "verb": "Played"
            },
            "magic_link_copy": null
        },
```

katiecrane: ooooo that seems like a pretty big difference to me for anything that uses this data. maybe @DFurnes knows if this is even used anywhere?

DFurnes: I'm not sure. Do you get more reasonable looking data with `Campaign::get($signup['campaign_id'])`?

DFurnes: If that works for most campaigns, you can wrap the `Campaign::get` in a try-catch to handle the situations where the NID doesn't exist so it doesn't give that nasty error you posted above.

chloealee: @DFurnes by using `Campaign::get($signup['campaign_id])`, the result exactly matches the phoenix result. By just doing `node_load($signup['campaign'])`, it doesn't recognize as a campaign [here](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_api/includes/Transformer.php#L205) so doesn't give the full data.

chloealee: omg @DFurnes you're brilliant!! it worked, thank you!! all set for a final review then :)

sbsmith86 (approved):
Looks great to me.

PR #7442, opened 2017-07-11
Make sure we can call user_logout in the init hook.

DFurnes:
#### What's this PR do?
This should fix the issue that a few people mentioned where `user_logout` was undefined. It turns out you need to "include" the user module before you're able to call it in the init hook, okay.

screen shot 2017-07-11 at 3 09 13 pm

👀

Here's the linked [StackOverflow post](https://drupal.stackexchange.com/a/188886).

#### Relevant tickets
Fixes 🐛.

#### Checklist
- [ ] Tested on staging.


sbsmith86 (approved):
Oh fun! yea I just ran into this with all the warnings the rogue module was throwing! 👍

PR #7446, opened 2017-07-13
Fix RB Emails when Rogue is on

sbsmith86:
#### What's this PR do?

This PR fixes a bug where the "We Received Your RB" Email was not being sent when rogue was turned on. This was because we were bypassing the `ReportbackController::save()` method when rogue was turned on and that method made the call to `dosomething_reportback_mbp_request()` that sent the message. 

However `dosomething_reportback_mbp_request()` took a reportback entity as a param and when we are working with Rogue posts the data is in a different format. So in this PR I created a new function, `dosomething_reportback_rogue_mbp_request()`, that accepts a Rogue post and signup object, formats the params, and makes the call to message broker. 

We will have to make sure we are getting the email on the first submission of a reportback and it is looking good. 

#### Relevant tickets
https://trello.com/c/WHWZhku8/492-5-%F0%9F%90%9B-when-rogue-is-turned-on-and-after-you-reportback-it-doesnt-seem-like-the-we-got-your-rb-email-triggers-%F0%9F%90%9B

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -1068,64 +1068,59 @@ function dosomething_reportback_mbp_request($entity) {
  * Sends mbp request for a reportback stored in Rogue.
  */
 function dosomething_reportback_rogue_mbp_request($rb) {
-  // dump($rb);
-  // die();
-  // This fid is the last file uploaded to the reportback.
-  // @see Reportback->save().
-  // $fid = $rb['id'];
-
-  // @TODO - Figure this out.
-  // $image_markup = dosomething_image_get_themed_image(variable_get('dosomething_campaign_permalink_failure_image_nid'), 'square', '480x480');
-
-  // dump($image_markup);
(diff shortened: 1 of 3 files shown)


katiecrane (approved):
This looks like it'll work to me!

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -1064,6 +1065,62 @@ function dosomething_reportback_mbp_request($entity) {
...
+  if (module_exists('dosomething_northstar')) {
+    $northstar_user = dosomething_northstar_get_user($rb['northstar_id']);
+    $account = user_load($northstar_user->drupal_id);
+    $node = node_load($signup['campaign_id']);
+    $language = dosomething_global_get_language($account, $node);
+    $campaign_language = !empty($node->language) ? $node->language : DOSOMETHING_GLOBAL_DEFAULT_LANG_CODE;
+    $campaign_country = dosomething_global_get_current_country_code();
+
+    if ($node) {
+      $wrapper = entity_metadata_wrapper('node', $node);
```
DFurnes, line 1084: Hissss! 😾

DFurnes (approved):
This looks good to me!

PR #7448, opened 2017-07-17
Move function definition to top level of code

sbsmith86:
#### What's this PR do?

I think there might have been a weird rebase issue where a new helper function moved into another function which was causing it not to be recognized. Or...i just messed up? This fixes it. Also fixes another stupid bug I introduced where I didn't use the `$` to denote the correct signup variable. 

Go to a campaign, signup, reportback for the first time, you should be taken to the permalink page and you should receive an email.

#### Checklist
- [ ] Tested on staging.


katiecrane (approved):
a-ha! looks much better

PR #7450, opened 2017-07-17
Move rogue signup logic into the single function that handles signups

sbsmith86:
#### What's this PR do?

This moves the logic of checking if rogue is on and sending to rogue instead of storing in phoenix to the `dosomething_user_signup_user` function. It turns out this is called in a [few places](https://github.com/DoSomething/phoenix/search?utf8=%E2%9C%93&q=dosomething_signup_user_signup&type=). And we could have been missing some funnels where the user wasn't getting sent to rogue and still getting stored in phoenix. This should clean that up by moving all the logic into the same handler. 

We were seeing and issue on thor where users were not being sent to the pitch page after signing up. This seemed to only occur on thor, but hoping this will clean up some of the logic so that the issue is either fixed, or allows for easier debugging. 

#### Relevant tickets
Addresses: https://trello.com/c/Jy2dPJoh/517-when-rogue-is-on-an-already-authenticated-user-is-not-being-redirected-to-the-pitch-page

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
@@ -97,18 +97,21 @@ function dosomething_signup_form_submit($form, &$form_state) {
     $params['affiliate_messaging_opt_in'] = (bool) $form_state['values']['affiliate_messaging_opt_in'];
   }
 
-  if (variable_get('rogue_collection', FALSE)) {
-    $rogue_signup = dosomething_rogue_send_signup_to_rogue($form_state['values']);
+  // Create signup for logged in user.
+  dosomething_signup_user_signup($nid, NULL, $source, $params, $form_state['values']);
 
-    // Make sure the signup exists before using it
-    if ($rogue_signup) {
-      $node = node_load($rogue_signup['data']['campaign_id']);
-      dosomething_signup_set_signup_message($node->title);
-    }
-  } else {
-    // Create signup for logged in user.
-    dosomething_signup_user_signup($nid, NULL, $source, $params);
-  }
+  // if (variable_get('rogue_collection', FALSE)) {
+  //   $rogue_signup = dosomething_rogue_send_signup_to_rogue($form_state['values']);
+
+  //   // Make sure the signup exists before using it
+  //   if ($rogue_signup) {
+  //     $node = node_load($rogue_signup['data']['campaign_id']);
+  //     dosomething_signup_set_signup_message($node->title);
+  //   }
+  // } else {
+  //   // Create signup for logged in user.
+  //   dosomething_signup_user_signup($nid, NULL, $source, $params);
+  // }
 }
 
 /**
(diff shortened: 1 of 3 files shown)


lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
```diff
@@ -97,18 +97,21 @@ function dosomething_signup_form_submit($form, &$form_state) {
...
-      dosomething_signup_set_signup_message($node->title);
-    }
-  } else {
-    // Create signup for logged in user.
-    dosomething_signup_user_signup($nid, NULL, $source, $params);
-  }
+  // Create signup for logged in user.
+  dosomething_signup_user_signup($nid, NULL, $source, $params, $form_state['values']);
+
+  // if (variable_get('rogue_collection', FALSE)) {
```
DFurnes: Should we remove these lines now? 👾

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -579,25 +579,38 @@ function dosomething_signup_user_signup($nid, $account = NULL, $source = NULL, $
     $account = $user;
   }
 
-  // Insert signup.
-  if ($sid = dosomething_signup_create($nid, $account->uid, $source)) {
-    $node = node_load($nid);
+  if (variable_get('rogue_collection', FALSE)) {
+    $rogue_signup = dosomething_rogue_send_signup_to_rogue([
```
DFurnes: Should we pass the `$account` variable here as the second parameter?

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -579,25 +579,38 @@ function dosomething_signup_user_signup($nid, $account = NULL, $source = NULL, $
...
+      'campaign_id' => $nid,
+      'source' => $source,
+    ]);
+
+    // Make sure the signup exists before using it
+    if ($rogue_signup) {
+      $node = $node = node_load($nid);
```
DFurnes: `$node = $node = ` 👾

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
```diff
@@ -579,25 +579,38 @@ function dosomething_signup_user_signup($nid, $account = NULL, $source = NULL, $
...
+    }
+  } else {
+    // Insert signup.
+    if ($sid = dosomething_signup_create($nid, $account->uid, $source)) {
+      $node = node_load($nid);
 
-    // Opt into competition if auto opt-in is enabled for this campaign.
-    $auto_opt_in = dosomething_helpers_get_variable('node', $nid, 'enable_auto_competition_opt_in');
+      // Opt into competition if auto opt-in is enabled for this campaign.
+      $auto_opt_in = dosomething_helpers_get_variable('node', $nid, 'enable_auto_competition_opt_in');
```
DFurnes: Should these lines be pulled out of the feature flag conditional (since competition & flash message should happen regardless)?

lib/modules/dosomething/dosomething_signup/dosomething_signup.module
sbsmith86: whoa

sbsmith86:
@DFurnes updated!

DFurnes (approved):
This looks good to me! ⛵️

sbsmith86:
https://trello.com/c/Jy2dPJoh/517-when-rogue-is-on-an-already-authenticated-user-is-not-being-redirected-to-the-pitch-page

PR #7452, opened 2017-07-17
Move Rogue feature flag check into signup module.

DFurnes:
#### What's this PR do?
This pull request moves the Rogue feature flags into the `dosomething_signup_create` and `dosomething_signup_exists` methods, since otherwise we had to check the feature flag in multiple places around the codebase which let some things fall through the cracks!

I roughly broke it into commits, but it's a lil' messy. Sorry in advance. 😬

This should mean we don't end up with surprise Phoenix signup records anymore.

#### Relevant tickets
Fixes 🐛.

#### Checklist
- [ ] Tested on staging.


sbsmith86 (approved):
looks great

PR #7453, opened 2017-07-18
Ensure we send transactionals for Rogue too.

DFurnes:
#### What's this PR do?
This pull request ensures we send transactional messages for Rogue signups – previously this was being called in the entity save hook, which would not fire for Rogue signups.

👀

I don't think we had a card for this, but it seemed like a quick fix.

#### Relevant tickets
Fixes 🐛.

#### Checklist
- [ ] Tested on staging.


sbsmith86 (approved):
Is this support for the truth opt-out/in?

DFurnes:
This will allow transactional messages to work the same as they do for traditional signups – I think they were sporadically working before because we were still creating Drupal entities, but after tidying up the `signup_create` function this needed to be pulled out. I'll check in on Truth opt-in, but that should now be working since it's [outside of the feature flag](https://github.com/DoSomething/phoenix/blob/994f5357e6939c1efc4ef42c1f654403ebb0689a/lib/modules/dosomething/dosomething_signup/dosomething_signup.module#L635-L639)!

PR #7454, opened 2017-07-18
Competitions prompt

sbsmith86:
#### What's this PR do?

This PR fixes the competitions prompt when rogue is turned on. 

* The `$config` variable that pulls the signup data form configuration variables (i.e. submit button text, form copy) was dependent on a phoenix `$signup` object. When rogue is on, the `$signup` object is different so this PR updates the logic to work with phoenix or rogue signup objects. 

* The decision to show the modal or not was tied to a `signup_data_form_timestamp` and a `signup_data_form_response` column on the signup. When rogue is on, we do not store these columns on the signup. The PR removes any dependencies on these columns when rogue is on. Instead we always show the competitions prompt unless we find the user in gladiator directly.

* This PR also updates the `signup_load()` function to return a `$signup` from rogue if rogue is on.

We made a decision to only support the competition prompt for Rogue's launch on 7/20/17. The signup data form supports the competition prompt, address collection, and the social share prompt. As a result, the code is pretty entangled and we need to do some work to separate some of the dependencies on the signup data form properties that we were stored on the signup. We will do this work after launch

#### Relevant tickets
https://trello.com/c/vGmYYcDa/511-%F0%9F%90%9B-competitions-prompt-not-working-when-rogue-is-turned-on%F0%9F%90%9B

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -762,8 +762,12 @@ function dosomething_campaign_add_signup_data_form_vars(&$node, $langcode) {
   $node->content['signup_data_form'] = drupal_get_form('dosomething_signup_user_signup_data_form', $signup);
 
   // If the signup_data_form is required and user has not submitted form yet:
-  $hasCompetitionRecordedLocally = $signup->signup_data_form_timestamp;
-  $shouldShowCompetitionPrompt = $config['required'] && !$hasCompetitionRecordedLocally;
+  if (!variable_get('rogue_collection', FALSE)) {
+    $hasCompetitionRecordedLocally = $signup->signup_data_form_timestamp;
+    $shouldShowCompetitionPrompt = $config['required'] && !$hasCompetitionRecordedLocally;
+  } else {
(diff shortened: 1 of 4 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -757,11 +757,17 @@ function dosomething_campaign_add_signup_data_form_vars(&$node, $langcode) {
...
   // Load the signup entity.
   $signup = signup_load($sid);
+
   // Pass to the user signup data form.
   $node->content['signup_data_form'] = drupal_get_form('dosomething_signup_user_signup_data_form', $signup);
+
   // If the signup_data_form is required and user has not submitted form yet:
-  $hasCompetitionRecordedLocally = $signup->signup_data_form_timestamp;
-  $shouldShowCompetitionPrompt = $config['required'] && !$hasCompetitionRecordedLocally;
+  if (!variable_get('rogue_collection', FALSE)) {
```
DFurnes, line 765: I'd be up for just always doing the "if Rogue is enabled" thing now, since we're gonna enable it so soon.

DFurnes (approved):

PR #7460, opened 2017-07-19
Return Post ID from Rogue when Creating an RB

sbsmith86:
#### What's this PR do?

Updates the return logic for the `/api/v1/campaigns/[nid]/reportback` Rogue proxy to return the `post_id` from rogue instead of storing anything in the reference table (which we don't use).

👀 

Reportbacks were successfully getting into rogue, but we were returning null because we don't use the reference table anymore.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -378,10 +378,7 @@ function _campaign_resource_reportback($nid, $values) {
   if (variable_get('rogue_collection', FALSE) && ! isset($headers['X-Request-Id'])) {
     $rogue_reportback = dosomething_rogue_send_reportback_to_rogue($values, $user);
 
-    // Make sure reportback made it back to Phoenix and store reference to the Rogue item if one was added.
-    $rb_info = dosomething_rogue_check_rbid_and_store_ref($rogue_reportback);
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -378,10 +378,7 @@ function _campaign_resource_reportback($nid, $values) {
   if (variable_get('rogue_collection', FALSE) && ! isset($headers['X-Request-Id'])) {
     $rogue_reportback = dosomething_rogue_send_reportback_to_rogue($values, $user);
 
-    // Make sure reportback made it back to Phoenix and store reference to the Rogue item if one was added.
-    $rb_info = dosomething_rogue_check_rbid_and_store_ref($rogue_reportback);
-
-    return $rb_info['rbid'];
+    return $rogue_reportback['data']['id']
```
DFurnes: Do we need to check for any error states here?

DFurnes (approved):

sbsmith86:
@DFurnes updated to return `FALSE` if we don't get anything back from rogue. The `dosomething_rogue_send_reportback_to_rogue()` function does a lot of error handling, including storing things in a failure table if it doesn't make it to rogue.

DFurnes:
Great! 👏

PR #7461, opened 2017-07-19
Always send Rogue northstar ids and not drupal ones

katiecrane:
#### What's this PR do?
When Rogue is on and you hit `GET /api/v1/signups`, requests can now include drupal ids or northstar ids or both! We go through all the ids given, swap out northstar for drupal where needed, and then put it all back together to pass over to Rogue.

Try a request like this `http://dev.dosomething.org:8888/api/v1/signups?users=1707996,1705523`

Want to make sure all of this is supported when we turn Rogue on.

#### Checklist
- [x] Tested on staging.


sbsmith86 (approved):
💯

PR #7466, opened 2017-07-20
Fix silly error to use the right index

katiecrane:
#### What's this PR do?
If given a `northstar_id` we reset it to live at `$values['uid']` and not ``$values['northstar_id']`, so use `$values['uid']` going forward. This fixes [this endpoint](https://github.com/DoSomething/phoenix/blob/dev/documentation/endpoints/campaigns.md#campaign-signup).

:eyes:

I think we made this error somewhere along the way (sing that part like [the Dawes song](https://open.spotify.com/track/01HHVuCx2hBQLl2aZExNAc)) yesterday.

#### Checklist
- [ ] Tested on staging.


DFurnes (approved):
👍 👍

sbsmith86:
Woot!

angaither:
niiiiice

PR #7468, opened 2017-07-20
Use the given user

katiecrane:
#### What's this PR do?
We were always using `global $user` but we want to use the user we were given instead.

👀 

#### Checklist
- [ ] Tested on staging.


sbsmith86 (approved):
💯

PR #7469, opened 2017-07-20
fixes NaN bug for first 6 photos of the gallery

chloealee:
#### What's this PR do?
Fixes reaction counter bug for first 6 photos of the gallery that should `NaN` in the counter.

Counter should show accurate number of reactions instead of `NaN` for first 6 photos in the gallery.

My local is a little borked and won't let me do view more. I'll double check that this doesn't break anything for the rest of the gallery but @ngjo mentioned she only saw this for the first 6 so hoping it doesn't!

#### Checklist
- [ ] Tested on staging.


DFurnes (approved):

katiecrane:
Does this still work if Rogue is off?

chloealee:
@katiecrane yes!

PR #7471, opened 2017-07-20
Use a compound post_id|northstar_id format for kid.

DFurnes:
#### What's this PR do?
This pull request updates Phoenix's `POST /api/v1/kudos` and `DELETE /api/v1/kudos/:kid` endpoints so that we can easily delete a reaction in Rogue. This solves a little conundrum where we needed the Post ID & Northstar ID in the "delete" endpoint, but were only getting the Post ID. Now they're both provided in a `<post_id>|<northstar_id>` formatted string.

👀

Here's the corresponding Rogue PR: DoSomething/rogue#316.

#### Relevant tickets
Fixes 🐞

#### Checklist
- [ ] Tested on staging.


sbsmith86 (approved):
💯

PR #7476, opened 2017-07-21
Return null reportback property if no rb exists

sbsmith86:
#### What's this PR do?

Fixes a bug in the rogue proxy response for `GET api/v1/signups/:id` 

* We were checking if `array_key_exists('posts', $signup)` to determine if there were posts on the signup and if we should return a reportback property. Turns out rogue always returns the `posts` property whether it is empty or not (should we fix this?). So this PR updates the check to see if posts are, in fact, empty.

* In the old phoenix response, `reportback: null` was returned in the response when a reportback didn't exists. This updates the proxy response to reflect that. 

👀 

Gambit was relying on the response being in this format. Without this fix, it thinks that all signups have reportbacks under them and are giving users the wrong flow.

#### Relevant tickets
Fixes 🐛

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
@@ -200,7 +200,7 @@ function dosomething_rogue_format_signup($signup) {
   ];
 
   // REPORTBACK IS CONDITIONAL
-  if (array_key_exists('posts', $signup)) {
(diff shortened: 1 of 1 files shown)


DFurnes (approved):

DFurnes:
> Turns out rogue always returns the posts property whether it is empty or not (should we fix this?).

I think that makes sense for Rogue to return `posts` as an empty array! 👍

PR #7477, opened 2017-07-23
User profile powered by Rogue 

chloealee:
#### What's this PR do?
- Updates user profile page to be powered by Rogue when Rogue is on. 

- Go to user profile page.
- You should see all of the campaigns you're doing and have done with data from Rogue.
- Make sure reactions work. **You should be able to add a reaction to the user profile and it will show up in Rogue. However, when you "unlike" something, it doesn't seem to send to Rogue correctly and I think this is a larger bug (tried in the gallery on the campaign page and saw same behavior). 

#### Relevant tickets
Fixes https://www.pivotaltracker.com/n/projects/2019429/stories/149207077

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -519,7 +519,38 @@ function dosomething_campaign_get_campaigns_doing($uid = NULL) {
   }
 
   // Get all signups.
-  $signup_nids = dosomething_signup_get_signup_nids_by_uid($uid);
+  if (variable_get('rogue_collection', FALSE)) {
+    $northstar_id = dosomething_user_get_northstar_id($uid);
+
+    $rogue_signups = dosomething_rogue_get_activity([
+      'filter' => [
+        'northstar_id' => $northstar_id,
+      ],
+      'limit' => 300,
+    ]);
+
+    // Filter through response and put all signup ids with no posts in an array
+    $signup_nids = [];
+
+    foreach ($rogue_signups['data'] as $signup) {
+      if ($signup['posts']['data']) {
+        continue;
+      } else {
+        array_push($signup_nids, $signup['campaign_id']);
+      }
+    }
+  }
+  else {
+    $signup_nids = dosomething_signup_get_signup_nids_by_uid($uid);
+
+    // Remove signups that have already been reported back for.
+    foreach ($signup_nids as $delta => $nid) {
+      if (dosomething_reportback_exists($nid, NULL, $uid)) {
+        unset($signup_nids[$delta]);
+      }
+    }
+  }
+
   if (empty($signup_nids)) {
     return [];
   }
@@ -527,11 +558,6 @@ function dosomething_campaign_get_campaigns_doing($uid = NULL) {
   $doing = [];
 
   foreach ($signup_nids as $delta => $nid) {
-    // If they reported back, they already did it.
-    if (dosomething_reportback_exists($nid, NULL, $uid)) {
-      continue;
-    }
-
     $node = node_load($nid);
 
     // If the campaign is closed, they're not doing it anymore.
--- a/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
+++ b/lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
@@ -526,7 +526,6 @@ function dosomething_campaign_get_campaigns_doing($uid = NULL) {
       'filter' => [
         'northstar_id' => $northstar_id,
       ],
-      'limit' => 300,
     ]);
 
     // Filter through response and put all signup ids with no posts in an array
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
@@ -1145,6 +1145,8 @@ function dosomething_reportback_get_reportbacks($uid = NULL) {
     $uid = $user->uid;
   }
 
+  // If Rogue is on, return array of themed reportbacks as done in https://github.com/DoSomething/phoenix/blob/76164b2006fb5a266e8166688472742556eae33e/lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_user.inc#L32-L51 for user profile page.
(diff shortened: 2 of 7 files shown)


lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -519,19 +519,45 @@ function dosomething_campaign_get_campaigns_doing($uid = NULL) {
...
   // Get all signups.
-  $signup_nids = dosomething_signup_get_signup_nids_by_uid($uid);
+  if (variable_get('rogue_collection', FALSE)) {
+    $northstar_id = dosomething_user_get_northstar_id($uid);
+
+    $rogue_signups = dosomething_rogue_get_activity([
+      'filter' => [
+        'northstar_id' => $northstar_id,
+      ],
+      'limit' => 300,
```
chloealee: Chose to do `limit` query param here instead of using paginated links to only make one API call each time. There are less than 300 campaigns so chose this number for now as a user can only sign up for a campaign once.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -1144,12 +1144,59 @@ function dosomething_reportback_get_reportbacks($uid = NULL) {
...
+    foreach ($response['data'] as $delta => $reportback) {
+      $reportback = (object) $reportback;
+
+      $total_kudos = $reportback->kudos['data'][0]['term']['total'];
+      $liked_by_current_user = $reportback->kudos['data'][0]['current_user']['reacted'] ? TRUE : FALSE;
+
+      $campaign = node_load($reportback->campaign['id']);
+
+      $impact = $reportback->reportback['quantity'] . ' ' . dosomething_helpers_extract_field_data($campaign->field_reportback_noun) . ' ' . dosomething_helpers_extract_field_data($campaign->field_reportback_verb);
+
```
chloealee, line 1168: note: this won't be a link like it is when Rogue is turned off because it links to `/reportback/:rbid` which doesn't work when Rogue is on

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -519,19 +519,45 @@ function dosomething_campaign_get_campaigns_doing($uid = NULL) {
   }
 
   // Get all signups.
-  $signup_nids = dosomething_signup_get_signup_nids_by_uid($uid);
```
katiecrane, line 522: If we're removing the old functionality do we still want to check for the feature flag?

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
katiecrane: Does this default to 25 or something like that if you don't include the limit?

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
katiecrane: oooo I see that you moved this into the else now I take it back!

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
chloealee: @katiecrane ohhh you're right ! I don't need to add this limit here - it works automatically to format and the view more button works too!

lib/modules/dosomething/dosomething_campaign/dosomething_campaign.module
```diff
@@ -519,19 +519,44 @@ function dosomething_campaign_get_campaigns_doing($uid = NULL) {
...
+      'filter' => [
+        'northstar_id' => $northstar_id,
+      ],
+    ]);
+
+    // Filter through response and put all signup ids with no posts in an array
+    $signup_nids = [];
+
+    foreach ($rogue_signups['data'] as $signup) {
+      if ($signup['posts']['data']) {
```
katiecrane, line 535: Just to confirm, Rogue doesn't include `posts` as part of the array unless there are posts?

chloealee: yes! it will just be an empty array

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
```diff
@@ -1144,12 +1144,59 @@ function dosomething_reportback_get_reportbacks($uid = NULL) {
...
+    $reportback_items = [];
+
+    $response = dosomething_rogue_get_reportback_items([
+      'as_user' => $northstar_id,
+    ]);
+
+    foreach ($response['data'] as $delta => $reportback) {
+      $reportback = (object) $reportback;
+
+      $total_kudos = $reportback->kudos['data'][0]['term']['total'];
```
katiecrane, line 1162: based on the [Rogue documentation](https://github.com/DoSomething/rogue/blob/master/documentation/endpoints/reportbacks.md) it looks like this lives at `kudos['total']`, are those docs out of date or am I looking at this wrong?

chloealee: @katiecrane yes the documentation is out of date (I have a note to update this in my notebook!)

katiecrane (approved):
I feel like the unliking stuff was something we found on launch day but I kind of thought it got fixed that day? Looks good to me aside from that though

chloealee:
@katiecrane yeah on Thor it seems to be working actually so I'll merge this and test and then confirm either way if it is an actual issue!

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
katiecrane: CHLOE 🙌  🙌  🙌

chloealee:
for documentation - reactions are working on thor - looks like it was just an issue on my local!

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.module
chloealee: updates in [this PR](https://github.com/DoSomething/rogue/pull/322) !

PR #7478, opened 2017-07-25
Return the rogue rejected status as flagged for phoenix

sbsmith86:
#### What's this PR do?

When transforming rogue posts to respond like phoenix reportback items, we were still setting an items status as `rejected` if it wasn't pending or approved. `rejected` is not a recognized phoenix status so this translates `rejected` statuses to `excluded`. 

👀 

Would particularly like the input of @katiecrane and @DFurnes . I don't think this will break anything else but you might have some insights into any red flags. 

This fixes a bug in gladiator where it is doing work expecting the old phoenix statuses. 

#### Relevant tickets
Fixes 🐛

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
+++ b/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
@@ -435,7 +435,7 @@ function dosomething_rogue_transform_rogue_status_to_phoenix_status($status) {
     return 'pending';
   }
   else {
-    return 'rejected';
+    return 'excluded';
   }
 }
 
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module
```diff
@@ -435,7 +435,7 @@ function dosomething_rogue_transform_rogue_status_to_phoenix_status($status) {
     return 'pending';
   }
   else {
-    return 'rejected';
+    return 'excluded';
```
katiecrane: Greatttt catch, this was definitely an error! I think it should be `flagged` instead of `excluded` though. The only things that are `rejected` in Rogue (that were reviewed in Phoenix originally) are things that were `flagged`. Based on [this function](https://github.com/DoSomething/phoenix/blob/fdaaee7c2306425e653c8c5708e7903b35d0720c/lib/modules/dosomething/dosomething_rogue/dosomething_rogue.module#L413-L423)

sbsmith86:
@katiecrane dope! updated that to `flagged`. Can i get some final 👀

katiecrane (approved):
sounds good to me 🎏

PR #7487, opened 2017-08-08
Tagger Pupper script

katiecrane:
#### What's this PR do?

1. Adds `getPosts` to `Rogue.php` so the tag script can see if the current post already has the current tag. We want to make sure we don't send the tag request if the tag already lives on that post because that will untag it!
2. Adds `postTag` to `Rogue.php` so we can apply tags in Rogue.
3. The script! [Here is all the documentation!](https://docs.google.com/a/dosomething.org/document/d/1mPmQcwbfkgkhlfWJqLM0T7lXrFZvzv3BwSy-oz_QG6w/edit?usp=sharing) Feel free to leave comments on the Google Doc or in the code.

Check the logic. Make sure I didn't say `promoted` where I meant `excluded` or vice versa.

We're doing this so Rogue will be able to show promoted/Good Photo items first in the gallery and will not show excluded/Hide In Gallery items. 

#### Relevant tickets
[Pivotal card](https://www.pivotaltracker.com/n/projects/2019429/stories/149207697)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/scripts/tag-as-excluded-promoted-in-rogue.php
+++ b/scripts/tag-as-excluded-promoted-in-rogue.php
@@ -32,26 +32,27 @@ else {
 }
 
 // Get all the excluded posts to send
-// if ($last_excluded) {
-// 	$excluded_posts_in_rogue = db_query("SELECT r.rogue_post_id, rb.uid, rb.nid
-// 										FROM dosomething_reportback_file f
-// 										JOIN dosomething_rogue_reportbacks r ON f.fid = r.fid
-// 										JOIN dosomething_reportback rb ON f.rbid = rb.rbid
-// 										WHERE f.status = \"excluded\"
-// 										AND r.rogue_post_id > $last_excluded
-// 										ORDER BY r.rogue_post_id");
-// }
-// else {
-// 	$excluded_posts_in_rogue = db_query("SELECT r.rogue_post_id, rb.uid, rb.nid
-// 										FROM dosomething_reportback_file f
-// 										JOIN dosomething_rogue_reportbacks r ON f.fid = r.fid
-// 										JOIN dosomething_reportback rb ON f.rbid = rb.rbid
-// 										WHERE f.status = \"excluded\"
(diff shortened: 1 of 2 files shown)


scripts/tag-as-excluded-promoted-in-rogue.php
```diff
@@ -0,0 +1,195 @@
...
+										JOIN dosomething_rogue_reportbacks r ON f.fid = r.fid
+										JOIN dosomething_reportback rb ON f.rbid = rb.rbid
+										WHERE f.status = 'excluded'
+										ORDER BY r.rogue_post_id");
+}
+
+$client = dosomething_rogue_client();
+
+// Tag promoted posts
+foreach ($promoted_posts_in_rogue as $promoted) {
```
DFurnes, line 56: It looks like each of these steps are very similar, could we have a `tag_posts($posts, 'good-photo')` method that handles the looping and querying logic for them all?

DFurnes (approved):
One tiny suggestion & some indentation looks a little wonky, but totally good to merge IMO!

scripts/tag-as-excluded-promoted-in-rogue.php
katiecrane: @DFurnes I thought about this, but I wasn't sure if it would be worth it for something that will only be used in this script. Or did you mean have the new method live in this file too?

DFurnes: Yeah, definitely can live in this file (if it is something you want to do).

katiecrane: @DFurnes This file will only ever be run a couple of times so I think I'm going to leave it as is if that's cool, just don't want to spend more time on it than necessary

DFurnes: Yeah, fine by me!

PR #7489, opened 2017-08-10
Fixing path alias on confirmation

itsjoekent:
#### What's this PR do?
The url for campaigns on the reportback confirmation had a double `/us` in them because of the drupal function screwing stuff up. just removed it and made the anchor myself. this is what we did for the image tag above, not sure why we didn't do that as well below. 

go to the confirmation page and check the link works.


weerd (commented):
👍

DFurnes (approved):

weerd (approved):

PR #7491, opened 2017-08-16
Update API endpoints to use Northstar ID instead of Drupal ID

chloealee:
#### What's this PR do?
Updates `/campaigns/:id/signup` and `/campaigns/:id/reportback` to use the user's Northstar ID instead of Drupal UID in logic in case a user doesn't have a Drupal UID. 

Sign up for a campaign via `/campaigns/:id/signup` and it should be reflected in Rogue.
Reportback for a campaign via `campaigns/:id/reportback` and it should be reflected in Rogue. 

- @DFurnes this should squash the bug and not introduce any problems assuming everyone has a Northstar ID?
- Did I miss any other endpoints? 

#### Relevant tickets
Fixes https://github.com/DoSomething/gambit/issues/922
https://www.pivotaltracker.com/n/projects/2019429/stories/150138796

cc @aaronschachter 
#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -261,7 +261,7 @@ function _campaign_resource_index($ids, $staff_pick, $term_ids, $count, $random,
  *   The Node nid to post the signup to.
  * @param array $values
  *   The signup data to post. Expected keys:
- *   - northstar_id: The user's Northstar ID uid (string).
+ *   - northstar_id: The user's Northstar ID (string).
  *   - source (string).
  *   - transactionals (bool). Optional flag to disable sending transactional messages.
  *       Defaults to sending messages.
@@ -303,7 +303,7 @@ function _campaign_resource_signup($nid, $values) {
  *   The Node nid to post the reportback to.
  * @param array $values
  *   The reportback data to post. Expected keys:
- *   - uid: The user uid (int).
+ *   - northstar_id: The user's Northstar ID (string).
  *   - file: Base64 encoded file string to save (required if using 'filename').
  *   - filename: The filename of the file provided as file (required if using 'file').
  *   - file_url: The URL of the reportback file to save (used if no 'file'/'filename' exist).
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -261,7 +261,7 @@ function _campaign_resource_index($ids, $staff_pick, $term_ids, $count, $random,
  *   The Node nid to post the signup to.
  * @param array $values
  *   The signup data to post. Expected keys:
- *   - uid: The user uid (int).  Optional, uses global $user if not set.
+ *   - northstar_id: The user's Northstar ID uid (string).
  *   - source (string).
  *   - transactionals (bool). Optional flag to disable sending transactional messages.
  *       Defaults to sending messages.
@@ -274,8 +274,8 @@ function _campaign_resource_signup($nid, $values) {
     unset($values['northstar_id']);
   }
 
-  // If we pass either a Northstar ID or Drupal UID to `uid`, convert it to a real UID.
-  $values['uid'] = dosomething_user_convert_to_legacy_id($values['uid']);
+  // If a Drupal UID is passed to `uid`, convert it to the user's Northstar ID.
+  $values['uid'] = dosomething_user_convert_to_northstar_id($values['uid']);
 
   if (!isset($values['uid'])) {
     return services_error('Cannot create signup without a `northstar_id` or `uid`.');
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -327,14 +327,14 @@ function _campaign_resource_reportback($nid, $values) {
     unset($values['northstar_id']);
   }
 
-  // If we pass either a Northstar ID or Drupal UID to `uid`, convert it to a real UID.
-  $uid = dosomething_user_convert_to_legacy_id($values['uid']);
+  // If a Drupal UID is passed to `uid`, convert it to the user's Northstar ID.
+  $uid = dosomething_user_convert_to_northstar_id($values['uid']);
 
   if (empty($uid)) {
     return services_error('Cannot create reportback without a `northstar_id` or `uid`.');
   }
 
-  $user = user_load($uid);
+  $user = dosomething_user_get_user_by_northstar_id($uid);
 
   if (!$user) {
     return dosomething_helpers_basic_http_response(404, 'The specified user was not found.');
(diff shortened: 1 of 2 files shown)


DFurnes:
Ah, so there's an extra piece here that wasn't really documented in the ticket: we want this endpoint to work even if the user doesn't have a Drupal ID at all, so we need to bypass any code that relies on a `uid` or loaded `$user` object if a Northstar ID is provided. ↪️

chloealee:
Thanks for clarifying, @DFurnes ! I think this will work though even if the user doesn't have a Drupal ID at all, assuming all users have a Northstar ID? Because the Northstar ID will get passed each time and be used. The places where a loaded `$user` object is loaded is the DS API user in these cases (I'm pretty sure)?

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -274,8 +274,8 @@ function _campaign_resource_signup($nid, $values) {
     unset($values['northstar_id']);
   }
 
-  // If we pass either a Northstar ID or Drupal UID to `uid`, convert it to a real UID.
-  $values['uid'] = dosomething_user_convert_to_legacy_id($values['uid']);
+  // If a Drupal UID is passed to `uid`, convert it to the user's Northstar ID.
+  $values['uid'] = dosomething_user_convert_to_northstar_id($values['uid']);
```
DFurnes, line 278: I'm not sure this will work as is, since [`dosomething_signup_create`](https://github.com/DoSomething/phoenix/blob/6e80ae77499c1fd8b30508c1bad7ab6e8dd6d192/lib/modules/dosomething/dosomething_signup/dosomething_signup.module#L255) expects the value provided in the `$uid` parameter here to be a UID stored in the database and [passes it along to `user_load`](https://github.com/DoSomething/phoenix/blob/6e80ae77499c1fd8b30508c1bad7ab6e8dd6d192/lib/modules/dosomething/dosomething_signup/dosomething_signup.module#L268).

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
```diff
@@ -327,14 +327,14 @@ function _campaign_resource_reportback($nid, $values) {
...
-  $uid = dosomething_user_convert_to_legacy_id($values['uid']);
+  // If a Drupal UID is passed to `uid`, convert it to the user's Northstar ID.
+  $uid = dosomething_user_convert_to_northstar_id($values['uid']);
 
   if (empty($uid)) {
     return services_error('Cannot create reportback without a `northstar_id` or `uid`.');
   }
 
-  $user = user_load($uid);
+  $user = dosomething_user_get_user_by_northstar_id($uid);
```
DFurnes, line 337: Unfortunately [this function](https://github.com/DoSomething/phoenix/blob/21792f5658e5542dde85e9e5bd2f7fbcb24022f9/lib/modules/dosomething/dosomething_user/dosomething_user.module#L971-L989) relies on the local `field_data_field_northstar_id` table, which won't be populated if the user doesn't have a Drupal ID. We want to make sure that this whole endpoint works okay without ever touching the local users table.

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
chloealee: @DFurnes Since we have the `northstar_id` of the user (assuming everyone has it), would it work to send the `northstar_id` directly to [`dosomething_rogue_send_signup_to_rogue`](https://github.com/DoSomething/phoenix/blob/6e80ae77499c1fd8b30508c1bad7ab6e8dd6d192/lib/modules/dosomething/dosomething_signup/dosomething_signup.module#L286) and update this to take that instead of `account`? We can also delete [this](https://github.com/DoSomething/phoenix/blob/6e80ae77499c1fd8b30508c1bad7ab6e8dd6d192/lib/modules/dosomething/dosomething_signup/dosomething_signup.module#L279-L281) since Rogue won't create a new Signup if one already exists.

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
chloealee: ahhhh I see I see.

lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
DFurnes: > @DFurnes Since we have the `northstar_id` of the user (assuming everyone has it), would it work to send the `northstar_id` directly to [`dosomething_rogue_send_signup_to_rogue`](https://github.com/DoSomething/phoenix/blob/6e80ae77499c1fd8b30508c1bad7ab6e8dd6d192/lib/modules/dosomething/dosomething_signup/dosomething_signup.module#L286) and update this to take that instead of `account`?

I think that's a great idea! 👍

> We can also delete [this](https://github.com/DoSomething/phoenix/blob/6e80ae77499c1fd8b30508c1bad7ab6e8dd6d192/lib/modules/dosomething/dosomething_signup/dosomething_signup.module#L279-L281) since Rogue won't create a new Signup if one already exists. 

Let's make sure that Phoenix returns the same response (the signup ID) if we do.

chloealee: ah actually I'm getting a little mixed up here @DFurnes - sorry about this! So unless I'm understanding incorrectly, I don't think we can just bypass any code that relies on a `uid` or loaded `$user` object if a Northstar ID is provided because of functions like [`dosomething_third_party_subscribe`](https://github.com/DoSomething/phoenix/blob/dev/lib/modules/dosomething/dosomething_signup/dosomething_signup.module#L293). This function uses others to find out user information (age, language, country)... unless it is ok to bypass this as well? 

Is the idea here to just make sure the user's signup and reportback gets logged in Rogue?

DFurnes: Oh no, that definitely makes things more complicated. 😩 We may want to hold off on this work until we've started sending these notifications via Blink/Customer.io.

> Is the idea here to just make sure the user's signup and reportback gets logged in Rogue?

We want to remove the requirement that every user _has to have_ a Phoenix UID, since that allows users who don't have Phoenix UIDs to signup or reportback for campaigns via Gambit. (Normally, Northstar makes a HTTP request to Phoenix to create the Drupal UID whenever a user is created but that sometimes breaks... I think due to network issues.)

Ideally, we should only have to make a Drupal UID when a user first logs in to Phoenix via the web, and no other systems should rely on it. This makes Northstar/Rogue more resilient, and makes it easier for us to remove Phoenix altogether one day.

chloealee: Ah that totally makes sense, thanks for clarifying, @DFurnes ! So then I think you're right - when going through the code as far as I could tell, the only thing that still rely on absolutely having more info. about the user (that is stored in P-A) is for sending to third party. I'll make a note in the card and let @ngjo know that this is blocked until we are able to send notifications via Blink/Customer.io.

DFurnes:
Closing this, since most other systems should be moving to talk directly to Rogue soon!

PR #7493, opened 2017-08-23
Don't send null query strings to Rogue.

DFurnes:
#### What's this PR do?
This should fix the issue where `api/v1/signups` was showing an empty page. Before, we'd send null values for some query strings, for example `?filter%5Bnorthstar_id%5D=&limit=25&page=1`. Now we don't: `?limit=25&page=1`.

👀

Was this always broken?! 🤔

#### Relevant tickets
[#150481585](https://www.pivotaltracker.com/story/show/150481585)

#### Checklist
- [ ] Tested on staging.


sbsmith86 (approved):
🥇

PR #7496, opened 2017-08-29
Rogue sends tags not tagged now

katiecrane:
#### What's this PR do?
Update the Rogue tags script to work with the response from Rogue's /post endpoint.

👀  Make sure this is actually how Rogue returns this data.

I mean to keep everything the same as before, but I accidentally switched this to `tags` and then the Rogue frontend started relying on it so we are changing it here too!

#### Relevant tickets
🐮 

#### Checklist
- [ ] Tested on staging.


sbsmith86 (approved):
💯

Issue #7498, opened 2017-09-14
Can't un-react to first 6 posts in gallery


katiecrane:
_BUG_
-------------------
#### Current Behavior
When a user re-visits a campaign page and tries to unlike a post (in the first 6 of the gallery) that they had previously liked, the heart will turn white and the counter will tick down. However, upon refreshing the page, the post will be liked again.

If a user on ONE page visit likes and unlikes a post, everything behaves as expected.

This is happening because an incomplete request is being sent to Rogue. Rogue receives the `post_id` but is getting a `null` `northstar_id`.

When the page loads, it _does_ hit [dosomething_kudos_get_kudos_data_for_fid](https://github.com/DoSomething/phoenix/blob/df6364925f9dcfd4f11c493315820a8ecd028438/lib/modules/dosomething/dosomething_kudos/dosomething_kudos.module#L323-L346) and set the `fid` as `post_id|northstar_id`. However, when the delete request hits [_kudos_resource_delete](https://github.com/DoSomething/phoenix/blob/fe60ffb9964d6f6253b2cbfd4970a43b108b206d/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc#L240-L282), the `$kudos_id` provided is somehow just the `post_id`.

#### Desired Behavior
When a user tries to unreact, the request sent to Rogue should include the `post_id` AND the `northstar_id`.

#### Steps to Replicate
1. Go to a campaign page.
2. Like one of the first 6 posts in the gallery.
3. Refresh the page.
4. Go unlike the post that you previously liked. See the heart turn white and the counter tick down.
5. Refresh the page.
6. See that the post is appearing as liked again.

PR #7499, opened 2017-09-14
Include pending posts on user profile

katiecrane:
#### What's this PR do?
Previously, the user profile page was using the Rogue `GET /reportbacks` endpoint, which also powers the gallery. This endpoint does not provide the option to include posts that have statuses other than `accepted`. Now the profile page uses the Rogue `GET /posts` endpoint, which provides all the data that we need. That endpoint gets hit with a filter for the `northstar_id` of the user whose profile page is being viewed. The `signup` is also included, and reaction data is retrieved `as_user` of the currently logged in user.

In order to unlike posts on the user profile, the Northstar ID had to be included with the Post ID in `dosomething_kudos_get_kudos_data_for_fid`.

Will this show all Posts a user has submitted on their profile page? Will you be able to properly react to these posts?

While working on this I found a bug which is documented here: https://github.com/DoSomething/phoenix/issues/7498

#### Relevant tickets
[Pivotal card](https://www.pivotaltracker.com/story/show/150312987)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/kudos_resource.inc
@@ -247,17 +247,13 @@ function _kudos_resource_delete($kudos_id) {
     $api_user_is_an_admin = user_access('view any reportback', $user);
 
     // Reactions from the user profile page will not have $northstar_id
-    if ($northstar_id) {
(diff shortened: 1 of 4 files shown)


DFurnes (approved):
This looks good to me! 👍

PR #7501, opened 2017-09-22
Don't enforce email/mobile uniqueness for Phoenix users.

DFurnes:
#### What's this PR do?
Phoenix's "create user" API endpoint is still enforcing email/mobile uniqueness, which we don't want anymore since the canonical source for these fields is in Northstar. By continuing to enforce this, we end up with some edge cases where accounts can't be created if a email/mobile was changed in Aurora (but the change wasn't pushed to Phoenix).

👀

This will fix an issue that was causing some Northstar accounts to be missing Phoenix IDs – specifically @lkpttn's, since he changed his mobile on Aurora to a new "test account" but the old Phoenix account was not updated (and so the new account with the mobile was a conflict oy oy oy).

There isn't a database uniqueness constraint on this column & we don't treat these `email` or `mobile` values as a source of truth on Phoenix, so there's no reason to enforce it on this endpoint.

#### Relevant tickets
Fixes 🐛.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/member_resource.inc
@@ -212,26 +212,9 @@ function _member_resource_create($request) {
   }
 
   // Check if email is formatted correctly and not already in use.
-  $email = $request['email'];
-  if ($user = user_load_by_mail($email)) {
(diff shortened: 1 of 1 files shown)


itsjoekent (approved):
I'm assuming we're ok dropping the `dosomething_user_valid_mobile` and email validation because NS handles that?

DFurnes:
Yup yup, we never sync that information into Northstar anymore so should be safe to bypass.

PR #7502, opened 2017-09-25
Expose campaign run start date in /api/v1/campaigns/:nid

chloealee:
#### What's this PR do?
Exposes campaign run state dates in `/api/v1/campaigns/:nid` in order to show this in the Rogue individual signup page for [this card](https://www.pivotaltracker.com/n/projects/2019429/stories/150849332). 

- Hit the `/api/v1/campaigns/:nid` endpoint. 
- You should now see a `start_date` returned for each campaign run. 
- e.g. 
```
{"data": {
    "id":"1485",
    "title":"Test Campaign GL",
    "campaign_runs": {
      "current": {
        "en": {
          "id":"1860",
          "start_date":"2015-10-19 00:00:00"
        },
        "en-global":{
          "id":"1860",
          "start_date":"2015-10-19 00:00:00"
        },
        "es-mx":{
          "id":"1860",
          "start_date":"2015-10-19 00:00:00"
        },
        "pt-br":{
          "id":"1860",
          "start_date":"2015-10-19 00:00:00"
        }
      },
      "past":[]
     },
...
```
#### Relevant tickets
Fixes https://www.pivotaltracker.com/n/projects/2019429/stories/151416179

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/documentation/endpoints/campaigns.md
+++ b/documentation/endpoints/campaigns.md
@@ -56,7 +56,8 @@ Example response:
     campaign_runs: {
       current: {
         en: {
-          id: "1227"
+          id: "1227",
+          "start_date":"2015-10-19 00:00:00"
         }
       },
       past: [ ]
--- a/lib/modules/dosomething/dosomething_api/includes/Transformer.php
+++ b/lib/modules/dosomething/dosomething_api/includes/Transformer.php
@@ -236,6 +236,11 @@ abstract class Transformer {
           }
         }
 
+        foreach ($data->campaign_runs['current'] as $language => $value) {
+          $run = node_load($value['id']);
+          $output['campaign_runs']['current'][$language]['start_date'] = $run->field_run_date[$language][0][value];
+        }
+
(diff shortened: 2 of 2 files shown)


sbsmith86:
Are we only returning `current` runs here? We will probably need `previous` ones as well.

sbsmith86:
Or maybe there is a way to pass in a run id param and then the run object will return the information for that run?

chloealee:
Great call! I think including the `previous` ones as well is the easier route to go because we're sending through multiple campaign ids to this `/campaigns/:nid` in [this call](https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/CampaignsController.php#L44), returning many campaigns. We have the `run_id` so I think we can just parse through the response when it gets back to Rogue after?

sbsmith86:
@chloealee sounds sane to me!

sbsmith86 (approved):

PR #7506, opened 2017-10-23
Fix redirect URI validation issue.

DFurnes:
#### What's this PR do?
This pull request fixes a configuration issue with the OAuth redirect URI when refreshing tokens – we had been providing the wrong value, and once we set specific whitelisted URLs for each client this started failing validation.

👀

Normal logins were still working because we offload those to the [OpenID Connect module](https://www.drupal.org/project/openid_connect) which gets it's redirect URL elsewhere. (Why does it validate redirect URI at all when fetching a refresh token? I'm really not sure, but the logic to include it is internal to [oauth2-client](https://github.com/thephpleague/oauth2-client) & the logic to validate it is standard for all grants in [oauth2-server](https://github.com/thephpleague/oauth2-server).)

#### Relevant tickets
Fixes [#152079260](https://www.pivotaltracker.com/story/show/152079260)

#### Checklist
- [ ] Tested on staging.


itsjoekent (approved):
interestin

PR #7507, opened 2017-10-23
Gets rid of all Rogue conditional flags since Rogue is always on

chloealee:
#### What's this PR do?
- Gets rid of Rogue conditional logic because Rogue will never be turned off!

- Make sure all the logic is still working that you can: 
  - Signup
  - Reportback
  - Update quantity/why participated
  - Gallery is working
  - Gallery pagination is working
  - Kudos / reactions are working (and are persisted when you refresh the page) 
  - User profile page and is working and all data shows up as expected 
  - @sbsmith86 does anything with competitions need to be tested for this?
  - Did I miss anything? 

#### Relevant tickets
Fixes the issue in this [Pivotal Card](https://www.pivotaltracker.com/n/projects/2019429/stories/151415698).

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
+++ b/lib/modules/dosomething/dosomething_api/resources/campaign_resource.inc
@@ -390,30 +390,18 @@ function _campaign_resource_reportback($nid, $values) {
 
   $headers = getallheaders();
 
-  if (variable_get('rogue_collection', FALSE) && ! isset($headers['X-Request-Id'])) {
-    $rogue_reportback = dosomething_rogue_send_reportback_to_rogue($values, $user);
+  $rogue_reportback = dosomething_rogue_send_reportback_to_rogue($values, $user);
 
-    // Get signup associated with the post.
-    $rid = $rogue_reportback['data']['signup_id'];
-    $rogue_signup = dosomething_rogue_get_activity(['filter' => ['id' => $rid]]);
+  // Get signup associated with the post.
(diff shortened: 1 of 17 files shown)


DFurnes (approved):
This is so good!! 🔥 🔥 🔥

chloealee:
yay thanks for the review, @DFurnes ! @sbsmith86 before I merge, just want to double check re: competitions work!

DFurnes:
Yeah, competitions and signup data form are definitely the most "dangerous" changes. Are we currently using either of those on any legacy campaigns?

chloealee:
@DFurnes asking around now to see! But for the signup data form, I tested that signups are working and all looked good on my local!

chloealee:
Asked Ashely in [this thread ](https://dosomething.slack.com/archives/C02BBNZFA/p1508866641000183)and she said that we are no longer doing competitions for any legacy campaigns on ashes! So I think these changes should be safe then?

sbsmith86:
@chloealee I'm good with killing storing competition info on phoenix-ashes. The only think I _could_ be concerned with is PN trying to find out if a user is in a competition or not, but they should be getting that from Gladiator and since there is nothing new happening in Gladiator this quarter we can probably kill it. 

As we look forward to Q4 we will probably handle competition signup information differently anyway.

chloealee:
ok cool, thanks @sbsmith86 ! I'll merge this now :)

PR #7508, opened 2017-10-24
Signupless Reportbacks -> Rogue script

katiecrane:
#### What's this PR do?
This is a script to migrate Reportbacks that do not have a corresponding signup over to Rogue. It sends the lonely posts to the Rogue `/posts` endpoint, which creates a signup if there is not one already. Errors are handled gracefully, and no data will be duplicated in Rogue.

Documentation in progress [here](https://docs.google.com/document/d/157aE2SPT_rElkEBEclONwP_1XUYT1qtjY5glv5S9lN4/edit).

Will this mess up any data? Will this mess up Rogue?

That darn pre-Gambit activity!

#### Relevant tickets
[Pivotal](https://www.pivotaltracker.com/story/show/151761727)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_rogue/includes/Rogue.php
+++ b/lib/modules/dosomething/dosomething_rogue/includes/Rogue.php
@@ -134,4 +134,17 @@ class Rogue extends RestApiClient {
 
     return $response;
   }
+
+  /**
+   * Send a POST request to create a new post in Rogue
+   *
+   * @param array $data
+   * @return array - JSON response
+   */
+  public function postPost($data)
+  {
+    $response = $this->post('v2/posts', $data);
+
+    return $response;
+  }
 }
--- a/scripts/export-signupless-reportbacks-to-rogue.php
+++ b/scripts/export-signupless-reportbacks-to-rogue.php
@@ -46,7 +46,7 @@ foreach ($reportbacks as $reportback) {
     echo "\t" . 'Reportback ' . $reportback->rbid . ' has no files, sending just a signup' . PHP_EOL;
 
     // Match Rogue's timestamp format
-    $rb_created_at = date('Y-m-d H:i:s', $photo->timestamp);
+    $rb_created_at = date('Y-m-d H:i:s', $reportback->timestamp);
 
     $data = [
       'northstar_id' => $northstar_id,
@@ -69,12 +69,12 @@ foreach ($reportbacks as $reportback) {
 
       // Handle getting a 404
       if (!$response) {
-        echo '***ERROR: 404 on reportback ' . $reportback->rbid . ' file ' . $photo->fid . PHP_EOL;
+        echo '***ERROR: 404 on reportback ' . $reportback->rbid . PHP_EOL;
       }
     }
     catch (GuzzleHttp\Exception\ServerException $e) {
       // These aren't yet caught by Gateway
-      echo '***ERROR: SERVER EXCEPTION on reportback ' . $reportback->rbid . ' file ' . $photo->fid . PHP_EOL;
+      echo '***ERROR: SERVER EXCEPTION on reportback ' . $reportback->rbid . PHP_EOL;
     }
     catch (DoSomething\Gateway\Exceptions\ApiException $e) {
       echo '***ERROR: API EXCEPTION on reportback ' . $reportback->rbid . PHP_EOL;
--- a/scripts/export-signupless-reportbacks-to-rogue.php
+++ b/scripts/export-signupless-reportbacks-to-rogue.php
@@ -6,7 +6,7 @@
  * drush --script-path=../scripts/ php-script export-signupless-reportbacks-to-rogue.php
  */
 
-// get all the signupless reportbacks
+// Get all the signupless reportbacks
 $reportbacks = db_query("SELECT rb.rbid, rb.nid, rb.run_nid, rb.quantity, rb.why_participated, rb.rbid, rb.flagged, rb.uid
   FROM dosomething.dosomething_reportback rb
   LEFT JOIN dosomething.dosomething_signup ds on rb.uid = ds.uid AND rb.run_nid = ds.run_nid
@@ -19,7 +19,7 @@ if (!$reportbacks) {
   echo 'Nothing to migrate!';
 }
 
-// for each reportback, send a request for each reportback file to POST /posts and Rogue will automatically create the signup
+// For each reportback, send a request for each reportback file to POST /posts and Rogue will automatically create the signup
 foreach ($reportbacks as $reportback) {
   echo 'On rbid ' . $reportback->rbid . '...' . PHP_EOL;
   $sent_at = date('Y-m-d H:i:s');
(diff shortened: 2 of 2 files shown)


scripts/export-signupless-reportbacks-to-rogue.php
```diff
@@ -0,0 +1,147 @@
...
+  echo 'On rbid ' . $reportback->rbid . '...' . PHP_EOL;
+  $sent_at = date('Y-m-d H:i:s');
+
+  // Try to get Northstar ID
+  $northstar_id = dosomething_user_get_northstar_id($reportback->uid, 'drupal_id');
+
+  if (!$northstar_id) {
+    echo "\t" . 'No northstar id, that is terrible ' . $reportback->rbid . PHP_EOL;
+
+    continue;
```
chloealee, line 33: do we log the reportback anywhere (in a failed logs table maybe?) that doesn't have a northstar id to try to re-send/investigate later?

scripts/export-signupless-reportbacks-to-rogue.php
```diff
@@ -0,0 +1,147 @@
...
+
+  // Just send a signup if there are no photos (there are about 25 of these)
+  if (!$photos) {
+    echo "\t" . 'Reportback ' . $reportback->rbid . ' has no files, sending just a signup' . PHP_EOL;
+
+    $data = [
+      'northstar_id' => $northstar_id,
+      'campaign_id' => $reportback->nid,
+      'campaign_run_id' => $reportback->run_nid,
+      'created_at' => $photo_created_at,
```
chloealee: does `$photo_created_at` exist yet?

scripts/export-signupless-reportbacks-to-rogue.php
katiecrane: @chloealee we do not, the thinking is that if there is no user there's not much else we can do. These ones will be tried every time we run the script though, and we can see in the logs which ones don't have northstar ids if we want to

chloealee: ah sorry! I see in line 120 the logic behind this that whenever we run the script these will always be tried and if successful, we save the information for that record!

scripts/export-signupless-reportbacks-to-rogue.php
katiecrane: @chloealee great catch! I'm surprised this never errored out on me! I'll also want to use the timestamp of the reportback here

katiecrane: @chloealee oooo I bet it didn't get mad because it was still set as something, but it's an irrelevant timestamp here

scripts/export-signupless-reportbacks-to-rogue.php
katiecrane: @chloealee yup that's exactly right!

scripts/export-signupless-reportbacks-to-rogue.php
katiecrane: @chloealee hmm I thought this comment was in the spot above where I was using `$photo_created_at` and shouldn't have been. In this spot it gets created on line 92 😺

chloealee: @katiecrane I think it was! Weird!

chloealee:
This looks great to me @katiecrane ! Not sure if you'd want other 👀  on this too for good measure?

katiecrane:
@chloealee more 👀  never hurt but it will be tested on Thor first! @sbsmith86 think it's safe to test on Thor or would you like to take a closer look?

lib/modules/dosomething/dosomething_rogue/includes/Rogue.php
```diff
@@ -134,4 +134,17 @@ public function postTag($data)
...
     return $response;
   }
+
+  /**
+   * Send a POST request to create a new post in Rogue
+   *
+   * @param array $data
+   * @return array - JSON response
+   */
+  public function postPost($data)
```
sbsmith86, line 144: oooo fun name.

scripts/export-signupless-reportbacks-to-rogue.php
```diff
@@ -0,0 +1,150 @@
...
+      'quantity' => $reportback->quantity,
+      'why_participated' => $reportback->why_participated,
+    ];
+
+    try {
+      // We do not need to worry about keeping track of these because sening them to Rogue multiple times will not create multiple signups
+      $response = $client->postSignup($data);
+
+      if ($response) {
+        echo "\t\t" . 'Migrated reportback ' . $reportback->rbid . ' to Rogue.' . PHP_EOL;
```
sbsmith86: hmm we aren't migrating the RB here though, correct? I think just saying we create the signup in rogue is enough.

scripts/export-signupless-reportbacks-to-rogue.php
```diff
@@ -0,0 +1,150 @@
...
+        echo "\t\t\t" . 'Rogue Signup ID: ' . $response['data']['signup_id'] . PHP_EOL;
+      }
+
+      // Handle getting a 404
+      if (!$response) {
+        echo '***ERROR: 404 on reportback ' . $reportback->rbid . ' file ' . $photo->fid . PHP_EOL;
```
sbsmith86: Where is `photo` coming from here?

scripts/export-signupless-reportbacks-to-rogue.php
```diff
@@ -0,0 +1,150 @@
...
+
+  // Get reportback files
+  $photos = db_query('SELECT rbf.fid, rbf.remote_addr, rbf.caption, rbf.status, rbf.reviewed, rbf.reviewer, rbf.source, rbf.status, rblog.timestamp
+                          FROM dosomething_reportback_file rbf
+                          INNER JOIN dosomething_reportback_log rblog ON rbf.fid = substring_index(rblog.files, \',\',-1)
+                          WHERE rbf.rbid = '. $reportback->rbid . '
+                          AND rbf.fid NOT IN (SELECT fid from dosomething_rogue_reportbacks)
+                          GROUP BY rbf.fid')->fetchAll();
+
+  // Just send a signup if there are no photos (there are about 25 of these)
```
sbsmith86, line 44: I think I might be a little confused on this piece. I don't see where you are looking through the RBS with out photos

katiecrane: @sbsmith86 the query above gets all the reportback files associated with the reportback. There are a few cases where that query won't return anything. In those cases, below (`if (!photos)`), we send over that data as just a signup.

scripts/export-signupless-reportbacks-to-rogue.php
katiecrane: @sbsmith86 good catch! this language is confusing. technically we _are_ migrating over the data the lives on the reportback in Phoenix, but in doing so we create a signup in Rogue.

I want to maintain having the `rbid` in the log, so how about `Migrated data on reportback 45 to Rogue` and the the Rogue signup ID prints below?

scripts/export-signupless-reportbacks-to-rogue.php
katiecrane: @sbsmith86 that should be removed

scripts/export-signupless-reportbacks-to-rogue.php
sbsmith86: sounds good!

sbsmith86 (approved):

PR #7519, opened 2018-03-05
Adding HelloBar script.

weerd:
#### What's this PR do?
This PR adds a script for using HelloBar to add a persistent bar at the top of all Ashes pages to prompt visitors to register to vote. HelloBar will also let us do some customizations and A/B tests to the bar to see which variations perform better.

image

#### Relevant tickets
Refs [Pivotal #155590930](https://www.pivotaltracker.com/story/show/155590930)


weerd:
Before deploying to Production, @ashleybaldwin should confirm that the settings we want in HelloBar for the initial prompt are all set!

DFurnes (approved):
A true beaut! 🌷

PR #7520, opened 2018-03-06
Turn off cropping

sbsmith86:
#### What's this PR do?

Turns of image cropping on Phoenix Ashes! 

Instead of popping up the cropping modal. We just add a preview image to the `.gigantor` element and unhide the caption field so it is displayed on the main form (instead of the modal).

Video example (it was cut a little short but submission still successfully goes to rogue): https://cl.ly/1i3t0L1f3N0J

👀 

In Rogue, we want to start using [Glide](http://glide.thephpleague.com/) to handle image orientation and cropping when serving images. 

To get this working on rogue, we need to turn off the user submitted cropping options so we can just default to auto orientation and a square crop in glide. 

#### Relevant tickets

Addresses https://www.pivotaltracker.com/story/show/154979852

#### Checklist
- [ ] Tested on staging.


DFurnes (approved):
🐦 🔥

PR #7527, opened 2018-05-15
Updating version of grunt-sass to fix compiling error

mirie:
Build of international site is failing: 

```
 DEBUG [8ed474a1]     Loading "sass.js" tasks...
 DEBUG [8ed474a1]     ERROR
 DEBUG [8ed474a1]     >> Error: `libsass` bindings not found. Try reinstalling `node-sass`?
 DEBUG [8ed474a1]     Warning: Task "sass:prod" not found. Use --force to continue.
 DEBUG [8ed474a1]     
 DEBUG [8ed474a1]     Aborted due to warnings.
 ```

I think this is due to an outdated version of grunt-sass

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/package.json
+++ b/lib/themes/dosomething/paraneue_dosomething/package.json
@@ -37,7 +37,7 @@
     "grunt-eslint": "^16.0.0",
     "grunt-modernizr": "^0.6.0",
     "grunt-postcss": "^0.2.0",
-    "grunt-sass": "^0.18.0",
+    "grunt-sass": "^2.0.0",
     "grunt-sasslint": "0.0.5",
     "grunt-webpack": "^1.0.8",
     "load-grunt-tasks": "^3.1.0",


blisteringherb (approved):
This looks good to me. If we're at all concerned that it'll block Phoenix deploys in the future, we can worry about Int'l now and then test for the US site later this week.

mirie:
@blisteringherb wait, what do you mean re: possibly blocking Phoenix deploys in the future? Wanna make sure I understand before I merge!

blisteringherb:
@mirie I was thinking that we haven't tested it to see if the old versions of Sass are needed in some way for the current Phoenix deploys and whether making this change will break those deploys. It's probably unlikely, but just mentioning it, and we can cross that bridge when we get to it. This is most likely also broken in those deploys as well.

mirie:
Ah, I see - Looking at package.json on dev branch: https://github.com/DoSomething/phoenix/blob/dev/lib/themes/dosomething/paraneue_dosomething/package.json

Looks like that package is no longer being used. International is pretty outdated

PR #7530, opened 2018-05-15
Create node_modules dir in forge dir and cp node_modules/susy to this directory

mirie:
#### What's this PR do?
Context:

We still have a build issue (issue with Forge and susy dependency):

alt text

From @DFurnes:
> NPM 2.x nested `node_modules` within the modules that depended on the package. From 3.x onwards, dependencies are all flattened in a single `node_modules` directory.
Unfortunately that path is inside a vendored `forge.scss` file, and we expect Webpack as our bundler in newer releases of Forge.

This fix creates the `node_modules` directory in the forge directory and copies over the `susy` directory to this location.

Hmmmm, untested because I can’t get a local build working! @_@

#### Relevant Tickets
https://github.com/DoSomething/devops/issues/406

### DIFF
--- a/bin/ds
+++ b/bin/ds
@@ -421,6 +421,9 @@ function grunt {
     eval "/usr/bin/npm install"
   fi
 
+  mkdir ./node_modules/@dosomething/forge/node_modules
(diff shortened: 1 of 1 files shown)


DFurnes (approved):
🤞 🔨

PR #7532, opened 2018-05-16
Pinned messagebroker-phplib to version 0.2.9

mirie:
#### What's this PR do?
Need to pin to this older version of messagebroker-phplib because the method DoSomething\MessageBroker\MessageBroker::publishMessage() that’s being called in message_broker_producer.module doesn’t exist in the newer version of this library @_@

#### Relevant tickets
Related to https://github.com/DoSomething/devops/issues/406


blisteringherb (approved):
Ah ha! Makes sense! 👍

PR #7533, opened 2018-05-23
Get user profile info from Northstar

sbsmith86:
#### What's this PR do?

This PR updates the `dosomething_user_preprocess_user_profile` function to pull user data directly from NS, bypassing Drupal user field tables entirely. 

👀 

GDPR task.

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.theme.inc
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.theme.inc
@@ -69,16 +69,14 @@ function dosomething_user_get_register_link($label, array $class = []) {
 function dosomething_user_preprocess_user_profile(&$variables) {
   // Collect User Account data.
   $uid = $variables['elements']['#account']->uid;
-  $user = user_load($uid);
-  $variables['fake'] = "fake";
   $nsUser = dosomething_northstar_get_user($uid, 'drupal_id');
-  dump($nsUser->first_name);
-  die();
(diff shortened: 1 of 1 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.theme.inc
```diff
@@ -69,12 +69,17 @@ function dosomething_user_get_register_link($label, array $class = []) {
...
-  $user = user_load($uid);
-  $variables['first_name'] = dosomething_user_get_field('field_first_name', $user, 'ucwords');
-  $variables['last_name'] = dosomething_user_get_field('field_last_name', $user, 'ucwords');
+
+  $nsUser = dosomething_northstar_get_user(
+    dosomething_user_get_northstar_id($uid)
+  );
+
+  $variables['first_name'] = $nsUser->first_name;
+  $variables['last_name'] = $nsUser->last_name;
```
weerd: Will this error if `last_name` is not set? Or is it `null` or something? Just curious since Last Name isn't a required field.

weerd (approved):
One question otherwise looks good to me! 👍

DFurnes (approved):

lib/modules/dosomething/dosomething_user/dosomething_user.theme.inc
sbsmith86: Maybe! i will add a check for it. Good flag.

PR #7534, opened 2018-05-24
Remove field dependencies

sbsmith86:
#### What's this PR do?

The PR attempts to start cleaning up dependencies we have on `dosomething_user_get_field()` which grabbed fields from the PA DB.

In the places I saw fit, I just pulled the user information directly from NS. 

👀 

I tried to write in the commit messages why I felt it was safe to remove the code that I removed.

GDPR. 

Also, there could certainly ways to be cleaner about this, but we are on a deadline of 5/25, so instead of being super DRY and clean, i opted to just get this working so things don't break obviously.

#### Relevant tickets

GDPR
Addresses https://github.com/DoSomething/devops/issues/403


DFurnes (approved):
This seems okay to me!

PR #7536, opened 2018-05-24
Remove tables we don't need

sbsmith86:
#### What's this PR do?
Removing tables as outlined in the GDPR list. 

Only gotcha is that we only truncated `dosomething_northstar_request_log` and `dosomething_rogue_request_log` since we do need these tables. 

👀 

GDPR

### DIFF
--- a/lib/modules/dosomething/dosomething_user/dosomething_user.install
+++ b/lib/modules/dosomething/dosomething_user/dosomething_user.install
@@ -143,3 +143,12 @@ function dosomething_user_update_7012() {
   field_delete_field('field_school_id');
   field_delete_field('field_user_registration_source');
 }
+
+/**
+ * Remove mail field for GDPR compliance.
+ */
+function dosomething_user_update_7013()
+{
+  db_drop_index('users', 'mail');
+  db_drop_field('users', 'mail');
+}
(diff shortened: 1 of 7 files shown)


lib/modules/dosomething/dosomething_user/dosomething_user.install
```diff
@@ -143,3 +143,19 @@ function dosomething_user_update_7012() {
   field_delete_field('field_school_id');
   field_delete_field('field_user_registration_source');
 }
+
+/**
+ * Remove blocked Ips table that isn't used and remove IPs from sessions table.
+ */
+function dosomething_user_update_7013()
+{
+  $tables = 'blocked_ips';
```
DFurnes: Is this a Drupal core thing? It might be safest to just truncate for now.

sbsmith86: it's actually empty right now. I can just remove this. There is nothing in it on prod

DFurnes (approved):

lib/modules/dosomething/dosomething_user/dosomething_user.install
```diff
@@ -143,3 +143,11 @@ function dosomething_user_update_7012() {
   field_delete_field('field_school_id');
   field_delete_field('field_user_registration_source');
 }
+
+/**
+ * Remove IPs from sessions table.
+ */
+function dosomething_user_update_7013()
+{
+  db_drop_field('sessions', 'hostname');
```
DFurnes, line 152: Actually, one more question - does this remove the column altogether or just zero it out? Are we safe to remove, or will Drupal continue to try to write to this?

PR #7540, opened 2018-05-25
Fix fatal error during authentication.

DFurnes:
#### What's this PR do?
This pull request fixes a fatal error where the OpenID Connect module attempts to parse birthdates incorrectly (even though it shouldn't be using them...) This fixes things up.

👀

The original fix was [here](https://github.com/DoSomething/phoenix/commit/7a52be8b9ed84ab7a14e4c04473bf23f2e2d85ee#diff-b5fe2464497701d1518b40a8d5e00e7c), but we don't care about real data in Ashes now.

#### Relevant tickets
N/A

#### Checklist
- [ ] Tested on staging.


katiecrane (approved):
👋 birthday!

PR #7166, opened 2016-11-01
Adds advertising tracking for FB & Twitter

itsjoekent:
#### What's this PR do?
- Adds advertising tracking for FB & Twitter
- (Replaces/Removes social auth prototype code)

#### How should this be reviewed?
Does the checkbox properly enable / disable the tracking code?

#### Any background context you want to provide?
Needed for VCard ad campaign asap

#### Relevant tickets
Fixes #n/a

PR #7545, opened 2018-06-22
Adding feature flag to include Facebook Pixel script.

weerd:
#### What's this PR do?
This PR adds a feature flag that will include the Facebook Pixel ad tracking script on every page on Ashes. Since Facebook only gives you one FB Pixel ID and we don't want local development/qa staging activity muddying up the data, we only want to toggle the script to show up on production. The feature flag defaults to `0` (false) via the `paraneue_dosomething.info` file.

As it turns out, after I finished coding this all up, I scrolled a bit farther down on the `html.tpl.php` file, only to find that a Facebook Pixel script was already included in 2016 via #7166 😱! 

Apparently it was behind a `enable_ad_tracking` toggle in the `dosomething_settings` module. However, that toggle bundles multiple scripts (a twitter ad tracking script as well), together. So to have finer grain control I felt that the feature flag I added would be more ideal. Thoughts?

Also, the FB Pixel ID for the old script doesn't seem to be something that we have on our account, so even if activated, not sure if it was tracking anything.

#### Relevant tickets
Refs [Pivotal ID #158462761](https://www.pivotaltracker.com/story/show/158462761)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_html.inc
+++ b/lib/themes/dosomething/paraneue_dosomething/includes/preprocess/preprocess_html.inc
@@ -28,6 +28,8 @@ function paraneue_dosomething_preprocess_html(&$vars) {
   $member_count = dosomething_user_get_member_count(TRUE);
   $vars['head_title'] = token_replace($vars['head_title'], ['member-count-readable' => $member_count]);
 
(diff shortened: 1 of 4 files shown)


DFurnes (approved):
Yeah, 👍 to separate feature flags & 😂 to it already being there. Oy vey!

PR #7546, opened 2018-06-22
Updates signup and post source from pheonix-ashes to phoenix-web

chloealee:
#### What's this PR do?
Updates signup and post `source` from `phoenix-ashes` to `phoenix-web`.

This will pull `phoenix-ashes` from using `variable_get('dosomething_northstar_app_id');` for consistency. 

Is there any other place I missed where `source` should be updated?

#### Relevant tickets
Fixes [this](https://www.pivotaltracker.com/n/projects/2019429/stories/158349333) Pivotal Card. 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -402,7 +402,7 @@ function dosomething_reportback_form_submit($form, &$form_state) {
   if (array_key_exists('storage', $form_state)) {
     $file = $form_state['storage']['file'];
     $values['file'] = dosomething_helpers_get_data_uri_from_fid($file->fid);
-    $values['source'] = variable_get('dosomething_northstar_app_id');;
+    $values['source'] = variable_get('dosomething_northstar_app_id');
   } else {
     // If there is no file and there is a RB in Phoenix, check the dosomething_rogue_reportback table to see if the RB already exists in Rogue.
     if ($rbid) {
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -402,7 +402,7 @@ function dosomething_reportback_form_submit($form, &$form_state) {
   if (array_key_exists('storage', $form_state)) {
     $file = $form_state['storage']['file'];
     $values['file'] = dosomething_helpers_get_data_uri_from_fid($file->fid);
-    $values['source'] = 'phoenix-ashes';
+    $values['source'] = variable_get('dosomething_northstar_app_id');;
   } else {
     // If there is no file and there is a RB in Phoenix, check the dosomething_rogue_reportback table to see if the RB already exists in Rogue.
     if ($rbid) {
@@ -413,7 +413,7 @@ function dosomething_reportback_form_submit($form, &$form_state) {
         $reportback_file = entity_load('reportback_item', array($fid));
         $values['file'] = dosomething_helpers_get_data_uri_from_fid($fid);
         $values['caption'] = $reportback_file[$fid]->caption;
-        $values['source'] = isset($reportback_file[$fid]->source) ? $reportback_file[$fid]->source : 'phoenix-ashes';
+        $values['source'] = isset($reportback_file[$fid]->source) ? $reportback_file[$fid]->source : 'phoenix-web';
       }
     }
   }
--- a/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
+++ b/lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
@@ -402,7 +402,7 @@ function dosomething_reportback_form_submit($form, &$form_state) {
   if (array_key_exists('storage', $form_state)) {
     $file = $form_state['storage']['file'];
     $values['file'] = dosomething_helpers_get_data_uri_from_fid($file->fid);
-    $values['source'] = 'phoenix-web';
+    $values['source'] = 'phoenix-ashes';
   } else {
     // If there is no file and there is a RB in Phoenix, check the dosomething_rogue_reportback table to see if the RB already exists in Rogue.
     if ($rbid) {
@@ -413,7 +413,7 @@ function dosomething_reportback_form_submit($form, &$form_state) {
         $reportback_file = entity_load('reportback_item', array($fid));
         $values['file'] = dosomething_helpers_get_data_uri_from_fid($fid);
         $values['caption'] = $reportback_file[$fid]->caption;
-        $values['source'] = isset($reportback_file[$fid]->source) ? $reportback_file[$fid]->source : 'phoenix-web';
+        $values['source'] = isset($reportback_file[$fid]->source) ? $reportback_file[$fid]->source : 'phoenix-ashes';
       }
     }
   }
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
@@ -45,6 +45,10 @@ function dosomething_signup_form($form, &$form_state, $nid, $label = "Sign Up") 
 
   // Check query string for signup source value.
   if ($source = dosomething_signup_get_query_source()) {
+    if ($source === NULL) {
+      $source = 'phoenix-ashes';
+    }
+
     $form['source'] = [
       '#type' => 'hidden',
       '#value' => $source,
@@ -88,7 +92,7 @@ function dosomething_signup_form_submit($form, &$form_state) {
 
   $nid = $form_state['values']['nid'];
 
-  $source = NULL;
+  $source = 'phoenix-ashes';
   if (isset($form_state['values']['source'])) {
     $source = $form_state['values']['source'];
   }
--- a/lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
+++ b/lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
@@ -44,17 +44,13 @@ function dosomething_signup_form($form, &$form_state, $nid, $label = "Sign Up") 
   ];
 
   // Check query string for signup source value.
-  if ($source = dosomething_signup_get_query_source()) {
-    if ($source === NULL) {
-      $source = 'phoenix-ashes';
-    }
-
-    $form['source'] = [
-      '#type' => 'hidden',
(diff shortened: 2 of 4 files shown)


lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -402,7 +402,7 @@ function dosomething_reportback_form_submit($form, &$form_state) {
   if (array_key_exists('storage', $form_state)) {
     $file = $form_state['storage']['file'];
     $values['file'] = dosomething_helpers_get_data_uri_from_fid($file->fid);
-    $values['source'] = 'phoenix-web';
+    $values['source'] = 'phoenix-ashes';
```
DFurnes: Can we use `phoenix-client` (Ashes' OAuth client) for consistency with other `source` values? You should be able to load it in from the site's settings using `variable_get('dosomething_northstar_app_id');`

lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
```diff
@@ -45,6 +45,10 @@ function dosomething_signup_form($form, &$form_state, $nid, $label = "Sign Up")
 
   // Check query string for signup source value.
   if ($source = dosomething_signup_get_query_source()) {
```
DFurnes: Should this query-string source be applied to `source_details` instead? Ideally `source` is always a "trusted" value.

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
chloealee: @DFurnes that makes sense to me! Data had requested to change all `NULL` sources to `phoenix-ashes` and we wanted to make sure that all new records from Ashes also had this same source value to stay consistent. Should I check with the data team to see if it is ok to change everything to `phoenix-client` instead of `phoenix-ashes` and also change any existing sources in the database to `phoenix-client`? 

One q though - will this make it confusing to differentiate what comes from Ashes vs Phoenix?

lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
chloealee: @DFurnes I don't think the signups table has a `source_details` column but it does have a `source` column! Should I put it there instead?

chloealee: an example of something that might be in that column is this:
```
{"campaign_id":"6uFn4194vCOYmoGm88EEwK","legacy_campaign_id":"8064","legacy_campaign_run_id":"8065"}
```

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
DFurnes: > One q though - will this make it confusing to differentiate what comes from Ashes vs Phoenix?

Yeah, I can see that! I'm up for creating a duplicate of [phoenix-client](https://aurora.dosomething.org/clients/phoenix-client) called `phoenix-ashes` to make that clearer. My preference is to always have our source columns to something [we can find in Aurora](https://aurora.dosomething.org/clients) though, if possible.

lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
DFurnes: Oh, bummer. I wonder if we should add a `source_details` there for consistency?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
chloealee: @DFurnes great call - that makes sense to me!

lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
chloealee: I think that makes sense! I can work on that now and then update this to populate `source_details` instead of `source` :)

DFurnes: Thanks @chloealee!! 👍

chloealee: Of course - thank you for flagging this, @DFurnes ! Looking into a few other records that have the `details` field populated, I'm seeing some other trends: 

`
{"hostname":"dosomething.turbovote.org","referral-code":"***","partner-comms-opt-in":"","created-at":"2018-05-20T22:02:42Z","updated-at":"2018-05-20T22:02:57
`

`
affiliate-opt-out
`

I'm wondering if I should just change `details` to `source_details` instead of adding another column? This would probably affect some other apps like chompy though cc @sbsmith86

DFurnes: I think we use `details` as more of a catch-all bucket, so it probably should stay as is.

chloealee: ah that makes sense! [Here's the PR](https://github.com/DoSomething/rogue/pull/744) to add the `source_details` column :)

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
```diff
@@ -402,7 +402,7 @@ function dosomething_reportback_form_submit($form, &$form_state) {
   if (array_key_exists('storage', $form_state)) {
     $file = $form_state['storage']['file'];
     $values['file'] = dosomething_helpers_get_data_uri_from_fid($file->fid);
-    $values['source'] = 'phoenix-web';
+    $values['source'] = variable_get('dosomething_northstar_app_id');;
```
DFurnes: An extra `;` snuck in at the end here!

lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
```diff
@@ -88,7 +92,7 @@ function dosomething_signup_form_submit($form, &$form_state) {
 
   $nid = $form_state['values']['nid'];
 
-  $source = NULL;
+  $source = 'phoenix-ashes';
```
DFurnes: Should this read the Client ID as well, like above?

lib/modules/dosomething/dosomething_reportback/dosomething_reportback.forms.inc
chloealee: ah yikes - thanks @DFurnes !

chloealee:
@DFurnes cleaned up a bit - ready again for a look! 

Also, once this is merged, how do we make a duplicate client (of `phoenix-client`) called `phoenix-ashes` so that will be the source value when going through Ashes' OAuth client?

chloealee:
Also, @DFurnes question relating to the reason we are changing to get `source` from OAuth client in this PR - we did this to make it more consistent and to always have our source columns to something we can find in Aurora though, if possible. 

I noticed that for voter-reg posts in Rogue, Chompy can update the logic for source to be `Turbovote` - should we update these to be the `source_details` and have the `source` be Chompy's OAuth client?  cc @DoSomething/team-bleed

DFurnes:
> Also, once this is merged, how do we make a duplicate client (of `phoenix-client`) called `phoenix-ashes` so that will be the source value when going through Ashes' OAuth client?

You can create a new client by hitting "New Client" on the [Clients page in Aurora](https://aurora.dosomething.org/clients). It should use the "Machine (Client Credentials)" and be named `phoenix-ashes`. You can give it the `user`, `activity`, and `admin` scopes.

DFurnes:
> I noticed that for voter-reg posts in Rogue, Chompy can update the logic for source to be `Turbovote` - should we update these to be the `source_details` and have the `source` be Chompy's OAuth client?  cc @DoSomething/team-bleed 

Yeah, my preference would be to have the `source` be `importer-client`, and to store a specific CSV source in the `source_details` so that we can keep the "source" field trusted. Another option would be to have Chompy swap out between multiple client IDs per-source, but that seems like it might be a bit over-complicated.

chloealee:
Cool, thanks @DFurnes ! I added it [here](https://aurora.dosomething.org/clients/phoenix-ashes) and also gave it the `write` scope in addition to the ones you mentioned! Is there anything else I need to do now to get Ashes to use this so that it will pull in the correct source?

chloealee:
> Yeah, my preference would be to have the source be importer-client, and to store a specific CSV source in the source_details so that we can keep the "source" field trusted. Another option would be to have Chompy swap out between multiple client IDs per-source, but that seems like it might be a bit over-complicated.

This sounds good to me! Should I create cards to write a script to clean this up in Rogue and then update the logic in Chompy for new Turbovote imports?

DFurnes:
> Is there anything else I need to do now to get Ashes to use this so that it will pull in the correct source? 

Just swap it out here on production! [Configuration → Web Services → Northstar](https://www.dosomething.org/admin/config/services/northstar).

> This sounds good to me! Should I create cards to write a script to clean this up in Rogue and then update the logic in Chompy for new Turbovote imports?

Yeah, that sounds great to me!

chloealee:
Awesome, thanks @DFurnes ! This is ready then for a final review when you have a chance! :) Once it is merged, I'll test on Thor and then update where you linked to above in Prod right after deploy!

chloealee:
And [here](https://www.pivotaltracker.com/n/projects/2019429/stories/158771503) is the card for the Chompy imports!

lib/modules/dosomething/dosomething_signup/dosomething_signup.forms.inc
```diff
@@ -44,13 +44,13 @@ function dosomething_signup_form($form, &$form_state, $nid, $label = "Sign Up")
...
 
   // Check query string for signup source value.
-  if ($source = dosomething_signup_get_query_source()) {
-    $form['source'] = [
-      '#type' => 'hidden',
-      '#value' => $source,
-      '#access' => FALSE,
-    ];
-  }
+  // if ($source_details = dosomething_signup_get_query_source()) {
```
DFurnes: Have you tested that a campaign with a source query string properly fills in this field without the hidden form element? I'm thinking that the call to `dosomething_signup_get_query_source` might not work since it'd be running in the form's POST, where the query string would no longer exist.

(If this does work, let's just remove these commented lines!)

chloealee:
@DFurnes oops sorry I thought I had commented those out already! Updated! And yes - tested both with the query string and without and it properly saves to `source_details` (and works without)!

DFurnes (approved):
Looks good to me!

PR #7548, opened 2018-07-02
Update to remove Facebook Pixel related code.

weerd:
#### What's this PR do?

This PR removes the code related to the Facebook Pixel.

Round and round we go! After some further clarification, we only want to include the Google Tag Manager script, since you can use Google's container script to include other scripts, such as the Facebook Pixel script with a bit more fine-tuned control.

#### Relevant tickets
Refs [Pivotal ID #158462761](https://www.pivotaltracker.com/story/show/158462761)


DFurnes (approved):
We hardly knew ye! 👋

PR #7552, opened 2018-09-05
Intl sites

crittermike:
#### What's this PR do?
This switches from sending mail via messagebroker to sending mail via Mandrill and Mailsystem, using Mandrill template mappings.

Check out the code, run DB updates, then do things that would normally send emails, such as the forgot password link or signing up for a campaign.

### DIFF
--- a/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
+++ b/lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
@@ -3,7 +3,7 @@
  * @file
  * Installation and schema hooks for dosomething_mbp.module.
  */
- 
+
 /**
  * Implements hook_install().
  */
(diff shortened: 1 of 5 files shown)


lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
```diff
@@ -135,9 +135,68 @@ function dosomething_mbp_update_7007() {
   dosomething_mbp_install_productions();
 }
 
+/**
+ * Add config for Mailsystem and Mandrill modules.
+ */
+function dosomething_mbp_update_7008() {
+  module_enable('mailsystem', 'mandrill');
+  $mailsystem_keys = array(
```
mirie, line 143: Do we need to have these settings set in the install hook as well?

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
```diff
@@ -135,9 +135,68 @@ function dosomething_mbp_update_7007() {
...
+    'default-system' => 'MandrillMailSystem',
+    'mandrill_user_register' => 'MandrillMailSystem',
+    'mandrill_user_password' => 'MandrillMailSystem',
+    'mandrill_campaign_signup' => 'MandrillMailSystem',
+    'mandrill_campaign_group_signup' => 'MandrillMailSystem',
+    'mandrill_campaign_reportback' => 'MandrillMailSystem',
+  );
+  variable_set('mail_system', $mailsystem_keys);
+  variable_set('mandrill_api_key', 'Zqk_59h4hMerx3ZlEvC13A');
```
mirie: I'm not sure if we want to have the API key publicly available and set. What are your thoughts?

crittermike: We could always just manually set it after deployment. Unless you all have another method for handling API keys (env vars perhaps)?

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
crittermike: I was basing this on the assumption that this module would never be installed again (on a new site or an environment where it's not already installed), as opposed to updated. Is that not a safe assumption?

mirie: Yeah, I think that's a valid assumption - just wanted to ask and see what you thought / if it was unnecessary.

lib/modules/dosomething/dosomething_mbp/dosomething_mbp.install
mirie: We're deploying this to Pantheon, right? Do they have anything like that available? Otherwise, yeah, I guess we could set this manually. If it's set manually after this is deployed, would anything in this update fail?

crittermike: Pantheon supports https://github.com/pantheon-systems/terminus-secrets-plugin which writes to a `secrets.json` file that you can read from, as well as https://pantheon.io/docs/guides/lockr/ which requires 2 new modules so it's probably overkill. 

I think just manual setting is probably fine though. Nothing will fail with the update - emails just won't send until we add the API key. That sound OK?

mirie: Sounds good to me! I'll generate a new API key for Mandrill for the Indonesia site now.

crittermike:
Cool cool, I just removed that from the PR.

mirie (approved):

mirie:
Okay it looks like I actually have to delete and recreate the key, so I'm going to do that instead now.... Done!

PR #7567, opened 2019-03-02
Drupal Security Updates for DS Indonesia

mirie:
#### What's this PR do?
Well, I went to update the services module to 3.23 when I realized that I neglected to update the `intl-sites` branch with the last set of changes. So this PR has the changes from before the final Pantheon switch and also the services module update.

Updates are as follows:
- Update Drupal Core to 7.64
- Fixed install hook for `dosomething_mbp.install`
- Other security module updates: blockreference, entityconnect, google_analytics, link

Also updates the docksal command for pushing to Pantheon. There were some missing files that weren't being copied over to the temporary pantheon build directory from drops 7.

### DIFF
--- a/.docksal/commands/push-to-pantheon
+++ b/.docksal/commands/push-to-pantheon
@@ -19,6 +19,12 @@ rm -rf pantheon-codebase/*
 cp -r web/* pantheon-codebase
 
 git clone git@github.com:pantheon-systems/drops-7.git
+cp -r drops-7/includes pantheon-codebase/
+cp -r drops-7/modules pantheon-codebase/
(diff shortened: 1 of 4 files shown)


mshmsh5000 (approved):
One question: The `drops-7` repo [specifies PHP 7.2 as default](https://github.com/pantheon-systems/drops-7/blob/default/pantheon.upstream.yml#L6), but I've seen issues with 7.2 on the US site—in fact, recently saw issues with 7.1 as well. What is Indo running on?

mirie:
@mshmsh5000 Mike C. ran into the same problem and had to downgrade back to 7.1 too

